PR #2895, opened 2017-09-25
fix: do not print stack trace for invalid option

### LINKED ISSUE #2891: Stack trace is printed when providing invalid option choice
```
$ prettier --version
1.7.0

$ echo 'test' | prettier --parser foo
stdin: Error: Couldn't resolve parser "foo"
    at resolveParseFunction (/home/lydell/bugs/prettier/node_modules/prettier/index.js:3774:13)
    at Object.parse (/home/lydell/bugs/prettier/node_modules/prettier/index.js:3782:25)
    at formatWithCursor (/home/lydell/bugs/prettier/node_modules/prettier/index.js:21730:22)
    at Object.formatWithCursor (/home/lydell/bugs/prettier/node_modules/prettier/index.js:21991:12)
    at format (/home/lydell/bugs/prettier/node_modules/prettier/bin/prettier.js:16346:21)
    at getStream.then.input (/home/lydell/bugs/prettier/node_modules/prettier/bin/prettier.js:16411:19)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

ikatyang:
Fixes #2891

before/after

image

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -1,0 +1,5 @@
+"use strict";
+
+class ConfigError extends Error {}
+
+module.exports = { ConfigError };
--- a/src/parser.js
+++ b/src/parser.js
@@ -38,7 +38,7 @@ function resolveParseFunction(opts) {
       return eval("require")(path.resolve(process.cwd(), opts.parser));
     } catch (err) {
       /* istanbul ignore next */
-      throw new Error(`Couldn't resolve parser "${opts.parser}"`);
+      throw `Couldn't resolve parser "${opts.parser}"`;
     }
   }
   /* istanbul ignore next */
--- a/src/parser.js
+++ b/src/parser.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const path = require("path");
+const ConfigError = require("./errors").ConfigError;
 
 const parsers = {
   get flow() {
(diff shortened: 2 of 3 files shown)


src/parser.js
```diff
@@ -38,7 +38,7 @@ function resolveParseFunction(opts) {
       return eval("require")(path.resolve(process.cwd(), opts.parser));
     } catch (err) {
       /* istanbul ignore next */
-      throw new Error(`Couldn't resolve parser "${opts.parser}"`);
+      throw `Couldn't resolve parser "${opts.parser}"`;
```
lydell: This works, but is it the correct fix? Do we want to catch errors somewhere instead?

lydell:
Nice!

There are two occurrences of throwing strings in cli-util.js as well, if you feel like looking if we can use a custom error there as well. :)

```
src/cli-util.js
117:      throw "prettier(input) !== prettier(prettier(input))\n" + diff(pp, pppp);
128:        throw "ast(input) !== ast(prettier(input))\n" +
```

src/errors.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
+
+class ConfigError extends Error {}
```
azz: I seem to recall there being issues in some engines with subclassing builtins but I can't remember if it affects `Error`...

lydell: Are you perhaps thinking about that it used to be more or less impossible to subclass builtins before ES6 (and that it therefore can’t be transpiled correctly)?

azz: Ah, yeah that could be it. If someone was to downlevel compile this to es5 for browser use will it still work?

I've always done it this way. Avoids nasty `instanceof` problems as well.

```js
function createConfigError(message) {
  const error = new Error(message);
  error.type = "ConfigError";
  return error;
}

function isConfigError(error) {
  return error && error.type === "ConfigError";
}
```

lydell: Extending `Error` seems to work flawlessly in Node.js 4+. Let’s stick with that until somebody brings up an issue.

PR #2627, opened 2017-08-15
Normalize numbers in CSS like JS

lydell:
Unlike JS, `1.0` is also normalized to `1`.

PR #2903, opened 2017-09-27
Don't use ANSI codes if stdout isn't a TTY

Narigo:
I've tried to pipe the output of `--list-differences` into `xargs git add`, but due to the progress information, the stdout is not really usable.

This commit could fix that by suppressing output information. I think it might make sense to change `process.stdout`/`console.log` to `process.stderr`/`console.error` in case of errors, but I did not want to put multiple things into one PR.

Doing `yarn` did not work for me as it currently cannot find the Git commit of `postcss-values-parser` referenced in the `package.json`. I've installed the `latest` version of it to get it to run, but this is probably the cause why the `css_numbers/jsfmt.spec.js` test fails for me locally.

**Edit:** Forgot `yarn lint` initially, which is now fixed :)

### DIFF
--- a/tests_integration/__tests__/__snapshots__/piped-output.js.snap
+++ b/tests_integration/__tests__/__snapshots__/piped-output.js.snap
@@ -1,15 +1,15 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`no file diffs with --list-different + formated file 1`] = `Array []`;
+exports[`no file diffs with --list-different + formatted file 1`] = `Array []`;
 
-exports[`no file diffs with --list-different + formated file 2`] = `Array []`;
+exports[`no file diffs with --list-different + formatted file 2`] = `Array []`;
 
-exports[`output with --list-different + unformated differs when piped 1`] = `
+exports[`output with --list-different + unformatted differs when piped 1`] = `
 "unformated.js[2K[1Gunformated.js
 "
 `;
 
-exports[`output with --list-different + unformated differs when piped 2`] = `
+exports[`output with --list-different + unformatted differs when piped 2`] = `
 "unformated.js
 "
 `;
--- a/tests_integration/__tests__/__snapshots__/piped-output.js.snap
+++ b/tests_integration/__tests__/__snapshots__/piped-output.js.snap
@@ -1,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`no file diffs with --list-different + formated file 1`] = `Array []`;
+
+exports[`no file diffs with --list-different + formated file 2`] = `Array []`;
+
+exports[`output with --list-different + unformated differs when piped 1`] = `
+"unformated.js[2K[1Gunformated.js
+"
+`;
+
+exports[`output with --list-different + unformated differs when piped 2`] = `
+"unformated.js
+"
+`;
--- a/tests_integration/__tests__/piped-output.js
+++ b/tests_integration/__tests__/piped-output.js
@@ -1,0 +1,50 @@
+"use strict";
+
+const runPrettier = require("../runPrettier");
+
+test("output with --list-different + unformated differs when piped", () => {
+  const result0 = runPrettier(
+    "cli/write",
+    ["--write", "--list-different", "unformated.js"],
+    { stdoutIsTTY: true }
+  );
+
+  const result1 = runPrettier(
+    "cli/write",
+    ["--write", "--list-different", "unformated.js"],
+    { stdoutIsTTY: false }
+  );
+
+  expect(result0.stdout).toMatchSnapshot();
+  expect(result0.status).toEqual(1);
+
+  expect(result1.stdout).toMatchSnapshot();
+  expect(result1.status).toEqual(1);
+
+  expect(result0.stdout).not.toEqual(result1.stdout);
+  expect(result0.stdout.length).toBeGreaterThan(result1.stdout.length);
+  expect(result0.write).toEqual(result1.write);
+});
+
+test("no file diffs with --list-different + formated file", () => {
+  const result0 = runPrettier(
+    "cli/write",
+    ["--write", "--list-different", "formated.js"],
+    { stdoutIsTTY: true }
+  );
+  const result1 = runPrettier(
+    "cli/write",
+    ["--write", "--list-different", "formated.js"],
+    { stdoutIsTTY: false }
+  );
+
+  expect(result0.write).toMatchSnapshot();
+  expect(result0.status).toEqual(0);
+
+  expect(result1.write).toMatchSnapshot();
+  expect(result1.status).toEqual(0);
+
+  expect(result0.stdout).not.toEqual(result1.stdout);
+  expect(result0.stdout.length).toBeGreaterThan(result1.stdout.length);
+  expect(result0.write).toEqual(result1.write);
+});
--- a/tests_integration/__tests__/piped-output.js
+++ b/tests_integration/__tests__/piped-output.js
@@ -2,16 +2,16 @@
 
 const runPrettier = require("../runPrettier");
 
-test("output with --list-different + unformated differs when piped", () => {
+test("output with --list-different + unformatted differs when piped", () => {
   const result0 = runPrettier(
     "cli/write",
-    ["--write", "--list-different", "unformated.js"],
+    ["--write", "--list-different", "--no-color", "unformated.js"],
     { stdoutIsTTY: true }
   );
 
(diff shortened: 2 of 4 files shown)


lydell:
About the `yarn install` error – sorry, that was by bad. It should work now again. See https://github.com/prettier/prettier/pull/2627#issuecomment-332411202

suchipi:
I'm concerned that printing to stderr might flag the run as an error in some CI servers...

Also, I personally prefer `--no-progress` over `--silent`

lydell:
Warnings are usually printed to stderr, and I have not seen CI failures because of warnings. As far as I know, only the exit code matters.

Narigo:
`--no-progress` is better, yes. Could also be `--quiet`. Just tell me what you prefer, if you think adding this feature makes sense at all and I will change it.

azz:
Can we use [`process.env.isTTY`](https://nodejs.org/api/tty.html#tty_tty) instead of a new option?

Narigo:
Will change it to `process.stdout.isTTY` and remove the option.

Narigo:
Should be updated now and use `process.stdout.isTTY`.

tests_integration/__tests__/piped-output.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
+
+const runPrettier = require("../runPrettier");
+
+test("output with --list-different + unformated differs when piped", () => {
```
lydell: Typo: "formated" → "formatted"

There are more instances of this below, including some file names.

tests_integration/__tests__/__snapshots__/piped-output.js.snap
```diff
@@ -0,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`no file diffs with --list-different + formated file 1`] = `Array []`;
+
+exports[`no file diffs with --list-different + formated file 2`] = `Array []`;
+
+exports[`output with --list-different + unformated differs when piped 1`] = `
+"unformated.js[2K[1Gunformated.js
```
lydell: Can you add `--no-color` to these tests? We've had CI failures because of this before. (Also, the tests become more readable.)

tests_integration/__tests__/piped-output.js
Narigo: I've copied the texts from the `write.js` tests. Shall I rename the files in there as well or do you want to just have the string typos fixed and the file changes in another PR?

lydell: I see. Let's do it in another PR (if you feel like it).

tests_integration/__tests__/__snapshots__/piped-output.js.snap
```diff
@@ -0,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`no file diffs with --list-different + formatted file 1`] = `Array []`;
+
+exports[`no file diffs with --list-different + formatted file 2`] = `Array []`;
+
+exports[`output with --list-different + unformatted differs when piped 1`] = `
+"unformated.js[2K[1Gunformated.js
```
lydell: I thought this was a color code. Oh well, `--no-color` doesn't hurt anyway. I guess it's from moving the cursor.

Narigo: Yes, and that's exactly the problem I had when piping the output: The next command in the pipe complained about that... :)

Narigo:
I'd rename `formated.js` -> `formatted.js` and `unformated.js` -> `unformatted.js`, but I'd like to wait until this is merged. Otherwise I'd have to change the other things first, rebase here and ... I admit I just finished the renaming on this branch and am now too lazy to do it the other way around... ;)

tests_integration/runPrettier.js
```diff
@@ -44,9 +44,11 @@ function runPrettier(dir, args, options) {
   const originalArgv = process.argv;
   const originalExitCode = process.exitCode;
   const originalStdinIsTTY = process.stdin.isTTY;
+  const originalStdoutIsTTY = process.stdout.isTTY;
 
   process.chdir(normalizeDir(dir));
   process.stdin.isTTY = !!options.isTTY;
+  process.stdout.isTTY = !!options.stdoutIsTTY;
```
suchipi: I think we can just use the existing `isTTY` here; I don't know of any real-world situations where stdin is a tty and stdout isn't (or vice-versa)

mitermayer:
Would this issue also solve something like this https://github.com/prettier/vim-prettier/issues/59? 

Having issues now with prettier 1.7.1 since we are printting the message

> --parser with value postcss is deprecated. Prettier now treats it as: --parser=css.

Sad :(

azz:
> Would this issue also solve something like this prettier/vim-prettier#59?


I don't think so. That warning should be printed with `console.warn` which should go to `stderr`. @ikatyang any ideas?

ikatyang:
@mitermayer 

If `vim-prettier` somehow can't understand the difference between stdout/stderr, I'd recommend to write a wrapper script to execute prettier so that you can do whatever you want, e.g. hide warnings.

```js
const child_process = require("child_process");
const pathToPrettierCli = "node_modules/.bin/prettier";

const result = child_process.spawnSync(pathToPrettierCli, ["--parser", "postcss"], {
  encoding: "utf8",
  input: ".test { display: none; }"
});

console.log('stdout', result.stdout);
console.log('stderr', result.stderr);
console.log('status', result.status);
```

@azz 

Should we add some kind of logger system so that they can easily get what they want and we can also log some debug information so as to trace what the root cause is? e.g. [`--loglevel`](https://github.com/lerna/lerna#--loglevel-silenterrorwarnsuccessinfoverbosesilly).

azz:
Sounds good to me.

tests_integration/runPrettier.js
lydell: https://nodejs.org/api/process.html

> ```
> $ node -p "Boolean(process.stdin.isTTY)"
> true
> $ echo "foo" | node -p "Boolean(process.stdin.isTTY)"
> false
> $ node -p "Boolean(process.stdout.isTTY)"
> true
> $ node -p "Boolean(process.stdout.isTTY)" | cat
> false
> ```

I think it is correct to use `process.stdout.isTTY` here.

azz:
Renamed the PR and merged. Thanks!

azz:
This broke this build.

https://travis-ci.org/prettier/prettier/jobs/283636656

```
Summary of all failing tests
FAIL tests_integration/__tests__/piped-output.js
  ● output with --list-different + unformatted differs when piped
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - "unformated.jsated.js
    "
    + undefined
      
      at Object.<anonymous>.test (tests_integration/__tests__/piped-output.js:18:26)
          at Promise (<anonymous>)
          at <anonymous>
  ● no file diffs with --list-different + formatted file
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Array []
    + undefined
      
      at Object.<anonymous>.test (tests_integration/__tests__/piped-output.js:41:25)
          at Promise (<anonymous>)
          at <anonymous>
FAIL tests_integration/__tests__/write.js
  ● write file with --write + unformated file › (stdout)
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Snapshot
    + Received
    
    - "unformated.js%s %dms
    + "%s %dms
      "
      
      at Object.test (tests_integration/runPrettier.js:106:25)
          at Promise (<anonymous>)
          at <anonymous>
  ● do not write file with --write + formated file › (stdout)
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Snapshot
    + Received
    
    - "formated.js%s %dms
    + "%s %dms
      "
      
      at Object.test (tests_integration/runPrettier.js:106:25)
          at Promise (<anonymous>)
          at <anonymous>
  ● do not write file with --write + invalid file › (stdout)
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Snapshot
    + Received
    
    - "invalid.js
      "
    + "
      
      at Object.test (tests_integration/runPrettier.js:106:25)
          at Promise (<anonymous>)
          at <anonymous>
```

Narigo:
I have no idea. After fetching again, I see that my branch diverged from yours. Probably some git issue?
Did the behaviour on `--write` change in the meantime? I've copied most of the tests from the write integration tests.

azz:
On mobile at the moment but maybe #2957 or #2951

Issue #2910, opened 2017-09-27
Long lines never reflow in test declarations.


dreid:
This seems to be related to https://github.com/prettier/prettier/pull/2877

**Prettier 1.7.1**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22describe(%5C%22foo%5C%22%2C%20()%20%3D%3E%20%7B%5Cn%20%20const%20MINIMAL_STATE%20%3D%20%7B%20active_features%3A%20%5B%5D%2C%20csrf%3A%20%7B%20csrf_token_value%3A%20%5C%22csrf-token-value%5C%22%20%7D%2C%20successful_campaigns%3A%20%5B%5D%2C%20user%3A%20null%20%7D%3B%5Cn%7D)%3B%5Cn%5Cnit(%5C%22bar%5C%22%2C%20()%20%3D%3E%20%7B%5Cn%20%20%20%20const%20MINIMAL_STATE%20%3D%20%7B%20active_features%3A%20%5B%5D%2C%20csrf%3A%20%7B%20csrf_token_value%3A%20%5C%22csrf-token-value%5C%22%20%7D%2C%20successful_campaigns%3A%20%5B%5D%2C%20user%3A%20null%20%7D%3B%5Cn%7D)%3B%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22flow%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)
```sh
--parser flow
```

**Input:**
```jsx
describe("foo", () => {
  const MINIMAL_STATE = { active_features: [], csrf: { csrf_token_value: "csrf-token-value" }, successful_campaigns: [], user: null };
});

it("bar", () => {
    const MINIMAL_STATE = { active_features: [], csrf: { csrf_token_value: "csrf-token-value" }, successful_campaigns: [], user: null };
});
```

**Output:**
```jsx
describe("foo", () => {
  const MINIMAL_STATE = { active_features: [], csrf: { csrf_token_value: "csrf-token-value" }, successful_campaigns: [], user: null };
});

[...]

PR #2877, opened 2017-09-21
Fix line break in test declarations with 2nd argument as a function

duailibe:
Fixes #1473

Basically this prevents breaking a test declaration when the 2nd argument is a function declaration:

```js
// current output
test("long string", function(
  done
) {
  // body
});

// with this PR
test("long string", function(done) {
  // body
});
```

I wasn't sure how recommended it is to use `removeLines` but since this case was quite specific, I didn't think it would be much trouble. Let me know if I should solve this some other way!

PR #2913, opened 2017-09-27
Revert "Fix line break in test declarations with 2nd argument as a function"

lydell:
Reverts prettier/prettier#2877

This caused #2910


lydell:
I tried to use the "Revert" button in the PR here on github, but that seems to have reverted way too much. :(

Issue #2291, opened 2017-06-27
Removes linebreaks from styled-components tagged template literals in an odd way


dgcoffman:
Using prettier 1.5.0.

before
``` javascript
const Direction = styled.span`
  ${({ up }) => up && `color: ${color.positive};`}
  ${({ down }) => down && `color: ${color.negative};`}
`;
```

after
``` javascript
const Direction = styled.span`
  ${({ up }) => up && `color: ${color.positive};`} ${({ down }) =>
       down && `color: ${color.negative};`};
`;
```

I'd expect it to preserve the existing linebreak.

vjeux:
Thanks, i'll look into fixing it :)

vjeux:
this is rewritten as

```js
  @prettier-placeholder
  @prettier-placeholder
```

which is being parsed as a media query. I managed to get it working for this case but then there are things like

```js
  @prettier-placeholder
  color: red;
  @prettier-placeholder
```

which are getting busted. If you add a `;` at the end of the expressions, then it prints as is:

```js
const Direction = styled.span`
  ${({ up }) => up && `color: ${color.positive};`};
  ${({ down }) => down && `color: ${color.negative};`};
`;
```

I'm not sure we can make it work without unfortunately :(

cc @pomber in case you have ideas.

dgcoffman:
That's cool -- I can add semicolons. You'll likely hear about this a bit since this pattern is common when using styled-components.

dgcoffman:
Another styled-components thing:
``` javascript
const Dropdown = styled.div`
  position: relative;
`;
```
becomes
``` javascript
const Dropdown = styled.div`position: relative;`;
```

which causes:
"block-closing-brace-space-after": "always-single-line",
"block-closing-brace-space-before": "always-single-line",
"block-opening-brace-space-after": "always-single-line",
"block-opening-brace-space-before": "always-single-line",

stylelint rules to fail. We'd prefer always line-breaking.

pomber:
First issue looks hard, maybe we could always preserve linebreaks after expressions in CSS template literals? No idea how to do it, but I know it is being done for [some cases](https://github.com/prettier/prettier/blob/master/Rationale.md#multi-line-objects).  

[...]

Issue #1893, opened 2017-06-02
Ugly line break in a template string


clebert:
Not really an error, but in my opinion, ugly.

Input:

```js
const description =
  `The value of the ${cssName} css of the ${this._name} element`;
```

Output:

```js
const description = `The value of the ${cssName} css of the ${this
  ._name} element`;
```

Expected:

```js
const description = `The value of the ${cssName} css of the ${
  this._name
} element`;
```

Or something like that...

Skywalker13:
Maybe templates should never be wrapped because it looks ugly most of time.

This input
```js
const description =
  `The value of the ${cssName} css of the ${this._name} element`;
```

looks better that
```js
const description = `The value of the ${cssName} css of the ${
  this._name
} element`;
```

even if it's longer than 80 cols

ndresx:
I agree @Skywalker13, such strings should never wrap.

O4epegb:
Do i understand correctly, that wrapping also adds new line characters? If so, then it can break some formatting when someone not expecting additional newline there, doesnt it?

Skywalker13:
@O4epegb not here, because the new line is added in the JS code `${... HERE ...}`, not in the string.

O4epegb:
@Skywalker13 Yes, i mean otherwise, even if it breaks correctly inside the string, it will add new line character, right?

Skywalker13:
Yes, in the source code; it's the case between `v1.3.x` and `v1.4.x`. It was looking better with `v1.3.x`.

Widdershin:
I used `git bisect` to determine the commit where this behaviour changed:

commit 374a68201c966ee9064a1173f5dfeb37980cfbe3 (pr #1664)
Author: Joseph Frazier <1212jtraceur@gmail.com>
Date:   Mon May 22 12:00:01 2017 -0400

    Don't removeLines() from TemplateLiteral quasis (#1664)

    * Add tests for newlines in template literals

    These are related to the following issues:

    * https://github.com/prettier/prettier/issues/1662
    * https://github.com/prettier/prettier/issues/821
    * https://github.com/prettier/prettier/issues/1183
    * https://github.com/prettier/prettier/issues/1626

    * Don't removeLines() from TemplateLiteral quasis

    This fixes https://github.com/prettier/prettier/issues/1662
    and fixes https://github.com/prettier/prettier/issues/1626

    It also affects https://github.com/prettier/prettier/issues/821 and
    https://github.com/prettier/prettier/issues/1183

[...]

PR #2926, opened 2017-09-28
Force template literals to break after ` for styled-components

duailibe:
There isn't a specific issue for this, but it was mentioned in https://github.com/prettier/prettier/issues/2291#issuecomment-311489067 and https://github.com/prettier/prettier/issues/2350#issuecomment-328346243.

```js
// input
const Dropdown = styled.div`position: relative`;

// output
const Dropdown = styled.div`
  position: relative;
`;
```

### DIFF
--- a/src/multiparser.js
+++ b/src/multiparser.js
@@ -42,10 +42,8 @@ function fromBabylonFlowOrTypeScript(path) {
   switch (node.type) {
     case "TemplateLiteral": {
       const isCss = [isStyledJsx, isStyledComponents].some(isIt => isIt(path));
-      const isEmpty =
-        node.quasis.length === 1 && !node.quasis[0].value.raw.trim();
 
-      if (isCss && !isEmpty) {
+      if (isCss) {
         // Get full template literal with expressions replaced by placeholders
         const rawQuasis = node.quasis.map(q => q.value.raw);
         const text = rawQuasis.join("@prettier-placeholder");
@@ -179,6 +177,13 @@ function fromHtmlParser2(path, options) {
 
 function transformCssDoc(quasisDoc, parent) {
   const parentNode = parent.path.getValue();
+
+  const isEmpty =
+    parentNode.quasis.length === 1 && !parentNode.quasis[0].value.raw.trim();
+  if (isEmpty) {
+    return "``";
+  }
+
   const expressionDocs = parentNode.expressions
     ? parent.path.map(parent.print, "expressions")
     : [];
--- a/src/multiparser.js
+++ b/src/multiparser.js
@@ -42,8 +42,10 @@ function fromBabylonFlowOrTypeScript(path) {
   switch (node.type) {
     case "TemplateLiteral": {
       const isCss = [isStyledJsx, isStyledComponents].some(isIt => isIt(path));
+      const isEmpty =
+        node.quasis.length === 1 && !node.quasis[0].value.raw.trim();
 
-      if (isCss) {
+      if (isCss && !isEmpty) {
         // Get full template literal with expressions replaced by placeholders
         const rawQuasis = node.quasis.map(q => q.value.raw);
         const text = rawQuasis.join("@prettier-placeholder");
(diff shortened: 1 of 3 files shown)


duailibe:
Ok, I ran this version on https://github.com/hawkrives/gobbldygook. The diff looks good, nothing in special that came from this commit, except for this case:

```js
const ListElement = styled.li``;
# gets converted to
const ListElement = styled.li`

`;
```
I'll try to fix it

hawkrives:
… I'll have to go update gobbldygook to prettier 1.7 today, then. Glad it helped!

suchipi:
@hawkrives 1.7.2 doesn't have these changes in it yet- you could use the github version if you wanted once this is merged, though

duailibe:
@lydell I updated the PR to account for the empty case.

@hawkrives there are quite a few differences in formatting. Most of them are related to expressions in template literals, like #1893

The problem described in #2291 OP, for example, is one of them. You'd have to include `;` in them.

hawkrives:
@duailibe Those are the two things I've noticed, as well.

@suchipi I just haven't gotten it updated past prettier@1.4 yet… don't want to clutter the diffs with the excellent JSX changes that we've had since then.

src/multiparser.js
```diff
@@ -46,12 +46,14 @@ function fromBabylonFlowOrTypeScript(path) {
       if (isCss) {
         // Get full template literal with expressions replaced by placeholders
         const rawQuasis = node.quasis.map(q => q.value.raw);
-        const text = rawQuasis.join("@prettier-placeholder");
-        return {
-          options: { parser: "css" },
-          transformDoc: transformCssDoc,
-          text: text
-        };
+        if (rawQuasis.some(q => q)) {
```
lydell: In `` `${a}${b}${c}` `` all quasis will be the empty string (falsy), right? What happens then?

duailibe: Yeah it prints `${a}${b}${c}` even without this change.

Let me try to fix it!

duailibe: @lydell I'm not sure what we should do on this case: [playground](https://prettier.io/playground/#%7B%22content%22%3A%22const%20a%20%3D%20styled.div%60%24%7Ba%7D%24%7Bb%7D%24%7Bc%7D%60%3B%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)

lydell: I don't really care about that case, I was just confused by this `if` statement. I like your new commit where you explicitly check for a completely empty template literal. :+1:

src/multiparser.js
```diff
@@ -42,8 +42,9 @@ function fromBabylonFlowOrTypeScript(path) {
   switch (node.type) {
     case "TemplateLiteral": {
       const isCss = [isStyledJsx, isStyledComponents].some(isIt => isIt(path));
+      const isEmpty = node.quasis.length === 1 && !node.quasis[0].value.raw;
```
lydell: Actually, I think we should add `.trim()` so that ``styled.div`  ` `` is formatted to ```styled.div`` ``` and not broken across several lines.

duailibe: @lydell wouldn't that change the AST?

lydell: No, why would it?

suchipi: It's changing the contents of the literal; the value of the first quasis changes from `" "` to `""`

suchipi: (However, technically, adding newlines also changes the value of the literal. It just happens to not matter for styled-components. So changing it here might be fine)

lydell: Yes, since we know that the contents is CSS we are free to change it however we want – which we also already do.

duailibe: You're right, that's fair. That means the solution I proposed won't work, since it skips the CSS formatting at all, leaving it up to the default printer.

duailibe:
Let me know what you think now!

lydell:
Thanks!

PR #2930, opened 2017-09-28
Try to fix cosmiconfig in the built version of Prettier

### LINKED ISSUE #2924: 1.7.2 js module config fails to load
**Prettier 1.7.2**

```
# myConfig.js
module.exports = { semi: true };
```
```
$ prettier --config ./myConfig.js index.js
Error: Invalid configuration file.
Failed to parse "[..]/myConfig.js" as JSON, JS, or YAML.
```

**Expected behavior:**
Should load js module as config.

lydell:
Fixes  #2924 (hopefully). Not sure if this is the correct fix, but it seems to work. Please help me test! We should also write integration tests that catch the error...

### DIFF
--- a/scripts/build/rollup.bin.config.js
+++ b/scripts/build/rollup.bin.config.js
@@ -11,7 +11,14 @@ export default Object.assign(baseConfig, {
   banner: "#!/usr/bin/env node",
   plugins: [
     replace({
-      "#!/usr/bin/env node": ""
+      "#!/usr/bin/env node": "",
+      // The require-from-string module (a dependency of cosmiconfig) assumes
+      // that `module.parent` exists, but it only does for `require`:ed modules.
+      // Usually, require-from-string is _always_ `require`:ed, but when bundled
+      // with rollup the module is turned into a plain function located directly
+      // in bin/prettier.js so `module.parent` does not exist. Defaulting to
+      // `module` instead seems to work.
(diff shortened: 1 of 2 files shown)


ikatyang:
I can't reproduce the issue in this PR, it seems worked.

Not sure how to write tests for it, can we just run tests for bundled version before releasing?

---

Unrelated question: our CI's `NODE_ENV=production` job is always failed, what does it want to test for?

azz:
I think it's supposed to run the tests on `dist/`. cc @josephfrazier 

Speculation: I think it is failing because `yarn --production=false` doesn't do what it did in yarn v0. Maybe switching to `NODE_ENV=development yarn` or `yarn --production=false install` will fix it.

lydell:
When running the build scripts, it says this at the end:

```
How to test against dist:
  1) NODE_ENV=production yarn test
```

I've tried that, and everything passes. It would be nice to construct a test that would fail there.

Btw, I also need to test this with eslint-plugin-prettier. Might be more rollup:ed files to tweak.

lydell:
I've tested with eslint-plugin-prettier now. I've force-pushed a fix to also do the module.parent thing in index.js.

lydell:
Turns out we don't have to write any new tests: #2932

lydell:
It can’t make things worse at least.

Issue #2825, opened 2017-09-15
Generate the options part of README.md and options.md from code


ikatyang:
From https://github.com/prettier/prettier/pull/2767#issuecomment-327744705:

> Probably something for a future PR, but I've been thinking about generating the options part of README.md and options.md (which is probably out of date) from code.

Possible solution: 
- [markdown-magic](https://github.com/DavidWells/markdown-magic)

PR #2937, opened 2017-09-30
[WIP] Generate options.md and README.md#Options

azz:
This is super-early progress on #2825.

I'm open to ideas on where to put all the existing documentation for each option. Maybe we could add a JSDoc comment to each option in `cli-constant.js` and use that? We could keep it simple and use template strings but that's code that we ship. We could also put it in a separate file as yml.


src/cli-constant.js
```diff
@@ -186,15 +186,24 @@ const detailedOptions = normalizeDetailedOptions({
...
-      "babylon",
-      "typescript",
-      "css",
+      { value: "flow", description: "https://github.com/facebook/flow" },
+      { value: "babylon", description: "https://github.com/babel/babylon" },
+      {
+        value: "typescript",
+        description: "https://github.com/eslint/typescript-eslint-parser"
+      },
+      { value: "css", description: "https://github.com/postcss/postcss" },
```
lydell: This description is incorrect. We don’t use the postcss parser (and never have). The "css" option tries to use postcss-less first, and if that fails, postcss-scss (unless a regex says to flip the order).

scripts/markdown-magic-prettier-options.js
```diff
@@ -0,0 +1,81 @@
...
+  ].join(" | ");
+
+  return [header, description, choices, [tableHeader, tableRow].join("\n")]
+    .filter(Boolean)
+    .join("\n\n");
+}
+
+// https://github.com/chjj/marked/issues/285
+function backtick(string) {
+  if (string.indexOf("|") > -1) {
```
lydell: Pro-tip: There’s `String.prototype.includes` (and, yes, it is supported on Node.js 4).

src/cli-constant.js
azz: Oops, thanks for catching that.

scripts/markdown-magic-prettier-options.js
azz: Every now and then I get it confused with `Array.prototype.includes` which isn't supported by node v4.

lydell: So do I!

ikatyang: Always get confused with these two:

- [`String.prototype.includes`](http://node.green/#ES2015-built-in-extensions-String-prototype-methods-String-prototype-includes) is supported
- [`Array.prototype.includes`](http://node.green/#ES2016-features-Array-prototype-includes-Array-prototype-includes) is not supported

README.md
```diff
@@ -594,7 +594,7 @@ Valid options:
 * `flow`        https://github.com/facebook/flow
 * `babylon`     https://github.com/babel/babylon
 * `typescript`  https://github.com/eslint/typescript-eslint-parser
-* `css`         https://github.com/postcss/postcss
+* `css`         scientifically try both postcss-less and postcss-scss
```
lydell: I love this new description! 😆

azz:
I implemented docblock parsing... Not because it is necessarily a good idea but I felt like the challenge. Now the source of truth for documentation is in one place.

src/options-definitions.js
```diff
@@ -0,0 +1,366 @@
...
+ *     getter?: (value: any, argv: any) => any;
+ *   }
+ * }
+ *
+ * Note: The options below are sorted alphabetically.
+ */
+module.exports = {
+  "bracket-spacing": {
+    type: "boolean",
+    category: categories.FORMAT,
```
lydell: Early on in one of the CLI refactor PRs, the categories used to be specified as string literals each time, such as `category: "format"`. Then I changed to `category: CATEGORY_FORMAT`. I thought about making a `categories` object like you have here, but settled on all those separate variables because then ESLint catches typos for us. Now, that “security” is lost. But I guess the tests would catch an unexpected "Undefined options:" line in the `--help` output.

README.md
```diff
@@ -602,7 +602,7 @@ Which parser to use.
 Both the `babylon` and `flow` parsers support the same set of JavaScript
 features (including Flow). Prettier automatically infers the parser from the
 input file path, so you shouldn't have to change this setting.
-
+
```
lydell: Not a big deal, but if you manage to figure out where these trailing spaces come from it would be nice if we could remove them :)

src/options-definitions.js
azz: Ooh that's a good point. I only moved it because it is depended on in two files, and it would have been a circular dependency otherwise. But yes, the tests should catch this.

README.md
azz: That's just my naive docblock regex, I realized that `require('jest-docblock').parseWithComments()` does a `.trim()` which removes wanted left-padding. I'll probably do a PR there to fix that.

azz: @lydell you might appreciate this regex fu! https://github.com/facebook/jest/pull/4576/files#diff-b5f6adcc40fabaaa89acc7ec660da205

azz:
I'm thinking this PR is being a bit too smart and maybe we should just use markdown-magic to sync across the contents of options.md to the main README.md.

azz:
See #3116

PR #2942, opened 2017-09-30
Fix production tests 2

lydell:


tests_integration/runPrettier.js
```diff
@@ -50,12 +57,14 @@ function runPrettier(dir, args, options) {
...
 
   jest.resetModules();
-  jest.setMock("get-stream", () => ({
-    then: handler => handler(options.input || "")
-  }));
+  jest
+    .spyOn(require(prettierApi).__debug, "getStream")
+    .mockImplementation(() => ({
+      then: handler => handler(options.input || "")
+    }));
```
lydell: @ikatyang All the other `.spyOn(...).mockImplementation(...)` are _above_ the `jest.resetModules()` call. Does this matter in any way?

ikatyang: The `jest.resetModules()` is for `require`s, it should not matter.

lydell:
Unless somebody has any notes, I think this is good to go.

PR #2844, opened 2017-09-16
Split Less and SCSS parsing into different parsers

lydell:
Now, .less files are always parsed with postcss-less, and .scss files
areare always parsed with postcss-scss. This:

- Is less hacky.
- Is meant to avoid issues like #2829.
- Is probably more performant.

`parser: "postcss"` and `--parser postcss` continue to work like before:
First trying postcss-less, and if that fails, postcss-scss, unless a
regex says that we should try in the opposite order. The new values for
the parser option are "postcss-less" and "postcss-scss".

PR #11, opened 2017-01-10
Fix tests for Windows.

demoneaux:
~~https://github.com/demoneaux/prettier/commit/95a89e9cdf9a7eadbafad51f17e51c46ec1286c7~~ https://github.com/jlongster/prettier/pull/11/commits/258575af9ade68a28813ea79a829468b61c94f22 allows the tests to run on Windows.

~~https://github.com/demoneaux/prettier/commit/8e2eaccfc1c5c4c456d5787992d2d19b67e11b31~~ ~~https://github.com/jlongster/prettier/pull/11/commits/29334081d2f7cd0ffcd8535b145e36d2f266e3d4~~ https://github.com/jlongster/prettier/pull/11/commits/e72347e51b3cd9d1694462856fad0047c4c6b38f attempts to solve EOL problems with Windows. ~~but there are still some issues with that.~~

Closes #10.

PR #2943, opened 2017-09-30
feat: support markdown

### LINKED ISSUE #2444: Markdown integration
Hi everyone!

I've decided to dive into the Prettier source code with the intention of ultimately building full Markdown support into it. I can find my own way around the codebase OK for now, but I'd like to discuss here (given the truth of [this tweet](https://twitter.com/Vjeux/status/884533137891524608)) some fundamental issues as I encounter them, to increase the chances of this actually landing in Prettier at some point.

## Goal

A Markdown integration for Prettier, which, the way I envision it, involves formatting Markdown files and ideally also JS/CSS/etc code blocks within them.

**Q: Is such a Markdown integration even desired?**

## Basic approach

Using [`remark`](https://github.com/wooorm/remark/) as a parser and as a reference printer:

1. Take a direct dependency on [`remark-parse`](https://github.com/wooorm/remark/tree/master/packages/remark-parse) to implement `parser-remark.js` (this part is straightforward).
2. Fully port [`remark-stringify`](https://github.com/wooorm/remark/tree/master/packages/remark-stringify) to Prettier's formatting primitives (and coding conventions) to implement `printer-remark.js`.
3. Make stylistic tweaks to `printer-remark.js` once it has reached parity with `remark-stringify`.
4. _Multiparser??? - Haven't explored in any depth yet_

### Passing user options to the parser/printer

`remark` supports several flavours of Markdown and takes various options; would we expose any of this to the user? (How?) The alternative is to essentially commit to a single flavour of Markdown, but that may be _too_ opinionated to be viable.

### A new document type for blockquotes

If we want to print Markdown blockquotes the "standard" pretty way - with hard wrapping and a `>` marker at the beginning of each line:

```
> This is a blockquote. Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aliquam hendrerit mi posuere lectus.
> This is a blockquote. Lorem ipsum dolor sit amet, consectetuer adipiscing
> elit. Aliquam hendrerit mi posuere lectus.
```

Then AFAICT we need a new document type to represent it _and_ probably a different implementation of `indent` and `align` in `doc-printer.js` (supporting arbitrarily mixed indentation and non-space markers, including nested blockquotes etc). Can someone confirm this or point out something I've missed?

### Multiparser architecture

I have not begun to tackle this at all, but would appreciate any thoughts about it nonetheless. ~My current thinking is that we can extract parts of the existing JS-centric "multiparser" for reuse in a Markdown multiparser, and potentially uncover interesting generalisations.~ Just realised the existing multiparser also does CSS in HTML; it's probably already general enough.

--- 

Thanks for a fantastic tool and thanks in advance for any help on this!

ikatyang:
Main:

- Use [`remark-parse`](https://github.com/wooorm/remark/tree/master/packages/remark-parse) with [commonmark](https://github.com/wooorm/remark/tree/master/packages/remark-parse#optionscommonmark) option enabled ([GFM](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown)  is based on commonmark)
- [Commonmark Spec](http://spec.commonmark.org/0.25/)
- enforce
  - header: `#`-style
  - emphasis: `_`
  - strong: `**`
  - listItem: `-`(always), `+`(switch between two style for continuous lists)
  - listItem (ordered): `1.` (always), `1)`(switch between two style for continuous lists)
  - thematicBreak: `- - -`(always), `* * *` (in list)
  - indented code block: 4 whitespaces
  - fenced code block: ` ``` `(always), `~~~`(in js template)
  - table: leading/trailing`|` and align table cell separator and respect alignment
- `fill` word by word (whitespace-based).
- See the formatted part of the [snapshot file](https://github.com/ikatyang/prettier/blob/feat/markdown/tests/markdown/__snapshots__/jsfmt.spec.js.snap).
  ```js
  export[xxx] = `

  unformatted
  formatted

  `;
  ```
Side:

- extend `align`'s `n` to be string available. (use for `quoteblock`)

Question:

- Are the styles above ok?
- ~~Should name this parser `markdown` (current) or `remark`?~~ [Ans: `markdown`](https://github.com/prettier/prettier/pull/2943#issuecomment-333329728)

NOTE:

- the content of ~~`code`~~/`html`/`yaml` still remains the same.
- ~~does not support `break` (trailing two spaces).~~
- remark is available on [astexplorer](http://astexplorer.net/).

TODO:

- [x] `<!-- prettier-ignore -->`
- [x] format the content of `code`
- [x] ~~escape html entities (`<`, `>`, `&`)~~ html entities remain the same (`&lt;`, etc.).
- [x] pass the test case: https://raw.githubusercontent.com/adamschwartz/github-markdown-kitchen-sink/master/TEST.md
- [x] add a special case for `<!-- prettier-ignore -->` not to print additional line break
- [x] use `~~~`-style code block for markdown in template string
- [x] pass all the examples in the [spec](http://spec.commonmark.org/0.25/) using `AST_COMPARE` (`607/616`)
- [ ] figure out a better way to escape necessary characters
- anything else?

Closes #2444


src/printer-markdown.js
```diff
@@ -0,0 +1,434 @@
...
+        "```"
+      ]);
+    case "yaml":
+      return concat(["---", hardline, node.value, hardline, "---"]);
+    case "html":
+      return node.value;
+    case "list":
+      return printChildren(path, options, print, {
+        processor: (childPath, index) => {
+          const prefix = node.ordered ? `${node.start + index}. ` : "- ";
```
SimenB: why include the index? One of the things I like about markdown is that I can move around the entries in an ordered list without having to renumber them

(loving this PR btw, prettier keeps on making new stuff pretty :D)

ikatyang: Well, I think it looks better with the index and the index will be renumbered by Prettier, you don't have to edit it manually.

SimenB: so

```md
1. one
3. two
2. three
```

is changed to

```md
1. one
2. two
3. three
```
?

In that case I guess it's fine 🙂

ikatyang: Yes, only the initial number matters.

lydell:
Awesome!

> Are the styles above ok?

To me personally, yes. I just wonder if you picked your personal preferences, what you feel like is common, or did some kind of research?

> Should name this parser `markdown` (current) or `remark`?

`markdown`. I guess we kind of decided to name the `parser` values after language until we have a reason not to (like babylon vs flow) in #2844.

> the content of code/html/yaml still remains the same.

On purpose, or just because you haven't done so yet, or out of scope for this PR?

yarn-error.log
```diff
@@ -0,0 +1,4654 @@
+Arguments:
```
azz: You accidentally checked this in. Could you remove it and add `*.log` to `.gitignore`?

azz:
Wow, this is awesome!

ikatyang:
> I just wonder if you picked your personal preferences, what you feel like is common, or did some kind of research?

It's currently my personal preferences, I have no idea how to do research for it, [searching symbols on GitHub seems not working](https://github.com/search?l=&q=%22-+%22+language%3AMarkdown&ref=advsearch&type=Code&utf8=%E2%9C%93).

> On purpose, or just because you haven't done so yet, or out of scope for this PR?

- For `code`, I thought someone may not want to format their `code`s, and currently I haven't implemented the `<!-- prettier-ignore -->` thing, still investigating.

- For `html`, the AST for html in `remark` looks something like:

  ```js
  [
    { type: "html", value: "<a>" }
    { type: "text", value: "content" }
    { type: "html", value: "</a>" }
  ]
  ```

  Don't know how to group them properly, any idea?

- For `yaml`, we don't support it (yet?).

src/printer-markdown.js
azz: awesome!

azz:
Release notes are going to be amazing.

azz:
r.e. YAML, I did a quick search a while ago and only found https://github.com/mulesoft-labs/yaml-ast-parser which doesn't parse comments.

src/printer-markdown.js
```diff
@@ -458,4 +458,63 @@ function mapDoc(doc, callback) {
...
+  if (parser === null) {
+    return node.value;
+  }
+
+  const prettier = require("..");
+  const indentation = getIndentationLength(path);
+
+  const formatted = prettier
+    .format(
+      node.value,
```
ikatyang: This seems not working here, help wanted!

src/printer-markdown.js
```diff
@@ -458,4 +458,63 @@ function mapDoc(doc, callback) {
...
+
+function getIndentationLength(/* path */) {
+  // TODO:
+  return 0;
+}
+
+function getParserName(lang) {
+  switch (lang) {
+    case "js":
+    case "javascript":
```
azz: `case "jsx"`

src/printer-markdown.js
```diff
@@ -458,4 +458,63 @@ function mapDoc(doc, callback) {
...
+      return "babylon";
+    case "ts":
+    case "typescript":
```
azz: `case "tsx"`

src/printer-markdown.js
```diff
@@ -93,7 +93,7 @@ function genericPrint(path, options, print) {
         "```",
         node.lang || "",
         hardline,
-        node.value,
+        getFormattedCode(path, options),
```
azz: No way to use multiparser here?

ikatyang: Actually, I don't know how to use multiparser, need some helps. 😅

azz: * Add another case to `getSubtreeParser()` for `markdown`
* call into a new function `fromMarkdown(path)`
* Switch over the `node.type` like the other functions
* For `case "code"`, do something like this:

    ```js
    const parser = getParserName(node.lang);
    if (parser) {
      return {
        options: { parse },
        transformDoc: doc => concat([
          "```",
          node.lang,
          hardline,
          doc,
          "```"
        ]),
        text: node.value
      }
    }
    ```

azz: It should handle indentation automatically as it ends up being one big `doc`.

ikatyang: @azz thanks, it works! ❤️

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -997,7 +997,12 @@ It fits in a single line so it's going to stay as is. However, we've all run
 into this situation:
 
 \`\`\`js
-foo(reallyLongArg(), omgSoManyParameters(), IShouldRefactorThis(), isThereSeriouslyAnotherOne());
+foo(
```
azz: We'll need to `<!-- prettier-ignore -->` this one 😉.

azz: (in the regular README.md)

ikatyang:
For nested emphasis/strong, e.g. `_21**3**_`, should we toggle them between two style , or just enforce one of strong or emphasis to be `_` style?

azz:
Found a nice test case: https://raw.githubusercontent.com/adamschwartz/github-markdown-kitchen-sink/master/TEST.md

Think there will any issues leaving another copy of the README lying around?

src/doc-printer.js
```diff
@@ -12,8 +12,8 @@ function rootIndent() {
   return {
```
azz: Could you explain why these changes were required?

ikatyang: Use for `blockquote` (leading `> `)

src/printer-markdown.js
```diff
@@ -0,0 +1,461 @@
...
+      return concat([
+        "![",
+        node.alt || "",
+        "](",
+        node.url,
+        node.title ? ` "${node.title}"` : "",
+        ")"
+      ]);
+    case "blockquote":
+      return concat(["> ", align("> ", printChildren(path, options, print))]);
```
azz: Couldn't this be something like 

```js
concat(["> ", join([hardline, "> "],  printChildren(path, options, print)));
```

src/doc-printer.js
azz: https://github.com/prettier/prettier/pull/2943/files#r142024814

azz:
> For nested emphasis/strong, e.g. `_21**3**_`, should we toggle them between two style , or just enforce one of strong or emphasis to be `_` style?

I usually use `_` for _emphasis_, and `**` for **strength**, but I have no idea which versions of markdown that is compatible with.

src/printer-markdown.js
ikatyang: Isn't the `join` for concat docs with specified separator? The `>` should be printed on every line break (just like `indent` but print `>`) and the `blockquote` can contains any kind of nodes.

My first attempt is to print `ifBreak("> ")` on every whitespace, but caused a lot of weird results, so I switched to use `align`.

ikatyang:
Should be compatible with all versions I think:

- https://github.github.com/gfm/#emphasis-and-strong-emphasis
- https://daringfireball.net/projects/markdown/syntax#em

lydell:
Minor thing to discuss:

So far, we've generally tried to make Prettier output code that minimizes unnecessary git diff. For example, we don't do this:

```diff
-const foo = 1;
+const foo              = 1;
+const longVariableName = 2;
```

So what about markdown?

- I really like that numbered lists are numbered with `1.`, `2.`, `3.`, etc. (for readability), but inserting a new item causes every following item to change, as opposed to using `1.`, `1.`, `1.`
- Tables. I really like when all cells line up. It's hard to read otherwise. But making one table cell longer can cause re-flow of the entire table.

Thoughts? Do we need to document some of these decisions?

azz:
I did some research using the sample datasets on Google BigQuery: https://bigquery.cloud.google.com/savedquery/652929483875:80c462386a5b4d3294cf6151b31c197c


single_under | double_under | single_star | double_star 
 -- | -- | -- | -- 
 274,827 | 22,877 | 30,932 | 89,270 

So it seems my previous post:

> I usually use `_` for _emphasis_, and `**` for **strength**, but I have no idea which versions of markdown that is compatible with.

Is the common case.

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -996,13 +997,10 @@ foo(arg1, arg2, arg3, arg4);
 It fits in a single line so it's going to stay as is. However, we've all run
 into this situation:
 
+<!-- prettier-ignore -->
+
```
azz: Is this blank line intentional?

ikatyang: It always prints additional line break for continuous nodes that their types are different, should we add a special case for `prettier-ignore`?

azz: Yes, I think so. It will probably be reported as a bug.

ikatyang:
> I really like that numbered lists are numbered with 1., 2., 3., etc. (for readability), but inserting a new item causes every following item to change, as opposed to using 1., 1., 1.

I forgot it'll cause unnecessary git diff, should we switch back to `1.`? though I really like the readability of auto-numbered list.

> Tables. I really like when all cells line up. It's hard to read otherwise. But making one table cell longer can cause re-flow of the entire table.

I think this one should be fine since it only happened if it was so long.

---

I even respect the alignment just for the readability, not sure if this is good since it'll cause unnecessary git diff.

```
| a     |   b   |     c |
| :---- | :---: | ----: |
| d     |   e   |     f |
```

azz:
R.e. numbered lists, this might be one of those cases where we want to preserve the choice to use `1.` repeatedly.

ikatyang:
Should we enable the [commonmark option](https://github.com/wooorm/remark/tree/master/packages/remark-parse#optionscommonmark)? as it seems GitHub uses this format and we can support `break` in this mode.

src/clean-ast.js
```diff
@@ -228,14 +228,15 @@ function massageAST(ast, parent) {
       quasis.forEach(q => delete q.value);
     }
 
-    // styled-components and graphql
+    // styled-components and graphql and markdown
     if (
```
azz: nit: replace `and`s with commas

src/multiparser.js
```diff
@@ -91,11 +140,40 @@ function fromBabylonFlowOrTypeScript(path) {
         };
       }
 
+      /**
```
azz: This will be the first time it is possible to infinitely nest the multiparser...

azz:
```js
markdown`
     ```js
     markdown`
          ```js
          console.log('hi');
          ```
     `
     ```
`
```

ikatyang: Oh, I think we should use `~~~`-style code block for markdown in template string.

azz: I didn't even know that existed!

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1862,8 +1862,8 @@ directory](https://github.com/prettier/prettier/tree/master/editors/vim).
 
 ### Visual Studio Code
 
-Can be installed using the extension sidebar. Search for
-\`Prettier - JavaScript formatter\`.
+Can be installed using the extension sidebar. Search for \`Prettier - JavaScript
+formatter\`.
```
azz: Is this linebreak a problem?

ikatyang: From [commonmark spec](http://spec.commonmark.org/0.25/#example-314), it shouldn't be a problem.

azz:
Looking forward to merging this! What's the next steps for completing this item?

>  pass all the examples in the spec using AST_COMPARE

ikatyang:
There are some tests failed in the [spec example](http://spec.commonmark.org/0.25/#example-616) re-parsing (passed `571/616`), I'm working on it.

Once I passed all the spec example, it should be good to go.

---

The last item `figure out a better way to escape necessary characters` is about removing unnecessary escapes (backslash), currently we just remain the same backslashes since I haven't figured out how to handle cases like the following, but this can be in the future PR.

```md
\~~123~~
\~\~123\~\~
```

moranje:
In the `multiparser.js` on line 183 you have a function called `escapeBackstick`. You may want to change that to `escapeBacktics` or `escapeBacktic` unless I misunderstood and you really meant to introduce the concept of a back stick :-)

ikatyang:
Most of the changes are bug fixes.

Notable changes:

- change thematicBreak style: `---` -> `- - -` to avoid conflict with yaml syntax
- use `* * *`-style thematicBreak in list to avoid conflict with list `-`-style

---

- the `AST_COMPARE` test for spec example passed `607/616`, the rest of them are `remark-parse` parsing bugs and just some edge cases (I put them in `tests/markdown_spec/remark-bug`), this PR should be good to go now.

Graham42:
Have you considered hard wrapping text for source readability? 

There's a styles guide [here](http://www.cirosantilli.com/markdown-style-guide/#line-wrapping) which details several options for wrapping and rationale for why each style might be chosen.

I feel that source readability is the most important consideration and for that reason would choose to wrap all text at 80 characters.

The one argument I see against this would that diffs aren't as readable. However git does word diffs pretty well when given the flag to do so (`--word-diff`). And github highlights word diffs along with the line by default. [Here's a recent example](https://github.com/prettier/prettier/commit/592f99fde69880c69fadb3887db911d87fa306a6#diff-04c6e90faac2675aa89e2176d2eec7d8L702).

src/parser-markdown.js
```diff
@@ -40,28 +40,55 @@ function map(ast, handler) {
   })(ast, null, null);
 }
 
-function transformInlincode() {
-  return ast =>
+function transformInlinCode(originalText) {
```
lydell: transformInlin<strong>e</strong>Code?

azz:
Do we have any concerns with adding ~600 new files?

ikatyang:
@Graham42 

I'm not sure I understand what you mean, the current behavior is similar to [Option wrap:space](http://www.cirosantilli.com/markdown-style-guide/#option-wrap-space) but the max length is configurable (`--printWidth`, `80` by default), it seems this is already what you want?

lydell:
> Do we have any concerns with adding ~600 new files?

A ~17% increase. I don't think it is a problem though. What could potential problems be?

azz:
I don't think it's a problem, and having good spec coverage is super valuable. Just thought I'd mention it as it is _quite a lot_ of files. Only downside is slower test run, I guess.

Graham42:
@ikatyang That's great. I was trying to find anything about wrapping in this PR and wasn't seeing it. And the [example file](https://github.com/ikatyang/prettier/blob/feat/markdown/tests/markdown/__snapshots__/jsfmt.spec.js.snap) linked above wasn't wrapped so I wanted to raise the concern.

Thanks for clarifying.

ikatyang:
@Graham42 

That's a test snapshot file, the top part is unformatted, the bottom part is formatted. 😅

Graham42:
O! well that explains it. That looks _much prettier_ 💯

azu:
I have a question.
Does this markdown formatter support multi-bye language like Japanese or emoji?

It seem that current implementaion has two problement.

First, current implementation will be broken in Chinese-Japanese-Korean(CJK) and emoji.
Because, some printing algorithm depened on `string.length`.

For example, `"❤️".length` is `2`. Following algorithm jolt out of alignment.

- https://github.com/prettier/prettier/pull/2943/files#diff-385bd78c43a57ae55923bfea744a5ae3R359

This issue is known as East Asian Width or Unicode problem.

- [UAX #11: East Asian Width](http://unicode.org/reports/tr11/ "UAX #11: East Asian Width")
- [JavaScript has a Unicode problem · Mathias Bynens](https://mathiasbynens.be/notes/javascript-unicode "JavaScript has a Unicode problem · Mathias Bynens")

I know that East Asian Width problem is very difficult.( I don't know perfect solution...)
Some library get unicode length.

- [timoxley/wcwidth: Port of C's wcwidth() and wcswidth()](https://github.com/timoxley/wcwidth)
- [sindresorhus/string-width: Get the visual width of a string - the number of columns required to display it](https://github.com/sindresorhus/string-width)
- [power-assert-runtime/packages/power-assert-util-string-width at master · twada/power-assert-runtime](https://github.com/twada/power-assert-runtime/tree/master/packages/power-assert-util-string-width)

Socnd, Following tokenizer can not tokenize non-English text.
Because, non-English text like cheniese doesn't put a space between words.

```
    .split(/(\s+)/g)
```

- https://github.com/prettier/prettier/pull/2943/files#diff-0f09e16c6ee7c5a40fef83b315027002R149

It will be resolved by using tokenizer like [nlcst-parse-japanese](https://github.com/azu/nlp-pattern-match/tree/master/packages/nlcst-parse-japanese "nlcst-parse-japanese"), [parse-english](https://github.com/wooorm/parse-english "parse-english"), [rakutenma](https://github.com/rakuten-nlp/rakutenma "rakutenma").
But, It is just not realistic. because toknizer is heavy weight.

Toknizer example is [Text Tokenizer · Hivemall User Manual](https://hivemall.incubator.apache.org/userguide/misc/tokenizer.html "Text Tokenizer · Hivemall User Manual")


Thanks.

ikatyang:
@azu

Thanks for the suggestion, I've already thought about it and just merged #3003 to fix the printer first, the CJK support is working in progress now, but it should be in a separate PR I think, this PR is somehow too large.

I think there's no need to use tokenizer for CJK, since AFAIK they can be broke in any place, and that's how CJK books printed, e.g.

```
一串很長很長很長很長很長很長很長很長很長很長很長的中文字，一串很長很長很長很長很長很
長很長很長很長很長很長的中文字，一串很長很長很長很長很長很長很長很長很長很長很長的中
文字，一串很長很長很長很長很長很長很長很長很長很長很長的中文字。
```

For the string width thing, use [`string-width`](https://github.com/sindresorhus/string-width) should be enough, but we have to disable the `stripAnsi` feature.

---

@azz 

Do we have any concern to merge this PR? I'd like to send some followed up PRs instead of pushing to the current branch since this PR is somehow too large to review easily.

azz:
I've been incrementally reviewing the changes so we're fine to merge!

vjeux:
So cool!

nhoizey:
Amazing!

Is it possible to also support Mardown dialects, such as Kramdown, used by Jekyll?

ikatyang:
@nhoizey 

It seems [`kramdown` does not follow the commonmark spec](https://github.com/gettalong/kramdown/issues/404#issuecomment-275213633) so that it should have its own printer instead of merging into this one.

For new language support, see https://github.com/prettier/prettier/issues/3017#issuecomment-336179856.

nhoizey:
@ikatyang thanks, I'll see what I can do!

lipis:
Now we should format all *.md files in this project! :)

lydell:
@lipis Good idea!

lipis:
#3022 quite a few things to take care of though

azz:
Big Data research part two: https://bigquery.cloud.google.com/savedquery/652929483875:69be015faec8444c8a2e7d055eeb32f8

`* item` | `- item`
 -- | --
642,106 | 413,627

type | regex
--|--
`* item` | `/^\s*\*\s/gm`
`- item` | ` /^\s*-\s/gm`

`* item` wins!

ikatyang:
There's also the `+`-style list item, we should do research for it too, and take the top two results to be used in:

> listItem: `-`(always), `+`(switch between two style for continuous lists)

(current)

```md
- 123
- 123

+ 456
+ 456
```

azz:
`+` is not very popular. Edited my previous comment.

src/printer-markdown.js
```diff
@@ -133,24 +135,121 @@ function genericPrint(path, options, print) {
...
+    rowPath.map(cellPath => {
+      rowContents.push(
+        printDocToString(cellPath.call(print), options).formatted
+      );
+    }, "children");
+
+    contents.push(rowContents);
+  }, "children");
+
+  const columnMaxWidths = contents.reduce(
```
revelt: We should also limit the max length of the columns. For example, All-Contributors tables contain HTML code and table dash columns processed by Prettier get very very long.

j-f1: Can you open a new issue so we can track this better?

revelt: Done https://github.com/prettier/prettier/issues/4195

PR #2948, opened 2017-10-01
Respect --stdin-filepath, regardless of config source

### LINKED ISSUE #2946: stdin with overrides
Hi,

I don't know if it's a bug or a unsolvable problem. Without the cli option `--stdin` it's working like expected. I cannot provide a playground example, because it only happens with the stdin option.

See yourself:

**Working** (without --stdin)
```bash
prettier --config .prettierrc.json --config-precedence prefer-file index.js
```

**Not working** (with --stdin)
```bash
cat index.js | prettier --stdin --config .prettierrc.json --config-precedence prefer-file
```

.prettierrc.json
```json
{
  "tabWidth": 2,
  "overrides": [{
    "files": "*.js",
    "options": {
      "parser": "babylon",
      "tabWidth": 4
    }
  }]
}
```

input.js
```js
const hello = {
  beatiful: "world"
};
```

output.js
```js
const hello = {
  beatiful: "world"
};
```

expected-output.js
```js
const hello = {
    beatiful: "world"
};
```

I think the problem here is, that prettier cannot detect which file extension the source has, because it comes from the stdin. I'm right?

I'm using the JsPrettier Plugin for Sublime Text.
https://github.com/jonlabelle/SublimeJsPrettier/issues/77

azz:
Fixes #2946

### DIFF
--- a/src/cli-util.js
+++ b/src/cli-util.js
@@ -159,11 +159,14 @@ function getOptionsOrDie(argv, filePath) {
   }
 }
 
-function getOptionsForFile(argv, filePath) {
-  const options = getOptionsOrDie(argv, filePath);
-  return applyConfigPrecedence(
-    argv,
-    options && normalizeConfig("api", options, constant.detailedOptionMap)
+function getOptionsForFile(argv, filepath) {
+  const options = getOptionsOrDie(argv, filepath);
+  return Object.assign(
+    { filepath },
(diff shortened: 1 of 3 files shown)


mitermayer:
This is awesome can we get a release soon with this ?

azz:
Released 1.7.4

PR #2950, opened 2017-10-01
Mention labels docs in CONTRIBUTING.md

azz:
Also remove `--` that was required in `yarn` v0.


azz:
Maybe we should move `commands.md` and `Rationale.md` to the wiki, too?

ikatyang:
Sounds good to me.

PR #2767, opened 2017-09-07
refactor(cli): centralize options

ikatyang:
main:

- centralize option information
- auto-generate usage from option information
- validate choice options from option information
- use api default options instead of its own cli default options

side:

- sort usage options
- rename `e`/`err` to `error`
- rename `resolveConfig` to `logResolvedConfigPathOrDie`

PR #2951, opened 2017-10-02
test(integration): update title and catch all outputs

ikatyang:
Context: https://github.com/prettier/prettier/pull/2767#discussion_r137940458

Add title `(stdout)`, `(stderr)`, etc. and make sure there's no unexpected outputs at all.

before:

```js
const result = runPrettier(...);
expect(result.stdout).toMatchSnapshot();
expect(result.status).toEqual(0);
```

after

```js
runPrettier(...).test({ status: 0 }); /*=>
  const result = runPrettier(...);
  test("(stdout)", () => {
    expect(result.stdout).toMatchSnapshot();
  });
  test("(stderr)", () => {
    expect(result.stderr).toMatchSnapshot();
  });
  test("(write)", () => {
    expect(result.write).toMatchSnapshot();
  });
  test("(status)", () => {
    expect(result.status).toEqual(0);
  });
*/
```

### DIFF
--- a/tests_integration/__tests__/__snapshots__/arg-parsing.js.snap
+++ b/tests_integration/__tests__/__snapshots__/arg-parsing.js.snap
@@ -1,57 +1,69 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`boolean flags do not swallow the next argument 1`] = `
+exports[`boolean flags do not swallow the next argument (stderr) 1`] = `""`;
+
+exports[`boolean flags do not swallow the next argument (stdout) 1`] = `
 "console.log('could be single quote and without semi');
 "
 `;
 
-exports[`boolean flags do not swallow the next argument 2`] = `""`;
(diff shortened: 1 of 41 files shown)


tests_integration/__tests__/__snapshots__/arg-parsing.js.snap
```diff
@@ -1,57 +1,69 @@
...
 
-exports[`negated options work 1`] = `
-"console.log(\\"could be single quote and without semi\\")
+exports[`deprecated options are warned (stdout) 1`] = `
+"console.log(\\"could be single quote and without semi\\");
 "
 `;
 
-exports[`negated options work 2`] = `""`;
+exports[`deprecated options are warned (write) 1`] = `Array []`;
```
azz: Why are these arrays?

ikatyang: Something like this:

```
Array [
  Object {	
    "content": "var x = 1;",
    "filename": "unformated.js",
  },
]
```

PR #2786, opened 2017-09-10
Break closing paren of ConditionalExpression in member chains

duailibe:
Still related to #2775, maybe a better way to print this case.

```js
(a
  ? b
  : c
).prop;

(a
  ? b
  : c
).call();

(a
  ? b
  : c
)
  .call1()
  .call2();
```

If you guys think this is cool, let me know if I should do the same for LogicalExpression (I believe so) and we can do it in a single commit

Thanks!

Issue #2918, opened 2017-09-27
Find an open-source "real world" React app and add it to our snapshots


suchipi:
There was a bug introduced in the 1.7.1 release that we would have caught in a real app pretty quickly, but we missed it.

If we could find a medium-to-large open-source React app to include in our snapshot suite, then we could see when we introduce changes that break things.

I say React because then we would see how JSX formats. Once we have better Vue support, it would make sense to add a vue app to the suite, too etc.

Ideally it'd be an app with:
* Some small components
* Some large components
* A few different patterns / somewhat inconsistent code
* Tests

AKA a production, "real world" app.

hawkrives:
I can volunteer either of the apps I work on, https://github.com/hawkrives/gobbldygook or https://github.com/StoDevX/aao-react-native - they're both formatted with prettier, although I need to take a bit of time to upgrade them to 1.7. 

I don't know how large of an app you're looking for, though; they might be too large.

suchipi:
Both look great to me- I actually would like to include both, unless the other collaborators don't want to take the size hit.

PR #2958, opened 2017-10-02
Break parens for binaries in member expression

### LINKED ISSUE #2934: Invalid indentation
**Prettier 1.7.2**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22const%20someLongVariable%20%3D%20(%5Cn%20%20idx(%5Cn%20%20%5Ctthis.props%2C%5Cn%20%20%5Ctprops%20%3D%3E%20props.someLongPropertyName%2C%5Cn%20%20)%20%7C%7C%20%5B%5D%5Cn).map(edge%20%3D%3E%20edge.node)%3B%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)

**Input:**
```jsx
const someLongVariable = (
  idx(
    this.props,
    props => props.someLongPropertyName,
  ) || []
).map(edge => edge.node);
```

**Output:**
```jsx
const someLongVariable = (idx(
  this.props,
  props => props.someLongPropertyName
) || []
).map(edge => edge.node);

```

**Expected behavior:**
The last 2 lines should't have the same indentation.
Either the `)` from the last line should be on the line before and `.map` indented once or something similar to the input.

duailibe:
Fixes #2934 (previous related work: #2786)

This will make binaries break like:
```js
(
  a &&
  b &&
  c
).member
```

The snippet in the issue would be formatted as:

```js
// before
const someLongVariable = (idx(
  this.props,
  props => props.someLongPropertyName
) || []
).map(edge => edge.node);

// after
const someLongVariable = (
  idx(this.props, props => props.someLongPropertyName) || []
).map(edge => edge.node);

### DIFF
--- a/tests/method-chain/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/method-chain/__snapshots__/jsfmt.spec.js.snap
@@ -570,6 +570,12 @@ Object.keys(
 `;
 
 exports[`logical.js 1`] = `
+const someLongVariableName = (idx(
+  this.props,
+  props => props.someLongPropertyName
+) || []
+).map(edge => edge.node);
+
 (veryLongVeryLongVeryLong || e).map(tickets =>
   TicketRecord.createFromSomeLongString());
 
@@ -582,6 +588,10 @@ exports[`logical.js 1`] = `
 (veryLongVeryLongVeryLong || anotherVeryLongVeryLongVeryLong || veryVeryVeryLongError).map(tickets =>
   TicketRecord.createFromSomeLongString()).filter(obj => !!obj);
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+const someLongVariableName = (
+  idx(this.props, props => props.someLongPropertyName) || []
+).map(edge => edge.node);
+
 (veryLongVeryLongVeryLong || e).map(tickets =>
   TicketRecord.createFromSomeLongString()
 );
--- a/tests/method-chain/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/method-chain/__snapshots__/jsfmt.spec.js.snap
@@ -590,12 +590,14 @@ exports[`logical.js 1`] = `
   .map(tickets => TicketRecord.createFromSomeLongString())
   .filter(obj => !!obj);
 
-(veryLongVeryLongVeryLong ||
+(
+  veryLongVeryLongVeryLong ||
   anotherVeryLongVeryLongVeryLong ||
   veryVeryVeryLongError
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by lydell** [comment]
Nice! Should we add the snippet to the tests?

**Comment by duailibe** [comment]
Done!

File: tests/method-chain/__snapshots__/jsfmt.spec.js.snap [hunk 1]
```diff
@@ -590,12 +600,14 @@ exports[`logical.js 1`] = `
   .map(tickets => TicketRecord.createFromSomeLongString())
   .filter(obj => !!obj);
 
-(veryLongVeryLongVeryLong ||
+(
```
azz: This _might_ be considered a regression...

azz:
I was going to merge this in for 1.7.4 but I am a bit concerned it might be a regression. Would like to run it on some larger codebases when I get a chance.

duailibe:
I just ran it on the codebase at work and only diff I got was this:

```diff
-    const chapterProgressIcons = ((chapterSummary && chapterSummary.chapterProgresses) ||
+    const chapterProgressIcons = (
+      (chapterSummary && chapterSummary.chapterProgresses) ||
       []
     ).map(c => { /* stuff */ })
```

azz:
I kind of liked the inlineness of it before but maybe I'm just used to it  ¯\_(ツ)_/¯

duailibe:
Yeah I can see that. For context, we'd like to not print:

```js
(foo &&
  bar &&
  baz).call()
```

But also not print:
```js
(func(
  foo,
  bar
) || var
).call()
```

I'm open for suggestions

duailibe:
Not sure why I closed this

suchipi:
Maybe we could run this against gobbldygook or aao-react-native? https://github.com/prettier/prettier/issues/2918

I still need to get around to adding those codebases in as part of our normal test suite...

hawkrives:
@suchipi I might be running it wrong, but I ran this branch on both [gobbldygook](https://github.com/hawkrives/gobbldygook/compare/prettier__fix-binary-in-member-expression) and [All About Olaf](https://github.com/StoDevX/AAO-React-Native/compare/prettier__fix-binary-in-member-expression), and neither had any changes. (Maybe that's not so surprising… I think I usually pull out binary expressions into their own variables instead of doing them inline.)

If you check out those branches locally, `npm run prettier` should run this branch's `prettier` against all of their source files.

suchipi:
Thanks everyone

PR #2957, opened 2017-10-02
test: fix snapshots

ikatyang:
Not sure why, but the build failed.

- remove `--no-color`
- use `strip-ansi` to enforce no colored snapshots
- update snapshots

Fixes #2956

PR #2960, opened 2017-10-02
Include duration in --write snapshots

lydell:
Context: https://github.com/prettier/prettier/pull/2957#discussion_r142209753

### DIFF
--- a/src/cli-util.js
+++ b/src/cli-util.js
@@ -325,13 +325,13 @@ function formatFiles(argv) {
       // mtime based caches.
       if (output === input) {
         if (!argv["list-different"]) {
-          console.log(chalk.grey(filename), `${Date.now() - start}ms`);
+          console.log(`${chalk.grey(filename)} ${Date.now() - start}ms`);
         }
       } else {
         if (argv["list-different"]) {
           console.log(filename);
         } else {
-          console.log(filename, `${Date.now() - start}ms`);
+          console.log(`${filename} ${Date.now() - start}ms`);
         }
 
         try {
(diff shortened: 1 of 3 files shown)


src/cli-util.js
```diff
@@ -327,13 +327,20 @@ function formatFiles(argv) {
...
       if (output === input) {
         if (!argv["list-different"]) {
-          console.log(chalk.grey("%s %dms"), filename, Date.now() - start);
+          // For the tests, we only support the first argument for
+          // `console.log`. This means that the duration will appear as `%dms`
+          // in the snapshots, instead of for example `40ms`. This needs to be
+          // done since otherwise the tests would only pass (match the snapshot)
+          // if they happened to run in the exact amount of time as when
+          // created.
+          console.log(chalk.grey(`${filename} %dms`), Date.now() - start);
```
ikatyang: Maybe we can mock `Date.now()` to always output `0` and replace `%d`/`%s` with their actual value?

tests_integration/__tests__/__snapshots__/write.js.snap
```diff
@@ -3,7 +3,7 @@
 exports[`do not write file with --write + formated file (stderr) 1`] = `""`;
 
 exports[`do not write file with --write + formated file (stdout) 1`] = `
-"formated.js%s %dms
+"formated.jsformated.js %dms
```
ikatyang: I just realize I removed the `[2K[1G` (remove line) control character in the snapshot output (`strip-ansi`) so that it looks weird here, I think we can add a special case for it to actually do the same thing in the mocked stdout/stderr.

azz:
Is this good to go, or do you want to do more ANSI changes?

lydell:
I don't want to do more ANSI stuff, but I want to try mocking `Date.now`.

lydell:
Ok, I ended up messing with the ANSI stuff anyway. @ikatyang What do you think?

lydell:
Works locally but fails on Travis :(

azz:
Failure is unrelated to your changes. #2903

lydell:
I scrapped my previous work and started over from master. Now, I've only added the duration to the snapshots (by mocking `Date.now`).

PR #2961, opened 2017-10-02
[WIP / RFC] Support expressions in CSS selectors

### LINKED ISSUE #2883: Prettier breaks with interpolation and complex CSS selector
**Prettier 1.7.0**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22export%20const%20foo%20%3D%20css%60%5Cn%26.foo%20.%24%7Bbar%7D%3A%3Abefore%2C%26.foo%5Bvalue%3D%5C%22hello%5C%22%5D%20.%24%7Bbar%7D%3A%3Abefore%20%7B%5Cn%5Ctposition%3A%20absolute%3B%5Cn%7D%5Cn%60%3B%5Cn%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22typescript%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A4%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)
```sh
--parser typescript
--tab-width 4
```

**Input:**
```tsx
export const foo = css`
&.foo .${bar}::before,&.foo[value="hello"] .${bar}::before {
	position: absolute;
}
`;

```

**Output:**
```txt
Error: Error: Couldn't insert all the expressions
```

**Expected behavior:**

```tsx
export const foo = css`
    &.foo .${bar}::before,
    &.foo[value="hello"] .${bar}::before {
        position: absolute;
    }
`;
```
If I replace `before,&` by `before, &` (space before `&`), it works but is not pretty...

Thank you to look at this :D you are wonderful!

duailibe:
Fixes #2883

We don't currently support expressions in CSS selectors because postcss-selector-parser can't parse `@prettier-placeholder` correctly. With this commit we can support expressions in the CSS selector, but I'm not sure... this might be overkill.

I'd like some feedback, please tell me what you think!

### DIFF
--- a/src/multiparser.js
+++ b/src/multiparser.js
@@ -227,19 +227,23 @@ function replacePlaceholders(quasisDoc, expressionDocs) {
       const rest = parts.slice(2);
       parts = [at + placeholder].concat(rest);
     }
-    if (
-      typeof parts[0] === "string" &&
-      parts[0].startsWith("@prettier-placeholder")
-    ) {
-      const placeholder = parts[0];
-      const rest = parts.slice(1);
-
-      // When the expression has a suffix appended, like:
-      // animation: linear ${time}s ease-out;
-      const suffix = placeholder.slice("@prettier-placeholder".length);
-
-      const expression = expressions.shift();
-      parts = ["${", expression, "}" + suffix].concat(rest);
+    const index = parts.findIndex(
+      p => typeof p === "string" && p.indexOf("@prettier-placeholder") !== -1
+    );
+    if (index > -1) {
+      parts = parts.reduce((acc, part) => {
+        if (
+          typeof part !== "string" ||
+          part.indexOf("@prettier-placeholder") === -1
+        ) {
+          acc.push(part);
+        } else {
+          const affix = part.split("@prettier-placeholder");
+          const expression = expressions.shift();
+          acc = acc.concat([affix[0] + "${", expression, "}" + affix[1]]);
+        }
+        return acc;
+      }, []);
     }
     return Object.assign({}, doc, {
       parts: parts
--- a/src/parser-postcss.js
+++ b/src/parser-postcss.js
@@ -2,12 +2,41 @@
 
 const createError = require("./parser-create-error");
 
+// Postcss won't parse correctly selectors with the "@prettier-placeholder",
+// so we change it to a placeholder it can parse and then change it back
+// before printing.
+function fixSelectorPlaceholder(root) {
+  if (root.nodes) {
+    root.nodes.forEach(node => {
+      if (node.value && node.value.indexOf("PRETTIER_PLACEHOLDER") > -1) {
+        node.value = node.value.replace(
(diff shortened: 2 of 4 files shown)


brikou:
Not sure about implementation, as I don't have enough experience with prettier as a dev, but as a user I can tell that this bug is really annoying, a single used expression in a big JS file prevent prettier to work at all. So thank you for your work, and can't wait to have this fix :+1:

src/parser-postcss.js
```diff
@@ -2,12 +2,41 @@
...
 
+// Postcss won't parse correctly selectors with the "@prettier-placeholder",
+// so we change it to a placeholder it can parse and then change it back
+// before printing.
+function fixSelectorPlaceholder(root) {
+  if (root.nodes) {
+    root.nodes.forEach(node => {
+      if (node.value && node.value.indexOf("PRETTIER_PLACEHOLDER") > -1) {
+        node.value = node.value.replace(
+          /PRETTIER_PLACEHOLDER/g,
```
azz: What happens if this string is in the source?

duailibe: @azz it gets replaced by `@prettier-placeholder`

Maybe add a sanity check? Throws if the counts don’t match

src/multiparser.js
```diff
@@ -227,19 +227,23 @@ function replacePlaceholders(quasisDoc, expressionDocs) {
...
-      const expression = expressions.shift();
-      parts = ["${", expression, "}" + suffix].concat(rest);
+    const index = parts.findIndex(
+      p => typeof p === "string" && p.indexOf("@prettier-placeholder") !== -1
+    );
+    if (index > -1) {
+      parts = parts.reduce((acc, part) => {
+        if (
+          typeof part !== "string" ||
+          part.indexOf("@prettier-placeholder") === -1
```
azz: `!part.includes("@prettier-placeholder")`

azz:
Was there a reason we're using the specific string `@prettier-placeholder`? Could we just use `__PRETTIER_PLACEHOLDER__` or something like that?

duailibe:
@azz that wouldn't work for "expression as rule":

```jsx
<style jsx>{`
div: {
  ${expr};
}
`}</style>

azz:
What if we _didn't_ put a placeholder and tracked the locations, similar to how we track comments?

duailibe:
I can see how that working for the case above, but not `div.${expr}`. We eventually need to parse this, right?

Or there's something I'm not visualizing about your suggestion 😅

azz:
Yeah you're right. Just trying to work through some alternatives.

duailibe:
Not gonna continue with this right now

Issue #2857, opened 2017-09-19
Broken code result when formatting GraphQL queries inside `graphql` calls


freiksenet:
**Prettier 1.7.0**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22graphql(schema%2C%20%60%7B%20query%20test%20%7B%20id%20%7D%7D%20%24%7Bfragment%7D%60)%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)
```sh
# Options (if any):
--single-quote
```

**Input:**
```jsx
graphql(schema, `{ query test { id }} ${fragment}`)
```

**Output:**
```jsx
graphql(
  schema,
  `
    {
      query
      test {
        id
      }
    }
  ${fragment}
    {
      query
      test {
        id
      }
    }
  `
);
```

**Expected behavior:**

It's okay to ignore invalid GraphQL output or format it badly, but this actually breaks it and produces faulty code.

freiksenet:
Related #2640

azz:
Fixed in #2975

Issue #2640, opened 2017-08-19
GraphQL, gql template literals, and including fragments


baerrach:
Currently I am writing my graphql in a *.js file, where the query is referencing the fragment from a JavaScript variable that contains the fragment gql. However prettier doesn't recognise the ${} stuff and isn't reformatting the code.

Its entirely possible I'm not doing this in the recommended way (e.g. not using *.graphql files). Updates to existing documentation to put people on the right path would be appreciated.

```graphql
fragments {
  all: gql`...`,
};

query = gql`
query allPartsByManufacturerName(
  $name: String!
) {
  allParts(filter:{
    manufacturer: {
      name: $name
    }
  }) {
    ...PartAll
  }
}
${fragments.all}
```

azz:
We don't currently support expressions inside tagged template GraphQL literals. We should add this.

lydell:
This was fixed by #3370 :tada:

PR #2975, opened 2017-10-04
Prevent formatting GQL template literals with expressions

duailibe:
Related to #2640 and #2857.

I'll take a shot in trying to support expressions in GQL template literals, but anyway, failing to do that, this PR should prevent breaking people's code.

### DIFF
--- a/src/multiparser.js
+++ b/src/multiparser.js
@@ -68,7 +68,7 @@ function fromBabylonFlowOrTypeScript(path) {
        */
       if (
         // We currently don't support expression inside GraphQL template literals
-        parent.quasis.length === 1 &&
+        parent.expressions.length === 0 &&
         parentParent &&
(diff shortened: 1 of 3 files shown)


src/multiparser.js
```diff
@@ -67,9 +67,10 @@ function fromBabylonFlowOrTypeScript(path) {
        * gql`...`
        */
       if (
+        // We currently don't support expression inside GraphQL template literals
+        parent.quasis.length === 1 &&
```
suchipi: You don't necessarily need to make this change, but it might be more readable if we checked the expressions length instead of the quasis length (either works, though, since they're associated with each other).

duailibe: Indeed

azz:
Thanks for the fix!

PR #2981, opened 2017-10-05
Fix build

azz:
@lydell this will conflict with your PR. Maybe you want to cherry pick this and resolve conflicts?

### DIFF
--- a/tests_integration/__tests__/__snapshots__/write.js.snap
+++ b/tests_integration/__tests__/__snapshots__/write.js.snap
@@ -3,7 +3,7 @@
 exports[`do not write file with --write + formated file (stderr) 1`] = `""`;
 
 exports[`do not write file with --write + formated file (stdout) 1`] = `
-"formated.js%s %dms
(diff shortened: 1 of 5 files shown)


tests_integration/__tests__/__snapshots__/write.js.snap
```diff
@@ -15,14 +15,14 @@ exports[`do not write file with --write + invalid file (stderr) 1`] = `
 `;
 
 exports[`do not write file with --write + invalid file (stdout) 1`] = `
-"invalid.js
```
azz: Not sure if this is supposed to be logged when the file has a syntax error?

ikatyang: It should not be logged, but how did it get removed here? I didn't see code changes related to it. 😱

azz: Must have just been when I updated the snapshot.

PR #2982, opened 2017-10-06
Fix typo in test files (`*formated` -> `*formatted`)

Narigo:
As a @lydell mentioned in the review for #2903 , there was a typo in two test files that the integration tests for write and piped-output used. I've renamed them and updated the snapshots accordingly.


azz:
Could you fix the typo in the file name `cursot-offset.js`?

Narigo:
Are there any more typos? I could fix them all, but I did not check the entire code base... :)

azz:
Not off the top of my head. I kept forgetting to fix the `cursot` one 😄

PR #2987, opened 2017-10-07
fix(cli): do not early exit for `--debug-check`

### LINKED ISSUE #2983: debug-check cli option only displays first diff error
The cli option `--debug-check` in previous versions would output all the diff errors.
Now it only outputs the first diff error.

ikatyang:
Fixes #2983


duailibe:
@ikatyang After some changes in #4364, this test started failing and I realized we clean the AST differences of `{"key": value}` -> `{key: value}` in https://github.com/prettier/prettier/blob/6a17474e37c3287ee161cc3ffb385f001801b6a1/src/language-js/clean.js#L61-L73

So why does `--debug-check` outputs that diff?

PR #2992, opened 2017-10-08
feat(cli): add `--loglevel` option

### LINKED ISSUE #2939: Add `--loglevel` option
from https://github.com/prettier/prettier/pull/2903#issuecomment-333285580

Possible option choices:

- `silent`: nothing
- `error`: error message
- `warn` (default): error message + warning
- `debug`: error message + warning + debug message

ikatyang:
Fixes #2939

- Use environment variable `PRETTIER_LOG_LEVEL` internally.
- Use `chalk` directly since `debug` enforces trailing `+0ms`s and it looks weird in some cases.
- Prefix every message to `stderr` with its loglevel, e.g. `error`, `warn`, etc.
- Move `unknown` minimist option to `normalizeConfig`

Question: 

- Should we output `--debug-check` results to stdout? We now prefix every message to stderr with `error/warn/etc.` and its functionality looks similar to `--list-different`.


tests_integration/__tests__/__snapshots__/arg-parsing.js.snap
```diff
@@ -22,7 +22,7 @@ exports[`deprecated option values are warned (stdout) 1`] = `
 exports[`deprecated option values are warned (write) 1`] = `Array []`;
 
 exports[`deprecated options are warned (stderr) 1`] = `
-"\`--flow-parser\` is deprecated. Use \`--parser flow\` instead.
+"warn \`--flow-parser\` is deprecated. Use \`--parser flow\` instead.
```
azz: Could we change it to 

```
warn: `--flow-parser` is deprecated...
```

or

```
[warn] `--flow-parser` is deprecated...
```

ikatyang: image

`[xxx]` looks better to me.

azz: Looks good. It was just but strange without punctuation and without color.

tests_integration/__tests__/__snapshots__/loglevel.js.snap
```diff
@@ -0,0 +1,37 @@
...
+"
+`;
+
+exports[`do not show warnings with --loglevel error (stdout) 1`] = `""`;
+
+exports[`do not show warnings with --loglevel error (write) 1`] = `Array []`;
+
+exports[`show all logs with --loglevel debug (stderr) 1`] = `
+"[warn] Ignored unknown option: --unknown-option
+[debug] normalized argv: {\\"_\\":[\\"not-found.js\\"],\\"bracket-spacing\\":false,\\"color\\":true,\\"debug-check\\":false,\\"debug-print-doc\\":false,\\"flow-parser\\":false,\\"insert-pragma\\":false,\\"jsx-bracket-same-line\\":false,\\"list-different\\":false,\\"require-pragma\\":false,\\"semi\\":false,\\"single-quote\\":false,\\"stdin\\":false,\\"use-tabs\\":false,\\"version\\":false,\\"with-node-modules\\":false,\\"write\\":false,\\"loglevel\\":\\"debug\\",\\"parser\\":\\"unknown-parser\\",\\"ignore-path\\":\\".prettierignore\\",\\"config-precedence\\":\\"cli-override\\"}
```
azz: Is this test prone to breaking?

ikatyang: It seems it'd be better to use `expect(...).toMatch(/[loglevel]/)` in this case.

tests_integration/__tests__/loglevel.js
```diff
@@ -2,51 +2,39 @@
...
+  ]);
+
+  expect(result).not.toEqual(0);
+
+  if (patterns) {
+    patterns.forEach(pattern => {
+      expect(result.stderr).toMatch(pattern);
+    });
+  } else {
+    expect(result.stderr).toMatch(/\s*/);
```
lydell: What is this supposed to do? _Any_ string matches `/\s*/`, right?

ikatyang: Oops, I forgot to add `^`+`$`

PR #3010, opened 2017-10-11
docs: fix table for `insertPragma`

bennyn:
I visited https://prettier.io/docs/en/options.html#content and noticed that the Markdown table for the `insertPragma` option is broken:

Screenshot

So I fixed it. :)

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -156,7 +156,7 @@ or
 Default | CLI Override | API Override
 --------|--------------|-------------
 `false` | `--require-pragma` | `requirePragma: <bool>`
-<!-->
+<!--
 ## Insert Pragma
 Prettier can insert a special @format marker at the top of files specifying that the file has been formatted
 with prettier.  This works well when used in tandem with the `--require-pragma` option.  If there is already a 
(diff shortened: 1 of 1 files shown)


docs/options.md
```diff
@@ -156,7 +156,7 @@ or
 Default | CLI Override | API Override
 --------|--------------|-------------
 `false` | `--require-pragma` | `requirePragma: <bool>`
-<!-->
```
azz: Actually this section is supposed to be hidden because the feature is unreleased. I think you just need to delete that `>` instead.

bennyn:
@azz Thanks for the hint! I brought back the `<!-->`. 😃  👍

azz:
Not quite. These are HTML comments so you want to replace `<!-->` with `<!--` and keep the `-->` as it was.

docs/options.md
bennyn: Ah, I see! I was already wondering why Prettier is telling me: "Ignored unknown option: insertPragma".

bennyn:
Done! :)

PR #3026, opened 2017-10-14
feat(markdown): support CJK and emoji

### LINKED ISSUE #3013: Support CJK and emoji in markdown
from @azu https://github.com/prettier/prettier/pull/2943#issuecomment-335819324:

> I have a question.
> Does this markdown formatter support multi-bye language like Japanese or emoji?
> 
> It seem that current implementaion has two problems.
> 
> First, current implementation will be broken in Chinese-Japanese-Korean(CJK) and emoji.
> Because, some printing algorithm depened on `string.length`.
> 
> For example, `"❤️".length` is `2`. Following algorithm jolt out of alignment.
> 
> - https://github.com/prettier/prettier/pull/2943/files#diff-385bd78c43a57ae55923bfea744a5ae3R359
> 
> This issue is known as East Asian Width or Unicode problem.
> 
> - [UAX #11: East Asian Width](http://unicode.org/reports/tr11/ "UAX #11: East Asian Width")
> - [JavaScript has a Unicode problem · Mathias Bynens](https://mathiasbynens.be/notes/javascript-unicode "JavaScript has a Unicode problem · Mathias Bynens")
> 
> I know that East Asian Width problem is very difficult.( I don't know perfect solution...)
> Some library get unicode length.
> 
> - [timoxley/wcwidth: Port of C's wcwidth() and wcswidth()](https://github.com/timoxley/wcwidth)
> - [sindresorhus/string-width: Get the visual width of a string - the number of columns required to display it](https://github.com/sindresorhus/string-width)
> - [power-assert-runtime/packages/power-assert-util-string-width at master · twada/power-assert-runtime](https://github.com/twada/power-assert-runtime/tree/master/packages/power-assert-util-string-width)
> 
> Second, Following tokenizer can not tokenize non-English text.
> Because, non-English text like cheniese doesn't put a space between words.
> 
> ```
>     .split(/(\s+)/g)
> ```
> 
> - https://github.com/prettier/prettier/pull/2943/files#diff-0f09e16c6ee7c5a40fef83b315027002R149
> 
> It will be resolved by using tokenizer like [nlcst-parse-japanese](https://github.com/azu/nlp-pattern-match/tree/master/packages/nlcst-parse-japanese "nlcst-parse-japanese"), [parse-english](https://github.com/wooorm/parse-english "parse-english"), and [rakutenma](https://github.com/rakuten-nlp/rakutenma "rakutenma").
> But, It is not realistic and is not perfect. because toknizer is heavy weight(File size is large and Parse speed is slow)
>
> Toknizer example is [Text Tokenizer · Hivemall User Manual](https://hivemall.incubator.apache.org/userguide/misc/tokenizer.html "Text Tokenizer · Hivemall User Manual")
> 
> 
> Thanks.

ikatyang:
Fixes #3013 

Rules:

- CJK
  - put `line`(`" "` or `"\n"`) between `non-cjk` and `cjk-character`
  - put `softline`(`""` or `"\n"`) between `cjk` and `cjk`
  - put nothing between `non-cjk` and `cjk-punctuation`, i.e. they're considered not breakable
- emoji
  - they're considered 2-char width

Question:

- ~~Is `--split-cjk-text` an suitable option name?~~ removed the option and enabled it by default.

---

cc @azu as you might be interested in this.


tests/markdown_table/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -11,6 +11,28 @@ exports[`align.md 1`] = `
...
+`;
+
+exports[`emoji.md 1`] = `
+| abc | def | ghi |
+| --- | --- | --- |
+| 👍👍👍 | 👍👍👍 | 👍👍👍 |
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+| abc    | def    | ghi    |
+| ------ | ------ | ------ |
+| 👍👍👍 | 👍👍👍 | 👍👍👍 |
```
azz: image

These don't line up in my diff. Is this intentional?

azz:
Can we just turn splitting on and remove the option? If you don't want your text to wrap you can use `<!-- prettier-ignore -->` or `.prettierignore` if the whole file has CJK. There's minimal benefit to running prettier on the file if you don't want it to split the text for you.

tests/markdown_table/__snapshots__/jsfmt.spec.js.snap
ikatyang: It's based on the font, if you use [monospace-font](https://github.com/ikatyang/configs/tree/master/fonts) that supports CJK character then you'll see:

image

ikatyang:
I'm fine with it but most of the markdown converters do not support CJK splitting, e.g. GitHub.

src/util.js
```diff
@@ -636,7 +642,102 @@ function mapDoc(doc, callback) {
...
+  const nodes = [];
+
+  (!options.splitCjkText
+    ? text
+    : text.replace(
+        new RegExp(`(${cjkRegex.source})\n(${cjkRegex.source})`, "g"),
+        "$1$2"
+      )
+  )
+    .split(/(\s+)/g)
```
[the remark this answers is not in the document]
ikatyang: I think we can just treat full-width whitespace as CJK punctuation.

src/util.js
```diff
@@ -645,24 +645,25 @@ function mapDoc(doc, callback) {
...
-    .split(/(\s+)/g)
+  text
+    .replace(
+      new RegExp(`(${cjkRegex.source})\n(${cjkRegex.source})`, "g"),
+      "$1$2"
+    )
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
+    // `\s` but exclude full-width whitspace (`\u3000`)
+    .split(
+      /([ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\ufeff]+)/g
```
lydell: This regex can be written as `/([^\S\u3000]+)/`. Not sure if that's better. (It can even be written as `/((?:(?!\u3000)\s)+)/`, but that's definitely not better!)

(Also note that the `g` flag is not needed for `.split()`.)

ikatyang:
@azu any further comments? otherwise I'd like to merge this tomorrow.

azu:
@ikatyang Look good to me.

azz:
I'm not completely up-to-speed with the nuances of CJK chars but if the tests are passing I'm happy!

azz:
Oh, I thought you said @azz 😆

ikatyang:
azz vs azu 😆

PR #2976, opened 2017-10-04
[WIP] Refactor JSX to use intermediate representation

karl:
This is a currently broken work in progress refactor of how we handle JSX. It aims to introduce an intermediate representation that will make it easier for us to output following different formatting rules.

This should make it possible to re-use the same code for JSX, HTML, and FBT (Facebook translations).

---

The basic idea I wanted to explore here was an intermediate representation format.

The idea is to have a collection of alternating children and separators (one of space, new line, or nothing).

```jsx
<div><br /><br /></div>
children = [nothing, BR, nothing, BR, nothing];

---

<div> Some text <br /> </div>
children = [space, "Some text", space, BR, space];

---

<div>
  <br />
  <br />
</div>
children = [newline, BR, newline, BR, newline];
```

From this intermediate representation we could then generate the output using one of 3 sets of rules:

* JSX
* FBT
* HTML

When playing around with this the idea seemed to work OK, although if I recall correctly there are some edge cases that make it difficult to cleanly separate the `AST -> Intermediate -> Prettier doc` stages.

PR #3033, opened 2017-10-14
Implement getSupportInfo() and use it for inference

### LINKED ISSUE #3023: Expose list of supported parsers, etc.
Currently every time Prettier adds another language, a lot of editor integrations have to release another version of their plugin to support it. This might be adding more language types, extensions, etc.

I'm wondering if we'd be able to support some kind of feature detection within Prettier that editor integrations can call and run more dynamically. Something along the lines of:

```js
require('prettier').getSupportInfo();
```

Which might return:

```js
{
  languages: [
    {
      language: "javascript",
      parsers: ["babylon", "flow"],
      ids: ["javascript", "javascriptreact"],
      extensions: [".js", ".jsx", ".mjs"]
    },
    {
      language: "typescript",
      parsers: ["typescript"],
      ids: ["typescript", "typescriptreact"],
      extensions: [".ts", ".tsx"]
    },
    ...
  ]
}
```

Would be nice to release this alongside Markdown support.

Love to get some feedback from some plugin maintainers on whether this kind of thing could be useful.

cc. @robwise @CiGit @mitermayer

### LINKED ISSUE #3058: The experimental HTML support unexpectedly leaks through the CLI
Edit by @lydell:

```
prettier foo.html
```

The above formats foo.html as HTML. It should try to format it as JS (as usual with unknown file extensions, until  #2884 is resolved).

Original issue description:

`<script src="/build.js"></script>`

is being prettified to

`<script src="/build.js" />`

which prevents the script from running for some reason.

azz:
TODO

 - [x] CLI
 - [x] Docs

This PR adds a new function to the prettier API: `getSupportInfo()`.

It returns an object, currently with a `languages` property, which is an array of:

```js
{
  name: "SCSS", // display name for the language
  since: "1.4.0", // version it was added to prettier
  parsers: ["scss"], // built-in parsers that support this language
  group: "CSS", // metadata used to group related languages
  tmScope: "source.scss", // TextMate scope, used by prettier-atom
  aceMode: "scss",
  codemirrorMode: "css",
  codemirrorMimeType: "text/x-scss",
  extensions: [".scss"], // extensions used in parser inference
  files: [], // filenames without extensions, also used in parser inference
  liguistLanguageId: 329, // GitHub linguist language ID
  vscodeLanguageIds: ["scss"] // used by prettier-vscode
},
```

`getSupportInfo` accepts an optional argument, `version`, e.g. `prettier.getSupportInfo("1.4.0")`, this will filter the list to only include languages that version 1.4.0 of prettier support. This includes changing `less/scss/css` parsers back to `postcss`.

In addition, this table is now used by Prettier internally to perform parser inference based on filename and extension.

Maybe we should also add a CLI version?

Closes #3023 
Fixes #3058


src/support.js
```diff
@@ -0,0 +1,220 @@
...
+function getSupportInfo(version) {
+  if (!version) {
+    version = currentVersion;
+  }
+
+  const usePostCssParser = semver.lt(version, "1.7.1");
+
+  const languages = supportTable
+    .filter(language => {
+      return language.since && semver.gte(version, language.since);
```
ikatyang: Can be simplified to `language => language...`

tests_integration/__tests__/support-info.js
```diff
@@ -0,0 +1,17 @@
...
+describe("API getSupportInfo()", () => {
+  test("no arguments", () => {
+    expect(prettier.getSupportInfo()).toMatchSnapshot();
+  });
+
+  const testVersions = ["0.0.0", "1.0.0", "1.4.0", "1.5.0", "1.7.1", "1.8.0"];
+
+  testVersions.forEach(version => {
+    test(`with version ${version}`, () => {
+      expect(prettier.getSupportInfo(version)).toMatchSnapshot();
```
ikatyang: Can we only snapshot the `{ [language]: parsers }` information? It's hard to read its snapshot and other information seems unrelated here.

src/support.js
```diff
@@ -0,0 +1,220 @@
+"use strict";
+
+const semver = require("semver");
+const currentVersion = require("../package").version;
+
+// Based on:
+// https://github.com/github/linguist/blob/master/lib/linguist/languages.yml
+
+const supportTable = [
```
ikatyang: Maybe we should add a script to generate this table from `linguist` so that it'd be easy to update in the future.

ikatyang:
> Maybe we should also add a CLI version?

Yeah, the CLI seems necessary for non-JS editor, e.g. vim-prettier.

package.json
```diff
@@ -42,6 +42,7 @@
     "postcss-values-parser": "1.3.1",
     "remark-frontmatter": "1.1.0",
     "remark-parse": "4.0.0",
+    "semver": "^5.4.1",
```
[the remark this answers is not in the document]
azz: Has `save-prefix ""` been removed from `yarn`?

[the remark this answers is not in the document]
ikatyang: No, it works for me (v1.2.1).

src/support.js
azz: Maybe. I ended up changing quite a few things though, added more extensions, filenames, added groups, etc.

src/support.js
```diff
@@ -0,0 +1,217 @@
...
+      ".workbook"
+    ],
+    filenames: ["README"],
+    tmScope: "source.gfm",
+    linguistLanguageId: 222,
+    vscodeLanguageIds: ["markdown"]
+  },
+  {
+    name: "HTML",
+    since: undefined, // unreleased
```
suchipi: No plans to release this yet (in eg 1.8.0), right?

azz: Probably not. Depends on #2976

mitermayer:
Are there any plans on having this function exposed through the CLI ? I believe the info table supplied by the CLI does not  need to be so verbose so would probably be a much simpler output than the one returned by `getSupportInfo()`

azz:
What kind of output would you expect from the CLI version?

azz:
Ping @mitermayer, would JSON output or a parseable table be good for CLI?

mitermayer:
Hi @azz,

Vim has a function `json_decode()` however to better leverage that we need to make sure that the output of `getSupportInfo()` would be a plain json with nothing else, otherwise we would still need to use regex to extract the JSON node, and in that case a table would make more sense.

This is the documentation info on json_decode() function for vim 8:

```
json_decode({string})					*json_decode()*
		This parses a JSON formatted string and returns the equivalent
		in Vim values.  See |json_encode()| for the relation between
		JSON and Vim values.
		The decoding is permissive:
		- A trailing comma in an array and object is ignored.
		- More floating point numbers are recognized, e.g. "1." for
		  "1.0".
		However, a duplicate key in an object is not allowed. *E938*
		The result must be a valid Vim type:
		- An empty object member name is not allowed.
		- Duplicate object member names are not allowed.
```

mitermayer:
I actually am not sure we can leverage from the `json_decode` to parse arbitrary json. So I would suggest printing as a table since it would also be nicer for CLI users, and on the vim side we can use regex to extract info output

azz:
Let's just do 

```js
console.log(
  prettier.format(
    JSON.stringify(prettier.getSupportInfo()),
    { parser: "json" }
  )
);
```

?

mitermayer:
We can always start with that, then i can try to to add that to vim-prettier and see how that goes. Can always submit another PR later to allow the getSupportInfo() cli command to accept other output formats other than JSON if required

azz:
@mitermayer do you want to give this a try with json_decode before we merge?

mitermayer:
I won't be able to test this today, so feel free to merge it as it is! Can always wait for next release in case this is not sufficient.

Issue #1957, opened 2017-06-04
Preserve pre-comment newlines in long chained method calls


mafredri:
I believe this falls under the feature request category. Sometimes I use newlines to break long chained method calls into logical groups for (IMO) added clarity. I would like for prettier to preserve these newlines.

EDIT: I apologize if this has been discussed or reported before, I tried to search the issues but there are quite a few to go through.

**Input:**

```javascript
angular.module('AngularAppModule')
  // Constants.
  .constant('API_URL', 'http://localhost:8080/api')
  
  // App configuration.
  .config(appConfig)
  .run(appRun);
```

**Output:**

```javascript
angular
  .module('AngularAppModule')
  // Constants.
  .constant('API_URL', 'http://localhost:8080/api')
  // App configuration.
  .config(appConfig)
  .run(appRun);
```

**Expected:**

```javascript
angular
  .module('AngularAppModule')
  // Constants.
  .constant('API_URL', 'http://localhost:8080/api')

  // App configuration.
  .config(appConfig)
  .run(appRun);
```

As far as I see it, there are two ways to go about this:

1. Only allow a newline before a comment
2. Always allow one newline between chained method calls

Arguably, this is not a make-it-or-break-it feature, but one I would like to see implemented.

[Reproduction link](https://prettier.github.io/prettier/#%7B%22content%22%3A%22angular.module('AngularAppModule')%5Cn%20%20%2F%2F%20Constants.%5Cn%20%20.constant('API_URL'%2C%20'http%3A%2F%2Flocalhost%3A8080%2Fapi')%5Cn%20%20%5Cn%20%20%2F%2F%20App%20configuration.%5Cn%20%20.config(appConfig)%5Cn%20%20.run(appRun)%3B%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Atrue%2C%22trailingComma%22%3A%22es5%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%7D%7D)

vjeux:
Yeah, this is consistent with the rest of prettier, we should respect empty lines there as well.

jackyho112:
@azz 

Thanks for pinging me! I would love to give this a crack. Will look into it this weekend!

PR #3035, opened 2017-10-14
Account for empty lines in long member call chain

jackyho112:
Fixes https://github.com/prettier/prettier/issues/1957, so that for common long chained method call patterns, one empty line will be preserved before each call if there is any originally.

For example, 

```
angular.module('AngularAppModule')

  // Constants.
  .constant('API_URL', 'http://localhost:8080/api')
  
  // App configuration.
  .config(appConfig)
  .run(appRun);
```

will be formatted to 

```
angular
  .module("AngularAppModule")

  // Constants.
  .constant("API_URL", "http://localhost:8080/api")

  // App configuration.
  .config(appConfig)
  .run(appRun);
```

. If two calls are separated by a parenthesis, one of the empty lines outside the parentheses before the next call will be preserved.

Not sure if the decisions I made here are acceptable. 😅  I am happy to make any change regarding the implementation and how we should format this sort of styles. Please let me know if you have any improvement in mind. 😄

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -3796,9 +3796,8 @@ function printMemberChain(path, options, print) {
     shouldMerge ? groups.slice(1, 2)[0] : groups[0]
   ).node;
   const shouldHaveEmptyLineBeforeIndent =
-    lastNodeBeforeIndent.type !== "CallExpression"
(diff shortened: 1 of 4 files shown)


azz:
Thanks!

PR #3036, opened 2017-10-15
Print pipeline and nullish-coalescing operators

azz:
Adds printer for two new operators!

## Pipeline operator

Proposal: https://github.com/tc39/proposal-pipeline-operator/

This operator makes functional chains much easier to read.

```js
// Before
baz(bar(foo(x)));

// After
x |> foo |> bar |> baz;
```

## Nullish-coalescing operator

Proposal: https://github.com/tc39-transfer/proposal-nullish-coalescing

This operator is much like `||` except it only evaluates the RHS if the LHS is `null` or `undefined`, instead of falsy.

```js
const foo = object.foo ?? "default";
```

----

* Bumps babylon to beta 28

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -4359,11 +4359,17 @@ function printBinaryishExpressions(
       parts.push(path.call(print, "left"));
     }
 
-    const right = concat([
-      node.operator,
-      shouldInlineLogicalExpression(node) ? " " : line,
-      path.call(print, "right")
-    ]);
+    const shouldInline = shouldInlineLogicalExpression(node);
+    const lineBeforeOperator = node.operator === "|>";
+
+    const right = shouldInline
+      ? concat([node.operator, " ", path.call(print, "right")])
+      : concat([
+          lineBeforeOperator ? softline : "",
+          node.operator,
+          lineBeforeOperator ? " " : line,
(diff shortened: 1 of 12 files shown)


src/printer.js
```diff
@@ -4359,11 +4359,17 @@ function printBinaryishExpressions(
...
 
-    const right = concat([
-      node.operator,
-      shouldInlineLogicalExpression(node) ? " " : line,
-      path.call(print, "right")
-    ]);
+    const shouldInline = shouldInlineLogicalExpression(node);
+    const lineBeforeOperator = node.operator === "|>";
+
+    const right = shouldInline
```
azz: This change prints the `|>` operator on the LHS. All other operators are printed before the line break.

olstenlarck:
Nowhere in the release notes i didn't seen that we can disable that? Or am i not getting it right? It's pretty strange formatter to transform such things. Even more, they are stage 1. I really like them and i'm 99 sure that they will land in the language, but officially for now they are pretty risky.

suchipi:
Prettier's ability to format them is independent from your choice to use them or not. If your code has them, prettier will format them. If your code doesn't have them, prettier will keep working fine :)

Issue #3012, opened 2017-10-11
CSS/SCSS indenting


holloway:
This is a suggestion for a ~~change~~ **option** in formatting SCSS/CSS.

When reading JavaScript we all know how beneficial indentation can be to express nesting blocks, but in CSS there are also defacto and popular ways of expressing nesting that I think Prettier could automate.

[BEM](http://getbem.com/) codifies some defacto standards in naming conventions for CSS, in which classes are named as,

```scss
.block { /*props*/ } 

  .block__child1 { /*props*/ } 

    .block__child1--modifier1 { /*props*/ } 

    .block__child1--modifier2 { /*props*/ } 

  .block__child2 { /*props*/ } 

    .block__child2--modifier { /*props*/ } 
```

Here nesting is expressed in a naming convention *but also whitespace*, and when I'm reading CSS I often find that classNames have hierarchy expressed in naming conventions (with SCSS of course programmers can also nest selectors and use `&` but that's a separate issue).

**Would you be open to accepting a patch to indent CSS/SCSS blocks based on the previous selector?**

I think the formatting rule could be that if the previous selector is a prefix of the current selector then we consider that to be nesting and the block is nested a level. Any name that doesn't match would be formatted as Prettier currently does it (pushed left-most). Selectors would need to be contiguous to be indented. Of course there might be some details to work out but I just want to know if there's interest in this type of formatting for CSS/SCSS?

Just to be more specific...

**Input:**
```scss
.reposition {
	display: flex;
	align-items: center;
}

.reposition button[disabled] {
	opacity: 0.25;
}

.maps {
	display: block;
	overflow: auto;
}

.maps--disabled {
	opacity: 0.5;
	pointer-events: none;
}
```

**Current Output:**
```scss
.reposition {
  display: flex;
  align-items: center;
}

.reposition button[disabled] {
  opacity: 0.25;
}

.maps {
  display: block;
  overflow: auto;
}

.maps--disabled {
  opacity: 0.5;
  pointer-events: none;
}
```

**Expected behavior:**

**Output:**
```scss
.reposition {
  display: flex;
  align-items: center;
}

  .reposition button[disabled] {
    opacity: 0.25;
  }


.maps {
  display: block;
  overflow: auto;
}

  .maps--disabled {
    opacity: 0.5;
    pointer-events: none;
  }
```

suchipi:
[...]

PR #3038, opened 2017-10-15
CSS Hierarchy Indent option (off by default)

holloway:
First some stats about the level of popularity of this formatting:
* SitePoint surveys over two years of 5000 devs indicate about ~25% of CSS is written this way. [A Twitter poll](https://twitter.com/lucasazzola/status/933297415460872192) of 725 devs indicates that 10% of CSS is written this way, and because these are probably Prettier users this may indicate that 10% of devs disable Prettier CSS printing to maintain this style. So it's probably 10% hard support ranging to 25% soft support for this feature.
* It's very similar to the default output of Sass (both libsass and rubysass) and their devs say indenting hierarchy helps with debugging and readability.
* Recommended best practice by [CSSGuidelin.es](https://cssguidelin.es/#indenting) / Harry Roberts who's fairly well known in the CSS community

----------------------

As discussed in #3012 this indents adjacent styles based on implied or explicit hierarchies in the CSS selector. It can be enabled with the `--css-hierarchy-indent` flag (formerly the `--indent-styles`).

As this is my first PR to Prettier I'd appreciate as much feedback as you can give!

In terms of approach, the code makes `indent` wrappers around CSS Rules without disturbing the rest of the codebase (so when `indent-styles` is `false` there's minimal differences in the code path)

Previous formatting

```css
.reposition {
  display: flex;
  align-items: center;
}

.reposition button[disabled] {
  opacity: 0.25;
}

.maps {
  display: block;
  overflow: auto;
}

.maps--disabled {
  opacity: 0.5;
  pointer-events: none;
}
```
New formatting
```css
.reposition {
  display: flex;
  align-items: center;
}

  .reposition button[disabled] {
    opacity: 0.25;
  }

.maps {
  display: block;
  overflow: auto;
}

  .maps--disabled {
    opacity: 0.5;
    pointer-events: none;
  }
```

Previous formatting:
```css
table {
  background: papayawhip;
}

table td.inactive {
  background: palegoldenrod;
}

table tr {
  background: palegreen;
}

table tr.active {
  background: palegreen;
}

div {
  border-color: firebrick;
}
```
New formatting
```css
table {
  background: papayawhip;
}

  table td.inactive {
    background: palegoldenrod;
  }

  table tr {
    background: palegreen;
  }

    table tr.active {
      background: lime;
    }

div {
  border-color: firebrick;
}
```


azz:
Why would

```css
table tr.active
```

be a different indentation level to

```css
table td.inactive
```
?

holloway:
Because of the `tr` vs the `td`. That's why it's at the same level as 2nd CSS Rule in the document

azz:
So it's supposed to approximate DOM hierarchy? That seems difficult to define programatically.

holloway:
No, and I agree that would be too complicated. Instead it's (almost!) just treating selectors like strings and comparing preceding ones for infer indentation.

If we just look two levels deep into the hierarchy it looks like,

```
table
  table tr
  table td
```
...because it's not inferring elements, it's just that the a preceding selector starts with a `table `.

This also means that it works with common CSS naming conventions, such as:

```
.container {
  display:block;
}

  .container_col3 {
    width: 25%;
  }

  .container_col6 {
    width: 50%;
  }

    .container_col6--disabled {
      opacity: 0.5;
    }

  .container_col9 {
    width: 75%;
  }

```

holloway:
@azz sorry, I only realised that I misunderstand your question and I've updated my [2nd comment](https://github.com/prettier/prettier/pull/3038#issuecomment-336698447)

holloway:
Just to give a better example

I understand that Prettier's purpose is to create more uniformity in codebases by removing personal style (except for a few cases like `'` vs `"` etc.), and so CSS indenting could be viewed as a personal style...

...however I'd say that primarily it has a function (that 24% of people use, about 47% of those without Sass nesting use) and that is to group CSS and to [help catch typos early](https://github.com/prettier/prettier/issues/3012#issuecomment-339537099) by using naming conventions that devs already have, and as an example of that here's how this PR looks on a larger codebase: **here's all of Bootstrap** (the compiled output as plain CSS),

  https://gist.github.com/holloway/66cb1585cc11ac5c96848dfcca5d1c5c#file-bootstrap-with-indenting-css-L1414

Now of course this is just some arbitrary CSS, but on a codebase that wasn't organised with CSS Indenting in mind you can see [tables starting on line 1414](https://gist.github.com/holloway/66cb1585cc11ac5c96848dfcca5d1c5c#file-bootstrap-with-indenting-css-L1414) have hierarchies that are made clearer, and [`.form-control`s](https://gist.github.com/holloway/66cb1585cc11ac5c96848dfcca5d1c5c#file-bootstrap-with-indenting-css-L1648) also adhere to a useful grouping.

(by the way -- I should say that I'm sorry if I'm repeating myself @azz but I'm not quite sure how to progress this. E.g. with your emoji response on the issue thread I'm not sure if I should explain the PR code in detail, or give high-level examples like this, or wait until after Prettier 1.8 is out, or promise to maintain the code for years (i will!) etc.)

lipis:
Can we pass on this one?!

```css
table {
  background: papayawhip;
}

  table td.inactive {
    background: palegoldenrod;
  }

  table tr {
    background: palegreen;
  }

    table tr.active {
      background: lime;
    }

div {
  border-color: firebrick;
}
```

and usually with SASS, or Less you would nest those and in those cases nobody cares about the output as it's usually not in the codebase..

holloway:
@lipis Yep and in the issues thread you can see some stats around preprocessor nesting usage which indicates that nesting in Sass/Less is slightly less popular than not nesting or using manual CSS indenting (52% vs. 44%), which probably means that about half of all Sass/Less codebases don't nest much, and as you rightly imply there's always plain CSS which can't nest.

Personally I think that managing large JavaScript codebases gets a lot of discussion in the community (older yet essential books like JavaScript: The Good Parts saying to use a subset of the language so don't use `with`, etc), but CSS is a simpler and yet also difficult language with it's global (non-scoped) selectors and `!important` and such, so managing larger Sass/Less/CSS codebases is also often about removing problematic features. There are parts of the CSS community who consider nesting to be an anti-pattern as it encourages overly specific selectors (e.g. just like how there's 'JavaScript: The Good Parts', some people use a subset of Sass/Less, perhaps just variables and mixins...).

Originally I just wanted this issue considered so I'm happy either way on how you judge this feature request. Again, thanks for you time.

README.md
```diff
@@ -363,6 +363,12 @@ table {
 ```
 
 This option is off by default.
+=======
```
lydell: Did you mean to add these equal signs here?

holloway: Sorry that was a leftover from a merge conflict, now removed!

(there are linting errors too that I can't reproduce on my local machine... I'll try to clean up this PR over the next day)

lydell:
Interesting related reading: stylelint used to have this feature (https://github.com/stylelint/stylelint/issues/278, https://github.com/stylelint/stylelint/pull/305), but it was removed (https://github.com/stylelint/stylelint/issues/1309, https://github.com/stylelint/stylelint/pull/1385).

holloway:
Thanks, that is interesting! I wonder where they were getting their stats from and if they agree with SitePoint survey.

From reading through the Stylelint code they seem to have a very sophisticated (and comparatively complicated) approach to hierarchy. The comparable code for indenting in this PR is in [getIndentTree](https://github.com/prettier/prettier/pull/3038/files#diff-f1e024c851619765fd816942326795c1R588), and regarding the code complexity (in general) I was trying to minimise the amount of codepath differences with [a pass-through function](https://softwareengineering.stackexchange.com/questions/275490/name-for-a-function-which-returns-its-arguments) so hopefully that issue doesn't apply, but obviously that's your call about whether it's justified in this case.

lipis:
I agree with Stylelint, it looks weird to me to indent CSS (or SASS, Less) files. My point was that if the style get's too big, usually it's broken to multiple files to organize it better, so there is no need to indent to group the related classes.

holloway:
Upgraded to Prettier 1.8.0 and now all tests pass.

ryami333:
@lipis to me this type of indentation is essentially just a more explicit alternative to nesting with a pre-processor, which is wildly popular. Using an ampersand makes sense to some people (and this keeps out of their way, because this formatting is behind an option/flag), but to others it just doesn't. To those other people, the ampersand just represents an unnecessary abstraction which results in selectors that you can't 'search' your codebase for.

I agree that stylesheets should be split when they get too big, but I don't really see how that factors into this conversation at all. Using that logic you could argue that we wouldn't need prettier at all if we would just refactor _all_ of our code into small enough modules.

holloway:
Just adding more credence to indenting, Sass itself (both libsass and rubysass) have an output formatting option called `nested` which indents the _output CSS_ in a very similar way.

[Their docs](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#_nested) describe it as:

> Nested style is the default Sass style, because it reflects the structure of the CSS styles and the HTML document they're styling. Each property has its own line, but the indentation isn't constant. Each rule is indented based on how deeply it's nested.
>
> Nested style is very useful when looking at large CSS files: it allows you to easily grasp the structure of the file without actually reading anything.

(I'm not so sure about the '_without actually reading anything_' part but you get the idea!)

The source file needs to be formatted using Sass nesting and then the output CSS is indented like this,

```css
.reposition {
  display: flex;
  align-items: center; }
  .reposition button[disabled] {
    opacity: 0.25; }

.maps {
  display: block;
  overflow: auto; }
  .maps .maps--disabled {
    opacity: 0.5;
    pointer-events: none; }
```

Personally I think a few more linebreaks between the CSS Rules would help readability, but regardless this seems to be a major example of a similar indenting style done for similar reasons of readability and seeing hierarchy at a glance.

src/printer-postcss.js
```diff
@@ -538,4 +565,59 @@ function maybeToLowerCase(value) {
...
+  // Stringifies selectors for comparison patterns.
+  // Doesn't stringify to normal CSS selectors, because it's just used to infer
+  // indenting depth.
+
+  const SEPERATOR = "\u0000"; // NUL invalid in CSS so it's a safe separator
+  return selector.nodes
+    .map(
+      node =>
+        `${SEPERATOR}${node.type}${SEPERATOR}${node.value}${
+          node.type === "selector-tag" // Element "train" preceded by "tr" shouldn't be indented
```
lydell: Perhaps a real-word example is clearer?

```
// Element "section" preceded by "s" shouldn’t be indented
```

src/printer-postcss.js
```diff
@@ -538,4 +565,59 @@ function maybeToLowerCase(value) {
     : value.toLowerCase();
 }
 
+function getSelectorPattern(selector) {
+  // Stringifies selectors for comparison patterns.
+  // Doesn't stringify to normal CSS selectors, because it's just used to infer
+  // indenting depth.
+
+  const SEPERATOR = "\u0000"; // NUL invalid in CSS so it's a safe separator
```
lydell: Typo: `SEPERATOR` → `SEPARATOR`

```
const SEPARATOR = "\0"; // NUL is invalid in CSS, so it's a safe separator.
```

tests/css_indent_styles/dump.css
```diff
@@ -0,0 +1,52 @@
...
+  table {
+    tr {
+      background: papayawhip;
+
+      td {
+        background: papayawhip;
+      }
+    }
+  }
+}
```
lydell: I would like to see crazy SCSS and Less tests as well.

- What happens when an indented rule has nested rules?
- What happens when SCSS/Less/PostCSS interpolations are used in selectors?
- What about pseudo-classes (`:hover`) and pseudo-elements (`::after`)?
- What about selectors with commas? `table th, table td`?

And some case insensitiviy stuff. I guess `TABLE TR` should be indented under `table`?

docs/options.md
```diff
@@ -165,3 +165,47 @@ docblock at the top of the file then this option will add a newline to it with t
...
+    background: lime;
+  }
+```
+
+Advocates of CSS Indenting say it soon becomes easier to read because top-level selectors are like section headings,
+and that it can help developers organise CSS and identify hierarchy bugs. We'll let you be the judge.
+
+Default | CLI Override | API Override
+--------|--------------|-------------
+`false` | `--indent-styles` | `indentStyles: <bool>`
```
lydell: We need to add this to the CLI code as well.

azz:
Gotta admit, I'm still completely confused as to how it doesn't cause massive git diffs.

From your example:

```css
table {
  background: papayawhip;
}

  table tr {
    background: palegreen;
  }

    table tr.active {
      background: lime;
    }
```

If I delete the `table tr` block:

```patch
-  table tr {
-    background: palegreen;
-  }
-
```

Then format, and the final diff will be:

```patch
-  table tr {
-    background: palegreen;
+  table tr.active {
+    background: lime;
   }
-
-    table tr.active {
-      background: lime;
-    }
```

Similarly, the reverse has the same problems. Start with:

```css
table {
  background: papayawhip;
}

  table tr.active {
    background: lime;
  }
```

Then add in between the two rules:

```patch
+  table tr {
+    background: palegreen;
+  }
+
```

Format, then the final diff is:

```patch
-  table tr.active {
-    background: lime;
+  table tr {
+    background: palegreen;
   }
+
+    table tr.active {
+      background: lime;
+    }
```

Both of those diffs are incredibly unfriendly for such a small change.

lydell:
@azz That’s a really good point. A withdraw half of my +1. :+0.5:

holloway:
Good point @azz, and it certainly has that ramification to diffs. There are relationships between CSS Rules that are being shown and changed, and there's no avoiding the need to reformat the 'child' branch now that 'parent' is gone. The flow-on effect of indenting (and diffs) would only affect 'child' CSS rules, and any other adjacent rules would be left as-is. To some degree it's like nesting in almost any language (XML, JS, Sass, SAX output, etc) when removing an intermediary level, however specific to CSS is that the hierarchy hasn't changed it's just our visualisation of it. If we transpose your example of _deletion_ to nesting with a preprocessor there are similar issues, and we'd also manually replace `&` with `tr`, and likewise the hierarchy hasn't changed but we still need to update adjacent lines and indentation levels:

```diff
table {
-  tr {
-    background: papayawhip;
-    &.active {
+  tr.active {
-      background: lime;
+    background: lime;
-    } 
  }
}
```

Ultimately I think your diff, and the Sass diff, is what people who use those would expect. In the opposite scenario of a file that contains `table` and `table tr.active` then _inserting_ a `table tr` between them then indenting the `table tr.active` would also be expected (to show that `table tr.active` is more specific than `table tr`). Affecting adjacent lines is a feature, not a bug!

tests/css_indent_styles/dump.css
holloway: Awesome, thanks @lydell! I'll try to address your feedback over the next few days. Commas are already supported, and the [Bootstrap output CSS formatting example has some examples](https://gist.github.com/holloway/66cb1585cc11ac5c96848dfcca5d1c5c#file-bootstrap-with-indenting-css-L150)  (and of course that's just some arbitrary CSS I chose, I don't imagine that people would normally format their output).

Just to add another question to the list:

* What happens with code comments?

holloway: For performance testing is there a tool that Prettier devs usually use? If there's nothing currently I could use my [whats-the-damage](https://www.npmjs.com/package/whats-the-damage) tool if that's ok... it can measure durations, CPU, and memory usage.

suchipi:
I'm okay with including this as long as @lydell is okay with including this, because he's most familiar with the CSS printer, and my main concern in regard to adding options is maintenance cost. Based on what he said, the maintenance cost shouldn't be too high, because this is relatively isolated- that said, I also wonder about performance.

tests/css_indent_styles/dump.css
lydell: Don't focus to much on the performance thing. It's enough to just run the CLI on a big piece of CSS with and without the option to see if there's a huge difference. For example, a mistake in the selector-checking loop could cause a massive slowdown, but if it's all correct we should be fine.

Also not that my questions in the bullet points in my above comments were intended more to be requests for additions to the tests, rather than just questions. Sorry for being unclear.

holloway: I've just pushed a new version with CLI mode, comment indenting, nested styles, and more tests. The indenting or passthrough code is now a single line of code too.

For nested styles the code skips indentation changes for that block, but it retains any indentation from the outer selector. It's hard to explain but I think it looks right,

before:
```css
div {
  background: papayawhip;
}

div s {
  background: palegreen;
}

div s.ghost {
  background: palegoldenrod;

  .coast {
      background-color: gray;
  }
}

div s.dancer {
  // this should be indented with previous nested style
  background: palevioletred;
}

div section {
  // this shouldn't be indented with "div s" because although it starts with the same characters
  // we treat element name as atomic.
  background: papayawhip;
}
```

after
```css
div {
  background: papayawhip;
}

  div s {
    background: palegreen;
  }

    div s.ghost {
      background: palegoldenrod;

      .coast {
        background-color: gray;
      }
    }

    div s.dancer {
      // this should be indented with previous nested style
      background: palevioletred;
    }

  div section {
    // this shouldn't be indented with "div s" because although it starts with the same characters
    // we treat element name as atomic.
    background: papayawhip;
  }
```

So for `div s.ghost` that outer selector for the nested style is used for indenting, but the inner ones are ignored and they just use the existing Prettier formatting. If it didn't skip the `.coast` then I'm not sure what would be expected...the code might have to resolve the full selector and perhaps resolve `&` and so on, which (I think) would make the code far too complex.



The comment indenting looks like this,

before
```css
a {
    text-decoration: none;
}
// a comment that should be indented at same column as following selector
a:hover {
    text-decoration: underline overline; 
}
```

after
```css
a {
  text-decoration: none;
}
  // a comment that should be indented at same column as following selector
  a:hover {
    text-decoration: underline overline;
  }
```

This way if there's an indentation shape any comments will adhere to it. This example also shows pseudo classes, and pseudo elements work the same way.

For performance just using `time` on the command line it's averaging 0.254s without indent-styles and 0.267s with.

> And some case insensitiviy stuff. I guess TABLE TR should be indented under table

This makes sense per-spec for HTML, but I suppose there are extremely niche cases like [XML CSS](http://www.xmlplease.com/xml/xmlcss/) that afaik are case-sensitive. More popular would be [Attribute values](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) which seem to be case-sensitive (`a[data-amiga="cOmMoDoRe"]` !== a[data-amiga="CoMmOdOrE"]`) unless there's an `i` flag. These per-spec rules are somewhat complex, and I wonder if people writing `TABLE TR` followed by `table tr td` is also a niche case to support. Maybe this aspect of the code should be left as-is to keep the code simpler in `getSelectorPattern`, even if it is wrong per-spec. I don't have a strong opinion here though so I'll make elements insensitive with the next PR revision unless you suggest otherwise.

TODO:

* More tests for every CSS node type appearing in the file.
* Confirm whether any of these nodes should be indented too. I'm hoping not because I wouldn't want to affect much more of the main code-paths. The current test file has a Sass variable declaration between two selectors 
* "What happens when SCSS/Less/PostCSS interpolations are used in selectors?" 
* Anything else?

I'll try to get a new version in the next week.

Thanks so much Lydell!

lydell: Re. `.coast`: That's the expected output. I wonder if it is possible to trick the indenter into indenting too much if you use something like this?

```css
div {
  div.foo {}
}
div.foo.bar {}
```

The expected output is:

```css
div {
  div.foo {}
}
  div.foo.bar {}
```

and not:

```css
div {
  div.foo {}
}
    div.foo.bar {} // A mistake in the logic could cause double indentation here. 
```

Re. performance: Great, nothing to worry about there, I guess.

Re. case insensitivity: I totally understand that it is an edge case, but Prettier already lowercases stuff, so we _have_ to do it in order to make the formatting stable. We don't want this to happen:

```css
TABLE {}
table.zebra {}
```
→
```css
table {}
table.zebra {}
```
→
```css
table {}
  table.zebra {}
```

lipis:
In case this one will make it to a release, I guess there will be an option to enable it and by default it will be disabled.

lydell:
Regarding the gotcha around causing large-y diffs, I think @holloway's example makes sense. When writing SCSS, it happens every now and then that I split up or join nested rules due and therefor cause indentation-only changes of some blocks of code, just like in the example. So this option is not much worse than using SCSS.

holloway:
@lipis Yes the proposal is for it to be an option that's disabled by default (see [vjeux's feedback and 4.](https://github.com/prettier/prettier/issues/3012#issuecomment-336312637))

holloway:
(just copying my comment here for visibility because I didn't realise that I posted it on a hidden outdated diff...sorry for the Github notifications spam)


I've just pushed a new version with CLI mode, comment indenting, nested styles, and more tests. The indenting or passthrough code is now a single line of code too.

For nested styles the code skips indentation changes for that block, but it retains any indentation from the outer selector. It's hard to explain but I think it looks right,

before:
```css
div {
  background: papayawhip;
}

div s {
  background: palegreen;
}

div s.ghost {
  background: palegoldenrod;

  .coast {
      background-color: gray;
  }
}

div s.dancer {
  // this should be indented with previous nested style
  background: palevioletred;
}

div section {
  // this shouldn't be indented with "div s" because although it starts with the same characters
  // we treat element name as atomic.
  background: papayawhip;
}
```

after
```css
div {
  background: papayawhip;
}

  div s {
    background: palegreen;
  }

    div s.ghost {
      background: palegoldenrod;

      .coast {
        background-color: gray;
      }
    }

    div s.dancer {
      // this should be indented with previous nested style
      background: palevioletred;
    }

  div section {
    // this shouldn't be indented with "div s" because although it starts with the same characters
    // we treat element name as atomic.
    background: papayawhip;
  }
```

So for `div s.ghost` that outer selector for the nested style is used for indenting, but the inner ones are ignored and they just use the existing Prettier formatting. If it didn't skip the `.coast` then I'm not sure what would be expected...the code might have to resolve the full selector and perhaps resolve `&` and so on, which (I think) would make the code far too complex.



The comment indenting looks like this,

before
```css
a {
    text-decoration: none;
}
// a comment that should be indented at same column as following selector
a:hover {
    text-decoration: underline overline; 
}
```

after
```css
a {
  text-decoration: none;
}
  // a comment that should be indented at same column as following selector
  a:hover {
    text-decoration: underline overline;
  }
```

This way if there's an indentation shape any comments will adhere to it. This example also shows pseudo classes, and pseudo elements work the same way.

For performance just using `time` on the command line it's averaging 0.254s without indent-styles and 0.267s with.

> And some case insensitiviy stuff. I guess TABLE TR should be indented under table

This makes sense per-spec for HTML, but I suppose there are extremely niche cases like [XML CSS](http://www.xmlplease.com/xml/xmlcss/) that afaik are case-sensitive. More popular would be [Attribute values](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) which seem to be case-sensitive (`a[data-amiga="cOmMoDoRe"]` !== a[data-amiga="CoMmOdOrE"]`) unless there's an `i` flag. These per-spec rules are somewhat complex, and I wonder if people writing `TABLE TR` followed by `table tr td` is also a niche case to support. Maybe this aspect of the code should be left as-is to keep the code simpler in `getSelectorPattern`, even if it is wrong per-spec. I don't have a strong opinion here though so I'll make elements insensitive with the next PR revision unless you suggest otherwise.

TODO:

* More tests for every CSS node type appearing in the file.
* Confirm whether any of these nodes should be indented too. I'm hoping not because I wouldn't want to affect much more of the main code-paths. The current test file has a Sass variable declaration between two selectors 
* "What happens when SCSS/Less/PostCSS interpolations are used in selectors?" 
* Anything else?

I'll try to get a new version in the next week.

Thanks so much Lydell!

lydell:
Well, then I'll have to move my comment as well.

Re. `.coast`: That's the expected output. I wonder if it is possible to trick the indenter into indenting too much if you use something like this?

```css
div {
  div.foo {}
}
div.foo.bar {}
```

The expected output is:

```css
div {
  div.foo {}
}
  div.foo.bar {}
```

and not:

```css
div {
  div.foo {}
}
    div.foo.bar {} // A mistake in the logic could cause double indentation here. 
```

Re. performance: Great, nothing to worry about there, I guess.

Re. case insensitivity: I totally understand that it is an edge case, but Prettier already lowercases stuff, so we _have_ to do it in order to make the formatting stable. We don't want this to happen:

```css
TABLE {}
table.zebra {}
```
→
```css
table {}
table.zebra {}
```
→
```css
table {}
  table.zebra {}
```

tests/css_indent_styles/dump.css
holloway: Thanks, so I'll add that indenter trick as a test (currently the indenter passes that test).

For case insensitivity.. that makes a lot of sense, so I'll definitely do it and try to adhere to the same normalization rules that Prettier already has.

holloway:
Thanks, so I'll add that indenter trick as a test (currently the indenter passes that test).

For case insensitivity.. that makes a lot of sense, so I'll definitely do it and adhere to the same normalization rules that Prettier already has.

lipis:
Awesome..

tests/css_indent_styles/indent-styles.scss
```diff
@@ -81,4 +81,9 @@ a::after {
...
 // another comment
 /* whatever 
-do does*/
\ No newline at end of file
+do does*/
+
+div {
+  div.foo {}
+}
+div.foo.bar {} // should only be indented against div not div.foo
```
lydell: Can you add a newline at then end of this file, please? That'll make the snapshot easier to read.

lydell:
You can see other things besides tag names that we lower-case here:

- https://github.com/prettier/prettier/blob/master/tests/css_case/case.scss
- https://github.com/prettier/prettier/blob/master/tests/css_case/case.less

It's probably also a good idea looking at the `maybeToLowerCase` function and its usages: https://github.com/prettier/prettier/blob/2332c5c8bf777fb5c78de4c9fa92749a8373d6aa/src/printer-postcss.js#L531-L539

holloway:
Given that the normalization code is somewhat complicated do you think I should try to reuse the CSS normalization code, or duplicate it? I tried to reuse it by making `getSelectorPattern()` call `genericPrint(path, options, print)` to stringify the CSS selector as needed, but --because  `getIndentStyleTree()` needs the previous or next selectors-- the `path` that I'm passing to `genericPrint()` is for the current node not the previous or next ones. Then I found `FastPath.call` which seems to insert items on the FastPath stack, but those are resolved by selecting strings on `path` (`name1, name2`  ➜ `path[name1], path[name2]`) rather than arbitrary nodes. Next I tried making a new `FastPath.prototype.pushCall` to put arbitrary items on the stack (and restore the stack, like `.call`) but that didn't work either. Am I heading down the wrong path? (so to speak)

The selector nodes that I'm trying to stringify are ones like `selector-selector`, `selector-combinator`, `selector-id`... and I think I need those as individual nodes so that I can do different things based on their type, such as keeping `selector-tag` atomic (the `s` vs `section` example). I'm hoping there's a cleaner way that I just don't know about. Any ideas?

lydell:
Really good questions. Unfortunately, I don't have any good answers. But I do have further complications – quotes normalization: `[attr="value"]`, `[attr='value']` and `[attr=value]`.

holloway:
No probs, those complications are still useful to know because (I think) it favours the idea reusing existing code. I'll continue with that approach for now.

Thanks!

holloway:
Phew, almost there! The code now reuses the existing normalization in `genericPrint()` so it supports your quotes normalization scenario. b54e786 retains the previous test and snapshot to show that there were no regressions overall, and 9730af8 adds a new test with those attribute quotes.

Specifically in `getSelectorPattern()` it uses a new `pushCall` method on the `path` to call `genericPrint()` with a `childPath`. I added pushCall to the bottom of printer-postcss.js rather than fast-path.js (as a prototype) just because I'm trying to keep these changes out of the way.

I'll go through any remaining issues over the next few days. Cheers :)

holloway:
Ok the code now has tests for variable interpolation, and _I think_ that's all of the specific issues responded to. Are there any other things this PR should deal with @lydell?

holloway:
Just because there's been some changes in codepaths I thought I'd redo the benchmarks and fyi they're basically the same in terms of ratios (average of 10 runs with option enabled is 0.305s, disabled is 0.297s, and for comparison master is also 0.297s). This is only with CSS formatting of course.

In terms of maintenance cost (the point raised by @suchipi) I'm hoping that it's minimal because it's now reusing the existing selector normalization code, and all the changes are confined to the bottom of the `printer-postcss.js`, and as said I'm more than happy to maintain this bit of the code. If anyone has suggestions for how to improve maintainability (more comments?) please let me know.

Thanks! :smile:

src/cli-constant.js
```diff
@@ -167,6 +167,12 @@ const detailedOptions = normalizeDetailedOptions({
     default: ".prettierignore",
     description: "Path to a file with patterns describing files to ignore."
   },
+  "indent-styles": {
+    type: "boolean",
+    category: CATEGORY_FORMAT,
+    forwardToApi: true,
+    description: "Indent adjacent CSS Rules based on CSS selector hierarchy"
+  },
```
lydell: Sorry for the nitpick, but the options here should be sorted alphabetically.

src/printer-postcss.js
```diff
@@ -34,35 +34,51 @@ function genericPrint(path, options, print) {
       if (n.raws.content) {
         return n.raws.content;
       }
+      const isIndentStyle = options.indentStyles
```
lydell: `isIndentStyle` sounds like a boolean to me, but it seems to be a function.

src/cli-constant.js
holloway: No problem changing this of course but I don't quite understand which part you're referring to...I guess you're referring to the "type", "category", "forwardToApi" "description" but other options ("insert-pragma") don't follow this ordering, so could you confirm that I should change that? The string "indent-styles" is alphabetically between "ignore-path" and "insert-pragma", so I guess you don't mean that one?

src/printer-postcss.js
holloway: Yep it's a function. Would `indentStyles` or perhaps `isIndentStyleFn` be a better name, even when `options.indentStyles` is `false`?

I suppose I was thinking about the later use of the function when it's wrapping `isIndentStyle(concat(...))` to show that it might be doing nothing (just using a pass-through) when the option is `false`. Originally I didn't have a pass-through and it was a conventional `const formatted = concat(/* the existing concat */); return options.indentStyles ? getIndentStyle(path, options, print)(formatted) : formatted;` but for the `css-comment` area this would mean more duplicate code.

src/cli-constant.js
lydell: Sorry, I was very unclear. This is what I meant: https://github.com/prettier/prettier/blob/c0fa47d2d26eb211156a1628d9370efb36c6afc0/src/cli-constant.js#L77

src/printer-postcss.js
duailibe: `maybeIndentStyle` ?

src/cli-constant.js
holloway: Hmm... sorry, I feel really dumb because I must be missing something. The string "**ind**ent-styles" is already alphabetically between "**ign**ore-path" and "**ins**ert-pragma" is it not?

```js
> ["indent-styles", "ignore-path", "insert-pragma"].sort()
["ignore-path", "indent-styles", "insert-pragma"]
```

Very sorry about this because I'm trying to adjust the code but I'm a bit confused by this.

src/printer-postcss.js
holloway: Changed!

holloway:
Hmm.. I notice that there's [a poll on this indenting PR](https://twitter.com/lucasazzola/status/933297415460872192) retweeted by the new Prettier twitter (tweeted originally by Azz) that's not linked here... so there you go, there's the link :) Certainly it seems that people on twitter don't like it very much compared to the other data points that has been posted.

Semi-related, but I had a look at the previous year's SitePoint CSS survey from March 2016 in which they received 3439 responses, and 30% of respondents use CSS Indenting

css-indenting-in-2016

ryami333:
@azz I'm interested to know the purpose of the twitter survey you've created? If it's to assess the appetite for this PR, then I don't think that that is very fair. We've provided some much fairer and more formal statistics which prove (even by the most cynical interpretation) that there is a significant appetite for this formatting, even though it's not a majority (but that's why it's an opt-in feature) as well as one or two objective benefits of such an approach. I'm worried that the first couple of comments on the survey are just going to set the tone for all of the remaining commentary, and that as a result Prettier's users are just going to be alienated against this formatting before they've even had a chance to try it.

docs/options.md
```diff
@@ -193,3 +193,25 @@ By default, Prettier will wrap markdown text at the specified print width. In so
 | Default | CLI Override      | API Override        |
 | ------- | ----------------- | ------------------- |
 | `true`  | `--no-prose-wrap` | `proseWrap: <bool>` |
+
+## Indent Styles
```
ikatyang: I think we need some discussion about the option naming, it's really confusing me what does `--indent-styles` mean at first glance, I personally would recommend something like `--css-hierarchy-indent`.

And there should be an `_available in v1.9.0+_` or something indicating that's an unreleased option.

docs/options.md
```diff
@@ -193,3 +193,25 @@ By default, Prettier will wrap markdown text at the specified print width. In so
...
+
+This indents adjacent CSS Rules based on their selectors naming conventions. This works well with popular CSS naming conventions such as [BEM](http://getbem.com/) or any naming convention that uses prefixes to indicate hierarchy.
+
+For example, a file with the following CSS will be indented as:
+
+```
```
ikatyang: ` ```css `?

src/cli-constant.js
```diff
@@ -167,6 +167,12 @@ const detailedOptions = normalizeDetailedOptions({
     default: ".prettierignore",
     description: "Path to a file with patterns describing files to ignore."
   },
+  "indent-styles": {
+    type: "boolean",
+    category: CATEGORY_FORMAT,
+    forwardToApi: true,
+    description: "Indent adjacent CSS Rules based on CSS selector hierarchy"
```
ikatyang: There should be an ending dot for consistency.

docs/options.md
holloway: Good point. I'll do that. I certainly agree that `indent-styles` is a bit unnecessarily confusing. I chose "styles" to encompass Sass/Less/CSS but "css" is probably simpler and clearer language. I've just searched through the various links posted in this PR/Issue and "indenting" is the most popular term amongst the community, with "hierarchy" being a close second (although it isn't mentioned in [guides like this](https://cssguidelin.es/#indenting)), so your suggestion makes sense to me.

docs/options.md
holloway: Unfortunately `css` was causing my Prettier plugin to flatten the example CSS (I have a feeling the CI rejected it too). So while it's not nicely syntax-highlighted CSS I'm guessing that I would need to change this later. Is that right?

ikatyang: `<!-- prettier-ignore -->`

lipis:
@holloway TBH I'm not sure if we would have added the same question with the example as in the poll on Twitter, I'm sure there would be more negative answers..

We maybe have to rethink the whole thing if it's worth it to include it in the Prettier and keep the code simpler (because the cost of maintaining it is still going to be high), because it looks like the majority wouldn't style their CSS like this..

https://twitter.com/lucasazzola/status/933297415460872192

holloway:
@ikatyang Ok I've renamed a lot of things! Does that look right now?

holloway:
@lipis Well... maybe we should agree to disagree? I think that two years of responses on SitePoint from 3,400 devs and 1,600 devs is probably statistically significant and that indicates about 25-30% of all CSS is written this way, which means devs are tediously and manually indenting to achieve what this PR can automate with the help of Prettier. That's who I'm trying to help. The Twitter poll of 215 people is valid too of course, but it's obviously a smaller sample size.
In terms of maintenance cost that's always been on my mind, and I've tried to isolate the changes as much as I can, reuse selector normalization, and offer to support it myself for years. As to whether that's enough that's not my call, but I'd urge you to read the Sass developers comments about seeing hierarchy and debugging.
Thanks for your comments. It helps having some somewhat negative views (because all features need to answer those kinds of questions) and the stats that I'm showing agree with the idea that it's not for everyone. I hope we can help 25-30% who use it though :)

lipis:
Point taken :)

holloway:
Cool, thanks dude :)

src/cli-constant.js
lydell: > sorry, I feel really dumb because I must be missing something.

No, it is I who is dumb – can't even do the alphabet anymore … sorry, you are correct. Nothing to do here :)

thibaudcolas:
I just made a repo with some real-world code using this formatting if anyone wants to further play with it: https://github.com/thibaudcolas/prettier-indent-styles.

To me formatting code like this is useful in a codebase _with a preprocessor_ (and BEM/ITCSS), to discourage nesting and keep specificity in check. Preprocessor nesting and `&` interpolation are easily abused, and make individual nested selectors appear deceptively short compared to their actual specificity in the resulting CSS. This formatting makes it easier to break up styles into separate blocks, while perserving the visual aid of nesting.

Here is an example, shorter than what real-world styles would look like but still demonstrates what problem this helps address:

```sass
// The pyramid of doom in CSS land. Menus are a classic .
.nav-list {
  li {
    display: inline;
    a {
      text-transform: none;

      // This will become `.nav-list li a.active`.
      &.active {
        text-transform: underline;
      }
    }
  }
}
```

```sass
// Breaking up the pyramid.
.nav-list {
  li {
    display: inline;
  }
}

  // This makes it clear the `li` is actually not needed.
  .nav-list li a {
    text-transform: none;
    &.active {
      text-transform: underline;
    }
  }
```

<details>
    <summary>BEM equivalent</summary>

```sass
// The BEM pyramid of doom.
.nav-list {
  &__item {
    display: inline;
    &__link {
      text-transform: none;
      &--active {
        text-transform: underline;
      }
    }
  }
}
```

```sass
// Breaking up the pyramid.
.nav-list {
  &__item {
    display: inline;
  }
}

  .nav-list__link {
    text-transform: none;
    &--active {
      text-transform: underline;
    }
  }
```

</details>

For diffs, I haven't suffered much from this. The changes caused by this are only whitespace, which diff tools often ignore. Preprocessor nesting has the same drawback, except you also need to replace the `&` interpolation.

Here's what I get for the example above with `-w`:

```diff
table {
   background: papayawhip;
 }

-  table tr {
-    background: palegreen;
-  }
-
   table tr.active {
     background: lime;
   }
```

The main drawback of this formatting is that doing it manually takes lots of tedious indenting and de-indenting to maintain the visual nesting as you refactor selectors. But that's exactly why it would be good to have this automated.

azz:
I wanted to get a (relatively) unbiased sampling of opinions on Twitter. I have a suspicion that the SitePoint survey suffers from people reading "Do you indent your CSS to match HTML hierarchy" and clicking "Yes" (because they use a postprocessor) without reading the responses carefully. People often whiz through those surveys really quickly, and it didn't have an image to describe what people were voting for.

> I'm worried that the first couple of comments on the survey are just going to set the tone for all of the remaining commentary

I don't think that really happens - people are always keen to argue things on the internet. 

---

I suppose my main concern with this is that it's an alternative style, when Prettier historically has only supported one "opinionated" style. The options introduced are for highly controversial things (tabs vs spaces, quotemarks, etc), and none of them significantly affect formatting (tabWidth excluded).

Every now and again, someone will ask to format vars like this:

```js
var name     = 1;
var longName = 2;
```

Even though that style is perfectly valid and somewhat popular (perhaps more popular than this CSS style), we haven't implemented it to encourage consistency. The variable alignment style also suffers git diff problems when automated, same as this PR.

azz:
Twitter results are in:

[image](https://twitter.com/lucasazzola/status/933297415460872192)

If you take out the 14% who voted for <kbd>(show results)</kbd>, **90% don't indent CSS**.

Not going to make any claims about the veracity of a Twitter poll, but it confirms my suspicion that this isn't a hugely popular style.

holloway:
Yep, the Stylelint example was mentioned already and so I had a good look at their code. Their technical approach certainly was very complex (and, more to the point, very different) whereas Prettier has been fairly straight-forward to adapt. This PR is essentially using the existing CSS printer while wrapping some `indent()`s if preceding selectors are substrings of the current selector. As I'm sure you can appreciate implementations of features can vary greatly. (hey just for my knowledge is there a way of replacing my `path.pushCall()` with a normal `path.call()`? Would I set values on the `path` and use `.call()` to put them on the stack by name? that was one bit of my code that I wasn't too happy with).

> I suppose my main concern with this is that it's an alternative style

Now this is a point that I think is quite reasonable, and yeah that's the main point for me too. I guess that's why Lydell responded saying:

> And it is not just a "I prefer once space before that" type of thing, but something that can be argued to actually improve the overall readability of the entire style sheet.

For me that's the main point. It's a bit more than tabs vs spaces and (imo) tedious arguments like that... instead it's like auto-generating headings in your CSS. And I guess you've read the Sass devs talk about debugging stylesheets with this, and so on?

Anyway... is this your call, and you're killing the PR? Or is it something where all the existing contributors vote ... what's next?

I'll accept the decision of course, and no hard feelings. Obviously I think this PR warrants inclusion because it's more like a feature than a mere style, but I'll wait for whatever happens next. I didn't want to talk about your 'survey suspicions' because, well, obviously there's no way to prove or disprove hunches so I'm not sure if that would be very productive. Cheers :)

azz:
There isn't really a procedure in place for these decisions, usually there is some consensus among collaborators. If not, I guess @vjeux has the final word.

vjeux:
Our guidelines around options is that the default stance is that we should not add any. If there’s a huge demand for a new option and adding it would unblock a sizeable amount of people that couldn’t use it otherwise, then, and only then, should we consider adding it.

This particular option doesn’t seem like it falls in the second category I’m afraid.

holloway:
@vjeux Ok. It's good to finally know.

I really hope you're not basing it on a twitter poll of ~750 though, when there are much larger datapoints around.

impressivewebs:
Hi everyone,

I'm the one that organized the SitePoint surveys over the past two years. It's possible that [@azz is correct](https://github.com/prettier/prettier/pull/3038#issuecomment-346554470), that many people clicked "yes" without thinking it through.

But, that being said, what were they answering "yes" to? I don't buy the answer that they were equating "indenting" to using a pre- or post-processor. Nesting isn't the same as indenting (if that's what was assumed). I'd like to think the respondents were smart enough to understand that difference. (But then again, in the first survey, tons of respondents said they used CSS Hacks for IE11, when in reality, they were saying they supported IE11).

I think it's safe to assume in that case that most understood the question. The numbers still  might be skewed because of those who didn't understand it, but that would still leave a decent number, IMO. Also, there was an option for "I don't know what that means", to cover those who didn't get it. So again, I feel like it was pretty clear and didn't encourage false positives.

To be honest, the Typeform survey app is nice, but it would be better if I could include images and code examples with questions, without just linking to stuff. If I do this survey again, I'll definitely try to improve on the wording of that question.

In my opinion, I think it would be fine if Prettier had an option for this, but I'm not necessarily married to the feature. I do it in my own code, but not consistently; only in areas where I feel it will help with maintenance.

duailibe:
I originally was against this feature, now I'm neutral and wouldn't mind having it. That said, IMO, this would be better living in a separate tool that can run after prettier (`prettier-eslint`, `prettier-standard`, etc.). But I think the same about almost all of our other options (the exception being `tabWidth`).

I think this falls in the same category of the other options we started to support. I missed the discussion of when they were initially supported so I don't know how much demand there was, but I think this brings more value than something like `noBracketSpacing` or `jsxBracketSameLine` (no wonder people use it so much as an argument "why don't you support X option if you support Y?").

xiaoyu0519:
good

aaronkahlhamer:
An associate was dissuaded from using Prettier because of it removing css hierarchy indent. The person heavily invests in nesting to get one's bearings. I appreciate hierarchy indent, but prefer Prettier over it. A fear is, I'll be the only person using Prettier and will accidentally overwrite their formatting when providing an assist. All members of the team have to be on board to make Prettier the authoritative code formatter.

A baker's dozen, many who are freshman coders who recently started learning how to code this year! and have no clue about the 2016 SitePoint survey, will be using Prettier once this feature is enabled.  Looking forward to it!

j-f1:
@holloway Can you merge (or rebase) your PR so it’s mergeable again?

@lydell Does your ❌ review still stand?

holloway:
@j-f1 done!

Initially I didn't know why the snapshot tests were different, but then I realized that since v1.9.0 (#3317) the CSS normalization rules had changed. So it's a positive sign that whatever the normalization rules Prettier has this code will follow :blush:

lipis:
So this is going in? :|

holloway:
@lydell Cheers! I think I'd responded to all the specific things that you wanted changed (and everyone else's like @ikatyang too).

I've updated the first comment at the top of this PR with a summary of the polls so far.

One thing I thought might make sense to decrease the `maybeIndentStyle` calls in the CSS comment code so I've just done that...

before:
```javascript
if (text.indexOf(rawText) === -1) {
  if (n.raws.inline) {
    return maybeIndentStyle(
      concat([path.indentStyleAsWhitespace || "", "// ", rawText])
    );
  }
  maybeIndentStyle(
    concat([path.indentStyleAsWhitespace || "", "/* ", rawText, " */"])
  );
}
return maybeIndentStyle(
  concat([path.indentStyleAsWhitespace || "", text])
);
```

after:
```javascript
let concatComment;
if (text.indexOf(rawText) === -1) {
  if (n.raws.inline) {
    concatComment = ["// ", rawText];
  } else {
    concatComment = ["/* ", rawText, " */"];
  }
} else {
  concatComment = [text];
}
concatComment.unshift(path.indentStyleAsWhitespace || "");
return maybeIndentStyle(concat(concatComment));
```

@lipis No idea, sorry.

azz:
I'm doing a clean up of stale PRs, closing this one. As @vjeux https://github.com/prettier/prettier/pull/3038#issuecomment-346675287 mentioned there isn't significant demand for this feature, and it would mean Prettier could produce significantly different code (where most options are fairly small/low maintenance). 

@holloway I'd recommend adding this option to a fork of Prettier, or adding it as a `stylelint` plugin used with [`prettier-stylelint`](https://github.com/hugomrdias/prettier-stylelint).

PR #3040, opened 2017-10-15
Update Sublime Text and VSCode logos

fernahh:
**Issue:**
[Sublime Text](https://www.sublimetext.com/) and [VSCode](https://code.visualstudio.com/) has updated your logos, but in prettier website there are old versions.

**Solution:**
Just updating the images.

**Preview:**
screenshot


fernahh:
@ikatyang nice! I'm change.

PR #3055, opened 2017-10-18
Delete $schema from resolved config

### LINKED ISSUE #3047: Ignored unknown option: $schema
**Prettier 1.7.4**

When using `prettier` cli with a config file defining the schema of prettier, the cli complains about not knowing that option 

**Input:**
```json
{
  "$schema": "http://json.schemastore.org/prettierrc",
  "parser": "babylon",
  "bracketSpacing": true,
  "printWidth": 80,
  "semi": true,
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "es5",
  "useTabs": false
}
```

**Output:** (on terminal)
```bash
Ignored unknown option: $schema
```

**Expected behavior:**
I think this should not be ignored and should be validated with only one possible value, the $schema itself.

azz:
Fixes #3047

### DIFF
--- a/src/resolve-config.js
+++ b/src/resolve-config.js
@@ -9,7 +9,13 @@ const getExplorerMemoized = mem(opts =>
   cosmiconfig("prettier", {
     sync: opts.sync,
     cache: opts.cache,
-    rcExtensions: true
+    rcExtensions: true,
+    transform: result => {
+      if (result && result.config) {
+        delete result.config.$schema;
+      }
+      return result;
+    }
   })
 );
 
--- a/tests_integration/__tests__/config-resolution.js
+++ b/tests_integration/__tests__/config-resolution.js
@@ -109,3 +109,23 @@ test("API resolveConfig.sync overrides work with absolute paths", () => {
     tabWidth: 6
   });
 });
+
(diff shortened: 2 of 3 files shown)


src/resolve-config.js
```diff
@@ -9,7 +9,13 @@ const getExplorerMemoized = mem(opts =>
   cosmiconfig("prettier", {
     sync: opts.sync,
     cache: opts.cache,
-    rcExtensions: true
+    rcExtensions: true,
+    transform: result => {
+      if (result && result.config) {
+        delete result.config.$schema;
+      }
+      return opts.sync ? result : Promise.resolve(result);
```
ikatyang: It seems we do not need to handle Promise ourselves:

https://github.com/davidtheclark/cosmiconfig/blob/930926e7ca4e4f4c6cd57c371748a6c9d45083d8/src/createExplorer.js#L90-L93

azz: Ah, cool. I might send a PR to cosmiconfig to clarify that in the docs

suchipi:
Sorry, I forgot to squash-merge again 😅  luckily it was just one commit

PR #3069, opened 2017-10-19
Split source elements relative to their language.

### LINKED ISSUE #3068: [JS,TS] Range formatting ObjectExpression throws
Trying to format an ObjectExpression -- with at least 2 props -- throws. 

**Prettier master**

**Input:**
```sh
echo "const y = {a:1,b:2}" |  prettier --range-start 12 --range-end 14
```

**Output:**
```sh
[error] stdin: SyntaxError: Unexpected token, expected ; (1:7)
[error] > 1 | {a:1,b:2}
[error]     |       ^
```

**Expected behavior:**
Shouldn't throw and format the object as an object.
Note: if they were only 1 prop in the object it wouldn't throw but it would be considered as a `LabeledStatement`
```js
const y = {
     a: 1 ; // It would work with no-semi 
}
```
Other issue which seems highly related:
```sh
> echo "const y = [1]" |  prettier --range-start 11 --range-end 12
const y = [1;]
```

CiGit:
Fix #3068 : Colliding node types which are not source elements in every language.
Example: `ObjectExpression` in JSON / JS

Source Elements are now based on the parser.


index.js
```diff
@@ -204,59 +204,73 @@ function isSourceElement(opts, node) {
...
+    "OperationTypeDefinition",
+    "InterfaceTypeDefinition",
+    "UnionTypeDefinition",
+    "ScalarTypeDefinition"
+  ];
+  switch (opts.parser) {
+    case "flow":
+    case "babylon":
+    case "typescript":
+      return jsSourceElements.includes(node.type);
```
CiGit: Let's end tests, but I'm not sure .includes is in every node ...

index.js
```diff
@@ -204,59 +204,73 @@ function isSourceElement(opts, node) {
...
-    case "InputValueDefinition": // GraphQL
-    case "InputObjectTypeDefinition": // GraphQL
-    case "SchemaDefinition": // GraphQL
-    case "OperationTypeDefinition": // GraphQL
-    case "InterfaceTypeDefinition": // GraphQL
-    case "UnionTypeDefinition": // GraphQL
-    case "ScalarTypeDefinition": // GraphQL
-      return true;
+  // JS and JS like to avoid repetitions
+  const jsSourceElements = [
```
CiGit: What's the rule with global variables? those `const ...SourceElements` could be defined once.

tests/range/object-expression2.js
```diff
@@ -0,0 +1,9 @@
+
+const y =       [
+    {
+                a: 1,
+    },<<<PRETTIER_RANGE_START>>>
+    {
+              a: 1,
```
azz: Could you add an additional property to this object? Otherwise it could be treated as a block + label.

CiGit: Right, it's certainly clearer than trying to find a semicolon `{a: 1;}`. 
Done.

index.js
```diff
@@ -204,59 +204,73 @@ function isSourceElement(opts, node) {
...
-    case "EnumValueDefinition": // GraphQL
-    case "InputValueDefinition": // GraphQL
-    case "InputObjectTypeDefinition": // GraphQL
-    case "SchemaDefinition": // GraphQL
-    case "OperationTypeDefinition": // GraphQL
-    case "InterfaceTypeDefinition": // GraphQL
-    case "UnionTypeDefinition": // GraphQL
-    case "ScalarTypeDefinition": // GraphQL
-      return true;
+  // JS and JS like to avoid repetitions
```
josephfrazier: Is this supposed to be "JS and *TS* like to avoid repetitions"?

azz: I interpreted as `JS-like` = flow, TypeScript.

josephfrazier:
Oh, that makes more sense. I had overlooked "like".

Issue #3056, opened 2017-10-18
Format SCSS maps like CSS rules


tzi:
Hi!

It seems weird to me that prettier doesn't format SCSS maps like CSS rules.
What do you think? 

**Input:**
```scss
$map: (
  color: #111111,
  text-shadow: 1px 1px 0 hotpink 
);

.selector {
  color: #111111;
  text-shadow: 1px 1px 0 hotpink 
}
```

**Output:**
```scss
$map: (color: #111111, text-shadow: 1px 1px 0 hotpink);

.selector {
  color: #111111;
  text-shadow: 1px 1px 0 hotpink;
}
```
Here is a [Playground demo](https://prettier.io/playground/#%7B%22content%22%3A%22%24map%3A%20(%5Cn%20%20color%3A%20%23111111%2C%5Cn%20%20text-shadow%3A%201px%201px%200%20hotpink%20%5Cn)%3B%5Cn%5Cn.selector%20%7B%5Cn%20%20color%3A%20%23111111%3B%5Cn%20%20text-shadow%3A%201px%201px%200%20hotpink%20%5Cn%7D%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22scss%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D) (prettier 1.7.4).

Cheers,
Thomas.

PR #3070, opened 2017-10-20
Format SCSS maps like CSS rules

asmockler:
Resolves prettier/prettier#3056

This makes prettier format SCSS maps like CSS rules (line breaks after each rule). e.g.

```sass
$map: (color: #111111, text-shadow: 1px 1px 0 salmon)
```

will format to

```sass
$map: (
  color: #111111,
  text-shadow: 1px 1px 0 salmon
);
```

### DIFF
--- a/src/printer-postcss.js
+++ b/src/printer-postcss.js
@@ -345,7 +345,10 @@ function genericPrint(path, options, print) {
           indent(
             concat([
               softline,
-              join(concat([",", line]), path.map(print, "groups"))
+              join(
+                concat([",", parent.type === "value-value" ? hardline : line]),
+                path.map(print, "groups")
+              )
             ])
           ),
           softline,
--- a/src/printer-postcss.js
+++ b/src/printer-postcss.js
@@ -339,6 +339,12 @@ function genericPrint(path, options, print) {
         return group(indent(fill(res)));
       }
 
+      const declaration = path.getParentNode(2);
+      const isMap =
+        declaration &&
+        declaration.type === "css-decl" &&
+        declaration.prop.startsWith("$");
+
       return group(
         concat([
           n.open ? path.call(print, "open") : "",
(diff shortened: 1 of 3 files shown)


src/printer-postcss.js
```diff
@@ -345,7 +345,10 @@ function genericPrint(path, options, print) {
           indent(
             concat([
               softline,
-              join(concat([",", line]), path.map(print, "groups"))
+              join(
+                concat([",", parent.type === "value-value" ? hardline : line]),
+                path.map(print, "groups")
```
lydell: As you've probably noticed, the AST nodes in CSS/SCSS/Less are super vague and super general. While this PR looks good, I'm worried about what unintended side effects this might/will have. Our tests aren't very comprehensive. But I guess we could also just release it and do the regular couple of point releases afterwards with CSS regression fixes ¯\\\_(ツ)\_/¯

asmockler: Would it help if we checked `startsWith('$')` on the parent value? If that doesn't give any more assurance, I am fine fielding any regressions that arise related to this.

asmockler:
Added the `startsWith` check; it now reaches up 2 nodes and checks for a `css-decl`+ `startsWith("$")` instead of just `value-value` 👍

src/printer-postcss.js
```diff
@@ -339,13 +339,20 @@ function genericPrint(path, options, print) {
         return group(indent(fill(res)));
       }
 
+      const declaration = path.getParentNode(2);
+      const isMap =
+        declaration.type === "css-decl" && declaration.prop.startsWith("$");
+
```
azz: Might want to be a bit defensive and prefix with `declaration && `. It might not exist?

asmockler: Good call - just updated.

azz:
Thanks!

Issue #3071, opened 2017-10-20
problem in "fill" call to fit()


acthp:
Sorry, no examples of bad output. Experimenting with "fill", I found that it would inexplicably break after the first word, if the fill had small indentation. With larger indentation, it would not break after the first word, which is somewhat backwards.

I believe the width parameter passed to fit() is incorrect, here:

https://github.com/prettier/prettier/blob/master/src/doc-printer.js#L274

I suspect it should be ```rem```, not ```width - rem```, which I believe is the current position (explaining why it happens only with small indent: if the current indent is less that the current word length, it causes a break).

I'm unaware of any particular code example that is incorrectly formatted due to this, but perhaps the JSX printer is affected.

azz:
cc @karl

karl:
Good catch!

I have a PR that fixes that and updates the test that hit this edge case:
https://github.com/prettier/prettier/pull/3081

karl:
I've merged in the fix to `fill`, thanks for the great bug report!

PR #3081, opened 2017-10-21
Fix calculation of whether doc parts fit

karl:
It turns out there was a subtle bug in the `fill` algorithm that was spotted by @acthp in https://github.com/prettier/prettier/issues/3071.

This PR fixes that issue, and updates the snapshot tests for the few edges cases it changes.

### DIFF
--- a/tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
@@ -469,7 +469,11 @@ facebook_translation_leave_text_around_tag = (
 
 this_really_should_split_across_lines = (
   <div>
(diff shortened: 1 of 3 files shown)


tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -651,9 +655,7 @@ jsx_whitespace_after_tag = (
 
 x = (
   <div>
-    ENDS IN <div>
```
azz: This one looked better. ¯\_(ツ)_/ ¯.

karl: Yeah this was right on the limit for what would fit on a line, one more character and it would have formatted as before!

PR #3082, opened 2017-10-22
Add older versions of Prettier to the playground

azz:
Thought it would be cool to switch between different versions of Prettier in the playground. This PR adds the following versions:

* latest (`website/static/lib` on master)
* 1.7.4
* 1.6.1
* 1.5.3
* 1.4.4

Older versions are retrieved from GitHub via https://rawgit.com.

It's a little bit hacky - need to resolve file path based on version number, but it seems to be working. I'm using a fresh worker thread for each version (on demand), so there's no way they can pollute each other. Looks cool in DevTools 😆 

image

**Not handled**:

* switch to postcss parser for prettier < 1.7.1
* hide parsers not available in older versions

Both of those could make use of `getSupportInfo()` #3033. 

I'm a little concerned that this may make maintaining the playground more difficult... I guess if it gets too hard we can always drop this feature.

----

Also added the markdown parser to the playground.


azz:
Deployed this to https://azz.github.io/prettier/playground/ if you want to play around with it.

azz:
@ikatyang I noticed I got a couple of odd stack overflows with the markdown parser. They appeared for a half a second before returning a valid formatting. Managed to grab a screenshot:

image

ikatyang:
I'm not able to reproduce, is my test case wrong? I'm using the default sample + ` ```js `/` ``` `.

(Chrome 62)

(https://azz.github.io/prettier/playground)
image


(local build 4b0c99b26fee15154e4eb0b2cc06c81023de24e6)
image

azz:
Try changing `if (!greeting)` to `if (!greeting && foo)`.

website/static/worker.js
```diff
@@ -93,7 +101,7 @@ function formatCode(text, options) {
     // Multiparser may throw if we haven't loaded the right parser
     // Load it lazily and retry!
     if (e.parser && !parsersLoaded[e.parser]) {
```
ikatyang: It seems this is the root cause, it always tries to parse again if failed, and the blinking is caused by the syntax error while we're typing, the same thing also occurred in other multi-parser, e.g. css in js-template, etc.

azz: but `parsersLoaded.markdown` should be true after the first attempt?

ikatyang: Oh, it seems you forgot to add the version to this check:

(L120)

```diff
-parsersLoaded[actualParser] = true;
+parsersLoaded[parserAndVersion] = true;
```

azz: Do'h that was when I was trying to get multiple versions working in the same worker thread, `parserAndVersion` should no longer exist.

ikatyang:
Should we point `latest` to the "version" on master's `package.json`? so that we could add a `master` label (with issue template `version: master (hash)`) point to master's `website/static/lib`, which should be always up-to-date (always deploy) so that it should be more easily to debug and see if some issues are fixed on master.

azz:
I was thinking we'd just rename latest to master once we see up proper CD

lydell:
As you might now, the "Copy markdown" and "Report issue" buttons contain "Prettier vX.Y.Z" in their output, as well as a link to the playground. If those links always use version X.Y.Z that would be great. But it also means that we have to support _every_ release from now on in the "versions" dropdown. Not sure what to do about `master (hash)`, though.

Is the above the reason for this feature? Or is there another use case? Just for fun? To make it easier for people to verify that something changed between versions?

azz:
> To make it easier for people to verify that something changed between versions?

Mostly this. If it complicates things too much it might not be worth it.

azz:
Nah.

Issue #3076, opened 2017-10-20
Raise awareness with lint-staged problems with staging hunks.


juanjoLenero:
The proposed **lint-staged** workflow in the readme:
```
{
  "scripts": {
    "precommit": "lint-staged"
  },
  "lint-staged": {
    "*.{js,json,css}": [
      "prettier --write",
      "git add"
    ]
  }
}
```
Conflicts with git workflows that include commiting partial files via hunks. This can lead to bad commits where the whole file is pulled into the commit instead of just the selected hunks (file belonging to hunk is prettified -> whole new file is added).

A visible warning next to the snippet would probably help people avoid hitting this issue in the future. At least until okonet/lint-staged#62 is resolved.

duailibe:
There is already a warning about this issue in README: https://github.com/prettier/prettier/blame/b53a426e153247b309ac693763a4613905aa7bfb/README.md#L406

But not in the site docs. Besides that, how we could make this issue clearer?

juanjoLenero:
My bad, I completely missed the warning.

The message could be highlighted or at the very least more strongly worded. Adding it to the code-block as a comment right next to the `git add` line would increase visibility.

```
There is a limitation where if you stage specific lines this approach will stage the whole file after regardless. See this issue for more info.
```
->
```
This approach breaks staging separate lines/hunks in a file. See (Link to github issue detailing problem)
```

lydell:
@juanjoLenero Feel free to send a PR! :)

silvenon:
I like the idea of Prettier, this is just a current limitation of lint-staged, but I don't understand how anyone who partially stages files is ok with this limitation… 😕

Yay for submitting a PR with the warning!

PR #3083, opened 2017-10-22
Improve visibility of lint-staged usage warning (#3076)

juanjoLenero:


README.md
```diff
@@ -398,7 +398,7 @@ and add this config to your `package.json`:
   "lint-staged": {
     "*.{js,json,css}": [
       "prettier --write",
-      "git add"
+      "git add" // Warning: This breaks staging specific lines. See footnote below.
```
lydell: Hmm, this is JSON, which doesn't support comments ...

tests/markdown/README.md
```diff
@@ -388,7 +388,7 @@ and add this config to your `package.json`:
   "lint-staged": {
     "*.{js,json,css}": [
       "prettier --write",
-      "git add"
+      "git add" // Warning: This breaks staging specific lines. See footnote below.
```
lydell: You've edited the input to a test without changing its expected output, which is why Travis fails. I'm not sure we should update the test, though.

lydell:
Thanks for putting together a PR! Really appreciate it. However, I'm not sure this is any better.

README.md
juanjoLenero: Well isn't this embarassing 😅

juanjoLenero:
I guess you are right, the warning is already there. I could maybe just add bold markup to it or prefix it with "**warning:**" if you think that would help. Otherwise I don't think there's much that can be done.

tests/markdown/README.md
azz: You don't want to update this file. @ikatyang maybe we should rename it so people don't accidentally find it :P

PR #3085, opened 2017-10-22
Fixing playground report issue button and options parsing

mitermayer:
Unfortunately with PR https://github.com/prettier/prettier/pull/3063 I introduced a few bugs to the playground

1. The previous playground could parse config directly from the URL without having to wait any dependency to be available, due to the inclusion of lz-string as a dependencie for URL parsing we should have moved the logic of extracting config to go inside `window.onload`

2. I Naively thought that I could have simply update all encodeUriComponent to the lz-string variant, but that was not the behaviour that we needed for `report-bug` button, that particular one should have kept the old behaviour of encodeUriComponent.

Apologies for leading to a broken PR being merged into `prettier`. Tested this PR and it seems to fix all the above issues.

### DIFF
--- a/website/static/playground.js
+++ b/website/static/playground.js
@@ -26,77 +26,80 @@ var OPTIONS = [
 
 var IDEMPOTENT_MESSAGE = "✓ Second format is unchanged.";
 
-var state = (function loadState(hash) {
-  var parsed;
-  try {
-    // providing backwards support for old json encoded URIComponent
-    if (hash.indexOf("%7B%22") !== -1) {
-      parsed = JSON.parse(decodeURIComponent(hash));
-    } else {
-      parsed = LZString.decompressFromEncodedURIComponent(hash);
-    }
-  } catch (error) {
-    return {
-      options: undefined,
-      content: [
-        'function HelloWorld({greeting = "hello", greeted = \'"World"\', silent = false, onMouseOver,}) {',
-        "",
-        "  if(!greeting){return null};",
-        "",
-        "     // TODO: Don't use random in render",
-        '  let num = Math.floor (Math.random() * 1E+7).toString().replace(/\\.\\d+/ig, "")',
-        "",
-        "  return <div className='HelloWorld' title={`You are visitor number ${ num }`} onMouseOver={onMouseOver}>",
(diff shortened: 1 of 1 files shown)


mitermayer:
@azz, Sorry just noticed that my previous PR broke the playground since lz-string needs to be downloaded and we were doing the hash parsing outside of the window.onload.

 I did not caught that while developing locally since resources was always available.

website/static/playground.js
```diff
@@ -26,77 +26,80 @@ var OPTIONS = [
...
+      docEditor.setValue(message.data.doc || "");
+      astEditor.setValue(message.data.ast || "");
+      output2Editor.setValue(
+        message.data.formatted === ""
+          ? ""
+          : message.data.formatted2 === message.data.formatted
+            ? IDEMPOTENT_MESSAGE
+            : message.data.formatted2 || ""
+      );
+      document.getElementById("button-report-issue").search =
```
mitermayer: On the previous PR I accidently changed the report-issue to use lz-string for encoding. But it actually only needed the normal `encodeURIComponent`

website/static/playground.js
```diff
@@ -26,77 +26,80 @@ var OPTIONS = [
...
+      }
+    } catch (error) {
+      return DEFAULT_OPTIONS;
+    }
+    // Support old links with the deprecated "postcss" value for the parser option.
+    if (parsed && parsed.options && parsed.options.parser === "postcss") {
+      parsed.options.parser = "css";
+    }
+
+    return parsed || DEFAULT_OPTIONS;
```
mitermayer: Making sure we do return the DEFAULT_OPTIONS in case parsed is falsey

mitermayer:
cc @lydell @azz

mitermayer:
@suchipi

suchipi:
I assume with these changes that 2169357c17fb6e8552af2ca69e80d8fc385aa664 doesn't need to be reverted anymore, right?

mitermayer:
@suchipi , that's right this fixes that regression, thanks for merging

suchipi:
Sure thing, thanks for the fix

PR #3106, opened 2017-10-25
feature(jsx): Break opening elements when it has more than 3 props

### LINKED ISSUE #3101: Option to retain multi-line props in JSX/HTML
Right now I'm finding that Prettier collapses JSX attributes onto a single line, up until the column width. This (in my opinion) feels much harder to read as it requires the eyes to constantly scan left to right rather than "glance" at a single block. 

Would you be open to an option that disables this forced single-line output? 

**Prettier 1.7.4**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22const%20f%20%3D%20()%20%3D%3E%20(%5Cn%20%20%3CFoo%5Cn%20%20%20%20hello%3D%7Bbaz%7D%5Cn%20%20%20%20how%3D%7Bare%7D%5Cn%20%20%20%20you%3D%7Ba%7D%5Cn%20%20%20%20hi%3D%7Bc%7D%5Cn%20%20%2F%3E%5Cn)%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)

**Input:**
```jsx
const f = () => (
  <Foo
    hello={baz}
    how={are}
    you={a}
    hi={c}
  />
)
```

**Output:**
```jsx
const f = () => <Foo hello={baz} how={are} you={a} hi={c} />;
```

**Expected behavior:**
Retain line breaks. 

### Update: 

I opened this a long while back but after two years of using Prettier I absolutely, without question, and without hestitation, support the settings currently in Prettier and further -- believe that all of these personal preferences simply vanish from sight after using Prettier for more than a few days, as if one never even had them. They vanish into the delight of not having to worry about preferences anymore, and all of the energy / time savings that come from that. I am a 👎 on adding additional configuration in this regard.

duailibe:
Related #3101

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1782,6 +1782,8 @@ function genericPrintNoParens(path, options, print, args) {
             (n.name.comments && n.name.comments.length))
         );
 
+      const separator = n.attributes.length > 3 ? hardline : line;
+
       return group(
         concat([
           "<",
@@ -1789,7 +1791,7 @@ function genericPrintNoParens(path, options, print, args) {
           concat([
             indent(
               concat(
-                path.map(attr => concat([line, print(attr)]), "attributes")
+                path.map(attr => concat([separator, print(attr)]), "attributes")
               )
             ),
             n.selfClosing ? line : bracketSameLine ? ">" : softline
--- a/tests/flow/new_react/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow/new_react/__snapshots__/jsfmt.spec.js.snap
@@ -66,7 +66,12 @@ var FeedUFI = React.createClass({
     };
     var ignored = <UFILikeCount {...props} />;
     return (
-      <UFILikeCount className="" key="" feedback={feedback} permalink="" />
+      <UFILikeCount
+        className=""
+        key=""
+        feedback={feedback}
+        permalink=""
+      />
     );
   },
 
--- a/tests/jsx/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/jsx/__snapshots__/jsfmt.spec.js.snap
@@ -622,6 +622,10 @@ onClick={() => {
 onClick={() => {
   a
 }}>{header}<showSort attr="long long long long long long long long long long long"/></td>;
+
+// max 3 props per line
+<Component prop1="val" prop2="val" prop3="val" />;
+<Component prop1="val" prop2="val" prop3="val" prop4="val" />;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <td
   onClick={() => {
@@ -640,6 +644,15 @@ onClick={() => {
   {header}
   <showSort attr="long long long long long long long long long long long" />
 </td>;
+
+// max 3 props per line
+<Component prop1="val" prop2="val" prop3="val" />;
+<Component
+  prop1="val"
+  prop2="val"
+  prop3="val"
+  prop4="val"
+/>;
 
 `;
 
--- a/tests/jsx/open-break.js
+++ b/tests/jsx/open-break.js
@@ -7,3 +7,7 @@ onClick={() => {
 onClick={() => {
   a
 }}>{header}<showSort attr="long long long long long long long long long long long"/></td>;
+
+// max 3 props per line
+<Component prop1="val" prop2="val" prop3="val" />;
+<Component prop1="val" prop2="val" prop3="val" prop4="val" />;


damassi:
@duailibe - To be clear, this isn't exactly a fix for the issue I filed above. It's _definitely_ better than how it is now, but this

```jsx
const f = () => <Component prop1="val" prop2="val" prop3="val" />;
```
Is not as readable as this:

```jsx
const f = () => (
  <Component 
    prop1="val" 
    prop2="val" 
    prop3="val" 
  />;
)
```

given the fact that props can (and often) have a much longer character count, and the "three props" rule can still extend the reading length way out forcing more reading time. 

What I'm requesting in the issue above is that Prettier leaves this stuff alone, unless a user _likes_ writing his component out in one line and it reaches the max column width. 

If you would like to file a new issue and reference that in the above PR, that would be appreciated 😄

duailibe:
@damassi That's fair, I edited the description to remove the "Fixes".

azz:
I don't think this is the right move. I was hesitant to add the 3 rule with method chains, but it fixes a lot of weird issues. It's not a perfect fix either.

With jsx, it can be more readable to have multiple attrs on a line. Especially when there is a list of similar components.

```jsx
<Box width={5} height={2} color="black" />
<Box width={5} height={7} color="red" />
<Box width={7} height={2} color="black" />
```

lydell:
@azz Your example is not very convincing, though, because you have _three_ props! ;)

azz:
Edited my example.

damassi:
I agree with @azz in this case:

```jsx
<Container>
  <Box width={5} height={2} depth={2} color="black" />
  <Box width={5} height={7} depth={1} color="red" />
  <Box width={7} height={2} depth={2} color="black" />

  <Controller
    basePath={foo}
    someOtherData={toPass}
    routes={{
      foo: 'foo',
      bar: 'bar'
    }}
  /> 
</Container
```
In some instances we should leave it on one line, in others, allow a break. Structural markup is different than code.

duailibe:
I personally don't think this is less or more readable than:

```jsx
<Box
  width={5}
  height={2}
  depth={2}
  color="black"
/>
<Box
  width={5}
  height={7}
  depth={1}
  color="red"
/>
<Box
  width={7}
  height={2}
  depth={2}
  color="black"
/>
```

But I can see your point

lydell:
@damassi Your example already prints the way you want?

**Prettier 1.7.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAwtGBDAllOAJwD4AdKAAgtQCEIAPCgd1wBMYALAXmAFYBfChzi4A5hxg8ATINZwADp2mDIAGwiEupEACNV2MAGttFAPRlK1OoxbtufQcLESeAdlkKlwAIwqI6zW1COFYTc3IqWgZmNi93IRFxSWAZCjlFe1S1DS1dfSMwiwjqTFhCf1UiYqodbABnOAAFbC8AMwgIfmqKOogAWzgAeU4iABEW7B4YCGa6uq7LKnKAV3g6nmBuqnaIJAoAch39gBotilrCPf2L-e7+BapzCnJUU1KcfCJiEGOQCEVcNA6shQPUYMhWthVA1fjpCAZDHAYABleQGfCiZAwQjLOC-VgQMAQqEwkAAKzq9Bo8KMSOR2AGABlPsToXi-qt5KspKzSWjCA1CMhdNgdABPdRQH4geSEfAwADqsQ4yAAHAAGX4NPq4LE49l1DGVACKywg8F57JwOiVdmQUl+2LwqgxmD6fWwwqg0Dg0uWDQAKqLgUhIWz7kA)

**Input:**
```jsx
<Container>
  <Box width={5} height={2} depth={2} color="black" />
  <Box width={5} height={7} depth={1} color="red" />
  <Box width={7} height={2} depth={2} color="black" />

  <Controller
    basePath={foo}
    someOtherData={toPass}
    routes={{
      foo: 'foo',
      bar: 'bar'
    }}
  /> 
</Container>
```

**Output:**
```jsx
<Container>
  <Box width={5} height={2} depth={2} color="black" />
  <Box width={5} height={7} depth={1} color="red" />
  <Box width={7} height={2} depth={2} color="black" />

  <Controller
    basePath={foo}
    someOtherData={toPass}
    routes={{
      foo: "foo",
      bar: "bar"
    }}
  />
</Container>;

```

damassi:
@lydell - yup, because it hits the max-width. It was more of an example of how sometimes, for readability, code that is inline can be complimented by code that has line breaks. Here's a better example:

[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAwtGBDAllOAJwD4AdKAAgtQCEIAPCgd1wBMYALAXmAFYBfChzi4A5hxg8ATINZwADp2mDIAGwiEupEACNV2MAGttFAPRlK1OoxbtufQcLESeAdlkKlwAIwqI6zW1COFYTc3IqWgZmNi93IRFxSWAZCjlFe1S1DS1dfSMwiwjqTFhCf1UiYqodbABnOAAFbC8AMwgIfmqKOogAWzgAeU4iABEW7B4YCGa6uq7LKnKAV3g6ngBtbqp2iAAaCm2KWsIDo9qAL26AXQWqc0OoVFNSnHwiYhA9kAhFXGg6shQPUYMhWthVA1vjpCAZDHAYABleQGfCiZAwQjLODfVgQMBgiFQkAAKzq9BosKMCMR2AGABl3oTITifqt5KspMziSjCA1CMhdNgdABPdRQL4geSEfAwADqsQ4yAAHAAGb4NPq4DFY1l1NGVACKywg8G5rJwOgVdmQUm+mLwqjRmD6fWwgqg0DgkuWDQAKsLAUhwSz+PwgA)

```jsx
<Container>
  <Box width={5} height={2} depth={2} color="black" />
  <Box width={5} height={7} depth={1} color="red" />
  <Box width={7} height={2} depth={2} color="black" />

  <Controller
    basePath={foo}
    someOtherData={toPass}
    routes={[
      foo, 
      bar, 
      baz
    ]}
  /> 
</Container>
```

suchipi:
Maybe a better solution for that particular case would be to always break arrays inside of JSX expression containers

duailibe:
Closing this for now

PR #3113, opened 2017-10-27
don't break expressions in template string regex constructors

jwbay:
The line of thinking is that being able to read straight left-to-right is important for a human to 'parse' a regex and keep track of opening/closing markers, especially if there are a few declared at one per line.

Output without this:
```js
const shouldNotBreak = new RegExp(
  \`11111111111111111111111111111 should not break here \${this.x
    .y} 11111111111111111111111111111\`,
);
```

Worth noting that an alternate solution might be to strip newlines for *all* template string expressions unless at least one newline exists in the source expression. This is related to #1893. I'd honestly prefer this and it would arguably fix 1893, but it breaks a few more tests.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1934,8 +1934,16 @@ function genericPrintNoParens(path, options, print, args) {
             tabWidth
           );
 
+          const parent = path.getParentNode(1);
+          const shouldBreak = !(
+            parent &&
+            parent.type === "NewExpression" &&
+            parent.callee.type === "Identifier" &&
+            parent.callee.name === "RegExp"
+          );
+
           const aligned = addAlignmentToDoc(
-            expressions[i],
+            shouldBreak ? expressions[i] : docUtils.removeLines(expressions[i]),
             indentSize,
             tabWidth
           );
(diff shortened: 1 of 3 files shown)


src/printer.js
```diff
@@ -1934,8 +1934,16 @@ function genericPrintNoParens(path, options, print, args) {
...
+          const shouldBreak = !(
+            parent &&
+            parent.type === "NewExpression" &&
+            parent.callee.type === "Identifier" &&
+            parent.callee.name === "RegExp"
+          );
+
           const aligned = addAlignmentToDoc(
-            expressions[i],
+            shouldBreak ? expressions[i] : docUtils.removeLines(expressions[i]),
```
suchipi: We've gotten bitten by using removeLines in the past, because it will remove lines in nested nodes, eg functions, arrays, etc. I don't think this is a problem for template strings in RegExp literals, but it's worth being aware of.

vjeux: Yeah, please don't use remove lines. What you want to do instead is to reprint the few patterns you identified here.

You can start with member expressions with only identifiers and member expressions and no comments.

azz:
We should try to fix the general case rather than just RexExp.

PR #3115, opened 2017-10-27
[WIP] A prettier site

joecritch:
This is only like an hour's work so far, as I've mostly been learning how Docusaurus works.

I might even delete this PR and re-submit again later, but I just wanted to check that this is the right approach with Docusaurus? It seems a lot of the layout isn't customisable with React components, and can only be changed with a custom stylesheet.

Although I might be limited to certain styles, I quite like Docusaurus so far. 

cc @vjeux

### DIFF
--- a/website/static/css/prettier.css
+++ b/website/static/css/prettier.css
@@ -1,0 +1,202 @@
+h1, h2, h3, h4, body {
+  font-family: -apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Oxygen,Ubuntu,Cantarell,Fira Sans,Droid Sans,Helvetica Neue,sans-serif;
+}
+
+/**
+ * Fixed header
+ */
+
+.fixedHeaderContainer {
+  padding-top: 13px;
+  padding-bottom: 13px;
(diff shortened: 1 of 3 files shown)


vjeux:
cc @joelmarcey

vjeux:
Could you post a screenshot of how this looks? The code looks good for me. Are there things you wanted to change but couldn't? @joelmarcey is actively working on open sourcing Docusaurus for real :)

duailibe:
M

website/static/css/prettier.css
```diff
@@ -0,0 +1,202 @@
+h1, h2, h3, h4, body {
```
lipis: run Prettier on this file :)

joecritch:
@vjeux I'm closing this for now, as it didn't warrant a PR at this stage really. I'll re-open when it's looking healthier 👍

As for Docusaurus feedback, it'd be nice to be able to change the header, by decorating some base components, perhaps? I also feel like overriding the base styles could easily break with a version update, as they're targeting the internals of the base theme.

PR #3116, opened 2017-10-29
Sync docs from README.md to docs/*.md

### LINKED ISSUE #2825: Generate the options part of README.md and options.md from code

azz:
This is a lower-tech variant of #2825.

* Copies content from the main README.md file to the docs directory using `markdown-magic`. 
* Validated in CI via git diff.
* Replaces existing `markdown-toc` with `markdown-magic`

Closes #2825
Supersedes #2937


azz:
Thoughts on whether we should generate from `README.md` to `docs/` or the other way around? Usually build steps are compositional rather than extracting bits, but people landing on the repo will probably submit edits to README and it would be strange to have part of that file generated, half hand-written.

I've redirected the <kbd>EDIT THIS DOC</kbd> buttons on the website to the `README.md` so we don't get people accidentally submitting changes to the docs files.

duailibe:
Or maybe we should not have so much info in the README and link to docs instead? I know this might be controversial but I think it’s more sane

azz:
Yeah that's also an option. We'd have to make sure everything in the README is covered elsewhere though. [Babel's README](https://github.com/babel/babel) works like this.

package.json
```diff
@@ -33,6 +33,7 @@
     "jest-docblock": "21.3.0-beta.1",
     "jest-validate": "21.1.0",
     "leven": "2.1.0",
+    "markdown-magic": "^0.1.19",
```
ikatyang: pin?

azz: Gah, I keep doing this. Need to figure out why my yarn isn't respecting `.yarnrc`.

azz:
@vjeux thoughts on this vs just redirecting from Readme to website and building that out more?

vjeux:
I don't feel strongly about the two variants. Either we redirect to the website or we make sure the two are kept in sync.

azz:
Seems more sane to only keep the docs in one place. Better for maintenance and contribution via GitHub (without needing to run build tools). Opened #3139

PR #3124, opened 2017-10-31
Improve template literal expresions line breaks

### LINKED ISSUE #1893: Ugly line break in a template string

### LINKED ISSUE #2752: JSX ternary looks weird
screen shot 2017-09-05 at 9 57 21 pm

[prettier.io](https://prettier.io/playground/#%7B%22content%22%3A%22const%20renderInput%20%3D%20()%20%3D%3E%20%7B%5Cn%20%20%20%20return%20(%5Cn%20%20%20%20%20%20%3Cdiv%5Cn%20%20%20%20%20%20%20%20onBlur%3D%7Bthis.onInputBlur%7D%5Cn%20%20%20%20%20%20%20%20className%3D%7B%60mdl-textfield%20mdl-js-textfield%20%24%7BclassName%7D%20%24%7Bcontent.length%20%3E%200%5Cn%20%20%20%20%20%20%20%20%20%20%3F%20'is-dirty'%5Cn%20%20%20%20%20%20%20%20%20%20%3A%20''%7D%20combo-box__input%60%7D%5Cn%20%20%20%20%20%20%3E%5Cn%20%20%20%20%20%20%20%20%3Clabel%20className%3D%5C%22mdl-textfield__label%5C%22%20htmlFor%3D%5C%22combo-box__input%5C%22%3E%5Cn%20%20%20%20%20%20%20%20%20%20%7Blabel%7D%5Cn%20%20%20%20%20%20%20%20%3C%2Flabel%3E%5Cn%20%20%20%20%20%20%3C%2Fdiv%3E%5Cn%20%20%20%20)%5Cn%20%20%7D%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D)

duailibe:
Fix #1893
Fix #2752

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -1519,15 +1519,17 @@ d //comment
 \`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 \`
-\${a // comment
+\${
+  a // comment
 }
 
 \${b /* comment */}
 
 \${/* comment */ c /* comment */}
 
-\${// comment
-d //comment
+\${
+  // comment
+  d //comment
 };
 \`;
 
--- a/tests/template/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/template/__snapshots__/jsfmt.spec.js.snap
@@ -63,17 +63,19 @@ g
 }\`;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 \`
-(?:\${escapeChar}[\\\\S\\\\s]|(?:(?!\${// Using \`XRegExp.union\` safely rewrites backreferences in \`left\` and \`right\`.
-// Intentionally not passing \`basicFlags\` to \`XRegExp.union\` since any syntax
-// transformation resulting from those flags was already applied to \`left\` and
-// \`right\` when they were passed through the XRegExp constructor above.
-XRegExp.union([left, right], "", { conjunction: "or" })
-  .source})[^\${escapeChar}])+)+
+(?:\${escapeChar}[\\\\S\\\\s]|(?:(?!\${
+  // Using \`XRegExp.union\` safely rewrites backreferences in \`left\` and \`right\`.
+  // Intentionally not passing \`basicFlags\` to \`XRegExp.union\` since any syntax
+  // transformation resulting from those flags was already applied to \`left\` and
+  // \`right\` when they were passed through the XRegExp constructor above.
+  XRegExp.union([left, right], "", { conjunction: "or" }).source
+})[^\${escapeChar}])+)+
 \`;
 
-\`a\${/* b */ c /* d */}e\${// f
-g
-// h
+\`a\${/* b */ c /* d */}e\${
+  // f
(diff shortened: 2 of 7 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1519,15 +1519,17 @@ d //comment
 \`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 \`
-\${a // comment
+\${
+  a // comment
```
duailibe: I personally think this is better, but I can understand if this is considered a regression...

tests/template/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -63,17 +63,19 @@ g
...
-// \`right\` when they were passed through the XRegExp constructor above.
-XRegExp.union([left, right], "", { conjunction: "or" })
-  .source})[^\${escapeChar}])+)+
+(?:\${escapeChar}[\\\\S\\\\s]|(?:(?!\${
+  // Using \`XRegExp.union\` safely rewrites backreferences in \`left\` and \`right\`.
+  // Intentionally not passing \`basicFlags\` to \`XRegExp.union\` since any syntax
+  // transformation resulting from those flags was already applied to \`left\` and
+  // \`right\` when they were passed through the XRegExp constructor above.
+  XRegExp.union([left, right], "", { conjunction: "or" }).source
+})[^\${escapeChar}])+)+
```
duailibe: ... but this is 100x better than it was before.

Perhaps we could differentiate single comment vs. multiple comments? Or another heuristic..

tests/template/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -63,17 +63,19 @@ g
...
 \`;
 
-\`a\${/* b */ c /* d */}e\${// f
-g
-// h
+\`a\${/* b */ c /* d */}e\${
+  // f
```
suchipi: I could see someone opening an issue because this comment was moved, but in the XRegExp example, moving it is way better. Not sure how to best handle that; maybe we should treat multiline comments differently from single-line comments? Might not actually be a big deal though.

duailibe: I agree. I think it's a good tradeoff for keeping the logic simple but it's possible we could find a good heuristic for these cases

suchipi: Let's merge this as-is for now, and if more issues are opened, those might help us come up with a heuristic.

vjeux: I'm not too worried about comments here, they are extremely rare and this output looks decent.

jasonLaster:
Great Work!

Issue #575, opened 2017-02-02
watch task


thejameskyle:
It would nice to be able to use this as just:

```sh
yarn add --dev prettier
yarn prettier -- --write src/**/*.js --watch/-w
```

And only update the files that are changed as they are changed.

vjeux:
I wish there was a unix utility for that instead of each project having to reimplement it in a crappy way :(

xtuc:
@vjeux there is actually unix `inotify` but you will still need some sort of Node binding.

vjeux:
@xtuc I was more thinking about

```js
watch src/**/*.js | xargs prettier -- --write
```

thejameskyle:
I am using https://github.com/mikeal/watch to do effectively that:

```
watch 'prettier --write src/**/*.js' src
```

The problem is that it will just spawn up a new prettier process and go through every single file again

xtuc:
Yeah that would be awesome.

lydell:
There's also [cli-watch](https://www.npmjs.com/package/cli-watch) and [chokidar-cli](https://www.npmjs.com/package/chokidar-cli). Example:

```
chokidar 'src/**/*.js' -c "test '{event}' = 'change' && prettier --write '{path}'"
```

jlongster:
Do you have your editor instantly show changes from the filesystem? If not, how would save a file a file twice, since the second time there will have been changes to the file and you'll overwrite them? (My editor warns me about this and asks to confirm)

thejameskyle:
My editor automatically shows changes. It's fast enough that it doesn't cause me any issues for now. If the watch task only rebuilds the file I'm working on it should stay that way

jlongster:
Interesting, that's the opposite workflow from usual, where we have editor plugins that will automatically reformat on save. The editor invokes prettier before saving, formats it, and then it's saved.

My experience with watchers has always been painful trying to get them working across operating systems. Is there any reason you don't want to have the editor invoke prettier, @thejameskyle ? Which editor are you using?

thejameskyle:
Because I want it to run out of the box for everyone working on the project

xtuc:
With Vim, i'm not notified that the file has changed unless I try to save. If I really save I will just override my version with the formated one.

Prettify before save is actually a better solution in my case.

I know many other editors will just load the new version of the file you're currently on.

jlongster:
[...]

PR #3126, opened 2017-10-31
Add watching for changes to README

eedrah:
To tie up #575, I thought it would be good to have a note on how to watch for files in the README, considering it is a much requested feature in Issues. However, I feel like a watcher doesn't need to be in the code base considering there are other packages out there that implement this.

Hopefully this pull request will please everyone that wants a watcher.


eedrah:
(Side note - that's a really clever test :smile: )

PR #3133, opened 2017-11-01
Preserve line breaks in css grid declaration

### LINKED ISSUE #2703: messes up formatting of css grid layout
CSS Grid layout encourages a somewhat special syntax where the lines of the rule represent the grid being described. For example this rule:

```css
.item {
    grid-template-areas: 
        "header header"
        "main sidebar"
        "footer footer";
}
```
describes the grid areas for a 2-column, 3-row grid. You'd want to keep the formatting just like this because it's easier to understand when it's laid out the same way as the grid being described. 

Same goes for the `grid-template-columns`, `grid-template-rows`, and `grid-template` properties. Another example:

```css
.item {
    grid-template: 
        "a a a" 200px
        "b b b" 200px
        / 200px 200px auto;
}
```

Having it on separate lines like this makes it easier to understand which part of the CSS correlates to specific rows/columns.

Prettier currently concats everything to a single line, when ideally it should only handle indents and semicolons for these cases.

duailibe:
Fix #2703 

So this basically preserves the user's line breaks in CSS `grid` (or `grid-template*`) declarations, but continues to normalize quotes, numbers, etc.

### DIFF
--- a/src/printer-postcss.js
+++ b/src/printer-postcss.js
@@ -283,8 +283,22 @@ function genericPrint(path, options, print) {
       return path.call(print, "group");
     }
     case "value-comma_group": {
+      const parent = path.getParentNode();
+      let declParent;
+      let i = 0;
+      do {
+        declParent = path.getParentNode(i++);
+      } while (declParent && declParent.type !== "css-decl");
+
+      const declParentProp = declParent.prop.toLowerCase();
+      const isGridValue =
+        parent.type === "value-value" &&
+        (declParentProp === "grid" ||
+          declParentProp.startsWith("grid-template"));
+
       const printed = path.map(print, "groups");
       const parts = [];
+      let didBreak = false;
       for (let i = 0; i < n.groups.length; ++i) {
         parts.push(printed[i]);
         if (
@@ -292,7 +306,17 @@ function genericPrint(path, options, print) {
           n.groups[i + 1].raws &&
           n.groups[i + 1].raws.before !== ""
         ) {
-          if (
+          if (isGridValue) {
+            if (
+              n.groups[i].source.start.line !==
+              n.groups[i + 1].source.start.line
+            ) {
+              parts.push(hardline);
+              didBreak = true;
+            } else {
+              parts.push(" ");
+            }
+          } else if (
             n.groups[i + 1].type === "value-operator" &&
             ["+", "-", "/", "*", "%"].indexOf(n.groups[i + 1].value) !== -1
           ) {
@@ -301,6 +325,10 @@ function genericPrint(path, options, print) {
             parts.push(line);
           }
         }
+      }
+
+      if (didBreak) {
+        parts.unshift(hardline);
       }
 
       return group(indent(fill(parts)));
--- a/tests/css_grid/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css_grid/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,234 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`grid.css 1`] = `
+/* quotes */
+div {
+  grid-template-areas:
+       'header header'
+       'main sidebar'
+       'footer footer';
+}
+
+/* numbers */
+div {
+  grid-template-columns:
+      [full-start] minmax(1.50em, 1fr)
+      [main-start] minmax(.40ch, 75ch)
+      [main-end] minmax(1em, 1.000fr)
+      [full-end];
+}
+
+/* casing */
+div {
+  GRID:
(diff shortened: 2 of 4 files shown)


tests/css_grid/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,79 @@
...
+  grid: [wide-start] "header header header" 200px [wide-end]
+      "footer footer footer" 25px
+      / auto 50px auto;
+}
+
+div {
+  grid-template-columns:
+      [full-start] minmax(1em, 1fr)
+      [main-start] minmax(40ch, 75ch)
+      [main-end] minmax(1em, 1fr)
```
vjeux: Can you add an example where it does the number normalization?

tests/aaaaaaaaaaaaaaaaaaaaaa/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,6 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`lero.js 1`] = `
```
vjeux: remove :)

vjeux:
I like the approach :)

tests/aaaaaaaaaaaaaaaaaaaaaa/__snapshots__/jsfmt.spec.js.snap
duailibe: ops 😅

tests/aaaaaaaaaaaaaaaaaaaaaa/lero.js
```diff
@@ -0,0 +1,28 @@
+a {
```
lydell: The name of this file

- why .js and not .css?
- why “lero”?
- why “aaaaaaaaaaaaaa”?

src/printer-postcss.js
```diff
@@ -283,16 +283,35 @@ function genericPrint(path, options, print) {
...
     case "value-comma_group": {
+      const parent = path.getParentNode();
+      let declParent;
+      let i = 0;
+      do {
+        declParent = path.getParentNode(i++);
+      } while (declParent && declParent.type !== "css-decl");
+
+      const isGridValue =
+        parent.type === "value-value" && declParent.prop.startsWith("grid");
```
lydell: This should be case insensitive: `declParent.prop.toLowerCase().startsWith("grid")`

tests/aaaaaaaaaaaaaaaaaaaaaa/lero.js
lydell: Just saw that @vjeux commented on this

duailibe: @lydell sorry about this, I like to create a file to minimize the repro case. "aaaaaaaa" is meant to be the first folder in the `tests/` folder. the ".js" is because the default is me working on JS and this is what's in `git stash`.

I just forgot removing this before creating the commit 😅

tests/css_grid/__snapshots__/jsfmt.spec.js.snap
duailibe: Added!

src/printer-postcss.js
```diff
@@ -283,16 +283,38 @@ function genericPrint(path, options, print) {
...
         if (
           i !== n.groups.length - 1 &&
           n.groups[i + 1].raws &&
           n.groups[i + 1].raws.before !== ""
         ) {
-          if (
+          if (isGridValue) {
+            if (
+              n.groups[i].source.start.line !==
+              n.groups[i + 1].source.start.line
```
suchipi: Nice, good thinking

duailibe:
This is currently turning:

```css
grid-template-columns: 1fr 100px 3em;
```

into

```css
grid-template-columns:
  1fr 100px 3em;
```

Let me try to fix this first :-)

lydell:
- Could you add a test for case insensitivity?
- Could you modify the comprehensive test to show that quotes and numbers and stuff is normalized?

tests/aaaaaaaaaaaaaaaaaaaaaa/lero.js
azz: Btw, we have `/test.*` ignored in `.gitignore` so if you want to have a dummy file you can save `/test.js` and do `./bin/prettier.js test.js`.

duailibe: @azz wow that's gonna make my life so much easier. Thanks!

PR #3154, opened 2017-11-06
fix(jsx): Break opening elements with single text attribute if there's a comment

### LINKED ISSUE #3152: Inline comments inside jsx tags is broken.
**Prettier 1.7.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAGgHSgAhwenxwEMYYAnbAPm1X3SoG5tatcSzL3COLrb6TFlAzY8pCgF5gARgC+YgkQlca7eAA8YAhsyitFKyZhAnFPFf3VwtOofpFtxnY6dPdlnQy7Pu1eTW0RQUYQABoQCAAHGABLaABnZFBiBJhkADNiABsEuAiAI3JiMABrOBgAZSiS2KgAc2QKAFd8kAATCDBMnLyIgCsEjQAhYrKKyuIAWzgAGTq4Hty2iGaYKLWAJiW+kBryPPJkEALiAoBPbOhwvfI6mAB1WPaYAAtkAA4ABgi8qdimuRWr86vVsnAAIrNCDwHZtGBnJ4vd5ITYRCjEWLZUEAYQgUymxGOUGgiwizTyABUzkkkFllnI5EA)

**Input:**
```jsx
<X
  // attr
>
</X>;

<X
  attr
  // attr
>
</X>;

<X
  attr={1}
  // attr
>
  text
</X>;

<X
  attr=""
  // attr
>
  text
</X>;

<X
  attr=""
  // attr
  attr=""
>
  text
</X>;
```

**Output:**
```jsx
<X
// attr
/>;

<X
  attr
  // attr
/>;

<X
  attr={1}
  // attr
>
  text
</X>;

<X attr="">
// attr
  text
</X>;

<X
  attr=""
  // attr
  attr=""
>
  text
</X>;
```

**Expected behavior:**
```jsx
<X
  // attr // <- wrong indent
/>;

<X
  attr
  // attr
/>;

<X
  attr={1}
  // attr
>
  text
</X>;

<X
  attr=""
  // attr // <- bug: inline comments should remain here
>
  text
</X>;

<X
  attr=""
  // attr
  attr=""
>
  text
</X>;
```

duailibe:
Fix #3152 

**Input:**
```jsx
<div
  className="foo" // comment
>
  {children}
</div>;

<div
  className="foo"
  // comment
>
  {children}
</div>;
```

**Current in master:**
```jsx
<div className="foo"> // comment
  {children}
</div>;

<div className="foo">
// comment
  {children}
</div>;
```

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1766,12 +1766,25 @@ function genericPrintNoParens(path, options, print, args) {
     case "JSXOpeningElement": {
       const n = path.getValue();
 
+      const hasComments =
+        (n.name &&
+          ((n.name.trailingComments && n.name.trailingComments.length) ||
+            (n.name.comments && n.name.comments.length))) ||
+        (n.attributes &&
+          n.attributes.length > 0 &&
+          n.attributes.some(
+            attr =>
+              (attr.comments && attr.comments.length) ||
+              (attr.trailingComments && attr.trailingComments.length)
+          ));
+
       // don't break up opening elements with a single long text attribute
       if (
         n.attributes.length === 1 &&
         n.attributes[0].value &&
         isStringLiteral(n.attributes[0].value) &&
-        !n.attributes[0].comments
+        !n.attributes[0].comments &&
+        !hasComments
       ) {
         return group(
           concat([
@@ -1784,13 +1797,7 @@ function genericPrintNoParens(path, options, print, args) {
         );
       }
 
-      const bracketSameLine =
-        options.jsxBracketSameLine &&
-        !(
-          n.name &&
-          ((n.name.trailingComments && n.name.trailingComments.length) ||
-            (n.name.comments && n.name.comments.length))
-        );
+      const bracketSameLine = options.jsxBracketSameLine && !hasComments;
 
       return group(
         concat([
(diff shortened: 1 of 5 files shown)


src/printer.js
```diff
@@ -1770,7 +1770,8 @@ function genericPrintNoParens(path, options, print, args) {
       if (
         n.attributes.length === 1 &&
         n.attributes[0].value &&
-        isStringLiteral(n.attributes[0].value)
+        isStringLiteral(n.attributes[0].value) &&
+        !n.attributes[0].comments
```
azz: Can the attribute node have `trailingComments`?

duailibe: Not sure but I don't think so.. it has a comment with `trailing: true`, but no `trailingComments`.

duailibe:
Just caught another bug:

***Input:***
```jsx
<div // comment
  className="foo"
>
  {children}
</div>
```

***Master:***

```jsx
<div className="foo"> // comment
  {children}
</div>

src/printer.js
duailibe: Actually I don't see `trailingComments` anywhere on the code..

duailibe:
@azz I think this covers all the cases now

PR #3165, opened 2017-11-07
Add markdown parser to playground

azz:


SimenB:
This is somewhat buggy. Pasting in some markdown without changing options gives a syntax error (as expected), but changing the parser does nothing. I have to refresh, then it's correct.

Is this expected?

azz:
Probably bad caching? Does it work the next time?

SimenB:
Ah, yes it does. Odd

azz:
Service worker voodoo... https://twitter.com/markdalgleish/status/921519602202316800

PR #3171, opened 2017-11-07
docs(users): add Algolia

Haroenv:
We use prettier everywhere we can 😄


azz:
Awesome! We're just setting up DocSearch for prettier.io at the moment!

Haroenv:
Perfect, let us know if there's any issues with setting it up 👍

PR #3179, opened 2017-11-07
fix: Dead link for the External Tool for WebStorm

lipis:


azz:
This should actually point to prettier.io/docs/en/webstorm.html

lipis:
Nice!! Haven't seen that yet.. looks great.

website/editors.json
```diff
@@ -32,6 +32,6 @@
     {
         "image": "/images/webstorm-128px.png",
         "name": "WebStorm",
-        "content": "[External Tool](https://github.com/prettier/prettier/blob/master/editors/webstorm/README.md)"
+        "content": "[External Tool](https://github.com/prettier/prettier/blob/master/docs/webstorm.md)"
```
duailibe: I think this should be `(/docs/webstorm.md)` so it links to the site

PR #3181, opened 2017-11-07
Don't add trailing commas in JSXAttribute for arrow functions

### LINKED ISSUE #3173: Trailing comma after function return shorthand produce SyntaxError output
**Prettier 1.8.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAgCU4BDMAgMxwnXwHIdTy6AdKNuAD2z3wBM4FEgFcANgQAUwfNABiECPgC+ASnwBeAHz4JbfPgA8fVADc9+mVADCo1GADW64BLVbzFy-IhT3H-ACMSHCR6QIAvOgAaX31VXyVzAHpNNhUQSJAITBhUaABnZFBMHFRYAHVUPhgAC2QADgAGDJgSfwqq2qQAJgy80oBzUTgARWEIeGQhUTy4ZpwSVFsofqsadBJkEBJRUXSQf3mHOBgAZUwyAeQYHGFZkAArPM4AIUP7Y5OSdDgAGVK4SbbGYZc44GY4TaBfwAT1E0D2M3QqCuNzuwhmABVWgUkFNgSA+BAwIDpncSHkYCT8RBhDBMLSulS4EolEA)
```sh
--trailing-comma all
```

**Input:**
```jsx
import React from 'react'

export default ({ onFoo }) => (
  <div
    onClick={() =>
      onFoo({
        bar: 'baz',
      })
    }
  />
)
```

**Output:**
```jsx
import React from "react";

export default ({ onFoo }) => (
  <div
    onClick={() =>
      onFoo({
        bar: "baz",
      }),
    }
  />
);

```

**Expected behavior:**

There should not be a trailing comma after `onFoo(...)`. This is causing an error when running prettier a second time on the file.

Thank you for your time, and for this project!

duailibe:
Fix #3173

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -556,6 +556,9 @@ function genericPrintNoParens(path, options, print, args) {
         (args && args.expandLastArg) ||
         path.getParentNode().type === "JSXExpressionContainer";
 
+      const printTrailingComma =
+        args && args.expandLastArg && shouldPrintComma(options, "all");
+
       // In order to avoid confusion between
       // a => a ? a : a
       // a <= a ? a : a
@@ -580,10 +583,7 @@ function genericPrintNoParens(path, options, print, args) {
                 ])
               ),
               shouldAddSoftLine
-                ? concat([
-                    ifBreak(shouldPrintComma(options, "all") ? "," : ""),
-                    softline
-                  ])
+                ? concat([ifBreak(printTrailingComma ? "," : ""), softline])
                 : ""
             ])
           )
--- a/tests/trailing_comma/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/trailing_comma/__snapshots__/jsfmt.spec.js.snap
@@ -210,6 +210,63 @@ a(
 
 `;
 
+exports[`jsx.js 1`] = `
+<div
+  onClick={() =>
+    doSomething({
+      foo: bar
+    })
+  }
+/>;
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div
+  onClick={() =>
+    doSomething({
+      foo: bar
+    })
+  }
+/>;
+
+`;
+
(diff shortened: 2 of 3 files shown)


src/printer.js
```diff
@@ -556,6 +556,9 @@ function genericPrintNoParens(path, options, print, args) {
         (args && args.expandLastArg) ||
         path.getParentNode().type === "JSXExpressionContainer";
 
+      const printTrailingComma =
+        args && args.expandLastArg && shouldPrintComma(options, "all");
```
azz: Not sure I understand the relationship between `expandLastArg` and the bug?

duailibe: We we using `expandLastArg` to add a softline in the end *and* print the trailing comma. Now there's another we want to add a softline in the end (`JSXExpressionContainer`) but not the trailing comma, so they are two differente checks now:
- to add a softline: `expandLastArg` or parent in `JSXExpressionContainer`;
- to add a trailing: `expandLastArg` and the option is enabled

Does that make sense?

azz: Wouldn't this be clearer?

```js
const printTrailingComma =
  path.getParentNode().type !== "JSXExpressionContainer" &&
  shouldPrintComma(options, "all");
```

duailibe: I'm not sure.. if `expandLastArg` is true it already says that the arrow function is 1) an argument and 2) the last argument.

Checking only if the parent is not a JSX expression works, but seems brittle. I don't know

azz: Question is, is this a fix for JSX or is it more general?

duailibe: Ah, yes. That's a fix for a regression on #3140

duailibe: So it's a fix for arrow functions inside a JSXExpression

azz: Which is why I suggested doing:

```
  path.getParentNode().type !== "JSXExpressionContainer" &&
```

Using `expandLastArg` might mean other code will hit this path and be missing a trailing comma.

azz: Ok sounds fine.  @vjeux you may want to take a look.

PR #3172, opened 2017-11-07
fix(markdown): do not break before special prefix

ikatyang:
Fixes #3168

The unstable is caused by the AST change, should be stable now.

PR #3194, opened 2017-11-07
refactor(markdown): simplify regex

ikatyang:
Context: https://github.com/prettier/prettier/pull/3172#discussion_r149451396 https://github.com/prettier/prettier/pull/3194#discussion_r149460006

### DIFF
--- a/src/printer-markdown.js
+++ b/src/printer-markdown.js
@@ -87,8 +87,8 @@ function genericPrint(path, options, print) {
       const index = parentNode.children.indexOf(node);
       const nextNode = parentNode.children[index + 1];
 
-      // special prefix that may cause different meaning
-      if (nextNode && /^(#{1,6}|>)|^[-+*]$/.test(nextNode.value)) {
+      // leading char that may cause different syntax
+      if (nextNode && /^>|^([-+*]|#{1,6})$/.test(nextNode.value)) {
         return node.value === "" ? "" : " ";
       }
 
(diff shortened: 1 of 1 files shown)


src/printer-markdown.js
```diff
@@ -88,7 +88,7 @@ function genericPrint(path, options, print) {
       const nextNode = parentNode.children[index + 1];
 
       // special prefix that may cause different meaning
-      if (nextNode && nextNode.value.match(/^(#{1,6}|>+)|^(-|\+|\*)$/)) {
+      if (nextNode && /^(#{1,6}|>)|^[-+*]$/.test(nextNode.value)) {
```
lydell: Actually, I just realized that the `{1,6}` check for `#` is unnecessary, too: `/^[#>]|^[-+*]$/`

Or did you mean to only match up to 6 `#`?

Now that I'm thinking about it, is there a reason that we check for _prefix_ for `#` and `>` but _exact string_ for `-`, `+` and `*`? Just curios (and a tiny bit confused by the comment the line above).

ikatyang: > Or did you mean to only match up to 6 #?

Yes, `#######` won't be considered heading.

---

Just realized we can also check the exact string for headings since [CommonMark](http://spec.commonmark.org/0.25/#atx-headings) does not allow the space after `#`.

special prefix

```md
- 123
* 123
+ 123
>123
# 123
```

not special prefix

```md
-- 123
** 123
++ 123
#123
```

Uh, would be better to call it leading char?

PR #3195, opened 2017-11-07
Update README to point to API doc

jblz:
API doc is here: https://prettier.io/docs/en/api.html

https://prettier.io/docs/en/usage-options.html `404`s


suchipi:
Thank you!

PR #3189, opened 2017-11-07
test: lowercase the test file to avoid cross platform issue

ikatyang:
Context: https://github.com/prettier/prettier/pull/3187#issuecomment-342487311

PR #3187, opened 2017-11-07
Print JSON top comments as leading comments of root node

duailibe:
Fix #3178

PR #3207, opened 2017-11-08
Make files execution order predictive

duailibe:
Ever since I started contributing to prettier, the `tests_integration` tests fail for me because of file ordering when there's a uppercase filename (e.g. `cli/jest/__best-tests__` and `cli/jest/Component.js`). Also recently a snapshot were rewrited for the same reason (see: https://github.com/prettier/prettier/pull/3187#issuecomment-342487311 and #3189).

Since this effectively affects the CLI (not drastically I hope), we could simply rename the test files, but that's really a temporary fix. Otherwise, I'm not sure how else to prevent this issue.

### DIFF
--- a/src/cli-util.js
+++ b/src/cli-util.js
@@ -271,6 +271,8 @@ function eachFilename(argv, patterns, callback) {
       .sync(patterns, { dot: true })
       .map(filePath => path.relative(process.cwd(), filePath));
 
+    filePaths.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
+
     if (filePaths.length === 0) {
       logger.error(`No matching files. Patterns tried: ${patterns.join(" ")}`);
       process.exitCode = 2;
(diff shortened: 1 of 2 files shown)


src/cli-util.js
```diff
@@ -271,6 +271,8 @@ function eachFilename(argv, patterns, callback) {
       .sync(patterns, { dot: true })
       .map(filePath => path.relative(process.cwd(), filePath));
 
+    filePaths.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
```
lydell: Just a curious question: Is there a specific reason you use `.localeCompare` instead of just `filePaths.sort()`?

duailibe: @lydell `filePaths.sort()` sorts using case sensitivity (i.e. the same as `filePaths.sort((a, b) => a.localeCompare(b))`

lydell: Sorry, let me rephrase my question: Is there any specific reason you don’t just use `filePaths.sort()`? Nicer output? Or is it really required to do it case insensitively?

duailibe: The output of `glob` is already sorted. But the case sensitivity is the same for matching (if you want case sensitive matches, it will sort with case sensitivity).

https://github.com/isaacs/node-glob/blob/master/common.js#L153

duailibe:
That's fair I'm not that happy with this either

PR #3204, opened 2017-11-08
fix(markdown): no break on link

ikatyang:
Fixes #3203

PR #2760, opened 2017-09-06
Respect EditorConfig settings

josephfrazier:
This is a simple way to address https://github.com/prettier/prettier/issues/42. It adds support for .editorconfig's `indent_style`, `tab_width`, and `max_line_length` properties.

I still need to add caching and a proper async implementation, but I think it's functionally correct. It doesn't support the `end_of_line` property as described in https://github.com/prettier/prettier/issues/42#issuecomment-327073326, but that could be added later.

PR #3213, opened 2017-11-08
Revert "Respect EditorConfig settings"

azz:
Reverts prettier/prettier#2760 which was merged before #3204, which is needed for 1.8.2


duailibe:
We should remember to comment the docs when including it again for 1.9

giggio:
The configuration file is not showing the .editorconfig support. https://github.com/prettier/prettier/blob/master/docs/configuration.md

PR #3022, opened 2017-10-13
Prettier Markdown files

lipis:
Some of the JS examples though must be ignored.

Let me know what would be the default print size and I'll edit the JS examples that needs to be reverted and not prettified.

Issue #238, opened 2017-01-16
Idea: bot that pretty prints on commit


vjeux:
I was thinking in the context of a github project, how to make sure that the files stay pretty printed.

The most obvious way is to add a CI rule that fails if the file is not pretty printed, but in practice that's super annoying as you only get the feedback way later and don't even get an email notification that it failed.

Instead, I was thinking of letting maintainers merge non pretty printed files and then have a CI hook that pretty prints all the changed files. If there's a change, then have a bot commit the pretty printed version to master directly.

Because github doesn't check the author field, we can even have the commit under the original author name and link to the actual commit to make sure that the blame and attribution is clean.

I'd love thoughts on this, is this crazy?

hzoo:
Not crazy, was thinking about a bot for this for `eslint --fix` for a long time. For our newly created https://github.com/babel/babel-bot we want to comment with the log failure and I thought if we could check the logs (not sure the most cost effective) to see if a failure was a linting failure then can make a command to autofix). Yeah could make a commit and if the diff is good keep it in the PR.

demoneaux:
This is a pretty common idea actually. Various projects make use of their CI for a build step in order to ensure builds are updated properly. For example, the HTML5 Boilerplate has a Travis script that will build it after every change. In fact, @alrra has made a few easy to use scripts to allow branches to be updated (if necessary) from Travis: https://github.com/alrra/travis-scripts.

jlongster:
Great minds! I think there are three workflows people can choose from:

1. Use [lint-staged](https://github.com/okonet/lint-staged) to run the formatter check as a precommit hook and will fail if anything isn't formatted (the user then runs a command to format all of them). lint-staged is a lot better than running the formatter across an entire project because it'll be a lot faster: just runs it on changed files. (disclaimer: I haven't looked at that project yet but it seems promising)
2. Have CI check if entire project is formatted and fail if not
3. Do what you described above, and "autofix" it afterwards!

I think #3 is a really cool idea. We basically need a "last resort" option to make sure everything is formatted. Ideally, a github bot could automatically check PRs and if it detects something isn't formatted it could offer to do a follow-up commit or something like that. I like PRs being completely finished before actually pulling the merge button, and that would give you the chance to squash the format PR in with the others. Can github bots add commits to PRs?

hzoo:
Can github bots add commits to PRs?

> Since maintainers can edit other people's PRs now via https://github.com/blog/2247-improving-collaboration-with-forks

[...]

PR #3224, opened 2017-11-09
chore: setup markdown formatting

ikatyang:
Re-apply #3022

This PR also update prettier to v1.8.2.


CONTRIBUTING.md
```diff
@@ -19,8 +19,8 @@ Here's what you need to know about the tests:
   sure they are semantically equivalent.
 * Each test folder has a `jsfmt.spec.js` that runs the tests. Normally you can
   just put `run_spec(__dirname);` there. You can also pass options and
-  additional parsers, like this:
-  `run_spec(__dirname, { trailingComma: "es5" }, ["babylon"]);`
+  additional parsers, like this: `run_spec(__dirname, { trailingComma: "es5" },
+  ["babylon"]);`
```
ikatyang: Should we disable the break in `inlineCode`? though it's breakable there.

duailibe: I personally would not break

package.json
```diff
@@ -85,6 +86,7 @@
     "test:dist": "cross-env NODE_ENV=production yarn test",
     "test-integration": "jest tests_integration",
     "lint": "cross-env EFF_NO_LINK_RULES=true eslint . --format node_modules/eslint-friendly-formatter",
+    "lint-docs": "prettylint {.,docs,website}/*.md",
```
azz: Would be really good if we could run --fix and auto-commit changes. Because if people contribute via GitHub UI we don't want the build to fail.

CONTRIBUTING.md
azz: Yeah it's fine as is.

docs/api.md
```diff
@@ -35,32 +39,37 @@ prettier.formatWithCursor(" 1", { cursorOffset: 2 });
...
+file.
 
 If `options.useCache` is `false`, all caching will be bypassed.
 
 ```js
 const text = fs.readFileSync(filePath, "utf8");
 prettier.resolveConfig(filePath).then(options => {
   const formatted = prettier.format(text, options);
-})
+});
```
azz: 😆

package.json
ikatyang: The auto-commit is available but it's dangerous since we'll need to expose our token to PR branches in the Travis, an alternative would be to enable the `--no-prose-wrap` option to reduce the possibility of linting errors.

CONTRIBUTING.md
ikatyang: Opened #3230

duailibe:
Auto commit in PRs might be a bit confusing for users, but I like the auto-fix idea. Perhaps it should be in follow up PRs after commit?

In #238 someone metioned https://github.com/GordyD/prettier-master

ikatyang:
- Option 1: `--no-prose-wrap`
  - pros: won't cause any additional effects
  - cons: PR authors may feel annoying for the lint errors
- Option 2: a Travis script runs on master and commit `yarn lint-docs --fix` results
  - pros: PR authors feel happy
  - cons: there will be a lot of `yarn lint-docs --fix` commits on master
- Option 3: a GitHub App commit `yarn lint-docs --fix` results to the PR
  - pros: won't cause any additional effects
  - cons: PR authors may feel unhappy since their PR had been changed and that kind of App is not easy to build

I prefer Option 1 personally but Option 2 is also fine to me. The Option 3 is too expensive.

What do you think?

azz:
@vjeux you may have some thoughts on this.

ikatyang:
Just realized if we choose Option 2 and there're some ASI issues in the code block or we do not want to format some code blocks, the `yarn lint-docs --fix` will be terrible.

lipis:
What about just failing the test if the format is wrong.. without committing anything.. by doing so people could change/fix manually if they are willing to.. otherwise someone can edit or run the --fix on top of it before merging..

vjeux:
I've always found the 80 column enforcement of emails and markdown files such a stupid thing. It's wayyy harder to write and the benefits are very small, nowadays most of the places where they are displayed have a way to do word-wrap. If there are some that don't, then I believe that they should be fixed, rather than pushing a worse editing experience on everyone.

Note that code is a different beast. In order to word wrap code you neee to be able to parse the language, so only a complex and language aware tool like prettier can do it. Also there isn't a clear single way to do it. But for markdown prose, it's just a flat list of characters.

A good example is writing github comments, it's not expected to fit 80 columns: you don't even have a monospace font nor a column marker.

My 2 cents :)

CHANGELOG.md
```diff
@@ -308,7 +309,7 @@ Lots of small fixes, mainly for TypeScript.
 * [JSX] Break if opening element breaks (#942)
 * Parenthesize function expressions in expression position (#941)
 * update the README to add a pre-commit hook (#944)
-* Fix #951: properly parenthesize ** expressions (#952)
+* Fix #951: properly parenthesize \** expressions (#952)
```
ikatyang: Should be formatted as `\*\*`, fixed in #3246.

prettier.json
```diff
@@ -0,0 +1,3 @@
+{
+  "proseWrap": false
+}
```
ikatyang: Rename as `prettier.json` since `.prettierrc` will cause `--find-config-path` tests to fail.

azz: We should fix the tests (maybe mock impl for `existsSync()`?

azz:
This one is good to merge, right?

ikatyang:
If there's no further comment, it should be good to merge.

PR #3237, opened 2017-11-10
Add support for JSX fragments

duailibe:
This adds support for JSX Fragments (spec: https://github.com/facebook/jsx/pull/93).

- [x] Update babylon
- [x] Update flow (#3238)
- [x] Update TypeScript (#3331)
- [x] Write more comprehensive tests

I only wrote simple tests for now, will probably update the already existing JSX tests to include fragments as soon as we can test with TypeScript. But regarding the core code this is ready for review.

PR #2336, opened 2017-06-28
Update flow tests

lydell:
Just felt like running the syncing script for some reason, but I'm not sure if all these added tests provide any value. Thoughts?

Issue #1481, opened 2017-05-02
Syntax error for flow code


philipp-spiess:
The following code: 

```js
class Foo {
  constructor: () => this;
}
```

will throw a SyntaxError in the `babylon` parser but will work fine in the `flow` parser:

https://prettier.github.io/prettier/#%7B%22content%22%3A%22class%20Foo%20%7B%5Cn%20%20constructor%3A%20()%20%3D%3E%20this%3B%5Cn%7D%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%7D%7D

vjeux:
I actually have no idea which parser is right! I'm asking around. Thanks for the report!

philipp-spiess:
@vjeux We use this as a hacky way to overwrite and "fix" the wrong flow types for Immutable.Record classes. Not pretty but it works w/ flow :)

vjeux:
That's great context, could you open up an issue on the babylon repository? Once it is fixed there, we can update our dependency and it's going to work for prettier :)

philipp-spiess:
Thank you @vjeux. I've filed an issue [here](https://github.com/babel/babylon/issues/508). I'll update the issue here if something changes. 👌

existentialism:
Just FYI, landed a [fix for this](https://github.com/babel/babylon/pull/510) in Babylon!

Issue #883, opened 2017-03-04
Unstable comments when using rare newlines


lydell:
Note: This is not reproducible in the online demo, since it appears to normalize newlines.

First off, note how the comments are stable when using the widely used `\n` or `\r\n` newlines:

```js
$ node -p '"1;/*a*///b\n/*c*/2"' | prettier
1; /*a*/ //b
/*c*/ 2;

$ node -p '"1;/*a*///b\n/*c*/2"' | prettier | prettier
1; /*a*/ //b
/*c*/ 2;

$ node -p '"1;/*a*///b\r\n/*c*/2"' | prettier
1; /*a*/ //b
/*c*/ 2;

$ node -p '"1;/*a*///b\r\n/*c*/2"' | prettier | prettier
1; /*a*/ //b
/*c*/ 2;
```

But when using any of the other valid JavaScript newlines (`\r`, `\u2028`, `\u2029`) the comments become unstable:

```js
$ node -p '"1;/*a*///b\r/*c*/2"' | prettier
1;
/*a*/ //b
/*c*/ 2;

$ node -p '"1;/*a*///b\r/*c*/2"' | prettier | prettier
1; //b
/*a*/ /*c*/ 2;

$ node -p '"1;/*a*///b\u2028/*c*/2"' | prettier
1;
/*a*/ //b
/*c*/ 2;

$ node -p '"1;/*a*///b\u2028/*c*/2"' | prettier | prettier
1; //b
/*a*/ /*c*/ 2;

$ node -p '"1;/*a*///b\u2029/*c*/2"' | prettier
1;
/*a*/ //b
/*c*/ 2;

$ node -p '"1;/*a*///b\u2029/*c*/2"' | prettier | prettier 
1; //b
/*a*/ /*c*/ 2;
```

vjeux:
Soooo, I fixed it but the snapshot tests don't pass because jest somehow normalizes the line endings in the snapshot :p

vjeux:
Looks like only the one with `\r` doesn't work: https://github.com/prettier/prettier/pull/1050

PR #3238, opened 2017-11-10
Update flow

duailibe:
In order to support JSX fragments (#3237) we need to update flow to version 0.59. Since there were some relevant changes I decided to do another sync of the tests.


duailibe:
cc @lydell

.gitignore
```diff
@@ -1,4 +1,4 @@
-/node_modules
+node_modules
```
duailibe: Some tests have a `node_modules` folder to test module resolution but `jest` doesn't run them and I'm not sure they would bring relevant information to tests anyway.

tests/flow_class_field/jsfmt.spec.js
```diff
@@ -1 +1 @@
-run_spec(__dirname, null, ["babylon"]);
+run_spec(__dirname, { parser: "babylon" });
```
duailibe: It's related to #1481 but now this code only parses in babylon and not flow anymore. Maybe the babylon fix should be reverted? cc @existentialism

tests/newline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3,13 +3,15 @@
...
 `;
 
 exports[`backslash_2029.js 1`] = `
 1;/*a*///b
/*c*/2
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-1; /*a*/ //b
/*c*/2
+1; /*a*/ //b
+/*c*/ 2;
```
duailibe: Related to #883

not sure what to do here

lydell:
Regarding syncing the Flow tests: See #2336. Do any tests contain new syntax or something that we'd like to copy? Otherwise it might not be worth syncing. Perhaps we should even remove the sync script.

duailibe:
I don't think they have any new syntax, they've only removed stuff. Maybe we should only update the existing ones?

duailibe:
I think something's changed on the flow-parser, now the dist tests are failing with:

`Dynamic requires are not currently supported by rollup-plugin-commonjs`

duailibe:
I managed to add a hack to work around flow-parser’s dynamic require. Hopefully tests will pass now!

scripts/build/rollup.parser.config.js
```diff
@@ -18,6 +18,15 @@ export default Object.assign(baseConfig, {
...
       : {},
+    // In flow-parser 0.59.0 there's a dynamic require: `require(s8)` which not
+    // supported by rollup-plugin-commonjs, so we have to replace the variable
+    // by its value before bundling.
+    parser === "flow"
+      ? replace({
+          "require(s8)": 'require("fs")',
+          include: "node_modules/flow-parser/flow_parser.js"
+        })
+      : {},
```
vjeux: cc @mroch, not sure it was intentional but would be great to remove dynamic requires

tests/flow/fetch/headers.js
```diff
@@ -2,15 +2,15 @@
 
 const a = new Headers("'Content-Type': 'image/jpeg'"); // not correct
 const b = new Headers(['Content-Type', 'image/jpeg']); // not correct
-const c = new Headers({'Content-Type', 'image/jpeg'}); // correct
+const c = new Headers({'Content-Type': 'image/jpeg'}); // correct
```
vjeux: How was it even working!?

mroch: IIRC, it was valid short notation, and then not being typechecked properly

scripts/build/rollup.parser.config.js
mroch: this is js_of_ocaml optimizing things... and comes from its internal library. hrm.

also, `s8` is the random name it happened to assign to that hoisted string so this is definitely unstable across releases. will have to think about this...

duailibe: Yeah, I'm aware this is a random name given in the ocaml -> JS build. But that shouldn't be a problem until we update flow again, and we don't do that often.

duailibe: But it's funny that the `require("constants")` remained like that and `require("fs")` got hoisted.

PR #3250, opened 2017-11-11
Don't break self closing elements with no attributes

### LINKED ISSUE #3234: Weird new line inside a self closing JSX tag with no attributes
**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeADgPgDpQAR7wDOMhcJZFxp1lN59VDdLtbzxqARgE4D0OfKw4imJXKj5YQAGhAR0MAJbQiyUAEMSyAGYaANkThzeGsAGs4MAMrozSqAHNkMHgFdjIACYQwug0ZyDkY8MAAKPBqOALYa-oaeAFZEAB4AQpEWVtYa0XAAMg5w8YHybjDo5QBMJZ52PCHIIFwaXACe+tCyIOg8DjAA6kpeMAAWyAAcAAxyvRBGA5HoLu6ePHAAjm5K6xFRsbVyRtFKKx5HDo76cACKbhDwhyAwrUMj40hVcq4aSvqXAGEINEDigoNBinI3EYACqtNRIPQJAC+yKAA)

**Input:**
```jsx
<p>
  test test test test test test test test test test test test test test tes<br/>
  test test test test test
</p>
```

**Output:**
```jsx
<p>
  test test test test test test test test test test test test test test tes<br
  />
  test test test test test
</p>;

```

**Expected behavior:**
Maybe put `<br />` on the second line as a whole?

duailibe:
Fix #3234

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1794,6 +1794,11 @@ function genericPrintNoParens(path, options, print, args) {
       const nameHasComments =
         n.name && n.name.comments && n.name.comments.length;
 
+      // Don't break self-closing elements with no attributes and no comments
+      if (n.selfClosing && (!n.attributes || !n.attributes.length) && !nameHasComments) {
+        return concat(["<", path.call(print, "name"), " />"]);
+      }
+
       // don't break up opening elements with a single long text attribute
       if (
         n.attributes.length === 1 &&
--- a/src/printer.js
+++ b/src/printer.js
@@ -1792,10 +1792,10 @@ function genericPrintNoParens(path, options, print, args) {
       const n = path.getValue();
 
       const nameHasComments =
-        n.name && n.name.comments && n.name.comments.length;
+        n.name && n.name.comments && n.name.comments.length > 0;
 
       // Don't break self-closing elements with no attributes and no comments
-      if (n.selfClosing && (!n.attributes || !n.attributes.length) && !nameHasComments) {
+      if (n.selfClosing && !n.attributes.length && !nameHasComments) {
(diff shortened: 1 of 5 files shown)


tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -702,4 +709,13 @@ x = (
 // NOTE: Multiple JSX whitespaces are collapsed into a single space.
 x = <div> </div>;
 
+// Don't break a self-closing element without attributes
+// ----------
+x = (
+  <p>
+    text text text text text text text text text text text text text text text<br />text
```
duailibe: To help review, the print width is in the middle of `<br />`

src/printer.js
```diff
@@ -1791,6 +1791,14 @@ function genericPrintNoParens(path, options, print, args) {
     case "JSXOpeningElement": {
       const n = path.getValue();
 
+      const nameHasComments =
+        n.name && n.name.comments && n.name.comments.length;
+
+      // Don't break self-closing elements with no attributes and no comments
+      if (n.selfClosing && (!n.attributes || !n.attributes.length) && !nameHasComments) {
```
lydell: As far as I can tell from the next if-statement, `n.attributes` is always an array, so there’s no need to check if it exists.

duailibe: Yeah I was just following some previous code. I never know when properties that can be empty arrays can be undefined.

PR #3255, opened 2017-11-12
Re-add EditorConfig support (undo #3213)

josephfrazier:
Reintroducing https://github.com/prettier/prettier/pull/2760 for the 1.9 release, as suggested in https://github.com/prettier/prettier/pull/3213#issuecomment-343009769

This reverts commit d2241fc0d52d807701d9a5ac580bc01010e7a93b.


docs/configuration.md
```diff
@@ -83,3 +83,13 @@ For more information on how to use the CLI to locate a file, see the [CLI](cli.m
 ## Configuration Schema
 
 If you'd like a JSON schema to validate your configuration, one is available here: http://json.schemastore.org/prettierrc.
+
+<!--
+  ## EditorConfig
```
ikatyang: I think we can add an `_available in v1.9.0+_` note just like <https://prettier.io/docs/en/options.html#prose-wrap>, so that we don't have to uncomment new feature every time and it's clearer which version is available for it.

lipis:
Will the following work out of the box?! Just by running Prettier on all supported files?

```
root = true

[*]
indent_style = space
indent_size = 2
max_line_length = 80

[*.{js,jsx}]
max_line_length = 120

[*.json]
indent_size = 4
max_line_length = 200
```

josephfrazier:
@lipis, I think so! If you have a clone of the prettier repo, you can test this branch using [hub](https://github.com/github/hub):

```bash
hub checkout https://github.com/prettier/prettier/pull/3255
yarn
./bin/prettier.js /the/globs/of/your/files
```

josephfrazier:
They are supported :) See https://github.com/prettier/prettier/blob/258de9dac85f4afff3ebcf40057622ec5a82b113/docs/configuration.md#configuration-overrides

lipis:
WHAT?! Nice.. thanks :)

josephfrazier:
You're welcome! If you do test this branch with the `.editorconfig` file you posted above, please let me know how it goes.

lipis:
I will once I'll get home.. :)

src/resolve-config.js
```diff
@@ -26,23 +28,43 @@ function getLoadFunction(opts) {
...
+  const unwrapAndMerge = arr => {
+    const result = arr[0];
+    const editorConfigured = arr[1];
+    const merged = Object.assign(
+      {},
+      editorConfigured,
+      mergeOverrides(Object.assign({}, result), filePath)
+    );
+
+    if (Object.keys(merged).length === 0) {
```
azz: This doesn't look the same as the current behaviour. We want an empty config file to be different than no config file. How this works with `.editorconfig`... I don't know. Editors want to support formatting only if a config file is found, so supporting EditorConfig might break that functionality.

josephfrazier: > This doesn't look the same as the current behaviour.

According to @robwise in https://github.com/prettier/prettier/pull/2760#discussion_r145477197, the current behavior treats missing/empty config files equivalently. I wanted to be sure, so here's a script you can run against a clean build of cbed0f487f2f69b8f8a70ba73ec7115762556cfb to try it out:

```bash
tmpdir=`mktemp -d`
tmpfile=$tmpdir/file.js
tmprc=$tmpdir/.prettierrc

echo -n > $tmpfile

echo "With no .prettierrc"
node -p "require('./').resolveConfig.sync('$tmpfile')"
echo

echo "With empty .prettierrc"
echo -n > $tmprc
node -p "require('./').resolveConfig.sync('$tmpfile')"
echo

rm -rf $tmpdir
```

On my machine, this prints:

```
With no .prettierrc
null

With empty .prettierrc
null
```

Because of this, I'd like to fix this missing/empty distinction in a separate PR, since it's pre-existing and (somewhat?) unrelated (see https://github.com/prettier/prettier/pull/2760#discussion_r146105548)

> How this works with .editorconfig... I don't know. Editors want to support formatting only if a config file is found, so supporting EditorConfig might break that functionality.

Hmm, I guess the use-case here is:

* The user has a prettier integration in their editor
* The integration is configured to format-on-save if a prettier configuration exists
* The user is working on a project that has a .editorconfig
* But the project's style guide is incompatible with prettier's formatting

Yeah, I could see how that might get tricky. I'm going to have to give it some more thought.

josephfrazier: > Yeah, I could see how that might get tricky. I'm going to have to give it some more thought.

josephfrazier: Yeah, I could see how that might get tricky. I'm going to have to give it some more thought.

josephfrazier: > Yeah, I could see how that might get tricky. I'm going to have to give it some more thought.

Ok, so one way to fix this is to make it so that `resolveConfigFile()` (and its CLI counterpart `--find-config-path`) return paths to empty `.prettierrc` files (it doesn't currently). Once that's the case, the editor integration can use it to determine whether an empty `.prettierrc` exists, even if `resolveConfig` returns an object derived from `.editorconfig`.

CiGit: Using `resolveConfigFile(path) != null` could be a solution.
I've not seen this function exposed for now, seems to be cli only.

josephfrazier: > Ok, so one way to fix this is to make it so that resolveConfigFile() (and its CLI counterpart --find-config-path) return paths to empty .prettierrc files (it doesn't currently).

Hmm, it looks like cosmiconfig intentionally ignores empty files: https://github.com/davidtheclark/cosmiconfig/commit/1adf5c2f10112322e12d53f7f779d559aef5c627

However, a `.prettierrc` file containing just a space or a newline is still found, so that could be the suggested workaround instead of this from https://github.com/prettier/prettier/pull/2760#discussion_r145477197:

lipis:
Are we there yet?! It works :)

josephfrazier:
I think there's still a risk that some editor integrations could break under certain circumstances (see https://github.com/prettier/prettier/pull/3255#discussion_r152259674 for details). Besides, we also want to make sure that there are no more patch releases before 1.9.0, so this is kinda just chilling for now.

josephfrazier:
It sounds like 1.9.0 is going to be the next release (https://github.com/prettier/prettier/issues/3333#issuecomment-347983167). @azz, do you have any thoughts on https://github.com/prettier/prettier/pull/3255#discussion_r152259674?

azz:
How about this:

Add a `editorconfig: true|false` to the options of `resolveConfig` to disable lookup of editorconfig. Default to `false` for backwards compat, but set it to true internally from the CLI.

josephfrazier:
That sounds like a great way to sidestep potential editor integration issues, while still making `prettier` friendly to new users that use EditorConfig in their projects! I'll see what I can do

azz:
Looks like this is the last one left in the 1.9 milestone. Did you get a chance to make that change?

josephfrazier:
It's still in-progress. I think there are a few more things in the milestone, though? https://github.com/prettier/prettier/milestone/7

azz:
#3379 has since been added. The rest are post release items

josephfrazier:
Oh, I see. I've added the new `editorconfig` option, but I still need to update the documentation. I plan to have that done by the end of the day (Eastern US).

josephfrazier:
Ok, I think it's ready now.

azz:
What's the use case for exposing it as a cli option too?

josephfrazier:
When I was updating the documentation, I noticed that there wasn't a CLI argument as there is for the other options, so I decided to add it in for consistency. I'm not sure if there's actually a use case, though.

src/cli-constant.js
```diff
@@ -158,6 +158,13 @@ const detailedOptions = normalizeDetailedOptions({
   "debug-print-doc": {
     type: "boolean"
   },
+  editorconfig: {
+    type: "boolean",
+    category: CATEGORY_CONFIG,
+    description: "Take .editorconfig into account when parsing configuration.",
+    forwardToApi: true,
```
ikatyang: IMO, it shouldn't be forwarded to API just like `--config`, it's a CLI-only flag.

src/cli-constant.js
```diff
@@ -158,6 +158,13 @@ const detailedOptions = normalizeDetailedOptions({
   "debug-print-doc": {
     type: "boolean"
   },
+  editorconfig: {
+    type: "boolean",
+    category: CATEGORY_CONFIG,
+    description: "Take .editorconfig into account when parsing configuration.",
```
ikatyang: We should also add `oppositeDescription` to show docs for `--no-editorconfig`.

docs/options.md
```diff
@@ -216,3 +216,17 @@ Valid options:
 | Default      | CLI Override                                                | API Override                                                |
 | ------------ | ----------------------------------------------------------- | ----------------------------------------------------------- |
 | `"preserve"` | <code>--prose-wrap <always&#124;never&#124;preserve></code> | <code>proseWrap: "<always&#124;never&#124;preserve>"</code> |
+
+## EditorConfig
```
ikatyang: Should be in `cli.md` I think.

src/cli-constant.js
josephfrazier: I'm sure if you're saying that it *is* a CLI-only flag, or that it *shouldn't be*, but it isn't currently. Even though the default behavior only affects the CLI usage, I think we should leave the API option in place, so that editor integrations can opt-in to supporting `.editorconfig` without having to shell out to the CLI.

docs/options.md
josephfrazier: See https://github.com/prettier/prettier/pull/3255#discussion_r154543449

src/cli-constant.js
ikatyang: I mean, it should be both a CLI flag (`--no-editorconfig` for backward compatibility I guess?) and an option in API's `resolveConfig`, but not an option in `.prettierrc`.

Sorry for the confusing.

src/cli-constant.js
josephfrazier: Good point, that makes the `--help` message much better. Fixed in https://github.com/prettier/prettier/pull/3255/commits/64a441384dafc6dec3d51e978dc068b6bbfb585f

src/cli-constant.js
josephfrazier: Ohh, you're saying that it will be parsed from `.prettierrc`? I just did a quick test, and it seemed not to be...

```bash
set -x
node ./bin/prettier.js tests_integration/cli/config/editorconfig/file.js
echo 'editorconfig: false' >> tests_integration/cli/config/.prettierrc
node ./bin/prettier.js tests_integration/cli/config/editorconfig/file.js
node ./bin/prettier.js --no-editorconfig tests_integration/cli/config/editorconfig/file.js
git checkout tests_integration/cli/config/.prettierrc
set +x
```

ikatyang: Oh, I though you wrote that way. Then you don't need `forwardToApi: true` which is for forwarding to prettier.format.options (`.prettierrc`), and so does `options.js`.

https://github.com/josephfrazier/prettier/blob/64a441384dafc6dec3d51e978dc068b6bbfb585f/src/options.js#L26

josephfrazier: Ah, thanks for the explanation! I'm not too familiar with how the current options system is set up, so I just copied code from the other options. Fixed in https://github.com/prettier/prettier/pull/3255/commits/53013cbdc0fe633d6094cb97bfdd165afe39b137

azz:
Having `--no-editorconfig` is fine, I was just thinking an option to `resolveConfig` only.

docs/options.md
ikatyang: Move docs to API's `resolveConfig` and some simple description to CLI's `--no-editorconfig`? `options.md` is for options from `.prettierrc`, which can be used in both CLI flag or API `prettier.format.options`.

josephfrazier: Oh, that makes much more sense. Fixed in https://github.com/prettier/prettier/pull/3255/commits/a36f9b9b577f3be370816b1fbbcd883d81da7657

src/resolve-config.js
```diff
@@ -26,23 +28,47 @@ function getLoadFunction(opts) {
...
```
azz: Shouldn't this be:

```js
if (!result && !editorConfigured) {
  return null;
}
```

?

azz: (e.g. `.prettierrc` file containing `{}` should return `{}`, not `null`)

josephfrazier: Good catch, fixed in https://github.com/prettier/prettier/pull/3255/commits/e3460b888f1752904c2af1cd50705da3f4ec1a39! This change exposed a bug in some of the tests added in this PR, so I had to remove them. Note that the bug in question would also affect a similar test added on master, so I think this is ok for now. See https://github.com/prettier/prettier/pull/3255/commits/a2c81fc0eea2e9c4b3cee70195d91d01aef0bd3c for details.

azz:
Sorry this took so long to get in!

lipis:
KaBoom! 💥 🎆

josephfrazier:
:tada: It's understandable, @azz. This turned out to be a tricky change to get right. I'll see if I can add a little more detail to the release draft.

PR #3257, opened 2017-11-13
Add Wire to users.json

lipis:
Most projects already using Prettier and eventually all them will :) 
https://github.com/wireapp/wire-webapp/pull/1729/files

https://wire.com


duailibe:
Thanks!

PR #3259, opened 2017-11-13
Please add 2 more users

jimmyn:
Please add 2 microapps.com projects to the users as we're big fun of prettier and use it everywhere now


jimmyn:
@ikatyang sorry, I've added properly formatted images.

ikatyang:
Thanks!

PR #3268, opened 2017-11-14
Skip assertDoc calls in production

duailibe:
Context: https://github.com/prettier/prettier/issues/3263#issuecomment-344275152

When `printDocToString` calls `fill`, it's not necessary to call `assertDoc` for all of the remaining elements since they already checked when the doc was originally created (somewhere in `printAstToDoc`).

### DIFF
--- a/src/doc-builders.js
+++ b/src/doc-builders.js
@@ -2,22 +2,19 @@
 
 function assertDoc(val) {
   /* istanbul ignore if */
-  if (process.env.NODE_ENV !== "production") {
-    if (
-      !(
-        typeof val === "string" ||
-        (val != null && typeof val.type === "string")
-      )
-    ) {
-      throw new Error(
-        "Value " + JSON.stringify(val) + " is not a valid document"
-      );
-    }
+  if (
+    !(typeof val === "string" || (val != null && typeof val.type === "string"))
+  ) {
+    throw new Error(
+      "Value " + JSON.stringify(val) + " is not a valid document"
+    );
   }
 }
 
 function concat(parts) {
-  parts.forEach(assertDoc);
+  if (process.env.NODE_ENV !== "production") {
+    parts.forEach(assertDoc);
+  }
 
   // We cannot do this until we change `printJSXElement` to not
   // access the internals of a document directly.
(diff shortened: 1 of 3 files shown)


suchipi:
If we *never* `assertDoc` in `fill`, are we still covered? Or are there situations where we call `fill` without asserting doc in `printAstToDoc`?

We might want to offload this kind of check to a type checker like flow or etc instead of taking the runtime hit (static type checking instead). But of course, adding flow could make things more overwhelming for new contributors. Not sure how often new contributors make changes to the doc printer though (I suspect rarely).

duailibe:
> If we never assertDoc in fill, are we still covered? Or are there situations where we call fill without asserting doc in printAstToDoc?

I'm not sure I understand this correctly. But if I did, I think we're not covered, since the case for `fill` is to receive values (not other docs: `group` or `concat`): `fill(["text", separator, "line", separator, ...])` so those would be "unchecked" docs.

I was wondering what would be the downside of skipping the `assertDoc` check entirely in production. Our own code is covered by tests and the only problem that could happen would be a parser misbehaving (like returning something not string when it should be a string?).

suchipi:
Sorry that I wasn't clear, but yes, you understood correctly.

I like the idea of skipping assertDoc in production. Let's do that instead of special-casing fill.

src/doc-builders.js
```diff
@@ -2,12 +2,17 @@
 
 function assertDoc(val) {
   /* istanbul ignore if */
-  if (
-    !(typeof val === "string" || (val != null && typeof val.type === "string"))
-  ) {
-    throw new Error(
-      "Value " + JSON.stringify(val) + " is not a valid document"
-    );
+  if (process.env.NODE_ENV === "production") {
```
suchipi: Should be `!==`

duailibe: 🙄 sorry about that

duailibe:
@suchipi Looks like just calling the function is expensive enough :(

duailibe:
I had to wrap the calls with the check and add the replace in rollup build to remove them from the build (optional but a minor additional gain). I thought it would not get removed without minification but it was, to my surprise.

It's a bit ugly that bunch of ifs lying around but I think it's worth the benefit. We could change them with a `__DEV__` call witch is a bit fancier, and we just have to change the rollup replace plugin to achieve that.

Let me know what you think!

src/doc-builders.js
```diff
@@ -2,7 +2,7 @@
 
 function assertDoc(val) {
   /* istanbul ignore if */
-  if (process.env.NODE_ENV === "production") {
+  if (process.env.NODE_ENV !== "production") {
```
azz: Not sure this is the right way of doing it, a lot of people will run prettier in `"development"` env.

duailibe: @azz I updated the PR, can you check again?

duailibe: To clarify, when building the bundle, we will replace that check with `"production"` to remove that entirely

azz: Oh, right. Cool.

PR #3273, opened 2017-11-15
Mutate the doc.parts array when printing fill

duailibe:
Context: https://github.com/prettier/prettier/issues/3263#issuecomment-344275152

Slicing the array means copying the array's content for every pair, which makes the algorithm quadratic. This PR changes the copying to mutating the array directly to prevent that. It's a more dangerous approach but the way `fill` works it's a valid solution.

I'm not 100% comfortable with that and I'll take suggestions.

### DIFF
--- a/src/doc-printer.js
+++ b/src/doc-printer.js
@@ -297,10 +297,15 @@ function printDocToString(doc, options) {
             break;
           }
 
-          const secondContent = parts[2];
-
+          // At this point we've handled the first pair (context, separator)
+          // and will create a new fill doc for the rest of the content.
+          // Ideally we wouldn't mutate the array here but coping all the
+          // elements to a new array would make this algorithm quadratic,
+          // which is unusable for large arrays (e.g. large texts in JSX).
           parts.splice(0, 2);
           const remainingCmd = [ind, mode, fill(parts)];
+
+          const secondContent = parts[0];
 
           const firstAndSecondContentFlatCmd = [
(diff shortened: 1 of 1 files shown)


src/doc-printer.js
```diff
@@ -297,10 +297,11 @@ function printDocToString(doc, options) {
             break;
           }
 
-          const remaining = parts.slice(2);
-          const remainingCmd = [ind, mode, fill(remaining)];
-
           const secondContent = parts[2];
+
+          parts.splice(0, 2);
```
suchipi: A comment noting why we're mutating here might be useful.

josephfrazier:
This is a bit out of scope, but what if we used something like [Immutable.js](http://facebook.github.io/immutable-js/) to allow us to avoid mutation without impacting performance (or with very little impact)? Obviously this would be a larger, separate change, but I wanted to throw it out there.

suchipi:
I think using immutable structures is a good idea to keep in mind... but adding it at this point seems excessive. Maybe if general performance concerns start to be more pressing, we should look into it.

vjeux:
@josephfrazier immutable.js has a non trivial performance impact. In a React context, the performance overhead of the immutable data structures on-top of the js ones is offset by the wins you get from shouldComponentUpdate.

In our case, those data structures are pretty performance sensitive, so I'd be willing to be that we'll see a non trivial regression.

Issue #2362, opened 2017-06-30
Combination of UTF-8 BOM, Block comment, and Interpolated string breaks


Kerrick:
As you can see in [this link](https://prettier.io/#%7B%22content%22%3A%22%EF%BB%BF%5Cn%5Cn%2F*%20Block%20comment%20*%2F%5Cn%5Cnhtml%20%7B%5Cn%20%20content%3A%20%5C%22%23%7B1%7D%5C%22%3B%5Cn%20%20%5Cn%7D%22%2C%22options%22%3A%7B%22printWidth%22%3A120%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22postcss%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D), there is a problem prettifying a file that meets the following conditions:

- The file begins with a UTF-8 BOM
- There is a CSS block comment within the file
- There is a Sass interpolation within the file

In that case, the Block comment's closing `/` is omitted, causing the rest of the file to be commented out.

anthony-redFox:
@vjeux  Why I get string without BOM? 

in change log wrote:
```
// After
[BOM]/* Block comment */
html {
  content: "#{1}";  
}
```
But I get without BOM string. And currently prettier converts file from UTF-8-BOM to just UTF-8.

azz:
I think this is intentional. Any reason you need a byte order mark?

anthony-redFox:
We use BOM for detect that file in UTF-8 encoding and tool opens js files and check BOM in build procedure. Unfortunately, I don't know the same easy way to detect just UTF-8.

We had the issue when one of developer added symbol NEL(https://en.wikipedia.org/wiki/Newline) and it is looks like '...' in Windows-1252 encoding. But Japanese Windows IE  it is parses like UTF-8 encoding as new line and the JavaScript file has syntaxes error. Yes, it is very old bug, but I don't know a way to 100% detect  that file is UTF-8.

azz:
I'd have to check, but I'm fairly sure prettier is hardcoded to parse files as utf-8, so even with BOM it probably won't work.

josephfrazier:
[...]

PR #3283, opened 2017-11-17
UTF8-BOM - if input string has BOM that mean output string also should have it

anthony-redFox:
The issue that prettier lost BOM and mean convert UTF-8 with BOM to just UTF-8

#2362 
Also version 1.6.0 release note:
```
// Before
[BOM]/* Block comment *
html {
  content: "#{1}";  
}
```
```
// After
[BOM]/* Block comment */
html {
  content: "#{1}";  
}
```


index.js
```diff
@@ -68,7 +68,11 @@ function formatWithCursor(text, opts, addAlignmentSize) {
     return { formatted: text };
   }
 
-  text = stripBom(text);
+  const UTF8BOM = 0xfeff;
+  const isStripBom = text.charCodeAt(0) === UTF8BOM;
+  if (isStripBom) {
```
suchipi: `isStripBom` should probably be named `hasUnicodeBOM` instead

anthony-redFox: agree, is better. done

index.js
```diff
@@ -110,7 +114,10 @@ function formatWithCursor(text, opts, addAlignmentSize) {
   const doc = printAstToDoc(ast, opts, addAlignmentSize);
   opts.newLine = guessLineEnding(text);
   const toStringResult = printDocToString(doc, opts);
-  const str = toStringResult.formatted;
+  let str = toStringResult.formatted;
+  if (hasUnicodeBOM) {
+    str = "\ufeff" + str;
```
josephfrazier: I think we should use `String.fromCharCode(UTF8BOM)` here instead of `"\ufeff"`. Other than that, this looks good to me!

anthony-redFox: @josephfrazier  done. Thanks.

anthony-redFox:
@azz Can you take a look?

index.js
```diff
@@ -68,7 +68,11 @@ function formatWithCursor(text, opts, addAlignmentSize) {
     return { formatted: text };
   }
 
-  text = stripBom(text);
+  const UTF8BOM = 0xfeff;
+  const hasUnicodeBOM = text.charCodeAt(0) === UTF8BOM;
+  if (hasUnicodeBOM) {
+    text = stripBom(text);
```
azz: Should we do `.substring(1)` instead?

josephfrazier: Yeah, I was kinda thinking the same (or `.slice(1)` for consistency with other code). There's almost no reason to continue using `strip-bom` here, now that we have to keep track of what a BOM is in order to re-print it.

anthony-redFox: @azz  done

suchipi:
Thank you for this fix!

PR #3291, opened 2017-11-20
fix #3289

### LINKED ISSUE #3289: unnecessary parentheses at typescript function type
**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAZgVyjAEZsBebACioA8lsp8BbAIzgCcBKcgPmwGcY7AJZQA5tmycA3AB1YmHASIAmcrXqNWHbmT6CR46SAA0ICBhjDo-ZKACGg5LnsAbfnDMt29sAGs4GABlDF9RMWQhfE8QABMIMGc3DzNRD3YYAAUfMSZ7JPcYgCt+GgAhH39AoPsmOAAZUTgClPN8GAx2lRaY0PZ05FRFfjARS1MQDEMYAHVhWJgAC2QADgAGMymIDxmfDEj2aLN2OABHfGET7Ptc-KQXQrMPJmEDo5B+cNc4AEV8CHgPTMMHsLDmC2WSBUwJ8wlc4QAwhAmHlBlBoM0zPgPAAVUG2e7JOAAX2JQA)
```sh
--parser typescript
```

**Input:**
```tsx
type func1 = ( (x: number) => string  );
type func2 = (x: number) => string;
```

**Output:**
```tsx
type func1 = ((x: number) => string);
type func2 = (x: number) => string;

```

**Expected behavior:**
Parentheses at func1 should be omitted.

dhcmrlchtdj:
Fixes #3289.

### DIFF
--- a/src/fast-path.js
+++ b/src/fast-path.js
@@ -379,7 +379,7 @@ FastPath.prototype.needsParens = function(options) {
           parent.type === "GenericTypeAnnotation" ||
           parent.type === "TSTypeReference") &&
         (node.typeAnnotation.type === "TypeAnnotation" &&
-          node.typeAnnotation.typeAnnotation.type !== "TSFunctionType" &&
+          node.typeAnnotation.typeAnnotation.type !== "TypeAnnotation" &&
           grandParent.type !== "TSTypeOperator")
       ) {
         return false;
--- a/src/fast-path.js
+++ b/src/fast-path.js
@@ -379,7 +379,6 @@ FastPath.prototype.needsParens = function(options) {
           parent.type === "GenericTypeAnnotation" ||
           parent.type === "TSTypeReference") &&
         (node.typeAnnotation.type === "TypeAnnotation" &&
-          node.typeAnnotation.typeAnnotation.type !== "TypeAnnotation" &&
           grandParent.type !== "TSTypeOperator")
       ) {
         return false;
--- a/tests/typescript/conformance/types/alias/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/typescript/conformance/types/alias/__snapshots__/jsfmt.spec.js.snap
@@ -1,12 +1,24 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`aliasArrow.ts 1`] = `
-type NameResolver = (() => string);
+type simple0 = ()=>string;
+type simple1 = ( ()=>string );
+type simple2 = ((x:number)=>string);
 
-type func = ((x:number)=>string);
+type f0 = number | (() => string);
+type f1 = () => string | number;
+type f2 = (() => string) | number;
+type f3 = (() => string) | (() => boolean);
+type f4 = (() => string) & (() => boolean);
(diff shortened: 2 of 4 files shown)


src/fast-path.js
```diff
@@ -373,7 +373,7 @@ FastPath.prototype.needsParens = function(options) {
           parent.type === "GenericTypeAnnotation" ||
           parent.type === "TSTypeReference") &&
         (node.typeAnnotation.type === "TypeAnnotation" &&
-          node.typeAnnotation.typeAnnotation.type !== "TSFunctionType" &&
+          node.typeAnnotation.typeAnnotation.type !== "TypeAnnotation" &&
```
suchipi: Was the node type renamed at some point? If not, we might still need the TSFunctionType here

dhcmrlchtdj: oh.. it's my mistake.
This test is added by @azz to keep parens around function type. (https://github.com/prettier/prettier/commit/e8f12cf62149bf9e6456208c49b503371395ac99)

Are there any special cases that we should keep parens?

azz:
```
type F = (() => string) | (() => boolean);
type F = (() => string) & (() => boolean);
```

azz:
```ts
type F = () => string | number;
type F = (() => string) | number;
```

There may be more

dhcmrlchtdj: @azz, do you have any suggestion?

1. keep current behaviors?
2. strip parens as more as possible?
3. add parens around function type to make the output more consistent?

azz: I'd say go with parens only for intersection and union types. There may be more but we'll catch them eventually. We usually run Prettier on big TS projects before a release so we should catch any regressions.

tests/typescript/conformance/types/alias/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,24 @@
...
+type f3 = (() => string) | (() => boolean);
+type f4 = (() => string) & (() => boolean);
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+type simple0 = () => string;
+type simple1 = () => string;
+type simple2 = (x: number) => string;
+
+type f0 = number | (() => string);
+type f1 = () => string | number;
+type f2 = (() => string) | number;
```
dhcmrlchtdj: @azz 
should we add parens around function type in `f1`?

azz:
```ts
() => string | number;
// is actually
() => (string | number);
```

dhcmrlchtdj: @azz, sorry for this delayed response. (I have no too much time at recently.

I just want to strip the parens around function type at first.
But it's more complicated than what my expected, 😂.

Is there any other case I should consider？
(Maybe someone else would like to take over this PR if I don't have time?

Issue #3280, opened 2017-11-16
Template strings with awkward line breaks


ismail-syed:
**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEaIFs4CqATgDYAqAFgIYwCSaAMtAOaYC8mABhTDAA5okAeiE58lAJZRmAOhjU6jFjMi4hCACZ8IUmEIAkwAGYQIAXyEB3CgE9KEAK7MKB4ACMqRCxIDicOjAAQg4wAOq2nADcADpQIAA0IBB8MBLoyKBUGMhGVCRocIluRFRgANb+AMp8pVLMyDBEDoUgGhBgOXkFiVIFRDAACiXMuFSd+S0AVmgAHoEl5VVU+AxScOPdSSF8IQBMGy01RH3IIB5uNiTQCSB8RLqhEhryyAAcAAyJdxAFoSV8DSaLSIcAAjg4JCChlQRmMkLkJokCrgJIDmki6iQ4ABFBwQeAHRIwKhuR7PCjIXZEkoSEh1ADCeFGpyg0HWiQcBTIJLQBzMZiAA)

**Input:**
```jsx
const someUrlThatIsLong = `https://someThing.thatIsLong.com/endpoint/${foo}/whyThough/${bar}/iGetItButWhy`;

```

**Output:**
```jsx
const someUrlThatIsLong = `https://someThing.thatIsLong.com/endpoint/${
  foo
}/whyThough/${bar}/iGetItButWhy`;

```

**Expected behavior:**
This doesn't feel right. Is this a side-effect or intentional?

duailibe:
It's intentional, this is a template literal so we either break inside the expressions (i.e. between `{` and `}`) or don't break at all. The only other option is to change this to a string concatenation, but we won't it (that would change the AST, see [rationale](https://prettier.io/docs/en/rationale.html#what-prettier-is-not-concerned-about), you'd need to do it manually.

I'm closing this because there's nothing we can do here, but thanks for the report and please comment if you think anything was not addressed.

SimenB:
Fwiw, I'd _really_ prefer not breaking, it's much harder to read. I'd maybe understand it if there were a lot of function invocations or member expressions, but for simple cases like this prettier makes it way worse. The only places at work we use `prettier-ignore` is above template strings. 

[...]

PR #3299, opened 2017-11-21
Template literals: Don't break on identifiers but break if comments

duailibe:
Context: https://github.com/prettier/prettier/issues/3280#issuecomment-345882038

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1988,7 +1988,7 @@ function genericPrintNoParens(path, options, print, args) {
           let printed = expressions[i];
 
           if (
-            n.expressions[i].type === "Identifier" ||
+            (n.expressions[i].comments && n.expressions[i].comments.length) ||
             n.expressions[i].type === "MemberExpression" ||
             n.expressions[i].type === "ConditionalExpression"
           ) {
--- a/tests/line_suffix_boundary/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/line_suffix_boundary/__snapshots__/jsfmt.spec.js.snap
@@ -42,8 +42,10 @@ ExampleStory.getFragment('story')}
   d //comment
 }
 
-\${// $FlowFixMe found when converting React.createClass to ES6
-ExampleStory.getFragment("story")}
+\${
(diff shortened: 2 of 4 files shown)


SimenB:
I love your responsiveness to issues. Thanks you! ❤️ 

What about throwing in some more examples of formatting? More as something to compare against if it changes in the future than anything.

Here's a couple of examples from Jest's migration to 1.8 (https://github.com/facebook/jest/pull/4852):

```js
descirbe('something', () => {
  test(`{pass: false} expect(${small}).toBeGreaterThanOrEqual(${big})`, () => {});  
})

throw new Error(`pretty-format: Option "theme" has a key "${key}" whose value "${value}" is undefined in ansi-styles.`,)
```

That last one is really long, but breaking it on only one identifier and not the other makes is difficult to read.

Anecdotally, one thing I came over in a personal project where I would like a break, but the break is wrong

```js
class Class {
  something() {
    // This would be WAY better if it took the whole `time.unit.charAt(0)` on a new line instead of just the `0`
    // Not breaking at all is better than that
    return () => {
      const timeQuery = time
      ? `time:(from:now-${time.value}${time.unit.charAt(0)},mode:quick,to:now)`
      : "";
    }
  }
}
```

duailibe:
Thanks for the kind words and the examples! <3

The last one is a really trick case because we have two separate places to break:

```js
`time:(from:now-${time.value}${
  time.unit.charAt(0)
},mode:quick,to:now)`
`time:(from:now-${time.value}${time.unit.charAt(
  0
)},mode:quick,to:now)`
```
And it currently tries to fit everything until it breaks. So we have no way of telling "it's ok to break here but try to break the parent first". This is another case that would benefit from #3014.

duailibe:
Actually, my bad.. this one is a different case. Including a softline for CallExpressions in the template literals expressions works for your example, but breaks in another one:

```js
const foo = `something:${JSON.stringify({
  a,
  b,
  c
})}`

// turns into:

const foo = `something:${
  JSON.stringify({
    a,
    b,
    c
  })
}`
```

PR #3305, opened 2017-11-22
Fix infinite recursion in the playground

### LINKED ISSUE #3300: Maximum call stack exceed in playground
image

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEAzAhgJ0wF5MBnGAT3QBMA6agSwDcADAMQgkwCECWBuADpQQAGhAQADjAbRSyULnLI8qUnHEAjfLjABrDAGVJuhlADmyGPgCuGkNQhgVuNfbPr8MAAo7zAW1wXN3EAK1IAD24dfSNcfzgAGTM4YPVxCBsYSSyAJjT7E3xPZBBNXE0qaDEQSXwzGAB1BmoYAAtkAA4ABnE6iHVGnUkrW3t8OABHGwYJ31wAoKRVdJB1fwZRu3FSM3N0AEUbCHgC8RgK5taOpFzznQZUPYBhCH9A0qhoVPEbdQAVCryZaudQAXzBQA

duailibe:
Fixes #3300

If there's an exception when formatting the code in playground, we check if it's because the parser was not yet loaded. This check was wrong for json and css|less|scss causing the code to recursive infinitely.

### DIFF
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -94,8 +94,7 @@ function formatCode(text, options) {
   } catch (e) {
     // Multiparser may throw if we haven't loaded the right parser
     // Load it lazily and retry!
-    if (e.parser && !parsersLoaded[e.parser]) {
-      lazyLoadParser(e.parser);
+    if (e.parser && !lazyLoadParser(e.parser)) {
       return formatCode(text, options);
     }
     return String(e);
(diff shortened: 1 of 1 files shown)


website/static/worker.js
```diff
@@ -111,8 +111,8 @@ function lazyLoadParser(parser) {
         : parser;
   var script = "parser-" + actualParser + ".js";
 
-  if (!parsersLoaded[actualParser]) {
+  if (!parsersLoaded[parser]) {
```
duailibe: This means it will call `importScripts` again if you try to load css after loading scss, for example. Is that a big deal?

I could change the check above in line 97 instead to get the actual parser, but this seemed simpler.

ikatyang: How about checking the status in `lazyLoadParser()`? so that the `parsersLoaded` is only used in one place.

image

PR #3309, opened 2017-11-22
fix(cli): respect `--ignore-path` when using `--stdin-filepath`

### LINKED ISSUE #3308: .prettierignore isn't respected when using --stdin-filepath
Let's say you have a project configured like so.

```
node_modules/
.prettierignore <- contains `app/` as its only line
app/app.ts
```

If you run `node_modules/.bin/prettier app/app.ts`, prettier will return no output. If you run `node_modules/.bin/prettier --stdin-filepath app/app.ts --stdin < app/app.ts`, then prettier will return the modified text. Prettier should instead search upwards from `--stdin-filepath` to find the `.prettierignore` file. This is important for integrating prettier with editor plugins like ALE.

ikatyang:
Fixes #3308

Output nothing if matched patterns in `--ignore-path`.


w0rp:
Thank you for the quick response. :+1:

PR #3317, opened 2017-11-25
Don't lowercase element names and attribute names in selectors

### LINKED ISSUE #3304: Styled-components integration lowercases SVG tag names in selectors
**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAxCFMC8mGAngDZwAmAdGgA4CGUABgDpSaZkCWUcDAJwDiAhpW4Iswdp04YIdALSQyEAUky8AFnAHcYAbhmYAvu2YGQAGhAKY3dMlAMMyAGYMyaODYBGosABrOBgAZUYwXgBzZBgBAFcfEEoIMHdPbxtebwEYAAVRKIBbBnSvJIArNAAPACEA4LCGIrgAGV44MszbeJg6XoAmLqTGARzkEF8GX3JoaxA6PVgAdW5KGC1kAA4ABhtFiG9l0TpYhKSBOABHeO5LgoZi0qQPcptvIu4zxPfoigBFeIQeDDGwwaardabJADMGibg8KBRADCECKJQmUGgnRs8W8ABVpmhhiYTEA)

**Input:**
```jsx
const Foo = styled.span`
  linearGradient {
    stop-color: inherit;
  }
`;
```

**Output:**
```jsx
const Foo = styled.span`
  lineargradient {
    stop-color: inherit;
  }
`;

```

**Expected behavior:**
Expected the selector of "linearGradient" to not be changed to "lineargradient".

The SVG element is &lt;linearGradient&gt;, but prettier rewrites the selector as "lineargradient" which no longer matches in Firefox or Safari.

lydell:
https://www.w3.org/TR/css3-selectors/#casesens

> All Selectors syntax is case-insensitive within the ASCII range (i.e.
> [a-z] and [A-Z] are equivalent), except for parts that are not under the
> control of Selectors. The case sensitivity of document language element
> names, attribute names, and attribute values in selectors depends on the
> document language. For example, in HTML, element names are
> case-insensitive, but in XML, they are case-sensitive.

Fixes #3304.

### DIFF
--- a/tests/css_case/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css_case/__snapshots__/jsfmt.spec.js.snap
@@ -3,7 +3,8 @@
 exports[`case.less 1`] = `
 // Convention in this test file:
 // - The case should be preserved for things prefixed with "Keep".
-// - The case should always be preserved for function names and property keywords.
+// - The case should always be preserved for element names and attribute names
+//   in selectors, as well as function names and property keywords.
 // - Other things should mostly be lowercase.
 // - The \`/*:*/\` comments are just to bust the \`isLikelySCSS\` check.
 
@@ -11,6 +12,7 @@ exports[`case.less 1`] = `
 
 HTML#KeepId.KeepClass,
 a[HREF=KeepAttrValue]:HOVER::FIRST-letter,
+svg[viewBox] linearGradient,
 :Not(:NTH-child(2N+1)) {
     COLOR: #AAbbCC;
     BACKGROUND-image: URL("KeepString");
@@ -49,7 +51,7 @@ a[HREF=KeepAttrValue]:HOVER::FIRST-letter,
 @KeepTopLevelVar: val;
 $KeepScssVar: val;
 
-.Keep(@Keep: 12e03PX) WHEN (@Keep=Case) /*:*/ {
+.Keep(@Keep: 12e03PX) when (@Keep=Case) /*:*/ {
     @KeepVar: KeepName; /*:*/
     @{KeepInterpolationVar}: val;
     $KeepScssVar: val;
@@ -59,7 +61,7 @@ $KeepScssVar: val;
       prop: val;
     }
 
-    &Keep & NoKeep {
+    &Keep & Element {
       prop: val;
     }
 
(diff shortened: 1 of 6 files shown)


tests/css_case/case.less
```diff
@@ -68,14 +70,14 @@ $KeepScssVar: val;
...
+    .Keep() when (@Keep=12PX);
+    .Keep() when (@Keep=Keep12PX);
 }
 
-.Keep (@Keep) WHEN (lightness(@Keep) >= 12PX) AND (@Keep > 0) {}
-.Keep (@Keep) WHEN (lightness(@Keep) != '12PX') AND (@Keep != "12PX") {}
-.Keep (@Keep) WHEN (lightness(@Keep) >= Keep12PX) AND (@Keep > @Keep12E5) {}
+.Keep (@Keep) when (lightness(@Keep) >= 12PX) and (@Keep > 0) {}
+.Keep (@Keep) when (lightness(@Keep) != '12PX') and (@Keep != "12PX") {}
+.Keep (@Keep) when (lightness(@Keep) >= Keep12PX) and (@Keep > @Keep12E5) {}
```
lydell: As far as I understand, Less only accepts `when` and `and` lowercase.

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -292,17 +300,17 @@ a[href="KeepAttrValue"]:hover::first-letter,
 }
 
 @keyframes KeepAnimationName {
-  from {
+  FROM {
```
lydell: `from` and `to` could technically be lowercased in `@keyframes`, but I don’t think it is worth the trouble.

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -89,15 +91,17 @@ $KeepScssVar: val;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // Convention in this test file:
 // - The case should be preserved for things prefixed with "Keep".
-// - The case should always be preserved for function names and property keywords.
+// - The case should always be preserved for element names and attribute names
+//   in selectors, as well as function names and property keywords.
 // - Other things should mostly be lowercase.
 // - The \`/*:*/\` comments are just to bust the \`isLikelySCSS\` check.
```
azz: Unrelated to this PR, but can we use `parser: "less"` instead of the comments now?

lydell: I don’t know.

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -292,17 +300,17 @@ a[href="KeepAttrValue"]:hover::first-letter,
...
 
 @keyframes KeepAnimationName {
-  from {
+  FROM {
     prop: val;
   }
 
   #{$KeepInterpolationVar},
-  #{$Keep + 15px},
+  #{$Keep + 15PX},
```
lydell: Oops, I didn't see this one when looking through the diff before. Need to lowercase that "PX".

azz:
Good to merge?

PR #2676, opened 2017-08-28
Arrow parens option

rattrayalex:
Addresses the rather lengthy https://github.com/prettier/prettier/issues/812
supersedes https://github.com/prettier/prettier/pull/2430 and https://github.com/prettier/prettier/pull/2473

I tried to keep commits reasonably granular, though there was some thrash. I can recondense if desired. 

I created an option with three choices: 
- "avoid" (prefer `x => x`)
- "default" (prefer `(x) => x` except curry-chains and callbacks)
- "always" (always do `(x) => x`)

We avoid options very heavily in prettier. However, @vjeux asked me to implement this and I think it makes sense here.

I set the default option to `default` because I think it is the best choice for most cases. This will cause some churn, but I don't think too much, and if it's the right thing to do, it's best to do it now. (In particular, the upcoming 1.6.0 release will have significant churn on react codebases, I think in a great way, so people should be okay with this being rolled in; in both cases, its mostly just adding parens). 

The `default` option was tricky to get right. It helped that `prettier` itself makes heavy use of callbacks, including nonstandard ones. The algorithm that I ended up going with is to omit parens when: 
- it is a part of a curry chain, denoted by being a child or parent of an arrow-fn-expression with an implicit return.
- it is a callback passed as the first argument to a function call, which does not contain other functions (to add parens when multiple callbacks are passed in a row). 
- it is a direct child of a template expression, (added for styled-components, basically) eg; 
    ``` `color: ${props => props.color}` ```

I also added support for the `it("is an asynchronous test with a long string", (done) => {})` edge-case highlighted in a previous PR (though I punted on the typeparams edge-edge-case). 

I am open to keeping `avoid` (the current behavior) as the default option. If the purpose of that would be to avoid this change on the prettier codebase itself, I would instead advocate for use of `.prettierrc` to override it. If the purpose is backwards-compatibility, well, people are going to have diffs when they upgrade to 1.6.0 anyway, and it's the release *after* this that we hope to minimize churn (afaik), so now would be the best time to squeeze in an upgrade like this (which I think will improve the editing experience for ~all prettier users). 

(EDIT: the "default" option was originally called "functional" which was silly)

PR #3324, opened 2017-11-26
Simple arrow function parens option

suchipi:
This is based on https://github.com/prettier/prettier/pull/2676 by @rattrayalex. Originally, I was going to rebase that branch and make changes there, but rebasing proved difficult due to unrelated changes in master.

This PR adds an "--arrow-function-parentheses" choice option with two options: "always" and "avoid". Avoid is the same as now (try not to print them if you can). Always is the new option (always print them). Avoid is the default, for backwards-compatibility.

Unlike #2676, this does not include the callbacks option AKA functional. Even though I find it interesting, I don't think there is community/collaborator buy-in or consensus, and I don't want to not add "always" because of that.

This PR also includes a fix that @rattrayalex started in his branch to inline type parameters on test callback functions (`it("very long description", <T>(done) => { /* ... */ })`).


docs/options.md
```diff
@@ -96,6 +96,19 @@ Put the `>` of a multi-line JSX element at the end of the last line instead of b
 | ------- | ------------------------- | ---------------------------- |
 | `false` | `--jsx-bracket-same-line` | `jsxBracketSameLine: <bool>` |
 
+# Arrow Function Parentheses
```
wKovacs64: This should probably be a `##` heading? 😅

src/printer.js
```diff
@@ -3272,6 +3255,19 @@ function printFunctionParams(path, print, options, expandArg, printTypeParams) {
...
+  if (options.arrowFunctionParentheses === "always") {
+    return false;
+  }
+
+  if (options.arrowFunctionParentheses === "avoid") {
+    const node = path.getValue();
+    return canPrintParamsWithoutParens(node);
+  }
+
+  return false;
```
rattrayalex: Probably worth adding a comment indicating that this should be unreachable

rattrayalex:
Oops, looks like you have some Travis failures (ironically looks like prettier needs to be run; do we have a command for that?).  Also need to fix the docs issue wKovacs64 mentioned.

src/printer.js
lipis: or just

```js
function shouldPrintParamsWithoutParens(path, options) {
  if (options.arrowFunctionParentheses === "avoid") {
    const node = path.getValue();
    return canPrintParamsWithoutParens(node);
  }
  // possible comment here to explain it :)
  return false;
}
```

rattrayalex: That works too, though I like the explicitness (and having something to grep for, and having a safe default-fallback).

azz:
Obligatory bikeshed... `arrow-function-parentheses` is a really long name... Should we call it `arrow-parens` like ESLint?

azz:
(Aside: should we continue using milestones, or use release labels?)

suchipi:
Let's use milestones instead of release labels; I forgot that milestones existed, and they're nicer for this kind of thing.

suchipi:
Looks like I'll need to add the option to the website, too.

duailibe:
And we should have a way to coordinate pre-release playground options too...

qleguennec:
Any news on adding a `block` parameter to `arrowParens`, like the `eslint` rule?

rattrayalex:
@qleguennec doubtful; personally, I've been using the "always" option and am much happier with it than I expected. I don't mind the extra parens when the computer does all the typing 😀

Which are you currently using / why is it problematic?

lydell:
@qleguennec You might be able to work around this using [prettier-eslint](https://github.com/prettier/prettier-eslint).

PR #3325, opened 2017-11-27
feat(markdown): preserve `break` style

### LINKED ISSUE #3319: Markdown: Preserve double-space line break syntax
In the original [Markdown syntax](https://daringfireball.net/projects/markdown/syntax#p) description, two spaces at the end of a line create a line break: 

> When you do want to insert a `<br />` break tag using Markdown, you end a line with two or more spaces, then type return.

This syntax is also supported in the [CommonMark spec](http://spec.commonmark.org/0.28/#hard-line-breaks):

> A line break … that is preceded by two or more spaces and does not occur at the end of a block is parsed as a hard line break

Prettier currently replaces this with CommonMark's single-backslash notation _(which works fine in some places like GitHub issue comments)_, but this syntax is **not** supported by `kramdown`, the default Markdown parser used by GitHub Pages, so the lines are joined and the backslash characters are visible in output when prettied Markdown is rendered via GitHub Pages.

The [kramdown syntax](https://kramdown.gettalong.org/syntax.html#paragraphs) requires **two** backslash characters at the end of the line for line breaks, but this renders elsewhere as two visible backslashes rather than a line break: `\\`.

**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAqABK1AjCAbAE3WwFc884Z1o4t10AdKe+vASwGcrJY2oSAhjDbRqUOEwxZchYmQpUYAdwh1GzFuy7oew-kJHNlEJiAA0ICAAdh0DslACuyAGYC8HOBewAnAWABrSgBlK38+AHNkGB8SLxACCDBXd08LPk8fGAAFPwiAWwEUj3iAKw4ADwAhP0CQgXy4ABk+OGK0yxIYKy6AJnb4sJ9M5BBCnwDEpShzECsfPhgAdTYCGAALZAAOAAYLeYhPJb8raNj4nzgARxI2S9yBAqKkNxKLT3y2M7j3yIoARRIEHgAwsMAE2BWa02SF6YL8bHYUAiAGEIPlCqMoDRZiRPAAVCH2F6pOAAXzJQA)
```sh
--parser markdown
```

**Input:**
```markdown
* **bold bullet one**  
    list continuation one
* **bold bullet two**  
    list continuation two

```

**Output:**
```markdown
* **bold bullet one**\
   list continuation one
* **bold bullet two**\
   list continuation two

```

**Expected behavior:**

The double-space line breaks should be preserved for maximum source compatibility with a broad range of Markdown implementations.

ikatyang:
Fixes #3319


src/doc-printer.js
```diff
@@ -398,7 +398,12 @@ function printDocToString(doc, options) {
                     out.pop();
                   }
 
-                  if (out.length) {
+                  if (
```
azz: Is there precedent for using the notion of "parser" in `doc-*` files?

ikatyang: No, but this is the easiest way to support double-space `break`.

That's why I chose the backslash style `break`, the trailing whitespace syntax is super weird and conflict with the `trim_trailing_whitespace` rule.

Issue #3322, opened 2017-11-26
Ignoring code section on website is kind of broken (I guess)


lipis:
https://prettier.io/docs/en/ignore.html

<img width="819" alt="screen shot 2017-11-26 at 22 35 06" src="https://user-images.githubusercontent.com/125676/33244555-2f232c84-d2fa-11e7-9270-840bbef6b8f9.png">

azz:
Looks like code fences with a `<!-- prettier-ignore -->` on the previous line are the cause.

~~~md
<!-- prettier-ignore -->
```js
code
```
~~~

@ikatyang any ideas here? I guess we could put a line break between the code fence and the comment, but it'd be a bit ugly.

ikatyang:
Since GitHub/CommonMark rendered it correctly, it should be a Docusaurus bug I guess, but don't know where to report.

azz:
Fixed in our end, not much else we can do

lipis:
What was that Docusaurus anyway?! Is it still active?

suchipi:
@lipis https://docusaurus.io/

lipis:
yes.. I found that one.. but the link to GitHub is dead https://github.com/facebookexperimental/docusaurus

azz:
It's a private repo

lipis:
I see :)

lipis:
Or more like.. I don't _see_ :P

PR #3326, opened 2017-11-27
chore(website): ignore `<!-- prettier-ignore -->` before passing into Docusaurus

ikatyang:
Workaround for #3322


azz:
Nice workaround!

Issue #2853, opened 2017-09-18
Comment gets moved away


cilice:
**Prettier 1.7.0**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22const%20API%20%3D%20%7B%5CnloadLocale%3A%20(locale%3A%20string)%20%3D%3E%20%5Cn%20%20%2F%2F%20%24FlowFixMe%5Cn%20%20import(%2F*%20webpackChunkName%3A%20%5C%22locale%2F%5Brequest%5D%5C%22%20*%2F%20%60..%2F..%2Fpublic%2Fstatic%2Flocales%2F%24%7Blocale%7D.json%60)%2C%5Cn%7D%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22flow%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)
```sh
--parser flow
```

**Input:**
```jsx
const API = {
loadLocale: (locale: string) => 
  // $FlowFixMe
  import(/* webpackChunkName: "locale/[request]" */ `../../locales/${locale}.json`),
}
```

**Output:**
```jsx
const API = {
  loadLocale: (locale: string) =>
    import(// $FlowFixMe
    /* webpackChunkName: "locale/[request]" */ `../../locales/${locale}.json`)
};

```

**Expected behavior:**
```jsx
const API = {
  loadLocale: (locale: string) =>
    // $FlowFixMe
    import(/* webpackChunkName: "locale/[request]" */ `../../locales/${locale}.json`)
};

```

lydell:
This is certainly unexpected behavior. Possibly a side effect of #2748.

A few related questions:

- Does Flow, webpack or both break because of the moved comment?
- Why do you need `$FlowFixMe` there in the first place?

cilice:
Hey @lydell 

Flow, at least currently in 0.52 that we use (we haven't got around to update it to 0.54.1 yet), doesn't understand `import` with variables, only static paths work. To ignore this, we added `// $FlowFixMe` to suppress the message.

[...]

PR #3334, opened 2017-11-28
Fix formatting of comment in paren of call expression in arrow expression

jackyho112:
Fixes https://github.com/prettier/prettier/issues/2853#issuecomment-330275116.

Fix an edge case where a comment inside the parentheses of a call expression of an arrow function expression with no bracket would have the arrow function expression, instead of the call expression, as the enclosing node. Therefore, when checking for whether there is an argument, we might get the wrong result.  

For example, 

```
const API = {
  loader: () => import('./test' /* webpackChunkName: "test" */),
};
```

would get formatted to 

```
const API = {
  loader: (/* webpackChunkName: "test" */) => import("./test")
};
```

before this fix. But with this fix, the comment won't get moved.

Let me know if there is any improvement I could make. 😄


azz:
Looks good, thanks!

Issue #3338, opened 2017-11-28
Support new options on playground


ikatyang:
- `arrowParens`
- `proseWrap: "preserve"`

PR #3340, opened 2017-11-28
feat(markdown): add `proseWrap: "preserve"` option

### LINKED ISSUE #3302: Markdown: one sentence per line
**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBiVACAQgVwJYA2AJnlAOYYCGURGCAZhAE5ikWUYDOMAngXBjL4icADpRxWHhnqUmGGAAsBAIzxkycbhiZxKnaDOZUiEAA4w2GAAq6YluPLycFELjHNUCBBcozQAWjIIKxEVSngXCAA3R3c+LQA6DABJGAxnQQRHSm9pSjAwOAs4WiUIjEVKaKsOSABbesNuBMFhAUzo3JxKFX4jeQ4zJggAKzgwdOoyvXqMFRx0zXsrJUcOlw5YpmkzSlJ6HB9pjBwoXQB3ORJyDGGIIs5OZOs4c37ljG3pOCbLaFyVBGZ1oeyYljARzkGCuPCi9BhTDwllukBEVBoGCgEBUECI0gIeAA1louGYEDcKJZ6gILkiUexMboingarcoMjniAADQgcz-KCcZCgfQwZCyAicOC8lRMAokmAAZT2rHIyBgTBw0pApjA4tyUt5pCl4NslDI9Uo+sl2tGnAAHlg5WAFYrKDSADKkODWw18xZmRYAJl92rBJuQIEtTCJpguUB5IGGpBgAHU8EQlMgABwABl59ylqblZnVmu1ugAjvhdGaLVakBK-VL6ngy1reZw2PwAIo4CDwUO8mC9dOZxTIIPDuWENgAYQgjQbIGxUB9vJwUoAKr0hY2DXAAL6HoA)
```sh
# Options (if any):
--parser markdown
```

**Input:**
```markdown
## Building and enforcing a style guide

By far the biggest reason for adopting Prettier is to stop all the on-going debates over styles. It is generally accepted that having a common style guide is valuable for a project and team but getting there is a very painful and unrewarding process. People get very emotional around particular ways of writing code and nobody likes spending time writing and receiving nits.
```

**Output:**
```markdown
## Building and enforcing a style guide

By far the biggest reason for adopting Prettier is to stop all the on-going
debates over styles. It is generally accepted that having a common style guide
is valuable for a project and team but getting there is a very painful and
unrewarding process. People get very emotional around particular ways of writing
code and nobody likes spending time writing and receiving nits.
```

**Expected behavior:**

I would _love_ an option to have the markdown formatting follow the _one sentence per line_ technique. There's are good explanation in [Asciidoctor's recommended practices](http://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line).

```markdown
## Building and enforcing a style guide

By far the biggest reason for adopting Prettier is to stop all the on-going debates over styles.
It is generally accepted that having a common style guide is valuable for a project and team but getting there is a very painful and unrewarding process.
People get very emotional around particular ways of writing code and nobody likes spending time writing and receiving nits.
```

ikatyang:
Fixes #3302 

Since detecting sentences in different natural languages is hard, preserving original linebreaks should be a good alternative.

Ref: https://github.com/prettier/prettier/issues/3302#issuecomment-346891816


azz:
Maybe this should be the default? Currently `proseWrap: true` causes issues in BitBucket server because its version of Markdown is much more sensitive to line breaks.

ikatyang:
It will be an actual breaking change if we change the default, though the change is to preserve linebreaks, not sure if we should do so.

azz:
@vjeux Thoughts on making this the default?

vjeux:
I am not a fan of the current default, breaking words to 80 columns for markdown is imo a bad idea. Honestly I think that the best default would be to put everything in one line, but making preserve would be a step in the right direction from my point of view.

ikatyang:
The playground option should be in another PR, tracked in #3338.

SimenB:
So this basically means prettier won't touch paragraphs all all? If so I'd prefer the default being all in a single line - I expect prettier to normalize stuff not just ignore inconsistencies.

I do think this option is a good alternative, but I do think the value proposition of either of the two existing options are higher

ikatyang:
I originally choose `"always"` by default as it looks the most "prettier"-style, wrapping everything fits the printWidth, but if we're talking about the correctness, `"preserve"` should be the best since the markdown renderer used by  GitHub comment/BitBucket are linebreak-sensitive, or if we're talking about which one is the most useful, it should be `"never"`.

I don't have strong preference here.

suchipi:
My vote goes towards making "preserve" the default.

azz:
@SimenB In principle I agree, but unfortunately (as @ikatyang mentioned) there are multiple variants of Markdown that are all a little different when it comes to line breaks, so, to me, favouring not breaking the rendering of existing Markdown makes sense for the default value.

src/options.js
```diff
@@ -72,6 +72,16 @@ function normalize(options) {
     );
   }
 
+  /* istanbul ignore if */
+  if (typeof normalized.proseWrap === "boolean") {
+    normalized.proseWrap = normalized.proseWrap ? "always" : "never";
+
+    console.warn(
```
azz: Do we still use `deprecated.js`? It seems to be outdated.

ikatyang: It seems it's only used for `useFlowParser` warning.

I think we can refactor those option warnings/redirections using #3352.

azz:
LGTM.

Issue #2546, opened 2017-07-26
Travis build errors on trusty


azz:
I had to lock travis to run only on Ubuntu precise in https://github.com/prettier/prettier/commit/675688c8147810359040a15b0369a69a394998c5, because in the trusty containers they're rolling out the build fails.

It seems that trusty has an old version of yarn built in (v0.17.8).

Might be https://github.com/travis-ci/travis-ci/issues/7566#issue-219234365

azz:
Unpinned in #3510

PR #3349, opened 2017-11-29
Initial Python formatting support

patrick91:
Hi everyone! I've started hacking around prettier to add support for Python files since I wasn't happy with autopep8 and yapf (and I'm using prettier at work).

I'm working on this in my free time so I can't say when it'll be ready, there's a lot of things to
do and to figure out, I'll update this issue while working on it :)

**TODO:**

- [ ] Add comments
- [ ] Add type annotations 
- [x] Error when ASTExport is not installed
- [ ] ExtSlice
- [x] Nested if (currently broken)
- [ ] Nested with on python 2
- [x] Handle Strings (and f strings)
- [x] Decorators
- [ ] ... 

#3354


vjeux:
This is awesome! I wouldn’t be opposed to merging it in the prettier repo so that other people can help with the effort, if you are interested.

Down the road, it would be nice to compile python to js via emscriptem so that we don’t need the binary dependency and we can show it in the playground. Here’s a build of python 2.7 as an example: https://github.com/replit/empythoned

suchipi:
If there's a python parser written in JS that we can leverage, that'd be our first choice (will work in the browser and etc), but if that's not available, we might want to find a way to vendor ASTExport with prettier. We vendor all of our other parsers so that we don't have to worry about the AST format differing when the user has a different parser version installed.

patrick91:
@vjeux oh, I'll have a look at empython! thanks.
I'm interested in merging this into the prettier repo, but maybe it is a bit early? :)

@suchipi vendorizing ASTExport shouldn't be a problem, it is a simple script :)
I'm not entirely opposed on using a JS based parser, but I wasn't able to find one that's up to date
and also it means that we'd need to maintain it as well to keep up to date with newer python version (for example python 3.7 is [introducing f-strings](https://www.python.org/dev/peps/pep-0498/))

duailibe:
> I'm interested in merging this into the prettier repo, but maybe it is a bit early? :)

The HTML support in the repo is super experimental as well :-)

suchipi:
We had a couple of issues where people used the experimental HTML support and were surprised it didn't work, though. We'd want to be careful to ensure that doesn't happen again.

duailibe:
Yeah, you're right. I thought that was fixed with `getSupportInfo()`

ikatyang:
Experimental parser should not be added in `cli-constant.js` so as to avoid exposing, and it should be added in `support.js` with `since: undefined` just like html:

https://github.com/prettier/prettier/blob/fbbfa525171f36632ae879010cba92af9a5fd896/src/support.js#L176-L189

patrick91:
If that's fine with all of you ok :) but I'd wait to have a decision on the parser as well, just to prevent having to change some pieces related to the current AST structure!

jquense:
can I just add that this would change my life for the much better 👍

taion:
Very neat.

People have run Python on the client with emscripten. Otherwise, the Python binary can be optional, no?

Note that the standard AST parser in Python is part of the core stdlib: https://docs.python.org/3/library/ast.html, so there should be a strong presumption in favor of using that instead of a separate parser.

vjeux:
Python provides a way to query the ast as part of the language itself, so there’s very little incentive for third party implementations to build competing ones. Building a faithful parser is not an easy task and it is a critical requirement if we want to reprint the code without introducing different semantics. So the only realistic choice here is to use the one provided by default, like you did.

It’s fine to require a binary to be installed on the machine for now, it’ll take a while for the whole language to be supported anyway. If we can get it to run using emscriptem, then that’s perfect. Otherwise we’ll have to invest in getting third party dependencies like this one installed and have a server for the playground. We’ve been lucky not to have had to do it so far, but it’s not the end of the world.

As for merging it in, if you don’t feel like it yet, that’s ok. The most important thing right now is that you churn code as fast as possible to print as much as the language as possible. Then there’s going to be a phase to make it actually correct and one to make it pretty. There’s basically a clock where you won’t be motivated anymore / won’t be able to spend the time anymore, so we should do whatever you feel best to get the maximum amount of work done :)

taion:
That said, it might also be worth taking a look at Baron and RedBaron. They're Python CSTs (instead of ASTs), but it may be unnecessary to fully duplicate the Python code printing logic, especially as the language adds new features.

taion:
You may need to vendor in something like astexport, though – the release on PyPI only supports Python 3 (despite having a Python 2 wheel for some reason).

Like I imagine this will work just fine with whatever virtualenv you're in, and will use that Python binary, but not supporting Python 2 might not be ideal.

patrick91:
@vjeux totally agree with you! I think it might be ready to merge when it prints most of the things, right now it is missing a lot! :)

@taion I'll definitely try to support python 2 as well!

src/printer-python.js
```diff
@@ -0,0 +1,179 @@
...
+        parts.push(concat(["*", path.call(print, "vararg")]));
+      }
+
+      // add keyword arguments (**kwargs)
+
+      if (n.kwarg) {
+        parts.push(concat(["**", path.call(print, "kwarg")]));
+      }
+
+      if (n.kwonlyargs.length > 0) {
```
taion: kwonlyargs need to go before `**kwargs`

taion:
Looking at this some more, this might be really tricky without using something like Baron. The Python AST is a pure AST, as with ESTree. Unlike Babylon, it doesn't include information about comments at all.

vjeux:
@taion comments are handling in a different way with prettier. We take the raw list of comments and positions and attach them to nodes ourself. All we need to do is to implement a small lexer to extract out comments. This is super easy to do, even if python tooling doesn't provide it for us.

taion:
Ah, cool. Looks like `tokenize` gives you that, then. Though docstrings are sort of their own thing and do end up in the AST (I think?), which is sort of weird.

src/printer-python.js
patrick91: @taion why? This seems to work just fine :)

screen shot 2017-11-29 at 21 17 45

taion: image

patrick91: oh, got it! thanks!

azz:
This is really cool, but I'm going to play devil's advocate and list some of my concerns:

1. Python doesn't have a specification. This has bitten us a bunch of times supporting Less and SCSS, and Python is probably a more complex language.
2. This will be the first language that doesn't fall in to the "languages used in front-end web" bucket. Not necessarily an issue, but it expands the scope of Prettier, and maybe people will expect we add other languages like Ruby, C#, etc.
3. [astexport](https://github.com/fpoli/python-astexport): no commits in 2 years.
4. Having a binary dependency that the user has to install is annoying for the user, and really difficult for us to work with. Currently we bundle in very specific versions of parsers, even an earlier patch release will often cause our tests to fail. It also means we can't use it in the playground, which makes triaging issues really hard.
5. It's been mentioned already but we've had issues with people running the HTML formatter despite it being experimental - we should find a way to outright disable the code in production releases.
6. Python has some complexities in when and where it is whitespace sensitive. We need to be really careful with the Prettier doc primitives so we don't break where it would be a syntax error.
7. People will naturally ask for Python 2 support. Will this be possible?
8. In terms of maintenance, I'm not sure how many of the collaborators are proficient in Python? We'd probably need to pick up a couple of collaborators to support it.

All that being said, I'm fine with merging this in an experimental status, provided we address (5.)

As @vjeux mentioned, being able to compile the parser down to JavaScript would be a huge step towards getting this released.

taion:
@azz

What do you mean by "Python doesn't have a specification"?

The Python 2/Python 3 support thing is trivial. astexport basically does very little, and it's a few little syntax things that make it not support Python 2. There's not much necessary to get it to work with Python 2, and otherwise it's just using `ast` from the stdlib, which is doing all the hard work. Literally all `astexport` is doing is traversing the output of `ast.parse` and exporting it as JSON, and it'd be quite easy to write one internal to this repo and to drop that binary dependency.

The AST parser here is literally built in to Python and exposed as such, so it's not an extra binary dependency unless you think people are going to try to format Python without having Python installed somehow.

In some sense you actually can't neatly bundle it here, though, because the AST parser is somewhat specific to individual versions of Python, and as such the correct version of Python to use would be the system (or virtualenv Python).

Though I think the biggest question here is – "what's the cost of bridging over the Python AST to JS, versus porting over Prettier's printer to Python?"

Is most of the value-add here a set of language-agnostic heuristics that are complicated and subtle, or is there a lot to handle JS-specific idioms? If it's the latter, then perhaps there's not much presumptive reason to implement Python tooling in Node.

patrick91:
@azz thanks for the comments!

1. not sure by what you mean by this. If you're talking about formatting, the PEP8 document should be a good reference
2. that'd be expected yeah, but I really want this built into prettier! (and looks like I'm not the only one)
3. ASTExport is a trivial script, it can be vendorized if needed, I can send an email to the maintainer.
4. I don't see this as a real issue, since you'll be likely having python on your machine if you're writing code. Anyway, for the playground we might check if this works: https://github.com/pybee/batavia
5. +1 one for this, I don't want to let people think that formatting python works, there's a lot of work to do still :)
6. Yup. Do you think we can add some tests for this as well? Like validating all the test codes after formatting them?
7. Yes!
8. I'm willing to help, I'm sure we can find some more people that will contribute!

azz:
> What do you mean by "Python doesn't have a specification"?

JavaScript has [a spec](https://tc39.github.io/ecma262/), which informs us of the rules in writing valid JavaScript, as well as the runtime semantics of the code. I forgot that Python has a [grammar reference](https://docs.python.org/3/reference/grammar.html), which helps with this.

> In some sense you actually can't neatly bundle it here, though, because the AST parser is somewhat specific to individual versions of Python [...]

This is the case with JavaScript, too, but we make sure we can handle all JavaScript language features (including a lot of experimental features). This way the output of Prettier doesn't vary based on which version of Node.js, or which browser you are using.

> Is most of the value-add here a set of language-agnostic heuristics that are complicated and subtle, or is there a lot to handle JS-specific idioms?

The doc-building/printing portion of Prettier is mostly independent of any programming language. There have been tweaks to support certain features, like quote blocks in Markdown, and template strings in JavaScript, but those might be useful for printing Python too (e.g. multi-line strings)

azz:
I've met a guy who did a talk on batavia. It's a cool project. I might have a play with it and try and get `ast` working.

taion:
@azz 

Python AST parsing is version-specific though. It's not like how all syntactically-valid ES5 code is also syntactically-valid "modern JS". If you run `ast.parse` (really just `compile`) from Python 3 against Python 2 code, it may will barf on syntax errors.

In general, the Python backward compat policy doesn't say "no backward-incompatible syntax changes": https://www.python.org/dev/peps/pep-0387/. If you want to use `ast`, you need to use the `ast` package corresponding to the version of Python for which a project runs.

taion:
The trade-off ends up being language-specific printing stuff. The astor docs for example specifically flag that many attempts to print the Python AST to source have failed on corner cases: https://github.com/berkerpeksag/astor#id1.

dhirschfeld:
If it lowers the dev/support burden, maybe just support Python3? Many projects will be removing support for Python2 at the end of next year anyway.

azz:
Are there syntax breaking changes from Python 3.0 to 3.7?

patrick91:
@azz none that I can't think of, I previously said that f-string will be supported in py 3.7, but they are already in (since 3.6)

taion:
Those are strictly additive, though.

One thing that I can see being sort of problematic is that the rules around trailing commas have changed multiple times between versions.

I guess the question resolves to a tradeoff between:
- Porting Python AST printing to JS, in light of the caveats from e.g. astor's docs
- Porting Prettier's printing logic (or perhaps encapsulating it by calling out to Node?) from Python

dhirschfeld:
AFAICR the rules around trailing commas haven't changed since at least 2.2. Before that I couldn't say because that was the first version I used.

taion:
The rules trailing commas have changed in Python 3, Python 3.5, and Python 3.6. To wit:

```python
foo(
    **kwargs,  # Legal only in Python 3.5+.
)


def foo(
    **kwargs,  # Legal only in Python 3.6+.
):
    pass
```

Both of the above cases are syntax errors in lower versions of Python than the ones specified in the comments.

-----

On Python 2 v. Python 3, while Python 3 is "the future" (and almost all of our projects are on Python 3.5 or 3.6), the bulk of larger Python codebases are still on Python 2.

For example, @google and @dropbox, two of the most prominent users of Python, still mandate Python 2.7, and have a migration timeline of "approximately never". If you want an example closer to some of the people involved in this discussion, @facebook's new-ish Caffe2 _only_ supports Python 2, and while PyTorch might support both Python 2 and Python 3, it's not hard to guess which version the developers tend to use from issues like https://github.com/pytorch/pytorch/pull/3209.

So announcing the death of Python 2, at least in a commercial environment, is probably highly premature.

suchipi:
Considering that the AST effectively *must* match the target Python version, and the grammar changes aren't always backwards-compatible... maybe a better approach would be to export prettier's doc printer as a separate package (that can consume JSON), and do most of the heavy-lifting in python itself (outputting a JSON doc AST), where you can get the correct python version without shelling out

taion:
@suchipi 

It think it might be possible to just vendor a Python 2 and 3 compatible script that produces the document as JSON, call into it from here, then pass that into `printDocToString`.

Could that work? Naively, porting `doc-builders` to Python doesn't look like a ton of work, and could instead allow using astor to handle printing small chunks of AST to source.

suchipi:
That sounds good; then the python formatter could remain a part of the prettier binary, which would be a nicer/cooler UX. But we'd still be able to leverage the system/virtualenv python version.

suchipi:
Plus, it'd probably be best for maintainability if the Python printer was written in Python, since it's more likely for there to be interested new potential maintainers who are familiar with Python than those who are familiar with both Python and JS.

azz:
Multiparser would not be possible then. e.g. you couldn't format this:

~~~md
Some code:
```py
def foo():
    pass
```

taion:
Why is that? In any case we'd still be shelling out to Python – it'd just be a question of where the hooks were.

lipis:
There is a lot of code still written in Python 2 and will remain like that for years to come.. let's not drop Python 2 completely from the get go. While of course supporting both perfectly would be the ideal scenario, if the effort is huge then there is not much we can do about it.

vjeux:
I've seen the idea of porting the core prettier formatter to Python a few times here. This is a totally possible thing to do (yeah, I may have done it on two occasions already... :p). However, prettier, the project, gives you a bunch more things:

- All the infrastructure for snapshot tests, ast-indempotence... is already setup.
- There's a very robust CLI already available.
- Integrations with all the text editors/IDE on earth already there.
- A well oiled release process.
- A community of people that have worked on the codebase and can help review / when you have problems.
- A lot of existing users.
- The prettier brand.

lipis:
While we are at it.. since Python is coming soon, maybe support for YAML would be a good idea as well.. I know that the current implementation doesn't support comments so maybe someone can actually make it work so we can use it here.. :D

azz:
That would need to be done in https://github.com/mulesoft-labs/yaml-ast-parser/

tests/python_args_mixed/jsfmt.spec.js
```diff
@@ -0,0 +1 @@
+run_spec(__dirname, { parser: "python" });
```
azz: There was a change in #3356 so now it should be `run_spec(__dirname, ["python"]);`

src/parser-python.js
```diff
@@ -0,0 +1,32 @@
+"use strict";
+
+const { spawnSync } = require("child_process");
```
azz: FYI we support node v4 so we can't use destructuring assignments.

src/options.js
```diff
@@ -59,6 +59,12 @@ function normalize(options) {
     normalized.trailingComma = "none";
   }
 
+  // TODO: check if this changes users' options
+
+  if (normalized.parser === "python") {
+    normalized.tabWidth = 4;
```
azz: There might be a case to _only_ support tabWidth of 4 in Python. (e.g. say that this setting does not apply to Python)

src/printer-python.js
```diff
@@ -0,0 +1,538 @@
...
+
+  // python AST represent arguments and default
+  // value in two different lists, so we grab
+  // the list of the arguments and the list of
+  // default values and we merge them together and sort
+  // them by column. Then we iterated one by one and
+  // if the next element is a default value we merge it with
+  // the current one
+
+  const merge = [...n[argsKey], ...n[defaultsKey]].sort(
```
azz: Node v4 doesn't support the spread operator, so you can use `n[args.key].concat(n[defaultsKey])`

src/printer-python.js
```diff
@@ -0,0 +1,538 @@
...
+      let parts = [
+        ifType,
+        path.call(print, "test"),
+        ":",
+        indent(concat([hardline, printBody(path, print)]))
+      ];
+
+      if (n.orelse.length > 0) {
+        if (n.orelse.length === 1 && n.orelse[0].ast_type !== "If") {
+          parts = [
```
azz: parts.push(line, "else:", indent(concat([line, concat(path.map(print, "orelse"))])))

src/options.js
taion: Google’s Python style guide says 2 spaces.

azz: Oh wow. Google doesn't follow PEP-8?

taion: The internal Google Python style guide mandates 2 spaces. You can see it in a number of their open-source projects: https://github.com/tensorflow/tensorflow/blob/v1.4.0/tensorflow/python/ops/nn_ops.py. It's real fun when you're backporting stuff into your own code base that is otherwise getting linted to ensure 4 spaces.

And of course if you've done anything with numpy, it's impossible to follow PEP 8 there, since implicit boolean casts for numpy arrays just raise exceptions.

And, like, `logging` doesn't even follow the camel_case thing. So, whatever... lol PEP 8? I dunno.

patrick91:
@azz many thanks for the review, I'll fix that soon :)

@vjeux and all, I've had some troubles with strings (as you have seen from twitter). I'll be trying this python library (which seems to be well maintained): https://github.com/gristlabs/asttokens
This will give us access to the original text for each token, so I should be able to resolve the issues with the strings :)

I'll keep you posted!

src/options.js
lipis: Please don't force tabWidth 4 for Python.. 2 is cool as well :)

patrick91: totally, when I did that the idea was to "force" 4 as the default, but then you could change it, I'll make sure that this will be the case :)

azz: Maybe something like this then:

```js
    if (normalized.parser === "python" && !normalized.tabWidth) {
      normalized.tabWidth = 4;
    }
```

src/printer-python.js
patrick91: I can't believe I didn't know that `push` accepted multiple parameters! All fixed now :)

patrick91:
Quick update, I haven't had much time to work on this, but I have a working python scripts that uses asttoken to fetch the original source code for a token, so I can work easily with strings. I'll give you another update hopefully soon :)

patrick91:
Before working on strings I've started working on Python2 support, I've a question, since we are using the current version of python available in the env there some tests that will file, for example tests using newer syntax (matrix multiplication, f strings and so on).

I'd like to test both python2 and 3, by moving the python3 only things in separated tests. What would be the best way to do this? Shall I create a python2/python3 parser, or a python specific configuration option?

@vjeux @azz

vjeux:
Argg, it's annoying, looks like they did change/remove some pieces of syntax in python 3: https://docs.python.org/3.0/whatsnew/3.0.html#changed-syntax

It looks like those are pretty subtle though. My inclination would be to only have a single python type and first parse it with python 3 and if it throws a syntax error, then parse it with python 2. If this  works in practice, then it's going to be great as, as a user, you don't have to think about the two and we can map .py files to python and be done with it.

If this solution doesn't end up working, then yeah, we should have two different python syntax. My recommendation would be to the default "python" to be python3 and "python2" if you really want to.

taion:
You should be able to handle both AST formats – i.e. given a syntax tree, you can probably emit valid Python for whatever version, net of maybe trailing commas. 

Can you just integrate tox (the Python package) into the test process? You can easily detect the Python version.

src/printer-python.js
```diff
@@ -0,0 +1,533 @@
...
+
+      // add varargs (*args)
+
+      if (n.vararg) {
+        parts.push(concat(["*", path.call(print, "vararg")]));
+      }
+
+      // add keyword only arguments
+
+      if (n.kwonlyargs.length > 0) {
```
azz:
```js
if (n.kwonlyargs && n.kwonlyargs.length > 0) {
```

azz:
Let's get the build green!

- You'll need to add `"python"` to these two files:

  https://github.com/prettier/prettier/blob/3fa2229a6291aa109a640aca199336b16c681509/scripts/build/build.js#L11-L19

  https://github.com/prettier/prettier/blob/3fa2229a6291aa109a640aca199336b16c681509/scripts/build/build-docs.js#L10-L18

- Delete `lineno` and `col_offset` from AST in this array:
  https://github.com/prettier/prettier/blob/3fa2229a6291aa109a640aca199336b16c681509/src/clean-ast.js#L36-L60

- Looks like there's a couple of unknown AST nodes still: `"Print"` and `"TryFinally"`.

- [#pullrequestreview-82345615](#pullrequestreview-82345615)

patrick91:
@vjeux I did a quick try/catch to use first python3 if available falling back to using python2 if not.

I still think we need to be able to test both python2 and python3 on the CI, there is at least one case where the AST is different between the 2 versions (the Print statement), I might add that in as an option (eg, forcePythonVersion=2).

@taion you mean to use tox so I can test multiple python versions?

@azz thanks for the review, I'll fix that soon :)

azz:
I've got a workaround for playground support here: https://github.com/prettier/prettier/issues/3354#issuecomment-350547062

patrick91:
I've added an option to change python version. There are still some differences on different python 3 versions (like f strings and @ multiplication), I'd like to handle all those cases, while retaining support for previous python versions, but I'm not sure where we should stop, we definitively shouldn't have all the python 3 versions running on ci.

I'd suggest to support the following versions:

- Python 2.7+
- Python 3.4+

vjeux:
What do you mean by differences between python 3 versions. Are those only additive? Or did they remove syntax in later versions?

It's fine to be able to print code that's invalid in earlier versions.

patrick91:
@vjeux yes, only additions :) ok then, I'll add the latest python version to travis :)

patrick91:
Almost green (I hope :))

@vjeux @azz is there a reason why travis is using ubuntu precise (12.04). The latest python version there is 3.2 and I wasn't able to find a PPA with python3.6 :(

Also when I install python3.5 the default python3 remains to python3.2 which isn't ideal for us.
A solution would be to use pyenv so that we can set the versions for python2 and python3 without changing any configuration. What do you think?

vjeux:
Since we’re likely not going to ship this by requiring people to have a specific version of python installed on their machine, I’d say do whatever it takes to get it working on your dev environment and in travis. We’ll figure something better once we actually want to ship it.

As for running pyenv or something similar, it’s not a crazy idea. At fb we’ve used nvm to control node versions on ci when the specific versions we wanted were not available by travis/circle.

patrick91:
Ok, I've added pyenv. It made the build quite slow. I'm sure when can optimise it by adding its folder to the cache :)
I'll have a look at it tomorrow!

I've noticed that on python3.2 the AST module is broken, for example doesn't return the column offset for the arguments. This seems to be fixed in python3.4.

azz:
> @vjeux @azz is there a reason why travis is using ubuntu precise (12.04). The latest python version there is 3.2 and I wasn't able to find a PPA with python3.6 :(

See #2546

taion:
BTW, it's more standard to use a Travis build matrix than to install different versions of Python in the same Travis build: https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

Travis has native support for multiple Python versions via their own setup, so it'd obviate a lot of the caching-related things.

patrick91:
@taion uhm, that would be better, but I have to run some tests in both python 2 and 3 and some tests only on python 3. How would you do it? Using environment variables, like the example:

```yaml
matrix:
  include:
    - python: '3.5' # this is not strictly necessary
      env: EXTRA_TESTS=true
```

scripts/build/build-docs.js
```diff
@@ -14,7 +14,8 @@ const parsers = [
   "graphql",
   "postcss",
   "parse5",
-  "markdown"
+  "markdown",
+  "python"
```
lipis: Shall we do it alphabetically?! Scales better..

src/clean-ast.js
```diff
@@ -54,7 +54,9 @@ function massageAST(ast, parent) {
       "trailingComma",
       "parent",
       "prev",
-      "position"
+      "position",
+      "lineno",
+      "col_offset"
```
lipis: Does the order matters? Can we sort them?

scripts/build/build-docs.js
azz: At the rate we add parsers I _really_ don't think it matter

taion:
@patrick91 Sure – or you can just check the Python version directly. Or use an env var.

If you want to get version in a nice machine-readable way without parsing the string, run:

```
$ python -c "import sys; print(sys.version_info.major)"
```

This will work on pretty much any Python version you'd care about.

taion:
Neat!

You may want to configure the build to not run the Node tests for both Python versions, though.

Maybe set up the build matrix to `include` the Python 2 build, and set some env var for running only Python tests.

patrick91:
@taion I'm still experimenting :D it seems that it still has `python3.2` as the default python:

vjeux:
What about using environment variable to tell where python is.

In JS:
```js
process.env.PYTHON_PATH
```

When you run the tests:

```js
PYTHON_PATH=/wherever/python2 npm run tests
PYTHON_PATH=/wherever/python3 npm run tests
```

taion:
My bad, Travis doesn't actually support multiple specified languages in the build matrix. Sorry.

patrick91:
I'm not entirely sure on how travis works when specifying multiple languages and versions: http://s.patrick.wtf/7ZiMNeU2hS

I'll do some other tests (and then squash the commits :) )

taion:
@patrick91 

It actually doesn't work at all: https://github.com/travis-ci/travis-ci/issues/4090

I guess that pyenv thing is the best available with Travis.

patrick91:
Seems to be slightly faster now :)

azz:
When this is green I'm happy to review and merge.

patrick91:
@azz would like to merge it even if it is missing support for comments?

azz:
Yeah, don't want the PR to get larger and larger, better to review smaller changes with greater focus.

.gitignore
```diff
@@ -13,3 +13,5 @@
 .DS_Store
 coverage
 .idea
+__pycache__/
+*.pyc
```
azz: nit: newline here

src/printer-python.js
```diff
@@ -0,0 +1,750 @@
...
+
+const docBuilders = require("./doc-builders");
+const concat = docBuilders.concat;
+const join = docBuilders.join;
+const hardline = docBuilders.hardline;
+const line = docBuilders.line;
+const softline = docBuilders.softline;
+const group = docBuilders.group;
+const indent = docBuilders.indent;
+// const ifBreak = docBuilders.ifBreak;
```
azz: nit: delete if not used.

src/printer-python.js
```diff
@@ -0,0 +1,750 @@
...
+
+function printPythonString(raw, options) {
```
azz: Any way to better re-use some of `util.js printString()`?

tests/python3_await/await.py
```diff
@@ -0,0 +1,6 @@
+async def example():
+    await channel.put(json.dumps({
+        'text': data.text,
+        'image': image,
+        'username': data.user.screen_name,
+    }))
```
azz: nit: newline here (affects snapshot)

src/parser-python.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (error) {
+    throw new Error(error);
+  }
+
+  return executionResult;
+}
+
+function parse(text, parsers, opts) {
+  const pythonExectuable = `python${opts.pythonVersion}`;
```
azz: Do we want do something like:

* try parse with `"python3"`,
* if error, then try parse with `"python"`

We actually do this with TypeScript ts and tsx.


Then we wouldn't need an option here.

patrick91: If not python version is specified it will use `python` directly, which means that will use the default python in your current environment, could be python2 or python3 depending on your setup (eg if you are in a virtualenv).

The option is mainly to run tests with both python2 and python3 :)

src/printer-python.js
patrick91: maybe, I think we need to make the js one a little bit more generic, in python you can have 2 different types of strings:
1. single quotes (eg. "hello")
2. triple quotes, for doc strings and multiline ("""hello world""")

All of them can have one or two modifiers (r, f and u). I'll think a bit on how to reduce code duplication :)

src/parser-python.js
azz: Ok cool. So we don't need to document it.

patrick91:
@azz can you do another review? :)

patrick91:
I've started another repo with (one for now) some unit tests that I would use to verify to validity of the formatted files, I had a bug where it was printing correct python code but it was doing a different thing.

https://github.com/patrick91/prettier-python-tests

I still have to add the steps to download prettier, prettify the code and then run the tests :)

azz:
I'll do another soon, but I'm considering the advice from @jlongster in https://github.com/prettier/prettier/issues/3503#issuecomment-352802184 and thinking about how we can share Prettier as a tool, while having separate maintainable codebases. To that end I'm working on a plugin API in #3536, and once that's done we can move this PR to a new repository (`prettier/prettier-python`) with a dependency on `prettier`.

This way we can achieve:
- Independent versioning, providing the flexibility to release fixes at a faster rate than Prettier.
- Seperate CI for faster iteration and ability to add a larger range of tests, potentially using more Python versions.
- Ability to watch only Python issues on GItHub

All the while keeping:
- The core Prettier platform, without having to install separate tools to format Python and the languages that Prettier currently supports.
- Editor integration support

We can do the same for the PHP PR in #3521, and potentially the HTML support in #3534.

Thoughts?

patrick91:
Sounds good to me, especially for this: 

> Ability to watch only Python issues on GItHub

I can see it being a problem when having many languages baked into prettier.

Let's see what the other folks think :)

jlongster:
Yes @azz thank you for working on that! I'd like to start watching prettier's issues again but I can't imagine trying to lump a whole new language into here and the amount of issues that it would add on top of everything. I really don't think another language should be included here. Markdown is passable because it's a big part of the JS ecosystem, and it's probably easy enough. But we should keep prettier as a JS tool.

taion:
I'm wondering about that community aspect, though. Like, as set up, the only people who can actually contribute to this are people who are familiar with Python AST manipulation _and_ who can work in Node.

So I'd ask then how this matches @vjeux's points in https://github.com/prettier/prettier/pull/3349#issuecomment-348348159.

vjeux:
What I want is for people to stop wasting time formatting their code. The reality is that people use many different programming languages on a daily basis and not just JS. It turns out that prettier is a really good foundation for building a pretty printer in any language so I feel like it would be a shame for it not to expand to more places now that it's covering the core use case very well.

I also found out that a big reason that languages build silos around themselves is that all the tooling is completely different. One of the reason I joined the Nuclide team was to break silos by not having to learn a completely new IDE when from switching languages, which is one less ramp up time. I believe that having the same pretty printer which shares similar decisions and design philosophy is going to be one more stone in that direction.

Now, this is --not-- going to be easy as forces to push language ecosystems away from each other are strong, but I don't think it's impossible and I believe that having PHP and Python pretty printers as part of prettier is something that can work. Having a separate repo to manage issues and PR is a good idea.

taion:
@vjeux 

I definitely agree with what you're saying.

My concern is maybe more around the API boundaries here. If the idea per #3536 is a plugin API, then, in that world where there is a level of separation, it would be great to implement the AST token printing in Python rather than in JS. That way Prettier could just leverage e.g. astor for Python, &c.

The core printing API, the platform API, the CLI stuff – that's all great. To me, especially if prettier-python lives in a different package entirely, it just looks weird that `printer-python.js` is JS instead of Python, especially when it's essentially just writing out a serializable data structure. Writing Python tooling in JS is a wall, too.

BTW, what's that about IDEs? Maybe I'm spoiled, but I use JetBrains IDEs for just about everything... they seem to support pretty much everything under the sun.

j-f1:
@taion Interesting idea. One of the important components of Prettier is the actual logic that figures out where to break lines. Maybe the Python component could generate the document using translated builders, then serialize the result to JSON and had that over to Prettier’s JS so that it can print the prettified code.

taion:
@j-f1 Yeah – I mentioned this upthread, but it's not a super-compelling thing for part of Prettier proper. As a Prettier plugin that's in a separate repo aimed at Python people, it's a better idea IMO. It looks like `doc-builders` just produces a serializable structure, so I'd imagine that would be a better API for plugins aimed at other languages.

azz:
I've copied all the commits from this PR to a new repository :tada: https://github.com/prettier/prettier-python. @patrick91 I've made you an admin of the new repo.

PR #3350, opened 2017-11-29
chore(playground): add new options

### LINKED ISSUE #3338: Support new options on playground

ikatyang:
Fixes #3338

- [x] wait for #3340

### DIFF
--- a/src/cli-constant.js
+++ b/src/cli-constant.js
@@ -82,7 +82,6 @@ const detailedOptions = normalizeDetailedOptions({
     category: CATEGORY_FORMAT,
     forwardToApi: true,
     description: "Include parentheses around a sole arrow function parameter.",
-    default: "avoid",
     choices: [
       {
         value: "avoid",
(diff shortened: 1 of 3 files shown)


src/cli-constant.js
```diff
@@ -82,7 +82,6 @@ const detailedOptions = normalizeDetailedOptions({
     category: CATEGORY_FORMAT,
     forwardToApi: true,
     description: "Include parentheses around a sole arrow function parameter.",
-    default: "avoid",
```
suchipi: Are you going to fill this back in once we've decided on the new default?

ikatyang: Oops, forgot to mention why I removed this line: we use the default value from `options.js` for all format options.

https://github.com/prettier/prettier/blob/fbbfa525171f36632ae879010cba92af9a5fd896/src/cli-util.js#L561-L579

suchipi: Hmm, that's feels a bit unintuitive, but changing it is out of the scope of this PR

azz:
Currently the playground uses the released version of Prettier (from package.json) so this won't actually work if we merge it.

duailibe:
I added the 1.9 milestone just to help us remember it's related but feel free to remove if you think someone might think this is a blocker.

suchipi:
So to clarify, this should be merged post release?

azz:
Yes. After prettier's own dependency on prettier is updated to 1.9.0

PR #3353, opened 2017-11-29
(docs) Fix broken footnote links

smably:
These links didn't work: https://prettier.io/docs/en/index.html#footnotes

Now they do!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -54,5 +54,5 @@ If you want to learn more, these two conference talks are great introductions:
 
 \* _Well actually, some original styling is preserved when practical—see [empty lines] and [multi-line objects]._
 
-[empty lines]: rationale.md#empty-lines
-[multi-line objects]: rationale.md#multi-line-objects
+[empty lines]: rationale.html#empty-lines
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -54,5 +54,5 @@ If you want to learn more, these two conference talks are great introductions:
 
 \* _Well actually, some original styling is preserved when practical—see [empty lines] and [multi-line objects]._
 
-[empty lines]: rationale.md#empty-lines
-[multi-line objects]: rationale.md#multi-line-objects
+[empty lines]: rationale.html#empty-lines
+[multi-line objects]: rationale.html#multi-line-objects
```
ikatyang: cc @JoelMarcey

This looks like a bug, the following two cases should be considered the same and [be transformed correctly](https://docusaurus.io/docs/en/doc-markdown.html#linking-other-documents).

```md
[aaa](bbb.md#ccc)
```

```md
[aaa]

[aaa]: bbb.md#ccc
```

JoelMarcey: Thanks for bringing this up @ikatyang -- we will take a look.

PR #3364, opened 2017-11-30
Add note about proseWrap option under printWidth for clarity

RyanZim:
A user in gitter was wondering how to disable prose wrapping in markdown, but hadn't noticed the option; he was thinking he'd need to set `printWidth` somehow to change this. I added a small note for clarity, since `proseWrap` is at the bottom of the page where some people don't notice it.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -19,6 +19,8 @@ Specify the line length that the printer will wrap on.
 | ------- | --------------------- | ------------------- |
 | `80`    | `--print-width <int>` | `printWidth: <int>` |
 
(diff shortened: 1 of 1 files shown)


docs/options.md
```diff
@@ -19,6 +19,8 @@ Specify the line length that the printer will wrap on.
 | ------- | --------------------- | ------------------- |
 | `80`    | `--print-width <int>` | `printWidth: <int>` |
 
+(If you don't want line wrapping when formatting Markdown, you can set the [Prose Wrap](#prose-wrap) option to disable it.)
```
duailibe: This `Prose Wrap` casing is weird.. shouldn't it be either `prose-wrap` or `proseWrap` right?

duailibe:
Thanks!

PR #3370, opened 2017-12-01
Support toplevel interpolations in graphql tags

### LINKED ISSUE #3318: Support top-level interpolations in gql tags
apollo supports _toplevel_ interpolations in `` gql`...` `` tagged template literals (it throws an error if you try to use interpolations in nested situations). This allows sharing GraphQL fragments between queries without loosing static analysis (both eslint-plugin-graphql and apollo-codegen support this).

Currently, we skip gql tags if they contain interpolations. It would be awesome if we could support toplevel ones though!

My idea is that we should _try_ to parse and print every literal part independently. If _all_ succeeds, we join the results up with the interpolations. If any of them fails, we bail out like today.

What do you think?

**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBFArnAJwE9MBeTAcwEcAbAAwB0or9jMBVNQzYJzTXFwIAKAJYATJAxAAjCDOkBKXlH78AdJs6EAInBgBDUTTR81m9QDECohONOrMAXyYuoTACTB2AZQCiAEoA+jp+ACoAggCSADI+QZYBEQDiALJ+AHJhTl6JUZk68Ykp6VludCAANCAQAA4woujIoAYYyABmBiZw1TIEBmAA1vo+tQOiUBTIMAT41eIQYB1dXNUTQjAACv0UALYGy93VAFZoAB4AQv1DIwa7cDETcIerNbgwte8ATC89IGMEITIWQGGREGjQKr-GywADqEhgAAtkAAOAAM1VqBAgXFh-Vq01mfwIcBYohJ2wMewOSE6RxAXF2okJcwZEwoNDgeAg8F+1UMMnh4iRyC+-P6xnZAGEILt9sCoNBntVBHAwqC0L8nE4gA)

**Input:**
```jsx
const Query = gql`
query User {
  user(id:"bob"){
    ...UserDetails
    ...Friends
  }
}

${USER_DETAILS_FRAGMENT}${FRIENDS_FRAGMENT}
`
```

**Output:**
```jsx
const Query = gql`
query User {
  user(id:"bob"){
    ...UserDetails
    ...Friends
  }
}

${USER_DETAILS_FRAGMENT}${FRIENDS_FRAGMENT}
`;

```

**Expected behavior:**

```jsx
const Query = gql`
  query User {
    user(id: "bob") {
      ...UserDetails
      ...Friends
    }
  }

  ${USER_DETAILS_FRAGMENT}
  ${FRIENDS_FRAGMENT}
`;
```

lydell:
Fixes #3318.

### DIFF
--- a/src/multiparser.js
+++ b/src/multiparser.js
@@ -23,17 +23,20 @@ function printSubtree(path, print, options) {
   }
 }
 
-function parseAndPrint(passedOptions) {
-  const options = Object.assign({}, passedOptions, {
-    parentParser: passedOptions.parser,
+function parseAndPrint(text, partialNextOptions, parentOptions) {
+  const nextOptions = Object.assign({}, parentOptions, partialNextOptions, {
+    parentParser: parentOptions.parser,
     trailingComma:
-      passedOptions.parser === "json" ? "none" : passedOptions.trailingComma
+      partialNextOptions.parser === "json"
+        ? "none"
+        : partialNextOptions.trailingComma,
+    originalText: text
   });
-  const ast = require("./parser").parse(options.originalText, options);
+  const ast = require("./parser").parse(text, nextOptions);
   const astComments = ast.comments;
   delete ast.comments;
-  comments.attach(astComments, ast, options.originalText, options);
-  return require("./printer").printAstToDoc(ast, options);
+  comments.attach(astComments, ast, text, nextOptions);
+  return require("./printer").printAstToDoc(ast, nextOptions);
 }
 
 function fromMarkdown(path, print, options) {
@@ -46,12 +49,7 @@ function fromMarkdown(path, print, options) {
       const style = styleUnit.repeat(
         Math.max(3, util.getMaxContinuousCount(node.value, styleUnit) + 1)
       );
-      const doc = parseAndPrint(
-        Object.assign({}, options, {
-          parser,
-          originalText: node.value
-        })
-      );
+      const doc = parseAndPrint(node.value, { parser }, options);
       return concat([style, node.lang, hardline, doc, style]);
     }
   }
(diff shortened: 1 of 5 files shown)


tests/multiparser_js_graphql/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -82,6 +78,153 @@ const query = gql\`
...
+    test # comment
+  } # comment
+  \${test} # comment
+  \${test} # comment
+
+  \${test} # comment
+
+  # comment
+  \${test} # comment
+\`
```
lydell: All the comments after the interpolations here are currently moved to the next line. Is this a deal-breaker? This seems to need some larger changes to change.

azz: Probably not a deal-breaker. I'm guessing comments are pretty uncommon here. I've only ever used comments in GraphQL schemas.

azz:
Multiparser refactor looks good. Haven't reviewed GraphQL changes properly yet but test outputs look good.

lydell:
I need to figure out why the `parentParser` thing doesn’t work, though.

lydell:
I need to figure out why the `parentParser` thing doesn’t work, though.

src/multiparser.js
```diff
@@ -5,44 +5,38 @@ const docUtils = require("./doc-utils");
...
-      return fromBabylonFlowOrTypeScript(path, options);
+      return fromBabylonFlowOrTypeScript(path, print, options);
     case "markdown":
-      return fromMarkdown(path, options);
+      return fromMarkdown(path, print, options);
   }
 }
 
-function fromMarkdown(path, options) {
+function parseAndPrint(passedOptions) {
```
duailibe: You'd need to pass the previous and next options here so the `parentParser` thing works.. see `printSubtree` how it builds `next.options` but gets the `parentParser` from `options`

lydell:
I think I'm happy with this now. Would be awesome if it could make it into the 1.9 release!

lydell:
> Haven't reviewed GraphQL changes properly yet 

@azz Do you want to take a look before merging, or should I just press the green button?

Issue #3335, opened 2017-11-28
Footer is empty on the mobile version of the website


lipis:
<img width="406" alt="screen shot 2017-11-28 at 11 46 38" src="https://user-images.githubusercontent.com/125676/33315761-ea56ef30-d431-11e7-978f-480f2cb906fc.png">

JoelMarcey:
Thanks for bringing this to our attention. I will look at having this fixed asap.

azz:
@JoelMarcey is there a fix in the pipes for this one? Otherwise we can go ahead with #3373

JoelMarcey:
@chrisburgin95 The repo exists, but isn't officially public yet. Very soon though. We are just finishing up a few things. Sorry for the confusion.

@azz I was planning to put work on a fix this upcoming week. How about you land #3373 and then once the fix has landed, and you `npm update` we can remove the override?

JoelMarcey:
@lipis Hi! Someone put a PR to fix this in the Docusaurus core - https://github.com/facebook/Docusaurus/pull/299 - I merged that in.

lipis:
@joelmarcey awesome!

PR #3373, opened 2017-12-01
Fixed website footer on mobile (#3335)

chrisburgin95:
The footer was not showing on mobile devices. This was because docusaurus was hiding it. I used `static/overrides.css` to override docusaurus and give the footer some mobile styling. 

I used the exact same size (735) max-width media query that docusaurus uses, that is where that width came from.

screen shot 2017-12-01 at 15 20 29


azz:
Thanks!

Issue #3352, opened 2017-11-29
Build playground options automatically


duailibe:
Recently after every a new option or a new language we have to manually add it to playground. @suchipi suggested building the HTML automatically and that would be nice. I've created this issue just so we don't forget it.

To keep in mind: we are now using the dist code to bundle the playground so we don't have access to the `cli-constant.js` at this point. We could get it from `src/` but that would mean exposing unsupported options. Another options is having a similar `getSupportInfo` but for options.

ikatyang:
`getOptionInfo()` looks good to me.

We can extract the format option part from `cli-constant`, proposed interface:

```ts
function getOptionInfo(version?: string): OptionInfo[];

interface OptionInfo {
  name: string;
  since: string; // available since version
  deprecated?: string; // deprecated since version
  type: 'int' | 'boolean' | 'choice';
  default: boolean | number | string | {
    [since: string]: boolean | number | string; // for default value that has changed before
  };
  range?: { start: number, end: number, step: number }; // for int, recommended range for playground
  choices?: ChoiceInfo[]; // for choice
}

interface ChoiceInfo {
  name: string;
  value: boolean | string; // boolean for the option that is originally boolean type
  since?: string; // undefined if available since the first version of the option
  deprecated?: string;
}
```

suchipi:
Would we use OptionInfo.since to decide whether it's shown on the playground? Or rather, how would we detect/handle options that aren't released yet?

ikatyang:
PR branch uses `getOptionInfo()`, which returns the latest options, and master branch uses `getOptionInfo(package_json_version)`, which returns the stable options.

It should work fine this way I guess.

duailibe:
Actually the official playground always builds from the NPM version and the PR preview runs `yarn build` on src/ and uses that for the playground.

So just the output would already work..

azz:
I left space in the return value of `getSupportInfo` for other things. Could add `options` to it.

```
{
  languages: [...],
  options: [...]
}
```

duailibe:
Nice!

duailibe:
I tried taking a stab at building the playground options automatically but my CSS-fu isn't that great. Do you have any suggestions on how to split the options in the groups like today? Or suggestions of a redesign to stop depending on those `div`'s?

ikatyang:
IMO, a vertical panel would be better as the scrolling is available, just like [babel REPL](https://babeljs.io/repl/).

duailibe:
[...]

PR #3374, opened 2017-12-01
Change playground options layout

duailibe:
Context: https://github.com/prettier/prettier/issues/3352#issuecomment-348387918

Preview: https://deploy-preview-3374--prettier.netlify.com/playground


duailibe:
Actually for now I think I'd prefer a :+1:/:-1: on the general layout, because I think the options layout can use some love (hiding some options depending on the parser) and I think these extensions would be great for a React-based site.

website/pages/playground/index.html
```diff
@@ -339,58 +357,14 @@ <h1>Prettier <span id="version"></span></h1>
 
   <div class="bottom-bar">
     <div class="bottom-bar-buttons">
+      <button type="button" class="btn" id="button-options">
+        See options
```
ikatyang: `Show options`?

ikatyang:
It'd be better to categorize options, but it should be in another PR.

- Global
  - printWidth
  - tabWidth
  - useTabs
  - parser
- JavaScript
  - semi
  - singleQuote
  - bracketSpacing
  - jsxBracketSameLine
  - trailingComma
- Markdown
  - proseWrap
- Special
  - rangeStart
  - rangeEnd
  - insertPragma
  - requirePragma
- Debug
  - show AST
  - show doc
  - show second format

lipis:
Is there a blocker on this one?! Can we have that in time for 1.9?

website/pages/playground/index.html
```diff
@@ -192,47 +189,27 @@
       }
     }
 
-    .options-details {
-      padding: 8px 0;
+    .options-container {
+      max-height: 100%;
+      overflow-y: scroll;
```
lydell: Could you change this to `overflow-y: auto;`. It looks strange to have a disabled scroll bar when there’s no overflow (that’s what it looks like on Windows and Linux).

website/pages/playground/index.html
lipis: That copy is better.. let's change that as well.

duailibe:
@lipis there’s no need to rush with this, not related to 1.9 at all. I’m not sure we shoud merge as-is yet, current layout has better options separation

website/pages/playground/index.html
duailibe: Yeah I do, thanks for noticing

website/pages/playground/index.html
```diff
@@ -192,47 +189,45 @@
...
+      padding-bottom: 5px;
+    }
+
+    .sub-options > * {
+      margin-left: 10px;
+    }
+
+    label {
+      font-family: "Consolas", "Courier New", Courier, monospace;
+      padding: 10px 0;
```
ikatyang:
```diff
-padding: 10px 0;
+margin: 10px 0;
+display: block;
```

image

image

website/pages/playground/index.html
```diff
@@ -382,27 +356,63 @@ <h1>Prettier <span id="version"></span></h1>
...
+
+          <label><input type="checkbox" id="insertPragma"> --insert-pragma</label>
+          <label><input type="checkbox" id="requirePragma"> --require-pragma</label>
+        </details>
+        <details class="sub-options">
+          <summary>Debug</summary>
+
           <label><input type="checkbox" id="ast"> show AST (debug)</label>
           <label><input type="checkbox" id="doc"> show doc (debug)</label>
           <label><input type="checkbox" id="output2"> show second format (debug)</label>
```
ikatyang: I think we don't need `(debug)` anymore :)

website/pages/playground/index.html
```diff
@@ -318,57 +313,36 @@ <h1>Prettier <span id="version"></span></h1>
...
-      </button>
-    </div>
-
-    <details class="options-details">
-      <summary class="options-summary">Options</summary>
+    <details id="options-details" class="options-container">
+      <summary class="options-summary"></summary>
+      <div class="options">
+        <details class="sub-options">
+          <summary>Global</summary>
```
ikatyang: It'd be better to open (`<details open>`) `Global` and `JavaScript` by default I think.

duailibe: Yeah that could work.. I'd rather persist the last state though.

(That'd be so easy to in a react component 😅)

lydell:
@duailibe Now the scrollbar only appears if needed, great! However, the whole page still scrolls instead of just the sidebar in Firefox.

duailibe:
Now tested on chrome, firefox and safari on mac :-)

azz:
The show options button doesn't seem to work on my phone (S8 Chrome)

azz:
Wait now it does - bad caching

duailibe:
I was just gonna ask if you could hard reload on the phone 😅

Issue #3108, opened 2017-10-26
Line break after object property


sharmilajesupaul:
Just to give you some context on this, I've been working with @ljharb to bring Prettier to @airbnb. 
We noticed a few cases where Prettier’s formatting could be adjusted to more closely align with our style guide.

cc. @vjeux

**Prettier 1.7.4**

If an object has a value that exceeds the specified `printWidth`, Prettier creates a line break after the object property. This ambiguates the value of the property when the value spans multiple lines.

[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAEnADYkQDqEATiQCYAUAlPsADpT5f4wAWOAZwB0AmAEN4+ALysO3bgLBioAWTh8IdJPlTkA7gGFo8AB4wAMtADmlqFYByYgLZxpUmQDFzAeQoB9A297ABUAUQANYIBlISiAVSiABVD7KIBJIL8AQUTkrPM5eSKAfnwAEVCANTSDUL9vROCM1KEDOIAldpTgvwrq2sKi7m0+mrqGpqCYlW8AITTzOpzEgBpB-ABfAG4ODZAVkAgAB1xoAWRQMVFkVDESATgDgCNqMTAAa3Uoo7ecO2QYNR0I8QHQIGAbncHgcAFYCEyzV4fL7OODmP5wSH3EEQdAwI54gBMWOhIB+1Ae1GQICeYieAE9yFB9mTqH8YBQcHQ+MgAIwABn5BweThwAKBIIEfysJDgAEV0BB4OLgQdxE9OdzeMhCWrXjgSNKjE4nGJqXcSCz0A9gnTzkhbtiNhsgA)
```sh
# Options (if any):
--single-quote
--print-width 100
--trailing-comma "all"
```

**Input:**
```jsx
function HelloWorld() {
  this.state = {
    scanMethod: flowContextLongLongName === FLOW_CONTEXTS.SUSPENSION_APPEAL
      ? DEVICE_OPTIONS.CURRENT_DEVICE
      : DEVICE_OPTIONS.MOBILE_APP,
  };
}
```

**Output:**
```jsx
function HelloWorld() {
  this.state = {
    scanMethod:
      flowContextLongLongName === FLOW_CONTEXTS.SUSPENSION_APPEAL
        ? DEVICE_OPTIONS.CURRENT_DEVICE
        : DEVICE_OPTIONS.MOBILE_APP,
  };
}
```

[...]

Issue #2482, opened 2017-07-14
Inconsistency breaking assignments


bakkot:
For example,

```js
aParticularlyLongAndObnoxiousNameForIllustrativePurposes = 
  anotherVeryLongNameForIllustrativePurposes;
```

[formats as](https://prettier.io/playground/#%7B%22content%22%3A%22aParticularlyLongAndObnoxiousNameForIllustrativePurposes%20%3D%20%5Cn%20%20anotherVeryLongNameForIllustrativePurposes%3B%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D)

```js
aParticularlyLongAndObnoxiousNameForIllustrativePurposes = anotherVeryLongNameForIllustrativePurposes;
```

Similarly there is no break after the `=` in

```js
aParticularlyLongAndObnoxiousNameForIllustrativePurposes =
  'a very long string for illustrative purposes'.length;
```
or
```js
aParticularlyLongAndObnoxiousNameForIllustrativePurposes =
  anotherVeryLongNameForIllustrativePurposes();
```
although there *is* in
```js
aParticularlyLongAndObnoxiousNameForIllustrativePurposes =
  anotherVeryLongNameForIllustrativePurposes.length;
```
and
```js
aParticularlyLongAndObnoxiousNameForIllustrativePurposes =
  anotherVeryLongNameForIllustrativePurposes + 1;
```

I'd expect every one of these to break after the `=`, since otherwise line length exceeds 80 characters. (Though I thought it used to... is this an intentional, or a regression?)

[...]

PR #2398, opened 2017-07-05
Improve JSX Formatting

suchipi:
* Arrow Function Expressions returning JSX will now add parens when the JSX breaks
* Conditional expressions within (or containing) JSX are formatted in a more natural way (for JSX), especially when chained
* JSX in logical expressions (|| or &&) is ~~always wrapped in parens~~ wrapped in parens when the JSX breaks
* `null` in conditional expressions within (or containing) JSX is never wrapped in parens

Fixes #2208

TODO:
- [x] Fix JSX-mode ternaries always breaking
- [x] Add comments to tests and make sure I have good coverage

PR #3394, opened 2017-12-05
Wrap object values on a newline in parens (refs: #3108)

sharmilajesupaul:
Hi there 👋🏽
This change addresses https://github.com/prettier/prettier/issues/3108. Object values that contain a break will be wrapped in parentheses.

i.e. This is how long values were printed originally:
```javascript
const g = classnames({
  "some-prop":
     longLongLongLongLongLongLongLongLongLongLongLongLongTooLongVar || 1337
});

const x = {
  ABC:
    "12345678901234567890123456789012345678901234567890123456789012345678901234567890"
};

const k = {
  blablah:
     "aldkfkladfskladklsfkladklfkaldfadfkdaf" +
     "adlfasdklfkldsklfakldsfkladsfkadsfladsfa" +
     "dflkadfkladsfklkadlfkladlfkadklfjadlfdfdaf"
};
```
This is how they are printed with this change: 
```javascript 
const g = classnames({
  "some-prop": (
     longLongLongLongLongLongLongLongLongLongLongLongLongTooLongVar || 1337
  )
});

const x = {
  ABC: ( 
    "12345678901234567890123456789012345678901234567890123456789012345678901234567890"
  )
};

const k = {
  blablah: (
     "aldkfkladfskladklsfkladklfkaldfadfkdaf" +
     "adlfasdklfkldsklfakldsfkladsfkadsfladsfa" +
     "dflkadfkladsfklkadlfkladlfkadklfjadlfdfdaf"
  )
}
```
cc. @ljharb, @vjeux

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1063,7 +1063,7 @@ function genericPrintNoParens(path, options, print, args) {
           printedLeft = printPropertyKey(path, options, print);
         }
         parts.push(
-          printAssignment(
+          printObjectProperty(
             n.key,
             printedLeft,
             ":",
@@ -4598,6 +4598,19 @@ function printAssignmentRight(rightNode, printedRight, canBreak, options) {
   return concat([" ", printedRight]);
 }
 
+function canBreakAssignment(leftNode, rightNode) {
+  return (
+    (isBinaryish(rightNode) && !shouldInlineLogicalExpression(rightNode)) ||
+    (rightNode.type === "ConditionalExpression" &&
+      isBinaryish(rightNode.test) &&
+      !shouldInlineLogicalExpression(rightNode.test)) ||
+    ((leftNode.type === "Identifier" ||
+      isStringLiteral(leftNode) ||
+      leftNode.type === "MemberExpression") &&
+      (isStringLiteral(rightNode) || isMemberExpressionChain(rightNode)))
+  );
+}
+
 function printAssignment(
   leftNode,
   printedLeft,
@@ -4610,15 +4623,7 @@ function printAssignment(
     return printedLeft;
   }
 
-  const canBreak =
-    (isBinaryish(rightNode) && !shouldInlineLogicalExpression(rightNode)) ||
-    (rightNode.type === "ConditionalExpression" &&
-      isBinaryish(rightNode.test) &&
-      !shouldInlineLogicalExpression(rightNode.test)) ||
-    ((leftNode.type === "Identifier" ||
-      isStringLiteral(leftNode) ||
-      leftNode.type === "MemberExpression") &&
-      (isStringLiteral(rightNode) || isMemberExpressionChain(rightNode)));
+  const canBreak = canBreakAssignment(leftNode, rightNode);
 
   const printed = printAssignmentRight(
     rightNode,
@@ -4628,6 +4633,36 @@ function printAssignment(
   );
 
   return group(concat([printedLeft, operator, printed]));
+}
+
+function printObjectProperty(
+  leftNode,
+  printedLeft,
+  operator,
+  rightNode,
+  printedRight,
+  options
+) {
+  const canBreak = canBreakAssignment(leftNode, rightNode);
+
+  const needsParens = options.parser !== "json" && canBreak;
+
+  let printed = printAssignmentRight(
+    rightNode,
+    printedRight,
+    canBreak,
+    options
+  );
+
+  if (needsParens) {
+    printed = concat([
+      indent(concat([ifBreak(" (", ""), line, printedRight])),
+      softline,
+      ifBreak(")", "")
+    ]);
+  }
+
+  return group(concat([printedLeft, ":", printed]));
 }
 
 function adjustClause(node, clause, forceSpace) {
--- a/tests/object-prop-break-in/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/object-prop-break-in/__snapshots__/jsfmt.spec.js.snap
@@ -23,8 +23,9 @@ const x = {
 };
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 const x = {
-  ABC:
+  ABC: (
     "12345678901234567890123456789012345678901234567890123456789012345678901234567890"
+  )
 };
 
 `;
@@ -63,8 +64,9 @@ const a = classnames({
 });
 
 const b = classnames({
-  "some-prop":
+  "some-prop": (
     this.state.longLongLongLongLongLongLongLongLongTooLongProp === true
+  )
 });
 
 const c = classnames({
@@ -84,8 +86,9 @@ const f = classnames({
 });
 
 const g = classnames({
-  "some-prop":
+  "some-prop": (
     longLongLongLongLongLongLongLongLongLongLongLongLongTooLongVar || 1337
+  )
 });
 
 `;
--- a/tests/objects/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/objects/__snapshots__/jsfmt.spec.js.snap
@@ -104,18 +104,20 @@ const blablah =
   "dflkadfkladsfklkadlfkladlfkadklfjadlfdfdaf";
 
 const k = {
-  blablah:
+  blablah: (
     "aldkfkladfskladklsfkladklfkaldfadfkdaf" +
     "adlfasdklfkldsklfakldsfkladsfkadsfladsfa" +
     "dflkadfkladsfklkadlfkladlfkadklfjadlfdfdaf"
+  )
 };
 
 somethingThatsAReallyLongPropName =
   this.props.cardType === AwesomizerCardEnum.SEEFIRST;
 
 const o = {
-  somethingThatsAReallyLongPropName:
+  somethingThatsAReallyLongPropName: (
     this.props.cardType === AwesomizerCardEnum.SEEFIRST
+  )
 };
 
 `;


ljharb:
(cc @bakkot)

ikatyang:
You need to rebase this PR so as to get `netlify` deployment success.

> error Command "build-docs" not found.

sharmilajesupaul:
oh ok, I was wondering what netlify was, just rebased @ikatyang

bakkot:
This seems... strictly worse.

I am confused as to why anyone would want these parentheses. What value do they have over the existing indentation?

bakkot:
Separately, either way I think it would be good to have some examples and/or snapshots with multiple properties:

```js
const g = classnames({
  "some-prop": (
    longLongLongLongLongLongLongLongLongLongLongLongLongTooLongVar || 1337
  ),
  "some-other-prop": (
    longLongLongLongLongLongLongLongLongLongLongLongLongTooLongVar || 1337
  )
});
```

ljharb:
We don't want arbitrary line breaks after `=` or `:`; multiline things should generally be wrapped in parens/brackets/braces so it's visually a single unit.

bakkot:
> multiline things should generally be wrapped in parens/brackets/braces so it's visually a single unit

It seems to me it is already visually a single unit. One of the fundamental assumptions of code formatting in general is that indentation indicates structure. The fact that subsequent lines are indented is sufficient to indicate it is a unit; that's precisely what indentation indicates. Why is more necessary?

ljharb:
In JavaScript, a line break isn’t sufficient; im not sure how to verbalize it concretely. Arbitrarily breaking immediately after : and = “looks weird”, and airbnb’s style is to forbid it.

bakkot:
A line break together with indentation _is_ sufficient as long as you can trust your formatting, which if you're using prettier you can.

To my eye the parentheses added by this PR look _considerably_ weirder than what Prettier outputs currently. They seem like noise. If airbnb wants them, [prettier-eslint](https://github.com/prettier/prettier-eslint) seems a better path than modifying prettier, especially at this otherwise relatively stable point in its life.

Also, if the idea is that this should hold for both `:` and `=` it seems inconsistent to make this change for one and not the other. There are currently some [issues](https://github.com/prettier/prettier/issues/2482) with linebreaks after `=`, mind.

(Of course, don't put too much weight on my opinion; for the purposes of discussion I'm just some random.)

sharmilajesupaul:
@bakkot fwiw, regardless if whether it _looks weird_ or not, I do think that the parens aid readability in some cases. 

i.e.
this is harder for me to read through:
```javascript
const b = classnames({
  "some-prop":
    this.state.longLongLongLongLongLongLongLongLongTooLongProp === true,
  "some-other-prop": this.state,
  "third-prop":
    this.state.longLongLongLongLongLongLongLongLongTooLongProp,
  "fourth": this.state.longProp,
});
```
than the clear delimiters in this eg:
```javascript 
const b = classnames({
  "some-prop": (
    this.state.longLongLongLongLongLongLongLongLongTooLongProp === true
  ),
  "some-other-prop": this.state,
  "third-prop": (
    this.state.longLongLongLongLongLongLongLongLongTooLongProp
  ),
  "fourth": this.state.longProp,
});
```
But a lot of these discussions go nowhere since 'readability' is very subjective.
 
It seemed like we got consensus on this issue which is why I opened this PR. And as far as I know, I don't think there is an eslint rule that will cover this case? (I may be wrong about that @ljharb)

I am happy to move forward exploring other options to fit our case if there are strong feelings against this though.

bakkot:
Hm. I disagree on readability there; the former seems more readable, where each top-level-indent item within the object is a property name (rather than `)`). It's subjective, like you say.

> It seemed like we got consensus on this issue which is why I opened this PR.

Ah. Well, if so, don't mind me. I don't think the discussion in #3108 amounts to what I'd normally consider consensus, though; I'd've commented there earlier if I'd realized that seemed to be the case.

> And as far as I know, I don't think there is an eslint rule that will cover this case?

That can be fixed! But... if there's not, how is it something airbnb is enforcing currently? If this would be a change even for people already using airbnb's style guide I'd be even more surprised to see it land.

> I am happy to move forward exploring other options to fit our case if there are strong feelings against this though.

Well, like I say, don't put too much weight on my opinions. I'd want to hear from more people, especially prettier maintainers (who are responsible for guarding the precious "minimal changes to formatting" resource), either way.

sharmilajesupaul:
ah ok, I was thinking that that's why #3108 was labeled a bug? but my apologies if not 😔

This case isn't in our style guide yet either, its not something we've had to think about. But I believe @ljharb plans on adding a section for it once we work out whether this belongs in Prettier or not.

We've recently updated the style guide to include the case where Prettier bumps implicit arrow function bodies to a newline. And plan on changing our `operator-linebreak` rule to disallow linebreaks after  `=` in assignments. 

This another step in a similar vein. We can definitely also propose a lint rule for this case.

ljharb:
Generally, the airbnb guide legislates the things people tend to type; some of the patterns Prettier outputs, while defensible/sensible, aren't things people tend to type on their own. In order to be compatible with Prettier, we have to come up with internal consensus on each pattern Prettier outputs (for, against, or "with tweaks"). In this case, we don't want line breaks after `=` or `:` unless the RHS is surrounded by parens/braces/brackets.

suchipi:
FWIW, I prefer the parens, but it feels like it goes against prettier's original philosophy of not adding unnecessary parens (not that I haven't [broken that rule before](https://github.com/prettier/prettier/pull/2398), though).

That said, I think that in this case, with parens and without parens both look kinda bad... So my preference towards parens is not very strong.

duailibe:
> ah ok, I was thinking that that's why #3108 was labeled a bug? but my apologies if not 😔

My understanding is that the following case is considered a bug:

```js
foo = {
  prop:
    condition
      ? consequent
      : consequent
};
```

because condition could remain in the same line as `prop` if fitting.

Then @vjeux commented:

> If no one is interested in writing the real fix, I'm not opposed to adding parenthesis, but that seems like a crutch.

Which means it's not 100% clearer that this PR is actually _fixing_ the bug.

duailibe:
As for the PR, I share @suchipi's opinion.

sharmilajesupaul:
hm ok, so I *think* if that was a bug, it's fixed now? Now it inlines the condition in a ternary beside the prop ([playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzCFMC8mwAOlJpgEYCGATkldWtS62+0x1689+2gMFCWvPuwBemAPyYA5AEZZmBgCYyAXxAAaEBAAOMAJbpkoOrQgB3AAp0EaZCGoA3CIYAm2pxmTZqAGzQ4HUpaajAAazgYAGU9cMMoAHNkGFoAV2CQdwgwXwCgnUSg2hhrMKSAW2p8wKyAKzQADwAhMMjomOpKuAAZRLhawt10mD1RlSGs+NoSxxpKAE9-aC89WkSYAHUPGAALZAAOAAYddYggrbC9R3W4EudBnVo4AEd0wxfy6iqapD86jogpVDKkMlk0Ikkv44ABFdIQeBTHQwaiUHbufbIFQosKGfxQgDCEEq1UcUGgTxA6SCABU0Q5-gU4Op1EA)). This happens regardless of whether the condition in a ternary fits on the same line as the prop.

So this change won't affect that behavior. That being said, this does need more discussion, should we close this and move it to the issue instead? @ljharb @vjeux?

duailibe:
@sharmilajesupaul I don't think it is, see [playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzCFMC8mwAOlJpgEYCGATkldWtS6202p11y8+-wMECAXoSIkQ1CZgD8mAOQBGeZgYAmMgF8QAGhAQADjACW6ZKDq0IAdwAKdBGmSSAbhGMATXZIzJs1ABs0OD1KWmowAGs4GABlAwjjKABzZBhaAFcQkA8IMD9A4L0k4NoYW3DkgFtqAqDsgCs0AA8AIXComNjqKrgAGSS4OqL9DJgDMbVh7ITaUucaSgBPAOhvA1okmAB1TxgAC2QADgAGPQ2IYO3wg2cNuFKXIb1aOABHDONXiupq2qR-PU9MEqsY0plsmgkskAnAAIoZCDwaZ6GDUSi7DwHZBqVHhYwBaEAYQgVRqzig0GeIAywQAKuinADCnBNJogA)

sharmilajesupaul:
@duailibe oh interesting, so it only creates a line break if the ternary has a conditional expression ([playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzCFMC8mwAOlJpgEYCGATkldWtS62+0x1689+2gMFCWvPuwBemAPyYA5AEZZmBgCYyAXzJlIUDDjyFiZCjXqNR3IVd4WxdvpIJESIai+lzFyzGqiaoIAA0IBAADjAAlujIoHS0EADuAAp0CGjIrgBuEBEAJkGuGMjY1AA2aHDBlLTUYADWcDAAyqG1EVAA5sgwtACulSC5EGDFZRXB7RW0MEk1HQC21KPlAwBWaAAeAEI19Y1N1PNwADLtcMvjIb0wodcqFwOttFMZNJQAnqXQBaG07TAAdTyMAAFsgABwABmCvwgFQBNVCGV+cCmmXOwVocAAjr0IljZtQFkskCUVsEKvMIt0+gM0O0OqU4ABFXoQeAPYIwaiUIG5UHIFRcmoRUoMgDCEHmiwyUGgGJAvQqABUeelSWM4Op1EA)).

I could update this PR to be specific to that case. But I'm not sure if parens are better than inlining the conditional and letting it exceed the max length?

duailibe:
Given the options (when the test exceeds the print width):

- breaking the conditional
- breaking the conditional (with parens)
- always inline the conditional

My favorite is inliing and my least favorite is adding the parens. But I don't know if we want that churn now when people upgrade (maybe it won't be a big deal).

j-f1:
What’s the status on this? Should we fix the conflicts and merge or close without merging?

ljharb:
Please don’t close; we’re happy to update whatever needs updating.

Beyond rebasing and resolving conflicts, what more is needed?

lipis:
So what do we do with this one? Almost two years since last comment.

brodybits:
I would downvote this idea because of the following kind of case:

There are cases such as ".config.js" files which could be mostly JSON-like objects, and this change could make some config objects no longer JSON compliant.

ljharb:
@brodybits if they're in a `.js` file they're already guaranteed not to be JSON; being JSON-like in format isn't a property that needs preserving (if it is, then they should be in a `.json` file).

sharmilajesupaul:
closing due to inactivity, feel free to re-open to push this forward

evilebottnawi:
Reopen - https://github.com/prettier/prettier/pull/3394#issuecomment-387458825

PR #3395, opened 2017-12-05
Upgrade Prettier to Prettier 1.9

### LINKED ISSUE #3385: Deprecation warning is printed in master
```
~/forks/prettier master ⇡ 
❯ git log -1
commit ceabc48e01b4b8977ae98f0b4bc774bd2659a294 (HEAD -> master, upstream/master, gql-cooked)
Merge: c1623d88 24036480
Author: Lucas Duailibe <lucasds@gmail.com>
Date:   2017-12-03 03:25:17 -0300

    Format CSS prop for emotion (#3381)

~/forks/prettier master ⇡ 
❯ echo a | ./bin/prettier.js 
[warn] `proseWrap` with value `false` is deprecated. Prettier now treats it as: `{ proseWrap: "never" }`.
a;

```

azz:
Fixes #3385

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,0 @@
-{
-  "proseWrap": false
-}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -315,7 +315,7 @@ Lots of small fixes, mainly for TypeScript.
 * [JSX] Break if opening element breaks (#942)
 * Parenthesize function expressions in expression position (#941)
 * update the README to add a pre-commit hook (#944)
-* Fix #951: properly parenthesize \** expressions (#952)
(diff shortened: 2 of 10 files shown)


.prettierrc
```diff
@@ -1,3 +0,0 @@
-{
-  "proseWrap": false
```
ikatyang: `"proseWrap": "never"`?

azz: I'm happy with the default?

ikatyang: I'm fine with both `"preserve"` (default) and `"never"`, just want to make sure it's intended.

azz: Yeah it was intentional.

azz:
Wow - I got a merge failed error because we clicked merge at the same time. Never seen that before!

Issue #3404, opened 2017-12-05
[playground] Parsers not loaded for multiparser


duailibe:
In the playground, we only load the parsers when necessary so that means the `format` may fail when we switch parsers in the playground. For that reason we have the following code:

https://github.com/prettier/prettier/blob/e09359d2428bdac40c30cd6a82c92a627d1e1d05/website/static/worker.js#L98-L104

But after the multiparser refactor in #3370 we no longer have the `parser` property in the exception:

https://github.com/prettier/prettier/blob/e09359d2428bdac40c30cd6a82c92a627d1e1d05/src/printer.js#L125-L137

This fails on the playground ([link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucBnVADDBuEAaECABxgEtpVlQBDAJ1ogHcAFOhSlagNwlIBN8IaqhjIAZtQA2qOAQBGtamADWcGAGUiS0lADmyGLQCuskHwhhxUmQR0zaMZot0BbalemmAVqgAeAIUUVNXVqFzgAGR04DxtCIxgiBIAmWNMtWntkEDlqOQBPSWhBIlodGAB1fhgAC2QADgAGAlKIGQrFImzSuHsuGIJaOABHI1Ihp2pXdyQJTwIZF1IDY1NUHV1JOABFIwh4NIIYPKq+WuRko8VSSQ2AYQgXN2yoaAGQIxkAFTyOOZkAL4AoA)):

```js
css``;
```

If you change the parser to css/less/scss, then go back to babylon, it starts working.

I'm not sure how to fix this with the current architecture of the multiparser. We'd have to have multiple try/catch around the code

duailibe:
@azz Looks like we can simply change here:

https://github.com/prettier/prettier/blob/e09359d2428bdac40c30cd6a82c92a627d1e1d05/website/static/worker.js#L22-L30

to call `importScripts`

azz:
You could, but we don't know the global export so it's difficult to determine what to return.

duailibe:
I don't understand, how's this any different from what we're doing right now?

```diff
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -26,6 +26,13 @@ self.require = function require(path) {
   if (path === "./third-party") {
     return {};
   }
+  if (~path.indexOf("parser-")) {
+    var parser = path.replace(/.+-/, "");
+    if (!parsersLoaded[parser]) {
+      importScripts("lib/parser-" + parser + ".js");
[...]

PR #3406, opened 2017-12-05
Fix multiparser on playground

azz:
One possible fix for #3404

@duailibe if you can get https://github.com/prettier/prettier/issues/3404#issuecomment-349298055 working then that might be better as this won't work until 1.9.1 is released.


duailibe:
Yeah this is pretty much the solution for the playground to continue to work.

I'm not that familiar with `importScripts` so I'm not sure the strategy in #3407 works. I don't have a preference either way for any of the solutions

PR #3407, opened 2017-12-05
Fix multiparser on playground

### LINKED ISSUE #3404: [playground] Parsers not loaded for multiparser

duailibe:
Fixes #3404, supersedes #3406

### DIFF
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -77,7 +77,6 @@ self.onmessage = function(message) {
   }
 
   if (message.data.doc) {
-    lazyLoadParser("babylon");
     try {
       doc = prettier.__debug.formatDoc(
         prettier.__debug.printToDoc(message.data.text, options),
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -26,7 +26,17 @@ self.require = function require(path) {
   if (path === "./third-party") {
     return {};
   }
-  return self[path.replace(/.+-/, "")];
+
+  if (~path.indexOf("parser-")) {
+    var parser = path.replace(/.+-/, "");
+    if (!parsersLoaded[parser]) {
+      importScripts("lib/parser-" + parser + ".js");
+      parsersLoaded[parser] = true;
+    }
+    return self[parser];
+  }
+
+  return self[path];
 };
 
 importScripts("lib/index.js");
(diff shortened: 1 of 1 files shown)


website/static/worker.js
```diff
@@ -26,7 +26,17 @@ self.require = function require(path) {
...
   }
-  return self[path.replace(/.+-/, "")];
+
+  if (~path.indexOf("parser-")) {
+    var parser = path.replace(/.+-/, "");
+    if (!parsersLoaded[parser]) {
+      importScripts("lib/parser-" + parser + ".js");
+      parsersLoaded[parser] = true;
+    }
+    return self[parser];
```
azz: Is this actually defined? If it works it's a good solution

duailibe: As I understand the previous code, that's exactly how it was working before. We called `importScripts` below just for this to work `return self[path.replace(/.+-/, "")]`

duailibe: It's a bit hackish, since it requires knowledge of the `moduleName` in the rollup config

PR #3415, opened 2017-12-05
Change cursor and remove outline in summaries on playground

lipis:
The focus outline doesn't makes much sense there


website/pages/playground/index.html
```diff
@@ -230,6 +230,11 @@
       font-size: 14px;
       font-weight: bold;
       padding-bottom: 5px;
+      cursor: pointer;
```
lipis: `default` could also work..

PR #3433, opened 2017-12-07
feat(support-info): add `options` field

ikatyang:
This will unblock #3352. cc @duailibe

And I'm going to refactor `options.js`/`cli-constant.js` using `getSupportInfo()` in another PR to address https://github.com/prettier/prettier/pull/3350#discussion_r153839262,  https://github.com/prettier/prettier/pull/3340#discussion_r154045632, and https://github.com/prettier/prettier/issues/3352#issuecomment-349683443.


duailibe:
Thank you for doing this! ❤️ 

Looks good to me, but I'd wait for another review :)

src/cli-constant.js
```diff
@@ -386,17 +387,12 @@ const minimistOptions = {
     )
 };
 
-const usageSummary = `
-Usage: prettier [options] [file/glob ...]
+const usageSummary = dedent(`
```
lydell: Isn’t this supposed to be used as a template tag?

duailibe:
If we can get it in 1.9.2, we wouldn't have to wait another release to do #3352 (since the site's playground only uses the released version)

src/support.js
```diff
@@ -189,28 +193,354 @@ const supportTable = [
...
+  {
+    name: "parser",
+    since: "0.0.10",
+    category: CATEGORY_GLOBAL,
+    type: "choice",
+    default: "babylon",
+    description: "Which parser to use.",
+    choices: [
+      { value: "flow", description: "Flow" },
+      { value: "babylon", description: "JavaScript" },
```
duailibe: Can you reorder flow and babylon? Just to keep the same order from the docs and the playground?

(not that it makes that much difference but still...)

duailibe:
I've started working on the playground rewrite and it's looking pretty good, I have it almost done with preact, with no build steps :-)

One of the problems I faced now is when in a PR preview, which version should I pass to `getSupportInfo` to get the unreleased opts? It would be nice if it was always `getSupportInfo()` so I don't have to special case PR previews or the official playground, but we don't bump the `package.json` until we release a new version...

(Actually the PR preview changes the package.json to `pr+XXXX` but that can be changed.. I don't know what we should change it to)

I can't simply append `-beta` because new stuff is likely to be on `1.10` and then new options would not be enabled. So I guess the correct approach would be to bump to something like `1.10-beta` right after the `1.9.X` release, or something...

Let me know if I wasn't clear!

ikatyang:
`getSupportInfo(null, { showUnreleased: !!isPR })`?

duailibe:
@ikatyang that would work. I hope that’s not too hackish though

ikatyang:
Anyone wants to do another review? I'd like to merge it and send a followed up PR.

duailibe:
@azz @lydell @suchipi anyone have any comments about this? Can we merge it?

tests_integration/__tests__/support-info.js
```diff
@@ -2,29 +2,66 @@
...
+    const info = getCoreInfo(version);
+    if (index === 0) {
+      test(`with version ${version}`, () => {
+        expect(info).toMatchSnapshot();
+      });
+    } else {
+      const previousVersion = testVersions[index - 1];
+      const previousInfo = getCoreInfo(previousVersion);
+      test(`with version ${previousVersion} -> ${version}`, () => {
+        expect(snapshotDiff(previousInfo, info)).toMatchSnapshot();
```
azz: Neat idea with snapshotDiff!

azz:
Unless I'm reading this wrong we now have two copies of each option definition (cli-constant.js and support.js`)?

ikatyang:
I have a followed up PR for it, see https://github.com/ikatyang/prettier/compare/feat/support-options...ikatyang:refactor/options, or should I add them in this PR?

duailibe:
I fixed all the conflicts, @azz can you take a look?

I had to change the `getSupportInfo`.. I'm not sure how useful it is to load external plugins there since the information is about prettier's own functionality?

azz:
We need to split it into multiple files:

- `src/common/options.js`
  - `--parser`
  - `--print-width`
  - `--tab-width`
  - `--use-tabs`
  - `--insert-pragma`
  - `--require-pragma`

- `laguage-js/options.js`:
  - `--no-semi`
  - `--single-quote`
  - `--no-bracket-spacing`
  - `--jsx-bracket-same-line`
  - `--arrow-parens`
  - `--trailing-comma`
 
- `language-markdown/options.js`
  - `--prose-wrap`
 
And then expose them as `options` on the `printer` objects.

j-f1:
`--single-quote` is also used by [Markdown](https://github.com/prettier/prettier/blob/a041914551c0c3886a90ba76d855dce95d3363cc/src/language-markdown/printer-markdown.js#L642) and [CSS](https://github.com/prettier/prettier/blob/a041914551c0c3886a90ba76d855dce95d3363cc/src/language-css/printer-postcss.js#L502).
`--no-bracket-spacing` is also used by [GraphQL](https://github.com/prettier/prettier/blob/a041914551c0c3886a90ba76d855dce95d3363cc/src/language-graphql/printer-graphql.js#L163).

azz:
That's fine, we should re-declare those flags in the printers that use them.

src/language-js/index.js
```diff
@@ -132,6 +133,7 @@ const printers = {
 };
 
 module.exports = {
+  options,
```
azz: `options` should probably be a property of a specific printer rather than a plugin, as a plugin can provide multiple languages/printers. In this case it should be `printers.estree.options`

duailibe:
@ikatyang is there anything else to tackle on this PR?

ikatyang:
@duailibe No, the CLI option generation will be in another PR. This one should be good to go.

PR #3444, opened 2017-12-10
Print comment after arrow function params

### LINKED ISSUE #3418: Flow comment syntax: misplaced return type of arrow function
The playground is throwing an error like ReferenceError: constants is not defined for the flow parser, so I'll do this manually:

**Prettier 1.9.1**
```sh
# Options (if any):
--single-quote
--trailing-comma es5
```

**Input:**
```jsx
const run = (cmd /*: string */) /*: Promise<void> */ => {
  debug(`Running ${cmd}`);
  return new Promise((resolve, reject) => {
    exec(cmd, (error, stdout, stderr) => {
      if (error) {
        debug(`Error running ${cmd}`);
        debug(stderr);
        return reject(error);
      }

      debug(stdout);
      return resolve();
    });
  });
}
```

**Output:**
```jsx
const run = (cmd /*: string */ /*: Promise<void> */) => {
  debug(`Running ${cmd}`);
  return new Promise((resolve, reject) => {
    exec(cmd, (error, stdout, stderr) => {
      if (error) {
        debug(`Error running ${cmd}`);
        debug(stderr);
        return reject(error);
      }

      debug(stdout);
      return resolve();
    });
  });
};
```

**Expected behavior:**

The return type (`: Promise<void>`) should not be moved inside the parenthesis because this is a syntax error.

duailibe:
Fixes #3418

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -506,8 +506,21 @@ function genericPrintNoParens(path, options, print, args) {
         );
       }
 
-      if (hasDanglingComments(n)) {
-        parts.push(" ", comments.printDanglingComments(path, options, true));
+      const dangling = comments.printDanglingComments(
+        path,
+        options,
+        /* sameIndent */ true,
+        comment => {
+          const nextCharacter = util.getNextNonSpaceNonCommentCharacterIndex(
+            options.originalText,
+            comment
+          );
+          return options.originalText.substr(nextCharacter, 2) === "=>";
+        }
+      );
+      if (dangling) {
+        console.log(dangling);
+        parts.push(" ", dangling);
       }
 
       parts.push(" =>");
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -506,6 +506,23 @@ function genericPrintNoParens(path, options, print, args) {
...
+        comment => {
+          const nextCharacter = util.getNextNonSpaceNonCommentCharacterIndex(
+            options.originalText,
+            comment
+          );
+          return options.originalText.substr(nextCharacter, 2) === "=>";
+        }
+      );
+      if (dangling) {
+        console.log(dangling);
```
vjeux: remove :)

duailibe: oops sorry!

duailibe:
Yeah I'm not a fan of it either but I'm afraid we don't have a better way to do this. We should try not to reuse this everywhere

vjeux:
Btw, we still need to fix the flow comments on the flow parser. Right now they are being reprinted as without the comments which break code. What we need to do is to check if they were comments in the first place and print them as comments instead.

I think that this would then be much more safe as we won't be able to mess it up.

duailibe:
Yep.. in that case I'm not sure how useful it is to try and keep these comments in the right place for babylon

vjeux:
We should do both imo :)

duailibe:
I'll just wait for the 1.9.2 release, and then merge this (probably not suitable for a patch release) and see if I can get the flow comments working :-)

eqyiel:
@duailibe thank you for taking the time to do this!

PR #3449, opened 2017-12-10
Print flow type annotation comments as comments

duailibe:
Flow AST does not expose that a type annotation `a /*: string*/` comes from a comment and so prettier currently prints them as if they were originally inserted as `a: string`. We're trying to identify if a type annotation was made as a comment and print it like so.

[Preview here](https://deploy-preview-3449--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAbOMAEAzCFMD0AVJqWZkuQM4wBOAllAOYDaAuuaUQZgLyYsAOiAAWcVKgjC2AbhAAaEBAAOMetCrJQAQ1q0IAdwAKuhJpTaAbhHoATBSG01k2baipxFAI1rawAawwAZWU-RiZkOgBXTxBbCDAXNw9FRg9aGCNfJgBbbST3WIArKgAPACFfAODtHLgAGUY4ApSlKJhldoAmFtjQ2nTkEGxJAwdlBlgAdTsYEWQADgAGRQmIDynfZSGJuHTLZsVaOABHKPpjrO1c-KRXQsUPHPpI2hjH8PQARSiIeF7FDBtF4ZrY5sguoDfPRUOEAMIQHJ5IZQaCHEBRDwAFWB5nuHgAvgSgA)


duailibe:
cc @vjeux

duailibe:
I'll update the PR soon-ish to support the ["comment type include"](https://flow.org/en/docs/types/comments/#toc-comment-type-include) syntax as well.

vjeux:
This is awesome! Can’t wait to see it ship :)

duailibe:
I pushed another commit to handle `/* *-/` comments inside the type annotation comments and have decided to handle `/*:: */` include comments in another PR :)

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -339,31 +339,24 @@ KEYPAD_NUMBERS.map((
 
 function f /* f */() {}
 function f(/* args */) {}
-function f() /* returns */ {
-}
-function f /* f */(/* args */) /* returns */ {
-}
+function f() /* returns */ {}
+function f /* f */(/* args */) /* returns */ {}
```
duailibe: This was necessary to align babylon and flow.. it was printing:

```js
// flow
function A() /*: T */ {}

// babylon
function A() /*: T */ {
}

duailibe:
This PR is ready for review now :-)

tests/comments/__snapshots__/jsfmt.spec.js.snap
vjeux: The fact that there was a newline here in the first place was just me being lazy :)

duailibe:
@vjeux I was just taking your feedback about formatting the comment. Can you take another quick look?

duailibe:
Thanks for the reviews!

Issue #3450, opened 2017-12-10
[RFC] Configuration for whitespace following `!`


GrahamCampbell:
How do people feel about adding configuration to control if `!foo` was kept as is, or fixed to `! foo`?

For example:

```js
if (!greeting) {
  //
}
```

being fixed to:

```js
if (! greeting) {
  //
}
```

I'm asking to see if such a PR already existed, or if there is a demand for it, so that one could be created, by myself, or someone else.

duailibe:
I don't think I've seen this style at all. Why do you need this kind of formatting? Is it common?

hawkrives:
I've seen this in a fair amount of PHP code but not in any JavaScript, myself.

lydell:
Regardless of the popularity, we have to close this for the same reasons as #1139.

GrahamCampbell:
That's a shame. For a tool like this to actually gain traction, it'll need to embrace configuration, otherwise there will just be 100 other tools with people that are pissed at this. This is why PHP-CS-Fixer is leading the way for PHP.

duailibe:
@GrahamCampbell I'm sorry but we don't actually believe we should have too many options. The point of Prettier is have a tool to format your code so you don't have to worry about that anymore. So ideally we wouldn't have any options and we'd always print the code the same.

But the community is already very used to styling in a code in some many different ways so we included options only to allow large number of people to use Prettier who wouldn't otherwise. I'm afraid this option doesn't fall into this bucket.

lydell:
We should write down our option philosophy in the docs, and perhaps even link to it in the issue template.

duailibe:
I'll send a PR

GrahamCampbell:
OK, makes sense. I think it will take a long time before the community will be aligned on standards. :/

JosephSilber:
@lydell unlike the parenthesis issue, which is just a style preference (I like the space, but it doesn't really matter. Can live without it), the space after the `!` is an important visual one.

Without a space, the `!` can easily be missed. This is the main thing that has been bothering me since we started using prettier: reading code became more difficult. We find that many times when reading code, your eyes simply gloss over the `!`, not realizing that the condition is being negated.

So while I agree that Prettier should avoid bike-shedding and not introduce merely-cosmetic options, I believe this is a more important one to consider.

lydell:
@JosephSilber I understand. However, in this case it could just as easily be argued that the space makes it less clear what is negated.

```js
if (! enabled && loggedIn) {
[...]

PR #3452, opened 2017-12-10
Clarify our philosophy around new options

duailibe:
So we can have something to link to in issues

Again, as a non-native speaker, let me know if something isn't clear or rephrase it entirely different, I don't mind :)

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -54,6 +54,19 @@ Here are a few examples of things that are out of scope for prettier:
 * Adding/removing `{}` and `return` where they are optional.
 * Turning `?:` into `if then else`.
 
+## Can prettier add an option to do X?
+
+A lot of the current options in prettier come from the `recast` options of which prettier is a fork, and some of them were added later due to popular demand. Ideally we would not want options in prettier at all to keep printing things consistently and have less maintenance overhead but the JavaScript community has already chosen a few different coding styles people are used to:
+
+* semicolons/no semicolons
+* single/double quotes
+* spaces between brackets/parens
+* more recently, parens in arrow function arguments
+
+Supporting all of these options are a maintenance burden because it's hard to keep track when they start conflicting with each other; and if you want to have a comprehensive test suite, it would grow exponentially with the number of options you support.
+
+For that reason we decided to only support new options that allowed a large number of people to adopt prettier that wouldn't otherwise. That means we'll have to say no to a lot of options that don't have a large demand for them. For everything else, we believe using ESLint (see [related projects](related-projects.md)) might fix most of the problems people encounter with prettier formatting.
+
 <!--
 ### Semi-colons
 


lydell:
Minor note: As far as I know, Prettier is spelled with an uppercase P.

Bigger note: I’d like to get something in about the “Use Prettier to avoid code style debates” thing I mentioned in: https://github.com/prettier/prettier/issues/3450#issuecomment-350578119

It might also be worth using “simpler” language, so that people actually read it. For important docs I usually aim for simplicity rather than 100% correctness. I should write a draft though, rather than giving you these unclear comments...

Also pinging @vjeux for your vision and communication skills :)

lydell:
Let’s also discuss whether we should add a link and short note to the issue template.

duailibe:
> Minor note: As far as I know, Prettier is spelled with an uppercase P.

I went with what was in the heading right above :)

Please do write a draft, I just meant to this started, let's rewrite it completely if necessary

docs/rationale.md
```diff
@@ -54,6 +54,19 @@ Here are a few examples of things that are out of scope for prettier:
 * Adding/removing `{}` and `return` where they are optional.
 * Turning `?:` into `if then else`.
 
+## Can prettier add an option to do X?
+
+A lot of the current options in prettier come from the `recast` options of which prettier is a fork, and some of them were added later due to popular demand. Ideally we would not want options in prettier at all to keep printing things consistently and have less maintenance overhead but the JavaScript community has already chosen a few different coding styles people are used to:
```
vjeux: > A lot of the current options in prettier come from the `recast` options of which prettier is a fork

This is not true. James Long didn't want any options and I wanted for prettier to work for Facebook but he didn't want to make the Facebook styleguide the default one, so we started adding a few options to make it work for Facebook.

docs/rationale.md
```diff
@@ -54,6 +54,19 @@ Here are a few examples of things that are out of scope for prettier:
...
+
+* semicolons/no semicolons
+* single/double quotes
+* spaces between brackets/parens
+* more recently, parens in arrow function arguments
+
+Supporting all of these options are a maintenance burden because it's hard to keep track when they start conflicting with each other; and if you want to have a comprehensive test suite, it would grow exponentially with the number of options you support.
```
vjeux: The reason eslint & co can get away with so many options is because their model is: when I see something bad, warn about it. Prettier on the other hand needs to be able to print every possible code you can throw at it using this option and make it look good. This is a --much-- harder problem.

We didn't even know if we could reasonably print JavaScript with no option at all, so it made sense to add as few options at the beginning.

Now that we know we can do it, it's still very challenging to add options and we're trying to only do so when a sizable amount of people are not willing to adopt prettier because of it. This way we can keep the project with a reasonable amount of complexity and when users want to use it, they are not overwhelmed by the number of options.

docs/rationale.md
duailibe: Ops I read someone saying that and thought it was true

duailibe:
Ok this requires a little more nuance and context on this project to be able to transmit better this idea than I do, I think someone else should do this and I'll help in any way I can :)

PR #3454, opened 2017-12-11
Fix PostCSS VS Code integration

ardevelop:
Fixes PostCSS integration for VS Code

Extension ([prettier-vscode](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)) is [querying](https://github.com/prettier/prettier-vscode/blob/master/src/utils.ts#L47) prettier for supported languages, and is not getting any information about `postcss`. As a result engineer sees that is no any formatter for `postcss`:

image

After exposing information that prettier supports `postcss` everything works fine.

### DIFF
--- a/src/support.js
+++ b/src/support.js
@@ -79,9 +79,9 @@ const supportTable = [
     aceMode: "css",
     codemirrorMode: "css",
     codemirrorMimeType: "text/css",
-    extensions: [".css"],
+    extensions: [".css", ".pcss", ".postcss"],
     liguistLanguageId: 50,
-    vscodeLanguageIds: ["css"]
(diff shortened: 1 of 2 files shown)


duailibe:
Hi @ardevelop, while I think I understand what's this supposed to fix, could you add a description of what exactly is the problem you are facing and why this fixes it?

Thanks for the PR! :)

ardevelop:
@duailibe thanks for feedback, updated description, PTANL

duailibe:
Thanks for updating @ardevelop!

@azz what do you think?

azz:
What does postcss as a language actually mean?

Issue #3321, opened 2017-11-26
SCSS: Preserve YAML front matter block


infotexture:
Prettier is smart enough to preserve YAML front matter blocks in Markdown files, but currently strips them from SCSS files, which breaks Jekyll sites such as those deployed via GitHub Pages.

Recent Jekyll versions provide native Sass support, but the Sass processing is only triggered if the file contains a valid YAML front matter block per [Jekyll docs](https://jekyllrb.com/docs/frontmatter/):

> Any file that contains a YAML front matter block will be processed by Jekyll as a special file. 

When Prettier removes the YAML header, site output appears unstyled, as the stylesheet is not parsed.

The same logic that allows Prettier to ignore YAML headers in Markdown files should be applied to SCSS.

azz:
Maybe not a bug in the strict sense, but we should fix it if we can.

lipis:
Maybe we should ask Jekyll to support/use valid comments per supported filetype for these headers instead of having something that is breaking the linters/editors :)

<img width="314" alt="screen shot 2017-11-27 at 15 55 00" src="https://user-images.githubusercontent.com/125676/33272742-5b229414-d38b-11e7-9127-de5af1fde8e3.png">

infotexture:
I don't think it's likely that Jekyll would be willing to change this, since so much of Jekyll's processing depends on the content of the YAML front matter, and every site on GitHub Pages depends on Jekyll, so the installed base that would be affected is huge.

Since Prettier already preserves YAML front matter blocks in Markdown files, it seems easier to enable similar processing in Prettier's SCSS parser than to hope for changes in Jekyll.

lipis:
My question was, why Jekyll is not supporting both..

lipis:
Will that make it to 1.9? #3333

duailibe:
If someone manages to get a PR in the next few weeks or so, sure.

infotexture:
**Until this is fixed, what's the best way to convince Prettier to ignore this single file?**

So far, I've tried the following entries in a `.prettierrc.yaml` file, but none of them seem to prevent Prettier from choking on the YAML header in the SCSS file:

_Skip the whole file:_
```yaml
overrides:
- excludeFiles: "styles.scss"
```

_Require pragma (since it doesn't have one, skip the file):_
```yaml
overrides:
- files: "styles.scss"
  options:
    requirePragma: true
```

_Exclude the range with the YAML header:_
```yaml
overrides:
- files: "styles.scss"
[...]

PR #3458, opened 2017-12-11
Docs: Add info on ignore file to "Ignoring Code"

infotexture:
While working on #3321, I needed to ignore an entire file, but the [Ignoring Code](https://prettier.io/docs/en/ignore.html) section of the docs only describes how to ignore blocks of code, so I wasn't aware of the default ignore file.

_(The `.prettierignore` file **is** mentioned in the [CLI](https://prettier.io/docs/en/cli.html) section, but easily overlooked there.)_

* Commit fccf98a extends the introduction sentence of the "Ignoring Code" section and adds a new subsection to explain how to ignore files.

    > Prettier offers an escape hatch to ignore a block of code <ins>or prevent entire files </ins>from being formatted.

    > ## Ignoring Files

    >To exclude files from formatting, add entries to a `.prettierignore` file in the project root or set the `--ignore-path` [CLI](https://prettier.io/docs/en/cli.html) option.

* Commit 420a12b fixes a nesting error in the heading hierarchy of the [CLI](https://prettier.io/docs/en/cli.html) section:

    The `--ignore-path` option currently appears as a third-level subsection of `--find-config-path` and `--config`, but this looks like an error. This commit promotes the `--ignore-path` heading to the same level as the other CLI options (H2).

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -44,7 +44,7 @@ You can also use `--config` if your configuration file lives somewhere where pre
 
 If you don't have a configuration file, or want to ignore it if it does exist, you can pass `--no-config` instead.
 
-### `--ignore-path`
+## `--ignore-path`
 
 Path to a file containing patterns that describe files to ignore. By default, prettier looks for `./.prettierignore`.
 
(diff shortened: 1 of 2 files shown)


duailibe:
Thanks!

PR #3471, opened 2017-12-13
Update tsep to v11 and update snapshot

JamesHenry:
Also has the bonus of fixing a bug by the looks of it!

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "semver": "5.4.1",
     "string-width": "2.1.1",
     "typescript": "2.6.2",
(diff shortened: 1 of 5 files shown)


duailibe:
@JamesHenry in that case we could remove `TSNamespaceFunctionDeclaration`? I'm not sure that's necessary

JamesHenry:
Yep, quite right, makes sense to remove it now. Updated!

duailibe:
Tagging this as part of 1.10 (even though it's mostly internal) because of the bug fix :)

PR #3473, opened 2017-12-13
Add Charlie Tango to users.json

thebuilder:


duailibe:
Thank you @thebuilder!

Can you replace the image for one with 100px height so the link remains the same height as the others?

thebuilder:
Sure! Added replaced with an image that contains white space to fill the height.

duailibe:
Thank you!

PR #3477, opened 2017-12-13
Add Webflow to users

iclanzan:
We are saving time at _Webflow_ thanks to _Prettier_!


duailibe:
Thanks @iclanzan!

Can you fill the image until its height is 100px? To be consistent with other logos

iclanzan:
@duailibe done!

duailibe:
Thanks! :heart:

PR #3478, opened 2017-12-13
Added Eventstag to users.json

ZeeCoder:
We're using Prettier at EventsTag for almost all of our projects now, and I don't see us going back. :D


duailibe:
Thanks!

PR #3479, opened 2017-12-13
Refactor imaged for Users, Editors and Languages

lipis:
Now that they are getting more and more let's have them in a separate folder.


lipis:
Do you guys agree with this change? Shall I fix the conflicts or close the PR?

PR #3484, opened 2017-12-14
docs: update babylon project url

nickarnold:
The babylon project URL has moved, as is noted by visiting the old URL.


azz:
Thanks!

Issue #3469, opened 2017-12-13
Can we enable the dangling comma on this project?


lipis:
Because even though one line was added, two were deleted and that's confusing when reviewing.

<img width="312" alt="screen shot 2017-12-13 at 11 39 30" src="https://user-images.githubusercontent.com/125676/33931987-5dfa3784-dffa-11e7-8193-6ba1e881638d.png">

Also if you would like to sort them, you would have to manually add the missing comma, because the there will be a syntax error if the last item will move in the middle.. (if someone will mention that these things are handled by Prettier 😃)

More reasons can be found here: https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

azz:
Prettier should probably use Prettier's defaults...

lipis:
That's what I thought.. but not really.. :) The default I guess comes based on popularity or what people wanted in a first place, but the world is evolving and maybe we can _teach_ them the _right_ way :P

Also if it's a hard requirement for this project to use the default values, let's change the default value :D Either case, having the dangling comma has more  pros than cons..

bakkot:
Please don't change the defaults without making a major version bump.

I don't see much reason for Prettier to always stick with the defaults, though, except those reasons that would apply to any other project (e.g. "use prettier's defaults" is the easiest rule which minimizes style bikeshedding).

lipis:
No I don't think that changing the default is the way to go.. Like I already mentioned.. there is no real need for Prettier to use the default.. we can adjust to what we want for the project.

duailibe:
Trying to make it clear: @lipis wants to change the Prettier settings we use for the Prettier codebase (i.e. only an internal thing)

lipis:
Yes! I (hopefully more) want to enable the following:

```yaml
trailingComma: all
```

azz:
We certainly can't set it to `all` because we support node v4. 

Prettier should be configuration free by default. Running Prettier on Prettier with Prettier's own defaults seems like it has more advantages. Don't want people to look at the Prettier source as an example and think that a config file is required.

josephfrazier:
> We certainly can't set it to `all` because we support node v4. 

For what it's worth, [node v4 will become EOL in April](https://github.com/nodejs/Release/tree/593a133b76fb8803b055716fd3aab7026a6cdf78#release-schedule), so I imagine this might not be a blocker at that point.

[...]

PR #3492, opened 2017-12-14
.eslintrc.js: Use `trailingComma: "es5"` with Prettier

josephfrazier:
This addresses https://github.com/prettier/prettier/issues/3469
as outlined in https://github.com/prettier/prettier/issues/3469#issuecomment-351758610

There are two commits at the moment, the first is the interesting change to
`.eslintrc.js`, and the second is the resulting formatting changes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,7 @@
 module.exports = {
   env: {
     es6: true,
-    node: true
+    node: true,
   },
   extends: ["eslint:recommended", "plugin:react/recommended"],
   plugins: ["prettier", "react", "import"],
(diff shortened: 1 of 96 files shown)


.eslintrc.js
```diff
@@ -22,15 +22,15 @@ module.exports = {
     "one-var": ["error", "never"],
     "prefer-arrow-callback": "error",
     "prefer-const": "error",
-    "prettier/prettier": "error",
+    "prettier/prettier": ["error", { trailingComma: "es5" }],
```
azz: Please don't put the setting here, it breaks Prettier editor integrations that rely on Prettier's config file.

josephfrazier: Ah, I didn't consider that. I had originally used `.prettierrc`, but the version of `eslint-plugin-prettier` we had didn't take it into account, so I had to upgrade that first. Fixed in https://github.com/prettier/prettier/pull/3492/commits/ed0b21faaa59db4ef162fa1e9341fe9644257cb7.

azz:
Let's discuss changing the default value in #3503

PR #3494, opened 2017-12-14
Update the command documentation and add types

j-f1:


commands.md
```diff
@@ -155,23 +155,59 @@ For an example, here's the implementation of the `ArrayExpression` node type:
...
+}
+
+declare function concat(docs: Doc<>[]): Doc<"concat">
+declare function indent(docs: Doc<>[]): Doc<"indent">
+declare function align(n: number, docs: Doc<>[]): Doc<"align">
+declare function group(docs: Doc<>[], opts?: {
+                        shouldBreak?: boolean,
+                        expandedStates?: Doc<>[]
+                      }): Doc<"group">
+declare function conditionalGroup(alternatives: Doc<>[], opts?: {
```
duailibe: `ident`, `align` and `group` receive only a doc, not array of docs

commands.md
```diff
@@ -155,23 +155,59 @@ For an example, here's the implementation of the `ArrayExpression` node type:
...
           path.map(print, "elements")
         )
       ])
-    ),
+    ]),
     line,
     "]"
   ])
-)
+])
```
duailibe: Per the other comment, this code was correct before

azz:
I worry that these types will become drift out of sync.

j-f1:
@azz I’ve moved the types to be inline with the docs. If anyone makes a breaking change, I’d assume they would update the docs too. Right now, it’s unclear what exactly each function takes as parameters.

azz:
Thanks for the docs!

PR #3496, opened 2017-12-15
Add "JSON with comments" to vscodeLanguageIds

thorn0:
Without this, the VS Code extension doesn't enable prettier for files of this type (settings.json, tsconfig.json, etc.).


thorn0:
done

azz:
Thanks!

felixfbecker:
Could we get a release for this? ❤️

PR #3502, opened 2017-12-15
Add Yelp to list of users

mxmul:
We use prettier in more than 100 repos at Yelp :)

screen shot 2017-12-15 at 12 43 21 pm


duailibe:
Thank you! ❤️

PR #3504, opened 2017-12-16
Add hasura to list of users

praveenweb:
Hasura is a platform to build and deploy app backends fast. We use it across all our frontend facing apps (mostly written in React.js)

prettier-hasura


azz:
Thanks :tada:

PR #3518, opened 2017-12-18
chore(package): upgrade to jest 22

SimenB:
The changes for you are just https://github.com/facebook/jest/pull/4460, and it impacts you because of https://github.com/prettier/prettier/blob/d1c97b362219d62669145bb6bdd66a4be5423257/tests_config/run_spec.js#L50

EDIT: rebasing


azz:
Looks like you might have done this on an old branch, (maybe wrong `<remote>/master`?) lots of merge conflicts.

SimenB:
Yup, have rebased and rerunning as we speak

ikatyang:
Is it safe to update to Jest 22? I saw `Good bye Node 4` on the post, or it's only for JSDOM, not the Jest repo?

azz:
We'll see if the build passes :P

SimenB:
Oh, you support node 4? In that case you might not want this actually.

The locked down versions you have make it safe, unless you use JSDOM to test, but might be better keep to the old versions

SimenB:
Fails because of `engines`. Can use `YARN_IGNORE_ENGINES=true` if you want, but not sure if it's worth it?

azz:
Better not. We might be stuck on the current version of jest for a while. Thanks anyway.

azz:
Just thinking - if the reason v4 support was dropped from jest is for jsdom, is there any way to guarantee v4 support if we're not using jsdom?

duailibe:
I think we can enable `YARN_IGNORE_ENGINES` on travis, that would still make the tests run with the node v4 and if they pass we're good.

azz:
Won't that env flag ignore all packages, not just jest?

SimenB:
It will, yes. But your test coverage is good enough, isn't it?

duailibe:
I believe if the tests pass we should be good, right?

azz:
_Should_ be. There may be edge cases we're not considering that are covered by the tests of our dependencies that don't pass on Node 4.

PR #3519, opened 2017-12-18
Add `root: true` to .eslintrc.yml

j-f1:
This ensures that peoples’ ESLint configurations in parent directories aren’t used when linting Prettier.


azz:
Thanks!

PR #3521, opened 2017-12-18
PHP formatting support

mgrip:
Inspiration taken from #3349 - this time for php! There's still a lot of work to be done here, but wanted to get eyes on it to make sure we're headed in the right direction. We're using https://github.com/glayzzle/php-parser for a parser - luckily its written in JS so no need for external dependencies.


TrySound:
Adding another one dependency will make prettier even larger which is not good. Look at python implementation https://github.com/prettier/prettier/pull/3349 without deps.

package-lock.json
```diff
@@ -0,0 +1,7827 @@
+{
```
lipis: Remove this file.. since we are using `yarn.lock`

mgrip: whoops sorry about that - all set!

azz:
Hi! Thanks for working on this. I think we should finish HTML printing before tackling PHP, because PHP has to be embedded in a HTML document.

azz:
Unless we just support PHP files without any HTML?

vjeux:
This is really exciting! I just want to make sure that you are aware that writing a pretty printer for a full language (and one as crazy as PHP) is going to be a big undertaking. If you are onboard, here are the three overall steps:

1) Keep doing what you're doing and print all the nodes.
2) Run --debug-check through a bunch of big codebases and make sure that you are every piece of code you can throw at it correctly.
3) Make sure the long tail of code is printed in a way that looks good.

Step 2 is absolutely critical. If your formatter outputs invalid code, people are not going to trust it, no matter how pretty the code is.

hawkrives:
@mgrip For step 2, you can let me know when you think you have something that should work – I can run it on a big-ish internal Wordpress theme / app.

mgrip:
thanks for the info all! so i cant make any promises in terms of timeline, but definitely interested in pushing this forward. I work with a large number of engineers using php daily, so im sure I could wrangle some help there as well (and we have a very large php codebase to run it against for testing, in addition to some open source ones like wordpress or drupal). mostly wanted to post this initially to make sure we’re headed in the right direction. in terms of next steps, that makes sense - thanks @vjeux. what are your guys thoughts on the parser we’re using? that could be expensive to pivot on, so just want to make sure thats set in stone before continuing on with this.

src/printer-php.js
```diff
@@ -0,0 +1,269 @@
...
+  const n = path.getValue();
+  if (!n) {
+    return "";
+  } else if (typeof n === "string") {
+    return n;
+  }
+  return handleNode(n);
+}
+
+function handleLiteral(node) {
```
j-f1: Should these be renamed `print*` to be similar to other printers’ methods?

mgrip: makes sense - updated!

azz:
👋 @mgrip, FYI: https://github.com/prettier/prettier/pull/3349#issuecomment-353311805.

mgrip:
makes sense, sounds much more scalable. so the intent would be to move this PR to a new prettier/prettier-php repo with a dependency on prettier?

src/printer-php.js
```diff
@@ -9,6 +9,16 @@ const indent = docBuilders.indent;
...
 const softline = docBuilders.softline;
 
+// polyfill for node 4
+function includes(array, val) {
+  for (let i = 0, len = array.length; i < len; i++) {
+    if (array[i] === val) {
+      return true;
+    }
+  }
+  return false;
```
j-f1: You can just do `array.indexOf(val) !== -1` here.

mgrip:
anyone know why the ci job fails w/ `NODE_ENV=production`? seems if dependencies are added in the PR they don't actually get installed?

borNfreee:
Sorry for such a question, but why have you decided to support PHP when there is already a known leader of PHP Code Style fixes: [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) ?

azz:
Looks like that's a fixer (rule based edits), not a full formatter.

azz:
I have copied all the commits from this pull request to a new repository using the in-progress plugin API. https://github.com/prettier/prettier-php 

@mgrip I have added you as an admin of the repository.

PR #3526, opened 2017-12-19
Sort parsers and more

lipis:
With this rate that the parsers are being added, I think it makes sense to have them in alphabetical order..


tests_integration/__tests__/__snapshots__/early-exit.js.snap
```diff
@@ -265,21 +265,21 @@ exports[`show detailed usage with --help no-semi (write) 1`] = `Array []`;
 exports[`show detailed usage with --help parser (stderr) 1`] = `""`;
 
 exports[`show detailed usage with --help parser (stdout) 1`] = `
-"--parser <flow|babylon|typescript|css|less|scss|json|graphql|markdown>
+"--parser <babylon|css|flow|graphql|json|less|markdown|scss|typescript>
```
azz: There was a logical grouping here that has broken

lipis: I can see that there are more errors and there are some trailing spaces that were removed..  do we really need those?

scripts/build/build.js
```diff
@@ -11,11 +11,11 @@ const rootDir = path.join(__dirname, "..", "..");
 const parsers = [
   "babylon",
   "flow",
-  "typescript",
   "graphql",
```
azz: Leave the one as is,  otherwise there will be conflicts in #3536

lipis:
Maybe it doesn't make sense if there is a huge refactor is about to happen

PR #1951, opened 2017-06-04
[WIP] Initial Handlebars/Glimmer support

azz:
Experimenting with HTML/Handlebars/Mustache/Glimmer support for #1882

Picked the glimmer parser as I thought it would have the widest range of support.

I've ran into a few issues so far:

* Some handlebars features (`{{> foo}}`, `{{../foo}}`) are not supported. [source](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/syntax/lib/parser/handlebars-node-visitors.ts#L158)
* `<!doctype>` doesn't parse.

The formatting is also super ugly at the moment because I figured the best way to go would be re-use some of the JSX formatting logic.

I also moved the postcss printer to a separate file.

Issue #1882, opened 2017-06-02
Supporting HTML/Mustache


vjeux:
Angular, Vue and Ember all use variants of html/mustache with JS and CSS embedded. It would be really awesome to make prettier support it as well. This would allow the project to be the one stop destination for all the front-end related code.

josephfrazier:
(related to https://github.com/prettier/prettier/issues/1674)

octref:
Really wish http://codeguide.co/#html-syntax by @mdo could be enforced, so people can write consistent html:

- Use soft tabs with two spaces—they're the only way to guarantee code renders the same in any environment.
- Nested elements should be indented once (two spaces).
- Always use double quotes, never single quotes, on attributes.
- Don't include a trailing slash in self-closing elements—the HTML5 spec says they're optional.
- Don’t omit optional closing tags (e.g. `</li>` or `</body>`).

Also, this feels like perfect use case for prettier 😉 :

```html
<foo
  class="i have so many classes"
  data-really-long-attr="bar"
  data-another-long-attr="bar"
></foo>
```

I don't know specifics for formatting ember/angular, but for Vue SFC, my Vue/VSCode plugin https://github.com/octref/vetur has code that takes each region and send them to respective formatters:

```vue
<template>
  <!-- send to html formatter -->
</template>
<script lang="ts">
  // send to ts formatter
</script>
<style lang="scss">
  // send to scss formatter
</style>
```

Previously I hooked up js-beautify as formatter for `script` and `style` regions, but now prettier has TS support so I'm switching over to it for formatting js/ts/css/scss/less.

Once html supports land in prettier, I can make a Vue formatter that takes any Vue file written with html, js/ts, css/less/scss/postcss and outputs consistent code. That is 🔥 🔥 🔥

donaldpipowitch:
Will this cover underscore/lodash templates, too? ♥

maraisr:
+1

thedamon:
This would be amazing. Other use case is working in not specifically js related environments such as CMSs that use html styled templates. A tool available to the command line to verify and correct HTML errors could be incredibly useful for processing user input in WYSIWYG HTML fields in CMS fields as well (Unfortunately these still do exist...).

One thing that could help a lot would be a list of acceptable (custom) attributes, though that may beyond the scope of prettier and more appropriate for a linter.

gabrielmaldi:
> This would allow the project to be the one stop destination for all the front-end related code.

👏👏👏

[...]

PR #3534, opened 2017-12-19
Support handlebars

robert-j-webb:
This work is me picking off where @azz left. This is his pr and my starting point: [here.](https://github.com/prettier/prettier/pull/1951)

#1882 is the issue related to this.

Feedback is extremely welcome. I'm doing my best and I have a good deal of hours to dedicate to this.


robert-j-webb:
Hi! This is ready for review. I'd be happy to fix any issues that there are with it. Thank you!

@azz

robert-j-webb:
I've been debugging this for a couple of hours and I'm getting a couple of errors. First of all, 
``` Dynamic requires are not currently supported by rollup-plugin-commonjs```

Secondly, ```... is not a token.``` This only happens when the build is node 4. I'm going to try changing the glimmer version to see if I can get something to work.

**Comment by azz** [comment]
Take a look at `rollup.parser.config.js` to see how to fix dynamic requires.

`...` isn't supported in Node.js v4. It might be used in Glimmer so you'll need to run it through babel. See `build-docs.js` for an example of that.

File: src/cli-constant.js [hunk 1]
```diff
@@ -231,6 +231,7 @@ const detailedOptions = normalizeDetailedOptions({
       "less",
       "scss",
       "json",
+      "glimmer",
```
azz: Just comment this line out and update snapshots for now. When we release it we'll uncomment.

tests/html_glimmer/element-modifier-statement.hbs
```diff
@@ -0,0 +1,15 @@
...
+  Hello
+</div>
+
+<div {{hello hashPair=value hashPair=value hashPair=value hashPair=value hashPair=value}}>
+  Hello
+</div>
+
+<div {{hello param param param param hashPair=value hashPair=value hashPair=value hashPair=value hashPair=value}}>
+  Hello
+</div>
```
azz: Newline here

tests/html_glimmer/element-node.hbs
```diff
@@ -0,0 +1,36 @@
...
+    {{hello}}
+  {{/block}}
+</div>
+
+<div>
+  {{hello}}
+</div>
+
+<div></div>
+<img />
```
azz: Newline here

tests/html_glimmer/else-if.hbs
```diff
@@ -0,0 +1,53 @@
...
+  <div>
+    {{#if a}}
+      b
+    {{else if c}}
+      d
+    {{else}}
+      e
+    {{/if}}
+  </div>
+</div>
```
azz: Newline here

robert-j-webb:
I added new lines to the test files and commented out the line you referenced. @azz

Turbo87:
@robert-j-webb how did you solve the whitespace handling issues that are inherited from HTML?

robert-j-webb:
I trim away all white space. At some point, I want to preserve the users’ extra line breaks between elements. However, I scoped this pr to be a much more manageable pretty printer by dealing with that problem later, preferably with lots of community feedback.

azz:
There was a massive refactor in #3536 that caused some conflicts here. Would you be able to solve them?

Go with `language-handlebars/parser-glimmer.js`

robert-j-webb:
Yeah I'll start working on it! I love the plugin approach.

robert-j-webb:
Ok I got the rebase to work @azz. I think so at least! It's really good that there's a solid plugin system now.

tests/html_handlebars/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,131 @@
...
+<h1>
+  {{#if isAtWork}}
+    Ship that code!
+  {{else if isReading}}
+    You can finish War and Peace eventually...
+  {{else}}
+    Go to bed!
+  {{/if}}
+</h1>
+<h2>
```
vjeux: We'll want to preserve the empty line above.

tests/html_handlebars/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,131 @@
...
+      {{body}}
+    </div>
+  </div>
+</script>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<script id="entry-template" type="text/x-handlebars-template">
+  <div class="entry">
+    <h1>
+      {{title}}
+    </h1>
```
vjeux: It seems overkill to always break

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,493 @@
...
+  Hello
+</div>
+<div
+  class="hello
+    {{if goodbye true}}
+    {{if goodbye false}}
+    {{if goodbye true}}
+    {{if goodbye false}}
+    {{if goodbye true}}"
+>
```
vjeux: Does it respect the jsx-bracket-same-line option?

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,493 @@
...
+  hi
+</div>
+<div>
+  A long enough string to trigger a line break that would prevent wrapping.
+</div>
+<div>
+  A long enough string to trigger a line break that would prevent wrapping more.
+</div>
+<div>
+  A long enough string to trigger a line break that would prevent wrapping more and more.
```
vjeux: You're not wrapping here?

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,493 @@
...
+</div>
+<div>
+  {{#block}}{{hello}}{{/block}}
```
vjeux: I don't understand the heuristics you chose. Sometimes it is in one line, sometimes it breaks, for things that look similar.

src/cli/constant.js
```diff
@@ -234,6 +234,7 @@ const detailedOptions = normalizeDetailedOptions({
       "less",
       "scss",
       "json",
+      // "glimmer",
```
vjeux: I would name all those "handlebars". Except for JavaScript, we started naming the parser after the language it parses and not the actual name of the parser. This is easier for people to figure out which to use. It would be nice to rename it to --language at some point but we'll see.

tests/html_basics/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -52,3 +52,55 @@ exports[`html-fragment.html 1`] = `
...
+
+<body>
+  <a href="#">Anchor</a>
+  <div hidden
+    class="foo"
+    id="bar" />
+
+</body>
+</html>
+`;
```
vjeux: The html parser/output seems to be completely broken :p

src/cli/constant.js
robert-j-webb: I agree that renaming from glimmer to handlebars would make sense. It's just that at this moment, glimmer doesn't support partial statements, which handlebars does support. So it's kind of a specific caveat that they're different.

vjeux: Do you know if there's a list of the differences between the two?

robert-j-webb: I'm not sure, I'll ask the maintainers of the project.

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
robert-j-webb: This is one the tricky ones to decide. You wouldn't break this string up to multiple lines:
```js
const hello = 'A string that is over 80 chars long';
```
I could take text and break on space, but I think that's against what I would expect from prettier.

tests/html_basics/__snapshots__/jsfmt.spec.js.snap
robert-j-webb: Yeah it's not great. I could fix (it's very similar to handlebars) it but I wanted to scope my PR around just getting handlebars working because it's already so much code.

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
robert-j-webb: No it doesn't. I think a ternary in a few places would fix that though. I didn't write any support for options! :(

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
vjeux: I would recommend looking at what prettier outputs for jsx and try to be as close as possible. This has been battle tested over time and it hasn’t been a big source of issues.

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
vjeux: It would be great to support it :)

Adding a new language to prettier works in two phases. The first one is to print everything in a correct way, which you mostly did here. The second one is to make all the things than can be thrown at it look good.

You’re now entering the second phase :) So expect to do a lot of tweaks everywhere.

tests/html_basics/__snapshots__/jsfmt.spec.js.snap
vjeux: Yeah, don’t worry about the html printer

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
robert-j-webb: I would prefer that HTML tags would break as frequently as block elements, but I remember running into some situations where that would cause some very ugly formatting. 

Perhaps block elements should also always break? What would be best?

vjeux: Here's a small example for JSX: [Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4D4A6UABEcAIYC+hqA9FngNzX27kW0tNQY5FmuUOOXFx7Y+RAEYD2dYaJZtgk2Z0IgANCAgAHGJmgBnZKDIAnMxADuABXMJjKMtgiZ0mkGUMxkAMzIANoZwWpJmZGAA1nAwAMo6EZhQAObIMGYAriEg6BBgfoHBWknBZjA24ckAtmQFQdkAVoYAHgBC4VExsWRVcAAySXB1RdoZMDpjAEzD2QlmpcggkmSSAJ4B0B46ZkkwAOpuMAAWyAAcAAxa2xDBe+E6i9twpdhDWmZwAI4ZmB8VZNVakh-PUtMEqpg0plsoYkskAnAAIoZCDwGZaGArA7oY7ISYY8KYAJwgDCECqNUWUGgbxAGWCABUVo4QcEKBQgA)

**Input:**
```jsx
<div>
  {a}
</div>;
<div>{a}</div>;
<div a>{a}</div>;
<div a b>{a}</div>;
<div>{a}{b}</div>;
```

**Output:**
```jsx
<div>{a}</div>;
<div>{a}</div>;
<div a>{a}</div>;
<div a b>
  {a}
</div>;
<div>
  {a}
  {b}
</div>;
```

It always break if there's >2 attributes or >2 values and inlines otherwise.

This is a small example but you should look at the JSX source code and see all the heuristics it does to make it look good.

vjeux:
I just want to be explicit about what the next steps are.

Prettier has been successful because it doesn’t just prints code but because it prints it extremely well. In order to achieve this level of quality, we need to add a bunch of complexity in the printer. The good news is that we have already done a lot of similar work for JSX so you should steal as much as possible. This way it has already been battle tested and we know it works well, and if people switch between React and Ember projects using prettier, they will already be familiar with the way code is formatted.

In order to make sure that things are well formatted, we want to run it in as many codebases as possible and go through the diffs and try to make all the outputs look as good as possible as well as running —debug-check to make sure that we’re printing correct code.

Once it’s in a really good shape, then we can release it to everyone. At this point it’s likely going to get a lot of people to try it out and give good feedback that needs to be acted on. After a while it’ll stabilize and it’ll just work and save people a bunch of time :)

I hope that it makes sense. Let me know if you have any questions.

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
robert-j-webb: I expected this! I'm really happy to do this part too. I figured that it would need tweaking because a lot of the decisions I made were based off asking a few people in the community and then picking from what is available. I will be in airplane for a few hours next week so I'll have plenty of time to implement whatever is decided upon.

robert-j-webb:
I absolutely understand! I'll look more into how JSX is doing things and try to mirror it. I don't know if it will be possible to reuse the JSX code itself, but I'll look into that. I'm excited to make this work extremely well!

vjeux:
Awesome! It’s going to be so good :)

For reusing code, I wouldn’t spend too much time, if it’s easy then do it but otherwise it’s not a big deal. It’s better for the behavior to be similar.

src/cli/constant.js
mmun: Handlebars is a plain text templating language. In practice it is often used as an HTML templating language, but it is also commonly used for templating configuration files (toml, ini, yaml, etc).

Glimmer is an HTML templating language, but there are a few differences from vanilla HTML:
- The parser will hard error if non-void tags aren't balanced. I don't think we plan to change that any time soon.
- There are a few extensions we've added to HTML that let us differentiate HTML attributes from component arguments:

```hbs
<MyComponent class="red" @data={{myData}}>
```

You can also pass blocks as arguments (I believe they are commonly called render props in React)

```hbs
<MyComponent>
  <@header=>This is a header!</@header>
  <@header=>This is a footer!</@header>
</MyComponent>
```

Turbo87: @mmun isn't it technically "HTMLBars" instead of "Glimmer" with "Glimmer" just implementing a parser for "HTMLBars"?

mmun: It's true that that is [how we launched it](https://emberjs.com/blog/2015/02/07/ember-1-10-0-released.html) but it was always for lack of a better name. You won't find references to HTMLBars in the guides/API docs anymore. The Glimmer brand now encompasses a DOM templating language, a runtime VM, and an experimental application framework (http://glimmerjs.com)

mmun: @vjeux Let me start by saying thank you very much for working on prettier and for reviewing this PR! And thanks to @robert-j-webb for working on it!

I completely agree with you that we should minimize confusion with the configuration. Taking a step back away from Glimmer for a second, what configuration would you propose to differentiate between `.txt.hbs`, `.html.hbs`, `.toml.hbs`, etc? Each variant would need different formatting for the non-Handlebars parts. It's also plausible to me that the stuff inside the curlies (`{{ ... }}`) might be slightly formatted differently depending on the variant as well.

duailibe:
@robert-j-webb A while ago @karl tried to come up with a intermediate representation that JSX, Handlebars and others would translate to and then apply the rules to that representation, if you're interested

src/cli/constant.js
vjeux: I’m not super familiar with handlebars but the philosophy we followed with prettier so far is to take use cases one by one and make them really good. So my recommendation would be to make it work perfectly for Ember and then we can figure out how to handle more use cases. My guy feeling is that there’s a big market for people eager to get prettier to work for the html variant and low interest in the other variants right now.

chriskrycho:
Quick question as a _very_ interested consumer: is the easiest way to try this out just installing `master`? I figure it may be useful to give feedback from our reasonably large codebase (3650 LoC of `.hbs` per [cloc](http://cloc.sourceforge.net/)).

ikatyang:
@chriskrycho 

```sh
yarn add prettier/prettier
yarn prettier **/*.hbs --parser glimmer
```

SimenB:
I tried running this at work (it's `mustache` and not `handlebars`, but close enough!), and the first issue I hit (as it failed on like 75% of all our templates) is that it doesn't support the old `else` (or `not`, no idea what it's called) syntax.

```sh-session
$ echo '{{^something}}{{/something}}' | node_modules/.bin/prettier --parser glimmer
[error] stdin: TypeError: Cannot read property 'blockParams' of undefined
[error]     at TokenizerEventHandlers.Program (/Users/simbekkh/repos/mfinn/node_modules/@glimmer/syntax/dist/commonjs/es5/lib/parser/handlebars-node-visitors.js:67:61)
[error]     at TokenizerEventHandlers.BlockStatement (/Users/simbekkh/repos/mfinn/node_modules/@glimmer/syntax/dist/commonjs/es5/lib/parser/handlebars-node-visitors.js:100:28)
[error]     at TokenizerEventHandlers.acceptNode (/Users/simbekkh/repos/mfinn/node_modules/@glimmer/syntax/dist/commonjs/es5/lib/parser.js:64:31)
[error]     at TokenizerEventHandlers.Program (/Users/simbekkh/repos/mfinn/node_modules/@glimmer/syntax/dist/commonjs/es5/lib/parser/handlebars-node-visitors.js:75:18)
[error]     at TokenizerEventHandlers.acceptNode (/Users/simbekkh/repos/mfinn/node_modules/@glimmer/syntax/dist/commonjs/es5/lib/parser.js:64:31)
[error]     at preprocess (/Users/simbekkh/repos/mfinn/node_modules/@glimmer/syntax/dist/commonjs/es5/lib/parser/tokenizer-event-handlers.js:336:61)
[error]     at Object.parse (/Users/simbekkh/repos/mfinn/node_modules/prettier/src/language-handlebars/parser-glimmer.js:27:12)
[error]     at Object.parse (/Users/simbekkh/repos/mfinn/node_modules/prettier/src/main/parser.js:61:19)
[error]     at formatWithCursor (/Users/simbekkh/repos/mfinn/node_modules/prettier/index.js:100:22)
[error]     at Object.formatWithCursor (/Users/simbekkh/repos/mfinn/node_modules/prettier/index.js:389:12)
[error]     at format (/Users/simbekkh/repos/mfinn/node_modules/prettier/src/cli/util.js:154:19)
[error]     at thirdParty.getStream.then.input (/Users/simbekkh/repos/mfinn/node_modules/prettier/src/cli/util.js:258:19)
[error]     at <anonymous>
[error]     at process._tickCallback (internal/process/next_tick.js:188:7)
```

This works on http://tryhandlebarsjs.com/.

image

SimenB:
It also doesn't support partials, which is sad :(

Another thing is that the column in the syntax errors are `0`, and should probably be ignored in the code frames.

image

Yet another issue is that it says that `Error: A block may only be used inside an HTML element or another block.` for valid stuff, like:

```mustache
<input {{#checked}}checked{{/checked}}>
```

This also works just fine in http://tryhandlebarsjs.com/

That's all I've got for now. Not sure which, if any, are fixable.

mmun:
@SimenB Thanks for the report! Yeah, we don't intend to support cases like `<input {{#checked}}checked{{/checked}}>` in Glimmer. Same for partials, decorators, etc. They don't make sense in Glimmer's component model.

PR #3541, opened 2017-12-20
add js version pre-commit

gengjiawen:
I find it very convenient and very easy to config.

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -31,7 +31,9 @@ and add this config to your `package.json`:
 See https://github.com/okonet/lint-staged#configuration for more details about how you can configure lint-staged.
 
 ## Option 2. [pre-commit](https://github.com/observing/pre-commit) (JS version)
+
 Install the package:
+
 ```bash
 yarn add pre-commit --dev
 ```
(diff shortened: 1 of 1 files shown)


docs/precommit.md
```diff
@@ -30,7 +30,27 @@ and add this config to your `package.json`:
...
+
+```json
+{
+  "scripts": {
+    "prettier": "prettier */**/*.js --ignore-path ./.prettierignore --write && git add . && git status"
+   },
+   "pre-commit": [
+    "prettier",
+  ]
+}
```
j-f1: Can you put a closing ` ``` ` here?

gengjiawen: silly me, forget that :)

gengjiawen:
Have not got a clue how to fix the ci ...

duailibe:
Thank you!

re CI: prettier was changing the contents of the `package.json` file, but your version was better so I just added `prettier-ignore` there :-)

azz:
Is this a good idea? This will not respect the files staged in git and commit everything. Better to recommend lint-staged approach.

gengjiawen:
yes, maybe should use something like pipe to redirect to the file list to git add.

gengjiawen:
I find someone use `lint-stage` with `pre-commit`, like https://github.com/apollographql/apollo-link/blob/bda4e08ae3baf53a4b90987e9b43766474e7ad78/package.json#L86. Maybe this is a better way.

PR #3553, opened 2017-12-22
Changed the alt attribute for monthly downloads

ntmonson:
Updated the alt attribute to reflect the image it is attached to.


azz:
Good catch, thanks!

Issue #3558, opened 2017-12-23
Don't read .editorconfig outside the project root


lydell:
See https://github.com/prettier/prettier/issues/3461#issuecomment-353732090

Want to help? Make a small reproduction case!

josephfrazier:
It looks like `editorconfig` supports a `root` option, so hopefully it's simply a matter of passing in the right path: https://github.com/editorconfig/editorconfig-core-js/tree/0c623dd8b5a8fc890c6e19fb0a17f7e983a1e015#parsefilepath-options

azz:
#3559

OliverJAsh:
I came here wanting the opposite behaviour. My editor respects my global EditorConfig (`~/.editorconfig`) and sets my tab width to that setting, but when I use my editor to run formatting (i.e. Prettier), it uses its own default tab width setting.

OliverJAsh:
@lydell That's one usage of it. However for projects where there isn't an `.editorconfig` defined (yet), e.g. new projects or temporary workspaces, it's useful to have a global config so you don't have to spend time setting things up. Of course as a project matures, it's good to share the config so everyone uses the same style.

lydell:
I usually use the default Prettier options for temporary workspaces. They’re temporary anyway.

PR #3559, opened 2017-12-23
editorconfig: Only search for .editorconfig up to the VCS directory

josephfrazier:
This addresses https://github.com/prettier/prettier/issues/3558

I'm not sure if this is the best way to find the "project root", but it
seems better than before.


rattrayalex:
Would this impact the behavior of something like 
`./prettier --write '../../some/where/else/**/*.js'`?

What about monorepo/lerna-style projects, which might have something like this: 

```
.editorconfig # establishes baseline
packages/
   foo/
     package.json
     .prettierrc # overrides something specific to foo
   bar/
     package.json
     .prettierrc
```
I actually don't know prettier's current behavior here, or what's desired.

azz:
I think finding the VCS directory might work? 

Something like https://npmjs.com/package/find-project-root? Not sure...

josephfrazier:
Yeah @azz, I think finding the VCS directory would work far better, and handle the cases that @rattrayalex pointed out. One potential issue is that projects that don't use source control wouldn't be able to take advantage of the `.editorconfig` support, but I imagine that use case is vanishingly small, and being able to format files (taking `.editorconfig` into account) outside the current directory is far more important.

azz:
Actually, any way we could add a test for this?

rattrayalex:
Do the docs for `.editorconfig` claim/imply that it must exist at a VCS root? (I also assume that it typically _would_). 

I'm not sure we need a test if we're using a third-party-package that has tests. _looks at npm package_ ah, it doesn't even have a repo listed.

azz:
Here's the tests: https://github.com/kirstein/find-project-root/blob/master/test/index.spec.js

I'd like to have one integration test for this though to document the expected behaviour.

josephfrazier:
I added a quick test in https://github.com/prettier/prettier/pull/3559/commits/d9b2447818965c3b146d52d47c5f0e0ff07bd440.

I also noticed that the `.prettierrc` search also doesn't stop at the VCS directory, currently. We should probably fix that too.

josephfrazier:
Good point @azz, thanks for merging.

Issue #2097, opened 2017-06-11
Vue Support


vjeux:
Let's start a master issue for all the related vue.js progress.

List of projects using Vue (based on tweet https://twitter.com/Vjeux/status/873726664630337536 ):

- https://github.com/gitlabhq/gitlabhq
- https://github.com/bootstrap-vue/bootstrap-vue
- https://github.com/ElemeFE/element/
- https://github.com/algolia/vue-instantsearch
- https://github.com/vuetifyjs/vuetify
- https://github.com/samiheikki/javascript-guessing-game
- https://github.com/phanan/koel
- https://bitbucket.org/sellvana/core/src/4231c995aff9785ca60a550b0ee06c286f4bf619?at=sn1
- https://github.com/search?utf8=%E2%9C%93&q=language%3AVue+stars%3A%3E1000&type=Repositories&ref=advsearch&l=Vue&l=

AustinPaquette:
Just want to add this to the list 😁 https://github.com/vuetifyjs/vuetify

vjeux:
Looks like we'll need to run some js expansion inside of `:class` value:

```vue
<template>
  <h2
    class="title"
    :class="{
      'issue-realtime-pre-pulse': preAnimation,
      'issue-realtime-trigger-pulse': pulseAnimation
    }"
    v-html="titleHtml"
  >
  </h2>
</template>
```

https://github.com/gitlabhq/gitlabhq/blob/master/app/assets/javascripts/issue_show/components/title.vue#L46-L49

vjeux:
Template supports a lang attribute. https://github.com/vuetifyjs/vuetify/blob/master/src/components/carousel/Carousel.vue

```js
<template lang="pug">
  div(class="carousel")
    div(class="carousel__left")
      v-btn(icon v-on:click.native.stop="prev")
        v-icon chevron_left
...
```

azz:
> Looks like we'll need to run some js expansion inside of :class value

Except that's not valid JS. The `{}` in this case would be parsed as a block, not an object. We'd need to wrap it with `()` before parsing.

samtgarson:
Didn't want to open a new issue, but feel free to tell me to shove off if this isn't the place...

I've got the `parse5` parser up and running on my large (unfortunately not open source) Vue codebase, and it seems to work well for `script` tags in `.vue` files, but for `.js` files and the `template` part of my `.vue` files it just sticks everything on one long line... 

[...]

PR #3563, opened 2017-12-24
[vue] Introduce proper support for Vue Single File Components (SFC)

vjeux:
There's a lot of demand for vue sfc (#2097). This introduces partial support for them: all the html is printed as is, except for the script and style tags which are printed using prettier. I believe that this should cover a lot of the use cases while being simple to support and if we want we can extend to more in the future.

I copy pasted the html parser used by vue (it's just a single 400 lines file) so that we don't run the chancesof conflicts. I'm also very conservative: I only print the style and script at the top level and for the lang attributes we support.

I expect this to be landable as is and provide value, review welcome :)

### DIFF
--- a/tests/multiparser_vue/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/multiparser_vue/__snapshots__/jsfmt.spec.js.snap
@@ -9,18 +9,16 @@ exports[`template-bind.vue 1`] = `
 </template>
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-
 <template>
 
-  <div v-bind:id="'list-' + id" />
-
-  <div v-bind:id="rawId | formatId" />
-
-  <div v-bind:id="ok ? 'YES' : 'NO'" />
-
-  <button @click="foo(arg, 'string')" />
-
+    <div v-bind:id=" 'list-'   +  id ">
+</div>
+    <div v-bind:id="  rawId | formatId ">
+</div>
+    <div v-bind:id=" ok ? 'YES' : 'NO' ">
+</div>
+    <button @click=" foo ( arg, 'string' ) ">
+</button>
 </template>
 `;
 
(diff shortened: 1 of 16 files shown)


j-f1:
Don’t forget [the README](https://github.com/vjeux/prettier/blob/vue_sfc/README.md#opinionated-code-formatter) and [the website](https://github.com/vjeux/prettier/blob/vue_sfc/website/data/languages.yml) :)

Haroenv:
Cc @octref who did a lot of work on Vetur, the vs code plugin which already uses prettier for css and js: https://github.com/vuejs/vetur

tests/multiparser_vue/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -68,28 +54,25 @@ p { font-size : 2em ; text-align : center ; }
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-
-<template>
-
-  <h1>{{greeting}} world</h1>
-</template>
+<template >
+  <h1 >{{greeting}}     world</h1 >
+  <script>kikoo ( ) </script>
```
suchipi: Why not format non-toplevel scripts? What's the concern?

vjeux: I don’t want to add a lot of heuristics to figure out the correct indentation level

vjeux:
Oops, looks like I broke the build! Sorry about that. It’ll be a day before I can fix it. Feel free to revert in the meantime.

tests/multiparser_vue/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -35,24 +33,23 @@ exports[`template-class.vue 1`] = `
...
-
-
 <template>
 
+  <h2
+    class="title"
+    :class="{ 'issue-realtime-pre-pulse': preAnimation,
+ 'issue-realtime-trigger-pulse': pulseAnimation}"
+    v-html="titleHtml"
+  >
```
octref: I wouldn't say this is the preferred way to write Vue interpolations.

octref:
@vjeux Thanks a lot for your hard work. It would be great to see prettier officially supporting Vue!

Meanwhile, I think cases such as
```vue
<template lang="jade">
div
 h1 Hello
</template>
```

I think you might want to output some error message for Vue that has non-supported sections. Same for stylus, sass, coffee.

tests/multiparser_vue/__snapshots__/jsfmt.spec.js.snap
vjeux: We started working on a full html pretty printer and added this vue file as an example. But it’s far from being usable so instead this PR just prints the html part as is and only reformats the js and css pieces.

vjeux:
What I went for is to just reprint as is things that prettier doesn’t support. So you don’t have the nice formatting but at least it doesn’t throw on you. This way you can just run prettier all the time and it will work when it can and do nothing when it can’t. The downside is that it makes troubleshooting harder but should help with adoption.

Thanks for looking at it!

chinchang:
@vjeux Any date for the 1.10 milestone release? Excited for this :)

PR #3570, opened 2017-12-25
add an example .prettierrc.js config

joshunger:
I thought a JS example would help too especially since comments are then valid.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -15,6 +15,16 @@ The options to the configuration file are the same as the [API options](options.
 
 ## Basic Configuration
 
+JS:
+
+```js
+// .prettierrc.js
+module.exports = {
+  "printWidth": 100,
+  "parser": "flow"
+}
+```
+
 JSON:
 
 ```json
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -15,16 +15,6 @@ The options to the configuration file are the same as the [API options](options.
 
 ## Basic Configuration
 
-JS:
-
-```js
-// .prettierrc.js
-module.exports = {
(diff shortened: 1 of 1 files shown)


azz:
Would you be able to put it below the JSON version?

FYI:

```bash
$ yarn lint-docs
  docs/configuration.md:22:2
  ⚠  22:2  Replace "printWidth" with printWidth  prettier/prettier
  ⚠  23:2  Replace "parser" with parser          prettier/prettier
  ⚠  24:1  Insert ;                              prettier/prettier
```

joshunger:
@azz fixed and thanks for the `lint-docs` issues

azz:
I wonder if we should be recommending `.prettierrc.js` or `prettier.config.js`. The latter seems to be more common, but eslint uses the former.

hawkrives:
My 2¢: `.prettierrc.js`

joshunger:
I came to the same conclusion that it should be `.prettierrc.js` because eslint uses `.eslintrc.js`.  Why do they both exist?

hawkrives:
@joshunger I believe it's because `cosmiconfig` (which I believe is still used for the config files) is not opinionated about where to store config options; they probably support "tool.config.js" and "toolrc.js" for the same reason that it handles .js, .json, and .yaml config files (and the package.json key too) – so that pretty much any tool can use cosmiconfig easily.

joshunger:
Do you want me to do something different?  I could put `.prettierrc.js or prettier.config.js` in the comment as one option.

azz:
That's fine, we already mention the alternative just above. Thanks.

PR #3572, opened 2017-12-26
Extract cleanAST to language-specific files

azz:


vjeux:
Nice! Can't wait to see it land :)

scripts/build/parsers.js
```diff
@@ -1,12 +1,8 @@
...
-  "language-css/parser-postcss",
-  "language-html/parser-parse5",
-  "language-markdown/parser-markdown",
-  "language-vue/parser-vue"
-];
+const globby = require("globby");
+const path = require("path");
+
+module.exports = globby.sync("language-*/parser-*", {
+  cwd: path.join(__dirname, "../../src")
```
j-f1: This should be `path.join(__dirname, "..", "..", "src")`.

azz: Why's that?

azz:
Moved to https://github.com/prettier/prettier/pull/3610

scripts/build/parsers.js
j-f1: On Windows, the path separators will be `\`, so `../../src` won’t work.

azz: It gets normalised correctly. 

```
$ node -p 'path.win32.join(__dirname, "../code/prettier")'
..\code\prettier
```

PR #3575, opened 2017-12-26
Split up comment printing

azz:
Unblocks https://github.com/prettier/prettier-python/pull/9

~This _works_ but the implementation leaves a bit to be desired. Need to improve the way the `printer`, `parser` and `plugins` are passed around rather than being looked up each time or added as more args everywhere.~

I've put `astFormat`, `plugins` and `printer` into the `options` object in the `normalize()` function. Not sure if this is the best solution but it cleans up the code significantly.


src/language-js/printer-estree.js
```diff
@@ -5167,9 +5167,77 @@ function willPrintOwnComments(path) {
   );
 }
 
+function whatShouldICallThis(node) {
```
azz: Seriously, no idea what to call this function...

azz: Going with `canAttachComment`.

vjeux:
I was going to suggest canAttachComment as well :)

index.js
```diff
@@ -179,16 +184,21 @@ function findSiblingAncestors(startNodeAndParents, endNodeAndParents) {
...
+function findNodeAtOffset(node, offset, options, predicate, parentNodes) {
   predicate = predicate || (() => true);
   parentNodes = parentNodes || [];
   const start = util.locStart(node);
   const end = util.locEnd(node);
   if (start <= offset && offset <= end) {
-    for (const childNode of comments.getSortedChildNodes(node)) {
+    for (const childNode of comments.getSortedChildNodes(
+      node,
+      undefined,
```
vjeux: nit: put an inline comment with that this variable is

PR #3576, opened 2017-12-26
Add Google Analytics on Playground

lipis:
Since playground is a different page, we should also include it in the analytics.


azz:
You didn't add any metrics? Do you just want pageviews?

lipis:
I guess for now it's a good start since we are checking pageviews for the rest of the website.

lipis:
Shall we go ahead with this one?

PR #3536, opened 2017-12-19
Prettier Plugin API

azz:
Very early progress - ~I've only moved files so far~. Want to keep the build green and get as much feedback as possible on this one.

Closes #3511

- [x] Refactor into language-based directories
- [x] Extract common code from JavaScript printer
- [x] Refactor printers to new API
- [x] Refactor parsers to new API
- [x] Implement extensibility via internal configuration
- [x] Investigate failing test
- [x] Test playground
- [x] Verify no breaking changes
- [x] Document the API
- [x] Refactor multiparser into per-language files

Follow ups (2.0 items):

- [x] Split out the support info and options to be language-specific
- [x] Make it externally configurable via CLI
- [x] Make it externally configurable via options/configuration file
- [ ] Switch to monorepo to enforce correct imports (?)
- [x] Use [prettier-python](https://github.com/prettier/prettier-python) as a proof-of-concept

PR #3584, opened 2017-12-26
Actually support plugins field in config

taion:


taion:
Sec, this only squelches the warning. It doesn't actually work.

taion:
Still one more bug to fix here

taion:
Good to go now. cc @azz

This fixes the hidden README example for prettier-python. Not sure what the test coverage policy here looks like as I didn't see any for #3536.

src/cli/util.js
```diff
@@ -619,7 +622,10 @@ function normalizeConfig(type, rawConfig, options) {
       return;
     }
 
-    const option = constant.detailedOptionMap[key];
+    let option = constant.detailedOptionMap[key];
+    if (type === "api" && option === undefined) {
```
taion: Otherwise we barf on unrecognized "plugins" in config file

src/cli/util.js
```diff
@@ -37,9 +37,12 @@ function getOptions(argv) {
   );
 }
 
-function dashifyObject(object) {
+function cliifyOptions(object) {
   return Object.keys(object || {}).reduce((output, key) => {
-    output[dashify(key)] = object[key];
+    const apiOption = constant.apiDetailedOptionMap[key];
+    const cliKey = apiOption ? apiOption.name : key;
```
taion: Need to map "plugins" to "plugin", else it always gets clobbered (even if `--plugin` isn't specified on CLI)

src/main/options.js
```diff
@@ -44,6 +44,8 @@ function normalize(options) {
   const normalized = Object.assign({}, options || {});
   const filepath = normalized.filepath;
 
+  normalized.plugins = loadPlugins(normalized);
```
taion: Need to load plugins first so `getSupportInfo` below can see Python

azz:
The original or didn't have tests because it was largely just a refactor. We should start adding integration tests for this kind of thing.

taion:
What do you think of adding an issue to track that and tackling the testing separately?

On master right now, the `plugins` field in config files doesn't really work at all, and parser inference for languages from plugins also doesn't work.

It's not exactly a blocker, but it makes testing prettier-python take a bit more effort than it needs to.

src/cli/constant.js
```diff
@@ -443,10 +443,19 @@ const detailedOptionMap = detailedOptions.reduce(
   {}
 );
 
+const apiDetailedOptionMap = detailedOptions.reduce(
+  (current, option) =>
+    option.forwardToApi && option.forwardToApi !== option.name
```
azz: Isn't forwardToApi `true` in a lot of places?

taion: At this point they're already normalized to the camelcased strings per https://github.com/taion/prettier/blob/70313e67950edb34a8cab0911c5b3ab1a90b7db6/src/cli/constant.js#L425

PR #3588, opened 2017-12-27
Move playground styles to the single CSS file

artiebits:
CSS that inside of playground.html probably was put there in the order to not concatenate it with Docusaurus's styles. But [Docusaurus provides `separateCss`](https://docusaurus.io/docs/en/site-config.html#optional-fields) property so we can move playground styles to the single file.


website/pages/playground/index.html
```diff
@@ -41,291 +41,7 @@
...
-      position: absolute;
-      top: 100%;
-      left: 50%;
-      transform: translateX(-50%);
-      border: 6px solid transparent;
-      border-bottom: none;
-      border-top-color: #000;
-    }
-  </style>
+  <link rel="stylesheet" href="/playground.css">
```
ikatyang: It seems you forgot to apply the `separateCss`, and the file should be in a folder since it requires a folder:

> `separateCss`: Folders inside which any css files will not be processed and concatenated to Docusaurus's styles.

image

artiebits: Oh, I am sorry, forgot to add `separateCss` property. 
It also works with files, and if there will be more than one file we can move them to the folder.

PR #3595, opened 2017-12-28
Support Relay Classic

taion:
Relay Classic uses `Relay.QL`, which was not previously picked up here.

### DIFF
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -39,8 +39,10 @@ function embed(path, print, textToDoc /*, options */) {
         parent &&
         ((parent.type === "TaggedTemplateExpression" &&
           ((parent.tag.type === "MemberExpression" &&
-            parent.tag.object.name === "graphql" &&
-            parent.tag.property.name === "experimental") ||
+            ((parent.tag.object.name === "graphql" &&
+              parent.tag.property.name === "experimental") ||
+              (parent.tag.object.name === "Relay" &&
+                parent.tag.property.name === "QL"))) ||
             (parent.tag.type === "Identifier" &&
               (parent.tag.name === "gql" || parent.tag.name === "graphql")))) ||
           (parent.type === "CallExpression" &&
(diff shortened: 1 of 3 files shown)


taion:
@azz 

Nope. That template is actually a marker for a Babel plugin anyway, and the only form that shows up in OSS code that's supported is `Relay.QL`: https://github.com/facebook/relay/blob/fe0e70f10bbcba1fff89911313ea69f24569464b/packages/babel-plugin-relay/getValidRelayQLTag.js#L22-L28

It'd also be a bad idea. Like, for the other Relay Classic exports, it'd be quite bad to see `createContainer` and `Renderer` running around your codebase instead of `Relay.createContainer` and `Relay.Renderer`.

j-f1:
Seems like `RelayQL` is also supported by the Babel plugin.

taion:
It is, but it's not used in OSS. I guess we can add it? But I'd imagine the FB folks _must_ have noticed this already :p

taion:
@kassens @vjeux You two seem to have touched the Prettier config in the Relay repo the most. Should we also pick up on `RelayQL`, `RelayClassic_DEPRECATED.QL`, or `RelayClassic.QL` here? I don't think anybody in OSS uses anything other than `Relay.QL`. Are those other forms still extant within FB?

kassens:
I think we only use `RelayClassic_DEPRECATED.QL` nowadays (keep pushing the deprecation!).

I think a good test here would be what happens when there are string interpolations which are no longer used in RelayModern eg.

```
Relay.QL`
  fragment on User {
    ${Comp.getFragment('user')}
  }
`;
```

vjeux:
prettier doesn’t support Relay Classic, only Relay Modern

taion:
Interesting, okay – it doesn't really work at all. I'm going to revert this to avoid confusion then.

azz:
Papertrail: this was reverted in #3606

Issue #3600, opened 2017-12-28
GraphQL: Support new multiple inheritance syntax


lydell:
https://github.com/graphql/graphql-js/pull/1169

```graphql
type Foo implements Bar & Baz { field: Type }
```

The above currently fails to parse.

It also replaces this syntax:

```graphql
type Foo implements Bar, Baz { field: Type }
```

We could try to support both by doing:

```js
function parseGraphQL(source) {
  try {
    return parse(source, {allowLegacySDLImplementsInterfaces: false})
  } catch {
    return parse(source, {allowLegacySDLImplementsInterfaces: true})
  }
}
```

lydell:
Merging into #3601.

vjeux:
Should also do https://github.com/prettier/prettier/issues/3619 at the same time.

PR #3605, opened 2017-12-29
Update GraphQL parser

### LINKED ISSUE #3601: GraphQL: Support new stuff in the latest version
- Block strings: https://github.com/facebook/graphql/pull/327
- Descriptions as strings: https://github.com/graphql/graphql-js/pull/927
- New multiple inheritance syntax: https://github.com/graphql/graphql-js/pull/1169
- `extend type`: https://github.com/graphql/graphql-js/pull/1117/files
- `UndefinedType`: https://github.com/graphql/graphql-js/pull/1117/files
- Empty types are no longer allowed: https://github.com/graphql/graphql-js/pull/1171

This file contains all syntax: https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/schema-kitchen-sink.graphql

To try to support both old and new syntax we should set these options:

```js
{
  allowLegacySDLEmptyFields: true,
  allowLegacySDLImplementsInterfaces: true,
}
```

vjeux:
Fixes #3601

This adds support for
- description as strings
- extending all the possible types
- block strings
- allow removing {} if there's no implementation for all graphql types

This is a breaking change but shouldn't be a big issue.
- Empty types are no longer allowed (there's an option to enable it but it hasn't been released yet). The fix is to remove `{}`

Something that hasn't been changed because not released:
- Doesn't support the new `&` separator for implementing multiple interfaces

A bug has been fixed:
- Now properly prints @directives for unions.

### DIFF
--- a/src/language-graphql/printer-graphql.js
+++ b/src/language-graphql/printer-graphql.js
@@ -31,9 +31,11 @@ function genericPrint(path, options, print) {
       ]);
     }
     case "OperationDefinition": {
+      const hasOperation = options.originalText[util.locStart(n)] !== "{";
+      const hasName = !!n.name;
       return concat([
-        n.name === null ? "" : n.operation,
-        n.name ? concat([" ", path.call(print, "name")]) : "",
+        hasOperation ? n.operation : "",
+        hasOperation && hasName ? concat([" ", path.call(print, "name")]) : "",
         n.variableDefinitions && n.variableDefinitions.length
           ? group(
               concat([
@@ -53,7 +55,7 @@ function genericPrint(path, options, print) {
             )
           : "",
         printDirectives(path, print, n),
-        n.selectionSet ? (n.name === null ? "" : " ") : "",
+        n.selectionSet ? (!hasOperation && !hasName ? "" : " ") : "",
         path.call(print, "selectionSet")
       ]);
     }
@@ -123,6 +125,15 @@ function genericPrint(path, options, print) {
       return n.value;
     }
     case "StringValue": {
+      if (n.block) {
+        return concat([
+          '"""',
+          hardline,
+          n.value.replace(/"""/g, "\\$&"),
+          hardline,
+          '"""'
+        ]);
+      }
       return concat(['"', n.value.replace(/["\\]/g, "\\$&"), '"']);
     }
     case "IntValue":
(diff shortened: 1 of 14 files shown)


src/language-graphql/printer-graphql.js
```diff
@@ -122,6 +124,15 @@ function genericPrint(path, options, print) {
       return n.value;
     }
     case "StringValue": {
+      if (n.block) {
+        return concat([
+          '"""',
+          hardline,
+          n.value.replace(/"""/g, "\\$&"),
+          hardline,
```
lydell: Doesn't this add newlines at the start and end of the block string – that is, changing the contents of the string? Is that safe?

lydell:
When merging this we should

- re-open #3600 
- open an issue for adding the new GraphQL parser options when released

src/language-graphql/printer-graphql.js
vjeux: Actually, there's an issue with the graphql parser where both of them are the same ( https://github.com/graphql/graphql-js/issues/1188 ). I'm going to just print the value from the original source for now.

lydell: Reading https://github.com/graphql/graphql-js/issues/1188#issuecomment-354614921 it seems like the spec says that "blank initial and trailing lines" are excluded. So I guess that’s a good thing? That means we can safely print this:

```
"""line one
line two"""
```

as

```
"""
line one
line two
"""
```

?

Not sure what "[exclude] uniform indentation with {BlockStringValue()}" means though...

j-f1: I *think* it means that code like this:

```
  """
  this text
  is indented
  """
```
is equivalent to:
```
"""
this text
is indented
"""
```

lydell: That would be really nice! If so, we could fix indentation inside block strings as well!

IvanGoncharov: > Not sure what "[exclude] uniform indentation with {BlockStringValue()}" means though...

@lydell Example from @j-f1 is fully correct. You can read detail algorithm here:
https://github.com/facebook/graphql/pull/327/files#diff-fe406b08746616e2f5f00909488cce66R789

BTW. Would be great if you [make it compatible](https://github.com/graphql/graphql-js/blob/master/src/language/printer.js#L260-L266) with `print` function from `graphql-js`.

IvanGoncharov: @lydell @j-f1 You can also see some tests here:
https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/printer-test.js#L74-L104

duailibe: Oh I didn't know graphql-js had a `print` function

IvanGoncharov: @duailibe Yes, it's part of `graphql-js` since it was first published in 2015.
It would be great if you figure out how to reuse it at least partly.

But I think most important is to have `prettier` in sync with `print` since it's de-facto standard in the GraphQL world. For example, people use `print` to print schema of remote APIs.

vjeux: The biggest differences between prettier and graphql printer is the fact that prettier is 80-columns aware and supports comments.

Thankfully, graphql is a pretty simple language so it has been easy to implement properly. I don't think that we're going to reuse any code from the project but we are definitely maintaining the same conventions that graphql is using.

The prettier graphql printer is currently used to pretty print all the graphql queries embedded inside of template literals at Facebook. So this part is rock solid. The graphql schema definition has seen less extensive use as far as I know (but correctly prints all the schema we have at Facebook, even if we don't actively format them using it).

For this specific example, we should do whatever you are doing. It seems like the only difference is to handle "inline" triple quote strings that have spaces around.

IvanGoncharov: @duailibe I forgot to mention that GraphiQL uses `print`:
image
So it's would be great to ensure that by default `prettier` produces the same output.

IvanGoncharov: > The biggest differences between prettier and graphql printer is the fact that prettier is 80-columns aware and supports comments.

@vjeux Make sense 👍  and thank you for the explanation.

PR #3610, opened 2017-12-30
Extract cleanAST to language-specific files

azz:
Easier to re-do that to rebase the previous PR. 

Supersedes #3572

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -20,7 +20,6 @@ const validator = require("./validator");
 const options = require("../main/options");
 const apiDefaultOptions = options.defaults;
 const normalizeOptions = options.normalize;
-const parsers = require("../main/parser");
 const logger = require("./logger");
 const thirdParty = require("../common/third-party");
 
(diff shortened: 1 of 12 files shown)


taion:
Ref https://github.com/prettier/prettier-python/pull/17#discussion_r159587398

At least in the Python case, the AST doesn't have a built-in JSON representation, and instead the conversion of the AST (from a Python object) to JSON is actually some of the code under test (for debugging).

However, there is a built-in way to serialize an AST to a string (`ast.dump` on a Python AST object) that can be used for equality checking in tests with the correct semantics there.

I guess there might not be a perfect answer here. Maybe if there were a base `Plugin` class that had a `parseAndMassage` method with a default implementation that called into `massageAstNode`? Or possibly just not worth the complexity here.

For purposes of `debug-check`, `astor.dump_tree` also pretty-prints Python ASTs with nice indentation, with roughly equivalent semantics to `cleanAST` here, including diffability as text.

PR #3573, opened 2017-12-26
New playground

duailibe:
Context: #3352

This is still WIP and just a POC of a Preact app with no build steps (I opened https://github.com/facebook/Docusaurus/issues/292 to discuss how to reuse Docusaurus infrastructure - then we could use classes and stuff). This is very early but it was usable (commit is a few weeks old, before the big refactor).

I'm opening before finishing to gather feedback: should we keep the "no build steps" for this? It's nice having only one command and not as nice having to use two separate workflows when working on the site (one for docusarus part, other for the playground only) but it's doable.. Ideally I'd want a way to integrate with docusaurus infrastructure (transpilation, dev server, React itself).. anyway, tell me what you think!

PR #3614, opened 2017-12-30
fix(playground): fix report version

ikatyang:
Revert #3613 and apply another solution since we don't actually expose `prVersion` from `prettier` module, which cause the wrong version on <kbd>copy markdown</kbd>.  😂

cc @azz

### DIFF
--- a/scripts/build/build-docs.js
+++ b/scripts/build/build-docs.js
@@ -24,7 +24,7 @@ const parserPaths = parsers.map(stripLanguageDirectory);
 
 if (isPullRequest) {
   const pkg = require("../../package.json");
-  pkg.prVersion = `pr-${process.env.REVIEW_ID}`;
+  pkg.version = `999.999.999-pr.${process.env.REVIEW_ID}`;
   pipe(JSON.stringify(pkg, null, 2)).to("package.json");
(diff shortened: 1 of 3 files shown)


scripts/build/build-docs.js
```diff
@@ -24,7 +24,7 @@ const parserPaths = parsers.map(stripLanguageDirectory);
 
 if (isPullRequest) {
   const pkg = require("../../package.json");
-  pkg.prVersion = `pr-${process.env.REVIEW_ID}`;
+  pkg.version = `999.999.999-pr.${process.env.REVIEW_ID}`;
```
j-f1: Why not use `[current version]-pr.1234`?

duailibe: I think a version followed by `-something` is meant to denote pre-release versions (which is not the case)

ikatyang: Ref #3433 and #3573

https://github.com/prettier/prettier/blob/7eca8011bc5e6f6071e94116619b3fda377befbb/src/common/support.js#L12-L17

We use this version to determine which `language` (and `option` in the future PR) can be used, which means it should be greater than the current version in PR builds so we can test it on the preview playground.

PR #3622, opened 2018-01-01
refactor(optios): use supportOptions to generate CLI options

ikatyang:
- use supportOptions to generate CLI options
- replace `deprecated.js`, `jest-validate` and warnings in `options.js` with `main/options-*.js`
- _this PR does not load options from external plugins, only internal plugins_


src/main/options.js
```diff
@@ -1,139 +1,81 @@
...
-  requirePragma: false,
-  semi: true,
-  proseWrap: "preserve",
-  arrowParens: "avoid",
-  plugins: [],
+const hiddenDefaults = {
   astFormat: "estree",
-  printer: {},
-  __inJsTemplate: false
+  printer: {}
```
azz: This was just to satisfy jest-validate. Might not be required now.

ikatyang: It's also used to bypass the option validation, see L62.

ikatyang:
We need to generate `cli-constant` dynamically, and probably need to parse cli args twice to get the proper `optionInfos` from `plugins`, it'd be a relative big change I guess, should be in another PR.

src/cli/util.js
```diff
@@ -597,135 +605,11 @@ function groupBy(array, getKey) {
...
-    }
-
-    if (type === "cli" && key.length === 1) {
-      // do nothing with alias
-      return;
-    }
-
-    let option = constant.detailedOptionMap[key];
-    if (type === "api" && option === undefined) {
-      option = constant.apiDetailedOptionMap[key];
```
taion: This was the only consumer for `apiDetailedOptionMap`, so probably just remove it entirely. This only dates back to https://github.com/prettier/prettier/pull/3584. It's not part of the public API and never made it into a release anyway.

taion: Oops, misread. Never mind. Sorry.

azz:
Looks good to me but I'm super tired and have had some 🍷 so maybe wait for someone else to review?

ikatyang:
Anyone wants to review this one?

tests_integration/cli/config/invalid/option/trailingComma
```diff
@@ -1,3 +1,3 @@
 {
-    "trailing-comma": "wow"
+    "trailingComma": "wow"
```
ikatyang: #3756

azz: Nice.

azz: I'll edit the OP.

lydell:
Let's just merge this?

PR #3632, opened 2018-01-02
Normalize GraphQL multi-line strings

vjeux:
See discussion in https://github.com/prettier/prettier/pull/3605#discussion_r159120321

### DIFF
--- a/src/language-graphql/printer-graphql.js
+++ b/src/language-graphql/printer-graphql.js
@@ -126,9 +126,13 @@ function genericPrint(path, options, print) {
     }
     case "StringValue": {
       if (n.block) {
-        // It is not possible to distinguish between """\nabc\n""" and """abc"""
-        // https://github.com/graphql/graphql-js/issues/1188
-        return options.originalText.slice(util.locStart(n), util.locEnd(n));
+        return concat([
+          '"""',
+          hardline,
+          n.value.replace(/"""/g, "\\$&"),
+          hardline,
+          '"""'
+        ]);
       }
       return concat(['"', n.value.replace(/["\\]/g, "\\$&"), '"']);
     }
--- a/tests/graphql_kitchen_sink/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/graphql_kitchen_sink/__snapshots__/jsfmt.spec.js.snap
@@ -110,9 +110,12 @@ fragment frag on Friend {
   foo(
     size: $size
     bar: $b
-    obj: { key: "value", block: """
+    obj: {
+      key: "value"
+      block: """
       block string uses \\"""
-  """ }
+      """
+    }
   )
 }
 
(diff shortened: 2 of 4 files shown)


j-f1:
Could you add a test for a legitimately-indented string?

```graphql
something {
  """
some
multiline
string
  """
}

# becomes

something {
  """
  some
  multiline
  string
  """
}

```

---


...or maybe the string content should be indented

vjeux:
This won't do what you suggest. It'll stay as is with this implementation.

src/language-graphql/printer-graphql.js
```diff
@@ -126,9 +126,13 @@ function genericPrint(path, options, print) {
...
-        // It is not possible to distinguish between """\nabc\n""" and """abc"""
-        // https://github.com/graphql/graphql-js/issues/1188
-        return options.originalText.slice(util.locStart(n), util.locEnd(n));
+        return concat([
+          '"""',
+          hardline,
+          n.value.replace(/"""/g, "\\$&"),
+          hardline,
+          '"""'
+        ]);
```
lydell: Don’t we need to split on newlines and `concat` them, so that all lines get the correct base indentation?

I’d like to see this test:

```graphql
type Foo {
"""
This is a description
of the `one` field.
"""
  one: Type
}
```

turn into:

```graphql
type Foo {
  """
  This is a description
  of the `one` field.
  """
  one: Type
}
```

However, when I tried that in the [PR playground](https://deploy-preview-3632--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAYhCbYAHSmJHMqgBUALASwGdsnsBDbAEzkbACd6GGPWikIAM2wxaOAAbQ4s7OPpwANpwB0pShQpRs2BUmzVMcUgF9SIADQgIQkVEbJQbPnwgB3AAoeEVxQ2ADcIek47EDZGGGRxNjVGOHsAIz42MABrOBgAZQxM+igAc2QYPgBXFJBOCDB4xOT7YuS+GF8MkoBbNkakmoArRgAPACEM7Ny8tm64ABliuH7mh0qYDHWAJhWawr425BASjIxaAEc1KIwBWAB1COlkAA4ABnsbiGS706ObnjgfBCy3sfDg50q9DBnTYPT6SASA3syW69HKVRqjGKJTUcAAipUIPBdvYYGxUg9OE8kFtSRl6GpsQBhCDdXpHKAKKKVZLUclBRHJSyWIA) that doc string wasn’t printed at all...

vjeux: Turns out i'm not printing the description for "FieldDefinition". We should look at all the places where descriptions are valid and make sure that they are all correctly printed.

vjeux:
@lydell thanks for the good feedback! I added a commit which does what you suggested :)

duailibe:
Also missing in:
- `EnumValueDefinition`
- `InputValueDefinition`

vjeux:
@duailibe I'm supposed to be working on my self-review right now... @duailibe would you be able to send a PR for them?

duailibe:
Sure! :)

lydell:
@duailibe Also for parameters:

```graphql
type T {
  q(
    """
    docs
    """
    a: String
  ): S
}
```

PR #3635, opened 2018-01-03
WIP - Preserve users white space when formatting handlebars

robert-j-webb:
I started working on this and it's pretty difficult! Going to give it my best shot.


src/common/clean-ast.js
```diff
@@ -54,6 +55,216 @@ function massageAST(ast, options, parent) {
     if (result) {
       return result;
     }
+=======
```
azz: Conflicts. This stuff lives in language specific files now, see #3610

src/language-handlebars/printer-glimmer.js
```diff
@@ -41,8 +41,19 @@ function print(path, options, print) {
 
   switch (n.type) {
     case "Program": {
-      return group(
-        join(softline, path.map(print, "body").filter(text => text !== ""))
+      return concat(
+        path.map(print, "body").map(node => {
+          if (node.trim && node.trim() === "") {
```
vjeux: Can't you use utils.hasNewLineAfter?

vjeux:
The way we handle it in the JS printer is when printing a list of elements, check if the next line is empty and force an empty line if that's the case and not the last element.

```js
    if (util.isNextLineEmpty(text, stmt) && !isLastStatement(stmtPath)) {
      parts.push(hardline);
    }
```

https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L2884-L2886

This should simplify the logic you have.

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -258,10 +264,18 @@ exports[`element-node.hbs 1`] = `
   A long enough string to trigger a line break that would prevent wrapping more and more.
 </div>
 <div>
+  A line break with its own new lines
+  inside of it forming something of 
+  a paragraph.
```
vjeux: For this, we use the `fill` primitive which adds whitespace when the first word would hit the limit. https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L4413-L4452

robert-j-webb: Hmm, so using fill would require `split` on the text characters because fill takes an array. One method is something like, `fill(n.chars.split(/\s+/).map(a => elem + " ")`, but that means that if a user has HTML which has double spacing in it, then we are going  to chew it away. Ex:

`<div>Hello   World</div>`
Becomes:
`<div>Hello World</div>`

I think we should do this. It's much better than preserving multiple spaces, because that gets ignored by HTML anyway, outside of `<pre>` tags.

vjeux: Someone posted a long comment as to why whitespace and html is going to be difficult. For example you can apply `white-space: pre` to arbitrary nodes in CSS which we have no way to know about when just looking at the template.

This means that the only want to guarantee correctness is not to change anything, which defeats the purpose. So we should use good judgement as to when we diverge and when we respect it and accept the fact that we can't get it right all the time.

src/language-handlebars/printer-glimmer.js
vjeux: Any reason you aren't using `util.isNextLineEmpty()`?

src/language-handlebars/printer-glimmer.js
```diff
@@ -41,8 +41,19 @@ function print(path, options, print) {
...
+            const newLines = node.match(RegExp("\n", "g"));
```
vjeux: Need to make sure that it works properly with `\r\n`

src/language-handlebars/printer-glimmer.js
```diff
@@ -183,7 +184,8 @@ function print(path, options, print) {
       return concat([n.key, "=", path.call(print, "value")]);
     }
     case "TextNode": {
-      return n.chars.replace(/^\s+/, "").replace(/\s+$/, "");
```
vjeux: This can be written as `/^\s+|\s+$/` in one regex fyi

vjeux: Or just call `.trim()` :p

src/language-handlebars/parser-glimmer.js
```diff
@@ -2,33 +2,27 @@
 
 const createError = require("../common/parser-create-error");
 function removeEmptyNodes(node) {
-  return (
-    node.type !== "TextNode" ||
-    (node.type === "TextNode" &&
-      node.chars.replace(/^\s+/, "").replace(/\s+$/, "") !== "")
-  );
+  if (node.type === "TextNode") {
+    node.chars = node.chars.replace(/[ ]+/, "").replace(/[ ]+.*$/, "");
```
vjeux: what does the second replace do?

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
j-f1: Also, if you know that you’re going to set `white-space: pre`, you can put a `<!-- prettier-ignore -->` before the element you apply it to.

robert-j-webb: Alright, those are great points. I'm going to pretend `<pre>` and `white-space: pre` don't exist for the purpose of developing something that's more useful to the user anyway.

robert-j-webb:
I'm not done with this at all! I'm still trying to figure out the approach. If you want to pair some time we can talk about what's going through my head but it's not doing what I want it to yet, so the code is definitely very messy.

j-f1:
Do you want to keep working on this @robert-j-webb or is it OK to close it?

PR #3641, opened 2018-01-03
Prevent adding softline after arrow attribute with comments

### LINKED ISSUE #3591: Double newline after commented arrow function in JSX attribute
**Prettier 1.9.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAzgBwIZQATYwwBOAvMADr54D0NekAtowjFXngBQCUpAfCQCucKgF8afANwgANCAiYYAS2jpkobMWIQA7gAVNCNSmwA3CEoAmskNnQxkAM2wAbdHDkAjYtjABrOBgAZRwwJSgAc2QhDxBLCDAnV3c5cPdiGD0fCMZsJLdYgCt0AA8AIR9-QKDsFgAZcLh8lPlBGEw2gCZm2JxidOQQT2xPAE8XaBtMYnCYAHUrGAALZABGAGY5aYh3OZ9MQem4dNMmuWI4AEdBJQus7By8pGcCuXdGJWjiYTfwiJc4ABFQQQeA9OQwEYLSzLZCdCE+JQuP4AYQgzCeICg0DOIEE7gAKiNjC93KJREA)
```sh
--print-width 13
```

**Input:**
```jsx
<span attr={
  // comment
  ()=>true
}/>;
```

**Output:**
```jsx
<span
  attr={
    // comment
    () =>
      true

  }
/>;
```

**Expected behavior:**
Just one newline after the arrow function. The two newlines are coming from:
1. printing the `ArrowFunctionExpression` node, where `shouldAddSoftLine` is `true`, because the parent node is `JSXExpressionContainer`.
2. printing the `JSXExpressionContainer` node, where `shouldInline` is `false` because of the comment.

I suspect that the condition to set `shouldAddSoftLine` to `true` is too broad:
```js
path.getParentNode().type === "JSXExpressionContainer"
```
The softline should be added only when the expression inside is printed in the "inline" mode.

duailibe:
Fixes #3591

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -499,8 +499,9 @@ function printPathNoParens(path, options, print, args) {
       // with the opening (, or if it's inside a JSXExpression (e.g. an attribute)
       // we should align the expression's closing } with the line with the opening {.
       const shouldAddSoftLine =
-        (args && args.expandLastArg) ||
-        path.getParentNode().type === "JSXExpressionContainer";
+        ((args && args.expandLastArg) ||
+          path.getParentNode().type === "JSXExpressionContainer") &&
+        !(n.comments && n.comments.length);
 
       const printTrailingComma =
         args && args.expandLastArg && shouldPrintComma(options, "all");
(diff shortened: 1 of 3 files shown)


vjeux:
<3

PR #3653, opened 2018-01-05
Update plugins docs

azz:
* Add new [Swift Plugin](https://github.com/prettier/prettier-swift).
* Document automatic plugin loading.


duailibe:
Maybe we should look into using docs versioning, since docusaurus supports it

PR #3655, opened 2018-01-05
Added Contiamo to list of users

TejasQ:
Simply added [my organization](https://github.com/Contiamo/operational-ui/blob/master/package.json#L18) to the list of satisfied Prettier users. :)


azz:
🔬 Thanks!

PR #3664, opened 2018-01-07
Allow Plugins to preprocess text

sirlantis:
As noted in prettier/prettier-swift#1

This allows plugins to preprocess the text that get's passed in and have Prettier update its `text` variable that is passed to `parse` and `attachComments` later.

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,7 +97,9 @@ function formatWithCursor(text, opts, addAlignmentSize) {
 
   addAlignmentSize = addAlignmentSize || 0;
 
-  const ast = parser.parse(text, opts);
+  const result = parser.parse(text, opts);
+  const ast = result.ast;
+  text = typeof result.text !== "undefined" ? result.text : text;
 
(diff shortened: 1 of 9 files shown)


azz:
How does this implementation differ from just calling `preprocess` inside `parse`? I thought you were trying to set this `text` local, which is in the parent scope?

https://github.com/prettier/prettier/blob/e4e3c757fb85f47f0336fc2f13678ae93facde40/index.js#L117

sirlantis:
@azz I should just not create PRs at 2am. :| Thanks for noticing. I added a regression test to `prettier-swift` and updated the PR. I hope this change is not too intrusive.

azz:
Cool, code change looks good. Could you add an integration test? (test code in `tests_integration/__tests__/plugin-preprocess.test.js` and fixtures in `tests_integration/plugins/preprocess`)?

sirlantis:
@azz thanks. Missed the comment, will include test immediately next time!

PR #3665, opened 2018-01-07
Force break if comment in return argument

### LINKED ISSUE #3648: comment in function chain causes function to be wrapped in `()` with an extra space
**Prettier 1.9.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMBGAFBrGAS2lwEoACYAHSgooCc4Z0G6CwAmW+3igen6QAtsIQwefCgDoARgEMG5ANy0AvmWUgANCAgAHYtADOyUIoYQA7gAVFCUynkA3CEQAmOkPOMxkqeQAbYzhdWQZ5MABrZgBlfUiiKABzZBgGdFCQdwgwfyCQ3SSQhhgbCOTheXzgrIArYwAPACEI6Lj5MQAZJLgawr10GH0hzn6shIYS5BAFWQBPQOgvfQYkmAB1DxgAC2QADgAGXVWIEI2I-RnVuBLnPt0mAEd0IiZy+UrqpADa3RDhEQ0hkssYkslAnAAIroCDwca6GDyWRbdy7ZCcRERIiBcEAYQgom+ICg0AeIHQIQAKsjHL8Qmo1EA)

**Input:**
```jsx
func1(function() {
  return func2
      //comment
      .bar();
});
```

**Output:**
```jsx
func1(function() {
  return (func2
      //comment
      .bar() );
});

```

**Expected behavior:**

 I wouldn't think it would wrap the function chain in `()` and at the least the extra space at the end wouldn't be necessary.

duailibe:
This one was kind of tricky, I'm still not sure why it was breaking in `(function () { /* code */ })` but not `fn(function() { /* code */ });`. Either way, if there's a comment in a return argument it probably requires a hardline, instead of a softline/line. Apparently, there weren't any specific tests relying on the softline/line so I guess this should be ok?

Fixes #3648

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -760,8 +760,8 @@ function printPathNoParens(path, options, print, args) {
           parts.push(
             concat([
               " (",
-              indent(concat([softline, path.call(print, "argument")])),
-              line,
+              indent(concat([hardline, path.call(print, "argument")])),
+              hardline,
               ")"
             ])
           );
--- a/tests/return/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/return/__snapshots__/jsfmt.spec.js.snap
@@ -55,9 +55,41 @@ exports[`comment.js 1`] = `
 function f() {
   return /* a */;
 }
+
+function f() {
+  return (
+    foo
+      // comment
+      .bar()
+  );
+}
+
+fn(function f() {
+  return (
+    foo
+      // comment
+      .bar()
+  );
+});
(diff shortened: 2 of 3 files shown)


azz:
Could you add another test case for

``js`
function x() {
  return func2
      //comment
      .bar();
};
```

**Comment by duailibe** [comment]
done!

**Comment by duailibe** [comment]
Actually I don't think this is the correct fix... 

This shouldn't have parens:

```js
function x() {
  return func2
      //comment
      .bar();
};
```

duailibe:
Looks like the above is the expected output of prettier... we should probably go forward with this fix then

PR #3666, opened 2018-01-07
docs(readme): travis master badge

ikatyang:
No more image caused by branches other than master.


azz:
Nice - noticed that once or twice but forgot to investigate!

PR #3668, opened 2018-01-07
fix(playground): avoid `assign to readonly` error on Safari

### LINKED ISSUE #3617: Preview playground seems broken on Safari
Tested on Safari 11.0.2 and iOS 10

For example: https://deploy-preview-3614--prettier.netlify.com/playground/

image

ikatyang:
Fixes #3617

### DIFF
--- a/scripts/build/build-docs.js
+++ b/scripts/build/build-docs.js
@@ -2,6 +2,7 @@
 
 "use strict";
 
+const fs = require("fs");
 const path = require("path");
 const shell = require("shelljs");
 const parsers = require("./parsers");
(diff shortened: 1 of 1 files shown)


ikatyang:
Got another error:

image

Any idea? 🤔

j-f1:
How did you get that error @ikatyang?

ikatyang:
It works now, tested on Safari 11.0.2 and iOS 10.

The root cause is that the top-level `var crypto = true` is not working since the global `crypto` is readonly, wrapping it in an IIFE so that it won't access to the global `crypto`.

scripts/build/build-docs.js
```diff
@@ -38,12 +39,13 @@ shell.cp(`${prettierPath}/index.js`, `${docs}/index.js`);
...
 );
-shell.sed(
-  "-i",
-  /(var crypto=true;)/,
-  "try{$1}catch(e){}",
-  `${docs}/index.js`
-);
+
+// wrap content with IIFE to avoid `assign to readonly` error on Safari
+(function(filename) {
```
azz: I think you can just set `format` to `iife` in `rollup.docs.config.js`?

ikatyang: We built `lib/index.js` from `rollup.index.config.js`, and we need to adjust a lot to get it worked, IMO this is the easiest fix.

PR #3675, opened 2018-01-08
Add pretty-quick pre-commit docs

azz:
Adding docs for integration with [`pretty-quick`](https://github.com/azz/pretty-quick).

I'm replacing the previous option (`pre-commit`) as it does a `git add .` which it really shouldn't be doing.

As it's a brand new package I'll leave it in the number two spot for now, but I think this could potentially be the recommended pre-commit hook in the future as it should be the fastest, and it's the easiest to set up.

[🌐](https://deploy-preview-3675--prettier.netlify.com/docs/en/precommit.html#option-2-pretty-quick-https-githubcom-azz-pretty-quick)

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -30,29 +30,25 @@ and add this config to your `package.json`:
 
 See https://github.com/okonet/lint-staged#configuration for more details about how you can configure lint-staged.
 
-## Option 2. [pre-commit](https://github.com/observing/pre-commit) (JS version)
+## Option 2. [pretty-quick](https://github.com/azz/pretty-quick)
 
-Install the package:
+Install it along with [husky](https://github.com/typicode/husky):
 
 ```bash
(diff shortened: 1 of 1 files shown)


duailibe:
> I think this could potentially be the recommended pre-commit hook in the future

Can you specify which files you want prettified and which you don't? I guess that should be set in `.prettierignore`...

azz:
Yes you'd opt-out rather than opt-in currently, `*.css`, etc. in `.prettierignore` would be the way to go.

asottile:
@azz pre-commit does *not* do a `git add .` -- could you clarify what you mean?

asottile:
oh I see now, another project with the same name... naming is hard :)

I was confusing:

- https://github.com/pre-commit/pre-commit (now the third option, used to be the second option, project I maintain) 
- https://github.com/observing/pre-commit

/me slinks away

PR #3676, opened 2018-01-08
fix(multiparser): respect 0-indent markdown-in-js

ikatyang:
Fixes #3670 

<details>
<summary>playground</summary>

**Prettier pr-3676**
[Playground link](https://deploy-preview-3676--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBbAJgAwDpQCoAWAlgM4AEANkVHGSQRAK4XpQDkMZARrdegvHRkAhp2EUKZamRgFaAJzglmMagHMyTGAAdGMADTc9UzqQB0uHFFwYAFFYAKimKrjyy6CEplyy8omoEnLLqmlA+pJTUtMJQQtqKJG4Abt6yvHECokTQlHCpFBZQmACUIPogENqq0CTIoMLy8hAA7g6NCHUowskQROjlIMIkMMgAZuJJFVzywmAA1nAwAMrac+rIMPKMcBWeYOOTuyDUSfIwTsJqqMKHFFMgAFYkAB4AQrMLS8vCqHAAMtE7g8tLoYAAmYHHNbyM7IEBcYRcACeFGggwS1BgAHV+rJkAAOAAMFQSECS2Nm2nhCSUKTgg0UAEdGERFJdrrckBN7sckqgiJttnz1BQ4ABFRgQeBQiowJG49D4pDguWzIhUKBqADCEFQN3hUGgDIqjCSeCRXR5SQAvjagA)

**Input:**
```jsx
md`
This line shouldn't be indented at all in the resulting output, but it is.
`

md(`
Prettier does the right thing on this line and preserves the indentation level.
`)
```

**Output:**
```jsx
md`
This line shouldn't be indented at all in the resulting output, but it is.
`;

md(`
Prettier does the right thing on this line and preserves the indentation level.
`);

```

</details>

---

Fixes #3740

<details>
<summary>playground</summary>

**Prettier pr-3676**
[...]

PR #3677, opened 2018-01-08
Add reset indent primitive

duailibe:
This PR adds a new primitive for reseting indentation (see https://github.com/prettier/prettier/pull/3676#discussion_r160227270 for context).

To test this, I've decided to refactor the `TemplateElement` printing but it seems the `literalline` does more than just "not indenting" the next line (e.g. it also includes trailing whitespaces) so I couldn't simply change those to `hardline` and wrap with a `resetIndent`.

What I did was remove the "not indent the next line" feature of `literalline` and added the `resetIndent`. With that, all tests passed!

Let me know what you think!

### DIFF
--- a/commands.md
+++ b/commands.md
@@ -138,7 +138,7 @@ Specify a line break that is **always** included in the output, no matter if the
 declare var literalline: Doc;
 ```
 
-Specify a line break that is **always** included in the output, and don't indent the next line. This is used for template literals.
+Specify a line break that is **always** included in the output, and doesn't trim whitespaces in the end of lines. This is used for template literals.
 
 ### lineSuffix
 
@@ -192,6 +192,16 @@ declare function indent(doc: Doc): Doc;
 ```
 
 Increase the level of indentation.
+
+
+### resetIndent
+
+```ts
+declare function resetIndent(doc: Doc): Doc;
+```
+
+Resets the level of indentation.
+
 
 ### align
 
--- a/commands.md
+++ b/commands.md
@@ -138,7 +138,7 @@ Specify a line break that is **always** included in the output, no matter if the
 declare var literalline: Doc;
 ```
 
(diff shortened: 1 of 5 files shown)


commands.md
```diff
@@ -138,7 +138,7 @@ Specify a line break that is **always** included in the output, no matter if the
 declare var literalline: Doc;
 ```
 
-Specify a line break that is **always** included in the output, and don't indent the next line. This is used for template literals.
+Specify a line break that is **always** included in the output, and doesn't trim whitespaces in the end of lines. This is used for template literals.
```
j-f1: > trim whitespaces **at** the end of lines

?

duailibe:
Another possible API `indent(doc, { reset: true })` or even `indent(doc, { length: 0 })` (this one would be more "powerful")

ikatyang:
I'm thinking of this case:

````md
- 1
  - 2
    - 3
      ```js
      md`
      # this is the root indent
      `
      ```
````

We probably also need another Doc to indicate that ` ```js ` is the root indent level, otherwise it'll be a broken codeblock:

````md
- 1
  - 2
    - 3
      ```js
      md`
# this is the root indent
      `
      ```
````

duailibe:
Yeah that's such an edge case... we could simply always indent in that case. It's unlikely someone will eval the contents of the ` ```js ` to conflict with an user function

vjeux:
I've never really understood this piece of the code. @ikatyang you seem to have been thinking about this a lot and thought about all sort of edge cases, would you be willing to take point on this. If not, let's just merge it in as it solves a problem, I don't want not to make progress because no one knows this piece of the code :)

ikatyang:
Sorry! It seems `align(-Infinity, doc)` already did the same thing, we should document it...

duailibe:
Cool! I saw that but ot didn’t seem to work (I didn’t test it though)

PR #3689, opened 2018-01-09
Prettier extends and implements when not fitting in same line

thorbenvh8:
Please see screenshot for how it looks like:
<a href="https://ibb.co/f6FZnR">Screen_Shot_2018_01_09_at_8_40_09_AM</a>

This PR was requested by @j-f1 in the prettier-java project, after changes there. See discussion here: https://github.com/thorbenvh8/prettier-java/issues/13

Had to update some unrelated tests, by adding some generic fake parameters.

Please take a close look if the comment logic (changed test) makes still sense to you, for me it even makes it more clear what part of the line the comment is regarding too.

This change will make it nearly impossible to break at "typescript_as".


tests/typescript_class/extends_implements.ts
```diff
@@ -0,0 +1,20 @@
...
+  extends AAAAAAAAAAAAAAbstractClass
+  implements
+    Interface1,
+    Interface2,
+    Interface3,
+    Interface4,
+    Interface5,
+    Interface6,
+    Interface7,
+    Interface8 {}
```
azz: Could you add a blank line at the end here.

tests/class_comment/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -66,10 +67,11 @@ class A extends B /* a */ {
...
 
 (class A // comment 1
-// comment 2
-extends B {});
+  // comment 2
+  extends B {});
 
-(class A extends B // comment1
+(class A
+  extends B // comment1
```
azz: I don't think it should break just because there's a comment.

tests/classes/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -75,16 +74,18 @@ class Readable extends events.EventEmitter implements NodeJS_ReadableStream {}
 class InMemoryAppender extends log4javascript.Appender
   implements ICachedLogMessageProvider {}
 
-class Foo extends Immutable.Record({
-  ipaddress: ""
-}) {
+class Foo
+  extends Immutable.Record({
+    ipaddress: ""
+  }) {
```
vjeux: Fwiw, we had the behavior you are changing it to in the early days of prettier but I changed to what we have now for this example and extending React component with types. It looks better in one line.

Now, this causes a bunch of problematic things (like all the rest that’s being fixed by this pr) so it may not be a bad idea to follow through. I just wanted to give some context

azz: Could we inline only if it is a `CallExpression`?

vjeux: What could be a good way is to keep the previous implementation if there’s just a single extend. Otherwise use your new logic

vjeux: React.Component<...> is not a call expression

vjeux:
Btw this is great but is going to cause a bunch of formatting changes so we should hold it off after tomorrow’s release

tests/classes/__snapshots__/jsfmt.spec.js.snap
j-f1: We do want a long identifier to break, though.

tests/typescript_class/extends_implements.ts
thorbenvh8: What do you mean by this?

    Interface5,
    Interface6,
    Interface7,
    Interface8*added*
{}
    
    Interface5,
    Interface6,
    Interface7,
    Interface8 {}
*added*

tests/classes/__snapshots__/jsfmt.spec.js.snap
thorbenvh8: Once you decided when not to break (which expression), i will continue with this :)

src/main/comments.js
```diff
@@ -976,6 +976,10 @@ function printTrailingComment(commentPath, print, options) {
   }
   const isBlock = util.isBlockComment(comment);
 
+  // Workaround for comments that are in the superClass definition ( extends ...)
```
thorbenvh8: Is this workaround ok? 
Feels dirty to me but there was no other way to get the parentNode of the parentNode

thorbenvh8: Is this workaround ok?
Feels dirty to me but there was no other way to get the parentNode of the parentNode

thorbenvh8: @azz

src/main/comments.js
```diff
@@ -976,6 +976,10 @@ function printTrailingComment(commentPath, print, options) {
...
+  // We don't want the line to break by default when its a comment 
+  const isParentSuperClass = commentPath.stack[commentPath.stack.length-6] === "superClass";
```
j-f1: `commentPath.getParentNode(3) === "superClass"`?

src/main/comments.js
j-f1: `commentPath.getParentNode(3) === "superClass"`?

or maybe it’s `2`.

thorbenvh8: i fixed it, its a little bit longer but better understandable and maintainable :)
We should definitely work on some kind of documentation :) There are cool features build in, but its hard for people who are new to find them :)

thorbenvh8: thanks for the quick feedback :)

tests/typescript_class/extends_implements.ts
thorbenvh8: @azz please check :)

azz: Files should have a blank line at the end of them so the snapshots look right.

thorbenvh8: done :)

tests/classes/__snapshots__/jsfmt.spec.js.snap
thorbenvh8: @azz @vjeux @j-f1 
Any update on this?
I mean if its a long statement it should def brake, and it makes it def more readable like this.
If we limit it to only break when there is a single extends
something like this 
class AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA extends AAAAAAAAAAAAAA
would still look shitty.

vjeux:
My ask is that if there's only a single extend then use the previous behavior. If there are more than one, feel free to change it however you like :)

vjeux:
Thanks, the examples on the last commit look good on my end :)

thorbenvh8:
@vjeux its now like you want it, only one extends is old behaviour :)
@azz can you check and merge? :)

src/main/comments.js
```diff
@@ -1001,11 +1012,13 @@ function printTrailingComment(commentPath, print, options) {
...
       concat([hardline, isLineBeforeEmpty ? hardline : "", contents])
     );
-  } else if (isBlock) {
+  } else if (isBlock || isParentSuperClass) {
     // Trailing block comments never need a newline
+    // console.log("trailing block comment");
     return concat([" ", contents]);
   }
 
+  // console.log("trailing comment");
```
azz: Could you delete these comments

thorbenvh8: oh man that's embarassing...
Fixed now :)

thorbenvh8:
@azz fixed :) finally done? :)

azz:
:tada: Thanks!

Issue #3667, opened 2018-01-07
Add playground examples


ikatyang:
Add examples to `getExample()`:

https://github.com/prettier/prettier/blob/5e2e7a2881d07ff0aa4b4673becc9cb4ce2f6cae/website/static/playground.js#L356-L381

- [x] JavaScript (babylon)
- [ ] JavaScript (flow)
- [x] TypeScript (typescript)
- [x] CSS (css)
- [ ] Less (less)
- [x] SCSS (scss)
- [x] JSON (json)
- [ ] GraphQL (graphql)
- [ ] Markdown (markdown)

arefaslani:
I have created a Next boilerplate that support importing scss, css, preconfigured with ESlint and prettier and many more features. Take a look at it and if you liked it, give it a star ;)
https://arefaslani.github.io/next-boilerplate

PR #3696, opened 2018-01-10
playground: add (s)css examples

Haroenv:
see #3667

These are derived from Bootstrap, should there be a license comment?

### DIFF
--- a/website/static/playground.js
+++ b/website/static/playground.js
@@ -407,7 +407,7 @@ function getExample(parser) {
         "  .bd-examples {margin-right: -.75rem;margin-left: -.75rem",
         "  }",
         "  ",
-        " .bd-examples>[class^="col-"]  {",
+        " .bd-examples>[class^=\"col-\"]  {",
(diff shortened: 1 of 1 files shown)


Haroenv:
It's MIT, so I'm not 100% sure if this is allowed without license text. It's such a small block that it doesn't matter I guess though 🤔  https://getbootstrap.com/docs/4.0/about/license/

azz:
Yeah I really don't think it matters for tiny extracts. I guess you could put a comment in the code linking to bootstrap?

j-f1:
What about putting something like this in the JS?

```js
// Excerpted from the Bootstrap source, which is licensed under the MIT license:
// https://github.com/twbs/bootstrap/blob/v4.0.0-beta.3/LICENSE
```

Haroenv:
done now (sorry for earlier mistakes, commited on GitHub so didn't see the linting errors)

Haroenv:
awesome 👌

PR #3712, opened 2018-01-11
Re-write Rollup Build Scripts

### LINKED ISSUE #3574: Revamp Build System
Prettier's build system is not pretty.

We support both install from both npm (`npm i prettier`) and GitHub (`npm i prettier/prettier`).

The structure of the npm production build is:

```
bin/
  prettier.js
index.js
third-party.js
parser-babylon.js
parser-flow.js
parser-....js
```

Parsers are packaged separately because they can be expensive to load into the browser or Node.js runtime.

#3296 also raises that it is non-trivial to load Prettier in the browser at all.

We use Rollup to create the bundles above. There is actually a copy of most of Prettier's source code in `index.js` and `bin/prettier.js` as they are separate entries without a common package.

We should look at a packages approach to creating bundles, and support UMD.

### LINKED ISSUE #3698: Improve distribution tests
`yarn test:dist` runs the tests against the built JS files, but it didn't catch an issue in 1.10 where a file with path `../../index` couldn't be resolved, because in the `dist` directory, `../../index` resolves to the original source. We could fix this by copying the files to a tempdir before running the tests.

It would be nice to test the actual `.tgz` file, too.

https://github.com/prettier/prettier/commit/cf511ec2e973ed6d794c5b99321f2b3a3458a919

#3697

azz:
Idea is to remove all the `eval("require")` calls, and replace them with Rollup `paths`/`externals`. 

Currently the `dist/` directory looks like this:

<details>

```shellsession
$ tree dist
dist
├── README.md
├── bin
│   └── prettier.js
├── externals
│   ├── babylon.js
│   ├── cosmiconfig.js
│   ├── flow-parser.js
│   ├── get-stream.js
│   ├── glimmer-syntax.js
│   ├── graphql-language.js
│   ├── parse5.js
│   ├── postcss-less.js
│   ├── postcss-media-query-parser.js
│   ├── postcss-scss.js
│   ├── postcss-selector-parser.js
│   ├── postcss-values-parser.js
│   ├── read-pkg-up.js
│   ├── remark-parse.js
│   └── typescript-eslint-parser.js
├── index.js
├── package.json
└── src
    ├── cli
    │   ├── constant.js
    │   ├── index.js
    │   ├── logger.js
    │   ├── util.js
    │   └── validator.js
    ├── common
    │   ├── clean-ast.js
    │   ├── errors.js
    │   ├── fast-path.js
    │   ├── load-plugins.js
    │   ├── parser-create-error.js
    │   ├── parser-include-shebang.js
    │   ├── support.js
    │   └── util.js
    ├── config
    │   ├── resolve-config-editorconfig.js
    │   └── resolve-config.js
    ├── doc
    │   ├── doc-builders.js
    │   ├── doc-debug.js
    │   ├── doc-printer.js
    │   ├── doc-utils.js
    │   └── index.js
    ├── language-css
    │   ├── clean.js
    │   ├── index.js
    │   ├── options.js
    │   ├── parser-postcss.js
    │   └── printer-postcss.js
    ├── language-graphql
    │   ├── index.js
    │   ├── options.js
    │   ├── parser-graphql.js
    │   └── printer-graphql.js
    ├── language-handlebars
    │   ├── index.js
    │   ├── parser-glimmer.js
    │   └── printer-glimmer.js
    ├── language-html
    │   ├── embed.js
    │   ├── index.js
    │   ├── parser-parse5.js
    │   └── printer-htmlparser2.js
    ├── language-js
    │   ├── clean.js
    │   ├── embed.js
    │   ├── index.js
    │   ├── options.js
    │   ├── parser-babylon.js
    │   ├── parser-flow.js
    │   ├── parser-typescript.js
    │   └── printer-estree.js
    ├── language-markdown
    │   ├── embed.js
    │   ├── index.js
    │   ├── options.js
    │   ├── parser-markdown.js
    │   └── printer-markdown.js
    ├── language-vue
    │   ├── embed.js
    │   ├── index.js
    │   ├── parser-vue.js
    │   └── printer-vue.js
    └── main
        ├── ast-to-doc.js
        ├── comments.js
        ├── deprecated.js
        ├── get-printer.js
        ├── multiparser.js
        ├── options.js
        └── parser.js

15 directories, 79 files
```

</details>

At the moment the dynamic requires are not being preserved by Rollup. I wonder if it is even _possible_ to preserve them due to the nature of Rollup. Maybe Webpack might be a better for the `index.js` file.

❓I wanted to get some feedback on whether this approach makes sense before sinking too much time into it.

Fixes #3574 
Fixes #3698


azz:
Another option is to ship the `src/` files as-is (but minified), then string replace `require("babylon")` with `require("./externals/babylon")`, etc.

duailibe:
Well I guess it comes down to testing if this is a performance regression or not

suchipi:
I would advise against using webpack for the bundle, because I've seen some strange issues occur when nesting webpack module wrappers (which is what happens when you pull a webpack-ified file into a webpack build). That could cause some pain for people using prettier in the browser.

azz:
I'm thinking maybe:

* Use rollup for individual external bundles.
* Use babel to copy all the internal files, using [`babel-plugin-rewrite-require`](https://www.npmjs.com/package/babel-plugin-rewrite-require) or [`babel-plugin-replace-require`](https://www.npmjs.com/package/babel-plugin-replace-require) to alias the requires to externals.

This way we could support:

* NodeJS v4 via`yarn add prettier` (npm)
* NodeJS v6+ via `yarn add prettier/prettier` (GitHub)

azz:
I think I like this approach so far. Implementation looks _much_ cleaner than before.

azz:
Has anyone used [`bundledDependencies`](http://npm.github.io/using-pkgs-docs/package-json/types/bundleddependencies.html) before? I find the docs very confusing. Is it supposed to pack the bundled dependencies for you? Do you still list the dependency in `dependencies`?

j-f1:
> Is it supposed to pack the bundled dependencies for you?

It seems like it includes them when you run `npm pack`.


> 

> Do you still list the dependency in `dependencies`?

I’d assume so — `bundledDependencies` looks like it’s only used when you run `npm pack`. Also, since you can’t specify a version, there would be no way for npm to know which version to install.

azz:
I'm doing a clean up of stale PRs, closing this one.

I didn't get where I wanted to get to with this and ran out of steam. Maybe I'll try again soon but the current system works ok for the moment.

PR #3715, opened 2018-01-11
Enable PostCSS in VUE SFC

DanielHeckrath:
This PR enables the css parser for style blocks that have the lang attribute set to postcss

### DIFF
--- a/src/language-vue/embed.js
+++ b/src/language-vue/embed.js
@@ -15,14 +15,12 @@ function embed(path, print, textToDoc, options) {
 
   if (node.tag === "style") {
     const langAttr = node.attrs.find(attr => attr.name === "lang");
-    if (!langAttr) {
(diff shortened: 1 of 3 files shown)


src/language-vue/embed.js
```diff
@@ -21,6 +21,8 @@ function embed(path, print, textToDoc, options) {
       parser = "scss";
     } else if (langAttr.value === "less") {
       parser = "less";
+    } else if (langAttr.value === "postcss") {
```
duailibe: Maybe move this to above?

```js
if (!langAttr || langAttr.value === "postcss") {
```

DanielHeckrath:
@azz I've added a style block that verifies that the css parser is used on style blocks with lang="postcss" to the multiparser_vue. Not sure if that is the correct location for that test.

azz:
Thanks!

Issue #2799, opened 2017-09-13
wrong indentation: SCSS map functions with !default; 


liubiantao:
**Prettier Version:** 1.6.1 / master

**Code**

```css
$theme-checkbox-colors: (
  default: $theme-color-border,
  checked: $theme-color-checked,
  disabled: $theme-color-disabled,
  disabled-font: $theme-color-font-secondary,
) !default;
```

**Expected behavior:**

```css
$theme-checkbox-colors: (
  default: $theme-color-border,
  checked: $theme-color-checked,
  disabled: $theme-color-disabled,
  disabled-font: $theme-color-font-secondary,
) !default;
```

**Actual behavior:**

```css
$theme-checkbox-colors: (
    default: $theme-color-border,
    checked: $theme-color-checked,
    disabled: $theme-color-disabled,
    disabled-font: $theme-color-font-secondary
  )
  !default;
```

and stylelint will throw this error:
```
 38:3  ✖  Expected single space before "!"   declaration-bang-space-before
 42:5  ✖  Expected indentation of 2 spaces   indentation
```

I also found the usage in bootstrap:

```css
$grid-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
) !default;
```
> https://github.com/twbs/bootstrap/blob/f7eb3c98b261930270c2f65c0e565c454d6cb647/scss/_variables.scss#L145-L151

lydell:
Given that it formats like the expected case when you remove `!default` (or change to `!important`), I consider this a bug.

JasonMSims:
I've stumbled onto this same error as well.

hudochenkov:
In similar situation valid CSS with `!important` works correctly.

**Prettier 1.10.2**
[...]

PR #3724, opened 2018-01-11
fix(scss): `!default` directive

evilebottnawi:
`postcss` add `node.important` when you use `!important`, but don't handle `!default`. 

Just remove `!default` from `value` and add `node.default` to detect when you use `!default`.

Issue: https://github.com/prettier/prettier/issues/2799


evilebottnawi:
@j-f1 done :+1:

azz:
Thanks!

PR #3730, opened 2018-01-12
Run production tests on npm package

### LINKED ISSUE #3698: Improve distribution tests

azz:
Runs the production tests agains the result of actually installing the package, rather than from the `dist` directory. 

Extracted from #3712
Fixes #3698

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,10 @@ node_js:
 cache:
   yarn: true
   directories:
-  - node_modules
+    - node_modules
 env:
   - NODE_ENV=development
   - NODE_ENV=production
-matrix:
-  fast_finish: true
 install:
   - NODE_ENV=development yarn install
 before_script:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -6,12 +6,10 @@ node_js:
 cache:
   yarn: true
   directories:
-  - node_modules
+    - node_modules
```
ikatyang: 😉

.travis.yml
```diff
@@ -6,12 +6,10 @@ node_js:
...
 env:
   - NODE_ENV=development
   - NODE_ENV=production
-matrix:
-  fast_finish: true
```
ikatyang: ❓

josephfrazier: Yeah, shouldn't we leave this, in order to learn about build failures faster? From https://blog.travis-ci.com/2013-11-27-fast-finishing-builds:

> With fast finishing enabled, Travis CI will mark your build as finished as soon as one of two conditions are met: The only remaining jobs are allowed to fail, or a job has already failed.

We don't have any [jobs allowed to fail](https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail), so it should be safe to leave `fast_finish` enabled.

azz: Ah, I was under the impression it only applied when jobs were allowed to fail. Reverted.

duailibe:
How likely it is that a build will pass in production but fail in development?

azz:
_Shouldn't_ be possible, but I don't know for sure.

PR #3735, opened 2018-01-12
Add support for skip() (#3706)

### LINKED ISSUE #3706: Print Width Whitelist should be optional
or at the very least, whitelisted methods should be exposed through config.

I would LIKE my testing methods to be prettier-fied. Also, since 'skip' is not one of the whitelisted words, when my bot goes in to auto-skip flakey tests, all of a sudden I have tons of prettier lint errors.

tmquinn:
Fixes #3706

https://api.qunitjs.com/QUnit/skip

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5121,7 +5121,7 @@ function isObjectType(n) {
 
 // eg; `describe("some string", (done) => {})`
 function isTestCall(n) {
-  const unitTestRe = /^(f|x)?(it|describe|test|skip)$/;
+  const unitTestRe = /^(skip|(f|x)?(it|describe|test))$/;
   return (
     ((n.callee.type === "Identifier" && unitTestRe.test(n.callee.name)) ||
       (n.callee.type === "MemberExpression" &&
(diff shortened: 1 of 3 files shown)


src/language-js/printer-estree.js
```diff
@@ -5121,7 +5121,7 @@ function isObjectType(n) {
 
 // eg; `describe("some string", (done) => {})`
 function isTestCall(n) {
-  const unitTestRe = /^(f|x)?(it|describe|test)$/;
+  const unitTestRe = /^(f|x)?(it|describe|test|skip)$/;
```
suchipi: Thanks for making a PR! This will also match `fskip` and `xskip`; could you adjust the regexp so it won't? I think `/^(skip|(f|x)?(it|describe|test))$/` should do it.

tests/test_declarations/test_declarations.js
```diff
@@ -85,6 +85,8 @@ ftest("does something really long and complicated so I have to write a very long
 
 xtest("does something really long and complicated so I have to write a very long name for the describe block", () => {});
 
+skip(`does something really long and complicated so I have to write a very long name for the test`, () => {});
```
j-f1: Could you also add a test with a double-quoted string?

tmquinn:
Thanks again for your prompt attention on this!

azz:
Thanks!

Issue #3671, opened 2018-01-08
Ignore package(-lock).json by default


azz:
@duailibe EDIT: see before implementing https://github.com/prettier/prettier/issues/3671#issuecomment-355933794

---

Proposal: If no `.prettierignore` file is found, default to the contents:

```
package.json
```

If an empty `.prettierignore` file is provided the behaviour is revoked.

Rationale: Every project I have created recently has a `.prettierignore` containing `package.json`, and I can't think of a reason why you _would_ want Prettier to format this file, as it is aggressively reformatted by other tools.

Disadvantage is that it creates a bit of hidden functionality. Note that there is _some_ precedent, globs automatically ignore `node_modules`.

👍 / 👎 ❓

SimenB:
>Rationale: Every project I have created recently has a `.prettierignore` containing `package.json`

Same here, +1 to this! They also contain `package-lock.json` for the same reason.

lydell:
Let’s say we add this feature, and a user does not add a .prettierignore. One day the user needs to ignore a file, so the user creates a .prettierignore with that file in it. Suddenly Prettier starts to format the user’s package.json. Is that unexpected?

j-f1:
Why not ignore it by default, like `node_modules` is? If you want it formatted for some reason, you can put `!package.json` in your `.prettierignore`.

duailibe:
I'm :-1: with the suggestion as is, but :+1: for the idea of ignoring `package(-lock).json`.

@j-f1's idea seems good.

azz:
Yeah I like @j-f1's alternative more. Anyone want to implement this?

duailibe:
And when should we run prettier on the file? `--with-package-json` ?

SimenB:
What about copying eslint's approach when you try to lint ignored files? `0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override`

azz:
There is an issue though... As prettier doesn't have a `prettier.isFileIgnored(path)` API, the changes would only affect the CLI, and not any editor integrations or other tools which are using `ignore` directly.

duailibe:
@SimenB that makes sense as a feature (we can do this now with `--ignore-path /dev/null` or something similar) but it's not related here since `node_modules` is ignored in addition do the ignore file.

@azz editors should be using the `filepath` with the API, right? We could ignore it then

lipis:
[...]

Issue #2882, opened 2017-09-23
Noooooo my .git is dead! / Ignore dotfiles?


testerez:
I was just trying prettier to see the result on my codebase. I should have been more cautious but I scrapped my `.git` directory. All my local branches are probably lost forever!

You should consider adding a check to prevent altering any file with `.git` in the path.

😢

testerez:
Actually, it seems like it was just my `.git/HEAD` that was scrapped. Fixed it and I can see my branches again 😅 
Still I think that securing `.git` directory would be a good idea.

lydell:
I don't understand what happened, and what you're proposing that Prettier should do?

testerez:
Ok, here is how to reproduce my issue:

```
mkdir test
cd test
git init
npx prettier --write "**/*"
git status
```
Result: `fatal: Not a git repository (or any of the parent directories): .git`

What I suggest is that prettier ignore any file in a `.git` directory.

azz:
What about other version control systems such as mercurial (`.hg/`) and subversion (`.svn/`)? Maybe ignore dot-directories and dotfiles by default?

Alternatively we could remove the default parser and rely entirely on extension or `--parser` option, this way `.git/HEAD` will not match a parser and won't be modified, this might solve a more general case.

But you probably shouldn't be running prettier on a glob without extension(s).

lydell:
@testerez Thanks for the steps to reproduce!

For reference, here are the modifications in a fresh git repo:

.git/HEAD
```diff
-ref: refs/heads/master
+ref: refs / heads / master;
```

.git/config:
```diff
-[core]
-	repositoryformatversion = 0
-	filemode = true
-	bare = false
-	logallrefupdates = true
+[core];
+repositoryformatversion = 0;
+filemode = true;
+bare = false;
+logallrefupdates = true;
```

testerez:
@azz I tried this command after reading this in the doc: "Prettier automatically infers the parser from the input file path"
You're probably right that I shouldn't have run it that way but I think that it can be pretty common mistake.

Thanks for the diff @lydell

lydell:
> Prettier automatically infers the parser from the input file path

Yeah, that could be considered a bit misleading since Prettier falls back to treating everything as JS if it can't infer the type of the file.

ikatyang:
This should be fixed by #4528.

PR #2720, opened 2017-08-31
Add support for multiple --ignore-path options

tgriesser:
Thought this might be useful, tried to use both `.gitignore` and `.prettierignore` and saw it didn't work.

PR #3737, opened 2018-01-13
Added support to specify more than one `ignore-path` entry

maoueh:
Previously, passing multiple ignore paths was not something possible. So if one wanted to use its `.gitignore` file as well as another file defining extra ignored files that should however not be part of the `.gitignore` file, then it was not possible to do so.

Now, the --ignore-path CLI option accepts being specified multiple times to use multiple ignore paths.

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -269,11 +269,25 @@ function formatStdin(argv) {
 }
 
 function createIgnorer(argv) {
-  const ignoreFilePath = path.resolve(argv["ignore-path"]);
-  let ignoreText = "";
+  let ignorePaths = argv["ignore-path"];
+  if (!Array.isArray(ignorePaths)) {
+    ignorePaths = [ignorePaths];
+  }
+
+  return ignorePaths
+    .map(readIgnoreFile)
+    .reduce(
+      (ignorer, ignoreFileContent) => ignorer.add(ignoreFileContent),
+      ignore()
+    );
+}
+
+function readIgnoreFile(ignorePath) {
+  const ignoreFilePath = path.resolve(ignorePath);
+  let ignoreFileContent = "";
 
   try {
-    ignoreText = fs.readFileSync(ignoreFilePath, "utf8");
+    ignoreFileContent = fs.readFileSync(ignoreFilePath, "utf8");
   } catch (readError) {
     if (readError.code !== "ENOENT") {
       logger.error(`Unable to read ${ignoreFilePath}: ` + readError.message);
@@ -281,7 +295,7 @@ function createIgnorer(argv) {
     }
   }
 
-  return ignore().add(ignoreText);
+  return ignoreFileContent;
 }
 
 function eachFilename(argv, patterns, callback) {
--- a/docs/ignore.md
+++ b/docs/ignore.md
@@ -9,6 +9,13 @@ Prettier offers an escape hatch to ignore a block of code or prevent entire file
 
 To exclude files from formatting, add entries to a `.prettierignore` file in the project root or set the [`--ignore-path` CLI option](cli.md#ignore-path).
 
+Multiple `--ignore-path` directives can be specified to ignore entries from multiple files like from
+your `.gitignore` file also:
+
+```
+prettier --ignore-path .gitignore --ignore-path .prettierignore
+```
+
 ## JavaScript
 
 A JavaScript comment of `// prettier-ignore` will exclude the next node in the abstract syntax tree from formatting.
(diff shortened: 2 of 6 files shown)


src/cli/util.js
```diff
@@ -269,7 +269,21 @@ function formatStdin(argv) {
...
+
+  const childIgnores = ignorePaths.map(createChildIgnore);
+
+  return childIgnores.reduce(
+    (parentIgnorer, childIgnorer) => parentIgnorer.add(childIgnorer),
+    ignore()
+  );
+}
+
+function createChildIgnore(ignorePath) {
```
duailibe: Does this work? Rename this function to `readIgnoreFile` and simply `return ignoreText`

Would make it more straight forward I believe.. but if it doesn't work it's fine =)

maoueh: When I read `ignore` library's README, I was under the impression it would not work but I did not try, following what they were saying about combining multiple `ignorer` together.

Let me try, I will report back.

maoueh: Yeah it works, modified code to use it now and renamed `ignoreText` to `ignoreFileContent` as I felt it was a better name, don't hesitate to tell if you want me to revert that change.

CiGit:
It seems to me this implementation makes `--ignore-path` order significant.
Is this expected?

maoueh:
@CiGit I guess you mean order between each `--ignore-path` entries.

You are right about that. As a wild guess, it's not a problem in 99% of the cases as people tend to only ignore files (not using the not (`!`) operator) which make stuff additive and as such not a problem.

It's only a problem if you ignore a certain path and then not ignore it in another file and would like to make it work properly:

```
# .gitignore
*.md

# .prettierignore
!CHANGELOG.md
```

in this case, only `--ignore-path .gitignore --ignore-path .prettierignore` would work. Even that, don't think it's a problem. We could mention the ordering stuff in the documentation however.

duailibe:
@CiGit I don't think there's an alternative for that if you want to support to multiple ignore paths. You either only support one or you support multiple and order is significant.

CiGit:
Thanks for your answer @maoueh, I also think it wouldn't be an issue in most cases.

@duailibe An alternative could be to test per file instead of concatenating them. (Is it ignored in one of them)

duailibe:
@CiGit ah yes, sorry you're right about that

maoueh:
@duailibe ESLint is having such feature. I started implementing it because I wanted to check that all files in my project were correctly formatted using prettier with the `-l` option.

So, I started with simply `prettier  '**/*'` but it was picking some files that were git ignored so I did `prettier --ignore-path .gitignore '**/*'`. But then, I realized it was still choking on some files like my `yarn.lock` file and that passing multiple `--ignore-path` was not possible while it exactly how I do it in ESLint.

While implementing the feature, I realized I could use file extension globbing to reduce the scope of prettier like `**/*.{js, json, md}`. Since the feature was mostly complete, I thought it would still be good to open a PR.

@CiGit That would prevent using the `!something` however as it would become a full OR. `!CHANGELOG.md` ignored? no, continue with `.gitignore`, `*.md`, ignored? yes. Not sure it's that useful however as it would mean somehow you git ignore a file but still pretty-format it with Prettier.

Tell me if you would like the feature or not.

ikatyang:
For reference, there's already a PR #2720 for this thing but closed due to inactivity.

maoueh:
@ikatyang Oups :$ Seems I missed it completely, I looked only at page 1 of search results while it showed up in page 2. Sorry if it caused any problem to anyone.

ikatyang:
No worry, just said that we probably won't reject this feature based on comments in that PR.

azz:
Don't use `**/*` with prettier. #2882

azz:
Closing this as stale. If someone (including yourself, @maoueh) wants to pick it up and implement https://github.com/prettier/prettier/issues/3671#issuecomment-355984739 then we will consider merging.

maoueh:
@azz I've added an initial draft of a possible ignore file API for prettier in this very branch. I could split the current PR in two if you prefer, one for the ignore API and a second one that enable multiple ignore paths.

Can we re-open the PR?

maoueh:
It seems GitHub does not update a close PR, I'll split the PR in two right now instead.

duailibe:
@maoueh Github won't let us reopen because the branch has been force-pushed or recreated

j-f1:
@maoueh If you reset the branch on your fork to https://github.com/prettier/prettier/pull/3737/commits/fdef738edb1f78d34e7f64410199a41b85d8853e, I might be able to re-open the PR, and then you should be able to push up the new commits.

maoueh:
I opened #4221 that contains only the API changes for now to discuss this part. Once finish, I'll re-open this very PR and rebase it on top of API implementation.

PR #3739, opened 2018-01-13
Use an SVG <object> tag to allow the Twitter button to be interactive

j-f1:
The interactive button links you to a page to follow the user directly, so I also added a direct link to the profile.


lipis:
We need to get the [`@prettier`](https://twitter.com/prettier) handle on Twitter.. :) Anyone from @Twitter?

PR #3277, opened 2017-11-15
fix(javascript): ternary with `--use-tabs`

ikatyang:
Fixes #2771

image

EDIT: After second thought, spaces should only be applied to the trailing parentheses, marking as WIP.

<details><summary>ideal v2</summary></details>

<details><summary><del>ideal</del></summary></details>

<details><summary><del>old</del></summary></details>

PR #1026, opened 2017-03-16
[RFC] Indent with tabs

rhengles:
I'm really sorry for having to create another PR, don't want to spam this project, but Github doesn't allow me to reopen the previous PR #920 because the branch has been rebased while the PR was closed.

PR #3745, opened 2018-01-15
fix(javascript): ternary with `--use-tabs`

### LINKED ISSUE #2771: Objects in ternary has extra indent when using --use-tabs
**Prettier Version:** 1.6.1 

**Code**

([playground](https://prettier.io/playground/#%7B%22content%22%3A%22const%20abc%20%3D%20props.something%5Cn%20%20%20%20%3F%20xyz%5Cn%20%20%20%20%3A%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20prop1%3A%20a%2C%5Cn%20%20%20%20%20%20%20%20%20%20prop2%3A%20b%2C%5Cn%20%20%20%20%20%20%20%20%20%20prop3%3A%20false%2C%5Cn%20%20%20%20%20%20%7D%3B%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A4%2C%22singleQuote%22%3Atrue%2C%22trailingComma%22%3A%22all%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Atrue%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D))

```js
const abc = props.something
    ? xyz
    : {
          prop1: a,
          prop2: b,
          prop3: false,
      };

```

**Expected behavior:**

The object in the ternary has the correct indentation as above JS snippet

**Actual behavior:**

We get an extra level of indentation in the object. This seems to only happen if using the `--use-tabs` option. If spaces are used, there is no issue.

```js
const abc = props.something
	? xyz
	: {
			prop1: a,
			prop2: b,
			prop3: false,
		};
```

ikatyang:
Rewrite #3277, fixes #2771.

Rules:
- trailing alignments:
  - still spaces.
- ~~middle alignments:~~ Removed, see https://github.com/prettier/prettier/pull/3745#issuecomment-357953038.
  - ~~every `tabsWidth` spaces will be transformed into a tab.~~

Example:
- before
  - `tabWidth = 2`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><tab>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><tab>`
  - `tabWidth = 4`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><tab>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><tab>`
- after
  - `tabWidth = 2`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><2 space>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><2 space>`
  - `tabWidth = 4`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><2 space>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><2 space>`

P.S. `align` with string `n` is not affected.

Diff:

`{ useTabs: true, tabWidth: 2 | 4 }`
image

---

**Prettier pr-3745**
[Playground link](https://deploy-preview-3745--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBDAOlABDrB+LYTXErVJLABmNIF8acKjtSdzCHWT3qWud6ffvg5DhuHp3GDxjUbNIBPJlIUSKvNbhlbSBZruGSxunYbkHzrY1YGrWZ7SAA0ICAAcYAS2gBnZKDuAE5esADqXgAmMAAWyAAclK4wqABGEdFxSAAsrr6hAOYANnAAigCuEPDIAGaoRb5wyUGoXkWFAMIQALbdqMggUNBwLiCpLWAA1nAwAMruqGCFyDBB5U0gAFa+AB4AQhPTc6jdcAAyoSNIdQ0bC0GNQQOpaYpF0KON3V4raxvljQAKml-EhVutXKFHjAAAotAp9Wr1RquIJwACO5S8aLhqAR-WuyLuQQgjTCLXcA2CcEeADcRq5UEESQB3GFMhCgtC0iBRUaRCBgJG3Rm+GDClFucowdzSgBMErgtFoQA)
```sh
--tab-width 4
--use-tabs
```

**Input:**
```jsx
a
    ? {
        a: 0
      }
    : {
        a: {
             a: 0
           }
            ? {
                a: 0
              }
            : {
                y: {
                    a: 0
                }
                    ? {
                        a: 0
                    }
                    : {
                        a: 0
                    }
            }
      }
```

**Output:**
```jsx
a
	? {
			a: 0
	  }
	: {
			a: {
				a: 0
			}
				? {
						a: 0
				  }
				: {
						y: {
							a: 0
						}
							? {
									a: 0
							  }
							: {
									a: 0
							  }
				  }
	  };

```

---

<details>
<summary>old</summary>

- before
  - `tabWidth = 2`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><tab>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><tab>`
  - `tabWidth = 4`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><tab>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><tab>`
- after
  - `tabWidth = 2`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><tab><2 space>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><tab><2 space>`
  - `tabWidth = 4`
    - `<tab><2 space><tab><2 space>` -> `<tab><tab><2 space>`
    - `<tab><4 space><tab><2 space>` -> `<tab><tab><tab><2 space>`

Diff:

`{ useTabs: true, tabWidth: 2 }`
image

`{ useTabs: true, tabWidth: 4 }`
image

---

**Prettier pr-3745**
[Playground link](https://deploy-preview-3745--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBDAOlABDrB+LYTXErVJLABmNIF8acKjtSdzCHWT3qWud6ffvg5DhuHp3GDxjUbNIBPJlIUSKvNbhlbSBZruGSxunYbkHzrY1YGrWZ7SAA0ICAAcYAS2gBnZKCoAE5BEADuAArBCP4oqABuEF4AJi5ovjDIAGaoADa+cK4ARkGoYADWcDAAyu5lXlAA5sgwQQCuhSDJEGDZeQWuDQVBMBGljQC2qH35nQBWvgAeAEKlFVXVqBNwADINcDMDbm0w7icATIeddUHDyCBFqEWKudBp7kENMADqKTAAFsgAByUVwfCAFb6ldz3D5wYbxA6uIJwACObS8KLGqEm0yQOVmrgKEy8LXanV8DUauTgAEU2hB4FdXDAnr9kgDkAAWFmlLy5KkAYQgEym9yg0CRIDaBQAKk9Yq0OrRaEA)
```sh
--tab-width 4
--use-tabs
```

**Input:**
```jsx
a
    ? {
        a: 0
      }
    : {
        a: {
             a: 0
           }
            ? {
                a: 0
              }
            : {
                y: {
                    a: 0
                }
                    ? {
                        a: 0
                    }
                    : {
                        a: 0
                    }
            }
      }
```

**Output:**
```jsx
a
	? {
		a: 0
	  }
	: {
		a: {
			a: 0
		}
			? {
				a: 0
			  }
			: {
				y: {
					a: 0
				}
					? {
						a: 0
					  }
					: {
						a: 0
					  }
			  }
	  };

```

</details>

### DIFF
--- a/tests/ternaries/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/ternaries/__snapshots__/jsfmt.spec.js.snap
@@ -346,15 +346,15 @@ a
 aaaaaaaaaaaaaaa
 	? bbbbbbbbbbbbbbbbbb
 	: ccccccccccccccc
-	  ? ddddddddddddddd
-	  : eeeeeeeeeeeeeee ? fffffffffffffff : gggggggggggggggg;
+		? ddddddddddddddd
+		: eeeeeeeeeeeeeee ? fffffffffffffff : gggggggggggggggg;
 
 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	  ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	    ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	    : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	  : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+		? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+			? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+			: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+		: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
 
 a
@@ -427,15 +427,15 @@ a
 aaaaaaaaaaaaaaa
 	? bbbbbbbbbbbbbbbbbb
 	: ccccccccccccccc
-	    ? ddddddddddddddd
-	    : eeeeeeeeeeeeeee ? fffffffffffffff : gggggggggggggggg;
+		? ddddddddddddddd
+		: eeeeeeeeeeeeeee ? fffffffffffffff : gggggggggggggggg;
 
 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	    ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	        ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	        : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	    : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+		? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+			? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+			: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+		: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
 
 a
--- a/tests/ternaries/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/ternaries/__snapshots__/jsfmt.spec.js.snap
@@ -440,26 +440,26 @@ aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 
 a
 	? {
-		a: 0
+			a: 0
 	  }
 	: {
-		a: {
-			a: 0
-		}
-			? {
+			a: {
 				a: 0
-			  }
-			: {
-				y: {
-					a: 0
-				}
-					? {
+			}
+				? {
 						a: 0
-					  }
-					: {
-						a: 0
-					  }
-			  }
+				  }
+				: {
+						y: {
+							a: 0
+						}
+							? {
+									a: 0
+							  }
+							: {
+									a: 0
+							  }
+				  }
 	  };
 
 `;
@@ -609,10 +609,10 @@ debug
 a =>
 	a
 		? () => {
-			a;
+				a;
 		  }
 		: () => {
-			a;
+				a;
 		  };
 a => (a ? a : a);
 a =>
@@ -801,29 +801,29 @@ const obj2 = conditionIsTruthy
 
 const obj3 = conditionIsTruthy
 	? {
-		some: "eeeeeeeeeeeeven looooooooooooooooooooooooooooooonger",
-		object: "with",
-		lots: "of",
-		stuff
+			some: "eeeeeeeeeeeeven looooooooooooooooooooooooooooooonger",
+			object: "with",
+			lots: "of",
+			stuff
 	  }
 	: shortThing;
 
 const obj4 = conditionIsTruthy
 	? shortThing
 	: {
-		some: "eeeeeeeeeeeeven looooooooooooooooooooooooooooooonger",
-		object: "with",
-		lots: "of",
-		stuff
+			some: "eeeeeeeeeeeeven looooooooooooooooooooooooooooooonger",
+			object: "with",
+			lots: "of",
+			stuff
 	  };
 
 const obj5 = conditionIsTruthy
 	? { some: "long", object: "with", lots: "of", stuff }
 	: {
-		some: "eeeeeeeeeeeeven looooooooooooooooooooooooooooooonger",
-		object: "with",
-		lots: "of",
-		stuff
+			some: "eeeeeeeeeeeeven looooooooooooooooooooooooooooooonger",
+			object: "with",
+			lots: "of",
+			stuff
 	  };
 
 `;
--- a/tests/ternaries/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/ternaries/__snapshots__/jsfmt.spec.js.snap
@@ -427,15 +427,15 @@ a
 aaaaaaaaaaaaaaa
 	? bbbbbbbbbbbbbbbbbb
 	: ccccccccccccccc
-	  ? ddddddddddddddd
-	  : eeeeeeeeeeeeeee ? fffffffffffffff : gggggggggggggggg;
+	    ? ddddddddddddddd
+	    : eeeeeeeeeeeeeee ? fffffffffffffff : gggggggggggggggg;
 
 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	  ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	    ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+	        ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+	        : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	    : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-	  : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
 
 a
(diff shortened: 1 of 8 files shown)


ikatyang:
Ref: previous discussion https://github.com/prettier/prettier/pull/3277#issuecomment-344665329

> We have to read through #1026 (the PR that added tabs support) again, because as far as I remember all of this was discussed back then.
> 
> When using spaces, Prettier indents with 2 spaces + 1 indent level. When using tabs, there are three ways to go:
> 
> - Indent with one tab.
> - Indent with two tabs.
> - Indent with 2 spaces + 1 tab.
> 
> As far as I remember, all of those were discussed with pros and cons, but I don't have the time right now to dig into it.

tests/ternaries/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -222,21 +318,69 @@ aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
...
 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 	? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-		? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-			? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-			: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-		: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+	  ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+	    ? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+	    : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+	  : aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
ikatyang: We need to use `dedentToRoot` from #3676 to re-indent with proper spaces/tab.

rhengles:
@ikatyang I'm amazed at the effort you put into this. I recognize that the output indeed looks very nice.

However, the reason the code is the way it is is because <tabs> _do not have an intrinsic length._

It looks very nice when you're formatting the code in your machine, with your editor tab length.

But the purpose of tabs is to be **tab length agnostic.**

When you share your tab indented code with another user who have an different tab length, the code will no longer be perfectly aligned.

That's why it is better to indent only with tabs, and to let the code be a little misaligned.

ikatyang:
It seems the trailing part is fine but the middle part should be still transformed into tabs per `align`?

rhengles:
@ikatyang Ah, the trailing part. It's simple when all you have inside a ternary is an literal, variable or object with simple properties (again, literal and variables).

But it gets messy _very_ fast - inside the object might be another object, an array or even a function, with endless combinations.

That's why I think in `useTabs` mode, every indentation level must be a `<tab>`.

ikatyang:
Removed the middle part and fixed the nested ternary case, @rhengles I think it should be fine now. If not, could you provide some examples so that I can improve it?

---

**Prettier pr-3745**
[Playground link](https://deploy-preview-3745--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBDdHNez3f9YA6UA-MQRZVfsWVNQ4xbeU2++sUqx79Vzz5CaUbvWETcxafWIACBXJJzg8xetRI5ABjXqAvnrlbVUdRpNHzFnVeuGz188tNO3czbcfvFDnwstvfwUAT0Dgt09dIOC-COsXO3ibaOTzOLSAlSTMqJzY-N87fRAAGhAIAAcYAEtoAGdkUFQAJxaIAHcABVaERpRUADcIGoATMrR6mGQAM1QAG3q4coAjFtQwAGs4GABlSo2aqABzZBgWgFdlkFGIMFmFpfKjpZaYLvXjgFtUB8XrgBW9QAHgAhdZbHa7VBfOAAGSOcD+TwqFxglTRACZkdcDi1XsgQCtUCsQvNoBNKi0jjAAOpjGAAC2QAA5tOUqRAlrT1pVCVS4K9Bkjyi04ABHC41MUfVDfX5IOb-cpLL41M6Xa71I7HeZwACKFwg8Bx5RgJPpoyZyAALGb1jV5jqAMIQL4-QlQaAikAXJYAFRJ-XOV30+iAA)
```sh
--tab-width 4
--use-tabs
```

**Input:**
```jsx
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
?
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
?
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
?
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa


a
    ? {
        a: 0
      }
    : {
        a: {
             a: 0
           }
            ? {
                a: 0
              }
            : {
                y: {
                    a: 0
                }
                    ? {
                        a: 0
                    }
                    : {
                        a: 0
                    }
            }
      }
```

**Output:**
```jsx
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
	? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
		? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
			? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
			: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
		: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
	: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;

a
	? {
			a: 0
	  }
	: {
			a: {
				a: 0
			}
				? {
						a: 0
				  }
				: {
						y: {
							a: 0
						}
							? {
									a: 0
							  }
							: {
									a: 0
							  }
				  }
	  };

```

rhengles:
@ikatyang Ah, I see it now. You only ever put spaces on the _closing_ bracket of an object who is a direct child of the result part of a ternary expression.

Just so we have a comprehensive test case to ensure no regressions, can you add this test case?
Of course no one should ever write code like this, but we have to make sure nothing breaks in the future 😃 

```javascript
a
	? {
			a: function() {
				return a
					? {
							a: [
								a
									? {
											a: 0,
											b: [
												a
													? [
															0,
															1
													  ]
													: []
											]
									  }
									: [
											[
												0,
												{
													a: 0
												},
												a
													? 0
													: 1
											],
											function() {
												return a
													? {
															a: 0
													  }
													: [
															{
																a: 0
															},
															{}
													  ];
											}
									  ]
							]
					  }
					: [
							a
								? function() {
										a
											? a(
													a
														? {
																a: a(
																	{
																		a: 0
																	}
																)
														  }
														: [
																0,
																a(),
																a(
																	a(),
																	{
																		a: 0
																	},
																	a
																		? a()
																		: a(
																				{
																					a: 0
																				}
																		  )
																),
																a()
																	? {
																			a: a(),
																			b: []
																	  }
																	: {}
														  ]
											  );
										a(
											a()
												? {
														a: 0
												  }
												: (function(a) {
														return a()
															? [
																	{
																		a: 0,
																		b: a()
																	}
															  ]
															: a(
																	[
																		a
																			? {
																					a: 0
																			  }
																			: {},
																		{
																			a: 0
																		}
																	]
															  );
												  })(
														a
															? function(a) {
																	return function() {
																		return 0;
																	};
															  }
															: function(a) {
																	return function() {
																		return 1;
																	}
															  }
												  )
										);
								  }
								: function() {
										
								  }
					  ];
			}
	  }
	: a;
```

wmertens:
At first I was like 🤔 and then I was like 😕 but it kinda looks ok so I'll just kill that eslint mixed spaces rule and be like 🤷‍♂️

jseminck:
I appreciate all the effort that has gone into this but wanted to give another view. We recently ran into this issue when prettier got upgraded and are not so excited about the change:

* Our company requires all projects to use tabs
* Our company provides a set of eslint rules to projects, that are not supposed to be overridden
* Prettier can be used by teams/projects if they want to, but this is not mandatory (as it can be too opinionated for some teams, their loss...)

This change doesn't play well with this setup. We cannot disable the ESLint rule related to mixing spaces and tabs (well we can, but we are asked not to).

So for us, that means we would have to stick to the older version of prettier. 😞

duailibe:
@jseminck Is/are the rule/rules auto-fixable?

In that case you can use `prettier-eslint` to fix the stuff after Prettier formats everything

jseminck:
I am not sure if this is fix-able ( https://eslint.org/docs/rules/no-mixed-spaces-and-tabs ). I will check how that could work. Thanks for the suggestion. 👍

lydell:
> - Our company provides a set of eslint rules to projects, that are not supposed to be overridden
> - Prettier can be used by teams/projects if they want to

That sounds like a constant uphill battle to me. Why not allow people to use [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier)?

jseminck:
It hasn't caused any problems until now. We have over 100 repositories and having a shared eslint config helps switching from one project to another a bit easier. 😄 

Is there any place where I can follow how this decision was made? I just don't understand why spaces are used when the option is set to use tabs.  Especially for people who have tabs set to something else than 4 spaces I think the new approach will look a bit odd (not sure why someone would to that though, maybe some people use 2 spaces) ?

This is the difference that we noticed:

previous:
```
function test() {
	const isNotLinkedToAnyOrganisation = false;
	const api = {};
	const organisation = {};

	return isNotLinkedToAnyOrganisation
		? Promise.resolve({ data: { id: null } })
		: api.get(`/endpoint`, {
				method: 'POST',
				body: JSON.stringify({ name: organisation.org_name }),
			});
}
```

to current:
```
function test() {
	const isNotLinkedToAnyOrganisation = false;
	const api = {};
	const organisation = {};

	return isNotLinkedToAnyOrganisation
		? Promise.resolve({ data: { id: null } })
		: api.get(`/endpoint`, {
				method: 'POST',
				body: JSON.stringify({ name: organisation.org_name }),
		  });
}
```

ikatyang:
As you can see in your example:

```diff
function test() {
	const isNotLinkedToAnyOrganisation = false;
	const api = {};
	const organisation = {};

	return isNotLinkedToAnyOrganisation
		? Promise.resolve({ data: { id: null } })
		: api.get(`/endpoint`, {
				method: 'POST',
				body: JSON.stringify({ name: organisation.org_name }),
-			});
+		  });
}
```

(I'm not sure if everyone's browser display tabs as 4 char width, so here's the screenshot.)

image

Its alignment looks odd before this PR if your tabs are not displayed as 2 char width, and it's not consistent with space-only:

```js
function test() {
    const isNotLinkedToAnyOrganisation = false;
    const api = {};
    const organisation = {};

    return isNotLinkedToAnyOrganisation
        ? Promise.resolve({ data: { id: null } })
        : api.get(`/endpoint`, {
              method: "POST",
              body: JSON.stringify({ name: organisation.org_name })
          });
}
```

The only solution I can think of is to add alignment spaces for it, don't know if there's a better one.

(I saw there's a [`smart-tabs`](https://eslint.org/docs/rules/no-mixed-spaces-and-tabs#options) option in ESLint, it should work in this case.)

wmertens:
note that the indentation after ?, : and the end is all two spaces, since ?
and : make up one space each, and thus the final indentation is always
correct, unlike when you are using tabs only. So just turn off the mixed
spaces rule and embrace prettier :)

PR #3758, opened 2018-01-17
Add plugin section in the readme

lipis:
I think it would be nice to link to the upcoming plugins for better visibility to attract more people to test/contribute.. :)

Maybe adding a few words or something..?

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,15 @@ Prettier can be run [in your editor](http://prettier.io/docs/en/editors.html) on
 
 ---
 
+## Plugins (under construction)
+
+* [prettier/prettier-python](https://github.com/prettier/prettier-python)
+* [prettier/prettier-php](https://github.com/prettier/prettier-php)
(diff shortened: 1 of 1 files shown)


lipis:
Am I missing any other language?

lipis:
I just noticed that there is a [plugin](https://prettier.io/docs/en/plugins.html) section in the docs, but maybe it's still a good idea to have something here as well.. wdyt?

azz:
We deleted all the duplication a while ago so the README is intentionally bare. I don't think we want to be adding this here. (we don't even have install instructions)

lipis:
maybe we could do in a less prominent way.. but the plugins section is not the most visited page in the docs.. but from here i thought that we could get more attention.. just an idea

azz:
We don't _want_ it to be the most visited page in the docs, especially since none of the plugins have been released yet.    #3762 should be fine

lipis:
fair enough

PR #3763, opened 2018-01-18
[vue] fix: *.vue with `lang='tsx'` is not formatted

wonderful-panda:
vue-language-server requires that lang must be "tsx" when <script> contains TSX syntax, like below:

```jsx
<script lang="tsx">
import { VNode } from "vue";
export default {
  render(): VNode { return <div>foo</div>; }
}
</script>
```

But prettier does not treat this code as TypeScript. This PR will fix this problem.


duailibe:
Thanks!

azz:
I didn't know vue used JSX?

wonderful-panda:
Yes. Using template is a most standard way, but JSX is also supported.
https://vuejs.org/v2/guide/render-function.html#JSX

Issue #3503, opened 2017-12-16
Prettier 2.0 (old)


azz:
**Note added by @j-f1:**

### This is NOT the current plan for Prettier v2.0. We’ve significantly scaled back our plans for what Prettier 2.0 will do, allowing us to get a less-controversial release out sooner: https://github.com/prettier/prettier/issues/6888

---

I've been thinking a bit about what we would do **if** (don't create a milestone just yet :wink:) we were to do a Prettier 2.0 release with some API/CLI breaking changes. No significant changes to code formatting other than perhaps changing some defaults.

### Easy?

* #68/#3469 - Change default value for `trailingComma` to `"es5"`.
  The diff advantages alone are excellent, and this seems to be commonplace among the most popular JS projects.

### CLI:

* #2691 - CLI defaults to `"**/*.{js,ts,...}"`.
  People should just be able to do `prettier --write` without passing messy globs _#0CJS_

* #2846 - Replace the "parser" option with "language".
  e.g. `echo ".foo {}" | prettier --language css`
  Keep `--parser flow` or perhaps support something like `--parser.js flow` for future extensibility. 

* Remove `--flow-parser` option.
  Has been deprecated for a long time.

### For debate

- #838 - Consistently add quotes to object keys.
  e.g. put quotes around `origin` in this example:
  ```js
  const headers = {
    'origin': 'foo.com',
    'access-control-allow-origin': '*',
  };
  ```

- Change the default value for `singleQuote` to `true`
  * Airbnb, standard, xo, and probably the majority of JS devs prefer this. Maybe do a Twitter poll to gauge opinions on this.

> To be clear, the Airbnb config requires single quotes for JS and double quotes for JSX.

### Outdated

* <del>#2884 - Remove the default parser.</del> <ins>Fixed by  #4528</ins>

* ~#1139 - Put a space after the `n` in `function () {}`~

vjeux:
All the changes you mentioned sound good (even switching single quote), now the big question is: do we think that the current suboptimal behavior is worth a breaking change. I’m not sure (in either way).

For consistent key quotes, I think we should just do it regardless, I’m not sure why we haven’t already.

azz:
I'm not certain either. The CLI breaking changes listed are quite small, and probably wont affect a large percentage of projects. The changes to the options are are trivial because you can just change them in a config file, so the "migration path" (if you can even call it that) is quite simple.

But there's always the risk with breaking changes that people will not know that and simply stick with the older version.

[...]

PR #3777, opened 2018-01-20
Split out location extraction into plugin config

czosel:
This is a first attempt to extract the parser-specific location extraction code (`locStart` and `locEnd` in `/common/util.js`) so that it can be configured by plugins - see https://github.com/prettier/prettier-php/pull/14 for more background.

This PR splits `util.js` in a "private" and a "public" part; the public one is configurable by passing `options` to it. It also shows how an example of it's usage for the HTML printer. In order to get this merged, all printers would have to be adapted to use new argument `util` in 
```javascript
genericPrint(path, options, print, args, util)
```
instead of importing it directly.

I thought I'd open this PR already to ask if this is going in the right direction. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,8 @@ const docblock = require("jest-docblock");
 
 const version = require("./package.json").version;
 
-const util = require("./src/common/util");
+const sharedUtil = require("./src/common/util-shared");
+const privateUtil = require("./src/common/util");
 const getSupportInfo = require("./src/common/support").getSupportInfo;
 
 const comments = require("./src/main/comments");
@@ -72,6 +73,7 @@ function formatWithCursor(text, opts, addAlignmentSize) {
     return { formatted: text };
   }
 
+  const util = sharedUtil(opts);
   const UTF8BOM = 0xfeff;
   const hasUnicodeBOM = text.charCodeAt(0) === UTF8BOM;
   if (hasUnicodeBOM) {
@@ -145,7 +147,7 @@ function format(text, opts, addAlignmentSize) {
   return formatWithCursor(text, opts, addAlignmentSize).formatted;
 }
 
-function findSiblingAncestors(startNodeAndParents, endNodeAndParents) {
+function findSiblingAncestors(startNodeAndParents, endNodeAndParents, util) {
   let resultStartNode = startNodeAndParents.node;
   let resultEndNode = endNodeAndParents.node;
 
@@ -187,6 +189,7 @@ function findSiblingAncestors(startNodeAndParents, endNodeAndParents) {
 }
 
 function findNodeAtOffset(node, offset, options, predicate, parentNodes) {
+  const util = sharedUtil(options);
   predicate = predicate || (() => true);
   parentNodes = parentNodes || [];
   const start = util.locStart(node);
@@ -297,6 +300,7 @@ function isSourceElement(opts, node) {
 function calculateRange(text, opts, ast) {
   // Contract the range so that it has non-whitespace characters at its endpoints.
   // This ensures we can format a range that doesn't end on a node.
+  const util = sharedUtil(opts);
   const rangeStringOrig = text.slice(opts.rangeStart, opts.rangeEnd);
   const startNonWhitespace = Math.max(
     opts.rangeStart + rangeStringOrig.search(/\S/),
@@ -335,7 +339,8 @@ function calculateRange(text, opts, ast) {
 
   const siblingAncestors = findSiblingAncestors(
     startNodeAndParents,
-    endNodeAndParents
+    endNodeAndParents,
+    util
   );
   const startNode = siblingAncestors.startNode;
   const endNode = siblingAncestors.endNode;
@@ -367,7 +372,10 @@ function formatRange(text, opts, ast) {
   );
   const indentString = text.slice(rangeStart2, rangeStart);
 
-  const alignmentSize = util.getAlignmentSize(indentString, opts.tabWidth);
+  const alignmentSize = privateUtil.getAlignmentSize(
+    indentString,
+    opts.tabWidth
+  );
 
   const rangeFormatted = format(
     rangeString,
--- a/index.js
+++ b/index.js
@@ -149,7 +149,6 @@ function format(text, opts, addAlignmentSize) {
 }
 
 function findSiblingAncestors(startNodeAndParents, endNodeAndParents, opts) {
-  const util = sharedUtil(opts);
   let resultStartNode = startNodeAndParents.node;
   let resultEndNode = endNodeAndParents.node;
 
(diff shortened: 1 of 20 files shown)


azz:
I think it's the right direction :+1:. I'll review properly tomorrow, but I noticed there's already a large number of functions exported. We'll need to review each function before committing to maintaining the API.

czosel:
Awesome! 
The current list of exports is not even complete. Here are all the calls to `util` in the various printers:
<details/>
  <summary>Show</summary>
  <pre>

  # html
  hasNewlineInRange
  locStart
  locEnd
  hasIgnoreComment

  # js
  getParentExportDeclaration
  isExportDeclaration
  hasIgnoreComment
  isNextLineEmpty
  shouldFlatten
  getNextNonSpaceNonCommentCharacterIndex
  startsWithNoLookaheadToken
  locStart
  hasNewlineInRange
  locEnd
  hasNodeIgnoreComment
  getLast
  printNumber
  isBlockComment
  getIndentSize
  getPenultimate
  getNextNonSpaceNonCommentCharacter
  isNextLineEmptyAfterIndex
  printString
  hasNewline
  skipWhitespace

  # markdown
  punctuationCharRange
  punctuationCharRange
  getLast
  getLast
  getMaxContinuousCount
  getMaxContinuousCount
  getLast
  getStringWidth
  getStringWidth
  getStringWidth
  getStringWidth
  mapDoc
  getLast
  hasIgnoreComment

  # graphql
  locStart
  isNextLineEmpty
  hasIgnoreComment

  # css
  locStart
  locEnd
  printString
  locStart
  locEnd
  hasNewline
  locStart
  isNextLineEmpty
  printString
  hasIgnoreComment
  </pre>
</details>

czosel:
Hi @azz, sorry i didn't get to follow up on this in the last couple of days - can you take a look and maybe give an idea of what needs to be done until we can merge this?

azz:
Sorry, I forgot too, and I've been on holiday until today. 

I think we should start with the following functions exported as public:

* `makeString`
* `getNextNonSpaceNonCommentCharacterIndex`
* `isNextLineEmptyAfterIndex`
* `isNextLineEmpty`
* `mapDoc`

czosel:
No problem! Unfortunately i also don't have time to work on this every day, but today i do 😃 

I reduced the set of exposed functions and refactored the implementation by passing the custom versions of `locStart`/`locEnd` directly to the affected `util` functions - I think it's much simpler now. `util-shared.js` is now just a very thin layer over `util.js`.

Looking forward to your feedback!

src/main/options.js
```diff
@@ -48,6 +50,8 @@ function normalize(options, opts) {
   }
 
   rawOptions.astFormat = resolveParser(rawOptions).astFormat;
+  rawOptions.locEnd = resolveParser(rawOptions).locEnd;
```
azz: Rather than call `resolveParser` 3 times, could we store the return value as a local variable?

src/common/util.js
```diff
@@ -238,7 +240,13 @@ function hasSpaces(text, index, opts) {
   return idx !== index;
 }
 
-function locStart(node) {
+function locStart(node, locStartFromPlugin) {
```
azz: I think we should delete `locStart` and `locEnd` from `util.js` and force each parser (in plugins) to provide it. This way we won't have to have awkward code like `util.locEnd(node, options.locEnd);`.

czosel: I agree, that would be much cleaner. It does require many changes though, because every callsite has to be changed (see https://github.com/prettier/prettier/pull/3777/commits/b2937f87d467eb97f57ba5fad065b41eb0cfc839 some first progress on this). I don't mind doing those changes, but let me know if you'd implement this differently!

azz: Looking good to me 👍

czosel: Awesome, thanks 😃

j-f1:
Can you merge in `master`? It looks like there are conflicts.

czosel:
@j-f1 done!

lydell:
Hmm, github still says there are conflicts…

czosel:
@lydell That was actually a new conflict - i rebased again.

src/language-markdown/printer-markdown.js
```diff
@@ -42,12 +42,12 @@ const INLINE_NODE_WRAPPER_TYPES = INLINE_NODE_TYPES.concat([
   "heading"
 ]);
 
-function genericPrint(path, options, print) {
+function genericPrint(path, options, print, args, util) {
```
j-f1: 🔥 `util`?

czosel: Thanks - fixed.

index.js
```diff
@@ -160,7 +164,8 @@ function findSiblingAncestors(startNodeAndParents, endNodeAndParents) {
     if (
       endParent.type !== "Program" &&
       endParent.type !== "File" &&
-      util.locStart(endParent) >= util.locStart(startNodeAndParents.node)
+      util.locStart(endParent, opts.locStart) >=
```
azz: Shouldn't this just be `opts.locStart(endParent) `?

src/common/util-shared.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+const util = require("./util");
+
+module.exports = function(options) {
```
azz: Could we delete `locStart` and `locEnd` here, then remove the factory/closure and pass `options` explicitly to functions that need it?

index.js
czosel: Definitely! fixed.

src/common/util-shared.js
czosel: Good idea - that also allows more reuse of the already existing `options` argument in the printer. Removing the closure would mean that we also won't pass `util` down to the printer anymore, but that might be a good idea? It's certainly removing an extra abstraction layer ...

czosel:
Tests should be passing now 🎉

src/language-css/index.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 const printer = require("./printer-postcss");
+const { lineColumnToIndex, getLast } = require("../common/util");
```
j-f1: Node 4 doesn’t like destructuring :(

czosel: Oops - fixed.

azz:
Code looks good! We should add a couple of unit tests to ensure the shape of `shared-util` doesn't accidentally break.

czosel:
Added a small test that asserts the presence of the five exported functions!

src/language-css/index.js
lipis: Node 4 is expiring soon :) There was a plan to drop it at one point..

czosel: That would/will be nice 😃

j-f1: @lipis:

> Looks good except for node v4 compat. Was chatting with **@​vjeux** and we think we shouldn't force people to upgrade versions, especially since Prettier is often run in CI.

https://github.com/prettier/prettier/issues/3503#issuecomment-359937346

lipis: I will just close that one then: #3919

czosel:
I rebased again - let me know if you see anything that still requires changes! @azz

azz:
Thank you! 🏆

czosel:
Thanks for merging! 😄

PR #3793, opened 2018-01-23
feat: remove extra spaces in `url` inside `@import`

evilebottnawi:


tests/css_import/url.css
```diff
@@ -24,3 +24,7 @@ $dir: 'fonts';
 @import 'rounded-corners', 'text-shadow';
 $family: unquote("Droid+Sans");
 @import url("http://fonts.googleapis.com/css?family=#{$family}");
+@import url( "foo.css");
+@import url("foo.css" );
+@import url( "foo.css" );
+@import url(  "foo.css"  );
```
j-f1: What about `@import url("foo bar baz.css");`?

src/language-css/printer-postcss.js
```diff
@@ -218,7 +218,7 @@ function genericPrint(path, options, print) {
       return adjustStrings(n.value, options);
     }
     case "media-url": {
-      return adjustStrings(n.value, options);
+      return adjustStrings(n.value.replace(/ /g, ""), options);
```
lydell: What about other types of whitespace?

evilebottnawi:
@j-f1 @lydell fixed and add tests :+1:

j-f1:
```
✘  prettier/prettier  Replace `.replace(/^url\(\s+/gi,·"url(")` with `⏎······.replace(/^url\(\s+/gi,·"url(")⏎······`  
  src/language-css/clean.js:97:32
```

evilebottnawi:
@j-f1 fixed
@lydell friendly ping

tests/css_import/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -63,6 +63,16 @@ $dir: 'fonts';
 @import 'rounded-corners', 'text-shadow';
 $family: unquote("Droid+Sans");
 @import url("http://fonts.googleapis.com/css?family=#{$family}");
+@import url( "foo.css");
+@import url("foo.css" );
+@import url( "foo.css" );
+@import url(  "foo.css"  );
+@import url("foo bar baz.css");
+@import url(  "foo bar baz.css"  );
+@import url("foo  bar  baz.css");
```
lydell: There are no tests for the case insensitivity?

evilebottnawi: @lydell right now media parsed don't works with `URL` case, i think better fix this in parser then add tests and fixes

Issue #1811, opened 2017-05-29
CSS: Improve arithmetic operators


lydell:
Goal: Print them more like JS. Flatten indentation and group operators.

```scss
.Image {
  @media (max-width: $image-width-responsive) {
    padding-bottom:
      (
        100% *
        $image-height /
        ($image-width-responsive + $image-margin-responsive * 2)
      );
  }
}
```
```scss
.Image {
  @media (max-width: $image-width-responsive) {
    padding-bottom: (
      100% *
        $image-height / ($image-width-responsive + $image-margin-responsive * 2)
    );
  }
}
```

Compare to this JS:

```js
{
  {
    padding_bottom = (
      100 *
        $image_height /($image_width_responsive + $image_margin_responsive * 2)
    );
  }
}
```
```js
{
  {
    padding_bottom =
      100 *
      $image_height /
      ($image_width_responsive + $image_margin_responsive * 2);
  }
}
```

---

```scss
a {
  b {
    img {
      max-height:
        calc(
          100vh -
          #{var(--margin)} * 2 - #{var(--padding)} * 2 -
          #{$extra-space-for-scrollbar}
        );
    }
  }
}
```
```scss
a {
  b {
    img {
      max-height: calc(
        100vh -
          #{var(--margin)} *
          2 -
          #{var(--padding)} *
          2 -
          #{$extra-space-for-scrollbar}
      );
    }
  }
}
```

Compare to this JS:

```js
{
  {
      max_height = calc(
        100 -
          $(car(__margin)) *
          2 -
          $(car(__padding)) *
          2 -
          $($extra_space_for_scrollbar)
        );
  }
}
```
```js
{
  {
    max_height = calc(
      100 -
        $(car(__margin)) * 2 -
        $(car(__padding)) * 2 -
        $($extra_space_for_scrollbar)
    );
  }
}
```

Extracted from: #1805

Important comment by @vjeux:

> Right now we only get a stream of tokens. We'd need to parse it to be able to reprint it properly. For this, we need to implement a small arithmetic parser and print it correctly. It shouldn't be too hard

j-f1:
Here’s the current status of this:

**Prettier 1.10.2**
[...]

PR #3795, opened 2018-01-23
tests: parens spaces

evilebottnawi:
just tests


tests/css_parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -22,4 +37,18 @@ a {
...
+.foo {
+  --paddingC: calc(var(--widthB) / 2);
+  content: attr(data-title);
+  color: var(--main-bg-color);
+  background-color: rgb(255, 0, 0);
+  background: element(#css-source);
+  width: calc(100% - (#{var(--g-spacing)} - #{$iframe-x-padding}) * 2);
+  padding-bottom: (
+    100% * $image-height / ($image-width-responsive + $image-margin-responsive *
+          2)
```
j-f1: This doesn’t look very pretty. Would you be willing to fix this in this PR, or should I open a new issue?

evilebottnawi: @j-f1 in this PR just tests, to ensure all extra spaces before and after parens and signs was removed. We already have issue https://github.com/prettier/prettier/issues/1811

PR #3798, opened 2018-01-23
feat(css): lowercase build-in functions

evilebottnawi:
Two question:
1. Can i create `reference` directory and put all `keywords` and build-in `css` stuff there?
2. In theory it can be break this code (only for scss):
```scss
@function URL($some-number, $another-number){
  @return $some-number + $another-number
}

.my-module {
  padding: URL(10px, 5px);
}
```
But it is very very bad practice (i don't see in popular framework this), maybe better use `prettier-ignore` for this and add docs?

### DIFF
--- a/src/language-css/clean.js
+++ b/src/language-css/clean.js
@@ -33,6 +33,118 @@ function clean(ast, newObj) {
   ) {
     newObj.value = newObj.value.toLowerCase();
   }
+  if (ast.type === "value-func") {
+    const lowerCasedValue = newObj.value.toLowerCase();
+    const functions = [
+      // URL
+      "url",
+      "local",
+      // Font format
+      "format",
+      // Color
+      "rgb",
+      "rgba",
+      "hsl",
+      "hsla",
+      "hwb",
+      "gray",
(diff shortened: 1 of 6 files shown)


lydell:
This is probably safe in practice, but not in theory. In SCSS you can define a function like:

```scss
@function ScaleX($x) {
  @return scaleX($x * 0.99);
}
```

If so, it is not correct to change `transform: ScaleX(5);` into `transform: scaleX(5);`.

j-f1:
How about only doing this for vanilla CSS?

lydell:
Personally, I don't think this is worth doing at all. I have never seen anyone use unusual casing for functions in CSS.

j-f1:
Should we fix the conflicts, make the required changes, and merge, or should we close this?

lydell:
I vote for closing, but I don’t feel strongly about this.

evilebottnawi:
@j-f1 we can merge this only for vanilla CSS, it is safe, for less/scss it is not safe, if we accept it for vanilla css i can rebase

PR #3794, opened 2018-01-23
feat: remove extra spaces in attribute selector

evilebottnawi:
:star:

PR #3799, opened 2018-01-23
fix(css): remove other types of whitespace in `clean`

evilebottnawi:
Based on https://github.com/prettier/prettier/pull/3794#discussion_r163313410

### DIFF
--- a/src/language-css/clean.js
+++ b/src/language-css/clean.js
@@ -66,7 +66,7 @@ function clean(ast, newObj) {
   }
 
   if (ast.type === "selector-attribute") {
-    newObj.attribute = newObj.attribute.replace(/ /g, "");
+    newObj.attribute = newObj.attribute.replace(/\s/g, "");
 
     if (newObj.value) {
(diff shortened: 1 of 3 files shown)


evilebottnawi:
@lydell how much approval do I need to make merge?

lydell:
A green check mark from Travis is all that is needed for small things like this, I'd say :) I think it's better not to slow you down on your CSS fixing spree. If somebody finds something bad after merge we can revert or do a follow-up PR.

lydell:
Looks like me and @j-f1 gave different answers :)

j-f1:
*\*defers to the collaborator with more time working with the project\**

lydell:
My point was that the CSS printing is generally buggy and not so polished like the JS printing, so I think it's fine to "let you loose" and fix things quickly. Many small PRs is still good since it's easy to review even after merges.

As opposed to PRs to the JS printing, which is very stable, or the core printing logic. That should be reviewed more before merging.

PR #3802, opened 2018-01-23
fix: handle YAML front matter block

### LINKED ISSUE #3321: SCSS: Preserve YAML front matter block

evilebottnawi:
Fixes #3321

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const createError = require("../common/parser-create-error");
+const grayMatter = require("gray-matter");
 
 function parseSelector(selector) {
   // If there's a comment inside of a selector, the parser tries to parse
@@ -189,7 +190,11 @@ function parseNestedCSS(node) {
         };
       }
     }
-    if (node.type && typeof node.value === "string") {
+    if (
+      node.type &&
+      node.type !== "css-comment-yaml" &&
+      typeof node.value === "string"
+    ) {
       try {
         if (node.value.endsWith(DEFAULT_SCSS_DIRECTIVE)) {
           node.default = true;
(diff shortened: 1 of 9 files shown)


src/language-css/parser-postcss.js
```diff
@@ -261,15 +274,29 @@ function parse(text, parsers, opts) {
     ? opts.parser === "scss"
     : IS_POSSIBLY_SCSS.test(text);
 
+  let frontMatter = null;
+
+  try {
+    frontMatter = grayMatter(text);
+  } catch (originalError) {
```
lydell: Why is this caught error called `originalError`?

evilebottnawi: @lydell other name?

lydell: `error` I guess

evilebottnawi:
@lydell done :+1:

src/language-css/parser-postcss.js
```diff
@@ -261,15 +274,29 @@ function parse(text, parsers, opts) {
     ? opts.parser === "scss"
     : IS_POSSIBLY_SCSS.test(text);
 
+  let frontMatter = null;
+
+  try {
+    frontMatter = grayMatter(text);
+  } catch (error) {
+    throw error;
+  }
```
lydell: Why try-catch if we immediately re-throw?

evilebottnawi:
@lydell done :+1:

PR #3810, opened 2018-01-24
fix(css): handle namespace in attribute selector

evilebottnawi:
Also add tests.
`[foo|att=val] {}` - namespace attribute selector - https://drafts.csswg.org/css-namespaces-3/#css-qnames

### DIFF
--- a/src/language-css/clean.js
+++ b/src/language-css/clean.js
@@ -66,7 +66,17 @@ function clean(ast, newObj) {
   }
 
   if (ast.type === "selector-attribute") {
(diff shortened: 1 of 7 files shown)


evilebottnawi:
@j-f1 @lydell friendly ping, it is bug and break code

lydell:
Sorry, I have no idea what this namespace thing is :)

evilebottnawi:
@lydell https://developer.mozilla.org/ru/docs/Web/CSS/@namespace

PR #3817, opened 2018-01-24
tests(css): `viewport` at rule

evilebottnawi:


evilebottnawi:
@j-f1 i think for css better leave `{` on the same line when no expression in media at rule, we use it for `function` and `scss maps`, the same behavior in this case seems to me more correct. Also we don't have any issue around this, we can change this in future if community vote for above example

j-f1:
@evilebottnawi I agree with you on that. I was just suggesting that as an additional test.

evilebottnawi:
@j-f1 friendly ping

PR #3818, opened 2018-01-24
docs: add information about precise-commits

SavePointSam:


azz:
I think we should add a summary at the top of this page explaining why you would use each one of these tools.

`lint-staged` if you need to do things other than Prettier (e.g. ESLint)
`pretty-quick` if your codebase is formatted with Prettier and you want to format the staged files.
`precise-commits` if your code is only partially formatted with Prettier, and you want to format just the staged lines.

SavePointSam:
Updated!

SavePointSam:
Any chance we can get this in? Thanks!

JamesHenry:
Haha how did you even discover this? I was holding off on telling anyone until I'd dogfooded it a bit 😄 

Anyway thanks a lot for adding! Hopefully this can complement the great existing tools, seems like we've got decent coverage on the various use-cases now 🙌

SavePointSam:
@JamesHenry, I saw a comment about someone (I believe you), wanting to make this in response to an issue that I was having like 10 mins after you pushed it to GH. So, I went and posted this PR to add it to the docs within the hour. :D

PR #3821, opened 2018-01-25
feat(css): lowercase `keyframes` keywords `to` and `from`

evilebottnawi:
Also don't parse `@keyframes` because it does not make sense and should be increase perf


tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -387,4 +460,28 @@ $KeepTopLevelVar: val;
 @media (#{$Keep}: #{$Keep}) {
 }
 
+@keyframes identifier {
+  FROM {
```
j-f1: This isn’t lowercase :)

evilebottnawi: @j-f1 oh sorry, I'm tired :tired_face:

evilebottnawi:
Close in favor https://github.com/prettier/prettier/pull/3954

PR #3825, opened 2018-01-26
tests(css): attr function

evilebottnawi:


tests/css_parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -128,6 +128,78 @@ a {
...
+    prop7:
+        attr(
+            data-foo
+        );
+    prop8:
+    attr(
+    data-foo
+    )
+    ;
+    prop9:"("attr(title)")";
```
evilebottnawi: This parsed as `value-comma_group`, seems it is invalid and should be `value-list_group` and we should implement logic

suchipi: We should probably open a tracking issue

evilebottnawi: @suchipi done https://github.com/prettier/prettier/issues/3842

Issue #3829, opened 2018-01-26
require and define callsare formatted differently


salemhilal:
**Prettier 1.10.2**
[...]

PR #3830, opened 2018-01-26
Treat define calls like require

salemhilal:
See issue #3829. This commit keeps define calls as a unit when possible,
to prevent an extra indent in the body of an AMD-style define.

Rather than adding the "define" check in the same case as the existing
"require" check, I added a separate condition to explicitly call out the
fact that this checks for both AMD and CommonJS modules.

`yarn test -u` yields no changes, and `yarn lint` passes successfully.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -802,16 +802,12 @@ function printPathNoParens(path, options, print, args) {
 
       const optional = printOptionalToken(path);
       if (
-        // We want to keep CommonJS require calls as a unit
-        (!isNew &&
-          n.callee.type === "Identifier" &&
-          n.callee.name === "require") ||
-        // As well as AMD-style require and define calls.
-        // Note that the previuos case covers AMD-style require calls as well.
+        // We want to keep CommonJS- and AMD-style require calls, and AMD-style
+        // define calls, as a unit.
         // e.g. `define(["some/lib", (lib) => {`
         (!isNew &&
           n.callee.type === "Identifier" &&
-          n.callee.name === "define") ||
+          (n.callee.name === "require" || n.callee.name === "define")) ||
         n.callee.type === "Import" ||
         // Template literals as single arguments
         (n.arguments.length === 1 &&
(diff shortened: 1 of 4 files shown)


src/language-js/printer-estree.js
```diff
@@ -802,10 +802,16 @@ function printPathNoParens(path, options, print, args) {
...
+        // We want to keep CommonJS require calls as a unit
         (!isNew &&
           n.callee.type === "Identifier" &&
           n.callee.name === "require") ||
+        // As well as AMD-style require and define calls.
+        // Note that the previuos case covers AMD-style require calls as well.
+        // e.g. `define(["some/lib", (lib) => {`
+        (!isNew &&
+          n.callee.type === "Identifier" &&
+          n.callee.name === "define") ||
```
j-f1: What do you think about moving this to the previous test?

```js
(!isNew &&
  n.callee.type === "Identifier" &&
  (n.callee.name === "require" ||
    n.callee.name === "define"))
```

salemhilal: I did that at first; I'm ambivalent there honestly. I ended up with this approach because it allows for a more obvious way to diverge CommonJs calls from AMD calls in the future. Lemme know and I'm down to move it to the previous test.

lydell:
Could you add your input from #3829 as a test?

salemhilal:
^ tests added

salemhilal:
^ added babylon parser to the test.

src/language-js/printer-estree.js
suchipi: Let's move it into the previous test

j-f1:
Thanks for contributing to Probot! 🎉

salemhilal:
Woo! Thank you.

salemhilal:
Hey, what's the release schedule like for prettier? I'm curious when this patch will be live (without having to checkout master).

suchipi:
We usually do a patch release every month or so, so it's around that time that we should do a new one.

PR #3840, opened 2018-01-30
Ignore package-lock.json and yarn-error.log

lipis:
It's happening quite often that these files are making it into PRs by accident, so why not ignoring them in general.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@
 .DS_Store
 coverage
 .idea
+package-lock.json
+yarn-error.log


.gitignore
```diff
@@ -13,3 +13,5 @@
 .DS_Store
 coverage
 .idea
+package-lock.json
+yarn-error.log
```
ikatyang: There's already a `*.log`.

lipis: oups. :)

SimenB:
Could also add an npmrc which disable its generation

lipis:
This is less intrusive than adding yet another config file :)

lipis:
While we are at it.. should I sort this file?

PR #3843, opened 2018-01-30
tests(css): trailing comma in maps

evilebottnawi:
just tests


tests/css_trailing_comma/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,6 +1,37 @@
...
+);
+$map: (
+  small: 767px,
+  medium: 992px,
+  large: 1200px
+);
+$map: (
+  "medium": (min-width: 800px),
+  "large": (min-width: 1000px),
+  "huge": (min-width: 1200px)
```
j-f1: Can you have this obey the `trailingComma` option?

evilebottnawi: @j-f1 let's do this

evilebottnawi:
@j-f1 `css` and `less` don't support `trailing comma`, `sass` partial, but very ugly, i think we should avoid using `trailing comma` in `css/less/scss` for pretty output. This test only to ensure all commas was removed.

Issue #3451, opened 2017-12-10
Expose config.resolveConfigFile in API?


darahak:
We have this [issue](https://github.com/prettier/prettier-atom/issues/270) with `prettier-atom` where it keeps using the cached config, even when the user modified their `.prettierrc` file.

We need the config file path if we want to watch it and clear the cache if it gets updated. 

I wanted to check with you, if you see any downsides/caveats in exposing this method or if you think there could be a better solution.

Thanks!

josephfrazier:
I think it's a good idea, there was a tad of discussion about it here: https://github.com/prettier/prettier/pull/2760#discussion_r145536461

darahak:
Thanks @josephfrazier! Totally missed this convo in the search results...

duailibe:
Hm, that's an unfortunate issue and I don't oppose this solution but I think a "Reload option" on prettier-atom could solve the issue.

Changing a config file is generally a set and forget it action.. I'm not sure how useful it is to keep watching this file. But that's your call and again I don't oppose exposing this method

darahak:
@duailibe I agree a "force reload config" command would be nice too, but ideally I would like the user to not have to do anything.

A watcher might consume too many resources for this kind of event.
A good compromise could be checking modification dates on config files when appropriate? (e.g. text editor focus + eligible file type)

We could have both.

SavePointSam:
I’d be happy to look at posting an update to PR to accomplish this. I was going to follow the pattern that Prettier#formatWithCursor uses. By creating a Prettier#resolveConfigWithFile that will encapsulate the tradition response in an `options` property and a `file` property that contains all the information an editor may need to watch configurations reliably.

duailibe:
FTR this was done in #4364, will be included in next release!

PR #3846, opened 2018-01-31
add resolveConfigWithFilePath API

SavePointSam:
This PR introduces a new API `prettier.resolveConfigWithFilePath`. This method uses the pre-existing internals to find and merge prettier configurations, but exposes the resolved file path for consumers. This will close https://github.com/prettier/prettier/issues/3451 and allow us to solve https://github.com/prettier/prettier-atom/issues/270.

I also reorganized the API docs a bit so that similar API sit next to each other.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -59,7 +59,7 @@ Use `prettier.resolveConfig.sync(filePath [, options])` if you'd like to use syn
 
 ## `prettier.resolveConfigWithFilePath(filePath [, options])`
 
-`resolveConfigWithFilePath` works that same as `resolveConfig` with one difference. The response will include both the configuration that was resolved, along with the file path to the configuration that was used to generate the configuration. However, it will not include the path to the located `.editorconfig` file when used as the core Prettier configuration acts as an override. This is useful for editor integrations, to subscribe to the resolved configuration and reload with any changes.
+`resolveConfigWithFilePath` works that same as `resolveConfig` with one difference. The response will include both the configuration that was resolved, along with the path to the file that was used to generate the configuration. However, it will not include the path any `.editorconfig` file when used, as the configuration file acts as an override. This is useful for editor integrations, to subscribe to the resolved configuration and reload with any changes.
 
 ```js
 prettier.resolveConfigWithFilePath(filePath).then(result => {
--- a/docs/api.md
+++ b/docs/api.md
@@ -16,10 +16,6 @@ prettier.format("foo ( );", { semi: false });
 // -> "foo()"
 ```
 
-## `prettier.check(source [, options])`
-
-`check` checks to see if the file has been formatted with Prettier given those options and returns a `Boolean`. This is similar to the `--list-different` parameter in the CLI and is useful for running Prettier in CI scenarios.
-
 ## `prettier.formatWithCursor(source [, options])`
 
 `formatWithCursor` both formats the code, and translates a cursor position from unformatted code to formatted code. This is useful for editor integrations, to prevent the cursor from moving when code is formatted.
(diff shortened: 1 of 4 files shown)


azz:
What if we just had `resolveConfig` return the path, too:

```js
prettier.resolveConfig.sync('some/path.js');
// -> 
//    {
//      useTabs: true,
//      configPath: "/abs/path/.prettierrc"
//    }
```

SavePointSam:
That's also an option, however I would suggest against it. It makes sense that the API returns the configuration as Prettier can later ingest it. Adding extra properties in this one case doesn't quite make sense to me and feels like a mix of concerns. Even more so when the rest of the time we are working with the configuration payload it is separated between `config` and `filePath`.

src/config/resolve-config.js
```diff
@@ -52,6 +62,13 @@ function _resolveConfig(filePath, opts, sync) {
       return null;
     }
 
+    if (!!includeFilePath) {
```
j-f1: Can this just be `if (includeFilePath)`?

SavePointSam: fixed!

docs/api.md
```diff
@@ -57,6 +57,18 @@ If `options.editorconfig` is `true` and an [`.editorconfig` file](http://editorc
 
 Use `prettier.resolveConfig.sync(filePath [, options])` if you'd like to use sync version.
 
+## `prettier.resolveConfigWithFilePath(filePath [, options])`
+
+`resolveConfigWithFilePath` works that same as `resolveConfig` with one difference. The response will include both the configuration that was resolved, along with the file path to the configuration that was used to generate the configuration. However, it will not include the path to the located `.editorconfig` file when used as the core Prettier configuration acts as an override. This is useful for editor integrations, to subscribe to the resolved configuration and reload with any changes.
```
josephfrazier: > The response will include both the configuration that was resolved, along with the file path to the configuration that was used to generate the configuration.

josephfrazier:
...wait a minute. Couldn't we just export `resolveConfigFile` from `src/config/resolve-config.js` and `index.js`, instead of adding a separate function?

After typing this out, I realized that the use case (editor integrations) will likely want both the config and the file path anyway, so it probably makes sense to combine them into a single function, instead of forcing the integration to call two functions (which would also unnecessarily repeat the path resolution). I'm going to leave this comment here anyway, for the record.

SavePointSam:
I thought the same thing at first. Though as someone who would be implementing this type of functionality, I'd much rather make a call that returns both the resolved/merged config as well as the file path of the config used. There is room for error when using those API separately.

olsonpm:
I believe a valid workaround until this is resolved would be to use [cosmiconfig's `load`](https://github.com/prettier/prettier/pull/3846) directly.  Anyway, I'm planning on using that to address "[Show which options are used on format](https://github.com/prettier/prettier-atom/issues/178)".

azz:
Is there a way to expose the path to the editorconfig file too?

josephfrazier:
I don't think https://github.com/editorconfig/editorconfig-core-js provides a way to see the path of the files being parsed (note that there can be more than one). However, assuming there's just one, we could use some combination of `find-up` and `find-project-root` to locate it. See also https://github.com/prettier/prettier/pull/3559

olsonpm:
"assuming just one" could produce false positives though the simplification would be nice.  Seems the devs behind editorconfig-core-js are pretty responsive though, so a PR adding a method such as this one wouldn't be out of the question.

SavePointSam:
Sorry for the silence! Having some crunch time at the office :(

I totally agree we should find a way to expose the path to `.editorconfig`. Can it actually pick up multiple to merge together? That's pretty rough..

olsonpm:
Bummer about the office time.  I'm in between jobs so I finally get to work on stuff I enjoy haha.

> Can it actually pick up multiple to merge together?

Unfortunately.  I created a [quick repo](https://github.com/olsonpm/test-editorconfig) if you wanted to mess around with it.

duailibe:
FTR this was done in #4364, did not include the editorconfig part.

suchipi:
I'm going to close this based on https://github.com/prettier/prettier/pull/3846#issuecomment-384495417

PR #3848, opened 2018-01-31
fix: Proper case for Prettier, JavaScript, TypeScript, GitHub, Less

lipis:


docs/rationale.md
```diff
@@ -34,47 +34,22 @@ Prettier outputs the minimum number of parentheses required to ensure that the b
...
-
-
-### JSX
-
-
-### Boolean expressions
-
-
-### String concatenation
--->
```
lipis: I'm guessing that is not needed at the moment..

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3861,7 +3861,7 @@ All of JSX and Flow syntax is supported. In fact, the test suite in \`tests/flow
 _is_ the entire Flow test suite and they all pass.
 
 Prettier also supports [TypeScript](https://www.typescriptlang.org/), CSS,
-[LESS](http://lesscss.org/), [SCSS](http://sass-lang.com),
```
ikatyang: We don't need to update test cases/snapshots. 😂

lipis: But there is no harm in it.. right?

ikatyang: Yeah, just 😂.

PR #3859, opened 2018-02-01
Add React Native Firebase to the list of users

chrisbianca:
We've just added Prettier to our codebase and it's made a world of difference!


lipis:
Build error is unrelated it seems..

PR #3718, opened 2018-01-11
Landing page revamp

azz:
Started working on #3669. Lots to do still, but it is starting to take shape.

## [🌐Live Preview](https://deploy-preview-3718--prettier.netlify.com/)

image

cc. @orta @vjeux @sandhose

PR #3870, opened 2018-02-01
feat: Randomize users, showcases (fixes #3868)

### LINKED ISSUE #3868: website: Randomize users on every reload of the page
https://prettier.io/en/users/

lipis:


website/data/users.yml
```diff
@@ -1,27 +1,21 @@
...
   infoLink: https://babeljs.io
-  pinned: true
 - caption: ZEIT
   image: "/images/users/zeit-200x100.png"
   infoLink: https://zeit.co
-  pinned: true
 - caption: Webpack CLI
   image: "/images/users/webpack-200x100.png"
   infoLink: https://webpack.js.org/api/cli/
-  pinned: true
```
lipis: Unless we want to keep the "big" guys..

website/siteConfig.js
```diff
@@ -13,7 +13,7 @@ function loadYaml(fsPath) {
   return parseYaml(fs.readFileSync(path.join(__dirname, fsPath), "utf8"));
 }
 
-const users = loadYaml("./data/users.yml");
+const users = loadYaml("./data/users.yml").sort(() => Math.random() - 0.5);
```
lydell: This isn't fair: https://stackoverflow.com/a/18650169/2010616

Is it possible to use npm modules here? Could we just use lodash's shuffle or something?

lipis: _but it's very cute_

lipis: no need to require lodash, just for this.. I'll add a little helper function..

lipis: From the same SO question..

website/siteConfig.js
```diff
@@ -13,7 +13,15 @@ function loadYaml(fsPath) {
   return parseYaml(fs.readFileSync(path.join(__dirname, fsPath), "utf8"));
 }
 
-const users = loadYaml("./data/users.yml").sort(() => Math.random() - 0.5);
+function shuffleArray(array) {
+  for (let index = array.length - 1; index > 0; index--) {
+    const rnd = Math.floor(Math.random() * (index + 1));
+    [array[index], array[rnd]] = [array[rnd], array[index]];
+  }
+}
```
lydell: Missing `// https://stackoverflow.com/a/12646864/2010616` comment :)

lydell: GitHub marked this as outdated, but it's not :)

website/siteConfig.js
```diff
@@ -13,7 +13,16 @@ function loadYaml(fsPath) {
   return parseYaml(fs.readFileSync(path.join(__dirname, fsPath), "utf8"));
 }
 
-const users = loadYaml("./data/users.yml");
+function shuffleArray(array) {
+  const copy = array.concat();
```
j-f1: I prefer `array.slice()`, but you can use whatever you’d like.

website/siteConfig.js
lipis: added

lipis:
We should still randomize them when showing the full list.. and since the front page is using very special images for special companies.. then maybe they shouldn't be taken from the user list at all.. or have two separate images in the #3718 for the ones that will appear on the homepage. Either case I think it's a good idea to randomize them and the new looks should adopt to it..

j-f1:
We could randomly choose from users who provide altered versions of their logos for use on the homepage.

lipis:
Yep! But what do you want me to do here? Should I load on the homepage only the pinned versions for now.. and in the #3718 we can remove the pinned and only check if there is a second version of the product to show it on homepage.

orta:
I'm less sold on this as a useful feature of the index page, these kind of sections exist so that you can see that the big projects that you probably rely on are using prettier. It's awesome that Wingify/n26/escrow.com use prettier, but it's not a great selling point. 

I wonder if taking this and switching out the call to action to be something more like this:

screen shot 2018-02-02 at 1 23 46 pm

Where a textual representation of x amount from the users list can be used instead would work? Might feel a bit nicer than the current CTA buttons for the section anyway

j-f1:
> Might feel a bit nicer than the current CTA buttons for the section anyway

Especially since the CTA doesn’t apply to most visitors.

suchipi:
I agree with @orta (as mentioned in https://github.com/prettier/prettier/issues/3868#issuecomment-362488722)

lipis:
shall we randomize it?

j-f1:
Do you want to do what @orta suggested in this PR or in another PR?

lipis:
Maybe we can do that in another PR..

orta:
Then this PR should probably be closed, there's not really any value in shuffling the images between depoys.

PR #3873, opened 2018-02-01
feat: Add support for no-response-bot

lipis:
More about the bot here: https://probot.github.io/apps/stale/

Let me know about the settings that we want if we are about to add this bot, to help with outdated issues.

Potential issues that could be marked as stale (older than 90 days): 
https://github.com/prettier/prettier/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+is%3Aissue+is%3Aopen+updated%3A%3C2017-11-01+-label%3Atype%3Abug+

We could also exclude maybe the `status:needs discussion`.


.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
```
lipis: I could remove the comments before merge

suchipi:
Does people adding :+1:s count as activity? Because if not, I could see this closing some issues that we're intentionally leaving open awaiting community demand.

lydell:
The number of open issues is growing, yes, but slowly. We're pretty good a triaging, so most of the open issues should be relevant. Perhaps we don't need a bot until we reach 500 or 1000 open issues or something.

lipis:
If we need a community demand then we can add the `Help Needed` label also to the exempt list.. but I've seen a bunch of them that we await response and response is not there..

just voting 👍 I don't think that it counts as activity.. but of course if someone comments that means it's _updated_.. I've seen this bot in many projects.. it's quite friendly :)

azz:
A lot of the issues are minor "ugly code" issues. They're actionable but quite low priority. We often link to duplicates. Won't stalebot close these issues?

lipis:
The bot is not closing them right away.. first it will post a comment there so all the affected parties will be notified.. and if anyone is going to comment that is needed or let's wait (or whatever) we have 90 more days until the next ping.. if nothing happens in 7 days the bot will close it down.. but usually humans will be closing the issues and not the bot.. unless nobody cares about a specific issue anyway :)

Of course once installed.. we might see a bunch of issues popping up because of that.. but it will also going to be a good time to revisit these issues and add the right label to it if needed more attention.. otherwise let's close some issues, to not ever reach 500 or 1000 open issues..

j-f1:
It would be really useful to have a custom bot that runs on each release and posts a comment if Prettier‘s output has changed in the example in the OP. Additionally, if the output matches the input, it should add a label indicating that we need to add a regression test for this.

lipis:
After we are done with this bot, we should create another one that actually fixes the bugs and creates new PRs with the fixes.. 🍻 🍷 💨 😄

suchipi:
I think this bot is cool, but it doesn't fit our current triage workflow, so we shouldn't add it.

j-f1:
`no-response` might be a good idea though — we can use it to automatically close issues that are <kbd>status:awaiting response</kbd>.

lipis:
@j-f1 There is no such option to include only specific labels.. https://github.com/probot/stale#usage

j-f1:
I was talking about https://probot.github.io/apps/no-response/ :)

lipis:
I update the PR with the no-response bot..

azz:
21 days seems very generous. Why not 5-7?

lipis:
I'm up for a week as well tbh..

lydell:
I think 5–7 is too little. People go on vacations sometimes :) Perhaps a compromise, then? 14 days? On the other hand, if a _bot_ is going to enforce the time limit it doesn't hurt if it's a little longer, does it?

azz:
🤷‍♂️ Lets go with 2 weeks and see how we go.

lipis:
14 it is :) we can adjust it later if needed.. or remove it completely if it's too aggressive.

lipis:
[Nice and polite](https://github.com/prettier/prettier/issues?q=is%3Aissue+is%3Aclosed+sort%3Aupdated-desc).. it works :)

PR #3885, opened 2018-02-04
Add prettierrc to prettier source code

xixixao:
I have default config in `prettier-atom`, and since this repo doesn't include `.prettierrc`, `prettier-atom` was formatting the code differently.

To avoid this, I added `.prettierrc` with all the `prettier` defaults (as per the JSON schema).

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,10 @@
+{
+  "bracketSpacing": true,
+  "jsxBracketSameLine": false,
+  "printWidth": 80,
+  "semi": true,
+  "singleQuote": false,
+  "tabWidth": 2,
+  "trailingComma": "none",
+  "useTabs": false
+}


.prettierrc
```diff
@@ -0,0 +1,10 @@
+{
+  "bracketSpacing": true,
+  "jsxBracketSameLine": false,
+  "printWidth": 80,
+  "semi": true,
+  "singleQuote": false,
+  "tabWidth": 2,
+  "trailingComma": "none",
+  "useTabs": false
```
lipis: Missing the: `arrowParens`, `proseWrap`, `requirePragma`

azz:
Seems like an issue with the extension rather than prettier itself. We intentionally don't have a config file as we use defaults for everything.

lipis:
It's a tricky one.. as it's kind of makes sense, because `prettier-atom` is not aware of the fact that this project is using Prettier so it falls back to user's preferences..

azz:
Shouldn't a config file containing `{}` just work? Defaults should _not_ be applied if there's a config file found. @robwise can you confirm this?

I'd be ok with adding `"prettier": {}` to `package.json` to fix this workflow.

robwise:
@azz The problem is that if you call `resolveConfig` and it resolves to an empty config file, it returns `null` instead of default options so prettier-atom can't tell the difference between no config and an empty config. So we added this: 



Of course, this is an entirely optional config option. Alternatives in prettier-atom are to always format on save or to always format on save if prettier is in your package.json.

azz:
> The problem is that if you call resolveConfig and it resolves to an empty config file, it returns null

Not if the file contains `{}`, right?

robwise:
@azz I haven't tried without the `{}`, so yes it still just returns `null` if I have only `{}` in the file.

.prettierrc
xixixao: @lipis Those are not present in the JSON schema: http://json.schemastore.org/prettierrc

xixixao:
My 2 cents:
1. The defaults don't change that often, so tracking them here shouldn't be a problem.
2. This gives a nice example to people on how to use `.prettierrc`, they can just copy and modify.

azz:
> I haven't tried without the {}, so yes it still just returns null if I have only {} in the file.

That sounds like a bug to me, I think we have tests to the contrary.

> This gives a nice example to people on how to use .prettierrc, they can just copy and modify.

We want to encourage people to roll with the defaults _unless_ they have reason to change each setting. A configuration file with 10 values where 8 of them are the default is not particularly useful. We also want to ensure that people know Prettier works perfectly _without_ a config file.

lipis:
What about just adding `"prettier": true` to the `package.json`? (works in VS Code)

robwise:
@olsonpm is this related to the empty (or just `{}`) file returning `null`? https://github.com/davidtheclark/cosmiconfig/issues/106

olsonpm:
@robwise - cosmiconfig will return an empty object upon seeing `{}` in `.prettierrc`.  Prettier does the same, so I'm confused at what is different in our setups.

My test consists of the file directory
```sh
test-prettier/
  .prettierrc

test.js
```

where `.prettierrc` contains just `{}` and `test.js` is:
```js
// test.js
const path = require('path'),
  prettier = require('prettier');

prettier.resolveConfig(path.join(__dirname, 'test-prettier'), {
  useCache: false
})
.then(result => {
  console.log(result);
  // prints '{}'
});
```

Though I still don't understand why `{}` is appropriate over an empty file given they support yaml syntax as well, and my understanding is {} is an empty config in json whereas an empty file is an empty config in yaml.  I don't use yaml though so i may misunderstand.

robwise:
@olsonpm I dunno maybe I was just misreading my node output or something, nvm.

olsonpm:
yeah no worries - unfortunately it seems the confusion surrounding empty config files resulted in [the decision to treat the cli differently from the api](https://github.com/prettier/prettier/pull/3255#issuecomment-348420546).  Making matters more confusing, cosmiconfig treats empty files differently in its own repo (throws an error when given a config file directly vs passing over a config file while traversing up directories).

Anyway, it's safe to say an empty *rc or config file is an error case that would ideally be shown to the user for them to address.

azz:
Closing this for reasons mentioned in https://github.com/prettier/prettier/pull/3885#issuecomment-362911470 and https://github.com/prettier/prettier/pull/3885#issuecomment-363024255. If there's an issue with the way editor integrations are set up we should fix those at the root cause.

xixixao:
@azz What do you suggest we do instead?

suchipi:
@xixixao the prettier repo has eslint-plugin-prettier set up, so if you enable eslint integration in the atom package, it should format correctly.

PR #3887, opened 2018-02-04
Add more plugins to the plugins section

lipis:
Just to give a bit more exposure to some of the other plugins.. I'm aware of the fact that they might be far from done, but promoting them from here could attract more people.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -38,7 +38,7 @@ If the plugin is unable to be found automatically, you can load them with:
 * [`@prettier/plugin-php`](https://github.com/prettier/prettier-php)
 * [`@prettier/plugin-swift`](https://github.com/prettier/prettier-swift)
 
-## Unofficial Plugins
+## Community Plugins
 
 * [prettier-plugin-java](https://github.com/thorbenvh8/prettier-java)
 * [iamsolankiamit/prettier-ruby](https://github.com/iamsolankiamit/prettier-ruby)
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -40,7 +40,7 @@ If the plugin is unable to be found automatically, you can load them with:
 
 ## Unofficial Plugins
 
-* [@thorbenvh8/prettier-java](https://github.com/thorbenvh8/prettier-java)
+* [prettier-plugin-java](https://github.com/thorbenvh8/prettier-java)
 * [@iamsolankiamit/prettier-ruby](https://github.com/iamsolankiamit/prettier-ruby)
 * [@benjie/prettier-plugin-pg](https://github.com/benjie/prettier-plugin-pg)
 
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -38,6 +38,12 @@ If the plugin is unable to be found automatically, you can load them with:
 * [`@prettier/plugin-php`](https://github.com/prettier/prettier-php)
 * [`@prettier/plugin-swift`](https://github.com/prettier/prettier-swift)
 
+## Unofficial Plugins
```
lipis: Other ideas for the title are more than welcome

docs/plugins.md
```diff
@@ -38,6 +38,12 @@ If the plugin is unable to be found automatically, you can load them with:
...
+
+* [@thorbenvh8/prettier-java](https://github.com/thorbenvh8/prettier-java)
```
azz: Please use the npm package names here.

lipis: should I just remove the `@` or what?

the following work..

```
yarn add thorbenvh8/prettier-java
# or
npm install --save-dev thorbenvh8/prettier-java
```

azz: `prettier-plugin-java`: https://www.npmjs.com/package/prettier-plugin-java

lipis: the other two are not on NPM and for the same reason even Python is not correct.. (https://www.npmjs.com/package/@prettier/python)

We can just point to the GitHub repositories until they are actually released, as we are looking for early adopters/testers and not the actual users.

azz: Yes, like to GitHub, but use the npm package names as the names.

docs/plugins.md
```diff
@@ -38,6 +38,12 @@ If the plugin is unable to be found automatically, you can load them with:
 * [`@prettier/plugin-php`](https://github.com/prettier/prettier-php)
 * [`@prettier/plugin-swift`](https://github.com/prettier/prettier-swift)
 
+## Unofficial Plugins
+
+* [prettier-plugin-java](https://github.com/thorbenvh8/prettier-java)
+* [@iamsolankiamit/prettier-ruby](https://github.com/iamsolankiamit/prettier-ruby)
```
azz: The `@`s are still incorrect as it implies it is an npm package name. Either `iamsolankiamit/prettier-ruby` or `prettier-plugin-ruby` is fine.

lipis: Really confused.. :| Is this fine now?

docs/plugins.md
suchipi: "Community Plugins" might be a bit more friendly

lipis: Much better!

lipis:
@azz?

PR #3899, opened 2018-02-06
Upgrade flow to 0.64

### LINKED ISSUE #3898: JSX spread operator syntax error on the flow parser
**Prettier 1.10.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAUgZQBoD4A6UABEcAHQVwAmA5nAM4C+hqA9FniADQgQAOMAJbR6yUAEMATpIgB3AApSEolOIBuEQVW4hx9GMgBm4gDb04PAEaTxYANZwYmPrcFQayGJICuFkFQgwI1NzHjdzSRh5GxoAW3Fgsz8AK3oADwAhG3tHTHFYuAAZNzhE0N5vGD5KgCYyvxdJCOQQQxM5HT5JNxgAdS0YAAtkAA4ABh4uiHNemz4WroY4STVSnkk4AEdvQQ3o8TiEpGMknnNYwU8fP3o3GhM4AEVvCHh6nhhxS36qIeQaj42QQmO4AYQgsXiLSg0DWIG85gAKl8VCdzIxGEA)
```sh
--parser flow
```

**Input:**
```jsx
<JSX>
  {...edges}
</JSX>
```

**Output:**
```jsx
SyntaxError: Unexpected token ... (2:4)
  1 | <JSX>
> 2 |   {...edges}
    |    ^^^
  3 | </JSX>
```

It works in babylon, so my guess is that we just need to upgrade flow and it should just go away.

vjeux:
This also enables spread operator on jsx on the flow parser:

```js
<div>{...list}</div>
```

I had no idea it was even a thing.

I had to comment out a flow test that no longer parses, but it's no longer in the flow repo and the comment said that it shouldn't be working anyway so I don't think it's going to be a problem.

Fixes #3898

### DIFF
--- a/tests/jsx_spread/jsfmt.spec.js
+++ b/tests/jsx_spread/jsfmt.spec.js
@@ -1,1 +1,1 @@
-run_spec(__dirname, ["babylon"]);
+run_spec(__dirname, ["babylon", "flow"]);
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "escape-string-regexp": "1.0.5",
     "esutils": "2.0.2",
     "find-project-root": "1.1.1",
-    "flow-parser": "0.59.0",
+    "flow-parser": "0.64.0",
     "get-stream": "3.0.0",
     "globby": "6.1.0",
     "graphql": "0.12.3",
--- a/scripts/build/rollup.parser.config.js
+++ b/scripts/build/rollup.parser.config.js
@@ -24,7 +24,7 @@ export default Object.assign(baseConfig, {
     // by its value before bundling.
     parser.endsWith("flow")
       ? replace({
-          "require(s8)": 'require("fs")',
+          "require(tf)": 'require("fs")',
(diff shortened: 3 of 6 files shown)


j-f1:
Can you add a test for the new syntax?

j-f1:
Also looks like there are a bunch of errors in the build.

tests/jsx_spread/jsfmt.spec.js
```diff
@@ -1 +1 @@
-run_spec(__dirname, ["babylon"]);
+run_spec(__dirname, ["babylon", "flow"]);
```
vjeux: @j-f1 there's already a test in the codebase: jsx_spread. I enabled it for the flow parser

vjeux:
Oops... I was trying out the https://github.com/sindresorhus/refined-github extension that has a merge when green... but it merged instantly :(

PR #3907, opened 2018-02-06
Normalize SVGs using svgo

lipis:
We can also minify them completely and leave only one script in the `package.json`


suchipi:
Whoops, I meant to approve that and I accidentally hit merge :x

lipis:
No problem by me :)

lipis:
The question remains if we want to keep it simpler... keep only one of the two scripts..
https://github.com/prettier/prettier/blob/master/website/package.json#L7

suchipi:
I like the SVGs you had in the diff. So we could remove the minify one.

lipis:
#3908

PR #3937, opened 2018-02-09
Mention vscode-status-bar-format-toggle in the docs

azz:
Ref #3634


Tom-Bonnike:
Ohhh totally missed that. Thanks for adding it to the docs! :) That explains the sudden surge in downloads.

PR #3938, opened 2018-02-09
Document new plugin API features

czosel:
* locStart / locEnd functions
* util-shared

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -64,14 +64,16 @@ export const languages = [
 
 Parsers convert code as a string into an [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree).
 
-The key must match the name in the `parsers` array from `languages`. The value contains a parse function and an AST format name.
+The key must match the name in the `parsers` array from `languages`. The value contains a parse function, an AST format name, and two location extraction functions (`locStart` and `locEnd`).
 
 ```js
 export const parsers = {
   "dance-parse": {
     parse,
     // The name of the AST that
-    astFormat: "dance-ast"
+    astFormat: "dance-ast",
+    locStart,
+    locEnd
   }
 };
 ```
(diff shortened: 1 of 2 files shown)


docs/plugins.md
```diff
@@ -135,6 +143,18 @@ function embed(
 
 If you don't want to switch to a different parser, simply return `null` or `undefined`.
 
+### Utility functions
+
+A `util` module from prettier core is considered a private API and is not meant to be consumed by plugins. Instead, the `util-shared` module provides the following limited set of utility functions for plugins:
```
azz: We should note that it is an argument to ` genericPrint`

docs/plugins.md
```diff
@@ -135,6 +143,18 @@ function embed(
...
+
+```ts
+makeString(rawContent: string, enclosingQuote: string, unescapeUnnecessarEscapes: boolean): string;
+getNextNonSpaceNonCommentCharacterIndex(text: string, node: object, options: object): number;
+isNextLineEmptyAfterIndex(text: string, index: number): boolean;
+isNextLineEmpty(text: string, node: object, options: object): boolean;
+mapDoc(doc: object, callback: function): void;
```
j-f1: Do you want to 📖 document these now or in another PR?

docs/plugins.md
czosel: Actually, it isn't anymore - i removed it a later version of the PR because we're passing `options` to every functions it exports - which means one can just import the module directly.

docs/plugins.md
czosel: It might make sense to add some minimal documentation for them now? I'd be happy if you could contribute to that, I'm not too familiar with some of them 😉

czosel: See also https://github.com/prettier/prettier-php/pull/14/files for a usage example

docs/plugins.md
azz: Ah cool, that's probably better. We might want to expose it as `require('prettier').util` as we bundle the code before publishing to npm.

czosel: Right - i think that's why CI is currently failing for my PR in the PHP plugin 😅 Shall i add it to this PR?

azz: Sure!

PR #3943, opened 2018-02-09
fix(css): scss interpolation

evilebottnawi:


lydell:
What does this PR fix?

evilebottnawi:
@lydell 
Before:
```scss
a {
  line-height: #{$var} em;
  line-height: 1 #{$var};
}
```

After:
```scss
a {
  line-height: #{$var}em;
  line-height: 1#{$var};
}
```

evilebottnawi:
@lydell right now don't normalize, i can do this in other PR, right now output as is

lydell:
How does the logic work? `if ("something[nospace]#{" in declarationValue) { printEntireThingAsIs(declarationValue) }`?

evilebottnawi:
@lydell add tests. What tests is still worth adding?

lydell:
I see. Personally, I think it's fine to add tests that print badly right now. It's nice when you can see in PRs how the snapshots changed for the better. But if you want to do it in another PR I guess that's fine too.

evilebottnawi:
@lydell let's add more tests :+1:

PR #3946, opened 2018-02-10
feat(css): improve at rule output

evilebottnawi:
Just workaround before implement better at rule parser


evilebottnawi:
@lipis i think it is bug solution, someone use this like (content: "  my string  ") and it is remove spaces, close this PR, try to search better solution. Thanks!

PR #3958, opened 2018-02-12
tests(css): colors

evilebottnawi:
Just tests.


tests/css_color/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -15,6 +49,27 @@ exports[`hexcolor.css 1`] = `
   -webkit-color: #873;
   -moz-color: #6bcdef;
   -ms-color: #aabbcc;
+  color: #F09F;
```
evilebottnawi: Bug in `postcss-values-parser`, PR - https://github.com/shellscape/postcss-values-parser/pull/44. After merge i will update version and snapshots.

PR #3966, opened 2018-02-13
Clarify that both packages are still needed

sharils:


suchipi:
You only need both if you want eslint to warn you about prettier violations. If you only want to disable formatting rules, but rely on prettier list different to identify violations, then you don't need both.

suchipi:
Using eslint-config-prettier without eslint-plugin-prettier provides a better experience in the editor in my opinion, but using eslint-plugin-prettier makes a lot of sense in CI or in projects with lots of public contributors where there's a chance they'll have eslint tooling configured but not prettier tooling. So it all depends on your use-case.

docs/eslint.md
```diff
@@ -57,3 +57,9 @@ There are a few rules that this disables that you may want to turn back on as lo
   "extends": ["plugin:prettier/recommended"]
 }
 ```
+
+Remember to install both `eslint-plugin-prettier` and `eslint-config-prettier`:
+
+```bash
+yarn add --dev prettier eslint-plugin-prettier eslint-config-prettier
```
lydell: Can you remove `prettier` from this line?

sharils:
@suchipi I suggested this patch because it's not clear in the [Why not both?](https://prettier.io/docs/en/eslint.html#why-not-both) section, there it didn't mention whether you should install both or either of them if you want to utilise the `plugin:prettier/recommended` settings. The patch is not implying in anyway that both should always be installed.

sharils:
@lydell Done. Thanks you for spotting that! :)

j-f1:
@lydell ✅/❌?

suchipi:
My bad @sharils, I didn't look at the context of where it's appearing. This makes total sense

Issue #3970, opened 2018-02-13
Angular.component.html files. Can we ignore specific file extensions?


Fallenstedt:
**Prettier 1.10.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEACdqA8BnADgQygD4AdKDDTARwFd8AneegWmzgBs4wYyL0AKGBADmwzgEU6jOPQCyCGgFEAtgEsYTAJQBeEiCGiJUpvKg1+AEjgA3BDE17effmAAWhYXEkMmK9Vt1wdyhPb2l6Sxs7BxAnChdgzwBNOAY-DRkdPTcPOBSGSNtYGLiMAG1VbDCmAHlcBABdQMrqmTqER3I+MvYIfAATVRCAZRh8eCa9XoGh4VHxuE6+VDKwGnp6O3FJ8HXN2HEl7rWNu3z6HZP9mHOjijL8dnYABQhsbFUAI05zuphsHaPF5vD7fPKpeh-bB3cq4EFfIw+GTYACSUHOOzh7wRXmMyLRt1iXXuADNVOwmADAmSKci9ERMAB6WhIlhsTjcUjEygk-BUVD8VwcXBZfQiMRwAASwsK0XpTN5VC5FCZeEIRBAABoQBBcDBVNBsMhQAx6BAAO7PBgII0ofDWCCqfpakD4bAwZC89hsbWfej4MAAazgMGGBDAs2QMHoNDg2v6EDAnsePpAQzYjGe-uEynwye9cZAACtsAAPABC-qDIeG+GUcAAMkM4PnUxAaDBcB2AEytwsEegZ5D6ACe9WwYHoqj1LtwU9gAHUnTBXMgABwABm1c7ecAX-tww7ncAzthdm1oqk2WfwObzSC9qbYaijMcLHxCRgg8D72rGnyXfoV2Qbs-39clZgAYQgZRc2HKBoBbbUaDYAAVfBPltR84AAXxwoA)
```sh
--parser typescript
```

**Input:**
```tsx
    <span>
      <quarter-select
      (toggleQuarterMenuEmitter)="toggleQuarterMenu($event)"
      (changeQuarterEmitter)="changeQuarter($event)"
      (changeYearEmitter)="changeYear($event)"
      [isQuarterOpen]="isQuarterOpen"
      [loadingState]="loadingState"
      [currentQ]="currentQ"
      [currentYear]="currentYear"
[...]

PR #3971, opened 2018-02-13
Add example .prettierignore file

Fallenstedt:
Adds more detail on how to use .prettierignore so people like me don't clog up issues with simple questions.

#3970

### DIFF
--- a/docs/ignore.md
+++ b/docs/ignore.md
@@ -11,8 +11,8 @@ To exclude files from formatting, add entries to a `.prettierignore` file in the
 
 For example: 
 
-<!-- .prettierignore -->
-```
+```sh
+# .prettierignore
 # ignores files with an .html extension
 *.html 
 ```


docs/ignore.md
```diff
@@ -9,6 +9,14 @@ Prettier offers an escape hatch to ignore a block of code or prevent entire file
 
 To exclude files from formatting, add entries to a `.prettierignore` file in the project root or set the [`--ignore-path` CLI option](cli.md#ignore-path).
 
+For example: 
+
+```sh
+# .prettierignore
+# ignores files with an .html extension
+*.html
```
suchipi: Prettier doesn't format HTML out of the box, so this is a bit misleading- could we maybe change the example to `*.test.js`?

suchipi: (Not that choosing not to format your tests is very common. Not sure what a better glob would be...)

j-f1: `*.min.js`?

azz: And/or
```js
dist
package.json
```

azz:
Ping @Fallenstedt. Would you be able to update the PR or check the "Allow edits from maintainers" checkbox?

Fallenstedt:
@azz This is from an `unknown repository` and I cannot find the "Allow edits from maintainers checkbox". I have no way of updating this PR without creating a new PR. Is that ok?

j-f1:
Yeah, let’s close this and open a new one, since I don’t think there’s a way to change this PR now.

PR #3984, opened 2018-02-16
feat(css): improve math output

evilebottnawi:
Feel free to feedback.

In short:
1. Always space before and after `*`.
2. Always space before and after `*` and `/` in calc function.
3. Don't touch `+` and `-` inside in `calc`, spec require always use space, but we don't prettify invalid syntax.
4. Always space before and after `/` if before or after `func`, `word`.
5. Refactor.
6. A lot of tests.
7. Ignore all in `url`.


lipis:
#4063 :D

PR #3987, opened 2018-02-16
feat(css): improve svg tags

evilebottnawi:
All tested in browser, all works as expected

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "resolve": "1.5.0",
     "semver": "5.4.1",
     "string-width": "2.1.1",
+    "svg-tag-names": "1.1.1",
     "html-tag-names": "1.1.1",
     "typescript": "2.7.0-insiders.20171214",
     "typescript-eslint-parser": "13.0.0",
--- a/src/language-css/clean.js
+++ b/src/language-css/clean.js
@@ -1,6 +1,11 @@
 "use strict";
 
 const htmlTagNames = require("html-tag-names");
+const svgTagNames = require("svg-tag-names");
+
+const lowercasedSvgTagNames = svgTagNames.map(svgTagName => {
+  return svgTagName.toLowerCase();
+});
 
 function clean(ast, newObj) {
   if (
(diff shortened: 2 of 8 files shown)


lydell:
I don't think it is safe to do this. Selectors will change from not matching anything to potentially matching.

evilebottnawi:
@lydell do you show example, for me it is very safe, in theory you can use `lInE` and `LiNeArGrAdIeNt` and it is valid

lydell:
See #3304.

Scenario: Somebody has accidentally left behind this piece of broken CSS in their stylesheet:

```css
LinearGradient {
  stop-color: inherit;
}
``

Then they upgrade Prettier, which would change it to:

```css
LinearGradient {
  stop-color: inherit;
}
``

Suddenly the gradient of soms SVG on their site has changed color. Which means that Prettier broke their site. Prettier is supposed to be 100% safe to run.

**Comment by evilebottnawi** [comment]
@lydell for `css` `LinearGradient` === `linearGradient` === `lInEaRgRaDiEnt`

**Comment by evilebottnawi** [comment]
@lydell http://cssdeck.com/labs

```html
<svg>
  <!-- Градиент -->
  <linearGradient id="linear-gradient">
    <stop offset="0%" stop-color="gold"/>
    <stop offset="100%" stop-color="teal"/>
  </linearGradient>

  <!-- Фигура с градиентной заливкой -->
  <rect fill="url(#linear-gradient)"
        width="100%" height="100%"/>
</svg>
```

```css
// Try to change case and look on result
svg LinearGradient stop {
  stop-color: blue;
}
```

evilebottnawi:
@lydell hm you are right, but according spec firefox do not works correctly, yep svg is xml, but when you output all in browsers it is html5. Some svg features don't allowed when you use inline svg in html5 (due to incompatibility). But we can't do here nothing :disappointed:

lydell:
For reference (copied from #3317):

https://www.w3.org/TR/css3-selectors/#casesens

> All Selectors syntax is case-insensitive within the ASCII range (i.e.
> [a-z] and [A-Z] are equivalent), except for parts that are not under the
> control of Selectors. The case sensitivity of document language element
> names, attribute names, and attribute values in selectors depends on the
> document language. For example, in HTML, element names are
> case-insensitive, but in XML, they are case-sensitive.

PR #3988, opened 2018-02-16
fix: Format Intersection same as Union

### LINKED ISSUE #3986: Typescript: Inconsistent union and intersection type multiline formatting
**Prettier 1.10.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAkrOAZzjBgEtoAVTHAXmwDU4AnNAGWgHNqsBBbAGSMW7LjzgAhQcNYco3GgGFpTWWJoARFSLkKsAUQA6UY+izYAqlApRx2eqtHzx-AD4yneydnePd45V8ddSwtILVnGn0QABoQCAxyaEJkUABDZmYIAHcABQyEFJQ0gDcIMgATWJA0whhkADM0gBtiOIAjZjSwAGs4GABlDG6yeWQYZgBXODiKiDBGlraQUeJmGFyuzgBbNMXWmZAAK0IADwku3v6BtO24NlG4feWISZgMN4AmZ8Ph5jXkKgaIQwMwyIlqhgwbAAOqVGAAC2QAA4AAxxKEQYgwroYQFQogsEpPOLMOAAR0mZDJmzSOz2SCaBzixG2ZHGU0OhFGnGacAAipMIPAfnEYGl2nCKojkJ8xV0yM0eYoINtdoCoNASSBJsRKBKikziABfY1AA)
```sh
--parser typescript
```

**Input:**
```tsx
type IntesectionType = VeryLongTypeA & VeryLongTypeB & VeryLongTypeC & VeryLongTypeD & VeryLongTypeE

type UnionType = VeryLongTypeA | VeryLongTypeB | VeryLongTypeC | VeryLongTypeD | VeryLongTypeE
```

**Output:**
```tsx
type IntesectionType = VeryLongTypeA &
  VeryLongTypeB &
  VeryLongTypeC &
  VeryLongTypeD &
  VeryLongTypeE;

type UnionType =
  | VeryLongTypeA
  | VeryLongTypeB
  | VeryLongTypeC
  | VeryLongTypeD
  | VeryLongTypeE;

```

**Expected behavior:**

```
type IntesectionType = 
  & VeryLongTypeA
  & VeryLongTypeB
  & VeryLongTypeC
  & VeryLongTypeD
  & VeryLongTypeE;

type UnionType =
  | VeryLongTypeA
  | VeryLongTypeB
  | VeryLongTypeC
  | VeryLongTypeD
  | VeryLongTypeE;

```

I think the Interections type should use the same format as the one used for the Union type.

albertorestifo:
Share the same logic to format Interection and Union types in TypeScript
and Flow.

Closes #3986

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -680,11 +680,13 @@ const rootEpic = (actions, store) =>
     });
 
 // Two extra levels of indentation because of the comment
-export type AsyncExecuteOptions = child_process$execFileOpts & {
-  // The contents to write to stdin.
-  stdin?: ?string,
-  dontLogInNuclide?: ?boolean
-};
+export type AsyncExecuteOptions =
+  & child_process$execFileOpts
+  & {
+      // The contents to write to stdin.
+      stdin?: ?string,
+      dontLogInNuclide?: ?boolean
+    };
 
 // optional trailing comma gets moved all the way to the beginning
 const regex = new RegExp(
--- a/tests/flow/intersection/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow/intersection/__snapshots__/jsfmt.spec.js.snap
@@ -79,12 +79,14 @@ function hasObjectMode_ok(options: DuplexStreamOptions): boolean {
  * @flow
  */
 
-type DuplexStreamOptions = ReadableStreamOptions &
-  WritableStreamOptions & {
-    allowHalfOpen?: boolean,
-    readableObjectMode?: boolean,
-    writableObjectMode?: boolean
-  };
+type DuplexStreamOptions =
+  & ReadableStreamOptions
+  & WritableStreamOptions
+  & {
+      allowHalfOpen?: boolean,
+      readableObjectMode?: boolean,
+      writableObjectMode?: boolean
+    };
 
 function hasObjectMode_bad(options: DuplexStreamOptions): boolean {
   return (
--- a/tests/flow/union_new/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow/union_new/__snapshots__/jsfmt.spec.js.snap
@@ -207,15 +207,19 @@ function printAll(val: MyType) {
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 type Common = {};
 
-type A = Common & {
-  type: "A",
-  foo: number
-};
+type A =
+  & Common
+  & {
+      type: "A",
+      foo: number
+    };
 
-type B = Common & {
-  type: "B",
-  foo: Array<number>
-};
+type B =
+  & Common
+  & {
+      type: "B",
+      foo: Array<number>
+    };
 
 type MyType = A | B;
 
@@ -260,15 +264,19 @@ function printAll(val: MyType) {
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 type Common = {};
 
-type A = {
-  type: "A",
-  foo: number
-} & Common;
+type A =
+  & {
+      type: "A",
+      foo: number
+    }
+  & Common;
 
-type B = {
-  type: "B",
-  foo: Array<number>
-} & Common;
+type B =
+  & {
+      type: "B",
+      foo: Array<number>
+    }
+  & Common;
 
 type MyType = A | B;
 
@@ -321,13 +329,17 @@ type DataBase = {
   name: string
 };
 
-type UserData = DataBase & {
-  kind: "user"
-};
+type UserData =
+  & DataBase
+  & {
+      kind: "user"
+    };
 
-type SystemData = DataBase & {
-  kind: "system"
-};
+type SystemData =
+  & DataBase
+  & {
+      kind: "system"
+    };
 
 type Data = UserData | SystemData;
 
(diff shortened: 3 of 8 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -680,11 +680,13 @@ const rootEpic = (actions, store) =>
...
-  stdin?: ?string,
-  dontLogInNuclide?: ?boolean
-};
+export type AsyncExecuteOptions =
+  & child_process$execFileOpts
+  & {
+      // The contents to write to stdin.
+      stdin?: ?string,
+      dontLogInNuclide?: ?boolean
+    };
```
vjeux: It looks worse to me

tests/flow/intersection/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -79,12 +79,14 @@ function hasObjectMode_ok(options: DuplexStreamOptions): boolean {
...
-    writableObjectMode?: boolean
-  };
+type DuplexStreamOptions =
+  & ReadableStreamOptions
+  & WritableStreamOptions
+  & {
+      allowHalfOpen?: boolean,
+      readableObjectMode?: boolean,
+      writableObjectMode?: boolean
+    };
```
vjeux: This is the only one that is arguable, but in my experience it's rare that you combine two variables with an object

suchipi:
I much prefer the current behavior

tests/flow/union_new/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -321,13 +329,17 @@ type DataBase = {
   name: string
 };
 
-type UserData = DataBase & {
```
azz: Strongly prefer the inline version and agree with @vjeux's comments.

j-f1:
How about inlining (for both operators) an identifier/member expression at the beginning xor at the end? i.e.

```ts
type foo = Foo & {
  bar: Baz
}

type bar =
  & Foo
  & Bar
  & {
      baz: Quux
    }
```

albertorestifo:
A real-world example of when this formatting comes in handy, is when combining React HOC in a GraphQL-Apollo app, where I needed to construct a type combining many verbose types:

```ts
// When creating an HOC
type Props = Subtract<InputProps, WithURLStateProps> &
  RouteComponentProps<ProjectViewRouteParams>;

// A component that uses many HOC
type WithMergeIntentsMutationProps = WithAddIntentToClassifierProps &
  MergeIntentsProps &
  RouteComponentProps<ProjectViewRouteParams>;
```

[Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACACgJwgwGdsBebAZQFcAjGfAQzBgB4BJKDamAo4gDTYA6gEsYACwCqAJQAylGI3h8SAPmwAybDIg84AYQgBbDNAS9CJVnwBWcFgDVRcAO679uRk2PE1AHShsbED0LBFxCQBZOHwAczhOeFhiKJ5lUWhVUgoxSQBBABNCpIsAFQgDABtGYmJRADMXfGytbBj4xNgLYlbtD3gjU3NYbJtCeycXdz0Vb0ZfNRABECIYTKhiZFBvQlcvfAQtlEYANwhRQuWQWphkBsYq4jgV2iYwAGs4GEoMZlEoHFkAxqC8QIUIGB7o9nisAc98JZGHFjIxoU8wbZiAAPABC7y+PwWcDkALg6Nhqx43BgACYKWC-vgEchUJg4MQwPhRBg7isMNzYGJCpJkAAOAAM-MIz2ETAwrIFHNip3JK0OAEdqKJDgRkaiGStnsZRMD8KCjQC4lU4ABFagQeCG1CMWjC0VIWkrBiMURVK1DA0oKDma7UZ5lV3HB4YgC+saAA)

It might not be a common use case, but the formatting proposed this PR makes this scenario much easier to read IMO.

I like the solution proposed by @j-f1

vjeux:
Thanks for the concrete use case. This is the case where it's indeed debatable as to which is the best way. I've personally never seen anyone write those with a leading & in the past. It also feels confusing to me: if there's a space after `&`, it looks like a binary operator but there's nothing before the first one.

Not sure how involved it would be to implement but adding a newline after `=` in this case may not be a bad solution:

```js
// A component that uses many HOC
type WithMergeIntentsMutationProps =
  WithAddIntentToClassifierProps &
  MergeIntentsProps &
  RouteComponentProps<ProjectViewRouteParams>;
```

In any case, I'm reasonably happy with the current output and don't think that we should drastically change it.

j-f1:
How do we feel about adding some heuristics to avoid breaking in some cases (for example, in `Foo & { ... }`) and merging this?

vjeux:
@j-f1 we already shouldn’t break for

```js
type A = Foo & {
  a: int
};

**Comment by j-f1** [comment]
We do, however, break for multiple identifiers:

**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAgtgXmwDEIJsAybAIQEMAnSm2gLyeAB0pttalsAlrC4BfANxcQAGhAQMMAdADOyUA3oQA7gAUGCFSloA3CAIAm0kACN6tMAGs4MAMoY7QgObIY9AK5wZISU4ehhtWw8AW1pkADNaABtgmQArJQAPalsHJ2daSLgAGSE4OMTkkDd6YPpka1orNAToSwx6IRgAdXMYAAtkAA4ABhk2iGDO2ww6trgao1KZejgAR18BZfDaKJikeKSAkGDIgW8-Q6VPBLgARV8IeDKDmRgG7rM+5AAmF9sBBM8AGEIJFonUoNBFiBfMEACoNAz7CpmCBgJ4VWhKGDow4QXwwDD4r44kQiIA)

**Input:**
```jsx
type A = Foo & Bar & Baz & {
  a: int
};

```

**Output:**
```jsx
type A = Foo &
  Bar &
  Baz & {
    a: int
  };

```

albertorestifo:
I'll close this as inconclusive

Issue #3972, opened 2018-02-14
Change option defaults in plugins


czosel:
Is it possible to overwrite existing options defaults in plugins?

Example: In the PHP plugin, we'd like to change `tabWidth` from `2` to `4` (https://github.com/prettier/prettier-php/issues/25)

I saw that #3615 might be related, but it seems that it covers only the case where plugins introduce new options.

azz:
I don't think it works with what we have now (happy to be proven wrong though, not 100% sure). Two ways come to mind:

1. Export a new option in the plugins `options` array and fix it in Prettier core to trump the existing option.

2. Add a separate `optionDefaults` that intentionally overrides some default values of an existing **core** option. (`tabWidth`, `useTabs`, or `printWidth`).

I think the latter might be a better way to go.

```js
exports.optionDefaults = {
  tabWith: 2,
};
```

ikatyang:
Option 2 looks good to me but it should only affect the specific plugin, e.g.

   ```js
   // php-plugin.js
   exports.optionDefaults = {
     tabWith: 2,
   };
   ```

   implicitly become something like:

   ```json5
   // .prettierrc
   {
     "overrides": [
       {
         "files": "*.php", // extname from supportLanguages
         "options": {
           "tabWith": 2
         }
       }
     ]
   }
   ```

ikatyang:
For reference, closed #3972 via #3991.

PR #3991, opened 2018-02-17
Allow plugins to override default options

czosel:
This is a first attempt to handle #3972. See https://github.com/prettier/plugin-php/pull/52 for a usage example. I attached `defaultOptions` to the printer to reuse existing logic (`getPrinter`). Like that, only files handled by the respective plugin should be affected by the changed defaults.

This doesn't handle multiparser support, though - but maybe that's something to be handled in the future when there is a concrete use case?


azz:
I think we want the API to be

```js
exports.defaultOptions = {...}
```

Rather than:

```js
exports.printers.php.defaultOptions = {...}
```

Because we may want to show this information in the CLI. e.g.:

```shellsession
$ prettier --help tab-width
--tab-width <int>

  Number of spaces per indentation level.

Default: 2
Plugin defaults:
* @prettier/plugin-php: 4
```

Do you have any thoughts on this @ikatyang?

czosel:
That makes sense! I just pushed some changes that add the plugin's options to a new property `pluginOptions` in `supportOptions`, which should make it easy to display them in the CLI's help. I read from the same object in `options.js` to extract the current defaults. 

In this implementation a plugin's default options are registered under the name of it's first print - which works because currently all plugins only have one printer. Maybe there's a better solution, for example a `name` property on the plugin?

If you think this is generally the right way to move forward, i'll continue by cleaning up the (currently a bit clumsy) implementation and add some tests!

src/common/support.js
```diff
@@ -249,6 +249,17 @@ function getSupportInfo(version, opts) {
...
 
       return newOption;
+    })
+    .map(option => {
+      const filteredPlugins = plugins.filter(
+        plugin => plugin.defaultOptions && plugin.defaultOptions[option.name]
+      );
+      const pluginDefaults = filteredPlugins.reduce((reduced, plugin) => {
+        const printerName = Object.keys(plugin.printers)[0];
+        reduced[printerName] = plugin.defaultOptions[option.name];
```
ikatyang: We probably need to attach the plugin name (package name, plugin filepath, or the `name` property if provided) in `load-plugins.js` and use it here as an ID.

https://github.com/prettier/prettier/blob/9643d3370233f2870c43b81d0855833fb983d9f4/src/common/load-plugins.js#L6

src/main/options.js
```diff
@@ -55,6 +55,21 @@ function normalize(options, opts) {
   rawOptions.locStart = parser.locStart;
   rawOptions.printer = getPrinter(rawOptions);
 
+  const pluginDefaults = supportOptions
+    .filter(
+      optionInfo =>
+        optionInfo.pluginDefaults && optionInfo.pluginDefaults[options.parser]
```
ikatyang: Same here, since the parser name might not equal to the printer name.

czosel:
@ikatyang Right - i updated the implementation.

src/cli/util.js
```diff
@@ -613,7 +613,13 @@ function createDetailedUsage(context, optionName) {
       ? `\n\nDefault: ${createDefaultValueDisplay(optionDefaultValue)}`
       : "";
 
-  return `${header}${description}${choices}${defaults}`;
+  const pluginDefaults =
+    option.pluginDefaults && Object.keys(option.pluginDefaults).length
+      ? `\nPlugin defaults:${Object.keys(option.pluginDefaults).map(
+          key => `\n* ${key}: ${option.pluginDefaults[key]}`
```
ikatyang: We should call `createDefaultValueDisplay(value)` here, though there is no non-primitive value in core options.

src/main/options.js
```diff
@@ -53,7 +53,23 @@ function normalize(options, opts) {
...
+    )
+    .reduce(
+      (reduced, optionInfo) =>
+        Object.assign(reduced, {
+          [optionInfo.name]: optionInfo.pluginDefaults[plugin.name]
+        }),
+      {}
+    );
+
+  Object.assign({}, defaults, pluginDefaults);
```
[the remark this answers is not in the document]
czosel: Oops, thats embarassing :sweat_smile: Probably still have to figure out how to rewrite it that i don't have to use `let` (and test is also still missing)

tests_integration/__tests__/early-exit.js
```diff
@@ -26,6 +26,26 @@ describe(`show detailed usage with --help l (alias)`, () => {
...
 
+describe(`show detailed usage with plugin options`, () => {
+  runPrettier("cli", [
+    "--help",
+    "tab-width",
+    "--plugin=../plugins/automatic/node_modules/prettier-plugin-bar",
+    "--parser=bar"
+  ]).test({
+    status: 0,
+    stdout: `--tab-width <int>
```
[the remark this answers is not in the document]
czosel: Of course - didn't think that doing that would be so easy :+1:

src/main/options.js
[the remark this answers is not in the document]
ikatyang: You can just assign it to a new variable, e.g. `mixedDefaults` or something else:

```js
const mixedDefaults = Object.assign({}, defaults, pluginDefaults);

// Object.keys(mixedDefaults) ...
```

tests_integration/__tests__/__snapshots__/early-exit.js.snap
```diff
@@ -514,6 +514,21 @@ exports[`show detailed usage with --help write (stdout) 1`] = `
...
+exports[`show detailed usage with plugin options (stderr) 1`] = `""`;
+
+exports[`show detailed usage with plugin options (stdout) 1`] = `
+"--tab-width <int>
+
+  Number of spaces per indentation level.
+
+Default: 2
+Plugin defaults:
+* ../plugins/automatic/node_modules/prettier-plugin-bar: 4
```
azz: Could we use the `package.json` name here? (Happy to do this in a follow up PR though)

ikatyang: That's the result from `--plugin=../plugins/automatic/node_modules/prettier-plugin-bar`, I guess it should be fine?

- `--plugin=path/to/plugin.js` (path)
   
  ```
  Plugin defaults:
  * path/to/plugin.js: 4
  ```

- `--plugin=prettier-plugin-bar` (package name)
   
  ```
  Plugin defaults:
  * prettier-plugin-bar: 4
  ```

azz: Ah, cool. I'm happy with that. Does it work for plugins that are automatically resolved?

ikatyang: It should work. @czosel could you add a test for that? just clone the test case and set its cwd to `path/to/plugins/automatic` then remove the `--plugin=xxx` part.

czosel: done!

src/main/options.js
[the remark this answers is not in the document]
czosel: right - done.

tests_integration/plugins/automatic/node_modules/prettier-plugin-bar/index.js
```diff
@@ -19,5 +19,8 @@ module.exports = {
     bar: {
       print: path => concat(["bar+", path.getValue().text])
     }
+  },
+  defaultOptions: {
```
azz: @ikatyang mentioned this above, but could you add an integration test to ensure that this `tabWidth` is applied?

azz:
LGTM other than missing an integration test.

src/main/options.js
```diff
@@ -53,14 +53,30 @@ function normalize(options, opts) {
...
+      {}
+    );
 
   Object.keys(defaults).forEach(k => {
     if (rawOptions[k] == null) {
       rawOptions[k] = defaults[k];
     }
   });
 
+  Object.assign(rawOptions, pluginDefaults);
```
czosel: Just realized that `rawOptions` already contains the default options, so assigning the "mixedDefaults" only if the option key is not present in rawOptions (line 73) won't work. I added some tests for it.

tests_integration/plugins/defaultOptions/plugin.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+module.exports = {
+  languages: [
+    {
+      name: "foo",
+      parsers: ["foo-parser"],
+      extensions: [".foo"],
+      since: "1.0.0"
```
azz: Can delete `since` now.

src/main/options.js
czosel: Hmm, if I'm not mistaken the order of option application should be
```
core defaults < plugin defaults < cli / file overrides
```
That would mean that we already have to find out which plugin is used before CLI options are applied, so the logic has to be moved to somewhere around here:

 https://github.com/prettier/prettier/blob/master/src/cli/util.js#L224

src/cli/util.js
```diff
@@ -835,11 +844,13 @@ function updateContextOptions(context, plugins) {
...
-        Object.assign(reduced, { [optionInfo.name]: optionInfo.default }),
-      Object.assign({}, optionsModule.hiddenDefaults)
-    );
+    .reduce((reduced, optionInfo) => {
+      // default are applied in main/options.js when pluginDefaults are present
+      if (Object.keys(optionInfo.pluginDefaults).length) {
+        return reduced;
+      }
+      return Object.assign(reduced, { [optionInfo.name]: optionInfo.default });
+    }, Object.assign({}, optionsModule.hiddenDefaults));
```
czosel: I managed to get the tests passing by adding this workaround - but I'm not really happy with it. I could imagine that there's a case where this breaks, and even if there isn't it doesn't look very clean to me.

I'd be glad if someone could explain the bigger picture on the way defaults are handled right now: To me, the two-step process of applying defaults seems confusing:

1. CLI defaults are applied in `cli/util.js`
https://github.com/prettier/prettier/blob/9ee113cd2e66dad5567ce03170d1c6c6dff1a310/src/cli/util.js#L226
2. Defaults are applied if an option is `null` in `main/options.js` 
https://github.com/prettier/prettier/blob/9ee113cd2e66dad5567ce03170d1c6c6dff1a310/src/main/options.js#L58

I saw that there are some use cases where the core options are still `null` in step two, but I'm not sure what those use cases are, and consequently how the proper process of applying defaults should look like in the future.

ikatyang: --> #4045

czosel:
Workaround is reverted, tests are passing :tada:

docs/plugins.md
```diff
@@ -149,6 +155,33 @@ function embed(
...
+```js
+options: {
+  openingBraceNewLine: {
+    type: "boolean",
+    category: "Global",
+    default: true,
+    description: "Move open brace for code blocks onto new line."
+  }
+}
+```
```
j-f1: Can you document each key in the option?

docs/plugins.md
```diff
@@ -149,6 +155,33 @@ function embed(
...
+
+### `defaultOptions`
+
+If your plugin requires different default values for some of prettier's core options, you can specify them in `defaultOptions`:
+
+```
```
j-f1: ` ```js `?

docs/plugins.md
czosel: Of course! Can you help my out with the `category` option? I'll do this in a new PR.

Issue #3985, opened 2018-02-16
`requirePragma` is not compatible with `rangeStart` and `rangeEnd`


JamesHenry:
<strike>**Note:** Cannot provide a playground as `rangeStart` and `rangeEnd` are not available options atm.</strike> EDIT by @josephfrazier: playground link from https://github.com/prettier/prettier/pull/3996#issuecomment-388919910: [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6AVOgOlABO3AAQAcAnOGGASzlJ31RwDcBDU3eAZxlwF4PSAVzggANCAjFq0TslBtSEAO4AFNglkoWTCFQAmYkACNSLMAGsKAZWJmqUAObIYQuOPudaMFaYcBbFmQAMxYAG09xACtOAA8AIVMLaxY-OAAZexEkEPC3EFtST1JkYxYjAE9Q6EMyexgAdX0YAAtkAA4ABnEyCE9602ISsjgiphFxcgBHQSpyHxZ-QKQXYXFPPypnVzX7B1C4AEVBCHhgsIiQGDLGvRbkACZxFxYqUN2AYQg-AJKoaHGQIJPAAVMqaHIXUyOOAAUSgBiQAGYAKwTFjQqxXUgwB4dAC+eKAA)

Raised through use of `precise-commits` here (https://github.com/nrwl/precise-commits/issues/9), but I can manually reproduce using the prettier CLI directly.

**Prettier 1.10.2**

```sh
# Options (if any):
--require-pragma=true
--range-start={SOME_CHAR}
--range-end={SOME_CHAR}
```

**Input:**
```jsx
// any valid pragma-containing source
```

**Output:**
Prettier makes no changes to the file/input

**Expected behavior:**
Prettier formats the range for the file/input (as long as the file contains the required pragma)

PR #3996, opened 2018-02-18
Don't format range if required pragma is missing outside range

josephfrazier:
Fixes https://github.com/prettier/prettier/issues/3985

`formatRange()` is only called from `formatWithCursor()`,
which already checks for the pragma.

EDIT: @JamesHenry, could you try this out and verify that it's working for you?

### DIFF
--- a/src/main/core.js
+++ b/src/main/core.js
@@ -227,14 +227,14 @@ function formatRange(text, opts) {
 }
 
 function format(text, opts) {
-  if (opts.rangeStart > 0 || opts.rangeEnd < text.length) {
-    return formatRange(text, opts);
-  }
-
   const selectedParser = parser.resolveParser(opts);
   const hasPragma = !selectedParser.hasPragma || selectedParser.hasPragma(text);
   if (opts.requirePragma && !hasPragma) {
     return { formatted: text };
+  }
+
+  if (opts.rangeStart > 0 || opts.rangeEnd < text.length) {
+    return formatRange(text, opts);
   }
 
   const hasUnicodeBOM = text.charCodeAt(0) === UTF8BOM;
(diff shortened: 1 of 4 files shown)


JamesHenry:
My naive initial assumption was that the solution would be to check if the source text/file contained the pragma, and then format the range within it if appropriate... (instead of silently setting requirePragma to false regardless of user config)

WDYT?

j-f1:
What’s the status on this? Should we fix the conflicts and merge or close it?

josephfrazier:
Hmm, looks like the then-relevant code has since changed a decent bit. @JamesHenry, can you still reproduce the issue with the latest Prettier?

JamesHenry:
I think my question about the decision to just set the `requirePragma` option is still outstanding, and now that range options are available in the playground (they weren't at the time of the original issue and this PR), it is easy enough to reproduce that the formatting will still _not_ occur in the range (unless the range itself includes the pragma comment):

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6AVOgOlABO3AAQAcAnOGGASzlJ31RwDcBDU3eAZxlwF4PSAVzggANCAjFq0TslBtSEAO4AFNglkoWTCFQAmYkACNSLMAGsKAZWJmqUAObIYQuOPudaMFaYcBbFmQAMxYAG09xACtOAA8AIVMLaxY-OAAZexEkEPC3EFtST1JkYxYjAE9Q6EMyexgAdX0YAAtkAA4ABnEyCE9602ISsjgiphFxcgBHQSpyHxZ-QKQXYXFPPypnVzX7B1C4AEVBCHhgsIiQGDLGvRbkACZxFxYqUN2AYQg-AJKoaHGQIJPAAVMqaHIXUyOOAAUSgBiQAGYAKwTFjQqxXUgwB4dAC+eKAA

duailibe:
Nevermind the comment, that’s not how range works

JamesHenry:
Excellent, thanks!

PR #4012, opened 2018-02-20
Allow new interface style for GraphQL.

### LINKED ISSUE #3600: GraphQL: Support new multiple inheritance syntax

ruiaraujo:
This is a breaking change since it will upgrade the old style to the new one.

Closes #3600.

### DIFF
--- a/src/language-graphql/parser-graphql.js
+++ b/src/language-graphql/parser-graphql.js
@@ -34,11 +34,19 @@ function removeTokens(node) {
   return node;
 }
 
+function fallbackParser(parse, source) {
+  try {
+    return parse(source, { allowLegacySDLImplementsInterfaces: false });
+  } catch (_) {
+    return parse(source, { allowLegacySDLImplementsInterfaces: true });
+  }
+}
+
 function parse(text /*, parsers, opts*/) {
   // Inline the require to avoid loading all the JS if we don't use it
   const parser = require("graphql/language");
   try {
-    const ast = parser.parse(text);
+    const ast = fallbackParser(parser.parse, text);
     ast.comments = parseComments(ast);
     removeTokens(ast);
     return ast;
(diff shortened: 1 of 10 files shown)


src/language-graphql/parser-graphql.js
```diff
@@ -34,11 +34,19 @@ function removeTokens(node) {
   return node;
 }
 
+function fallbackParser(parse, source) {
+  try {
+    return parse(source, { allowLegacySDLImplementsInterfaces: false });
+  } catch (_) {
+    return parse(source, { allowLegacySDLImplementsInterfaces: true });
```
duailibe: It'd be great if we could detect which style the code was originally written so we could support both styles, and we can with this code.

But I don't think it's worth the trouble though.

ruiaraujo: I would agree with that.

suchipi:
Will users on older versions of GraphQL libraries be inconvenienced if prettier starts using the new style? As in, will their libraries not be able to parse the new `&` style? If so, we should probably go through the effort to detect the style from the source.

ruiaraujo:
Yes, I would consider this change a breaking change and consumers should be made aware.

If you disagree and think that we should support it regardless, I would ask for your suggestions on how to do it since the AST provided does not include any notion of separator.

azz:
We have to check the source (`options.originalText`)

ruiaraujo:
@suchipi @azz I have done it. Feel free to take another look.

ruiaraujo:
Should I worry about the codecov/project check?

suchipi:
Nah

PR #4019, opened 2018-02-21
Add husky, lint-staged for SVG files

lipis:
In order to force the new SVG files to be formatted with SVGO before commiting.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -1,7 +1,11 @@
 {
   "license": "MIT",
+  "lint-staged": {
+    "*.svg": ["yarn svgo", "git add"]
+  },
   "scripts": {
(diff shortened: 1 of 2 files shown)


lipis:
I was also in doubt if that would work, but it works as is.

screen shot 2018-02-21 at 16 32 44

azz:
Is svgo idempotent? Feels a bit weird adding commit hook for files that aren't really ever updated.

j-f1:
Given the rate that we change SVGs, it would probably be easier to ask contributors to run SVGO after they open PRs.

lipis:
Yeah.. maybe we don't need it.

PR #4021, opened 2018-02-21
Add support for markdown with pre-commit

magicmark:
Prettier now runs on markdown files! Let's let pre-commit run on those files!

(There may be additional file extensions that we should also add?)

cc @asottile

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,5 +2,5 @@
     name: prettier
     entry: prettier --write
     language: node
-    # From https://github.com/prettier/prettier/blob/133303f47a30f6b3e46ffdf9d5c2d6609d65c416/src/options.js#L32-L42
-    files: \.(css|less|scss|html|ts|tsx|graphql|gql|json|js|jsx)$
+    # From https://github.com/prettier/prettier/blob/7a7eb170/docs/index.md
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -2,5 +2,5 @@
     name: prettier
     entry: prettier --write
     language: node
-    # From https://github.com/prettier/prettier/blob/133303f47a30f6b3e46ffdf9d5c2d6609d65c416/src/options.js#L32-L42
-    files: \.(css|less|scss|html|ts|tsx|graphql|gql|json|js|jsx)$
+    # From https://github.com/prettier/prettier/blob/7a7eb170824afa8ad8efe531532ed8d7cd6cc965/docs/index.md
```
j-f1: Can you shorten the hash to `7a7eb170` so it’s not as long?

magicmark:
ack, rebasing

j-f1:
Thanks for contributing to Prettier!

asottile:
fwiw, it looks like there's a bunch more that are supported:

https://github.com/prettier/prettier/blob/7a7eb170824afa8ad8efe531532ed8d7cd6cc965/src/language-markdown/index.js#L18-L28

PR #4023, opened 2018-02-21
refactor: at rules parse

evilebottnawi:
Note:
1. Use `postcss-media-query-parser` only for standard css at rules with params, because we can break other code (example postcss plugin which add own at rules or less).
2. Don't parse unknown at rules, it is also can break code (including change quotes, numbers and etc).
3. Refactor code to improve readability

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -310,26 +310,10 @@ function parseNestedCSS(node) {
         return node;
       }
 
-      if (
-        node.name === "charset" ||
-        node.name.toLowerCase() === "counter-style" ||
-        node.name.toLowerCase().endsWith("keyframes") ||
-        node.name.toLowerCase() === "page" ||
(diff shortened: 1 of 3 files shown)


j-f1:
I don’t really understand the CSS parser, sorry :(

lydell:
Can you show a couple examples of cases where we will avoid breaking code, and cases where we will miss out on normalizing quotes and numbers and stuff? Makes it easier to understand the trade-off.

evilebottnawi:
@lydell I see custom postcss at rule as `@font "Name font" src("path/to/my.eot") !woff !woff2`, `postcss-media-query-parser` only for standard css rules, we should avoid using this for non standard and print non standard as is. Using this strategy we don't break code never. Better create issue about handle specific at rule and if it is popular we can integrate in prettier

evilebottnawi:
By default we account all standard css at rules, Less, SCSS and some custom popular at rules it is 99% enough

PR #4024, opened 2018-02-21
refactor(css): printer

evilebottnawi:
Just refactor


evilebottnawi:
Slightly decreased coverage, but it is normal, some construction very rary example:
```js
if (!parentNode) {
  return false;
}
```

ParentNode always true if you are not root, but it is very very rary

PR #4025, opened 2018-02-21
Handle definite variable declarations

azz:


src/language-js/printer-estree.js
```diff
@@ -3204,6 +3204,12 @@ function printTypeAnnotation(path, options, print) {
   }
 
   const parentNode = path.getParentNode();
+  const printDefinite =
```
j-f1: `isDefinite`?

PR #4028, opened 2018-02-22
feat: ignore non standard IE filter values

evilebottnawi:
Why?
1. It is non standard.
2. It is break standard css specification.
3. Rare usage.
4. Many logic around this = decrease perf
5. I don't found right syntax. 

In 7-8 IE (with newlines) don't work, in IE 9 work :confused: 
```
progid:DXImageTransform.Microsoft.gradient(
  enabled='false', 
  startColorstr='#fff', 
  endColorstr='#000'
);
```
6. Very buggy property :smile: 

Let's output this property as is

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -279,6 +279,10 @@ function parseNestedCSS(node) {
           }
         }
 
+        if (value.startsWith("progid:")) {
+          return node;
+        }
+
         node.value = parseValue(value);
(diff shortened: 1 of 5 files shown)


evilebottnawi:
@j-f1 tests? We Already have tests around `progid:`

j-f1:
Can you add some tests of things that are no longer broken across lines?

lipis:
We should update the release notes #4035

PR #4035, opened 2018-02-23
1.11 Release Notes and Changelog

suchipi:
Here is the proposed Release Notes blog post and Changelog for the 1.11 release. Could someone please proofread it?

Rendered version here:
https://deploy-preview-4035--prettier.netlify.com/blog/2018/02/26/1.11.0.html

PR #4036, opened 2018-02-24
Bump own prettier dependency version to 1.11.0-rc.1

suchipi:
Context: https://github.com/prettier/prettier/pull/4035#issuecomment-368173012

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "eslint-plugin-react": "7.1.0",
     "jest": "21.1.0",
     "mkdirp": "0.5.1",
-    "prettier": "1.10.2",
+    "prettier": "1.11.0-rc.1",
     "prettylint": "1.0.0",
     "rimraf": "2.6.2",
     "rollup": "0.47.6",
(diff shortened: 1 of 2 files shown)


suchipi:
Yup!

package.json
```diff
@@ -97,7 +97,8 @@
     "test": "jest",
     "test:dist": "node ./scripts/test-dist.js",
     "test-integration": "jest tests_integration",
-    "lint": "cross-env EFF_NO_LINK_RULES=true eslint . --format node_modules/eslint-friendly-formatter",
```
azz: I think we don't usually format this file. (related: #3671). Should we add `package.json` to `.prettierignore`?

suchipi: I usually do so in my own projects. But I think maybe we should make prettier ignore package.json out of the box.

lipis: and usually for JSON files I'm having a huge printWidth..

azz: That's not really how updating JSON works. One of the big advantages of JSON over (e.g. YAML) is it is easy to make a program to read, patch, and write the file back to disk. `JSON.stringify` is the standard tool for this job and I don't think a couple of line breaks added by Prettier warrant the the added complexity of extra config and linters.

suchipi: For what it's worth, the only reason I disable prettier for package.json in my projects is because of this behavior it has with long string keys, since it only ever seems to introduce diffs

suchipi:
Superceded by https://github.com/prettier/prettier/pull/4058

PR #4038, opened 2018-02-24
Do not require plugins to set 'since' field

azz:
Fixes the issue raised in https://github.com/prettier/plugin-php/issues/55

### DIFF
--- a/src/common/support.js
+++ b/src/common/support.js
@@ -255,15 +255,12 @@ function getSupportInfo(version, opts) {
 
   const languages = plugins
     .reduce((all, plugin) => all.concat(plugin.languages), [])
-    .filter(language => {
-      if (language.since === null) {
-        return false;
-      }
-      if (language.since) {
-        return semver.gte(version, language.since);
-      }
-      return true;
-    })
+    .filter(
+      language =>
+        language.since
+          ? semver.gte(version, language.since)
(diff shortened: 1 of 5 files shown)


src/main/options.js
```diff
@@ -81,10 +81,9 @@ function inferParser(filepath, plugins) {
     pluginsLoaded: true
   }).languages.find(
     language =>
-      typeof language.since === "string" &&
```
ikatyang: `language.since !== null`? We should omit unreleased languages here.

src/common/support.js
```diff
@@ -255,7 +255,15 @@ function getSupportInfo(version, opts) {
...
-    .filter(language => language.since && semver.gte(version, language.since))
+    .filter(language => {
+      if (language.since === null) {
+        return false;
+      }
+      if (language.since) {
+        return semver.gte(version, language.since);
+      }
+      return true;
+    })
```
ikatyang: How about

```js
language.since
  ? semver.gte(version, language.since)
  : language.since !== null
```

?

azz: Yeah I originally had a ternary but it looked messy/hard to reason about so I switched it to if/else.

src/main/options.js
azz: Good catch.

PR #4039, opened 2018-02-25
webpack/webpack now uses Prettier

azz:
[`webpack/webpack-cli`](https://github.com/webpack/webpack-cli) has used Prettier for a while, but now [`webpack/webpack`](https://github.com/webpack/webpack) does, too!

https://github.com/webpack/webpack/pull/6081

### DIFF
--- a/website/data/users.yml
+++ b/website/data/users.yml
@@ -23,10 +23,10 @@
   greyImage: /images/users/used_by_babel.svg
   infoLink: https://babeljs.io
   pinned: true
-- caption: Webpack CLI
+- caption: Webpack
   image: /images/users/webpack-200x100.png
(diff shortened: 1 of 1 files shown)


website/data/users.yml
```diff
@@ -23,10 +23,10 @@
   greyImage: /images/users/used_by_babel.svg
   infoLink: https://babeljs.io
   pinned: true
-- caption: Webpack CLI
+- caption: Webpack
```
j-f1: I’m pretty sure it’s “webpack” (lowercase w).

azz: https://github.com/prettier/prettier/commit/79196030a78954359cb7885ba32b984e0e960510

PR #3990, opened 2018-02-17
feat(markdown): respect tabWidth for list items

ikatyang:
Context: https://github.com/prettier/prettier/issues/3459#issuecomment-366382620

This PR improves the indenting experience in VSCode (probably other editors as well) and also fixes #3459, I'm not sure if this is the proper fix but at least we can see the result and decide if it should be closed/merged.

- align list prefix with `tabWidth`
- it only affect ordered lists if `tabWidth: 2`, otherwise all lists.

vscode indent demo

cc @kachkaev

---

**Prettier pr-3990**
[Playground link](https://deploy-preview-3990--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBGAdAAgMIAsCGUA5nJjBJgJ4QCuATprhALakAmAlnXGOXZUgA6UYZjHixYNsOEAmLABkOAZxhlcpSLAQxlQkVAkSANsswBaYyAA0ICAAcYHaMuSh8dOhADuABQ8Irij4AG4QHGw2IPiqyABm+KZwtgBGdPhgANZwMADK9hkcxMgwdDTJIGwQYPGJyhVF9XQwvulEzPi1SbYAVsoAHgBC6Vk5ufisSlBwXfW2tDD2NDCyJWUVBXRNyCAddJlV3lBR9nRFMADqETC4yAAcAAy2pxD1F+n2O6dwTSEzttwAI40LhwVr4dqdJAJbogerMDhrcq2ZRFIjGOAARRoEHgswqMHwKSubBuyFktlK+A4xjR2BYHR2UGg-xANHqABUiUEYfUAL58oA)
```sh
--parser markdown
```

**Input:**
```markdown
1. Change to your home directory:

        cd

2. List the contents:

        ls -l
```

**Output:**
```markdown
1.  Change to your home directory:

        cd

2.  List the contents:

        ls -l

```

PR #4041, opened 2018-02-25
fix(markdown): list prefix alignment: count spaces from correct place

ikatyang:
4+ spaces will cause indented codeblock but I mis-count its range in #3990, it was over-strict.

before

```
1.···123 <---- (count: 3) paragraph 
  ^^^
1.····123 <--- (count: 4) paragraph
  ^^^^
1.·····123 <-- (count: 5) indented codeblock
  ^^^^^
```

after

```
1.····123 <---- (count: 3) paragraph
   ^^^
1.·····123 <--- (count: 4) indented codeblock
   ^^^^
1.······123 <-- (count: 5) indented codeblock
   ^^^^^
```


src/language-markdown/printer-markdown.js
```diff
@@ -357,12 +357,11 @@ function printListItem(path, options, print, listPrefix) {
 }
 
 function alignListPrefix(prefix, options) {
-  const prefixTrailingSpaces = prefix.match(/ *$/)[0].length;
   const additionalSpaces = getAdditionalSpaces();
   return (
     prefix +
     " ".repeat(
-      prefixTrailingSpaces + additionalSpaces >= 4 ? 0 : additionalSpaces // 4+ will cause indented code block
+      additionalSpaces >= 4 ? 0 : additionalSpaces // 4+ will cause indented codeblock
```
j-f1: `code block` feels more correct here to me.

PR #4042, opened 2018-02-25
fix(markdown): list prefix: no unnecessary trailing spaces

ikatyang:


src/language-markdown/printer-markdown.js
```diff
@@ -233,7 +233,11 @@ function genericPrint(path, options, print) {
 
       return printChildren(path, options, print, {
         processor: (childPath, index) => {
-          const prefix = alignListPrefix(
+          const listItem = childPath.getValue();
+          const alignFunc = listItem.children.length
+            ? alignListPrefix
+            : identity; // do not print trailing spaces for empty list item since it might be treated as `break` node by [doc-printer](https://github.com/prettier/prettier/blob/master/src/doc/doc-printer.js#L470-L480), we don't want to preserve unnecessary trailing spaces.
+          const prefix = alignFunc(
```
j-f1: Would it make sense to refactor this into:

```js
let prefix = ...
if (listItem.children.length) {
  prefix = alignListPrefix(prefix)
}
```
?

PR #4055, opened 2018-02-26
docs: note on defaults which may change

bakkot:
There seems to be consensus in #3503 that these options will change in Prettier 2.0. I think it's a good idea to point that out.

That issue also mentions removing `--parser postcss`, but I'm not sure how to properly note that.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -52,6 +52,10 @@ Valid options:
 
 ## Quotes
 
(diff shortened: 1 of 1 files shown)


bakkot:
I'll go ahead and close this, then. I still think this is a good idea to do if before the 2.0 release drops, though; the problems React ran into don't look like they'd be that relevant.

PR #4084, opened 2018-03-01
chore: update `postcss-scss` to latest version

evilebottnawi:
Tests included


duailibe:
If I understand correctly, this fixes some issue we were having with SCSS? If so, can you explain them in the PR description so it'll be easier add a note to a future changelog?

Thanks!

evilebottnawi:
@duailibe we ignore selector which contain `comments`, it is just upgrade package and add tests for future, no changelog.

duailibe:
@evilebottnawi Yes you are right. My bad :)

PR #4088, opened 2018-03-02
pre-commit.com is now served on https

asottile:
This commit was automated with the following script:

```bash
git grep -l 'pre-commit\.com' |
    xargs sed -i 's|http://pre-commit\.com|https://pre-commit.com|g'
```

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -67,7 +67,7 @@ Copy the following config into your `.pre-commit-config.yaml` file:
         -   id: prettier
 ```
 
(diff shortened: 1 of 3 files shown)


asottile:
I wasn't sure about the `.snap` file so I edited it anyway 🤷‍♀️

duailibe:
Yeah there was no need to update that.. it has probably diverged already from the actual docs and we don't really sync them.

But there's no harm in updating it anyway so 👍

PR #4089, opened 2018-03-02
No longer need to clarify that `pre-commit` is "Python version"

asottile:
Since pretty-quick replaced the _other_ pre-commit: https://github.com/prettier/prettier/pull/3675


asottile:
Part of the point of pre-commit is that it doesn't matter that it's written in python :)

PR #4101, opened 2018-03-05
Add more stuff to the Rationale page

lydell:
I tried to write something down for common issues.


docs/rationale.md
```diff
@@ -44,3 +142,4 @@ Here are a few examples of things that are out of scope for Prettier:
 * Turning single- or double-quoted strings into template literals or vice versa.
 * Adding/removing `{}` and `return` where they are optional.
 * Turning `?:` into `if`-`else` statements.
+* Sorting and hoisting `import`s. (Sorting is unsafe because of side effects, which would violate the [correctness](#correctness) goal.)
```
ericanderson: I agree on sorting but not on hoisting. Its valid ES6 to hoist imports. Having prettier do it will prevent surprises since the code will run the same way. Although it does break the AST rule. Anyway, this is complex.

PR #4104, opened 2018-03-05
chore: Upgrade some dependencies

lipis:


package.json
```diff
@@ -24,13 +24,13 @@
...
     "diff": "3.2.0",
-    "editorconfig": "0.14.2",
+    "editorconfig": "0.15.0",
     "editorconfig-to-prettier": "0.0.6",
     "emoji-regex": "6.5.1",
     "escape-string-regexp": "1.0.5",
     "esutils": "2.0.2",
     "find-project-root": "1.1.1",
-    "flow-parser": "0.64.0",
+    "flow-parser": "0.66.0",
```
duailibe: Luckily this didn't break anything but we can't simply upgrade flow, we special case the build process because of a dynamic require

https://github.com/prettier/prettier/blob/8e6b9dd5d9723e042b8949c318b971ed23a1763a/scripts/build/rollup.parser.config.js#L22-L30 (unfortunately the comment is outdated)

Issue #4143, opened 2018-03-13
SCSS code is different with single line comments inside mixins using CRLF line endings


aholtkamp:
**Prettier 1.11.1**

Using prettier on our SCSS files we experience different behaviour under linux than under windows. Further investigation revealed that the line endings make the difference. It seems when using CRLF as line endings prettier produces different output than with normal LF as line endings.

**Input (CRLF line endings):**
```scss
@mixin create-rules($padding) {
  //
  // single line comment
  //
  .abc {
    padding: $padding;
  }

  /**
   * multi line comment
   */
  .def {
    padding: $padding;
  }
}

```

**Output:**
```scss
@mixin create-rules($padding) {
  $p // single line comment
  //
  .abc {
    padding: $padding;
  }
  /* *
   * multi line comment */
  .def {
    padding: $padding;
  }
}

```

**Expected behavior:**
The file should not be changed as it already is properly formatted.

**Issues:**

- single line comments produce invalid code after formatting
- multi line comments are formatted incorrectly
- a line break between the two code blocks gets stripped

evilebottnawi:
@aholtkamp i add tests https://github.com/prettier/prettier/pull/4145 and all works as expected, what editor you use and how you run prettier, looks something wrong with cursor

aholtkamp:
I'm using IntelliJ IDEA. It seems your tests do not use "@mixin" but media-queries. For media-queries everything works as expected.

evilebottnawi:
@aholtkamp add you example

evilebottnawi:
@aholtkamp maybe you can create minimum reproducible repo or describe instruction how you run prettier?

aholtkamp:
I will see if I can provide a repo, probably in the next few hours

aholtkamp:
Oh, I'm sorry, it seems I have changed and formatted my code too often when trying to provide a minimal example for the issue (our affected file is much bigger).

The problem occurs when the mixin parameter list too long and will be split into multiple lines.

**Input (CRLF line endings):**
```scss
@mixin create-rules(
  $padding,
  $param2: 30px,
  $param3: 1px solid black,
  $param4: 1,
  $param5: 1
) {
  //
  // single line comment
  //
  .abc {
    padding: $padding;
  }

  /**
   * multi line comment
   */
  .def {
    padding: $padding;
  }
}

```

I hope you can reproduce it with this code, otherwise I will provide a repo.

aholtkamp:
[...]

PR #4145, opened 2018-03-14
tests: comments with `CRLF`

evilebottnawi:
For #4143


tests/css_comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,188 @@
...
+
+@mixin create-rules($padding) {
+  //
+  // Comment 8
+  //
+  .abc {
+    padding: $padding;
+  }
+  /**
+     * Comment 9
```
evilebottnawi: Ugly, but let's fix it in other PR

PR #4159, opened 2018-03-16
fix(playground): no undefined parsers

ikatyang:
I'm not sure how it happened, but *preview playground* seems broken:

(https://deploy-preview-4151--prettier.netlify.com/playground/)

image

image


duailibe:
I also noticed this, don't know how it happened :/

PR #4161, opened 2018-03-16
Add elm plugin

kachkaev:
This PR adds a link to a freshly baked [`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) and also reorders community plugins so that the list is easier to glance through 🎉 

I'm really happy to bring Elm support to Prettier given that [its philosophy](https://github.com/prettier/prettier/blob/master/docs/option-philosophy.md) is influenced by how formatting is done in Elm 😄

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -40,10 +40,10 @@ If the plugin is unable to be found automatically, you can load them with:
 
 ## Community Plugins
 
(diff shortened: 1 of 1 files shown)


duailibe:
Nice work!! 💯 

Loved the Elm plugin!

PR #4163, opened 2018-03-16
Add Elm plugin to website's home page

kachkaev:
Supplements #4161

Sorry for not doing all in one PR – I thought the website is build from a different repo. Can there be other lists of plugins that I have missed?

### DIFF
--- a/website/data/languages.yml
+++ b/website/data/languages.yml
@@ -33,10 +33,10 @@
   showName: true
   image: /images/languages/tools_wip.svg
   variants:
-  - "[Python](https://github.com/prettier/prettier-python)"
-  - "[PHP](https://github.com/prettier/prettier-php)"
-  - "[Swift](https://github.com/prettier/prettier-swift)"
-  - "[Java](https://github.com/thorbenvh8/prettier-java)"
-  - "[Ruby](https://github.com/iamsolankiamit/prettier-ruby)"
-  - "[PostgreSQL](https://github.com/benjie/prettier-plugin-pg-sql/)"
   - "[Elm](https://github.com/gicentre/prettier-plugin-elm)"
+  - "[Java](https://github.com/thorbenvh8/prettier-java)"
+  - "[PHP](https://github.com/prettier/prettier-php)"
+  - "[PostgreSQL](https://github.com/benjie/prettier-plugin-pg-sql/)"
+  - "[Python](https://github.com/prettier/prettier-python)"
+  - "[Ruby](https://github.com/iamsolankiamit/prettier-ruby)"
+  - "[Swift](https://github.com/prettier/prettier-swift)"
--- a/website/data/languages.yml
+++ b/website/data/languages.yml
@@ -33,7 +33,7 @@
   showName: true
   image: /images/languages/tools_wip.svg
   variants:
-  - "[Elm](https://github.com/gicentre/prettier-plugin-elm)"
+  - "[Elm](https://github.com/gicentre/prettier-plugin-elm) (via [`elm-format`](https://github.com/avh4/elm-format))"
   - "[Java](https://github.com/thorbenvh8/prettier-java)"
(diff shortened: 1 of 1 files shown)


duailibe:
I'm OK with this since it's under "Works with the tools you use" and that's a true statement.

But I'm worried people may think we're taking advantage of others' work and advertising as ours.

kachkaev:
If you are worried that elm plugin is just a wrapper for `elm-format` but it does not mention its core tool specifically, I don't think it's a problem. When someone clicks on [`Elm`](https://github.com/gicentre/prettier-plugin-elm), the first thing they see is references to @avh4's `elm-format` and the emphasis of its dominant role in the plugin.

`prettier-plugin-elm` has some similarity with how markdown or python languages work in Prettier – both use external tools for parsing. The only difference is that Elm plugin also leverages `elm-format`'s printer, which is one extra step.

@avh4 WDYT as the author of `elm-format`?

evilebottnawi:
@kachkaev just for information, `elm-format` is support `printWidth`?

kachkaev:
@evilebottnawi nope, `elm-format` is 100% opinionated in everything and it does not support any options.

Here is the output of `elm-format --help` for `0.6.1-alpha` (latest):

```

Usage: elm-format [INPUT] [--output FILE] [--yes] [--validate] [--stdin]
                  [--elm-version VERSION] [--upgrade]
  Format Elm source files.

Available options:
  -h,--help                Show this help text
  --output FILE            Write output to FILE instead of overwriting the given
                           source file.
  --yes                    Reply 'yes' to all automated prompts.
  --validate               Check if files are formatted without changing them.
  --stdin                  Read from stdin, output to stdout.
  --elm-version VERSION    The Elm version of the source files being formatted.
                           Valid values: 0.16, 0.17, 0.18. Default: 0.18
  --upgrade                Upgrade older Elm files to Elm 0.18 syntax

Examples:
  elm-format Main.elm                     # formats Main.elm
  elm-format Main.elm --output Main2.elm  # formats Main.elm as Main2.elm
  elm-format src/                         # format all *.elm files in the src directory

Full guide to using elm-format at <https://github.com/avh4/elm-format>
```

Same set of options applies to `0.7.0-exp` (experimental release).

Discussion of `printWidth`: https://github.com/avh4/elm-format/issues/324

evilebottnawi:
@kachkaev I think we should add new section (or add note before or after) for plugins which don't use logic of `prettier` but use api for DX.

kachkaev:
That could be a good thing to add to the home page once there is at least one more plugin in this new list – otherwise it will look a bit odd. I'd also support splitting official plugins from the community ones, just like in [`plugins.html`](https://prettier.io/docs/en/plugins.html#official-plugins).

I think I understand what you mean by logic of prettier,  but I'm not sure what its border is. OK, `--print-width` is a rather universal option, but what about `--tab-width` or `--trailing-comma`? It's quite likely that not all of the languages will be supporting all the options even if some of them could fit. In this case, does a plugin follow the logic or not? What are the criteria?

duailibe:
I think there's been a bit of confusion here :-)

My comment is only because of the difference is other plugins do use external tools but the core printing logic is implemented in the plugin, and not delegated to other tools (not because they use Prettier's options or not)

I think it's great that the integration interface we created is being useful for languages that already have their own formatter. I was just worried that we're *not* advertising something we didn't create (it's also possible I'm worried for nothing and this is fine 😄)

azz:
Could we just make the list item say "Elm (`elm-format` wrapper)"?

kachkaev:
@azz done!

avh4:
elm-format author here.  The latest patch looks good to me.  (Though afaik, no other Elm formatters exist, so I doubt there would have been confusion in any case.)

azz:
image

LGTM

azz:
We might eventually change it from "Work in Progress" to list them as "Plugins"

Issue #4177, opened 2018-03-19
Enable plugin-specific comment functionality


mgrip:
https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L4

Plugins need to be able to call comment utility functions like `printDanglingComments`, however they currently aren't exported

mgrip:
also this might be a separate issue, but I believe we're going to need to abstract how comments get attached to nodes, potentially allowing plugins to define an `attachComment()` function? based on [prettier core](https://github.com/prettier/prettier/blob/master/src/main/comments.js#L491) seems like it would need to accept at least `text, precedingNode, enclosingNode, followingNode, comment, options`

mgrip:
Seems like both issues might be a hard blocker for any plugins to be released in even a semi-stable state - do others agree?

vjeux:
I don’t have any issue exporting all those.

evilebottnawi:
@mgrip can you take care around this? I don't have many time right now

mgrip:
I think the more difficult piece of this is figuring out an api for plugins to abstract the logic around how comments get attached to the ast. most of it is specific to estree https://github.com/prettier/prettier/blob/master/src/main/comments.js#L615

mgrip:
I can try and get a WIP PR together in the next couple days as a poc

vjeux:
The attach function has is generic and then delegates in three places to a lot of tree specific ones. Would be nice to make those plugins of some sort

mgrip:
Added a PR https://github.com/prettier/prettier/pull/4182 - im torn if it makes sense to export `printDanglingComments` or not, theoretically the plugins can just handle printing on their own, but not sure. I did allow plugins to define a `handleComment` function, which can call `addLeadingComment`, `addTrailingComment`, `addDanglingComment` for different kinds of nodes

PR #4182, opened 2018-03-20
feat: enable plugin-specific comment functionality

mgrip:
This change enables language plugins to provide their own "handleComment" function. To do so, we're exposing the addLeadingComment, addTrailingComment, and addDanglingComment functions which attach comment nodes correctly in the ast,  which will later get handled by prettier core when printing.

WIP - feedback and contributions are welcome


mgrip:
Addresses https://github.com/prettier/prettier/issues/4177

src/common/util.js
```diff
@@ -756,6 +756,37 @@ function arrayify(object, keyName) {
...
+  const comments = node.comments || (node.comments = []);
+  comments.push(comment);
+  comment.printed = false;
+
+  // For some reason, TypeScript parses `// x` inside of JSXText as a comment
+  // We already "print" it via the raw text, we don't need to re-print it as a
+  // comment
+  if (node.type === "JSXText") {
+    comment.printed = true;
+  }
```
vjeux: This should probably be inside of estree printer plugin

mgrip:
@vjeux do you think we should move those estree-specific handlers out into a `language-js` `handleComment` function? and if so part of this PR or a follow-up?

vjeux:
`attach` has some boilerplate that is common and then the bulk of the parser-specific logic is in all the `handleXXXX`.

https://github.com/prettier/prettier/blob/master/src/main/comments.js#L167-L193

If you look, there are three cases:

1)
```js
    if (privateUtil.hasNewline(text, locStart(comment), { backwards: true })) {
```

2) 
```js
    } else if (privateUtil.hasNewline(text, locEnd(comment))) {
```

3)
```js
    } else {
```

Ideally, plugins should be able to give in a list of handlers for those three passes but not have to reimplement the rest of the logic.

vjeux:
As for this PR or not, I don't really care. This current PR seems to be working and provide value so we can merge it as is.

mgrip:
Ahh ok that makes sense. So like `handleComment`, `handleCommentOnNewLine`, `handleCommentWithTrailingNewLine`? (Naming is hard 🙄 )

vjeux:
The idea between the three cases are:

- Comment on its own line
- Comment at the end of a line
- The rest

mgrip:
Ok got it. I'll add `handleComment`, `handleEndOfLineComment`, `handleCommentOnOwnLine`? And I can separate out the JS specific ones as part of this

vjeux:
I think that it would be helpful is the interface was an array of handlers and have the generic code go through all the handlers one by one and stop if one returns false. The same way the logic is setup with lots of else ifs.

mgrip:
Sorry not sure I understand completely - do you mean only call the plugin-specific handlers in case 3?

vjeux:
The code looks like this:

```js
      if (
        handleLastFunctionArgComments(
          text,
          precedingNode,
          enclosingNode,
          followingNode,
          comment,
          options
        ) ||
        handleConditionalExpressionComments(
          enclosingNode,
          precedingNode,
          followingNode,
          comment,
          text,
          options
        ) ||
        handleImportSpecifierComments(enclosingNode, comment) ||
        handleIfStatementComments(
          text,
          precedingNode,
          enclosingNode,
          followingNode,
          comment,
          options
        ) ||
        handleClassComments(
          enclosingNode,
          precedingNode,
          followingNode,
          comment
        ) ||
        handleLabeledStatementComments(enclosingNode, comment) ||
        handleCallExpressionComments(precedingNode, enclosingNode, comment) ||
        handlePropertyComments(enclosingNode, comment) ||
        handleExportNamedDeclarationComments(enclosingNode, comment) ||
        handleOnlyComments(enclosingNode, ast, comment, isLastComment) ||
        handleTypeAliasComments(enclosingNode, followingNode, comment) ||
        handleVariableDeclaratorComments(enclosingNode, followingNode, comment)
```

Ideally the plugin would export something that looks like:

```js
handleComments: {
  ownLine: [
    handleLastFunctionArgComments,
    handleConditionalExpressionComments,
    handleImportSpecifierComments,
    ...
  ],
  endOfLine: [
    ...
  ],
  remaining: [
    ...
  ],
}
```

mgrip:
Got it. The only thing I'm a little hesitant about is the array form sort of implies precedence, based on the order of things in the array. If `ownLine` is just a function then the plugin developer has to explicitly define precedence

vjeux:
Thanks a lot for improving the internals of prettier :)

vjeux:
I was going to suggest doing the same for locStart/locEnd but looks like it's already been done :) https://github.com/prettier/prettier/blob/13103101ebd6bc563fb6cc765186244b90d8300e/src/language-js/index.js#L11-L37

This was the last thing that I was aware needed to be plugin-ified

mgrip:
I think this was the last thing we noticed too :) At this point I don't think there's much stopping us from getting to our first stable version for the php plugin! https://github.com/prettier/plugin-php/issues/181

azz:
This is awesome, great work @mgrip!

PR #4189, opened 2018-03-21
Add Jane.com to users

zacanger:
As of today we're 100% Prettier! Thanks for all your hard work!

The change to Brigad's logo was a `chmod -x`.


zacanger:
Thank you!

PR #4190, opened 2018-03-21
Remove outdated comment

vjeux:
We now always return a conditional group.


j-f1:
Does this part of the command docs still apply?

> [`conditionalGroup`] should be used as last resort as it triggers an exponential complexity when nested.

vjeux:
Absolutely!

Issue #4179, opened 2018-03-20
The tarball on npm contains another prettier tarball


sandorfr:
**Prettier 1.11.1**

The tarball currently available on npm contains another tarball `prettier-1.11.1.tgz`. It seems like there a bug in your publish process which trigger this tarball inception. 

![image](https://user-images.githubusercontent.com/2660262/37629108-015abdfe-2c31-11e8-9518-a75fb73b9a7a.png)

**Expected behavior:**

The package on npm should not contain this tarball. it adds unnecessary overhead.

suchipi:
Sounds like a good solution

PR #4197, opened 2018-03-23
Don't publish nested tarballs

josephfrazier:
Fixes https://github.com/prettier/prettier/issues/4179

I tried the `.gitignore` idea, but it didn't work, presumably because we
publish from `dist/` instead of the repo root. See here for more details
on how/why this solution works:
https://docs.npmjs.com/files/package.json#files

### DIFF
--- a/scripts/build/build.js
+++ b/scripts/build/build.js
@@ -92,6 +92,7 @@ pkgWithoutDependencies.scripts = {
   prepublishOnly:
     "node -e \"assert.equal(require('.').version, require('..').version)\""
 };
+pkgWithoutDependencies.files = ["*.js"];
 pipe(JSON.stringify(pkgWithoutDependencies, null, 2)).to("dist/package.json");
 
 shell.echo("Copy README.md");


suchipi:
Did you try `.npmignore`? That might be more discoverable.

josephfrazier:
Good point, it would definitely be more discoverable. However, I think we'd then have to keep it synchronized with `.gitignore`. From https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package:

> Use a .npmignore file to keep stuff out of your package. If there's no .npmignore file, but there is a .gitignore file, then npm will ignore the stuff matched by the .gitignore file. If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it. Like git, npm looks for .npmignore and .gitignore files in all subdirectories of your package, not only the root directory.
> ...
> If, given the structure of your project, you find .npmignore to be a maintenance headache, you might instead try populating the files property of package.json, which is an array of file or directory names that should be included in your package. Sometimes a whitelist is easier to manage than a blacklist.

I considered putting the `files` property in the top-level package.json, but thought it might be misleading, since the `dist/` directory is flat, but the top-level one isn't.

suchipi:
I didn't know `.npmignore` took precedenceover `.gitignore` - TIL

azz:
Any reason we don't just add `"files": ["*.js"]` to `package.json` directly?

lydell:
https://github.com/prettier/prettier/pull/4197#issuecomment-375766381

> I considered putting the `files` property in the top-level package.json, but thought it might be misleading, since the `dist/` directory is flat, but the top-level one isn't.

azz:
Ah missed that, makes sense to me.

PR #4208, opened 2018-03-26
refactor: `parseNestedCSS` function

evilebottnawi:
Improve speed and fix places where we can break code

### DIFF
--- a/tests/css_case/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css_case/__snapshots__/jsfmt.spec.js.snap
@@ -357,7 +357,7 @@ $KeepScssVar: val;
   .Keep();
   .Keep(4px) !important;
   .Keep() when (@Keep=Keep);
-  .Keep() when (@Keep=12px);
+  .Keep() when (@Keep=12PX);
   .Keep() when (@Keep=Keep12PX);
 }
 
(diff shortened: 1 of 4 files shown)


tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -357,7 +357,7 @@ $KeepScssVar: val;
   .Keep();
   .Keep(4px) !important;
   .Keep() when (@Keep=Keep);
-  .Keep() when (@Keep=12px);
+  .Keep() when (@Keep=12PX);
```
evilebottnawi: In last LESS version value case sensitivity, let's output as is

tests/css_comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1303,20 +1303,20 @@ $var: /* comment 1 */ all /* comment 2 */ !default /* comment 3 */ ; /* comment
   /* comment 20 */
 }
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-$var: /* comment 1 */ all /* comment 2 */ !default /* comment 3 */ ; /* comment 4 */
+$var: /* comment 1 */ all /* comment 2 */ !default /* comment 3 */; /* comment 4 */
```
evilebottnawi: Better output comments in value properties

tests/css_less/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2690,9 +2690,7 @@ div {
 header {
   background-color: blue;
 
-  .desktop-and-old-ie({
-    background-color: red;
-  });
+  .desktop-and-old-ie({background-color: red;});
```
evilebottnawi: We use `postcss-selector-parser` and it is can break code. We should use `postcss-value-parser` for this. I can do this in other PR.

PR #4214, opened 2018-03-27
Update ISSUE_TEMPLATE.md

suchipi:
We get a lot of dupes, option requests, and external issues opened. This intends to respond to those before they are opened so that we don't have to.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -14,9 +14,7 @@ BEFORE SUBMITTING AN ISSUE:
 3. If your issue is with a prettier editor extension or add-on, please open the
    issue in the repo for that extension or add-on, instead of this repo.
 
--->
-
-<!-- For ugly or incorrect code issues: Please use the below template.
+For ugly or incorrect code issues: Please use the below template.
 
 Tip! Don't write this stuff manually.
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,21 @@
-<!-- For ugly or incorrect code issues: Please use this template.
+<!--
+
+BEFORE SUBMITTING AN ISSUE:
+
+1. Search for your issue on GitHub: https://github.com/prettier/prettier/issues
+   A large number of opened issues are duplicates of existing issues.
+   If someone has already opened an issue for what you are experiencing,
+   you do not need to open a new issue.
+   
+2. If you are requesting a new option, your request will almost certainly be denied.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,4 +1,21 @@
-<!-- For ugly or incorrect code issues: Please use this template.
+<!--
+
+BEFORE SUBMITTING AN ISSUE:
+
+1. Search for your issue on GitHub: https://github.com/prettier/prettier/issues
+   A large number of opened issues are duplicates of existing issues.
+   If someone has already opened an issue for what you are experiencing,
+   you do not need to open a new issue.
```
j-f1: How about adding an recommendation to 👍 the issue so we see people want it?

suchipi: Good idea

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,4 +1,22 @@
...
+   you do not need to open a new issue- please instead add a 👍 rating to the
+   existing issue.
+   
+2. If you are requesting a new option, your request will almost certainly be denied.
+   Please read why here: https://prettier.io/docs/en/option-philosophy.html
+
+3. If your issue is with a prettier editor extension or add-on, please open the
+   issue in the repo for that extension or add-on, instead of this repo.
+
+-->
```
lydell: I think this will break the build script: https://github.com/prettier/prettier/blob/9da875258526154ca9e727d5f351dec72610e3fc/scripts/build/build.js#L71-L84

It assumes there's just one comment in this file. I guess it's the easiest to just join the two comments into one.

j-f1: IMO we should remove the second comment since it’s basically an advertisement for the playground, which the user has already used.

suchipi: @j-f1 not necessarily- they've used it if they got here from the "create issue" button on the playground, but not if they get here by clicking GitHub's "new issue" button

suchipi: Oh, in this build code. Gotcha

suchipi:
Please don't merge until https://github.com/prettier/prettier/pull/4214#discussion_r177849177 is done

.github/ISSUE_TEMPLATE.md
lydell: Are you talking about the "Report issue" button in the playground? When I click it, there are no comments at all in the pre-filled issue.

suchipi: Oh, you're right. Then maybe this is ready? @j-f1?

Issue #4206, opened 2018-03-26
Range formatter weird indent with ClassDeclaration


CiGit:
When range formatting a ClassDefinition, there is an offset which is the same as `class`.

Original issue prettier/prettier-vscode#398

`echo -e "export default class Foo{\n/**/\n}" | prettier --range-start 16 --range-end 31`
```js
export default class Foo {
                 /**/
               }
```
This behavior differs from:
`echo -e "export default function Foo(){\n/**/\n}" | prettier --range-start 16 --range-end 31`
```js
export default function Foo() {
  /**/
}
```

Those 2 following behavior look similar:
`echo -e "/*     */      function Foo(){\n/**/\n}" | prettier --range-start 16 --range-end 31`
```js
/*     */      function Foo() {
                 /**/
               }
```
`echo -e "/*     */      class Foo{\n/**/\n}" | prettier --range-start 16 --range-end 31`
```js
/*     */      class Foo {
                 /**/
               }
```

### Expected
I expect `export default` to be part of ClassDeclaration, they shouldn't increase indentation. Behave the same as with `function`
For **comment** and **whitespace** case, I don't know what to expect 😞 It may be already what I would expect.

wKich:
Is any progression?

evilebottnawi:
It is open source, PR welcome

wKich:
@evilebottnawi can you clarify where from can I start to debug this issue?

evilebottnawi:
https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L2349

thorn0:
Closing in favor of https://github.com/prettier/prettier/issues/4926

Issue #3789, opened 2018-01-22
Format range produces additional whitespace for method of exported class


undeadcat:
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACMAGwEMBnUvAMT2AB0o89s4AzHAW2JgBUALAS1IAKAJQ16jRsxgBXbAwCMAbgl4AvvVUATaDwgywvPoNF0Gk6XIYAmSXhXmNUJ-RAAaEBAwx+0UslBibGwIAHcABSCEfxRiADcIfi13EAAjbGIwAGs4GABlDEz+KABzZBhsGTgPYtI4XHCMks5kFmJCOo8AK1I0ACEM7Ny84nY4ABliuFb2zpBC7DrsZDTiVIBPQmgUjGximAB1JJheZAAOAAYPXYg6g4yMFd24Jbjpj2YARxl+ZkbiZrEGYdaogOrsfjlSqg0jFEqEOAARRkEHgwLmMDWRy0J2Q1g8FWI-EIcIAwhB2C0UFBoO8QDI6tw1jE2iCPGQYOjQTowFyPPoYBgZDBrHyQBlSnA8pjcMgFAB2D7ESUAUSgySQAGZrGo1EA)
```sh
--range-end 32
--range-start 17
```

**Input:**
```jsx
export class F {
  reformatThis() {
    return 1;
  }

  dontTouchThis(){
    return 2    ;
  }
}


```

**Output:**
```jsx
export class F {
         reformatThis() {
           return 1;
         }

         dontTouchThis() {
           return 2;
         }
       }


```

Same result with `--range-start 17 --range-end 18`

Expected behavior would be to reformat only the first method and not produce additional  whitespace that is not present using full-file reformat.

Reproducible with prettier 1.10.2, also with prettier-vscode if only first method is selected in editor.
Not reproducible if 'export' is removed.

JamesHenry:
@josephfrazier @azz We just hit something in this area too via some internal usage of `precise-commits`.

It may not be exactly the same issue, but I will keep things here for now.

In our case the minimal reproduction was this:

**initial.ts**
```ts
export class Rule {
  constructor(
    options: IOptions,
  ) {}
}
```

**staged.ts** (some extra whitespace between options and type annotation)
```ts
export class Rule {
  constructor(
    options:    IOptions,
  ) {}
}
```

`precise-commits` therefore identifies the range data for the affected line as:

[...]

PR #4216, opened 2018-03-27
Playground: Add support for rangeStart and rangeEnd

JamesHenry:
I was delighted to see that this appears to be very easy to add and would allow users to more easily reproduce and report issues such as:

- https://github.com/prettier/prettier/issues/3789
- https://github.com/prettier/prettier/issues/4206
...etc...

screen shot 2018-03-28 at 00 52 17


j-f1:
Looks cool 👍 ([live preview](https://deploy-preview-4216--prettier.netlify.com/playground/))

JamesHenry:
And full for reference, here is me recreating the issue I outline here (https://github.com/prettier/prettier/issues/3789#issuecomment-366273671) in the running local version of this PR branch: 

Reproduction in the updated playground from this PR:
screen shot 2018-03-28 at 01 04 01

JamesHenry:
Ah nice, yeah thanks @j-f1!

Here is the same reproduction on the live preview as an example: https://deploy-preview-4216--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACMAGwEMBnUvAJQFdC49gAdKPA6UmbasGHACmatWEDDACW7JELwBJAPKiJUUgBpBeAJQMAvs20gVIEePbJQxbNggB3AAoWEpZCGIA3CGIAmBlx2QAzYkJSOEMAI2xiMABrOBgAZQwosSgAc2ROalCQTwgwAKCQwxSQ3FtI1IBbYgLg7IArUjQAIUiYuPjiSrgAGRS4WqKjahgMEYAmQeyk7FLnGABPDDhSMGwxUR8MddgAdS8YAAtkAA4ABkNtiBDdyIxnbZW4bFcBw0i0uABRKG8kADYLiAPqk4PEYBYYMgAMwAFnecAAjtQxNg4OViFUakhAnVDCFKmIMlxsqQUqk6ABFagQeBTQwQsL7TxHZDjBmRMSEckAYQglWqzig0DeIGoIQAKsQwk4cYU4NptEA

website/static/playground.js
```diff
@@ -194,6 +196,9 @@ window.onload = function() {
 
 function setOptions(options) {
   OPTIONS.forEach(function(option) {
+    if (option === undefined) {
+      return;
+    }
```
suchipi: What's this check for? We're iterating over an array of known values, so there shouldn't be any undefined values here

website/pages/playground/index.html
```diff
@@ -95,6 +95,8 @@ <h1>Prettier <span id="version"></span></h1>
           <label>--print-width <input type="number" value="80" min="0" id="printWidth"></label>
           <label>--tab-width <input type="number" min="0" value="2" id="tabWidth"></label>
           <label><input type="checkbox" id="useTabs"> --use-tabs</label>
+          <label>--range-start <input type="number" min="0" id="rangeStart"></label>
+          <label>--range-end <input type="number" min="0" id="rangeEnd"></label>
```
ikatyang: I guess these should belong to `Special` category.

lydell:
Just an idea: What if you could set the range by selecting a piece of text somehow? Probably another PR though. Possibly not even useful.

suchipi:
I like that idea. If that proves difficult, we could render the start/end markers as lines behind the text.

JamesHenry:
Sorry for the delay on looping back to this, I finally found a few mins to look into CodeMirror.

I have hacked together a working version of the overlay API that CodeMirror provides to perform highlighting of the range. Pretty darn cool! 😄 

Here is the same example as the one above but with my local updates:
screen shot 2018-04-04 at 10 29 21

Happy for me to push that up as a part of this PR?

josephfrazier:
Oh, if you've already got it working, by all means include it! That would make it much easier to know what the range actually is and to get it right.

FWIW, I'm also curious about https://github.com/prettier/prettier/pull/4216#discussion_r177611888

JamesHenry:
I'm afraid I have no idea about the undefined check 🤷‍♂️ will remove it.

Here is my suggestion for the organisation, now that we have a bit more behaviour (the highlighting), I think it is worth having its own section:

screen shot 2018-04-04 at 11 58 28

JamesHenry:
Same example from above but with all of the new updates:

https://deploy-preview-4216--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACMAGwEMBnUvAJQFdC49gAdKPA6UmbasGHACmatWEDDACW7JELwBJAPKiJUUgBpBeAJQMAvs20gVIEePbJQGbGNgB1MQBMYAC2QAOAAyGYxAEa2HzpAAmQ1IrAHM6AEVqCHhkADNiQlI4T2xiMUJwgGEIAFs84mQQKGg4AxBvdLAAazgYAGUMYjBw5E5qVJAAK1I0ACFqusbiPLgAGStypETkrubsFOximABPDDhSMEtRCpS8sXauLuoUgBUfUgSklMMrJZgABXSwwuu5w2w4AEdqMS-nsRXkUZjd5tgIClrOkMMULJs4NgAG7lQzEbAQgDuj3RCCuKGISIg9gq6SgYTgDS8uGQAGYACyfYjkuAAUSgdmQADYPCA7BAwO9biAyDAhV0INQYBgpYFxdptEA

Ready for final review!

website/pages/playground/index.html
```diff
@@ -20,25 +20,25 @@
...
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/mode/css/css.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/mode/markdown/markdown.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/display/placeholder.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/display/rulers.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/search/searchcursor.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/edit/matchbrackets.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/edit/closebrackets.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/comment/comment.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/addon/wrap/hardwrap.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.26.0/keymap/sublime.js"></script>
```
lydell: You need to update all of these in website/static/service-worker.js as well

JamesHenry: Done

PR #4219, opened 2018-03-28
 Improve formatting of TS casts with generics and unions

### LINKED ISSUE #4168: Very ugly formatting for Typescript union type type assertion
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAHgBQCU+wAOlPvpFAM4z4AOcATgLY4MN6MBhABZwwAaw74AvNVq0APAAUO3XvwbDRE9vICGUAJ4A+fAB98mACZxUOKHEtHiew0bZcefaA1IBtdqJwOABuHAAqBmwAugDclDDUAL4gADQgECy43siguuzsEADuinkIDMggusEQOJaplUzIqLoANgxwaQBG7LricDAAyiy9dgDmyDDs6B0glhBgTa3taXbt7DCKPaOcuottMwBWDIQAQj19g7qccAAydnB7y+noMCwvAEyPM8PsaxUwkTgDDA7BwmXqLFBsAA6rUYEJkAAOAAMaUhEHa0J6LAqkKBHFC9QCAEd0DgAptdNtdkhmvs0u1uBMpjM+FBRi04ABFdAQeBfNIwXSdWGWeHId6Cno4FpjAQQTg7CpQaAPNLodphYXlWlLOCJRJAA)
```sh
--parser typescript
```

**Input:**
```tsx
function x() {
  const permissionsChecker =
    <PermissionsChecker<any> | undefined>(<any>permissions)[receiverType];	
}
```

**Output:**
```tsx
function x() {
  const permissionsChecker = <
    | PermissionsChecker<any>
    | undefined>(<any>permissions)[receiverType];
}

```

**Expected behavior:**
Ideally, I think it would leave the code as input.

### LINKED ISSUE #4171: Ugly formatting for Typescript generic type type assertion
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAngBQCU+wAOlPvpFAM4z6o5QCGANgMLSsDm6AE7tcBALzVatADwBJALYL0MdgCNOcAHQBZdgAcZTIW34AafOyiEAfDYYQFcAKIAPHE1O8oAvfq1OQvxwACJwcPrEqFycauxgANYA8vowDKQA3FLSObTZOKjEMELocORUNNL0TCxsXN4CwqJ4NJKVOfJKKuqaugZGxaYWVrb2ji7unlD8DTj8fgFwQaHhkdGcsfHJqS3pWZUAvtQHIGYgEDvQDMig7EJCEADuAAp3CNco7ABuEDgAJqcQOwmMh1gw4Gc1CJEnAYABlfTxUzIYqlM5-CBgUFccFnNjgoQwZ4ifgKdjYzi4kAAKwYrgAQtCErC4ewnAAZNhwClUiAqfQqABMPIhIERQgJyBAMEI+jgDDAJlSgP0JlgAHV-jAABbIAAcAAYzqqIOD1SJ9FLVfKll9uWchHAAI7oHCO4nsUnkpBg0XghQ4FElP2mTQARXQEHgIrOqjUmr+OuQgtjIhwnC8jjJUqg0HtIHQ4IAKuoPr6DgcgA)
```sh
--parser typescript
```

**Input:**
```tsx
function y() {
  const finalConfiguration =
    <Immutable.Map<string, any>>someExistingConfigMap.mergeDeep(fallbackOpts);
          
  if(true) {
    const finalConfiguration =
      <Immutable.Map<string, any>>someExistingConfigMap.mergeDeep(fallbackOptions);
  }
}
```

**Output:**
```tsx
function y() {
  const finalConfiguration = <Immutable.Map<
    string,
    any
  >>someExistingConfigMap.mergeDeep(fallbackOpts);

  if (true) {
    const finalConfiguration = <Immutable.Map<
      string,
      any
    >>someExistingConfigMap.mergeDeep(fallbackOptions);
  }
}

```

**Expected behavior:**

First assignment would be left as is. The second assignment, which can't fit on one line even with a break after the equals sign, would be rewritten to something like:

```ts
if(true) {
    const finalConfiguration = <Immutable.Map<string, any>>(
      someExistingConfigMap.mergeDeep(fallbackOptions)
   );
}
```

(Adding the parentheses around the expression that's subject to the type assertion is clearer, imo, than:

```ts
if(true) {
    const finalConfiguration = <Immutable.Map<string, any>>
      someExistingConfigMap.mergeDeep(fallbackOptions);
```

where the type assertion might be read as applying to `someExistingConfigMap` and not the result of `mergeDeep`).

kpdonn:
fixes #4171
fixes #4168

Implemented by adding a soft break between the cast and the expression. Also includes wrapping parentheses for clarity as suggested in #4171.

Avoids changing behavior at all though if casting an array or object literal because those already have good behavior where the array or object literal breaks before the cast does so including them would just result in a pointless extra layer of parentheses that would add no clarity because the `{ }` or `[ ]` wrapping the literal serves same purpose as the parentheses would.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -487,13 +487,32 @@ function printPathNoParens(path, options, print, args) {
         " = ",
         path.call(print, "right")
       ]);
-    case "TSTypeAssertionExpression":
+    case "TSTypeAssertionExpression": {
+      const shouldBreakAfterCast = !(
+        n.expression.type === "ArrayExpression" ||
+        n.expression.type === "ObjectExpression"
+      );
+
+      if (shouldBreakAfterCast) {
+        return group(
+          concat([
+            "<",
+            path.call(print, "typeAnnotation"),
+            ">",
+            ifBreak("("),
+            indent(concat([softline, path.call(print, "expression")])),
+            softline,
+            ifBreak(")")
+          ])
+        );
+      }
       return concat([
         "<",
         path.call(print, "typeAnnotation"),
         ">",
         path.call(print, "expression")
       ]);
+    }
     case "MemberExpression": {
       const parent = path.getParentNode();
       let firstNonMemberParent;
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -493,29 +493,30 @@ function printPathNoParens(path, options, print, args) {
         n.expression.type === "ObjectExpression"
       );
 
-      const typeAnnotationGroup = group(
+      const castGroup = group(
         concat([
+          "<",
           indent(concat([softline, path.call(print, "typeAnnotation")])),
-          softline
+          softline,
+          ">"
         ])
       );
 
+      const exprContents = concat([
+        ifBreak("("),
+        indent(concat([softline, path.call(print, "expression")])),
+        softline,
+        ifBreak(")")
+      ]);
+
       if (shouldBreakAfterCast) {
-        return group(
-          concat([
-            "<",
-            typeAnnotationGroup,
-            ">",
-            ifBreak("("),
-            indent(concat([softline, path.call(print, "expression")])),
-            softline,
-            ifBreak(")")
-          ])
-        );
+        return conditionalGroup([
+          concat([castGroup, path.call(print, "expression")]),
+          concat([castGroup, group(exprContents, { shouldBreak: true })]),
+          concat([castGroup, path.call(print, "expression")])
+        ]);
       }
-      return group(
-        concat(["<", typeAnnotationGroup, ">", path.call(print, "expression")])
-      );
+      return group(concat([castGroup, path.call(print, "expression")]));
     }
     case "MemberExpression": {
       const parent = path.getParentNode();
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -493,11 +493,18 @@ function printPathNoParens(path, options, print, args) {
         n.expression.type === "ObjectExpression"
       );
 
+      const typeAnnotationGroup = group(
+        concat([
+          indent(concat([softline, path.call(print, "typeAnnotation")])),
+          softline
+        ])
+      );
+
       if (shouldBreakAfterCast) {
         return group(
           concat([
             "<",
-            path.call(print, "typeAnnotation"),
+            typeAnnotationGroup,
             ">",
             ifBreak("("),
             indent(concat([softline, path.call(print, "expression")])),
@@ -506,12 +513,9 @@ function printPathNoParens(path, options, print, args) {
           ])
         );
       }
-      return concat([
-        "<",
-        path.call(print, "typeAnnotation"),
-        ">",
-        path.call(print, "expression")
-      ]);
+      return group(
+        concat(["<", typeAnnotationGroup, ">", path.call(print, "expression")])
+      );
     }
     case "MemberExpression": {
       const parent = path.getParentNode();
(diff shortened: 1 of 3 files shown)


tests/typescript_cast/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,159 @@
...
+    number,
+    object,
+    null,
+    undefined,
+    any,
+    void,
+    never
+  >>(
+    someExistingConfigMap.mergeDeep(fallbackOptions)
+  );
```
j-f1: What do you think about this formatting instead?

```ts
const stillTooLong = <
  Immutable.Map<
    string,
    boolean,
    number,
    object,
    null,
    undefined,
    any,
    void,
    never
  >
>(
  someExistingConfigMap.mergeDeep(fallbackOptions)
);
```

kpdonn: That looks better to me. It was a little more complicated to get working but I think I managed to do it.

tests/typescript_cast/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,210 @@
...
+    | PermissionsChecker<object>
+    | undefined
+    | number
+    | string
+    | boolean
+    | null
+    | never
+  >(
+    (<any>permissions)[receiverType]
+  );
```
duailibe: This looks like it would be better not to break here after the cast. Have you tried splitting this in 2 groups?

kpdonn: I'm not against the suggestion that it might be better not to break in this case but I'm not sure how to implement it.

The point of the PR is essentially to make the first choice for breaking lines with casts be between the cast and the expression being cast. The hope being better output in cases where both the cast and the expression then both fit on their own lines without further breaking like in:

```ts
  const breakAfterCast = <PermissionsChecker<any> | undefined>(
    (<any>permissions)[receiverType]
  );
```

The problem in the case you're talking about is that even after breaking the cast and the expression onto their own lines, the cast still doesn't fit onto a single line and has to break more.

So what I can't figure out is a way to tell it "try to break on the outer group first, but if that doesn't make the subgroups fit onto their own line then undo the outer break and just break the subgroups." 

Any thoughts?

src/language-js/printer-estree.js
```diff
@@ -490,29 +490,30 @@ function printPathNoParens(path, options, print, args) {
...
+      ]);
+
       if (shouldBreakAfterCast) {
-        return group(
-          concat([
-            "<",
-            typeAnnotationGroup,
-            ">",
-            ifBreak("("),
-            indent(concat([softline, path.call(print, "expression")])),
```
duailibe: I *think* you could just wrap this in a group. The thing here is, when the `typeAnnotationGroup` breaks, it will also break the parent group, which will then make the behavior you are seeing.

But if you wrap this part here in a group, the fact that the parent is "broken" won't affect a child. It will try to print in a single line before breaking.

I'll try my suggestion in a bit and report back :-)

lipis:
This have been recently talked in the Twitterverse :) What's the latest..?

suchipi:
Rebased on master; I'll merge this once CI finishes. Thank you for the fix @kpdonn!

suchipi:
Failure is unrelated

PR #4221, opened 2018-03-28
Added prettier API for ignored files

maoueh:
### Description

The actual logic for creating an ignorer has been moved to `common/ignore.js` so it can be shared easily between CLI and API.

The CLI has been updated to use the new `common/ignore.js` file by creating the ignored via the CLI context.

The API has been added as a factory pattern so that reading the ignored file contents is not repeated for each but only once at the creation of the ignorer.

### Notes

Once agreement has been given about implementation & API scope, I'm going to update the docs and the tests if needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 
 const version = require("./package.json").version;
 
+const createIgnorer = require("./src/common/ignore").createIgnorer;
 const privateUtil = require("./src/common/util");
 const sharedUtil = require("./src/common/util-shared");
 const getSupportInfo = require("./src/common/support").getSupportInfo;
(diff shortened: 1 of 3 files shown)


lipis:
@maoueh Can you please resolve the conflicts and maybe somebody could take a look on this one?

duailibe:
This functionality was implemented in #4341. Thanks for the implementation though!

duailibe:
Actually it's not exactly the same use case, looks like you might want to support multiple ignore paths. Reopening for discussion..

duailibe:
We should decide on one of the two APIs

cc @suchipi @kachkaev

maoueh:
This was implemented because you were not willing to merge my PR adding support for multiple ignore-path entries (#3737) without having an API entry point to determine if the file is ignored or not.

#4341 Implements this through a `getFileInfo`. I'm closing this PR. And I will re-work #3737 when I get time.

suchipi:
Sorry for all the roundabout @maoueh, it can be hard for us maintainers to keep track of so much on our own time when it comes to Prettier, especially in regards to the API, but we appreciate your contributions whether they are merged or not. Thank you for your help!

PR #4222, opened 2018-03-29
Format CSS number literals in the same rule as JS's

gfx:
Trimming trailing zeros as 1.0em ⇢ 1em in CSS removes an intention to express the number precision.

I mean, I write number literals intentinally:

```css
0.8em;
1.0em;
1.2em;
```

but currently prettier (v1.11.1) formats them to:

```css
0.8em;
1em;
1.2em;
```

It makes no sense!

If prettier was a minifier, the behavior might be desiable, though.

This is introduced in https://github.com/prettier/prettier/pull/2627 but I disagree with it.


evilebottnawi:
@gfx `1.0em` === `1em`. Why we don't remove `.0`?

gfx:
@evilebottnawi 

In general, `.0` is a precision of the value. That is, `1` indicates that the `.x` might be truncated or rounded. 

In addition, programmers that use languages that distinguish integer literals and floating point number literals, i.e. C/C++, Ruby, Java, Rust, Go, and so on,  always train themselves to distinguish `1` and `1.0` because  `1` is an integer literal and `1.0` is a floating number literal. Truncating `1.0` to `1` is confusing.

lydell:
> In general, .0 is a precision of the value.

`1.50` would still get lowered in precision to `1.5` so this PR doesn't solve your use case?

> In addition, programmers that use languages that distinguish integer literals and floating point number literals, i.e. C/C++, Ruby, Java, Rust, Go, and so on, always train themselves to distinguish 1 and 1.0 because 1 is an integer literal and 1.0 is a floating number literal. Truncating 1.0 to 1 seems to change the type of numbers.

I originally trimmed all unnecessary trailing zeroes even in JS, but other maintainers didn't dare do it because of this reason. But we figured CSS people hadn't hardwired their brains to this convention. IMO it's better to embrace the fact that JS only has floating point numbers and not pretend anything else, but people disagree...

gfx:
> 1.50 would still get lowered in precision to 1.5 so this PR doesn't solve your use case?

In fact, yes. However,  it causes no issues on my codebase, so I don't change the feature for now.

For CSS, I don't agree with this truncation so I am hesitating to apply prettier to my codebase.

Because this is an arguable issue, it might be an option?

evilebottnawi:
@gfx no option, all popular style guides recommended `1.0em` => `1em`. It is more readable and honestly for me `1.0em` in code base is very misleading. Why `1.0em`? Maybe some workaround for css bug or maybe somebody forget something (`1.05em`). What is magic number? Why not just `1em`? 

IMHO, :-1: on this. If you can use this style formatting numbers you can use prettier + stylelint (with fix option) as doing for js `prettier` + `eslint` for custom use case.

gfx:
@evilebottnawi Your opinion is okay but I see more 👍than 👎  of this issue:

screen shot 2018-04-03 at 19 46 14 (as of 2018/04/03)

meaning it is worth creating an option, isn't it?

evilebottnawi:
@gfx Let's give more time on voting. And feedback welcome for another people.

gfx:
@evilebottnawi Of course!

lydell:
> > 1.50 would still get lowered in precision to 1.5 so this PR doesn't solve your use case?
>
>In fact, yes. However, it causes no issues on my codebase, so I don't change the feature for now.

Then why does `1.0`→`1` cause issues? What issues does it cause?

gfx:
@lydell You mean, JavaScript formatter also truncates `.0`?

suchipi:
On the fence with this one 🤷‍♀️

duailibe:
I'm 👎 with this

alfaproject:
I tend to disagree, only because this is more than a mere formatting change. Shouldn't prettier only touch whitespace (by default at least)?

evilebottnawi:
Close in favor https://prettier.io/docs/en/option-philosophy.html

Issue #3296, opened 2017-11-20
Can't use Prettier within a browser


bvaughn:
Following up from [this Twitter comment](https://twitter.com/suchipi/status/932517379367608320); relates to babel/website/issues/1388 and babel/website/pull/1466.

The Babel REPL has been using Prettier via [bundle.run](https://bundle.run/) for some time. Version 1.7.0 broke our ability to do this with the error: "_cosmiconfig requires at least version 4 of Node_". Version 1.8.0 introduced a different error: "_os$2.homedir is not a function_".

It would be ideal if Prettier output a browser-friendly UMD bundle for use-cases like this, but in lieu of that maybe it could at least work via a solution like rundle.run? 😄 

For the time being, we've pinned Babel to use Prettier version 1.6.1. You can try it out [here](https://deploy-preview-1466--babel.netlify.com/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYewdgzgLgBAhjAvDA3gKBjARgLhgcmHwBoMYATPAIgFMrSBfIA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=react%2Cstage-2&prettier=true&targets=&version=6.26.0) if you'd like.

mitermayer:
This sounds reasonable to me, I can try looking into this during the weekend, unless @azz , @suchipi @ikatyang what’s to look into this first

bvaughn:
A UMD bundle would be so-so-so great ❤️  If there's support for that idea, I'd also be happy to help (although I am not familiar with Prettier's codebase)

suchipi:
I think we can fix "os$2.homedir is not a function" with the same approach as in this PR (adding to the rollup-plugin-replace config): https://github.com/prettier/prettier/pull/2930

"cosmiconfig requires at least version 4 of Node" might require a change to cosmiconfig.

suchipi:
Also, since we are already using rollup, we are probably not too far from generating a working UMD bundle.

ikatyang:
How about extracting `prettier.format()` to another file? The browser version seems no need to read config file (`fs`-related), it only cares about `prettier.format()`.

bvaughn:
A trimmed down version for the browser would be slick.

Would also be nice if there were multiple UMD targets - so we could avoid loading parsers for targets/languages we don't care about (eg CSS).

But I don't want to be greedy. 😁

duailibe:
[...]

PR #4230, opened 2018-03-30
(steps toward) A prettier that works cleanly on the web

tmcw:
Afaict, #3296 has a lot of hopes but no active PRs trying to do it. This, then, is trying to be that, and it's just a first gesture in that direction - though it does accomplish the initial goal of building a rollup-bundled JavaScript file that doesn't include prebundled parsers and doesn't throw on import or require a stub of globals to run. It does so by replacing references to process.

I think the eventual goal, as mentioned in that ticket, should still be to cleanly separate the parts of prettier that rely on a node environment and filesystem from those that don't. The main thing harshing that vibe is the `support` module: it's included even by the Markdown parser, which uses it to ask whether it can embed a parsing mode for another language. I suspect that to create a clean boundary there, we'd need to just pass the output of `getSupportInfo` into all embed calls, and that'd mean a change to the method signature, something like

```
//                                              new ↓
function embed(path, print, textToDoc, options, supportInfo) {
```


j-f1:
Looks cool! Would you be able to make the playground use the web version of Prettier so we can test it more easily?

tmcw:
Yep, kind of want to get tests passing first, though. Only three tests to go! If anyone wants to take a look, well - this is a few revisions in and will probably need more rethinking: prettier relies pretty heavily on the node platform, and there are a few things to think about, like for instance:

- whether it's best to have to entry points and use argument passing ("injection") for the difference between web and node targets, or if we should have some sort of conditional require that's remapped depending

aboyton:
I haven’t looked a lot at the code but I’m curious, what’s node specific about Prettier? I would have hoped that it was little more than a pure function `format` wrapped in some glue code to read and write files and take in a config? (Not to discount all the work in Prettier, it’s a very nice pure function and glue code.)

tmcw:
There isn't a ton of node-specific stuff, but there's enough to cause a lot of havoc with bundlers :)

- Relies on process.env to change behavior based on environment variables
- Relies on os.EOL to match system-specific line endings
- Code for requiring custom parsers and custom plugins that relies on require() (which in turn relies on filesystem access and the ability to synchronously, dynamically load code)
- Also relies on require() to lazy-load parsers, to keep the prettier startup time fast without losing the ability to support things like prettifying javascript snippets within Markdown

So, in the 'idempotent' sense of 'pure', sure - prettier's essentially that. But in the 'behaves the same way on all platforms' sense, prettier isn't (and pretty much no software fits that definition). And in terms of dynamically including code to fit inputs and intelligently cut down on load time, prettier has a pretty sophisticated, subtle implementation that you'd only find in other large, plugin-enabling architectures like eslint.

tmcw:
@j-f1 Okay, this now passes tests locally, and I'm trying to integrate it with the playground, but am baffled by the setup: I'm running (as documented in website/readme):

```sh
yarn start # in ./website

# and also
./node_modules/.bin/webpack --watch # in ./website
```

But the site depends on a file `/lib/index.js`, which is required by [service-worker.js](https://github.com/prettier/prettier/blob/74ad0014aa4fbec2f4c5a681f65c3484d9b57e51/website/static/service-worker.js#L10) and [worker.js](https://github.com/prettier/prettier/blob/74ad0014aa4fbec2f4c5a681f65c3484d9b57e51/website/static/worker.js#L64), but I can't find any place in prettier's codebase in which that file is generated.

src/language-css/index.js
```diff
@@ -54,7 +54,7 @@ const languages = [
 
 const postcss = {
   get parse() {
-    return eval("require")("./parser-postcss");
+    return require("./parser-postcss");
```
tmcw: My assumption in removing these evals is that they were used to stop browserify or webpack from bundling them as dependencies. Since the approach of this PR is to load 0 plugins from the web route, it should avoid that requirement entirely.

j-f1:
`lib/index.js` comes from [here](https://github.com/prettier/prettier/blob/master/scripts/build/build-docs.js).

src/language-css/index.js
lydell: I think the evals are there mainly to stop ... rollup? ... from bundling those files in the npm package. Remember, Prettier bundles all its dependencies into the npm package and has zero dependencies listed in package.json. I might remember all of this wrong though.

tmcw:
Okay, that makes sense, though it seems like we should be able to set them as `externals` in the rollup configuration, right?

suchipi:
Yes- assuming that works, I would prefer that over `eval("require")`.

duailibe:
@tmcw Last time we tried, the `externals` only stop including them in the bundle, but the `require` calls were hoisted. That's why we went forward with `eval`

scripts/build/rollup.web.config.js
```diff
@@ -0,0 +1,27 @@
...
+import globals from "rollup-plugin-node-globals";
+import json from "rollup-plugin-json";
+import alias from "rollup-plugin-alias";
+
+export default Object.assign(baseConfig, {
+  entry: "index.js",
+  dest: "prettier.js",
+  format: "umd",
+  plugins: [
+    alias({
```
tmcw: Stalled for the moment on this: the `alias` plugin has the following observed behavior:

- The code from `load-plugins-browser.js` is included
- The dependencies from `load-plugins.js` are also included

🤯 I have no idea. Will return to it with fresh eyes in the near future, or appreciate any help cracking the code in the meantime.

tmcw:
Stalled for the moment on getting rollup to require a different file depending on environment. Right now it's this puzzling, incorrect behavior when I try to redirect a require from load-plugins to load-plugins-browser:

- The code from `load-plugins-browser.js` is included
- The dependencies from `load-plugins.js` are also included

🤯 I have no idea. Will return to it with fresh eyes in the near future, or appreciate any help cracking the code in the meantime.

package.json
```diff
@@ -13,8 +13,7 @@
   "engines": {
     "node": ">=4"
   },
-  "dependencies": {
-    "@babel/code-frame": "7.0.0-beta.40",
+  "dependencies": { "@babel/code-frame": "7.0.0-beta.40",
```
lydell: Huh?

suchipi:
@tmcw I think this is superceded by #4449; does that cover all the use-cases you were looking to address here?

Also sorry this stayed open for so long, but I felt like your opening it reminded us of the need for it, so it wasn't for naught

tmcw:
👍 yep! As long as the thing is solved in the end :)

Issue #4238, opened 2018-04-01
Ugly TS output for type alias of a function type with generics


kpdonn:
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAYgVyjABVM4AeYgRmzgA94oATAZ2xZgCcBLKAcwA02YgCZaDBK2xR8AWwBGcTkOIBmcYykyFSgHzYAvNgAUAHSjZslACwbJbDj367jAQ058kw6wEpD+gHl8GAxgyioVERVVFWtdcx8AbhABEAgMGG5oFmRQd04IAHcABXcEHJRXADcIbiYUkFcOZAAzVwAbFjhU+U5XMABrOBgAZQx+3j5kLnxukCYIMFaOrtTeLs4YYr6+WVdlzrmAKxY6ACE+weGR11k4ABleOAPVtODQmBEXufHODeRUGQWGAeBkGhgnDAAOp1GAAC2QAA4AAypCEQLpQvoYAEQuAbKrPVKcOAAR3w3BJ21cu32SDah1SXVk3GmnFmTMm7TgAEV8BB4N9UjBXPIYUx4cgoqg+tx2pMAMIQWR7AFQaBEkD4LrEUUVBldAC+hqAA)
```sh
--parser typescript
```

**Input:**
```ts
type FuncType<T1 extends string, T2 extends number, T3 extends number> = (
  <T4 extends string>(arg: T4) => Output<T1, T2, T3, T4>
);
```

**Output:**
```ts
type FuncType<T1 extends string, T2 extends number, T3 extends number> = <
  T4 extends string
>(
  arg: T4
) => Output<T1, T2, T3, T4>;
```

**Expected behavior:**
Similar to input format where there is a line break added after the `=`.

PR #4239, opened 2018-04-01
[RFC] New 'moreCompact' feature that would pick option that fits more content onto following lines

kpdonn:
Edit: Updated to be more generalizable.

Not intended or ready to be merged as is. Just want to get feedback on if there is any interest in something like this.

What this does is add an `moreCompact` concept that takes two options for output and determines which fits more content into the next N lines. Motivated primarily by cases like #4238 where 
```ts
type FuncType<T1 extends string, T2 extends number, T3 extends number> = (
  <T4 extends string>(arg: T4) => Output<T1, T2, T3, T4>
);
```
currently becomes
```ts
type FuncType<T1 extends string, T2 extends number, T3 extends number> = <
  T4 extends string
>(
  arg: T4
) => Output<T1, T2, T3, T4>;
```

It chooses the more compact version by checking which ends up containing more non-whitespace characters before a configurable number of line breaks. The only case I used it for right now is adding an optional line break before the right hand side of assignments. For that case I have it check the current line and the following line. 

The assumption is that if adding an immediate line break after the `=` sign lets it fit more content onto the next line than it would fit on the current line and next line combined without the line break, that it is better to add the break.

Check the updated snapshots for changed output but it generally seems better to me, especially in several typescript situations.


suchipi:
I haven't looked at this yet but the idea sounds great.

j-f1:
What’s the status of this? (No pressure, just asking)

kpdonn:
@j-f1 Status is that no one expressed much interest so it's just been sitting. The current implementation was just a quick and dirty version to show the concept so it'd need some cleanup work before it'd be ready to merge.

I still think it'd be a good addition though so if there does end up being interest in adding it I'd be willing to do that work to clean it up.

j-f1:
I *think* https://github.com/prettier/prettier/pull/4239/commits/5b0046018365e75095b6ee0b3fb3ad1b8dcea310 fixes an error, but feel free to revert if it was intended to work the other way.

j-f1:
Seems like something is badly broken here.

lipis:
A bit outdated.. please ping back if this needs to remain open.

PR #4243, opened 2018-04-02
fix: don't strip empty jekyll front-matter blocks

### LINKED ISSUE #4162: Strips empty Jekyll front-matter blocks
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBadAdK7VawIwgBMBPYPGQgJyLiqQAIBGABwA8GBnCAGwEsiDKnCIBuLAF8QAGhAQWMPtE7JQAQypUIAdwAKGhCpRqAbhAEyQazjGQAzNT05xZ+KmrABrODADKLDz4oAHNkGCoAVxcQIggwe0dnWSDnKhhdd2CAWzUEp2iAK042ACF3Lx9fNSy4ABkguDykuQiYFlaAJibogKpU5BBOME4VWRYqIJgAdQEYAAtkAA4ABjGtZyn3FgHxuFSTRtlhAEcIvmEMtWzcpAd82WcsvjDI6M4g4J44AEUIiHhurIYGp8DMiPNkB0ge4+PwQgBhCBZHIDKDQQ4gCLOAAqIKMd2cEgkQA)
```sh
--parser scss
```

**Input:**
```scss
---
---

body{
	border: 1px solid red;
}
```

**Output:**
```scss
body {
  border: 1px solid red;
}

```

I noticed prettier will retain Jekyll-style front matter blocks if they have content but remove them if they are empty. I believe this to be incorrect behavior as Jekyll requires the front-matter block to be present to process the file, even if that block has no content. Specifically this is happening with a .scss file in Jekyll with empty front-matter. 

See #3321 & #3802

evilebottnawi:
fixes #4162

Also refactor logic to keep parse logic in one function

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "get-stream": "3.0.0",
     "globby": "6.1.0",
     "graphql": "0.13.2",
-    "gray-matter": "3.1.1",
+    "gray-matter": "4.0.1",
     "html-tag-names": "1.1.2",
     "ignore": "3.3.7",
     "jest-docblock": "22.2.2",
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -1,7 +1,7 @@
 "use strict";
 
-const createError = require("../common/parser-create-error");
 const grayMatter = require("gray-matter");
+const createError = require("../common/parser-create-error");
 
 function parseSelector(selector) {
   // If there's a comment inside of a selector, the parser tries to parse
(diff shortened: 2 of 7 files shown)


evilebottnawi:
Oh, minimum require version `node@6`. We have plans to drop `node@4`?

ikatyang:
Re: node@4, I mentioned it in [2.0 item](https://github.com/prettier/prettier/issues/3503#issuecomment-352154804) but it seems [rejected](https://github.com/prettier/prettier/issues/3503#issuecomment-359937346).

evilebottnawi:
@ikatyang any ideas how we can solve this on our board. In near future many projects update your code bases to `node@6` only support

ikatyang:
We might have to fork it, downgrade it, and use that fork here, though I don't think it's a good idea but it's better than using `YARN_IGNORE_ENGINES=true`.

`node@4` must be dropped in the future but I don't know when it will be.

vjeux:
Can you just match `/^---\r?\n[\s\r\n]*\r?\n---\r?\n/` and if there's a match then print `---\n---\n` at the beginning? We don't really need to upgrade to fix this.

evilebottnawi:
Already fixed in other PR

PR #4244, opened 2018-04-02
Upgrade Flow to 0.69

### LINKED ISSUE #4240: Flow – Unable to declare a class that implements an interface
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEATOYA2BDATnAAgFsJUBXTQgHRDgA9siAHSmg4KqAg9LPQgJaw4uAGbYwhAKIA3BDClEBMeLnadu3aAAoYATyZwkBAM4xcQgOYAaApgFmEI4wDEyUMDAHQAlMZkQAqgA3BqaEKKiugZGpuZWtvaOUM4Ebh5evv6BIWEAvpxhvDj4BPRMELgwBHwmJgQAKiJKUNiYBALMlETy9bLyisqq6lzhUNGGxmYWUDZ2DvApuK7unt5QfgQBQaGjWpETsdMJ88mp6WtZWzm73AVQ9yDWIBBMmVAmyKB4uBAA7gAFfgfZAgbDbVBPMFmZDiTAmODPABGuAkAGs4DAAMpMCRWZDmMiItAQMCwtoI55CBFVAGoyxEbDk+HEgBWJjoACFUWAMdjGHAADJCODMykvMgwJiSgBMYuJuNwNNBokw-yhTBmMAA6kEYAALZAADgADM9NRAEdrUUxQZq4DS5FD8ABHMgCfB07AMplIOHihFKAm4InPExWSgARTIEHg8ueMGwSN1qANyBlCdRAnsswAwhAiIzQVBoKLnmQEQ0k58-RS4Hk8kA)
```sh
--parser flow
```

**Flow v0.69.0**
[Try link](https://flow.org/try/#0CYUwxgNghgTiAEBbA9sArhBByEAPKiADplvAN4BQ88oksCAlgHYAuIMAZlGAgKIBuIVr0QMWbGOSrwAkMiYAKFgE9CIAFzwAzixjMA5gBp4EBjqHtNAMTRMwLBvICUm-sgbAA3NOrIOHJVUNbV0DY1NzJkt4GzsHZ1d3L2kAX2lpWmg4eDxCZBgWeDotLXgAFXZRJigIeAYiTEQhFlKBZpExCSlqOUUVNU0dPSYjEzM2KJhrW3tHJhd4Nw9val9-QIGQ4dGIiejY2YTFpJX4NLSgA)

**Input:**
```jsx
declare module "example" {
  declare interface EventEmitter {
    on(type: string, listener: Function): void;
    off(type: string, listener: Function): void;
  }

  declare export class Terminal implements EventEmitter {
    on(type: string, listener: Function): void;
    off(type: string, listener: Function): void;
  }
}

```

**Output:**
```jsx
SyntaxError: Unexpected identifier (7:33)
   5 |   }
   6 | 
>  7 |   declare export class Terminal implements EventEmitter {
     |                                 ^^^^^^^^^^
   8 |     on(type: string, listener: Function): void;
   9 |     off(type: string, listener: Function): void;
  10 |   }
```

**Expected behavior:**
This should be supported. It works in the `flow.org/try` but not in the `prettier.io/playground`.

duailibe:
Closes #4240

### DIFF
--- a/tests/flow/declare_implements/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow/declare_implements/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,28 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`test.js 1`] = `
+declare module "example" {
+  declare interface EventEmitter {
+    on(type: string, listener: Function): void;
+    off(type: string, listener: Function): void;
+  }
+
+  declare export class Terminal implements EventEmitter {
+    on(type: string, listener: Function): void;
+    off(type: string, listener: Function): void;
+  }
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+declare module "example" {
+  declare interface EventEmitter {
+    on(type: string, listener: Function): void;
+    off(type: string, listener: Function): void;
+  }
+
+  declare export class Terminal {
+    on(type: string, listener: Function): void;
+    off(type: string, listener: Function): void;
+  }
+}
+
+`;
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "escape-string-regexp": "1.0.5",
     "esutils": "2.0.2",
     "find-project-root": "1.1.1",
-    "flow-parser": "0.64.0",
+    "flow-parser": "0.69.0",
     "get-stream": "3.0.0",
     "globby": "6.1.0",
     "graphql": "0.13.2",
--- a/scripts/build/rollup.parser.config.js
+++ b/scripts/build/rollup.parser.config.js
@@ -19,15 +19,6 @@ export default Object.assign(baseConfig, {
           include: "node_modules/typescript-eslint-parser/parser.js"
         })
       : {},
-    // In flow-parser 0.59.0 there's a dynamic require: `require(s8)` which not
-    // supported by rollup-plugin-commonjs, so we have to replace the variable
-    // by its value before bundling.
-    parser.endsWith("flow")
-      ? replace({
-          "require(tf)": 'require("fs")',
-          include: "node_modules/flow-parser/flow_parser.js"
-        })
-      : {},
     json(),
     resolve({ preferBuiltins: true }),
     commonjs(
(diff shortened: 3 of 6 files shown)


tests/flow/declare_implements/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+declare module "example" {
+  declare interface EventEmitter {
+    on(type: string, listener: Function): void;
+    off(type: string, listener: Function): void;
+  }
+
+  declare export class Terminal {
```
ikatyang: The `implements EventEmitter` is missing.

duailibe: Looks like it's not exposed in the AST https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwCohgWwJYDsoJ5dsAHCEbEfAFzQFEA3G2hvW2reAbwCgAvnyA

bakkot: @duailibe Should open an issue on flow about this if you haven't. (Or I can if you want.)

At a glance it looks like it's exactly the same problem as with [mixins](https://github.com/facebook/flow/issues/5463), so if anyone is up to making a PR to flow you could probably solve both without much work!

duailibe: Opened #5463, thanks!

vjeux: If the flow parser doesn’t parse it, does it mean that the type checker doesn’t use it? So it’s not a blocker

bakkot: Unfortunately not: the AST exposed to JS like prettier by `flow-parser` is translated from the AST used internally, lossily. The type checker has access to an AST which contains those nodes, I'm almost certain.

That said I don't think it should necessarily be a blocker anyway, since prettier has been doing the same with [`mixins`](https://github.com/prettier/prettier/pull/3391) for months.

duailibe: Yep, @bakkot is right. I don't think there's much we can do...

I opened https://github.com/facebook/flow/pull/6118 so we can hopefully land those soon

Issue #4254, opened 2018-04-04
custom selectors case is changed to lowercase


Coobaha:
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABMBXAzjCBbAWizgBs4xcAnAAiQILAEM9SBhR46gOjwE8GsWIqXJUA3AB0oUug2ZsOcasCnVIJCJSTVKcACaSoAXykgANCAgAHGAEtoWZKCuVbsAOq29MABbIAHAAMFjCMAEae3n5IAEwWWG4A5mQAihgQ8MgAZowkxCGUjLYkSaz4eIzIIFDQcOYgYYVgANZwMADKVoxgScgwlBhwFgBWWAAeAEJNrR3yADJudUg5eUMgXZTElFVggvXEeLZ9A2vYcAAq4Y7LufkgblswAAqFiRXZt2u6AI4YtrovRhvSo3VYWFwQYjuQpWKouOBbABudQsjEolAgAHcnmiENcQIxERAvPU9BAwB8wQScJS7hAMDArAyYrS4EYjEA)
```sh
--parser css
```

**Input:**
```css
@custom-selector :--camelCase .my-css-selector;

:--camelCase {
 color: red;
}

```

**Output:**
```css
@custom-selector :--camelCase .my-css-selector;

:--camelcase {
  color: red;
}

```

**Expected behavior:**
```css
@custom-selector :--camelCase .my-css-selector;

:--camelCase {
  color: red;
}

```

evilebottnawi:
@Coobaha Thanks for issue. In near future i will fix it. BTW PR welcome it is easy fix

PR #4255, opened 2018-04-04
fix: don't lowercase custom selectors variable name #4254

Coobaha:


tests/stylefmt/custom-selectors/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3,20 +3,28 @@
...
 @custom-selector :--enter :hover, :focus;
+@custom-selector :--camelCase .my-css-selector;
 
 :--button {
   display: inline-block;
 }
 :--button:--enter {
   text-decoration: underline;
 }
+:--camelcase {
```
j-f1: Looks like this one is still camel-cased.

PR #4264, opened 2018-04-05
Handle comments before `else`

### LINKED ISSUE #2573: Prettier misplaces commented out code between if else
```js
let a = 2;

if (a==2) {
  console.log('2');
}
// else if (a==3) {
//   console.log('3');
// }
else if (a==4) {
  console.log('4');
}
```

becomes

```js
let a = 2;

if (a == 2) {
  console.log('2');
} else if (a == 4) {
  // else if (a==3) {
  // console.log('3');
  // }
  console.log('4');
}
```

[Live example](https://prettier.io/playground/#%7B%22content%22%3A%22let%20a%20%3D%202%3B%5Cn%5Cnif%20(a%3D%3D2)%20%7B%5Cn%20console.log(%272%27)%3B%5Cn%7D%5Cn%2F%2F%20else%20if%20(a%3D%3D3)%20%7B%5Cn%2F%2F%20console.log(%273%27)%3B%5Cn%2F%2F%20%7D%5Cnelse%20if%20(a%3D%3D4)%20%7B%5Cn%20console.log(%274%27)%3B%5Cn%7D%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Atrue%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D)

duailibe:
I chose to treat comments as:
- trailing comments of the consequent, if it's a `BlockStatement`
- dangling comments of the `IfStatement` otherwise. (I'm not quite sure this is conceptually correct. Thoughts?)

In this PR I'm always printing as:

```js
if (cond) {
}
// comment
else {
}

if (cond) {
} /* comment */ else {
}
```

Should `// comment`s always be printed in their own line or should we detect if they're on the same line? Same for `/* comment */`s should always be printed in the same line?

Closes #2573

### DIFF
--- a/tests/flow/refinements/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow/refinements/__snapshots__/jsfmt.spec.js.snap
@@ -2571,8 +2571,7 @@ type Breakfast = Apple | Orange | Broccoli | Carrot;
 function bar(x: Breakfast) {
   if (x.kind === "Fruit") {
     (x.taste: "Good");
-  }
-  // error, Apple.taste = Bad
+  } // error, Apple.taste = Bad
   else (x.raw: "No"); // error, Carrot.raw = Maybe
 }
 
--- a/tests/flow/refinements/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow/refinements/__snapshots__/jsfmt.spec.js.snap
@@ -2571,7 +2571,9 @@ type Breakfast = Apple | Orange | Broccoli | Carrot;
 function bar(x: Breakfast) {
   if (x.kind === "Fruit") {
     (x.taste: "Good");
-  } else (x.raw: "No"); // error, Apple.taste = Bad // error, Carrot.raw = Maybe
+  }
+  // error, Apple.taste = Bad
+  else (x.raw: "No"); // error, Carrot.raw = Maybe
 }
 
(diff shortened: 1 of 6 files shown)


tests/flow/refinements/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2571,7 +2571,9 @@ type Breakfast = Apple | Orange | Broccoli | Carrot;
 function bar(x: Breakfast) {
   if (x.kind === "Fruit") {
     (x.taste: "Good");
-  } else (x.raw: "No"); // error, Apple.taste = Bad // error, Carrot.raw = Maybe
+  }
+  // error, Apple.taste = Bad
```
suchipi: This one makes me think we should detect same-line. Same for `/* bla */`

duailibe:
@suchipi Thank you for the suggestion 🙃

PR #4265, opened 2018-04-05
Fix embedded GraphQL in JS with backticks

### LINKED ISSUE #4258: Escaped backticks in GraphQL descriptions prevents formatting
GraphQL description strings may contain markdown. Backticks for markdown inline code are escaped in the tagged template literal string using backslashes. Escaped backticks causes Prettier to not format the entire GraphQL string.

**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBzAjgGwAYB0oAEBuIuu2AZhBGdgQLYCuMAhjAJbQEAOLAnpggsAJgDoS+IjD7c4BAGLUACv0EiCwSUWIwARiwBOSAgGUYB9lFRaAvvmwgANCAjcO0AM7JQhgxADuKgYIXigsAG4Q7MJOICweMMgULJgecM66BixgANZwMCa8YJaoyOaM6SDCEGBJKWnOlmkGMEpZqPQsdamVAFYeAB4AQlm5+SYs9HAAMpZw3Q0uzNzMAEwLlbwGzcgg+roC0LHcFrAA6tEwABbIABwADM4nEGlnWdy7J3DN4fPOWVY4ABRKAxJCPEAA1BwMyGRLg-5wdCMdjBNosDpdJDJHrONL0dhlAwVPElTBwACKjAg8A2zlYuguwmuyFW9Ky7EwJQAwhB6J1dlBoH8QIw0gAVFi6UI4tI2GxAA)

**Input:**
```jsx
gql`
  "\`foo\` mutation payload."
  type FooPayload {
    	bar: String
  }
`
```

**Output:**
```jsx
gql`
  "\`foo\` mutation payload."
  type FooPayload {
    	bar: String
  }
`;
```

**Expected behavior:**
```jsx
gql`
  "\`foo\` mutation payload."
  type FooPayload {
    bar: String
  }
`;
```

duailibe:
This one was hard to write the title :sweat_smile:

When we parse embedded GraphQL, we use the `raw` value from the template literal, which includes escape characters, e.g.:

```gql
"\`foo\` mutation payload."
type FooPayload {
  bar: String
}
```

We should probably pass the `cooked` value instead, because they're only meaningful in the JS context, and escape the result of the "foreign" language printer (which is what the markdown embed does).

Closes #4258

### DIFF
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -77,7 +77,7 @@ function embed(path, print, textToDoc /*, options */) {
           const templateElement = node.quasis[i];
           const isFirst = i === 0;
           const isLast = i === numQuasis - 1;
-          const text = templateElement.value.raw;
+          const text = templateElement.value.cooked;
           const lines = text.split("\n");
           const numLines = lines.length;
           const expressionDoc = expressionDocs[i];
@@ -107,13 +107,17 @@ function embed(path, print, textToDoc /*, options */) {
               doc = docUtils.stripTrailingHardline(
                 textToDoc(text, { parser: "graphql" })
               );
-            } catch (_error) {
+            } catch (error) {
+              if (process.env.PRETTIER_DEBUG) {
+                throw error;
+              }
               // Bail if any part fails to parse.
               return null;
             }
           }
 
           if (doc) {
+            doc = escapeBackticks(doc);
             if (!isFirst && startsWithBlankLine) {
               parts.push("");
             }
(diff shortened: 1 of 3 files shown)


duailibe:
@lydell that's unfortunate :(

What do you suggest?

lydell:
As far as I understand, only the cooked value can be null.

I think that we need to choose between cooked and raw on a case-by-case basis for every template tag we support, and use the same as the template tag uses. If we use cooked and it is null, we should print the whole template as-is.

Would you mind making a follow-up PR handling cooked being null and adding a test?

duailibe:
> As far as I understand, only the cooked value can be null.

I mean any of the quasis

Will do a follow-up PR 👍

PR #4266, opened 2018-04-05
 Fix styled-components template strings formatting

### LINKED ISSUE #1872: Styled-components template strings formatting
I'm not sure if it's can be fixable and how. But it's very common when using `styled-components` to write constructions that look like that.
```javascript
const Something = styled.span`
 ${props => props.right && css`
    padding: 10px 134px 10px 20px;
    &:after {
      left: initial;
      right: 30px;
    }
  `
`
```
So it looks a bit cleaner, especially for multiple rules. After the latest update, it's broken and looks ugly.
```javascript
const Something = styled.span`
  ${props =>
  props.right &&
  css`
    padding: 10px 134px 10px 20px;
    &:after {
      left: initial;
      right: 30px;
    }
  `
`
```
To prove my words, this construction is on the main page first example of styled-components website: https://www.styled-components.com/

kinday:
This is an attempt to fix #1872. I added test case and it’s covered with code changes, yet I don’t feel this is a complete solution as there might be edge cases.

I’ll play with it a bit more, but initial feedback is welcome.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -659,6 +659,11 @@ function printPathNoParens(path, options, print, args) {
         (n.body.type === "ArrayExpression" ||
           n.body.type === "ObjectExpression" ||
           n.body.type === "BlockStatement" ||
+          (n.body.type === "TemplateLiteral" && isInsideTemplate(path)) ||
+          n.body.type === "TaggedTemplateExpression" ||
+          (n.body.type === "LogicalExpression" &&
+            isInsideTemplate(path) &&
+            !isLogicalChain(n.body)) ||
           isJSXNode(n.body) ||
           isTemplateOnItsOwnLine(n.body, options.originalText, options) ||
           n.body.type === "ArrowFunctionExpression")
@@ -4972,6 +4977,10 @@ function shouldInlineLogicalExpression(node) {
     return true;
   }
 
+  if (node.right.type === "TaggedTemplateExpression") {
+    return true;
+  }
+
   if (isJSXNode(node.right)) {
     return true;
   }
@@ -5070,7 +5079,8 @@ function printAssignmentRight(leftNode, rightNode, printedRight, options) {
   }
 
   const canBreak =
-    (isBinaryish(rightNode) && !shouldInlineLogicalExpression(rightNode)) ||
+    isBinaryish(rightNode) ||
+    isLogicalChain(rightNode) ||
     (rightNode.type === "ConditionalExpression" &&
       isBinaryish(rightNode.test) &&
       !shouldInlineLogicalExpression(rightNode.test)) ||
@@ -5369,6 +5379,19 @@ function returnArgumentHasLeadingComment(options, argument) {
   return false;
 }
 
+function isLogicalChain(node) {
+  if (node.type !== "LogicalExpression") {
+    return false;
+  }
+  if (
+    node.left.type === "LogicalExpression" ||
+    node.right.type === "LogicalExpression"
+  ) {
+    return true;
+  }
+  return isLogicalChain(node.left) || isLogicalChain(node.right);
+}
+
 function isMemberExpressionChain(node) {
   if (node.type !== "MemberExpression") {
     return false;
@@ -5629,6 +5652,23 @@ function isSkipOrOnlyBlock(node) {
 
 function isTemplateLiteral(node) {
   return node.type === "TemplateLiteral";
+}
+
+function isInsideTemplate(path, depth) {
+  const parent = path.getParentNode(depth);
+
+  if (!parent) {
+    return false;
+  }
+
+  if (
+    parent.type === "TemplateLiteral" ||
+    parent.type === "TaggetTemplateExpression"
+  ) {
+    return true;
+  }
+
+  return isInsideTemplate(path, ~~depth + 1);
 }
 
 function isIdentiferAsync(node) {
--- a/tests/binary-expressions/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/binary-expressions/__snapshots__/jsfmt.spec.js.snap
@@ -154,11 +154,12 @@ const avatar = hasAvatar && <Gravatar user={author} size={size} />;
 
 const avatar = (hasAvatar || showPlaceholder) && <Gravatar user={author} size={size} />;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-const user = renderedUser || (
-  <div>
-    <User name={this.state.user.name} age={this.state.user.age} />
-  </div>
-);
+const user =
+  renderedUser || (
+    <div>
+      <User name={this.state.user.name} age={this.state.user.age} />
+    </div>
+  );
 
 const user =
   renderedUser ||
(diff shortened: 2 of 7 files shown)


src/language-js/printer-estree.js
```diff
@@ -5233,6 +5245,15 @@ function templateLiteralHasNewLines(template) {
   return template.quasis.some(quasi => quasi.value.raw.includes("\n"));
 }
 
+function isStyledComponentsCssTemplate(n, text, options) {
+  return (
+    n.type === "LogicalExpression" &&
+    n.right.type === "TaggedTemplateExpression" &&
+    n.right.tag.name === "css" &&
+    !privateUtil.hasNewline(text, options.locStart(n), { backwards: true })
```
kinday: I copied it from `isTemplateOnItsOwnLine`, but honestly have no idea what it does. Is it something that should be there?

kinday:
It is a bit out of scope of initial problem, but this place asks for quick refactoring IMHO:

https://github.com/prettier/prettier/blob/2e73164b81cc165ab55c957ead7b7c9a957c9d69/src/language-js/printer-estree.js#L625-L631

Things I see:

- `isTemplateOnItsOwnLine` and `hasLeadingOwnLineComment` could have same signature
  ```js
  isTemplateOnItsOwnLine(n.body, options.originalText, options)
  hasLeadingOwnLineComment(n.body, options.originalText, options)
  ```

- `options.originalText` looks redundant as whole `options` object is passed

evilebottnawi:
@kinday Can we review? Maybe we can do refactoring in other PR?

kinday:
@evilebottnawi, separate PR with refactoring works fine for me.

I think we can wait with review as I found a thing that should be fixed probably.

If we take this example code as input:
```js
styled.div`
  ${props => props.display && props.block && props.inline && css`
    display: ${utils.display};
  `}
`
```

Currently it is formatted like this:
```js
styled.div`
  ${props => props.display &&
  props.block &&
  props.inline && css`
    display: ${utils.display};
  `};
`;
```

While it seems to be better to format this way:
```js
styled.div`
  ${props =>
    props.display &&
    props.block &&
    props.inline && css`
      display: ${utils.display};
    `};
`;
```

Or maybe this way (yet I find it awkward):
```js
styled.div`
  ${props =>
    props.display &&
    props.block &&
    props.inline &&
    css`
      display: ${utils.display};
    `};
`;
```

Should we have discussion in initial issue or we can make decision right away?

-----

I am sorry if I am taking too much time with all my questions, but I never contributed to such a large project and don’t really know how decisions are being made on that level.

evilebottnawi:
@kinday better:
```
styled.div`
  ${props =>
    props.display &&
    props.block &&
    props.inline &&
    css`
      display: ${utils.display};
    `};
`;
```

evilebottnawi:
@kinday hm, looks better, let's listen other opinions
/cc @vjeux @lydell @j-f1 @azz

src/language-js/printer-estree.js
duailibe: I don't see why this is necessary. Did you try without this?

src/language-js/printer-estree.js
```diff
@@ -5233,6 +5245,15 @@ function templateLiteralHasNewLines(template) {
   return template.quasis.some(quasi => quasi.value.raw.includes("\n"));
 }
 
+function isStyledComponentsCssTemplate(n, text, options) {
+  return (
+    n.type === "LogicalExpression" &&
+    n.right.type === "TaggedTemplateExpression" &&
+    n.right.tag.name === "css" &&
```
duailibe: This is confusing.. the function is called "is styled components css template" but it checks if it's a `LogicalExpression`.

I say you either rename the function or "expect" the node to be `TaggedTemplateExpression`. And `n.type === "LogicalExpression"` would be moved out of this function, i.e.:

```js
// instead of
isStyledComponentsCssTemplate(n)

// would be
n.type === "LogicalExpression" && isStyledComponentsCssTemplate(n.right)
```

Also, it doesn't seem too restrictive to be called "is styled components css template", looks like "is css template" would be equally descriptive.

kinday: You’re right. (:

This part will probably be gone if I implement formatting I have shown above.

duailibe:
Sorry I didn't notice your comment above. The changes look good to me! :)

kinday:
@evilebottnawi, sorry people, won’t be able to work on it until Sunday.

kinday:
Friendly notification: I’m back to working on it.

kinday:
Whew, that was one of a brain teaser for me!

I would like to take a fresh look at the code tomorrow myself and will try running it against two real codebases to see the impact. Also I have one test related to Flow, which I don’t know how to fix due to no experience with Flow itself.

Otherwise I am happy with the results so far and brave enough to say it is ready for review.

tests/binary-expressions/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -154,11 +154,12 @@ const avatar = hasAvatar && <Gravatar user={author} size={size} />;
 
 const avatar = (hasAvatar || showPlaceholder) && <Gravatar user={author} size={size} />;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-const user = renderedUser || (
-  <div>
-    <User name={this.state.user.name} age={this.state.user.age} />
-  </div>
-);
+const user =
```
duailibe: I think you went too far 😅 Changes like this are regressions...

I think this PR should only affect tagged templates inside a tagged template

kinday:
@evilebottnawi, real codebase reveals some issues not covered with tests. ): I’ll add test cases for them today or tomorrow and will fix the problems.

E.g. indentation and grouping are broken when there’s less space for lines:
```diff
                       playerProgress={this.props.completion}
-                      onClick={() =>
-                        this.props.completion >= item.progressPercent &&
-                        this.props.redirectTo('/adventure')
-                      }
+                      onClick={() => this.props.completion >=
+                        item.progressPercent &&
+                      this.props.redirectTo('/adventure')}
                     >
```

evilebottnawi:
@kinday good catch :+1:

kinday:
Hey folks, it’s finally green and running on real codebase doesn’t make undesirable changes for me.

Could you please take a look at what we got here?

Also I want to know if I should fix what @duailibe mentioned as regression?

lipis:
https://deploy-preview-4266--prettier.netlify.com/playground/ Doesn't seem to work from here. Chrome/macOS.

duailibe:
Needs this commit edcebb0d546c9a9997a796f6a620806efa9a411d

kinday:
@lipis, doesn’t seem like PR’s fault to me. Rebased on latest `master`. Hope it’ll help.

kinday:
Can I do something else to help or should I just be patient?

duailibe:
Hi @kinday sorry for the delay. I still think this PR should only focus on fixing the ``css` ``  cases.

While I don't think the way we print stuff is great, I don't to cause unnecessary churn when people upgrade for no clear benefits, like below:

```js
foo = bar &&
  baz && {
    spam
  };
// vs
foo =
  bar &&
  baz && {
    spam
  };
```

If you feel like the one below is better for ``css` `` go ahead with it. But I wouldn't change that for the others.

kinday:
@duailibe, I’m not as quick as I’d like to be this time, so we’re even now. (;

So, what I am going to do now is these steps:
- close this PR as it has grown too much;
- create a new with fix _only_ for Styled Components issue;
- create an RFC* on general JS formatting changes introduced in this PR.

* Are you practicing RFCs? Is there anything I should know before opening one?

Issue #2550, opened 2017-07-27
Keep multi-line formatting for destructuring assignment 


Brian-Gaffney:
I think multi-line destructuring should not be collapsed by Prettier.

Same concept as the [multi-line objects special case](https://github.com/prettier/prettier/blob/master/Rationale.md#multi-line-objects).

When doing destructuring (particularly nested) new lines can greatly enhance readability.

Example of current behaviour:
_Input_
```javascript
// Object with multiple lines uncollapsed by Prettier
const user = {
  name: {
    first: "Zoid",
    last: "Berg",
  },
};

// Similar destructuring assignment is collapsed by Prettier
const {
  name: {
    first,
    last,
  },
} = user;
```

_Prettier output_
```javascript
// Object with multiple lines uncollapsed by Prettier
const user = {
  name: {
    first: "Zoid",
    last: "Berg"
  }
};

// Similar destructuring assignment is collapsed by Prettier
const { name: { first, last } } = user;
```

azz:
The output seems perfectly readable to me.

The special-casing of objects is more of a failing in prettier than a feature, I don't think it makes sense to spread that to more cases.

Brian-Gaffney:
I agree that the above example is quite readable, I was only going for a minimal example case.

For larger and deeper destructuring assignments I find it harder to read.

**Larger example**
Input:
```javascript
const {
  name: {
    first,
    last,
  },
  organisation: {
    address: {
      street: orgStreetAddress,
      postcode: orgPostcode,
    },
  },
} = user;
```

Prettier output:
```javascript
const {
  name: { first, last },
  organisation: { address: { street: orgStreetAddress, postcode: orgPostcode } }
} = user;
```

If there's no interest in this change then I'm fine to drop it.
I found it a bit of a barrier to adoption at work where we have some very big props destructuring in React components and I thought others might have the same issue.

azz:
Agreed, the input looks better.

Maybe a better rule would be "always break nested object patterns onto multiple lines", rather than looking at the source. Though that rule might have to only apply to variable declarations, else this will be an issue:


```js
function f({ data: { name } }) {}
//Out:
function f({
  data: {
    name
  }
}) {}
```

Which would likely be seen as a regression.

Brian-Gaffney:
For React pure render functions I do find myself writing them similarly to your example output above.
Each destructed variable is on a new line.

```javascript
export default UserComponent ({
  name: {
    first,
    last,
  },
  organisation: {
[...]

PR #4267, opened 2018-04-06
Break `ObjectPattern`s with nested `ObjectPattern`s

duailibe:
Context: #2550

After significant demand, this PR introduces an heuristic to always break `ObjectPattern`s if there are any nested `ObjectPattern` (or `ArrayPattern` in that case) properties.

I chose not to use the "always break if any property has a value" for now and not to break the outermost `ObjectPattern` if it's a function argument.

I personally think it's a good compromise, but I didn't mark to close the issue yet.

Let me know what you think!

### DIFF
--- a/tests/destructuring/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/destructuring/__snapshots__/jsfmt.spec.js.snap
@@ -4,9 +4,56 @@ exports[`destructuring.js 1`] = `
 const [one, two = null, three = null] = arr;
 a = ([s=1,]) => 1
 const { children, ...props } = this.props
+
+const { user: { firstName, lastName } } = this.props;
+
+const {
+  name: { first, last },
+  organisation: { address: { street: orgStreetAddress, postcode: orgPostcode } }
+} = user;
+
+function f({ data: { name } }) {}
+
+const UserComponent = function({
+  name: { first, last },
+  organisation: { address: { street: orgStreetAddress, postcode: orgPostcode } },
+}) {
+  return
+};
+
+const { a, b, c, d: { e } } = someObject;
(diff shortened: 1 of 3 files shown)


tests/destructuring/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4,9 +4,56 @@ exports[`destructuring.js 1`] = `
...
+
+const { a, b, c, d: { e } } = someObject;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 const [one, two = null, three = null] = arr;
 a = ([s = 1]) => 1;
 const { children, ...props } = this.props;
 
+const {
+  user: { firstName, lastName }
+} = this.props;
```
duailibe: I'm on the fence with this one... I personally prefer it collapsed, but don't care either way

duailibe:
@tennisgent Would you open a new issue with this please?

tennisgent:
Certainly. Thanks for the quick response. 👍

amandakoster:
Hi Everyone,  I am using Eslint and Prettier, and am wondering if there is a Prettier rule I can add to my .eslintrc file to keep the destructure multi line vs. single line? Thank you!

j-f1:
@amandakoster Unfortunately, we don’t offer those kinds of options. Sorry :/

tennisgent:
As per https://github.com/prettier/prettier/issues/4384, the maintainers decided not to make an option for this. So it's all or nothing. Not configurable.

PR #4275, opened 2018-04-06
feat(markdown): support hasPragma/insertPragma

ikatyang:
_hasPragma_

**support**

```md
<!-- @prettier -->
```

```md
---
something
---

<!-- @prettier -->
```

```md
<!--
  MIT LICENSE 
  
  bla bla bla

  @format
-->
```

---

_insertPragma_

**support**

```diff
+ <!-- @prettier -->
+
something something
```

```diff
---
seomthing
---

+ <!-- @prettier -->
+
something something
```

**does not support**

```diff
<!--
  MIT LICENSE 
  
  bla bla bla
+
+ @format
-->
```

### DIFF
--- a/src/language-markdown/pragma.js
+++ b/src/language-markdown/pragma.js
@@ -1,0 +1,12 @@
+"use strict";
+
+const pragmas = ["format", "prettier"];
+const startWithPragmaRegex = new RegExp(
+  `^<!--\\s*@(${pragmas.join("|")})\\s*-->`
+);
+
+module.exports = {
+  startWithPragmaRegex,
+  hasPragma: text => startWithPragmaRegex.test(text.trimLeft()),
+  insertPragma: text => `<!-- @${pragmas[0]} -->\n\n${text}`
+};
--- a/src/language-markdown/pragma.js
+++ b/src/language-markdown/pragma.js
@@ -5,8 +5,25 @@ const startWithPragmaRegex = new RegExp(
   `^<!--\\s*@(${pragmas.join("|")})\\s*-->`
 );
 
+function extract(text) {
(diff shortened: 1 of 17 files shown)


src/language-markdown/pragma.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+const pragmas = ["format", "prettier"];
+const startWithPragmaRegex = new RegExp(
+  `^<!--\\s*@(${pragmas.join("|")})\\s*-->`
+);
+
+module.exports = {
+  startWithPragmaRegex,
+  hasPragma: text => startWithPragmaRegex.test(text.trimLeft()),
```
suchipi: Do we need to worry about this case?

suchipi: Or:
```md
<!--
  MIT LICENSE

  bla bla bla
-->

<!-- @format -->
```

ikatyang: We should worry about the first one, but I'm not sure about the second one, maybe just doing something like JS?
```md
<!--
  MIT LICENSE 
  
  bla bla bla

  @format
-->
```

suchipi: That looks good to me :+1:

tests/insert-pragma/markdown/simple.md
```diff
@@ -0,0 +1,5 @@
+
+
```
j-f1: Can you add a test without these blank lines?

ikatyang:
Don't know how to handle this case so I think we can just apply the common rule for now.

```diff
+<!-- @format -->

 <!--
   some
     thing
    not 
      indented
 well
 -->
```

Issue #1283, opened 2017-04-14
Are any plugins while use d3.js ?


litinskii:
Are any plugins while use d3.js ?

vjeux:
Hey, I'm not really sure what you have in mind with d3? Could you expand.

finetype:
My guess would be that when one is writing d3, which features function chaining very heavily, one might desire slightly different formatting.

vjeux:
Going to close this as it's not actionable.

Peeja:
[Here's an example](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VACAYgJwgWwwAsYYAHAZyXQCM4YBDAOghorhwDcGaAbOIgEcmkfKgAC+GhAowIYANaoA5r1YNeAWnj4yHBjACuOONpNQAJhQA6UW8FsYMkKLIwVOyjAF4MFgMxM7PxgMAAUACIA8gCyQZ5hAO4AlhYwRAA0xHDJyiQAlPkA3I4YpR7KTAxkepZh1iDKDfmlTk4iGrxhAB4Agt3JFMWl5Z5VNQgW9Y3NrW0dvF0Anv2Dw1BOo5XVtVMNTSAtG23tBjA407J4CnANWQ0AxAAML7PHJ1WkFw1XEDeaEDIDDAyRgSzuGCeTAATEcPsE4KFeotpiCcGB+G8PhYDAwwjjGHD5gh4Bd8uNdqjkujMYc5qcvqjuhCLD4AHwYbr4pgEuCFekYT7nVHgkBZVneDlLbncXiGPlEj5nb4gABmyUWLPZGAAXjKNPL+e95srpjgIdCmABWdabY4mIw4DYVJhQCAWOBhdYAXzFIEBMGS0AoyFADBweESAAVwwgQygGJwIKk-QxZMhVRp2BkQDQcMCbjAAMpAkFQZTIc7ynMWeQZrNwHPJVwcGBR-PKfAMeu8bMgABWFG6ACF84p6EWGPg4AAZZtwHt9iCGcgr6GLxsgIE4dg4ZC5nhLNRQP1kHDNmAAdVS6WQAA4njmzzI4Jf82R92e4LvOAucyZBEMak4HbBhO27JBM17Td2HwZJKxwasQAoZtVDgABFQwIHgDcc0YGhrzSIhkGhPD8w1VCAGECC7fc3SgP8QEMdgABUeHjKD2G9b0gA):

```js
[...]

PR #4285, opened 2018-04-09
Nicer line wrapping for d3

1wheel:
#1283 

```
d3.select('body')
  .append('circle')
  .at({width: 30, fill: '#f0f'})
  .st({fontWeight: 600})
```

instead of 

```
d3
  .select('body')
  .append('circle')
  .at({ width: 30, fill: '#f0f' })
  .st({ fontWeight: 600 })
```

I'm not familiar with `jsfmt` and I'm not sure how to add a test for this; please let me know what changes are needed.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -4108,9 +4108,9 @@ function printMemberChain(path, options, print) {
   //
   // In order to detect those cases, we use an heuristic: if the first
   // node is an identifier with the name starting with a capital letter,
-  // containing only _ and $, or shorter than tabWidth. The rationale
-  // is that they are likely to be factories.
-  function isFactory(name) {
+  // or shorter than tabWidth. The rationale is that they are likely 
+  // to be factories.
+  function isFactoryOrShort(name) {
     return name.match(/(^[A-Z])/) || name.length <= options.tabWidth;
   }
   const shouldMerge =
@@ -4119,12 +4119,12 @@ function printMemberChain(path, options, print) {
     ((groups[0].length === 1 &&
       (groups[0][0].node.type === "ThisExpression" ||
         (groups[0][0].node.type === "Identifier" &&
-          (isFactory(groups[0][0].node.name) ||
+          (isFactoryOrShort(groups[0][0].node.name) ||
             (groups[1].length && groups[1][0].node.computed))))) ||
       (groups[0].length > 1 &&
         groups[0][groups[0].length - 1].node.type === "MemberExpression" &&
         groups[0][groups[0].length - 1].node.property.type === "Identifier" &&
-        (isFactory(groups[0][groups[0].length - 1].node.property.name) ||
+        (isFactoryOrShort(groups[0][groups[0].length - 1].node.property.name) ||
           (groups[1].length && groups[1][0].node.computed))));
 
   function printGroup(printedGroup) {
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -4108,7 +4108,7 @@ function printMemberChain(path, options, print) {
   //
   // In order to detect those cases, we use an heuristic: if the first
   // node is an identifier with the name starting with a capital letter,
-  // containing only _ and $, or shorter than tabWidth. The rationale 
+  // containing only _ and $, or shorter than tabWidth. The rationale
   // is that they are likely to be factories.
   function isFactory(name) {
     return name.match(/(^[A-Z])|^[_$]+$/) || name.length <= options.tabWidth;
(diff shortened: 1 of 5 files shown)


src/language-js/printer-estree.js
```diff
@@ -4111,7 +4111,7 @@ function printMemberChain(path, options, print) {
   // letter, just a sequence of _$ or this. The rationale is that they are
```
j-f1: While you’re here, could you update this comment (remove the “or this” part and add something about your new changes)?

1wheel:
Got it. Updated the comment and added a test.

Thanks for the help!

petulla:
@j-f1 Will this be merged? Am in need of this same solution.

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1409,9 +1409,8 @@ function memberOutside() {
 
 function memberInAndOutWithCalls() {
   return (
-    a
```
vjeux: This changes the meaning of the text, can you add a few more characters here?

src/language-js/printer-estree.js
```diff
@@ -4107,11 +4107,11 @@ function printMemberChain(path, options, print) {
...
   // In order to detect those cases, we use an heuristic: if the first
-  // node is just an identifier with the name starting with a capital
-  // letter, just a sequence of _$ or this. The rationale is that they are
-  // likely to be factories.
+  // node is an identifier with the name starting with a capital letter,
+  // containing only _ and $, or shorter than tabWidth. The rationale
+  // is that they are likely to be factories.
   function isFactory(name) {
-    return name.match(/(^[A-Z])|^[_$]+$/);
+    return name.match(/(^[A-Z])|^[_$]+$/) || name.length <= options.tabWidth;
```
vjeux: You can get rid of the special case for _$ since it's always going to be <= tabWidth

src/language-js/printer-estree.js
```diff
@@ -4107,11 +4107,11 @@ function printMemberChain(path, options, print) {
   //     .map(x => x)
   //
   // In order to detect those cases, we use an heuristic: if the first
-  // node is just an identifier with the name starting with a capital
-  // letter, just a sequence of _$ or this. The rationale is that they are
-  // likely to be factories.
+  // node is an identifier with the name starting with a capital letter,
+  // containing only _ and $, or shorter than tabWidth. The rationale
+  // is that they are likely to be factories.
   function isFactory(name) {
```
duailibe: I don't think short names mean they are likely to be factories.

It's a different rationale.. so I'd have a check for `isFactory(name) || name.length <= options.tabWidth`

src/language-js/printer-estree.js
j-f1: @vjeux Note the `+` — `$$$$$`, `_____`, and `$_$_$` would all count as factory functions, but still be longer than `tabWidth`.

vjeux: If you write this kind of variables, I'm not going to feel sorry for you :p

j-f1: I agree it’s not a good idea, but I wanted to point it out anyway in case you missed an edge case.

src/language-js/printer-estree.js
1wheel: That'll be a little gross to call: 

```
isFactory(groups[0][groups[0].length - 1].node.property.name) || 
  groups[0][groups[0].length - 1].node.property.name.length <= options.tabWidth
```

What about renaming the function? `isFactoryOrShort` or `isNoWrap`?

tests/comments/__snapshots__/jsfmt.spec.js.snap
1wheel: Added.

src/language-js/printer-estree.js
1wheel: Removed the `$` special case; `$_$` no longer counts as a factory function.

src/language-js/printer-estree.js
duailibe: Sure, that works

1wheel: I updated the function name and comment. The travis build is failing now though, because of formatting. Both `yarn link` and `yarn lint --fix` just hang when I try to run them locally.

j-f1: It looks like you have some trailing whitespace in a comment. Removing that should appease the linter.

1wheel: Ah. Took it out.

1wheel:
Does this need anything else @j-f1?

j-f1:
Nope, not as far as I can see.

j-f1:
Thanks for contributing!

PR #4288, opened 2018-04-09
fix(glimmer): preserve leading/trailing spaces in text nodes undernea…

n8n8baby:
…th attribute nodes when next to mustache statements.


suchipi:
Does this only affect attributes containing mustache?

PR #4296, opened 2018-04-10
[flow] Upgrade flow to 0.69 and enable ?. support

vjeux:
Flow now support ?. so this PR enables it in prettier


bakkot:
@vjeux https://github.com/prettier/prettier/pull/4244 ?

vjeux:
Oh great, thanks for pointing it out to me!

PR #4299, opened 2018-04-11
Include webpack to bundle the playground

duailibe:
This is a step to unblock #3573 so we can use ES6 syntax and modules to have a better playground. Should not change anything in how the playground works today.


website/package.json
```diff
@@ -1,17 +1,21 @@
 {
   "license": "MIT",
   "scripts": {
-    "build": "docusaurus-build",
+    "build": "webpack -p && docusaurus-build",
     "publish-gh-pages": "docusaurus-publish",
     "start": "docusaurus-start",
```
j-f1: What about changing this to `docusaurus-start & webpack --watch` (just one `&`) to run them in parallel?

lydell: https://github.com/kimmobrunfeldt/concurrently

PR #4312, opened 2018-04-13
fix(schema): support array type

ikatyang:
Forgot to take `option.array` into account.

### DIFF
--- a/scripts/generate-schema.js
+++ b/scripts/generate-schema.js
@@ -2,14 +2,17 @@
 
 "use strict";
 
-const prettier = require("..");
-
-console.log(
-  prettier.format(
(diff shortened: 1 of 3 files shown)


scripts/generate-schema.js
```diff
@@ -2,7 +2,7 @@
 
 "use strict";
 
-if (process.env.NODE_ENV === "test") {
+if (process.env.NODE_ENV === "test" || process.env.CI) {
```
j-f1: What about using `if (requre.main === module)` as the condition ([more info on Stack Overflow](https://stackoverflow.com/a/6398335/5244995)), and inverting the two blocks?

This way, you can still `require()` the module outside of CI/tests.

PR #4317, opened 2018-04-15
Allow trailing comma in scss list and maps

### LINKED ISSUE #4076: trailingComma setting is not respected in SCSS lists/maps.
**Prettier 1.11.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASAtgTwLSQDYQBOAzkgAQAUAOlGWQOSFwAm9ANDXfQEZ4CucdjQCUAbhBsQEAA4wAltGLJQAQ0KEIAdwAKahEpQqAbhDnMJIFcRjIAZirzE4k7oRVgA1nBgBlae7koAHNkGEIBSWYIMDsHJ0lAp0IYbTcg9BVYx2cQACtiAA8AITdPbx8VdDgAGUC4LPipPhhpZoAmBpz-EjhCZBBiMGIlSWlCQJgAdTMYAAtkAA4ABlGNJ0m3aX6xuCSjeskmAEc+OSZUlXTMpHtsySd0OVDwnOJAoLw4AEU+CHhOyQwFTcabMObINqAtxyPDvADCEHQGX6DjwFj4TgAKsCDLcnABffFAA)
```sh
--parser scss
--trailing-comma all
```

**Input:**
```scss
$my-colors: (
  'red',
  'blue',
);
```

**Output:**
```scss
$my-colors: (
  "red",
  "blue"
);

```

**Expected behavior:**

This might be intentional, but is there a reason why the `trailingComma` setting isn't respected in other-than-JS languages (e.g. SCSS)?

ad1992:
fix #4076

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -15,7 +15,9 @@ const group = docBuilders.group;
 const fill = docBuilders.fill;
 const indent = docBuilders.indent;
 
+// utils
 const removeLines = doc.utils.removeLines;
+const utils = require("./utils");
 
 function genericPrint(path, options, print) {
   const node = path.getValue();
@@ -679,7 +681,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
-          isSCSS(options) &&
+          utils.isSCSS(options.parser, options.originalText) &&
           node.groups.length > 1 &&
           options.trailingComma !== "none"
             ? ","
@@ -985,13 +987,6 @@ function isWideKeywords(value) {
     ["initial", "inherit", "unset", "revert"].indexOf(
       value.replace().toLowerCase()
     ) !== -1
-  );
-}
-
-function isSCSS(options) {
-  const IS_POSSIBLY_SCSS = /(\w\s*: [^}:]+|#){|@import[^\n]+(url|,)/;
-  return (
-    options.parser === "scss" || IS_POSSIBLY_SCSS.test(options.originalText)
   );
 }
 
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -666,6 +666,7 @@ function genericPrint(path, options, print) {
         declNode &&
         declNode.type === "css-decl" &&
         declNode.prop.startsWith("$");
+
       return group(
         concat([
           node.open ? path.call(print, "open") : "",
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -679,7 +679,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
-          node.groups.length > 1 && options.trailingComma === "es5" ? "," : "",
+          node.groups.length > 1 && options.trailingComma !== "none" ? "," : "",
           softline,
           node.close ? path.call(print, "close") : ""
         ])
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -679,7 +679,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
-          shouldPrintTrailingComma(options.trailingComma, node) ? "," : "",
+          node.groups.length > 1 && options.trailingComma === "es5" ? "," : "",
           softline,
           node.close ? path.call(print, "close") : ""
         ])
@@ -982,16 +982,6 @@ function isWideKeywords(value) {
       value.replace().toLowerCase()
     ) !== -1
   );
-}
-
-function shouldPrintTrailingComma(choice, node) {
-  if (node.groups.length === 1) {
-    return false;
-  }
-  if (choice === "all") {
-    return true;
-  }
-  return false;
 }
 
 module.exports = {
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -679,7 +679,11 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
-          isSCSS(options) && node.groups.length > 1 && options.trailingComma !== "none" ? "," : "",
+          isSCSS(options) &&
+          node.groups.length > 1 &&
+          options.trailingComma !== "none"
+            ? ","
+            : "",
           softline,
           node.close ? path.call(print, "close") : ""
         ])
@@ -986,7 +990,9 @@ function isWideKeywords(value) {
 
 function isSCSS(options) {
   const IS_POSSIBLY_SCSS = /(\w\s*: [^}:]+|#){|@import[^\n]+(url|,)/;
-  return options.parser === 'scss' || IS_POSSIBLY_SCSS.test(options.originalText);
+  return (
+    options.parser === "scss" || IS_POSSIBLY_SCSS.test(options.originalText)
+  );
 }
 
 module.exports = {
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -666,7 +666,6 @@ function genericPrint(path, options, print) {
         declNode &&
         declNode.type === "css-decl" &&
         declNode.prop.startsWith("$");
-
       return group(
         concat([
           node.open ? path.call(print, "open") : "",
@@ -679,6 +678,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
+          shouldPrintTrailingComma(options.trailingComma, node) ? "," : "",
           softline,
           node.close ? path.call(print, "close") : ""
         ])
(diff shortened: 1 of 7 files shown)


src/language-css/printer-postcss.js
```diff
@@ -679,6 +679,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
+          shouldPrintTrailingComma(options.trailingComma, node) ? "," : "",
```
j-f1: How do you feel about refactoring this to `options.trailingComma && node.groups.length > 1` instead of a separate function?

Also, by testing the truthiness of `options.trailingComma`, instead of checking for `"all"`, we can print trailing commas if the value is `"es5"`.

src/language-css/printer-postcss.js
```diff
@@ -679,6 +679,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
+          node.groups.length > 1 && options.trailingComma === "es5" ? "," : "",
```
j-f1: Sorry, I was unclear in my first review. We should print a comma here if `options.trailingComma !== "none"`, since we don’t care which option the user picks in this case (`all` or `es5`).

tests/css_trailing_comma/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -11,16 +11,64 @@ a {
 
 `;
 
+exports[`declaration.css 2`] = `
+a {
+    margin: $bar,;
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+a {
+  margin: $bar;
```
lydell: What is this test trying to say? Are you specifically trying to remove that trailing comma, or is this just an observation you made? I guess it doesn’t really matter what happens to the comma in this case. Just curious.

ad1992: @lydell here the trailing comma should not be shown as it was previously in `line 9`. I had just added `trailingComma: "es5" ` in `jsfmt.spec.js` so this is showing here as all the tests are again run for `trailingComma: "es5" `

ad1992:
Sure I understand and will take care of the bug fixes when the release goes live.

tests/css_trailing_comma/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -11,16 +11,64 @@ a {
...
+$list: (a);
+$list: (
+  "Helvetica",
+  "Arial",
+  sans-serif
+);
+$colors: (
+  "red",
+  "blue"
+);
```
evilebottnawi: Let's add tests for nested maps and lists inside maps

src/language-css/printer-postcss.js
```diff
@@ -679,6 +679,7 @@ function genericPrint(path, options, print) {
               )
             ])
           ),
+          node.groups.length > 1 && options.trailingComma !== "none" ? "," : "",
```
evilebottnawi: Can we add check what it is `isSCSS()`, in theory we can break less code, some projects can have scss/less files togethe (legacy or migration period)

evilebottnawi: @ad1992 nested maps, when variable contain map and each key contain other map/list

tests/css_trailing_comma/__snapshots__/jsfmt.spec.js.snap
ad1992: I have added the tests. is this the expected behaviour when nested [playgroundlink](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASSAbCAnAzkgAgAsIYAHASygGsCATOOM3agTwCMMBXOAgMwrY47bBTDUANAG4AOlFQYKuGIQAUAQwkBKWfMXK1AcgAScDADc4MMesMSChgIKj1GOwVzqouALS44onzauugQWHhqcgQEMiBCdLESUTEgnDyxcjpyoVACAOaRUNEwRHAAtnD4BKrJ0WVKKtW10cRw6gzYhADEdGB86mAADElFLdGQsAiNXYODAJwAbAAc7CNj0XwQpAHdSwCsYADsc2DNWmstqLhkolAFTaNjpe07BF18ACz9RxdjE-CwbpgABMHz2Hzgvxam22nTefD47AAjHslmdklpMrpsiI2tQyBAqDAfGV1GRCtFcKSMBgKS16lAfAB3Ch0EqEKBcGlQ0kAD2ZrPZBA+xzIvKh7HU-h8m0BBCRCzFUMs2GsYFcPmwRFYJTKhCRADoAMwYi4VOgULh6h5jBkCtlEQgfJaDJXNPn2oVzPZzN2PAiS6WyxpIpZ+sYqtUarU6ojWw0fJEfU3JDDqbB5OB0upUT2OgiLV3i93qfksh36+a+4v+wNwGXQENhmsRgJRjCa7W6-UGvYp0a8tMZrM2+m58tCpFI2bh+mlvMcrkYCVS+vBwjApGz6KRmwdmPd+UGhahjFYuQgCQgCBkazQXDIUDp7AQJkABXTCAfKHU5kJdEvEApRgZB+gwfwrxEAZqCsABlMgBioPJkBgbAeCvOgIDAUDXAgkAqH8VU32wdQ8lJHDwMhEAACtcF5AAhEjxDg9QKgAGSoOAKLwiAuHIPjgW4qiELwAJkBAXAwFwB8rxuIkAHVBSIZAXVkl9-HkkiyHEm5KgCSxAKEABHLhBDgYjSPIpAwLw-x6hQtCqNwJCMDgABFLhtiEq8YHUdhFIdZBgR8kiKEUO4AGEIDKKyQCgaAuKvLh-AAFT878bLgABfLKgA) ?

According to style guide trailing comma should not be showing when in single line so is there any better way of detecting if its exceeding printWidth except for looping through the node node and checking the length occupied and then decide whether to put trailing comma ?

src/language-css/printer-postcss.js
ad1992: sorry i guess there is some confusion. i understood about adding nested map and lists. But i didnt get `Can we add check what it is isSCSS(), in theory we can break less code, some projects can have scss/less files togethe (legacy or migration period`

So is there any specifice check you meant in the above comment  except for nested map and lists ?

evilebottnawi: @ad1992 parsers some times change ast and introduce new nodes, better check directly what is `SCSS` to avoid break something in future

ad1992: @evilebottnawi so you mean we need to make sure trailing comma is added only in case of  `scss`. 
I referred to `parser-postcss.js` file and added check for `isSCSS`. 
Is this fine ? 
Sorry since I am new to prettier codebase and AST, I am asking too many questions.

ad1992:
@j-f1 made the changes. Can we make yarn lint as a part of pre-commit hook ? so that user is forced to fix the linting issues before he pushes the commit.

j-f1:
@ad1992 I’m personally not a fan of pre-commit hooks, since they make it take longer to commit. If you can, I’d recommend installing a Prettier editor integration so the code gets formatted when you save.

ad1992:
@lipis making it part of pre-commit will take the same time `yarn lint` takes. So if we have pre-commit hook then it will be helpful for all the contributors as they will be forced to fix the issues and they don't have to set up in local. 
Actually if we add a pre-commit then we should add two things
```
yarn lint 
yarn test
```
So that all checks pass in travis.

j-f1:
@ad1992 Since the Travis run for the last commit took ten minutes, it’s reasonable to expect that each commit would take at least five minutes to be saved if `yarn test` was included. IMO this is an excessively  long wait, especially since you can always push a commit that fixes the tests, or your PR is supposed to add a failing test case.

ad1992:
@j-f1  I understand and agree that it will make the commit process slower. But in case of `pre-commit` hook one can always commit with `-nm` if the user wants to skip it.  While pushing the commits I often forget to make the linting fixes and when travis fails I realize I forgot to make changes and push the linting fix and wait for it to pass. So I suggested if pre-commit hook if it can make it better.  No issue if we don't add it, I understand your concern, and will make the local settings changes as you suggested 👍

lydell:
I usually use a pre-push hook (in addition to an editor integration). I commit way more often than I push. Makes the push slower, but faster than having CI fail.

ad1992:
@j-f1 anything else I can do to help get this merged?

duailibe:
@ad1992 Thanks for the contributions! 💯

PR #4321, opened 2018-04-16
Add a redirects file

j-f1:
I was looking through [Google’s crawl errors](https://www.google.com/webmasters/tools/crawl-errors?siteUrl=https://prettier.io/) for the website <sup>(link only works if you have access)</sup>, and this should fix most of them.

### DIFF
--- a/website/static/_redirects
+++ b/website/static/_redirects
@@ -1,7 +1,3 @@
 # Redirect removed doc links
 /docs/:lang/usage.html          /docs/:lang/install.html
 /docs/:lang/usage-options.html  /docs/:lang/options.html
-
-# Accommodate links to https://prettier.io/foo.js:line:col
-/lib/index.js*                  /lib/index.js
-/worker.js*                     /worker.js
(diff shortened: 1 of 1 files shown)


ikatyang:
For reference: https://www.netlify.com/docs/redirects/

Test: https://deploy-preview-4321--prettier.netlify.com/docs/en/usage-options.html

j-f1:
Unfortunately, it doesn’t work for https://deploy-preview-4321--prettier.netlify.com/lib/index.js:2312:75 ☹️
I contacted support, and I’ll see what they say.

j-f1:
Support said that the `:foo` and `*` matches only work on a full path component (i.e. `/foo/:bar/baz`, not `/foo/item-:bar/baz`), so we can’t use that here. However, we could load all of the JS files with a trailing `?` which would make the line numbers go in the query string, or a trailing `#` which would make them go in the hash. I’ll remove those redirects for now and merge. If anyone wants to make the JS line references work later, feel free!

PR #4325, opened 2018-04-16
[flow] Fix union indent inside of function param with a name

### LINKED ISSUE #4323: Regression for flow union
**Prettier 1.12.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEATOYA2BDATnAAi2wGcSCAxCXAWwBFsZsCDgAdKF7AB24VQAUHFiwjcYAS2gkA-EmEiWAHwIkYuCVADmCxSvadFRgGYTMcbowAWc1es1aANLqMtT5qNhpxbajdudDVwBfQJYASiQCADcICVQAbg5gjhBHEDFJaWRQPFwIAHcABTwEEmQQbFj4tMq1ZGNsTBI4dIAjXGwwAGs4GABlSzAHZHUAV1a0CDAGppb0zRbcGCLOrRpsWebJgCsSAA8AIU6evv6vOAAZTTgt+YyxmG5HgCY7yctcJYq27DaAT0w0Fq3H8MAA6vEYFZkAAOAAM6VBEBa4M63AqoLgS2it3S+AAjmMJPhVth1pskI1tukWjQJKNcBNaQ5zABFMYQeDvdJMNqQ1DQ5AvXmdMwOADCEBoGwqUGgeJAYxaABU-uUqXM4MFgkA)

**Input:**
```jsx
declare class FormData  {
  append(
    options?:
      | string
      | {
          filepath?: string,
          filename?: string,
        },
  ): void;
}

```

**Output:**
```jsx
declare class FormData {
  append(
    options?: | string
    | {
        filepath?: string,
        filename?: string
      }
  ): void;
}

```

This is a regression from 1.9. It's blocking the release of 1.12 at Facebook. We should get it fixed and send a 1.12.1.

vjeux:
Fixes #4323

See #4323 for all the context.

I'd love to get it merged and ship a dot release if possible with it. It's affecting a bunch of code and prevents upgrading to the latest version of prettier at Facebook. I know it's been there for a while (since 1.9.3...)

### DIFF
--- a/tests/union_intersection/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/union_intersection/__snapshots__/jsfmt.spec.js.snap
@@ -49,6 +49,17 @@ type Props = {
       }
   ) => void
 };
+
+declare class FormData {
+  append(
+    options?:
+      | string
+      | {
+          filepath?: string,
+          filename?: string
+        }
+  ): void;
+}
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 type T5 =
   | "1"
@@ -117,5 +128,16 @@ type Props = {
       }
   ) => void
 };
+
+declare class FormData {
+  append(
+    options?:
+      | string
+      | {
+          filepath?: string,
+          filename?: string
+        }
+  ): void;
+}
 
 `;
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -2384,7 +2384,7 @@ function printPathNoParens(path, options, print, args) {
         parent.type !== "TSTypeParameterInstantiation" &&
         parent.type !== "GenericTypeAnnotation" &&
         parent.type !== "TSTypeReference" &&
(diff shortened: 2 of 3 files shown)


tests/union_intersection/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -118,4 +129,15 @@ type Props = {
   ) => void
 };
 
+declare class FormData {
+  append(
+    options?:
+      | string
+      | {
```
j-f1: Would this look better as `options?: string | {`?

vjeux: If I remember correctly, I only do the printing for undefined and null. We could extend it. I wanted to err on the safe side.

vjeux:
@suchipi: would you be able to cut 1.12.1 with this change by any chance? I really want to get the new release out to Facebook :p

suchipi:
@vjeux 1.12.1 is released

zertosh:
yay!

vjeux:
@suchipi thanks so much for the quick turnaround time!

suchipi:
You're welcome 😃

Issue #3557, opened 2017-12-23
--insert-pragma breaks GraphQL, CSS


azz:
**Prettier 1.9.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBHArnATgTwAQCKmuewAOlHngJYAmFAvhSADQgQAOM10AzsqACGWLBADuABWEJ+KQQDcIdViEG8YyAGaCANrzhsARlkFgA1nBgBlDqepQA5shhZMbWhDBbd+tvf1YMBImDgC2gs6uBiAAVrwAHgBCJuaWVoKhcAAy9nDeetEQ6DAcxQBM+b4gtlgByCAOJhwAFqg6KhxY9jAA6nQwzcgAHAAMbJ0Q+j1N9Z1wAfJ5bFhwGNQrwYJhEUjaBWz6odSRbiC89g46cEQQ8JXRMIKGfbQDyGVsLoLUOhcAwhBQuF6lBoEsQOh9AAVJ6yPb6BgMIA)
```sh
--insert-pragma
--parser graphql
```

**Input:**
```graphql
query Query {
  id
}

```

**Output:**
```graphql
SyntaxError: Syntax Error GraphQL request (1:1) Cannot parse the unexpected character "/".

1: /** @format */
   ^
2: 
 (1:1)
> 1 | /** @format */
    | ^
  2 | 
  3 | query Query {
  4 |   id
```

---
**Prettier 1.9.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBiABACzgGxxEAGhAgAcYBLaAZ2VAEMAnRiAdwAUmFaV6A3CBQAmREPWoxkAM3o5qcYgCNG9MAGs4MAMqlVFKAHNkMRgFcFIIRDDTZ84vvmMY7FQYC29Y2YsAragAeAEIq6ppa9O5wADL6cLZyFhCmMKQpAEwJ9iC6jE7IIJ6MalasUKKkjPowAOrCMJjIABwADMSVEPI1KqQFlXBOfPHEjHAAjqYUo670Hl5IMonE8u4U3ubL+gY4cACKphDwWRYw9Ip1Qg3I6cQm9BQ4WwDCEO6eBVDQwyCm8gAqZx4i3kAF8QUA)
```sh
--insert-pragma
--parser markdown
```

**Input:**
```markdown
# hello
```

**Output:**
```markdown
/\*_ @format _/

# hello

```

---
[...]

PR #4333, opened 2018-04-17
Fix pragma detection/insertion for JSON

duailibe:
Following https://github.com/prettier/prettier/issues/3557#issuecomment-356280409 we should always format JSON even with `--require-pragma` and never add a comment even with `--insert-pragma`.

Making `hasPragma` for JSON always return true makes Prettier always format the JSON file, and stop trying to insert pragma because it thinks the code already has a pragma.

### DIFF
--- a/src/language-js/index.js
+++ b/src/language-js/index.js
@@ -181,7 +181,7 @@ const parsers = {
   babylon,
   json: Object.assign({}, babylon, {
     hasPragma() {
-      return false;
+      return true;
     }
   }),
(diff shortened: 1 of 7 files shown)


j-f1:
We could also always quote props with `json`, and apply our current JS rules to `json5`.

duailibe:
@ikatyang Yeah, I think when we add JSON5 "proper" support, we can also make the correspondent pragma changes

Issue #4000, opened 2018-02-19
Plugin API not working with globally installed modules


czosel:
Currently, globally installed plugins don't seem to be discovered by the CLI:
```bash
npm install -g prettier @prettier/plugin-php
prettier --plugin=@prettier/plugin-php --parser=php test.php
# Error: Cannot find module '@prettier/plugin-php'

npm init --yes
npm install prettier @prettier/plugin-php
prettier --plugin=@prettier/plugin-php --parser=php test.php
# works
```
It seems that `read-pkg-up` is just looking for the closest `package.json` file, which doesn't work for global installs 
https://github.com/prettier/prettier/blob/1b61b819d7f58c645c97a89b26e86915cad3c1f1/src/common/load-plugins.js#L4

(in case you're wondering why i'm explicitly passing the plugin: There also seems to be an issue with automatic discovery, running the command without results in `SyntaxError: Unexpected token` (see https://github.com/prettier/plugin-php/issues/55))

ikatyang:
One solution is to set [`cwd`](https://github.com/sindresorhus/read-pkg-up#cwd) to its own directory instead of `process.cwd()`, but I'm not sure if there's a side effect.

https://github.com/prettier/prettier/blob/1b61b819d7f58c645c97a89b26e86915cad3c1f1/src/common/load-plugins.js#L22-L24

kachkaev:
@ikatyang looks like you're right – changing `cwd` helps indeed. I implemented a [quick patch with this change](https://github.com/kachkaev/prettier/commit/a5910f886d96cdaa748d2708da78de5a7d643d02) to check your theory and was able to get prettier working with locally installed modules inside Atom and VScode plugins (https://github.com/prettier/prettier-atom/issues/395 and https://github.com/prettier/prettier-vscode/issues/395).

Searching for packages in global modules is still not working because there is no package.json in `/usr/local/lib/` (where my global `node_modules` are). It feels like if Prettiers searches for `prettier-plugin-X`/`@prettier/plugin-x` _folders_ in `node_modules` rather than checking `package.json`, we can get a more universal solution, which can work for both local and global setups.

When `basedir` in my patch is not defined, `cwd` is `process.cwd()`, which equals to `/path/to/current/folder` for global prettier being executed from the command line. For some reason, it is equal to `/` inside Atom plugin.

[...]

PR #4341, opened 2018-04-18
Implement prettier.getFileInfo() method and --file-info CLI option

### LINKED ISSUE #4253: Feature request: prettier.getFileInfo(path) API and CLI option
We were [discussing Prettier plugin support in Atom](https://github.com/prettier/prettier-atom/issues/395) with @robwise and [faced](https://github.com/prettier/prettier-atom/issues/395#issuecomment-378357499) the default parser problem.

Until Prettier stops parsing unsupported files as suggested in https://github.com/prettier/prettier/issues/2884, its editor extensions need to be fed with a whitelist of supported scopes. Without this list, Prettier would throw an exception when a user presses `cmd+s` for an unsupported file type or _silently_ fail on real parse errors, which is also undesired. Because #2884 cannot be resolved until Prettier 2.0 and this might take a while, is there any room for a new API call like `prettier.isFileSupported(file)` and a corresponding CLI option? This feature could be used by the editor extensions before applying format on safe, and as a consequence, save a user from maintaining a whitelist of scopes.

How do you see the exterior of this feature look like? I can help with a PR if it has a potential to land to Prettier < 2.0.

kachkaev:
Closes #4253

* [x] get design approval from @robwise and @suchipi (participants in #4253)
* [x] ensure existing tests don't break
* [x] add tests
* [x] see how it works in https://github.com/prettier/prettier-atom/pull/404

---

This PR adds the following prettier method:

```js
const info = prettier.getFileInfo(filepath, opts);
// opts = {
//   ignorePath?: string,
//   withNodeModules?: boolean,
//   plugins: string[]
// }
```

and a corresponding CLI option:

```
prettier.js --help file-info
--file-info <path>

  Extract the following info (as JSON) for a given file path. Reported fields:
  * ignored (boolean) - true if file path is filtered by --ignore-path
  * inferredParser (string | null) - name of parser inferred from file path
```

The new API method is designed to help text editors decide whether a file needs to be formatted on save. A call to it is vital given the introduction of plugins and thus no fixed mapping between file extensions and parsers. A file with the given path does not have to exist for the method to be called.

## Usage examples

`prettier --file-info README.md`
```json
{ "ignored": false, "inferredParser": "markdown" }
```
* Expecting to use a `markdown` parser, according to the file extension
* No signs of `README.md` in `.prettierignore` → `ignored: false`

---

`prettier --file-info README`
```json
{ "ignored": false, "inferredParser": "markdown" }
```
* Same as above despite the lack of extension, because `README` is a [special case](https://github.com/prettier/prettier/blob/0c1d62079e31a045dc4b6b16734a65c9fbf3f3b6/src/language-markdown/index.js#L31)

---

`prettier --file-info src/something.js`
```json
{ "ignored": false, "inferredParser": "babylon" }
```
* Choosing `babylon` parser for a js file – usual thing

---

`prettier --file-info src/something.elm`
```json
{ "ignored": false, "inferredParser": null }
```
* Prettier does not know what to do with `.elm` files by default, so can't infer a parser

---

`yarn add prettier-plugin-elm`
`prettier --file-info src/something.elm`
```json
{ "ignored": false, "inferredParser": "elm" }
```
* Installing a corresponding plugin ([`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) in this case) helps `--file-info` pick the right parser

---

`prettier --file-info src/something.js --ignore-path .ignoresrc`
```json
{ "ignored": true, "inferredParser": "babylon" }
```

* Assuming that file `.ignoresrc` contains `src`, the result contains `ignored: true`

---

`prettier --file-info node_modules/module/something.js`
```json
{ "ignored": true, "inferredParser": "babylon" }
```
* Stuff in `node_modules` is ignored by default

---

`prettier --file-info node_modules/module/something.js --with-node-modules`
```json
{ "ignored": false, "inferredParser": "babylon" }
```

* An already existing CLI option `--with-node-modules` makes the file inside `node_modules` not `ignored`

<details>
<summary>outdated info (changed after further discussion)</summary>

`prettier --file-info README.md`
```json
{ "exists": true, "ignored": false, "inferredParser": "markdown" }
```
* Expecting to use a `markdown` parser, according to the file extension
* No signs of `README.md` in `.prettierignore` → `ignored: false`
* FIle exists (let's assume so)

---

`prettier --file-info README`
```json
{ "exists": false, "ignored": false, "inferredParser": "markdown" }
```
* Same as above despite the lack of extension, because `README` is a [special case](https://github.com/prettier/prettier/blob/0c1d62079e31a045dc4b6b16734a65c9fbf3f3b6/src/language-markdown/index.js#L31)
* File does not exist (again, let's assume so)

---

`prettier --file-info src/something.js`
```json
{ "exists": true, "ignored": false, "inferredParser": "babylon" }
```
* Choosing `babylon` parser for a js file – usual thing

---

`prettier --file-info src/something.elm`
```json
{ "exists": true, "ignored": false, "inferredParser": null }
```
* Prettier does not know what to do with `.elm` files by default, so can't infer a parser

---

`yarn add prettier-plugin-elm`
`prettier --file-info src/something.elm`
```json
{ "exists": true, "ignored": false, "inferredParser": "elm" }
```
* Installing a corresponding plugin ([`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) in this case) helps `--file-info` pick the right parser

---

`prettier --file-info src/something.js --ignore-path .ignoresrc`
```json
{ "exists": true, "ignored": true, "inferredParser": "babylon" }
```

* Assuming that file `.ignoresrc` contains `src`, the result contains `ignored: true`

---

`prettier --file-info node_modules/module/something.js`
```json
{ "exists": false, "ignored": true, "inferredParser": "babylon" }
```
* Stuff in `node_modules` is ignored by default

---

`prettier --file-info node_modules/module/something.js --with-node-modules`
```json
{ "exists": false, "ignored": false, "inferredParser": "babylon" }
```

* An already existing CLI option `--with-node-modules` makes the file inside `node_modules` not `ignored`
</details>

<details>
<summary>outdated info-2 (changed after further discussion)</summary>

`prettier --file-info README.md`
```json
{ "exists": true, "ignored": false, "inferredParser": "markdown" }
```
* Expecting to use a `markdown` parser, according to the file extension
* No signs of `README.md` in `.prettierignore` → `ignored: false`
* FIle exists (let's assume so)

---

`prettier --file-info README`
```json
{ "exists": false, "ignored": false, "inferredParser": "markdown" }
```
* Same as above despite the lack of extension, because `README` is a [special case](https://github.com/prettier/prettier/blob/0c1d62079e31a045dc4b6b16734a65c9fbf3f3b6/src/language-markdown/index.js#L31)
* File does not exist (again, let's assume so)

---

`prettier --file-info src/something.js`
```json
{ "exists": true, "ignored": false, "inferredParser": "babylon" }
```
* Choosing `babylon` parser for a js file – usual thing

---

`prettier --file-info src/something.elm`
```json
{ "exists": true, "ignored": false, "inferredParser": null }
```
* Prettier does not know what to do with `.elm` files by default, so can't infer a parser

---

`yarn add prettier-plugin-elm`
`prettier --file-info src/something.elm`
```json
{ "exists": true, "ignored": false, "inferredParser": "elm" }
```
* Installing a corresponding plugin ([`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) in this case) helps `--file-info` pick the right parser

---

`prettier --file-info src/something.js --ignore-path .ignoresrc`
```json
{ "exists": true, "ignored": true, "inferredParser": "babylon" }
```

* Assuming that file `.ignoresrc` contains `src`, the result contains `ignored: true`

---

`prettier --file-info node_modules/module/something.js`
```json
{ "exists": false, "ignored": true, "inferredParser": "babylon" }
```
* Stuff in `node_modules` is ignored by default

---

`prettier --file-info node_modules/module/something.js --with-node-modules`
```json
{ "exists": false, "ignored": false, "inferredParser": "babylon" }
```

* An already existing CLI option `--with-node-modules` makes the file inside `node_modules` not `ignored`

In the simplest scenario, should be enough to just call `if (prettier.getFileInfo(filePath).formattable) {...}` as @robwise [suggests](https://github.com/prettier/prettier/issues/4253#issuecomment-379582875). Additional fields in the output may be useful for ‘forced format on save’, status display in  UI and other cases. JSON output also makes this method expandable in future.

## Usage examples

`prettier --file-info README.md`
```json
{
  "formattable": true,
  "ignored": false,
  "parser": "markdown",
  "parserIsFallback": false
}
```
* Expecting to use a `markdown` parser, according to the file extension
* No signs of `README.md` in `.prettierignore` → `ignored: false`
* Conclusion: the file is `formattable`

---

`prettier --file-info README`
```json
{
  "formattable": false,
  "ignored": false,
  "parser": "markdown",
  "parserIsFallback": false
}
```
* Same as above despite the lack of extension, because `README` is a [special case](https://github.com/prettier/prettier/blob/0c1d62079e31a045dc4b6b16734a65c9fbf3f3b6/src/language-markdown/index.js#L31)

---

`prettier --file-info src/something.js`
```json
{
  "formattable": true,
  "ignored": false,
  "parser": "babylon",
  "parserIsFallback": false
}
```
* Choosing `babylon` parser for a js file – usual thing

---

`prettier --file-info src/something.elm`
```json
{
  "formattable": false,
  "ignored": false,
  "parser": "babylon",
  "parserIsFallback": true
}
```
* Prettier does not know what to do with `.elm` files by default, so picks `babylon` parser as a fallback one
* When a fallback parser is used, the file is considered as non-`formattable` (editors should not attempt to format it on save as this will produce an error anyway)

---

`yarn add prettier-plugin-elm`
`prettier --file-info src/something.elm`
```json
{
  "formattable": true,
  "ignored": false,
  "parser": "elm",
  "parserIsFallback": false
}
```
* Installing a corresponding plugin ([`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) in this case) helps `--file-info` pick the right parser and mark the file as `formattable`

---

`prettier --file-info src/something.js --ignore-path .ignoresrc`
```json
{
  "formattable": false,
  "ignored": true,
  "parser": "babylon",
  "parserIsFallback": false
}
```
* Assuming that file `.ignoresrc` contains `src`, the result contains `ignored: true`
* `ignored: true` implies `formattable: false`

---

`prettier --file-info node_modules/module/something.js`
```json
{
  "formattable": false,
  "ignored": true,
  "parser": "babylon",
  "parserIsFallback": false
}
```
* Stuff in `node_modules` is ignored by default and hence not `formattable`

---

`prettier --file-info node_modules/module/something.js --with-node-modules`
```json
{
  "formattable": true,
  "ignored": false,
  "parser": "babylon",
  "parserIsFallback": false
}
```
* An already existing CLI option `--with-node-modules` makes the file inside `node_modules` not `ignored` and thus `formattable`
</details>

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ const version = require("./package.json").version;
 
 const core = require("./src/main/core");
 const getSupportInfo = require("./src/main/support").getSupportInfo;
+const getFileInfo = require("./src/common/get-file-info");
 const sharedUtil = require("./src/common/util-shared");
 const loadPlugins = require("./src/common/load-plugins");
 
@@ -12,7 +13,7 @@ const config = require("./src/config/resolve-config");
 const doc = require("./src/doc");
 
 // Luckily `opts` is always the 2nd argument
-function withPlugins(fn) {
+function _withPlugins(fn) {
   return function() {
     const args = Array.from(arguments);
     const opts = args[1] || {};
@@ -21,6 +22,13 @@ function withPlugins(fn) {
     });
     return fn.apply(null, args);
   };
+}
+function withPlugins(fn) {
+  const resultingFn = _withPlugins(fn);
+  if (fn.sync) {
+    resultingFn.sync = _withPlugins(fn.sync);
+  }
+  return resultingFn;
 }
 
 const formatWithCursor = withPlugins(core.formatWithCursor);
@@ -47,6 +55,7 @@ module.exports = {
   resolveConfigFile: config.resolveConfigFile,
   clearConfigCache: config.clearCache,
 
+  getFileInfo: withPlugins(getFileInfo),
   getSupportInfo: withPlugins(getSupportInfo),
 
   version,
--- a/index.js
+++ b/index.js
@@ -23,6 +23,7 @@ function _withPlugins(fn) {
     return fn.apply(null, args);
   };
 }
+
 function withPlugins(fn) {
   const resultingFn = _withPlugins(fn);
   if (fn.sync) {
--- a/src/common/get-file-info.js
+++ b/src/common/get-file-info.js
@@ -1,0 +1,40 @@
+"use strict";
+
+const createIgnorer = require("../common/create-ignorer");
+const options = require("../main/options");
+
+/**
+ * @param {string} filePath
+ * @param {{ ignorePath?: string, withNodeModules?: boolean, plugins: object }} opts
+ *
+ * Please note that prettier.getFileInfo() expects opts.plugins to be an array of paths,
+ * not an object. A transformation from this array to an object is automatically done
+ * internally by the method wrapper. See withPlugins() in index.js.
+ */
+function _getFileInfo(filePath, opts) {
+  let ignored = false;
+  const ignorer = createIgnorer(opts.ignorePath, opts.withNodeModules);
+  ignored = ignorer.ignores(filePath);
+
+  const inferredParser = options.inferParser(filePath, opts.plugins) || null;
+
+  return {
+    ignored,
+    inferredParser
+  };
+}
+
+// the method has been implemented as asynchronous to avoid possible breaking changes in future
+function getFileInfo(filePath, opts) {
+  return new Promise((resolve, reject) => {
+    try {
+      resolve(_getFileInfo(filePath, opts));
+    } catch (e) {
+      reject(e);
+    }
+  });
+}
+
+getFileInfo.sync = _getFileInfo;
+
+module.exports = getFileInfo;
--- a/src/common/get-file-info.js
+++ b/src/common/get-file-info.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const createIgnorer = require("../common/create-ignorer");
+const createIgnorer = require("./create-ignorer");
 const options = require("../main/options");
 
 /**
--- a/src/main/options.js
+++ b/src/main/options.js
@@ -118,7 +118,7 @@ function inferParser(filepath, plugins) {
   }).languages.find(
     language =>
       language.since !== null &&
-      (language.extensions.indexOf(extension) > -1 ||
+      ((language.extensions && language.extensions.indexOf(extension) > -1) ||
         (language.filenames &&
           language.filenames.find(name => name.toLowerCase() === filename)))
   );
@@ -126,4 +126,4 @@ function inferParser(filepath, plugins) {
   return language && language.parsers[0];
 }
 
-module.exports = { normalize, hiddenDefaults };
+module.exports = { normalize, hiddenDefaults, inferParser };
--- a/docs/api.md
+++ b/docs/api.md
@@ -72,6 +72,23 @@ If `options.useCache` is `false`, all caching will be bypassed.
 
 As you repeatedly call `resolveConfig`, the file system structure will be cached for performance. This function will clear the cache. Generally this is only needed for editor integrations that know that the file system has changed since the last format took place.
 
+## `prettier.getFileInfo(filePath [, options])`
+
+`getFileInfo` can be used by editor extensions to decide if a particular file needs to be formatted. This method returns a promise, which resolves to an object with the following properties:
+
+```typescript
+{
+  ignored: boolean,
+  inferredParser: string | null,
+}
+```
+
+Setting `options.ignorePath` (`string`) and `options.withNodeModules` (`boolean`) influence the value of `ignored` (`false` by default).
+
+Providing [plugin](./plugins.md) paths in `options.plugins` (`string[]`) helps extract `inferredParser` for files that are not supported by Prettier core.
+
+Use `prettier.getFileInfo.sync(filePath [, options])` if you'd like to use sync version.
+
 ## `prettier.getSupportInfo([version])`
 
 Returns an object representing the parsers, languages and file types Prettier supports.
--- a/src/cli/constant.js
+++ b/src/cli/constant.js
@@ -133,7 +133,6 @@ const options = {
   },
   "file-info": {
     type: "path",
-    category: coreOptions.CATEGORY_CONFIG,
     description: dedent`
       Extract the following info (as JSON) for a given file path. Reported fields:
       * ignored (boolean) - true if file path is filtered by --ignore-path
--- a/src/cli/constant.js
+++ b/src/cli/constant.js
@@ -131,6 +131,15 @@ const options = {
     description:
       "Find and print the path to a configuration file for the given input file."
   },
+  "file-info": {
+    type: "path",
+    category: coreOptions.CATEGORY_CONFIG,
+    description: dedent`
+      Extract the following info (as JSON) for a given file path. Reported fields:
+      * ignored (boolean) - true if file path is filtered by --ignore-path
+      * inferredParser (string | null) - name of parser inferred from file path
+    `
+  },
   help: {
     type: "flag",
     alias: "h",
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -22,6 +22,11 @@ function run(args) {
       process.exit(1);
     }
 
+    if (context.argv["file-info"] && context.filePatterns.length) {
+      context.logger.error("Cannot use --file-info with multiple files");
+      process.exit(1);
+    }
+
     if (context.argv["version"]) {
       context.logger.log(prettier.version);
       process.exit(0);
@@ -50,10 +55,9 @@ function run(args) {
       context.argv["stdin"] || (!hasFilePatterns && !process.stdin.isTTY);
 
     if (context.argv["find-config-path"]) {
-      util.logResolvedConfigPathOrDie(
-        context,
-        context.argv["find-config-path"]
-      );
+      util.logResolvedConfigPathOrDie(context);
+    } else if (context.argv["file-info"]) {
+      util.logFileInfoOrDie(context);
     } else if (useStdin) {
       util.formatStdin(context);
     } else if (hasFilePatterns) {
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -92,7 +92,8 @@ function logFileInfoOrDie(context) {
   const options = {
     ignorePath: context.argv["ignore-path"],
     withNodeModules: context.argv["with-node-modules"],
-    plugins: context.argv["plugin"]
+    plugins: context.argv["plugin"],
+    pluginSearchDirs: context.argv["plugin-search-dir"]
   };
   context.logger.log(
     prettier.format(
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -5,13 +5,14 @@ const camelCase = require("camelcase");
 const dashify = require("dashify");
 const fs = require("fs");
 const globby = require("globby");
-const ignore = require("ignore");
 const chalk = require("chalk");
 const readline = require("readline");
 const leven = require("leven");
+const stringify = require("json-stable-stringify");
 
 const minimist = require("./minimist");
 const prettier = require("../../index");
+const createIgnorer = require("../common/create-ignorer");
 const errors = require("../common/errors");
 const constant = require("./constant");
 const coreOptions = require("../main/core-options");
@@ -76,13 +77,31 @@ function handleError(context, filename, error) {
   process.exitCode = 2;
 }
 
-function logResolvedConfigPathOrDie(context, filePath) {
-  const configFile = prettier.resolveConfigFile.sync(filePath);
+function logResolvedConfigPathOrDie(context) {
+  const configFile = prettier.resolveConfigFile.sync(
+    context.argv["find-config-path"]
+  );
   if (configFile) {
     context.logger.log(path.relative(process.cwd(), configFile));
   } else {
     process.exit(1);
   }
+}
+
+function logFileInfoOrDie(context) {
+  const options = {
+    ignorePath: context.argv["ignore-path"],
+    withNodeModules: context.argv["with-node-modules"],
+    plugins: context.argv["plugin"]
+  };
+  context.logger.log(
+    prettier.format(
+      stringify(prettier.getFileInfo.sync(context.argv["file-info"], options)),
+      {
+        parser: "json"
+      }
+    )
+  );
 }
 
 function writeOutput(result, options) {
@@ -255,7 +274,7 @@ function formatStdin(context) {
     ? path.resolve(process.cwd(), context.argv["stdin-filepath"])
     : process.cwd();
 
-  const ignorer = createIgnorer(context);
+  const ignorer = createIgnorerFromContextOrDie(context);
   const relativeFilepath = path.relative(process.cwd(), filepath);
 
   thirdParty.getStream(process.stdin).then(input => {
@@ -278,22 +297,16 @@ function formatStdin(context) {
   });
 }
 
-function createIgnorer(context) {
-  const ignoreFilePath = path.resolve(context.argv["ignore-path"]);
-  let ignoreText = "";
-
+function createIgnorerFromContextOrDie(context) {
   try {
-    ignoreText = fs.readFileSync(ignoreFilePath, "utf8");
-  } catch (readError) {
-    if (readError.code !== "ENOENT") {
-      context.logger.error(
-        `Unable to read ${ignoreFilePath}: ` + readError.message
-      );
-      process.exit(2);
-    }
+    return createIgnorer(
+      context.argv["ignore-path"],
+      context.argv["with-node-modules"]
+    );
+  } catch (e) {
+    context.logger.error(e.message);
+    process.exit(2);
   }
-
-  return ignore().add(ignoreText);
 }
 
 function eachFilename(context, patterns, callback) {
@@ -329,7 +342,7 @@ function eachFilename(context, patterns, callback) {
 function formatFiles(context) {
   // The ignorer will be used to filter file paths after the glob is checked,
   // before any files are actually written
-  const ignorer = createIgnorer(context);
+  const ignorer = createIgnorerFromContextOrDie(context);
 
   eachFilename(context, context.filePatterns, (filename, options) => {
     const fileIgnored = ignorer.filter([filename]).length === 0;
@@ -910,5 +923,6 @@ module.exports = {
   formatStdin,
   initContext,
   logResolvedConfigPathOrDie,
+  logFileInfoOrDie,
   normalizeDetailedOptionMap
 };
--- a/src/common/create-ignorer.js
+++ b/src/common/create-ignorer.js
@@ -1,0 +1,30 @@
+"use strict";
+
+const ignore = require("ignore");
+const fs = require("fs");
+const path = require("path");
+
+function createIgnorer(ignorePath, withNodeModules) {
+  let ignoreText = "";
+
+  if (ignorePath) {
+    const resolvedIgnorePath = path.resolve(ignorePath);
+    try {
+      ignoreText = fs.readFileSync(resolvedIgnorePath, "utf8");
+    } catch (readError) {
+      if (readError.code !== "ENOENT") {
+        throw new Error(
+          `Unable to read ${resolvedIgnorePath}: ${readError.message}`
+        );
+      }
+    }
+  }
+
+  const ignorer = ignore().add(ignoreText);
+  if (!withNodeModules) {
+    ignorer.add("node_modules");
+  }
+  return ignorer;
+}
+
+module.exports = createIgnorer;
--- a/tests_integration/__tests__/__snapshots__/early-exit.js.snap
+++ b/tests_integration/__tests__/__snapshots__/early-exit.js.snap
@@ -102,6 +102,19 @@ Default: true
 
 exports[`show detailed usage with --help editorconfig (write) 1`] = `Array []`;
 
+exports[`show detailed usage with --help file-info (stderr) 1`] = `""`;
+
+exports[`show detailed usage with --help file-info (stdout) 1`] = `
+"--file-info <path>
+
+  Extract the following info (as JSON) for a given file path. Reported fields:
+  * ignored (boolean) - true if file path is filtered by --ignore-path
+  * inferredParser (string | null) - name of parser inferred from file path
+"
+`;
+
+exports[`show detailed usage with --help file-info (write) 1`] = `Array []`;
+
 exports[`show detailed usage with --help find-config-path (stderr) 1`] = `""`;
 
 exports[`show detailed usage with --help find-config-path (stdout) 1`] = `
@@ -608,6 +621,9 @@ Config options:
                            Define in which order config files and CLI options should be evaluated.
                            Defaults to cli-override.
   --no-editorconfig        Don't take .editorconfig into account when parsing configuration.
+  --file-info <path>       Extract the following info (as JSON) for a given file path. Reported fields:
+                           * ignored (boolean) - true if file path is filtered by --ignore-path
+                           * inferredParser (string | null) - name of parser inferred from file path
   --find-config-path <path>
                            Find and print the path to a configuration file for the given input file.
   --ignore-path <path>     Path to a file with patterns describing files to ignore.
@@ -756,6 +772,9 @@ Config options:
                            Define in which order config files and CLI options should be evaluated.
                            Defaults to cli-override.
   --no-editorconfig        Don't take .editorconfig into account when parsing configuration.
+  --file-info <path>       Extract the following info (as JSON) for a given file path. Reported fields:
+                           * ignored (boolean) - true if file path is filtered by --ignore-path
+                           * inferredParser (string | null) - name of parser inferred from file path
   --find-config-path <path>
                            Find and print the path to a configuration file for the given input file.
   --ignore-path <path>     Path to a file with patterns describing files to ignore.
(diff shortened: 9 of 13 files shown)


kachkaev:
🤔 should we also have `exists` flag in the output? It makes sense to claim that a file is `formattable` only if it exists really/

robwise:
> should we also have exists flag in the output?

I don't have an opinion on this one way or the other.

> Also the logic behind formattable gets a bit too complicated, don't you think?

Well, this logic needs to exist *somewhere*, right? It's better to do it once in the main repo than reproduced in all of the editor implementations. Or maybe I'm misunderstanding you?


I've looked over each of the scenarios you've documented, everything looks good to me!

kachkaev:
Re `formattable`:

> It's better to do it once in the main repo than reproduced in all of the editor implementations. Or maybe I'm misunderstanding you?

After a bit of thinking, I guess that moving this logic to the editor implementation is a better thing. The main reason is that there are two ways of formatting a file:
* from stdin / text argument + file name as option to determine parser
* from CLI for a real file with `--write`

In the first approach, it's OK to set `formattable` to true when a file does not exist – no harm at all. However, in the second scenario, just checking `formattable` for a path and then calling `prettier --write` would throw `ENOENT`. Now if we add `exists` flag and only trigger `formattable` if it's true, the first scenario breaks. So depending on what strategy a tool like the Atom Package chooses, `formattable` will either work or not – it cannot be tuned so that it fits everyone.

What is also a bit confusing is that this flag may look over-promising. Someone may see that `formattable: true` may mean that prettier has already looked and the file syntax, successfully parsed it and has some changes to suggest. At least there is a chance that someone might interpret this flag that way and come back with an issue.

I understand that implementing the logic of `formattable` within `prettier-atom` makes the `if` a lot bigger, but on the other hand, it removes some questions and does not lock everyone into a certain opinionated flavour of this flag. With `formattable` removed and `exists: true/false` added, a consumer of `--file-info` will be free to use whatever formatting approach it finds best. Both stdin and `prettier --write` methods will be first-class choices.

WDYT?

kachkaev:
@robwise I can provide a prebuilt copy of prettier that you can test in https://github.com/prettier/prettier-atom/pull/404. The reason why a prebuilt version is better is because it does not have `eval()` in plugin-related code, which Atom does not like ([details](https://github.com/prettier/prettier/issues/4000#issuecomment-376119478)).

Would you like me to change the result of `getFileInfo()` to the following before that?

```json5
{
  "exists": true,
  "ignored": false,
  "inferredParser": "markdown", // or null, even in v1.x
}
```

robwise:
Oh just saw the previous comment.

> In the first approach, it's OK to set formattable to true when a file does not exist – no harm at all. However, in the second scenario, just checking formattable for a path and then calling prettier --write could throw ENOENT (file does not exist). Now if we add exists flag and only trigger formattable if it's true, the first scenario breaks. So depending on what approach a tool like the Atom Package chooses, formattable will either work or not – it cannot be tuned so that it fits everyone.

I'm not sure I understand this one, why would we return `formattable: true` if the file doesn't exist?

kachkaev:
> why would we return formattable: true if the file doesn't exist?

@robwise on first ‘format on save’. That is when the file name has already been given, there is some content in the file, but the file has never been saved yet. Details in https://github.com/prettier/prettier/pull/4341#issuecomment-382638660.

Apart from `formattable` existing or not, are you OK with these three options instead of the original ones?

```json
{
  "exists": true,
  "ignored": false,
  "inferredParser": "markdown", // or null, even in v1.x
}
```

robwise:
> @robwise on first ‘format on save’. That is when the file name has already been given, there is some content to prettify, but the file has never been put on disk yet.

I'm still confused about this point: if the file does not even exist on disk, why would we have Prettier claim it's formattable at that point? How would it know anything about the file at all—at this point, it doesn't exist except as a string of text inside of my editor's text buffer, and Prettier knows nothing about my editor's text buffer?

> IMO that's the way to go – prettier should not decide too much for the editor plugin.

Well, agree to disagree, but I can definitely get the job done with what you've shown in your examples (where you demonstrate combining some of the flags), so 👍from me.

> Apart from formattable existing or not, are you OK with these three options instead of the original ones? I'll update the implementation of so.

Yeah, the main differences being no `formattable` and no `isFallbackParser`? I'm fine with that IIUC.

kachkaev:
@robwise done! I also rebased on top of the latest master and updated the first comment in this PR.

If you'd like to try out `getFileInfo()` in https://github.com/prettier/prettier-atom/pull/404, feel free to `npm install github:kachkaev/prettier#file-info-dist` (with `-dist` in the end). It's a built version of my branch, which matches how prettier looks like on npm.

Waiting to hear your feedback and will proceed to writing tests after that.

kachkaev:
> if the file does not even exist on disk, why would we have Prettier claim it's formattable at that point?

In atom you would not be able to infer a parser while you don't know the file extension, i.e. until you save for the first time, that's true. But what about other editors? Imagine a hypothetical editor that always has a file path input filed on top of a text area with a file contents and you fill in both before you save for the first time.

I like that when we only have `exists`, `ignored` and `inferredParaser`, _all_ these fields are orthogonal (i.e. cannot be derived from each other). This means that the information to data ratio is the highest and there is no redundancy.

If you're still worried about the size of an `if` statement, there's always space for a new npm module that would shorten that for you 🙂 

```js
import { isFormattable } from 'prettier-file-info';

if (isFormattable(prettier.getFileInfo(filePath))) {
  // format on save
}
```

kachkaev:
@suchipi could you please have a look at this PR? If you're happy with its design and implementation, I'll tick the first checkbox and proceed to writing tests. It'd be also great if we could go through https://github.com/prettier/prettier/pull/4192 before 1.13 because that PR is also important for text editors in context of prettier plugins.

robwise:
@kachkaev I'm getting an error:

```
path.js:28
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received undefined
    at assertPath (path.js:28:11)
    at Object.resolve (path.js:1188:7)
    at createIgnorer (/Users/rob/github/prettier-atom/node_modules/prettier/index.js:1743:39)
    at Object.getFileInfo (/Users/rob/github/prettier-atom/node_modules/prettier/index.js:24170:19)
    at Object.<anonymous> (/Users/rob/github/prettier-atom/test.js:8:27)
    at Module._compile (module.js:624:30)
    at Object.Module._extensions..js (module.js:635:10)
    at Module.load (module.js:545:32)
    at tryModuleLoad (module.js:508:12)
    at Function.Module._load (module.js:500:3)
```

```js
// test.js
const prettier = require('prettier')
const path = require('path')

const filePath = path.join(__dirname, 'test.js')
const fileInfo = prettier.getFileInfo(filePath)

console.log('file info:', JSON.stringify(fileInfo, null, 2))

```

kachkaev:
@robwise – sorry for not mentioning what arguments `prettier.getFileInfo()` expects and so causing this exception on your side. The function has this API:

```ts
getFileInfo(
  filePath: string,
  opts: Options,
  ignorePath: string,
  doNotIgnoreNodeModules: boolean
): FileInfo
```

The reason why you were getting an exception was because you did not pass `ignorePath`, so an instance of ignorer could not be created.

This worked for me:

```js
const filePath = path.join(__dirname, "test.js");
const ignorePath = path.join(__dirname, ".prettierignore");
const fileInfo = prettier.getFileInfo(filePath, {}, ignorePath);
```

or just

```js
const fileInfo = prettier.getFileInfo("test.js", {}, '.prettierignore');
```
(paths are resolved based on `cwd`)

I'll make all arguments but the first one optional in the next few days, but you are free to test the method even before the changes. Sorry for not sharing this caveat with you – I was only testing the new functionality as  the `--file-info` flag and did not think about it.

robwise:
No worries.

I was curious about the `Options` as I assumed that they are irrelevant to just checking file info (no actual formatting is done here)?

I was thinking it might be more something like this, that way you aren't forced to provide an `ignorePath` in order to specify `doNotIgnoreNodeModules`:

```js
type getFileInfo = (
  filePath: string, 
  opts?: { configPath?: string, ignorePath?: string, doNotIgnoreNodeModules?: boolean }
) => FileInfo;
```

kachkaev:
Fair point, I’ll check that. I was not sure that mixing the options that are going to be normalized with two additional parameters is a good thing, but if the same is done in other methods, I’m happy to go for the trick.

Apart from this, do things work for you in general? I’m keen to cover the new feature with tests and remove the PR’s WIP status.

robwise:
@kachkaev All seems to work so far, now we're able to get rid of that `scopes` thing entirely :)

duailibe:
@kachkaev `index.js` has been significantly changed in `master`. Can you get the latest from master and update the PR? Thanks!

kachkaev:
@duailibe sure, just give me a few days ;–)

duailibe:
No hurry!

kachkaev:
@duailibe the changes have been rebased on top of the latest master.

@robwise thanks to your suggestion and the new internal `withPlugins()` function, I was able to simplify the API of `getFileInfo()` It looks like this now:

```js
const fileInfo = getFileInfo({
  filepath: 'path/to/file.js',
  ignorePath: 'path/to/.prettierignore', // optional, assumed '.prettierignore' by default, can be set to disable
  withNodeModules: true,  // optional, assumed false by default
  ...otherPrettierOptions, // optional
});
```

When `getFileInfo()` is called from CLI via `--file-info`, `ignorePath` is set to `.prettierignore` by default. I've decided to give this option the same default value in method call for consistency. If the file does not exist in `cwd`, nothing dramatic happens – an instance of ignorer is still created and this `withNodeModules` still works. As far as I understand, an [exception is only thrown](https://github.com/kachkaev/prettier/blob/f08710fc7be56ea298432cfceda9cc8901fab916/src/common/create-ignorer.js#L11-L19) when a file at `ignorePath` cannot be read due to permission issues.

It's unfortunate that `filepath` is not _camelCase_ while `ignorePath` is. This is because Prettier uses `filepath` everywhere else inside `options` and it's probably not a good idea to change this. I'm not naming `ignorePath` `ignorepath` for consistency with `--ignore-path` and other variable names throughout the code (`filepath` is the only exception here).

If everyone is happy with the new design, I'm happy to proceed to tests to remove WIP status from this PR.

suchipi:
@kachkaev sorry it took me so long to look at this, the API looks good to me. I'm not super familiar with the stuff in `src/cli/util.js` so I defer to you and the other maintainers on that part.

kachkaev:
@suchipi great to hear that you're OK with the API (hope you looked at the version in my last commit as I forgot to update the first one and only did this now). Could you please cc someone who could look at `src/cli/util.js`?

suchipi:
Yeah, I looked at your most recent commit.

@ikatyang @azz @duailibe @lydell @j-f1 could one of you take a look at this?

robwise:
@kachkaev if I don't pass an `ignorePath`, it's going to look for `.prettierignore` in the project's root directory, not in the directory of the `filepath`, right?

kachkaev:
@robwise I guess it will be cwd/.prettierignore – anything else would require too much changes throughout Prettier. Does it sound good to you?

src/common/get-file-info.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const inferredParser =
+    options.inferParser(opts.filepath, opts.plugins) || null;
+
+  return {
+    exists,
+    ignored,
+    inferredParser
+  };
+}
```
duailibe: My opinions:

- `getFileInfo(filepath, {ignorePath, withNodeModules, plugins})`
- returns `{ignored, inferredParser}`

kachkaev: > `getFileInfo(filepath, {ignorePath, withNodeModules, plugins})`

Looks a bit simpler to use and does not expose different naming styles for `filepath` and `ignorePath` – I like this! Will change the API again as you suggest. We can also avoid setting `ignorePath` when a method is called directly (not via CLI) – that may work a bit more predictably for Prettier API users. The only problem that remains is with plugins – they will not be found automatically unless `cwd` matches project path (which will be the case for editor plugins). We need https://github.com/prettier/prettier/pull/4192 to solve that.

> Last suggestion: isFormattable that returns true if path is not ignored and a parser can be inferred
see https://github.com/prettier/prettier/pull/4341#issuecomment-383473937 and https://github.com/prettier/prettier/pull/4341#issuecomment-382638660

duailibe: Sorry I wasn't clear. I'm suggesting an alternative API to `getFileInfo` not an additional return value.

What I mean is: most integrations will use this function and use something like:
```js
if (!fileinfo.ignored && fileinfo.parser) { const formatted = prettier.format(text, {options}); }
```

So maybe we could do instead:

```js
if (!prettier.isFileSupported(filepath, someOptions)) return;
const formatted = prettier.format(text, {options});
```

WDYT?

kachkaev: What you’re suggesting when work when using format on save. But when a person calls format manually, I can imagine a scenario when .prettierifnore is ignored. I’d let editors decide what to do in this case.

kachkaev: Hmm in the second scenario we can just avoid passing .prettierignore as an option and this will do the job. Guess you’re right and I should open a competing PR with your suggestion.

duailibe: I don't know if we need separate PRs, they'll probably be very similar, we just need to figure out the API but it's your call!

> What would be an appropriate CLI option name in this case?

Now you got me :joy:

duailibe:
Implementation looks good to me.. Let's just figure out if we need to expose the ignored and inferred parser separately, or a boolean `!ignored && inferredParser` would be enough.

kachkaev:
I'm open to change the implementation to just doing that – that'll look simpler. But I'm wondering what the name of the method and the name of the CLI option would be. After a bit of thinking, I rejected `inferParser` / `--infer-parser` and a few other. `inferParser` would create a confusion with an already existing `inferParser` that does not take ignore config into account.  Could it be `getFileSupport()` / `--file-support`? Not a big fan of the name, but it could be an option.

Apart from that, I quite like the idea of just getting true/false from a method or in stdout.

robwise:
> @robwise I guess it will be $(cwd)/.prettierignore – anything else would require too much changes throughout Prettier. Does it sound good to you?

I think it should reuse this logic instead of requiring the editor to do it: https://github.com/prettier/prettier/blob/master/src/cli/util.js#L281-L297 

This way:

- the implementation of `ignored` will explicitly be kept in sync with how the CLI actually ignores
- each editor won't have to re-implement this (or else the `ignored` option is going to have a lot of false negatives)

I think it would just require moving one or two functions from CLI utils to a more general utils file, purely to keep things organized.

suchipi:
Let's just expose them separately and call it file info, there could be value in editor integrations knowing which parser is selected (like checking if a plugin is being loaded correctly and showing the user a help message). Also, we might want to add more stuff to the return value later, which we can do with an object but not with a boolean.

kachkaev:
@suchipi I added tests and an entry to `api.md`, so the PR should be ready for your review.

There's still one thing I'm not sure about: should we make `getFileInfo()` asynchronous? It is sync by nature now, so returning a promise would be surplus. However, if we assume that `getFileInfo()` may support other output in future, going from sync to async would be come a breaking change. Should we  future-proof or is this an overkill?

kachkaev:
@suchipi thanks for adding this to 1.13! Do you mind leaving your review for this PR so that I could make the final changes if needed? You took part in the original discussion in #4253, so it'd be useful to know what you think in the end.

suchipi:
Yeah- I will plan on looking through this thoroughly tomorrow evening and leaving a review. I will also test it against a couple use-cases locally.

duailibe:
@kachkaev do you think it'd be too much trouble and make this async and add a `getFileInfo.sync`?

suchipi:
@kachkaev my bad, I thought this was your other PR :sweat_smile: I can review this one now, it's simpler

src/main/options.js
```diff
@@ -117,13 +117,12 @@ function inferParser(filepath, plugins) {
     plugins
   }).languages.find(
     language =>
-      language.since !== null &&
-      (language.extensions.indexOf(extension) > -1 ||
-        (language.filenames &&
-          language.filenames.find(name => name.toLowerCase() === filename)))
```
suchipi: This `language.since` check is present to prevent the HTML and Handlebars parsers from being inferred, so we need to keep it in. Did you need to remove this to support a third-party plugin? I don't think third-party plugins need to ever define `since`, so you could take it out of your plugin (since this is looking for null specifically). Alternatively, we could add an `unreleased` property to the HTML and Handlebars languages, and check for that instead. I think that would be more clear.

kachkaev:
@duailibe I don't think making `getFileInfo()` async would be a big pain unless I struggle to tweak early exit in CLI. Will try now.

src/main/options.js
kachkaev: Many thanks for spotting this @suchipi! I guess `language.since !== null` disappeared just by my stupid mistake when I was guarding `language.extensions.indexOf`. There was a null pointer exception somewhere without this; all fixed now.

I did not run the tests locally because I'm looking into async `getFileInfo()` now – hope they pass in CI 😅

kachkaev:
@duailibe @suchipi  `getFileInfo()` is async now – feel free to review again 😉 

The tests pass, but Travis builds struggle a bit. Here's the original one and the one I re-spawned by tweaking the last commit and then undoing it:
https://travis-ci.org/prettier/prettier/builds/376719918
https://travis-ci.org/prettier/prettier/builds/376725574

It'd be great if someone with enough permissions could poke failed jobs in the last build – they fail for some internal reason.

lipis:
Node 4 / Prod timed out.. :/

kachkaev:
@lipis in https://travis-ci.org/prettier/prettier/builds/376725574 – yes, while in https://travis-ci.org/prettier/prettier/builds/376719918 it went well. Do you see some button that would restart https://travis-ci.org/prettier/prettier/jobs/376725575? If not, I'll have to force-push again until all for jobs succeed.

lipis:
There was button indeed, to restart only that job. Let's see :) 🕥

lipis:
All green!

kachkaev:
Squashed and rebased on top of the latest master that contains #4192 (had to squash because placing intermediate commits on top of master was creating conflicts). Old commits are still available in [branch `file-info-before-rebase-03` of my fork](https://github.com/kachkaev/prettier/tree/file-info-before-rebase-03) (for history).

I'll sortly add a test case for `--plugin-search-dir` as suggested previously.

lipis:
Everything will be squashed anyways.. so no worries about the history! It's mostly for you

index.js
```diff
@@ -22,6 +23,13 @@ function withPlugins(fn) {
     return fn.apply(null, args);
   };
 }
```
lipis: nit: Empty line between functions :P

kachkaev: 👍 😅

lipis: Like in the good old days! :)

kachkaev:
All done 🎉  I guess th PR is ready to be merged after someone checks out my last two commits.

src/common/get-file-info.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
+
+const createIgnorer = require("../common/create-ignorer");
```
ikatyang:
```js
const createIgnorer = require("./create-ignorer");
```

lipis:
the second to last is OK :P

src/common/get-file-info.js
```diff
@@ -0,0 +1,40 @@
...
+  const inferredParser = options.inferParser(filePath, opts.plugins) || null;
+
+  return {
+    ignored,
+    inferredParser
+  };
+}
+
+// the method has been implemented as asynchronous to avoid possible breaking changes in future
+function getFileInfo(filePath, opts) {
```
ikatyang: This async method looks just like

```js
(...args) => Promise.resolve(_getFileInfo(...args))
```

We need to use `createIgnorer`/`createIgnorer.sync` if we'd like to make it really async.

ikatyang: It's just a quick demo, not meant to be used in node 4. 😂

> does this have to be done now?

Yeah, it can be another PR, just a reminder.

kachkaev: Changing the function to the following does not work:

```js
function getFileInfo(filePath, opts) {
  return Promise.resolve(_getFileInfo(filePath, opts));
}
```

It throws rather than rejects, because `_getFileInfo()` is sync.

ikatyang: Opened #4452 to track this issue.

src/common/get-file-info.js
```diff
@@ -0,0 +1,40 @@
...
+  return new Promise((resolve, reject) => {
+    try {
+      resolve(_getFileInfo(filePath, opts));
+    } catch (e) {
+      reject(e);
+    }
+  });
```
j-f1: This could just be `Promise.resolve().then(() => _getFileInfo(filePath, options))`.

kachkaev: I'm happy to include this to https://github.com/prettier/prettier/pull/4451, which has a couple of minor enhancements. See commit log.

robwise:
@kachkaev Why was this made async? I can't use it in prettier-atom :(

duailibe:
@robwise There's a sync and an async method.

robwise:
@duailibe ah excellent, that makes things much easier!

PR #4361, opened 2018-04-24
New: Format Angular Component styles

JamesHenry:
I had a free couple of hours on the flight back from ng-conf and so put together a POC for Angular Component inline template and style formatting. Would love feedback! More than a couple of folks at the conference asked me if this was something we could add to Prettier.

I can see from recent comments on issues that HTML support is not classed as stable yet? If this has to go in in two stages (first inline styles, then inline template once HTML is no longer experimental) then that would still be great.

The implementation is a little bit hacky as I did not have a WIFI connection and just did my best to emulate existing patterns from Vue formatting etc. so happy to make any updates you want me to. 

Example test case:

**UPDATED TO REMOVE HTML SUPPORT**

**Before:**
```ts
@Component({
       selector: 'app-test',
  styles: [   `
  
 :host {
   color: red;
 } 
 div { background: blue
 }
`

]
})
class     TestComponent {}
```

**After:**
```ts
@Component({
  selector: "app-test",
  styles: [
    `
      :host {
        color: red;
      }
      div {
        background: blue;
      }
    `,
  ],
})
class TestComponent {}
```

### DIFF
--- a/src/language-js/clean.js
+++ b/src/language-js/clean.js
@@ -122,6 +122,24 @@ function clean(ast, newObj, parent) {
     newObj.value.expression.quasis.forEach(q => delete q.value);
   }
 
+  // CSS template literals in Angular Component decorator
+  if (
+    ast.type === "Decorator" &&
+    ast.expression.type === "CallExpression" &&
+    ast.expression.callee.name === "Component" &&
+    ast.expression.arguments.length === 1 &&
+    ast.expression.arguments[0].properties.some(
+      prop =>
+        prop.key.name === "styles" && prop.value.type === "ArrayExpression"
+    )
+  ) {
+    newObj.expression.arguments[0].properties.forEach(prop => {
+      if (prop.value.type === "ArrayExpression") {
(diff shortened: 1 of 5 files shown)


suchipi:
The HTML formatter isn't ready for prime time yet, so let's do inline styles first

JamesHenry:
Ok thanks @suchipi! I'll clean up this PR and request another review once it only applies changes to styles

JamesHenry:
Updated! @suchipi Do you have any idea how I can work around the AST_COMPARE=1 failure on the whitespace in my example? I can reproduce locally, but I don't entirely understand the failure...

duailibe:
@JamesHenry you need to remove those nodes from compare, somewhere in: https://github.com/prettier/prettier/blob/bcdb6936ba42658b8dcdf1f70b4e5e0c8ae8ec06/src/language-js/clean.js#L115-L123

JamesHenry:
Thanks for the review, @duailibe! @suchipi Did you have any other thoughts/concerns?

suchipi:
Looks great, thanks for your contribution!

PR #4364, opened 2018-04-24
Refactoring: better boundaries for different parts of the project

duailibe:
Analyzing our current build of the `bin-prettier.js` we distribute on NPM and after several tries of having a better build for web I've noticed our current graph of dependencies is quite a mess.

I've started a bit of refactoring here, gonna try summarize my rationale and what changed here:

- we need to have core functionality independent of infrastructure. My rationale is: we can build another "entry" for prettier that knows how to load plugins (like in a web context, via `importScripts`) that would simply pass the loaded plugins to the core parts. So I've removed `loadPlugins` calls from `main` folder

- our `bin-prettier.js` is basically `index.js` + stuff for CLI, but it also requires `index.js` meaning users end up loading a lot of code twice. ideally the CLI would always use the public API as well (in `index.js`) so I've removed some direct calls to `getSupportInfo` to go through `prettier.getSupportInfo`.

I need help from people (cc @ikatyang and @azz) to understand a few things we have I found a bit complicated:

- what is the difference between `main/options.js`'s `normalize` function and the normalizers in `main/options-normalizers.js`. Do we need 3 normalizers? How can we make this simpler? We could rename them too, because we have way too much normalizers in the code base 😂 
- I've noticed we have duplicated logic to log unknown options and suggest another option with `leven` in two different places (one for API and one for CLI) but one ends up also including the other in the final build. Is there a way to reuse them? Should make sense to extract the core.

This is still WIP, you guys think this is worth pursuing? I can also share my views for the web build if all of this don't make sense.


ikatyang:
> - what is the difference between `main/options.js`'s `normalize` function and the normalizers in `options-normalizer`. How can we make them simpler?

- `options.js` is a prettier-specific normalizer which uses `options-*.js`.
- `options-*.js` could be a standalone package, I originally thought it'd be better there for maintainability reason, but now it seems extracting it would be better(?), what do you think?

> - I've noticed we have duplicated logic to log unknown options and suggest another option with `leven` in two different places (one for API and one for CLI) but one ends up also including the other in the final build. Is there a way to reuse them? Should make sense to extract the core.

There're actually 3 places (`api options, cli options`, and `cli --help`) where the first two are from `options-normalizer`, and the core is just a few lines:

https://github.com/prettier/prettier/blob/50101e98c6bf8a686e43913cc6fc7001ee035966/src/main/options-normalizer.js#L92-L97

> This is still WIP, you guys think this is worth pursuing?

I guess this is what you mean:

- core (no `fs`, web compatible, including `format()`, `formatWithCursor()`)
- api (require `core` and `fs`)
- cli (require `api` and `fs`)
- plugins (independent, web compatible)

👍

duailibe:
@ikatyang yeah I think you explained it better than me 😂 

> - `options.js` is a prettier-specific normalizer which uses `options-*.js`.
> - `options-*.js` could be a standalone package, I originally thought it'd be better there for maintainability reason, but now it seems extracting it would be better(?), what do you think?

Thank you for this! It helped me understand it better, but I'm still confused about what is `options-normalizer.js` responsibility.

ikatyang:
validate options, redirect deprecated options and apply default value.

ikatyang:
And yes, I'm not good at naming, so I just named it `normalizer`. 😂

index.js
```diff
@@ -17,6 +19,15 @@ const doc = require("./src/doc");
 const printDocToString = doc.printer.printDocToString;
 const printDocToDebug = doc.debug.printDocToDebug;
 
+function withPlugins(opts) {
+  opts = Object.assign({}, opts);
```
j-f1: This line is unnecessary.

duailibe: It was not but I removed it anyway :)

tests_config/run_spec.js
```diff
@@ -95,7 +95,8 @@ function stripLocation(ast) {
         key === "raw" ||
         key === "comments" ||
         key === "parent" ||
-        key === "prev"
+        key === "prev" ||
+        key === "__location"
```
j-f1: What do you think about replacing this with `["loc", "range", "raw", "comments", "parent", "prev", "__location"].indexOf(key) !== -1`?

duailibe:
@ikatyang I still think there's a some complexity that could be simplified but I think this is already good enough to have a `web` entry point which is what I want to achieve =)

src/language-css/options.js
```diff
@@ -2,7 +2,7 @@
 
 const jsOptions = require("../language-js/options");
 
-// format based on https://github.com/prettier/prettier/blob/master/src/common/support.js
+// format based on https://github.com/prettier/prettier/blob/master/src/core/options.js
```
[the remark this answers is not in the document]
duailibe: I was hoping to find a better place for it actually.. gonna move it to `main/`

duailibe:
Let's hope for the best 🙏 😂

lydell:
Thanks @duailibe ! :tada:

PR #4181, opened 2018-03-20
[WIP] feat: use `postcss-values-parser` for `at-rules`

evilebottnawi:
Why? Because `at-rule` contains same values as any value property.
Feel free to feedback :+1:

PR #4373, opened 2018-04-25
feat: use `postcss-values-parser` for `@namespace` at rule

evilebottnawi:
Part of https://github.com/prettier/prettier/pull/4181. `postcss-value-parser` is abadoned and bad parser many bleeding edge in `at-rules`

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -378,6 +378,8 @@ function parseNestedCSS(node) {
 
       if (
         [
+          "namespace",
+          "supports",
           "if",
           "else",
           "for",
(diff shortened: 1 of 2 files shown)


lydell:
No snapshots changed?

evilebottnawi:
@lydell yes

evilebottnawi:
@lydell yep, `postcss-media-query-parser` is don't respect css spec. According css spec all values in at-rules and value in properties have same syntax. Look https://www.w3.org/TR/css-syntax-3/#parser-diagrams on At-rule and Declaration - one algorithm for get tokens - i.e. ast very same, only except keywords.

evilebottnawi:
Example `cssnano` also use value parser for at rules and all works very very good. Maybe in future postcss implement basic support value parser and at-rule and we can remove some workaround, but it is don't break anything in theory :smile:

PR #4374, opened 2018-04-25
fix(json): jsonc is considered JSON instead of JSON5

ikatyang:
Just noticed that jsonc (JSON with Comments) in VSCode is indeed just JSON + comments...

image

### DIFF
--- a/src/language-js/index.js
+++ b/src/language-js/index.js
@@ -160,7 +160,7 @@ const languages = [
     extensions: [".json5"],
     filenames: [".babelrc"],
     linguistLanguageId: 175,
-    vscodeLanguageIds: []
+    vscodeLanguageIds: ["json5"]
   }
 ];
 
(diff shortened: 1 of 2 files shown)


src/language-js/index.js
```diff
@@ -160,7 +160,7 @@ const languages = [
     extensions: [".json5"],
     filenames: [".babelrc"],
     linguistLanguageId: 175,
-    vscodeLanguageIds: ["jsonc"]
+    vscodeLanguageIds: []
```
j-f1: Are there any language IDs for JSON5?

ikatyang: There's no official language id for JSON5, but there's a unofficial one (`json5`) from [community extension](https://github.com/mrmlnc/vscode-json5).

Should we use that instead?

j-f1: Sounds good. Providing an unofficial one should be fine if there isn’t an official one.

PR #4386, opened 2018-04-26
Separate entry for web bundle

duailibe:
Follow up on #4364. This PR extracts core functionality that was present in `index.js` to `main/` and creates a `web.js` entry point.

It's still not exposed after the build, gonna do that in follow-up PRs but I'd like to get that merged in already to prevent conflicts later :joy:

### DIFF
--- a/src/main/range-util.js
+++ b/src/main/range-util.js
@@ -1,0 +1,218 @@
+"use strict";
+
+const comments = require("./comments");
+
+function findSiblingAncestors(startNodeAndParents, endNodeAndParents, opts) {
+  let resultStartNode = startNodeAndParents.node;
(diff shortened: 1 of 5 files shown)


src/main/range-util.js
```diff
@@ -0,0 +1,218 @@
...
+      return {
+        node: node,
+        parentNodes: parentNodes
+      };
+    }
+  }
+}
+
+// See https://www.ecma-international.org/ecma-262/5.1/#sec-A.5
+function isSourceElement(opts, node) {
```
j-f1: Should this be extracted to the plugins?

duailibe: I think so, but no plugin asked for this yet

Issue #4370, opened 2018-04-25
Formtting a blank markdown files results in adding \n


kachkaev:
**Prettier 1.12.1**
~~~[Playground link](https://prettier.io/playground/#.....)~~~ (playground formats a default example when the input is blank)

**Current behavior:**

Unlike for `.js` and `.ts` files, formatting an empty `*.md` file results a `\n` character in the end of a non-existing line one.

**Expected behavior:**

I'd expect `blank.md` to remain zero bytes in size, just as this happens for `blank.js` and `blank.ts`.

ikatyang:
This one should be pretty easy, PRs welcome.

huyvohcmc:
I would like to take this issue, thanks.

lydell:
For quick experiments I usually run `./bin/prettier.js my-test-file.js` and look at the code printed to the console.

ikatyang:
- add `your-test-case.md` to `tests/markdown_blahblah`
- `yarn test --watch markdown_blahblah`
- make some changes to `src/language-js/printer-markdown.js` and see if it's fixed

See [CONTRIBUTING.md](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md) for more info.

(There's no notification from edits so you probably have to send a new comment next time.)

PR #4387, opened 2018-04-26
Remove newline at the end of .md files

huyvohcmc:
Remove newline at the end of markdown files #4370


duailibe:
This is not the right fix.. we do want a new line at the end of the files, we just don't want them for empty files :)

tests/markdown_empty/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`empty.md 1`] = `
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+`;
```
duailibe: Every test that changed are regressions except for this one

tests/markdown_empty/not_empty.md
```diff
@@ -0,0 +1 @@
+** Not empty **
```
duailibe: This one isn't necessary. All of the other .md files are already examples of `not_empty.md`

You could also move `empty.md` to `tests/markdown/` so you don't have a single file in the folder, but it's fine either way.

tests/markdown_empty/__snapshots__/jsfmt.spec.js.snap
duailibe: Every test that changed are regressions except for this one

tests/markdown_empty/not_empty.md
duailibe: This one isn't necessary. All of the other .md files are already examples of `not_empty.md`

You could also move `empty.md` to `tests/markdown/` so you don't have a single file in the folder, but it's fine either way.

tests/markdown_empty/__snapshots__/jsfmt.spec.js.snap
huyvohcmc: I'll fix it later, could you add a WIP label or something?

duailibe: Sure! Take your time

PR #4388, opened 2018-04-27
Remove newline in empty .md files

### LINKED ISSUE #4370: Formtting a blank markdown files results in adding \n

huyvohcmc:
Fixes #4370

Remove newline in empty .md files regarding to #4370


src/language-markdown/printer-markdown.js
```diff
@@ -68,6 +68,9 @@ function genericPrint(path, options, print) {
 
   switch (node.type) {
     case "root":
+      if (node.tokens.length === 0 && node.children.length === 0) {
```
ikatyang: There's no `tokens` attr in a standard [`root`](https://github.com/syntax-tree/mdast#root), `node.children.length === 0` should be enough.

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4235,14 +4245,14 @@ This is a paragraph.
...
 \`\`\`markdown
-* Red
-* Green
-* Blue
-
 - Red
 - Green
 - Blue
 
++ Red
```
ikatyang: These changes are caused by `node.tokens.length === 0` since there's no `tokens` in a `root` so it threw, then our multi-parser caught the error and decided not to format these codeblocks.

And sorry for the `TypeError: Cannot read property 'type' of undefined` error, I forgot to check `parent.children.length !== 0` before accessing `parent.children[0]`.

https://github.com/prettier/prettier/blob/8015e417c01a2417f9da8c15c2540258716378e7/src/language-markdown/printer-markdown.js#L805-L811

ikatyang:
FYI, you don't have to close #4387 and create another PR, you can push changes to the same branch and it'll reflect the changes here.

tests/markdown/__snapshots__/jsfmt.spec.js.snap
huyvohcmc: @ikatyang got it

huyvohcmc: @ikatyang So I'm gonna wait for the fix of `parent.children.length` or just create another PR for that?

ikatyang: We can fix it here since we found it in this PR.

ikatyang:
Thanks!

PR #4389, opened 2018-04-27
Fix cannot read property 'type' of undefined bug

huyvohcmc:
# Summary:
Fix "cannot read property 'type' of undefined" bug

# Changes proposed in this pull request:
- Add a condition check for `function clean` in `src/language-markdown/printer-markdown.js`.


lydell:
Can you add the piece of code that used to cause a crash as a test case?

ikatyang:
It'd be better to include this in #4388 since we found it there.

Issue #4360, opened 2018-04-24
Format untagged GraphQL template literals preceded by /* GraphQL */


jaydenseric:
Currently, Prettier formatting GraphQL template literals is only possible when the literals are tagged with a function such as `gql`. There are many places GraphQL strings are not tagged, here are a few examples:

- Type defs being provided to `graphql-tools` to initialize an Apollo GraphQL server ([example](https://github.com/apollographql/graphql-tools#example)).
- GraphqL queries for use in `graphql-react` `<Query />` components ([example](https://github.com/jaydenseric/graphql-react#query-1)).

This problem has long been solved in the editor syntax highlighting space with `` /* GraphQL */ `{ query }` `` syntax. Here are a few plugins for various editors that support it:

- Atom: [language-babel](https://github.com/gandm/language-babel#graphql-code-highlighting) (1.7m+ installs).
- VS Code: [vscode-language-babel](https://github.com/michaelgmcd/vscode-language-babel#why-this-package) (90k+ installs).
- VS Code: [graphql-for-vscode](https://github.com/kumarharsh/graphql-for-vscode/commit/91917f00c677ae5b650f5fabea5a922bc18a1a30) (191k+ installs).

At the moment, the ugly hack to get Prettier to format untagged GraphQL template strings is to use a fake, no-op tag like this:

```jsx
// For Prettier to detect and format GraphQL, otherwise pointless…
const gql = ([string]) => string

const foo = qgl`
  {
    foo {
      bar
    }
  }
`;
```

**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzCFMC8mA9AFSYDiATgIYAOAFgIoAympxmABsDnprwBGNKpgC+4riAA0ICHRgBLdMlAiqEAO4AFEQjTIQNAG4RFAExkhBtMAGs4MAMp0aYRVADmyGFQCucLIeaHBUMNq0ngC2NMjYNAA2IbIAVmgAHgBCtg7ONFFwLB5wcYnJIK5UIVSGwoIAngnQVnRUHjAA6hYwDMgAHAAMsq0QIR20dIatcNXGJbJUcACOfoqLETTRsUjxSYEgIVGKPv77aB6eCXBMfhDwpXuyMDSCXeY9yABMT7SKCRcAYQgURihig0HmID8IQAKi8DDsynAxGIgA)

**Input:**
```jsx
[...]

PR #4395, opened 2018-04-28
Added support for the graph ql comment tag

tjallingt:
As described in the issue https://github.com/prettier/prettier/issues/4360
Prettier still adds a space in between the comment tag and the template literal.


duailibe:
After reading https://github.com/prettier/prettier/pull/4395#commitcomment-28780404 I don't think we should do this. Feels really hack-ish. The suggestion in https://github.com/prettier/prettier/issues/4360#issuecomment-385070286 seems like a better solution

```js
const gql = String.raw;

gql`query MyQuery {}`;
```

tjallingt:
@duailibe well I mean this is the way those other tools do this as well so its not more or less hacky than those other tools ("would you jump off a bridge if your friends did" fallacy i know...) additionally 
```javascript
const gql = String.raw;

gql`query MyQuery {}`;
```
also feels kinda hacky...

but its your call, I understand your hesitation to merge this 😕

tjallingt:
Actually after thinking about it, the `String.raw` workaround doesn't work if you are using the actual `gql` tag in the same file. (Which i guess you could fix by using one of the other template tags prettier formats as graph ql but thats a workaround ontop of a workaround...)

Additionally I dislike it as a workaround because of someone sees ```gql`...```` but doesn't see the `const gql = String.raw` part they might get a completely wrong idea of what the code is doing only because prettier doesnt handle formatting arbitraty template literals as graph ql...

jaydenseric:
@duailibe https://github.com/prettier/prettier/issues/4360#issuecomment-385218271

suchipi:
@duailibe could you expound on why you think we shouldn't do this?

duailibe:
@suchipi Just feels like we already have a pretty consistent way to format non-JS stuff in JS, that is using template literals or `<style>` tags and it's been working really well.

Comments is an area of Prettier that works good enough for the effort we've put in but we still get reports of it's behavior so it seems a fragile strategy to rely on them for this kind of stuff.

That said, I changed my mind and I don't oppose to it. If other maintainers are comfortable with adding this functionality, I'm on board.

tjallingt:
So what is the status of this discussion?
Are you willing to merge this provided I fix the failing test?

j-f1:
I’m 👍 on the change, and @duailibe seems have changed their mind and is also 👍 (or at least not 👎).

duailibe:
@tjallingt to fix the test, you probably need to skip the check of the AST here: https://github.com/prettier/prettier/blob/master/src/language-js/clean.js

tjallingt:
That should fix the failing test 👍 

I really don't like that i had to duplicate the hasGraphQLComment stuff so maybe that should be exported to clean? (don't know what your stance on this is)

suchipi:
@tjallingt let's not worry about the space between the comment and the template literal tag for now. Also, I'm not sure we have a consistent approach with duplicated logic in clean; do whatever you feel is best.

tjallingt:
The it's ready for review and merge I think

lydell:
@tjallingt There are some merge conflicts to fix before merging, though

suchipi:
Fixed the conflicts

tjallingt:
oh mb didn't see them 😕 
Thanks for fixing @suchipi 👍

PR #4397, opened 2018-04-29
Fix cursor offset tracking

### LINKED ISSUE #1869: Allow `cursorOffset` to be used with `rangeStart` and `rangeEnd`
From https://github.com/prettier/prettier/pull/1637#discussion_r119144922:

> The cursor-translation functionality still needs to be unified with the range-formatting functionality (currently, either one or the other can be used). I think it might be easiest to do this by keeping track of the range start/end positions in the same way as the cursor position, then having an API that returns all 3 positions (if provided), and only formats the text necessary to satisfy the range.

### LINKED ISSUE #1981: --cursor-offset output is wrong sometimes
I just wanted to note that since the `cursorOffset` option (introduced in https://github.com/prettier/prettier/pull/1637) works by tracking the cursor position relative to an AST node (rather than a [CST](https://github.com/cst/cst) token), it can produce incorrect results. For example:

```bash
$ echo "return         15" | prettier --cursor-offset 15 # the cursor is positioned immediately before `15`
return 15;
15
```

(related to https://github.com/esbenp/prettier-vscode/issues/12#issuecomment-305913893)

ds300:
Howdy! 👋 😄 

This PR is fixing a few things related to the cursorOffset option:

1. Fix two bugs that were present before
    - If the cursor appeared before any ast nodes, it would error. (at least in TypeScript)
    - The cursor tracking was not accounting for comments preceding a node
    
2. Improve tracking with a more nuanced approach (fixes #1981)
3. Allow cursorOffset to be used with rangeStart and rangeEnd (fixes #1869)

I'll explain how I did number 2, to help you read this PR.

How cursor offset was tracked before, in broad terms:

1. find the 'cursor node', i.e. the smallest AST node which encompasses the cursor.
2. calculate the relative cursor offset from the start of that node
3. track that node until the final output text
4. calculate the final cursorOffset as being the start position of the cursor node in the final text + the relative offset calculated in step 2.

This was producing bad results because if the text between the start of the AST node and the cursor would change (e.g. whitespace being inserted or deleted) then the offset calculated in step 2 is useless and should be ignored.

The approach I take in this PR is to use that relative offset as long as the text didn't change. Otherwise, I calculate a character diff between the input and output text for the cursor node, and then figure out where the best place for the cursor would be from that diff.

In order to do that, i had to "wrap" the printed version of the cursor node using two 'cursor placeholder's, one at either end. Previously there was only one inserted at the beginning.

Let me know if you have any more questions.

Thanks for the amazing work on Prettier. I can't live without it. :bow:


tests/cursor/jsfmt.spec.js
```diff
@@ -36,3 +63,124 @@ foo('bar', cb => {
...
+  const y = 5
+
+     
+  
+  <|>  
+    
+
+
+  const z = 9
+`)
```
j-f1: What about

```js
const y = 5
<|>



const z = 9
```

IMO that should either put it on the blank line between the two or right after the `5`.

ds300: Yeah it puts it on the blank line between the two. I added this test case just now :+1:

src/main/core.js
```diff
@@ -66,41 +68,108 @@ function coreFormat(text, opts, addAlignmentSize) {
...
+      return {
+        formatted: result.formatted,
+        cursorOffset: newCursorNodeStart + cursorOffsetRelativeToOldCursorNode
+      };
+    }
+
+    // diff old and new cursor node texts, with a special cursor
+    // symbol inserted to find out where it moves to
+
+    const CURSOR = Symbol("cursor");
```
duailibe: I think we can move this to the file top level (together with `UTF8BOM`)

src/main/core.js
```diff
@@ -66,41 +68,108 @@ function coreFormat(text, opts, addAlignmentSize) {
...
+    const cursorNodeDiff = diff.diffArrays(
+      oldCursorNodeCharArray,
+      newCursorNodeCharArray
+    );
+
+    let i = 0;
+    for (const entry of cursorNodeDiff) {
+      if (entry.removed) {
+        if (entry.value.indexOf(CURSOR) > -1) {
+          break;
```
duailibe: Maybe?

```js
if (entry.removed && entry.value.indexOf(CURSOR) > -1) {
```

src/main/core.js
```diff
@@ -127,26 +196,47 @@ function formatRange(text, opts) {
...
+  } else if (rangeResult.cursorOffset !== undefined) {
+    // handle the case where the cursor was in the range
+    cursorOffset = rangeResult.cursorOffset + rangeStart;
+  }
+  // keep the cursor as it was if it was before the start of the range
 
-  return text.slice(0, rangeStart) + rangeTrimmed + text.slice(rangeEnd);
+  return {
+    formatted: formatted,
+    cursorOffset: cursorOffset
```
duailibe:
```js
return { formatted, cursorOffset }
```

=)

src/main/core.js
```diff
@@ -66,41 +68,108 @@ function coreFormat(text, opts, addAlignmentSize) {
...
-    return { formatted: formatted.trim() + opts.newLine };
+    const trimmed = result.formatted.trim();
+    const cursorNodeStart =
+      result.cursorNodeStart === undefined
+        ? undefined
+        : result.cursorNodeStart - result.formatted.indexOf(trimmed);
+    result = Object.assign({}, result, {
+      formatted: trimmed + opts.newLine,
+      cursorNodeStart
+    });
```
duailibe: You can change the `result` object here

src/main/core.js
```diff
@@ -66,41 +68,108 @@ function coreFormat(text, opts, addAlignmentSize) {
...
+    let i = 0;
+    for (const entry of cursorNodeDiff) {
+      if (entry.removed) {
+        if (entry.value.indexOf(CURSOR) > -1) {
+          break;
+        }
+      } else {
+        i += entry.count;
+      }
+    }
```
duailibe:
```js
let cursorOffset = newCursorNodeStart;

for (const entry of cursorNodeDiff) {
  /* code */
  cursorOffset += entry.count;
}

return { formatted: result.formatted, cursorOffset };
```

tests/cursor/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,305 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`puts the cursor in sensible places 1`] = `
+Object {
+  "formatted": "return <|>15;
+",
+  "original": "return        <|> 15",
+}
+`;
+
```
duailibe: I think we can improve these snapshots, see below

duailibe:
I'd like to adapt the tests to use our current infrastructure. Can you give me push access to this PR? Just check "Allow edits from maintainers"

src/main/core.js
j-f1: Removing `result =` and `{},` should have that effect.

src/main/core.js
ds300: It's not the same thing, because there's an `else` clause. Would need to change the `else` to `else if (!entry.removed)` in order for that to work, which IMO is a bit more confusing. But happy to change it if you prefer.

ds300:
Cool stuff with the snapshots! I'll try to get around to those small changes later today.

src/main/core.js
duailibe: Oops sorry!

ds300:
All done @duailibe !

ds300:
Haha, I wasn't sure whether you'd finish. I did the rest just now. Cheers!

duailibe:
Thank you!!! 💯 👏 🎉

tests/cursor/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,245 @@
...
+    thisWillBeFormatted    (2  ,3,   )
+
+    thisWontBeFormatted  (2, 9<|>0  ,)
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+thisWontBeFormatted  ( 1  ,3)
+
+    thisWillBeFormatted(2, 3);
+
+    thisWontBeFormatted  (2, 9<|>0  ,)
```
vjeux: Those snapshot formats are really cool, super easy to understand what’s going on!

ds300:
I forgot to say that the reason I was working on this is that I was building a little toy text editor to see what it would feel like if prettier's formatting happened with a smooth animated transition. Turns out it feels great — https://code-in-motion.herokuapp.com/

vjeux:
This sounds awesome!

I’m currently in vacation with just my iphone and it doesn’t seem to be working. I’d love if you could make a small video :)

ds300:
sure, here's a small contrived example

codesbad

vjeux:
That looks slick! Do you do it “on save” or are you able to do it on arbitrary keystrokes that would lead to a code that actually parses?

ds300:
just on save so far. I tried to do it after every keystroke but it doesn't work because sometimes prettier deletes things you type like commas :) so to work automatically it would need some basic heuristics about when to format and when not to format.

But also triggering it manually gives this nice cathartic feeling, for me at least.

PR #4394, opened 2018-04-28
avoid overflow in playground-container

johann-sonntagbauer:
fixes #4350

playground-container
overflow hidden forces the container to don't outgrow its space

sub-options
fixes Problem in Edge

PR #4406, opened 2018-05-01
[playground] fix overflow in safari

johann-sonntagbauer:
fixes problems in safari discussed in https://github.com/prettier/prettier/pull/4394.

Now the options panel will not outgrow editors panel.

### DIFF
--- a/website/static/separate-css/playground.css
+++ b/website/static/separate-css/playground.css
@@ -180,6 +180,7 @@ pre.CodeMirror-placeholder {
   display: none;
   box-sizing: border-box;
   width: 225px;
+  overflow-y: auto;
   border-bottom: 1px solid #ddd;
   border-right: 1px solid #ddd;
   flex: 0 1 auto;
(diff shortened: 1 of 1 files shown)


duailibe:
Thank you!! 💯 🎉

duailibe:
I don't think this was the right fix.. I guess we shouldn't mess with the CodeMirror classes, we broke it in another way

screen shot 2018-05-04 at 11 45 37

johann-sonntagbauer:
@duailibe sorry for that, I will try to test even further, what safari os combination do you face that problem?
Tested quite a few combinations without any issues.

j-f1:
@johann-sonntagbauer That’s Chrome, not Safari.

duailibe:
Yeah it's Chrome on macOS

johann-sonntagbauer:
yes you are right, the code mirror gutter styles work as expected, the problem seems to be related only to the placeholder. I will formulate a pr for the revert and dig a little deeper on the placeholder stuff later on.

PR #4408, opened 2018-05-01
fix spacing issue in postcss simple vars

### LINKED ISSUE #4348: postcss-simple-vars: computed variable names broken
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAjAhmA1gcwE4QCuUAJgLSQA2EeSABACQMAUAzjAJ6VwCUAwhGp4A3AB0oIADQgIABxgBLaK2Sh0eAgHcACuoQqU6AG4QFJKSHTtkAM3SVWcaajyYscGAGVZmBVBzIMHiETiAkEGC29o7Sfo54MNquOAC26FEOoQBWrAAeAEKu2B6e6ClwADJ+cBkxMoQwsg0ATLWhPnjxyOCsKtKyeH4wAOpmMAAWyAAcAAz9BI7DrrLdA3DxRjXSeHAAjoQKO0noqelIdpnSjikKgcGhrH443ACKhBDwbdIw6KijJBNkM1vq4FJQngIUmlulBoFsQIRHAAVX4GC6OAC+GKAA)
```sh
--parser css
```

**Input:**
```css
background-color: $$(style)Color;

```

**Output:**
```css
background-color: $$(style) Color;

```

**Expected behavior:**
Output should be same as input, i.e. no additional space should be introduced.

ad1992:
Fixes #4348

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -44,7 +44,7 @@ const hasComposesValueNode = utils.hasComposesValueNode;
 const hasLessExtendValueNode = utils.hasLessExtendValueNode;
 const hasParensAroundValueNode = utils.hasParensAroundValueNode;
 const maybeToLowerCase = utils.maybeToLowerCase;
-const hasNoSpaceBeforeValueNode = utils.hasNoSpaceBeforeValueNode;
+const isPostcssSimpleVar = utils.isPostcssSimpleVar;
 
 function genericPrint(path, options, print) {
   const node = path.getValue();
@@ -410,10 +410,7 @@ function genericPrint(path, options, print) {
         const iNextNode = node.groups[i + 1];
         const iNextNextNode = node.groups[i + 2];
 
-        if (
-          iNode.type === "value-func" &&
-          hasNoSpaceBeforeValueNode(iNextNode)
-        ) {
+        if (isPostcssSimpleVar(iNode, iNextNode)) {
           continue;
         }
         // Ignore after latest node (i.e. before semicolon)
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -409,6 +409,14 @@ function genericPrint(path, options, print) {
         const iNextNode = node.groups[i + 1];
         const iNextNextNode = node.groups[i + 2];
 
+        if (
+          iNode.type === "value-func" &&
+          iNextNode &&
+          iNextNode.type === "value-word" &&
+          !iNextNode.raws.before
+        ) {
+          continue;
+        }
         // Ignore after latest node (i.e. before semicolon)
         if (!iNextNode) {
           continue;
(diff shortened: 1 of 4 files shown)


ad1992:
@evilebottnawi made the change.

src/language-css/printer-postcss.js
```diff
@@ -409,6 +410,12 @@ function genericPrint(path, options, print) {
         const iNextNode = node.groups[i + 1];
         const iNextNextNode = node.groups[i + 2];
 
+        if (
+          iNode.type === "value-func" &&
+          hasNoSpaceBeforeValueNode(iNextNode)
```
evilebottnawi: Better add helper `isPostcssSimpleVar` and move all check inside this util function.

evilebottnawi:
@ad1992 yes, just need add tests when word have other value

ad1992:
@evilebottnawi I have added some specs but these specs are already covered in other files so I am not sure if its needed here for example `@@var` is already in `less.css`

src/language-css/utils.js
```diff
@@ -211,6 +211,14 @@ function hasParensAroundValueNode(node) {
   );
 }
 
+function isPostcssSimpleVar(currentNode, nextNode) {
```
lydell: The check in here is a bit hard to understand. At a glance it feels too generic. Is it? Could his have unintended consequences? For example, do we need to check that the name of something is `$`?

evilebottnawi: /cc @ad1992

lydell: Perhaps the function logic is fine, but the name `isPostcssSimpleVar` is misleading? Because that's not what it checks?

ad1992: initially the code was
```js
iNode.type === "value-func" &&
hasNoSpaceBeforeValueNode(iNextNode)
```
and then changed it to isPostcssSimpleVar. Yes its not explicitly checking whether is `PostcssSimpleVar`. 
should it be isLeadingSpaceInPostCssSimpleVar ? Please suggest some better name :)

evilebottnawi: @ad1992 just add check on `$` in your utils function and all

src/language-css/utils.js
```diff
@@ -211,6 +211,14 @@ function hasParensAroundValueNode(node) {
...
+  return (
+    currentNode.type === "value-func" &&
```
evilebottnawi: `currentNode.name === '$$'`

evilebottnawi: /cc @ad1992

ad1992: @evilebottnawi if I add the check  `currentNode.value === '$$'` then it will only work for functions with name `$$`, etc. Is that the correct behaviour ? Then I should also add test where function name is something else other than `$$`

evilebottnawi: @ad1992 postcss simple var have `$$(name)` syntax and it is normal, yes add test where we have `@@(name)`, maybe other tests too

ad1992: got it. Made the changes.

evilebottnawi:
@vjeux sorry, accidentally mentioned you :smile:

ad1992:
@j-f1 please review

j-f1:
I don’t know the CSS printer that well, so I’m not really qualified to review.

ad1992:
@evilebottnawi @lydell can this PR be merged if there are no more pending reviews ?

PR #3649, opened 2018-01-05
[scss] Print comments inside of selector as is

vjeux:
Fixes #3647

PR #4421, opened 2018-05-04
[FAILING] Upgrade to postcss-selector-parser v5.0 release candidate.

### LINKED ISSUE #4422: Upgrade postcss-selector-parser 🙏🏼
I'm the maintainer for postcss-selector-parser. The existing dependency on postcss-selector-parser `2.2.3` is quite old. The latest release is `5.0.0-rc.0` and I'd like to do the full release of 5.0 soon. If that release has any regressions that affects prettier, I'd like to know soon so i can fix them before the release is out.

The [CHANGELOG](https://github.com/postcss/postcss-selector-parser/blob/master/CHANGELOG.md) outlines the API changes and gives a lot of code examples for how to use them.

I provided patches to stylelint that demonstrate how to upgrade through the latest patch.

https://github.com/stylelint/stylelint/pull/2984
https://github.com/stylelint/stylelint/pull/3284

I don't have enough cycles to do upgrades like that for every major project that is using it. I made a [PR that demonstrates what's failing](https://github.com/prettier/prettier/pull/4421
).

Please file bugs for undocumented regressions at https://github.com/postcss/postcss-selector-parser/issues.

chriseppstein:
This PR just demonstrates the failing tests for your review. feel free to close it without merging.

Fixes #4422

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -17,11 +17,8 @@ function parseSelector(selector) {
     };
   }
   const selectorParser = require("postcss-selector-parser");
-  let result;
-  selectorParser(result_ => {
-    result = result_;
-  }).process(selector);
-  return addTypePrefix(result, "selector-");
+  const ast = selectorParser().astSync(selector);
+  return addTypePrefix(ast, "selector-");
 }
 
 function parseValueNodes(nodes) {
@@ -155,11 +152,12 @@ function addTypePrefix(node, prefix) {
   if (node && typeof node === "object") {
     delete node.parent;
     for (const key in node) {
-      addTypePrefix(node[key], prefix);
       if (key === "type" && typeof node[key] === "string") {
         if (!node[key].startsWith(prefix)) {
           node[key] = prefix + node[key];
         }
+      } else {
+        addTypePrefix(node[key], prefix);
       }
     }
   }
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -316,9 +316,9 @@ function genericPrint(path, options, print) {
           : "",
         node.attribute.trim(),
         node.operator ? node.operator : "",
-        node.value
+        typeof node.value !== "undefined"
           ? quoteAttributeValue(
-              adjustStrings(node.value.trim(), options),
+              node.quoteMark ? adjustStrings(node.raws.value, options) : node.value,
               options
             )
           : "",
--- a/tests/css_quotes/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css_quotes/__snapshots__/jsfmt.spec.js.snap
@@ -99,7 +99,7 @@ one 'two' three {}
 @charset 'UTF-8'; /* Single quotes are invalid here. Keep them since we don't know what the user is doing. */
 
 @supports (content: one "two" three "four") {
-  a[href="foo" y],
+  a[href="foo"],
   abbr[title^="It's a trap!"],
   img[src=""] {
     /* Simple strings. */
@@ -288,7 +288,7 @@ one 'two' three {}
 @charset 'UTF-8'; /* Single quotes are invalid here. Keep them since we don't know what the user is doing. */
 
 @supports (content: one 'two' three 'four') {
-  a[href='foo' y],
+  a[href='foo'],
   abbr[title^="It's a trap!"],
   img[src=''] {
     /* Simple strings. */
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - 4
+  - 6
   - stable
 cache:
   yarn: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - 6
+  - 4
   - stable
 cache:
   yarn: true
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "postcss-less": "1.1.5",
     "postcss-media-query-parser": "0.2.3",
     "postcss-scss": "1.0.5",
-    "postcss-selector-parser": "2.2.3",
+    "postcss-selector-parser": "5.0.0-rc.0",
     "postcss-values-parser": "1.5.0",
     "read-pkg-up": "3.0.0",
     "remark-frontmatter": "1.1.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "postcss-less": "1.1.5",
     "postcss-media-query-parser": "0.2.3",
     "postcss-scss": "1.0.5",
-    "postcss-selector-parser": "5.0.0-rc.0",
+    "postcss-selector-parser": "5.0.0-rc.1",
     "postcss-values-parser": "1.5.0",
     "read-pkg-up": "3.0.0",
     "remark-frontmatter": "1.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1363,6 +1363,10 @@ crypto-random-string@^1.0.0:
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/crypto-random-string/-/crypto-random-string-1.0.0.tgz#a230f64f568310e1498009940790ec99545bca7e"
 
+cssesc@^1.0.1:
+  version "1.0.1"
+  resolved "https://registry.yarnpkg.com/cssesc/-/cssesc-1.0.1.tgz#ef7bd8d0229ed6a3a7051ff7771265fe7330e0a8"
+
 cssom@0.3.x, "cssom@>= 0.3.2 < 0.4.0":
   version "0.3.2"
   resolved "https://registry.yarnpkg.com/cssom/-/cssom-0.3.2.tgz#b8036170c79f07a90ff2f16e22284027a243848b"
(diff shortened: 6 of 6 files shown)


j-f1:
Are you OK with dropping Node 4 support in the next release @lipis?

vjeux:
> I can relax the engine constraint in our package.json, afaik, our code still works in node 4, but it's not an officially supported version anymore.

I think that we should keep supporting node 4 for as long as it’s not becoming a huge overhead on our side.

Breaking compatibility is the worst thing you can do in terms of adoption, so we should strive to work as far back as possible, even if versions are no longer marked as officially supported.

Having been involved in upgrading node versions at Facebook, it’s not a straightforward process and if we can avoid prettier be the cause of having to do such an upgrade, it would be very benefic for the project

evilebottnawi:
@vjeux Two sides of the problem:
1. I agree we should keep supporting node 4 for as long as it’s not becoming a huge overhead on our side
2. postcss-selector-parser have very important fixes and improvement, what we can't solve with old version. And we need this update to solve this problem :disappointed:

duailibe:
Expanding on what I mentioned in the other thread I think we can drop it oficially but still make it work on node 4, which is:

- we upgrade our dependencies nevertheless;
- use `--ignore-engines` in the travis node 4 run and see if things still work.

If something breaks then we need to think about what to do

evilebottnawi:
@duailibe great solution :+1:

evilebottnawi:
But we should add docs about what some parser require `node >= 6`

chriseppstein:
I can relax the node engine check in postcss-selector-parser. But I can’t guarantee that future releases won’t have a bug because my tests can’t run on node 4 anymore because the latest versions of `npm` don’t support it and so my tests don’t run in that environment.

evilebottnawi:
@chriseppstein it is will be great. I think that we will be able to deal with the node support issue to next releases

chriseppstein:
I just released `postcss-selector-parser@5.0.0-rc.1` which changes the node engine check to `>= 4`.

j-f1:
Can you ⬆️ the parser version here and revert the commit dropping support for Node 4?

chriseppstein:
@j-f1 that's what I did in `postcss-selector-parser@5.0.0-rc.1`

duailibe:
@chriseppstein I'd like to discuss a few things I've found trying this branch. Do you prefer to discuss it in this PR or in the postcss-selector-parser repo?

evilebottnawi:
@duailibe feel free to ping me if you need help, i found some strange places in prettier.
Example: we pass `.foo { font: { family: value; weight: 900} }` to selector parser, we should not pass this to selector parser

tests/css_quotes/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -288,7 +288,7 @@ one 'two' three {}
 @charset 'UTF-8'; /* Single quotes are invalid here. Keep them since we don't know what the user is doing. */
 
 @supports (content: one 'two' three 'four') {
-  a[href='foo' y],
+  a[href='foo'],
```
evilebottnawi: @chriseppstein says what is 4/5 version postcss selector parser support custom modificator

evilebottnawi:
@duailibe you can assign issue to me,i have time on this tomorrow

duailibe:
Sure, go ahead

tests/css_quotes/__snapshots__/jsfmt.spec.js.snap
chriseppstein: We support nonstandard modifier. the value is stored in `raws.insensitiveFlag`

https://github.com/postcss/postcss-selector-parser/blob/master/src/__tests__/attributes.js#L396-L402

I think the API could be better around this, so I filed an [issue to track it](https://github.com/postcss/postcss-selector-parser/issues/135) a while ago.

evilebottnawi: @chriseppstein thanks for clarify. `raws.insensitiveFlag` looks some mesleading name for non standard flags

src/language-css/parser-postcss.js
```diff
@@ -17,11 +17,8 @@ function parseSelector(selector) {
     };
```
chriseppstein: regarding this comment above:

```
// If there's a comment inside of a selector, the parser tries to parse
// and reformat it.
```

This is no longer true. We have many tests around comments to ensure they are preserved across parse and re-stringification.

https://github.com/postcss/postcss-selector-parser/blob/master/src/__tests__/comments.js
https://github.com/postcss/postcss-selector-parser/blob/master/src/__tests__/attributes.js#L355-L394

But... the way you are using the selector api causes the comments to be dropped in some cases. Comments are stored in the `raws`. There are methods for accessing the value(s) of a node and the spaces around them in a consistent way. If you think we need other methods, feel free to suggest or send a PR.

src/language-css/parser-postcss.js
```diff
@@ -17,11 +17,8 @@ function parseSelector(selector) {
...
   }
   const selectorParser = require("postcss-selector-parser");
-  let result;
-  selectorParser(result_ => {
-    result = result_;
-  }).process(selector);
-  return addTypePrefix(result, "selector-");
+  const ast = selectorParser().astSync(selector);
+  return addTypePrefix(ast, "selector-");
```
chriseppstein: Mutating our value for `type` and removing the `parent` back-reference is unsupported and may cause bugs in future minor releases. I recommend that you stop doing this.

If you want to associate your own data with our nodes you can set your own attributes but this is likely to force v8 to de-opt (the delete on the `parent` property is likely already doing this -- what is the point of that?). A good solution is to use a WeakMap to set your own values to be associated with any of our nodes. these will get GC'd automatically.

src/language-css/parser-postcss.js
j-f1: If I remember correctly, people were opening issues about `//`-style comments, and I don’t see any tests for those after a quick glance through the links you provided.

tests/css_quotes/__snapshots__/jsfmt.spec.js.snap
chriseppstein: I agree that's why I filed the bug.

src/language-css/parser-postcss.js
duailibe: No, it was about regular CSS comments. See #3649

src/language-css/printer-postcss.js
```diff
@@ -316,9 +316,9 @@ function genericPrint(path, options, print) {
...
         node.attribute.trim(),
         node.operator ? node.operator : "",
-        node.value
+        typeof node.value !== "undefined"
           ? quoteAttributeValue(
-              adjustStrings(node.value.trim(), options),
+              node.quoteMark ? adjustStrings(node.raws.value, options) : node.value,
               options
             )
           : "",
```
chriseppstein: This entire approach to space trimming is hostile to intervening comments as it does not consider the `raws` setting for spaces which is where the comments are stored.

src/language-css/parser-postcss.js
duailibe: I don't understand, what exactly is the problem doing this? Once we get the AST, what exactly should be "supported"?

We're only interested in traversing the tree top-down and have our own structure to traverse it (and get the parents, see: https://github.com/prettier/prettier/blob/master/src/common/fast-path.js)

src/language-css/printer-postcss.js
duailibe: I don't have the context on this, but there must be a reason for doing this. I didn't touch this part of the code that much but in my (little) experience, the AST output on CSS in general is very brittle. We have all sorts of hacks to prevent printing anything wrong.

src/language-css/parser-postcss.js
chriseppstein: Given that it works, there is no problem at this time. But removing the parent has certainly introduced bugs in methods that you are not using. E.g. `node.replace()`. Changing type has broken our [typeGuard methods](https://github.com/postcss/postcss-selector-parser/blob/master/src/selectors/guards.js).

I'm just letting you know that it's unsupported and I can't guarantee patch release compatibility when you mutate internals like this. Please consider how you can structure this code so that such changes to our internals are not performed. 🙏

src/language-css/printer-postcss.js
chriseppstein: I added a commit on this other branch with some ideas for how this code could work better. But I don't see it working nicely with the tests, so I don't really understand what's the issue there. Hopefully it will add some thoughts for how you can do this differently without dropping intervening comments within the attribute.

https://github.com/chriseppstein/prettier/commits/selector-parser-v5-comment-preservation

src/language-css/parser-postcss.js
duailibe: Understood!

What I meant is, we treat the output of the parser as a simple object and don't intentionally use any "methods" of it.

src/language-css/parser-postcss.js
chriseppstein: Supporting `//` style comments isn't that hard. if it's really important I can add support to our lexer for them.

src/language-css/parser-postcss.js
chriseppstein: The AST for selectors has multiple properties that are used in coordination to handle a variety of edge cases. This approach will invariably lead to you implementing your own version of our methods or having bugs that are hard to catch due to incomplete test cases. 

In the latest version we handle escapes transparently for you. This is a semantic change that very likely fixes a number of bugs you didn’t know you had but if you don’t read or set the properties correctly you could result in changing the original source in a way that you don’t want.

Are there test cases in prettier around various escapes and extended Unicode characters in selectors? If so can you point me to them so I can verify that they are testing aspects that are especially tricky to get right?

chriseppstein:
@duailibe will you have time to work on this more soon? There's some pretty bad bugs in the last release and I'd like to get 5.0 shipped soon.

duailibe:
Hey @chriseppstein. sorry I didn't work on this anymore, I left it to @evilebottnawi who asked to work on that.

I think you can go ahead and release 5.0 if you're already comfortable with it. Don't feel you have to wait for us, sorry we couldn't deliver this on time

chriseppstein:
@duailibe @evilebottnawi it would be really helpful if someone could figure out what input is causing the hanging tests, that's the issue I'm most worried about right now.

duailibe:
There aren't any tests hanging in travis. That happened locally?

evilebottnawi:
Sorry for delay, WIP

evilebottnawi:
Close in favor. Feel free reopen if you think some discussions will be place here

PR #4430, opened 2018-05-06
refactor(massageAst): move target props to corresponding plugins

ikatyang:
I'm not sure why but it seems some of them are unnecessary.

```js
[
  "loc",
  "range",
  "raw",
  "comments",
  "leadingComments",   // js
  "trailingComments",  // js
  "extra",             // js
  "start",             // js, vue
  "end",               // js, vue
  "tokens",
  "flags",             // js
  "raws",              // css
  "sourceIndex",       // css
  "id",
  "source",            // css
  "before",            // css
  "after",             // css
  "trailingComma",     // css
  "parent",
  "prev",
  "position"           // md
].forEach(name => {
  delete newObj[name];
});
```

### DIFF
--- a/src/language-css/clean.js
+++ b/src/language-css/clean.js
@@ -3,6 +3,12 @@
 const htmlTagNames = require("html-tag-names");
 
 function clean(ast, newObj) {
+  ["raws", "sourceIndex", "source", "before", "after", "trailingComma"].forEach(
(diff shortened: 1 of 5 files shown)


duailibe:
@j-f1 Travis runs with `AST_COMPARE=1`

Issue #4313, opened 2018-04-13
Less: URLs that contain commas are destroyed


shroudedcode:
**Prettier 1.12.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAlgWwA4QE4wAEA5ABYwzYDOSA9LQGbQxUB0A5hBOwDZwCG2dG0iZaYKlQD8DfpnQ8AngF4AyhACueMHADUq-lCq6ACnghIALAAZrAGgBst4gG4AOlBB2QEbDHTQVMig-HjmAO4moQhBKPwAbhDoACZeIPxUMMiyPFRw3gBGePxgANZwMKrYJehQ7MgweBr5IMkQYNn8uS21eQRm-OyY-J3d3gBWVAAeAELFZRUGmHAAMrVwo3nempQaMABMmy3VeH3IIHySadh4tTAA6ikwpMgAHPYgNxB598XY5zc4H14htvHg4ABHDTocEDIYjJA5LYgPLyBpNFpUWq8OAARQ0EHgR28MH4BUeyWeyH2JOKCmxAGEIJhhucoNBQSANHkACpk2JIuAAXyFQA)
```sh
--parser less
```

**Input:**
```less
@import 'https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600';
```

*This is a perfectly valid URL which happens to contain a comma.*

**Output:**
```less
@import 'https://fonts.googleapis.com/css?family=Source+Sans+Pro:400, 600';
```

*Prettier has added a space after the comma destroying the URL.*

**Expected behavior:**
Prettier should leave the URL untouched.

ad1992:
@j-f1 can I take this up ?

evilebottnawi:
@ad1992 do it :+1:

ad1992:
the output if the postcss-media-query-parser at https://github.com/prettier/prettier/blob/3bc9b68cf69fbd1628923b6530f5b65dee024435/src/language-css/parser-postcss.js#L205 for the input
`@import 'https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600';`  is
```js
Node {
  after: '',
  before: '',
  type: 'media-query-list',
  value: '\'https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600\'',
  sourceIndex: 0,
  nodes:
   [ Node {
       after: '',
       before: '',
       type: 'media-query',
[...]

PR #4435, opened 2018-05-07
update postcss value parser for @import and at rule

ad1992:
fixes https://github.com/prettier/prettier/issues/4313.  WIP on fixing the specs.

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -366,7 +366,7 @@ function parseNestedCSS(node) {
 
       if (name === "at-root") {
         if (/^\(\s*(without|with)\s*:[\s\S]+\)$/.test(params)) {
-          node.params = parseMediaQuery(params);
+          node.params = parseValue(params);
         } else {
           node.selector = parseSelector(params);
           delete node.params;
@@ -423,7 +423,12 @@ function parseNestedCSS(node) {
           };
         }
 
-        node.params = parseMediaQuery(params);
+        if(lowercasedName === "import") {
+          node.params = parseValue(params);
+        }
+        else {
+          node.params = parseValue(params);
+        }
 
         return node;
       }
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -375,6 +375,18 @@ function parseNestedCSS(node) {
         return node;
       }
 
+      if (lowercasedName === "import") {
+        if (params.includes("#{")) {
+          // Workaround for media at rule with scss interpolation
+          return {
+            type: "media-unknown",
+            value: params
+          };
+        }
+        node.params = parseValue(params);
+        return node;
+      }
+
       if (
         [
           "namespace",
@@ -414,7 +426,7 @@ function parseNestedCSS(node) {
         return node;
       }
 
-      if (["import", "media", "custom-media"].indexOf(lowercasedName) !== -1) {
+      if (["media", "custom-media"].indexOf(lowercasedName) !== -1) {
         if (params.includes("#{")) {
           // Workaround for media at rule with scss interpolation
           return {
@@ -423,11 +435,7 @@ function parseNestedCSS(node) {
           };
         }
 
-        if (lowercasedName === "import") {
-          node.params = parseValue(params);
-        } else {
-          node.params = parseMediaQuery(params);
-        }
+        node.params = parseMediaQuery(params);
 
         return node;
       }
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -423,10 +423,9 @@ function parseNestedCSS(node) {
           };
         }
 
-        if(lowercasedName === "import") {
+        if (lowercasedName === "import") {
           node.params = parseValue(params);
-        }
-        else {
+        } else {
           node.params = parseMediaQuery(params);
         }
 
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -427,7 +427,7 @@ function parseNestedCSS(node) {
           node.params = parseValue(params);
         }
         else {
-          node.params = parseValue(params);
+          node.params = parseMediaQuery(params);
         }
 
         return node;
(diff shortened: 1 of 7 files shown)


src/language-css/printer-postcss.js
```diff
@@ -582,7 +582,19 @@ function genericPrint(path, options, print) {
           isNextForKeyword ||
           isNextEachKeyword
         ) {
-          parts.push(line);
+          if (atRuleAncestorNode && atRuleAncestorNode.name === "import") {
```
ad1992: @evilebottnawi I have added some more specs and added more checks in here but I think these checks are too many and the `css_comments` is still different from the original one. Can you tell me how can I make it better ?

src/language-css/printer-postcss.js
```diff
@@ -593,7 +605,13 @@ function genericPrint(path, options, print) {
       if (isControlDirective) {
         return group(indent(concat(parts)));
       }
-
+      if (
+        atRuleAncestorNode &&
+        atRuleAncestorNode.name === "import" &&
+        node.groups[0].value === "url"
```
evilebottnawi: Improct can be without `url`function

src/language-css/printer-postcss.js
evilebottnawi: Alright with comments, don't touch their right now

src/language-css/parser-postcss.js
```diff
@@ -423,7 +423,11 @@ function parseNestedCSS(node) {
           };
         }
 
-        node.params = parseMediaQuery(params);
+        if (lowercasedName === "import") {
+          node.params = parseValue(params);
+        } else {
+          node.params = parseMediaQuery(params);
+        }
```
evilebottnawi: Move this here https://github.com/ad1992/prettier/blob/e84c9949924e7b1046c07dff63a0c914abc0f6d7/src/language-css/parser-postcss.js#L394

src/language-css/parser-postcss.js
```diff
@@ -375,6 +375,18 @@ function parseNestedCSS(node) {
         return node;
       }
 
+      if (lowercasedName === "import") {
+        if (params.includes("#{")) {
```
evilebottnawi: Don't use this, `postcss-values-parser` works great with `#{$var}`

tests/css_atrule/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2288,6 +2295,8 @@ $dir: "fonts";
 @import url("very-very-very-very-very-very-very-very-very-very-long-name.css")
   projection,
   tv;
+@import url("very-very-very-very-very-very-very-very-very-very-long-name.css")
+projection tv;
```
evilebottnawi: Looks like problem with indent

src/language-css/printer-postcss.js
```diff
@@ -582,7 +582,19 @@ function genericPrint(path, options, print) {
...
+          if (atRuleAncestorNode && atRuleAncestorNode.name === "import") {
+            if (iNextNode.open) {
+              parts.push(iNextNode.open.raws.before.replace(/\s\s+/g, " "));
+            } else if (iNode.close || (iNode.group && iNode.group.close)) {
+              parts.push(line);
+            } else if (iNextNode && iNextNode.raws) {
+              parts.push(iNextNode.raws.before.replace(/\s\s+/g, " "));
+            } else if (iNode && iNode.raws) {
+              parts.push(iNode.raws.after.replace(/\s\s+/g, " "));
+            }
```
evilebottnawi: Looks very crazy, we should:
1. Search other solution
2. If no solution found add comments about each `if`

Even I can not understand what is happening here :confused:

ad1992: @evilebottnawi I added this code when I tried fixing comments and even I didn't like this code . But I am removing it now as you said to keep comments as it is and adding url check https://github.com/prettier/prettier/pull/4435/commits/ff3d67ff2d49e24351b8b3aa490edcac8cf45c9e

src/language-css/parser-postcss.js
ad1992: screen shot 2018-05-12 at 10 30 21 pm

It was adding a comma why I added this check here @evilebottnawi 
should I remove this check from here and try fixing it in `printer-postcss.js file` ?

evilebottnawi: @ad1992 it is very strange, here only `@import` check, code around `media` below and we have check on `#{`

ad1992: @evilebottnawi my bad. this is fixed. 
Though all the test cases except for comments pass now but there are still some issues on how the postcss value parser behaves when having `,` in import statement .  i am working on fixing it.
For example testcase like
`@import url("very-very-very-very-very-very-very-very-very-very-long-name.css") projection,tv,tv1;
`
output of media query parser
```css
@import url("very-very-very-very-very-very-very-very-very-very-long-name.css")
  projection,
  tv,
  tv1;
```
Output of post css value parser
```css
@import url("very-very-very-very-very-very-very-very-very-very-long-name.css")
  projection,
  tv,  tv1;
```

src/language-css/printer-postcss.js
```diff
@@ -593,7 +593,14 @@ function genericPrint(path, options, print) {
       if (isControlDirective) {
         return group(indent(concat(parts)));
       }
-
+      if (
+        atRuleAncestorNode &&
+        atRuleAncestorNode.name === "import" &&
+        node.groups[0].value === "url" &&
+        node.groups.length === 2
```
evilebottnawi: Why two? Can you add comments about this?

ad1992: @evilebottnawi Indent is not needed for import url when node has two groups

src/language-css/printer-postcss.js
```diff
@@ -593,6 +593,13 @@ function genericPrint(path, options, print) {
       if (isControlDirective) {
         return group(indent(concat(parts)));
       }
+
+      /**
+      * Indent is not needed for import url when node has two groups
+      * when type is value-comma_group
+      * example @import url("abc") projection,tv
+      */
```
evilebottnawi: Good comment, don't use `/*` and `*/` it is for JSDoc comments, use simple `//`

PR #4436, opened 2018-05-07
Add /coverage/ to .eslintignore

kachkaev:
I generated coverage reports and then ran `yarn lint`. The script took ages to run and printed lots of useless errors.

Adding `/coverage/` to `.eslintignore` fixes `yarn lint`'s performance.


azz:
Oh right. I noticed this once but assumed it was traversing `node_modules`, never let it run to the end. Thanks!

PR #4440, opened 2018-05-08
feat(markdown): change unordered list symbol to hyphen

### LINKED ISSUE #4251: Markdown Unordered List Item Symbol
Hello Prettier! I'm a big fan of using Prettier in my JavaScript projects, and I recommend it to all my friends an colleagues. I was really excited about the addition of Markdown formatting, until I found that unordered list items are delineated with `*` instead of `-` (see [this issue](https://github.com/prettier/prettier/issues/3025) for initial discussion).

I have never even seen `*` used for list items before Prettier started changing my markdown files. I think I knew it was part of the Markdown spec but I'd never seen it used in practice. I stopped using Prettier for Markdown as a result, and [I'm not the only one](https://github.com/strongloop/loopback-next/pull/1139). I do appreciate that Prettier is opinionated, but in this case the "opinion" was informed only by a measure of "what's more common" based on a [Big Data query](https://bigquery.cloud.google.com/savedquery/652929483875:69be015faec8444c8a2e7d055eeb32f8) that I can't seem to access (so I can't assess the quality of the data). I would argue that, absent an advocate from either side, this opinion lacks context. And even taking the data on face value, `-` use stood at 39%, which I believe is enough to warrant a discussion around an option to support it. 

Personally, `-` makes far more sense to me as a list item symbol because it matches the YAML notation for list items. It is also widely used in word processors to begin lists (type `- ` on a new line and both Word and Pages will automatically convert it into a list for you). Not so with `*`. 

A [popular Markdown style guide](http://www.cirosantilli.com/markdown-style-guide/#option-list-marker-hyphen) recommends `-` for list items because `*` can be confused for bold/italic markers. [Other style guides](https://github.com/carwin/markdown-styleguide) also recommend `-` (although I did [find one guide](https://github.com/google/styleguide/blob/gh-pages/docguide/style.md#lists) that _used_ `*` but did not explicitly recommending it).

Based on discussion [on the initial issue](https://github.com/prettier/prettier/issues/3025), there is strong interest in an option for `-`, and even interest in outright changing to `-` entirely. What do others think? Thanks for your consideration.

ikatyang:
Fixes #4251


lipis:
Looks good: [https://deploy-preview-4440--prettier.netlify.com/playground/#N4Igxg9g...](https://deploy-preview-4440--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAqABAGQJYGcYA6UGAZhBERgEYCGATkeugLTpkVROtW4AmJ9Rk3QBqdABtBnYWIC2U4cICMAOnRRsYOEMUAmNbM3bpi1RKM7lauhZOL0IADQgIABxjZouZKHp0IAO4ACvQI3ig0AG4Q2LxOIFR0NGAA1nAwAMquydhQAObIMHQArnDOubhwdDBBSXnyyALilc4AVrgAHgBCSanpGTSycDhQcI00zWUg2XSVdMgg8nQpvIFQ8a42sADqsTAAFsgAHAAMzpsQldtJrgubcHORY850cACOxdivtTT1NOOTZyVQyFEpTXC5PLiOAARWKEHgAJaIBgNCou14B2QumcRRo2HEkIAwhBZA0UFBoM8QMVKgAVNHhJqVAC+LKAA)

duailibe:
@kachkaev Every minor version of Prettier possibly changes the way you format the code. We've done similar stuff in the past, it doesn't warrant a 2.0.

lipis:
What about alternating the symbol for nested lists?

https://github.com/prettier/prettier/issues/4251#issuecomment-379020433 by @infotexture 

```markdown
- Item 1
  * Item 2
    + Item 3
  * Item 4
- Item 4
  * Item 5
```

ikatyang:
It's out of scope of this PR and I don't have strong opinion on that. We should open another issue to see how people think about it.

ashfurrow:
How exciting – thank you!

Issue #4437, opened 2018-05-07
Breaking BinaryExpression in ClassProperty initializer is not indented


suchipi:
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgMoQC2cMAFgJZQDmOwAOlDjpkXAJJSYzpRhwAKAJwgAHOIJgBPHAF4cZcpgB0I4SOUAzCBC07tug9pwAyYwyZ0YC5atHKARukGPnTlwG4GAXwYgANCCiMOTQmMigTsIA7vxOCGEo6ABuEOQAJv4g9oLoYADWJHgiuZRUyDCCAK5wAZSY4jBC6FSE6Mga6Kj1AQBWmAAeAEI5+YXoxAAylHDtnd0gxYL1gshZ6PaSqNCZqpQwAOrpZMgAHAAMAbb1+zkiq6pwy0kzAYJwAI6V5G9NLW1IHS6NRA9UI5HKVWBmFKqDgAEVKhB4LMgQFuPZDmljkgAExonLkVClADCRFaqyg0BeIEq9QAKusEoD6l4vEA)

**Input:**
```jsx
class Something {
  someInstanceProperty = this.props.foofoofoofoofoofoo &&
  	this.props.barbarbarbar;
}

```

**Output:**
```jsx
class Something {
  someInstanceProperty = this.props.foofoofoofoofoofoo &&
  this.props.barbarbarbar;
}

```

**Expected Output:**
Same as input.

This appears to be a regression from 1.11.1.

PR #4442, opened 2018-05-08
Fix: Correctly break class property initializers

nicolo-ribaudo:
This PR moves the `canBreak` check from `printAssignment` to `printAssignmentRight`, so that it can be used in class properties.

Fixes https://github.com/prettier/prettier/issues/4437


suchipi:
Thanks @nicolo-ribaudo!

PR #4445, opened 2018-05-09
(refactor): Move common options out of JS plugin

suchipi:
While seeing where we were at with web, I tried to make a build containing only the markdown plugin, and was surprised to see files from the JS plugin get pulled in. This refactor just makes it a bit more obvious that these settings are shared across plugins.

### DIFF
--- a/src/common/common-options.js
+++ b/src/common/common-options.js
@@ -1,0 +1,22 @@
+"use strict";
+
+const CATEGORY_COMMON = "Common";
+
+// format based on https://github.com/prettier/prettier/blob/master/src/main/core-options.js
+module.exports = {
+  bracketSpacing: {
+    since: "0.0.0",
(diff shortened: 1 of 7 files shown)


suchipi:
Is there a magic comment I can leave to ask travis to re-test this? There's a node 4 failure with no output, which seems to indicate an issue on travis's side.

With Jenkins, you an say "jenkins retest this please", etc

ikatyang:
Just click the `restart job` button?

lipis:
It's already running BTW.. so I guess somebody pressed that button.

suchipi:
I considered making them "global" in the documentation, but they aren't quite global, so it's a bit odd. But once I realized "Global" was a category, I did feel a bit weird introducing "Common", too.

duailibe:
Let's leave them as Common and figure this out later.

PR #4449, opened 2018-05-09
[RFC] New build system

duailibe:
This is a candidate of a new build system for Prettier. It's focused on being "not too difficult" to add new parsers, and automatically output a working web version. It's also heavily based on React's build system.

UNCHANGED:

bin-prettier.js
index.js
CHANGED:

parser-*.js are now UMD modules. We shouldn't rely on any fs, process, path, etc. modules anyway so they're all shimmed in the build. They're all exported as a plugin with only the parsers exposed.
NEW:

standalone.js - UMD build that has no config/ignore/plugin auto-loading features. It has a registerPlugin API and must receive the plugins: [string] object in format() (we could change the API to always auto-load all registered plugins or use @suchipi's of passing plugins: [object] instead of the registerPlugin)
I'm quite happy with this approach. load-plugins is quite ugly now but I think it's a good trade-off.

---
Outdated:
<details>
This doesn't break anything for Node use and just adds a new `web/` folder in the generated package with UMD bundles for Prettier core (`web/index.js`) and our parsers (`web/parser-*.js`):

```
- index.js
- bin-prettier.js
- parser-*.js
- web/
  - index.js
  - parser-*.js
```

This does not output the final form we desire for Prettier but already unlocks a key thing for running Prettier in the browser: lets us ship a web version and we could start using that directly (instead of needing to have a build step). We can update our playground to load Prettier from http://unpkg.com and dogfood the web bundle we ship.

One problem with this is it doesn't work with Node 4, which I don't think it's a problem. We must continue to support 4 users, IMO this means 2 things:

- install from NPM and using Prettier should work

- install from Github and using Prettier should work

The new build system wouldn't affect neither of those points. It does break our current Travis "node 4 + production" test, so the ideal solution would be to build and generate a tarball once, then install and run the tests on the node versions we're interested.. I'm currently investigating a solution for this.

If we do change the playground to load from unpkg, for PRs preview this means we could reuse the solution from above (use a built tarball) and just load in the playground.. i.e. no need for any build steps for a new playground.
</details>


scripts/build/plugins/uglify.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+const { minify } = require("uglify-es");
```
azz: Doesn't `rollup-plugin-uglify-es` do this?

duailibe: I don't know.. we were already using it like this, just moved to a separate file

TrySound: Actually original rollup plugin uglify uses uglify es.

duailibe: To be clear, I just copied what we were doing already in the old `scripts/build/rollup.parser.config.js`

duailibe:
I don't know how to move forward. I was aiming for an API like `babel-standalone` for the web build:
```js
prettier.registerPlugin({ parsers, printers, /* ... */ });
// now we can load externals plugins
```

But that would mean we'd have to preload all the plugins but re-reading #3574 maybe that's not what we want?

Also, for JS that would mean loading Flow and Typescript even if you only wanted Babel.

I'm not sure what I should do

suchipi:
@duailibe I haven't looked at the code yet but I have been thinking of this for a while:
```js
const createPrettier = require("prettier/web/createPrettier");
const plugins = [ /* ... */ ];
const prettier = createPrettier({ plugins });
prettier.format("foo");
```

So for example:
```js
const luaPlugin = require("prettier-plugin-lua");
const prettier = createPrettier({ plugins: [luaPlugin] });
prettier.format("function foo() return 5 end", { parser: "lua" })
```

And we could have a UMD build of the core plugins so that you can do:
```js
const corePlugins = require("prettier/web/core-plugins");
const prettier = createPrettier({ plugins: corePlugins });
prettier.format("function foo() { return 5; }", { parser: "babylon" })
```

And `prettier/web/index.js` could export:
```js
const corePlugins = require("prettier/web/core-plugins");
const prettier = createPrettier({ plugins: corePlugins });
module.exports = prettier;
```

duailibe:
@suchipi that build of the core plugins is likely to weigh almost 10MB :/

duailibe:
I've just had an idea: it's totally possible that a plugin has just the parser, instead of the parser(s) + printer combo.

I'm gonna see if that works

suchipi:
We could also split up the plugins, and provide them each piecemeal instead of one big "core plugins" thing. That line of thinking was what led to https://github.com/prettier/prettier/pull/4445

(Most users will not use JS, CSS, TS, Flow, GraphQL, Markdown, LESS, etc all in the same project)

duailibe:
So I kind of hacked something that actually works for all our cases. We continue to bundle a `parser-*.js` but instead of exposing the `parse()` function, we actually expose with a plugin interface:

```js
// parser-babylon.js
module.exports = {
  parsers: {
    babylon: { parse, hasPragma, locStart, locEnd },
    json: { parse, hasPragma: () => true, locStart, locEnd },
  }
};
```

In `main/options.js`, we now have:

```js
const internalPlugins = [
  require("./language-js"),
  require("./language-*"),
  {
    parsers: {
      // we just have to be careful with this descriptor but it's an easy refactor
      get babylon() {
        return eval("require")("./language-js/parser-babylon").parsers.babylon;
      },
      // others
    }
  }
};
```

For UMD it lets you just do:

```js
prettier.registerPlugin("babylon", prettierParserBabylon /* or other name */);
prettier.registerPlugin("postcss", prettierParserPostcss /* or other name */);
prettier.format(code, {plugins: ["babylon", "postcss]});
// or just automatically use the plugins registered
```

And in our playground you can do:
```js
prettier.registerPlugin("parsers", {
  parsers: {
    get babylon() {
      importScript("https://unpkg.com/prettier/umd/parser-babylon.js");
      return prettierParserBabylon;
    },
    // others
  }
});
```

That satisfies:
- node bundle continues working the same
- UMD bundle lets you register plugins as you wish
- playground still only loads on demand

@azz What do you think?

duailibe:
@azz @suchipi Friendly ping on ☝️

suchipi:
To clarify, you're separating the parsers from the printers because part of the parser API is that you can use the core parsers? Do we need to support that in web? 

Also, rather than registering parsers to strings, I'd rather pass the parsers in directly:
```js
prettier.format(code, {plugins: [prettierParserBabylon, prettierParserPostcss]});
```

But I understand that's a larger refactor.

suchipi:
Also, parsers/printers/plugins are getting blurred a bit here, so I'm still not 100% sure I understand (but I also haven't looked at the code yet, been pretty busy lately 😅)

duailibe:
@suchipi Let me try to explain better :)

First of all, let's remember the main (I think) "constraint" of our build system: only load a parser when we need it (for our Node version and for our playground). And for users of the "web" build, they should be able to only load parsers they'll actually use. So far we've assumed it's OK to load all the code related to a language (even if you don't need it), except for the parsers. We might want to revisit these later but that's how it works now :-)

The API we have now for that is plugins. A plugin consists of: parsers, printers, options and languages specifications. In node we always load the core plugins, but have "glue" code to only load a parser when needed. That doesn't work for the web (and that's why we hack the `require` function for that)... you either have to load the plugin completely (and today our "JS plugin" would have babylon, typescript and flow parsers) or split plugins like "JS-babylon plugin", "JS-typescript plugin", etc.

Another option I envisioned was splitting the JS in:
- plugin with "language specification + printer(s) + options" 
- plugin with only the babylon parser
- plugin with only the typescript parser
- plugin with only the flow parser

Luckily, our plugin infrastructure supports that use case (I don't know if intentionally). Node usage wouldn't change. Playground would be able to lazy load a plugin. Web "regular" users can pick the parsers they want to load.

Now to answer the question directly: I'm separating parsers and printers because we've always wanted to only load parsers when needed. I think it's desirable for the playground

Does that make sense?

j-f1:
@duailibe To allow lazy-loading, why not make `parser` optional and add an optional `getParser` which is called with no arguments and is written `() => require('./parser')`?

duailibe:
@j-f1 Thanks but that's already (kind of) what we do.. the point is how to do that on the web.

There's some context already discussed in #3574

duailibe:
@suchipi (and anyone who's still listening) Here's what I arrived:

UNCHANGED:
- bin-prettier.js
- index.js

CHANGED:
- parser-*.js are now UMD modules. We _shouldn't_ rely on any `fs`, `process`, `path` modules anyway so they're all shimmed in the build. They're all exported as a plugin with only the `parsers` exposed.

NEW:
- standalone.js - UMD build that has no config/ignore/plugin auto-loading features. It has a `registerPlugin` API and must receive the `plugins: [string]` object in `format()` (we could change the API to always auto-load all registered plugins or use @suchipi's of passing `plugins: [object]` instead of the `registerPlugin`)

I'm quite happy with this approach. `load-plugins` is quite ugly now but I think it's a good trade-off.

Gonna fix the playground and this should be ready for merge if this approach is wanted =)

duailibe:
It works! :)

https://deploy-preview-4449--prettier.netlify.com/playground/

azz:
> Luckily, our plugin infrastructure supports that use case (I don't know if intentionally)

Yep it was intentional. Wanted to allow external plugins to provide just printers or just parsers if need be. 

Haven't had a chance to read through the code since the last time I mentioned I did, but just caught up on the thread.

Is it possible to get this to a point where the API looks like this?

Global:
```html
<script src="prettier/standalone"/>
<script src="prettier/parser-graphql"/>
<script>
prettier.format("query { }", { parser: "graphql", plugins: prettierPlugins });
</script>
```

AMD:
```js
define(
  [
    'prettier/standalone',
    'prettier/parser-graphql'
  ], 
  (prettier, ...plugins) => {
    prettier.format("query { }", { parser: "graphql", plugins });
  }
);
```

CJS:
```js
const prettier = require('prettier/standalone');
const plugins = [require('prettier/parser-graphql')];
prettier.format("query { }", { parser: "graphql", plugins });
```

Worker:
```js
importScripts('prettier/standalone');
importScripts('prettier/parser-graphql');
prettier.format("query { }", { parser: "graphql", plugins: prettierPlugins });
```

duailibe:
@azz This is almost where we are now.. CJS and AMD work like this right now.

Global/Worker doesn't yet, I don't know if rollup supports `global.prettierPlugins = global.prettierPlugins || [], global.prettierPlugins.push(factory())` but should be easy enough to write a plugin for that

duailibe:
Also I'll write some integration tests to prevent regression later for the standalone build

azz:
Yeah that's what I was thinking, better push to a single global array than create one global per plugin. Easier to manage.

azz:
Do we have a way to get a list of all parsers that can be loaded? (Will `getSupportInfo()` work before the plugins are loaded?)

duailibe:
@azz since only the parser are deferred, all language options will still be available. And the parsers are still hardcoded in `main/core-options.js`:
https://github.com/prettier/prettier/blob/0b8a8d84f6efb5ff9af9f6f4ba04e49a3dfb5c61/src/main/core-options.js#L84-L117

standalone.js
```diff
@@ -8,40 +8,35 @@ const getSupportInfo = require("./src/main/support").getSupportInfo;
...
-          .map(
-            plugin =>
-              typeof plugin === "string" ? externalPlugins[plugin] : plugin
-          )
-          .filter(Boolean)
-      )
+    let plugins = (args[1] && args[1].plugins) || [];
+    if (!isArray(plugins)) {
+      plugins = Object.values(plugins);
+    }
```
duailibe: @azz We already expose the plugins to `global.prettierPlugins` so this will accept both `{ plugins: { string: [object] } }` and `{ plugins: [object] }`

azz: Makes sense.

scripts/build/rollup-plugins/babel.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
```
azz: https://github.com/rollup/rollup-plugin-babel?

scripts/build/rollup-plugins/executable.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
```
azz: https://github.com/developit/rollup-plugin-preserve-shebang?

scripts/build/util.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+}
+
+function getRelativePath(filepath) {
+  return `./${path.basename(filepath).replace(/\.js$/, "")}`;
+}
+
+function readJson(file) {
+  return new Promise((resolve, reject) => {
+    fs.readFile(file, (error, data) => {
```
azz: Know about [`return util.promisify(fs.readFile)(file)`](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original)?

src/main/parser.js
```diff
@@ -50,16 +62,16 @@ function resolveParser(opts, parsers) {
...
+  // Create a new object {parserName: parseFn}. Uses defineProperty() to only call
+  // the parsers getters when actually calling the parser `parse` function.
   const parsersForCustomParserApi = Object.keys(parsers).reduce(
     (object, parserName) =>
-      Object.defineProperty(
-        object,
-        parserName,
-        Object.getOwnPropertyDescriptor(parsers[parserName], "parse")
-      ),
+      Object.defineProperty(object, parserName, {
```
azz: Nice.

azz:
Noticed the `old-build` stuff, does this mean we'll have a UMD and a CJS-only copy of each parser? If so, is there anything we can do to just ship the UMD build?

scripts/build/rollup-plugins/babel.js
duailibe: I got confused with lots of stuff on the README, just wanted to ship something.. I'll check it out again

scripts/build/rollup-plugins/executable.js
duailibe: Looks like it does the same thing, minus *chmod*ing the file (+x)

duailibe:
No, the current status of this PR is we only have the UMD build of the parsers. The only thing "duplicate" is `index.js` and `standalone.js` which share a lot of code but have different APIs.

The `old-build` is present because we still need to build IIFE bundles for the current released version (which is CJS) for the playground.

scripts/build/rollup-plugins/babel.js
duailibe: Also we should probably use `babel-preset-env` for this but I wanted to make follow-up PRs since this was already too big

azz:
Ah right, so we can delete it after the next patch release, cool.

azz:
Not what I meant. My understanding is the playground uses a version built for a specific release of prettier, so that's why we need `old-build`. If we merge this, then after the next release (no matter what type) we can delete the old build system. Is that correct?

duailibe:
That's correct, I was just confused by "the next patch release".

azz:
Yeah should have just said "next release". Sorry for the confusion!

scripts/build/babel-plugins/transform-eval-require.js
```diff
@@ -0,0 +1,33 @@
+"use strict";
+
+//
+// BEFORE:
+//   __r("path/to/file")
```
suchipi: Since this isn't transforming `eval("require")` but now `__r`, we could rename the plugin.

scripts/build/build.js
```diff
@@ -1,108 +1,230 @@
...
+  const config = {
+    babelrc: false,
+    presets: [],
+    plugins: ["external-helpers"]
+  };
+  if (bundle.type === "core") {
+    config.plugins.push(require("./babel-plugins/transform-eval-require"));
+  }
+  if (bundle.transpile) {
+    config.presets.push(["es2015", { modules: false }]);
```
suchipi: I know we were using es2015 before, but we should change this to use `babel-preset-env` instead of `babel-preset-es2015`, as es2015 is deprecated.

scripts/build/build.js
```diff
@@ -1,108 +1,230 @@
...
+      if (
+        warning.code === "UNRESOLVED_IMPORT" &&
+        bundle.target === "universal"
+      ) {
+        throw new Error(
+          `Unresolved dependency in universal bundle: ${warning.source}`
+        );
+      }
+
+      console.log(warning);
```
suchipi: We might want to use `console.warn` here, so it goes to stderr instead

scripts/build/build.js
```diff
@@ -1,108 +1,230 @@
...
+        json(),
+        nativeShims(path.resolve(__dirname, "shims")),
+        resolve({
+          extensions: [".js", ".json"],
+          preferBuiltins: false
+        }),
+        commonjs(bundle.commonjs || {}),
+        nodeGlobals()
+        // babel(),
+        // uglify()
```
suchipi: Why are these commented out? Should we remove these lines, or do they need to be uncommented?

scripts/build/shims/assert.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  ok() {},
+  strictEqual() {}
```
suchipi: We could make these actually work if we need to- where are they used?

src/common/internal-plugins.js
```diff
@@ -0,0 +1,110 @@
+"use strict";
+
+const __r = require;
```
suchipi: A comment here indicating the babel plugin that looks for this identifier would be good to have.

standalone.js
```diff
@@ -8,40 +8,35 @@ const getSupportInfo = require("./src/main/support").getSupportInfo;
 const internalPlugins = [
   require("./src/language-js"),
   require("./src/language-css"),
-  require("./src/language-handlebars"),
+  // require("./src/language-handlebars"),
   require("./src/language-graphql"),
   require("./src/language-markdown"),
-  require("./src/language-html"),
+  // require("./src/language-html"),
```
suchipi: I'd rather remove the lines than keep commented code. We can add them once we're ready for them

scripts/build/bundles.js
```diff
@@ -0,0 +1,109 @@
...
+  },
+  {
+    input: "src/language-js/parser-typescript.js",
+    target: "universal",
+    minify: true
+  },
+  {
+    input: "src/language-css/parser-postcss.js",
+    target: "universal",
+    bundler: "webpack",
```
suchipi: This is the only bundle that uses webpack. Why doesn't it work with rollup? If we can get it working with rollup, then we won't need to support webpack, which simplifies the build script a lot.

scripts/build/bundles.js
```diff
@@ -0,0 +1,109 @@
...
+    minify: true
+  }
+].map(parser => {
+  const name = getFileOutput(parser)
+    .replace(/\.js$/, "")
+    .split("-")[1];
+  return Object.assign(parser, { type: "plugin", name });
+});
+
+const bundles = [
```
suchipi: Could we add a JSDoc comment explaining the shape of a bundler and its allowed fields? You have to read through a lot of separate files in the build script right now to understand it. I know @ikatyang knows how to make those kind of JSDoc comments (I don't).

scripts/build/build.js
duailibe: Yeah I was gonna do it in a follow up since there are a LOT of changes already, but I guess it doesn't matter much now.

scripts/build/bundles.js
ikatyang: FYI, https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript

scripts/build/build.js
duailibe: The `babel()` is just a reminder to myself that we should probably just use `babel-preset-env` and include `node 4` in the targets

I commented `uglify()` because it takes too long to run, but I wasn't supposed to commit that

scripts/build/shims/assert.js
duailibe: We should probably remove `assert` calls actually. It's used in places like `assert.ok(path instanceof FastPath)`.

Not sure these should go to production anyway

src/common/internal-plugins.js
duailibe: Yeah I just wasn't sure if we should do this or keep `eval("require")`.

scripts/build/bundles.js
suchipi: Thanks

scripts/build/bundles.js
duailibe: Oh I forgot to bring the comment from the old `build/build.js`. postcss has a dependency cycle and doesn't work with rollup. Will add a comment

duailibe:
Thank you for the review!

src/common/internal-plugins.js
suchipi: We could name it `externalRequiredThatGetsTransformedByBabel` 😉

scripts/build/shims/assert.js
suchipi: I think those are currently removed in production, but I agree that we shouldn't rely on assert for those.

suchipi:
This will work with third-party plugins, too, right? You can bundle them yourself and pass them in the `plugins` array?

suchipi:
Also for the third-party plugins case, we might actually want a build that has *no* internal plugins (my understanding is that standalone omits expensive internal parsers, but not their printers)

scripts/build/shims/assert.js
duailibe: They're not removed yet, that's why we need those shims

We should consider doing that in the future

duailibe:
@suchipi I think I addressed all the review comments

> This will work with third-party plugins, too, right? You can bundle them yourself and pass them in the `plugins` array?

(Assuming you mean "work in the browser") Yes! As long as you can load them in the browser, everything should work.

> (my understanding is that standalone omits expensive internal parsers, but not their printers)

This is correct

> Also for the third-party plugins case, we might actually want a build that has *no* internal plugins

Yeah that might be a valid case

duailibe:
Applying `@babel/preset-env` is transforming our custom `Error` classes.. looks like we need this PR https://github.com/babel/babel/pull/7533, still unreleased

duailibe:
@azz @suchipi Let's do this?

j-f1:
Does this mean we can ES6ify constructs throughout the code?

duailibe:
@j-f1 We're not transpiling our own code yet, only the dependencies, I ran into some bugs doing that.. we can probably try it again later

suchipi:
@duailibe sorry, I was asleep, but 🎉

PR #4450, opened 2018-05-09
feat: support JSON.stringify formatting

### LINKED ISSUE #3671: Ignore package(-lock).json by default

ikatyang:
The new parser only matches `package.json` and `package-lock.json` by default.

Similar to `JSON.parse`/`JSON.stringify` but also supports `a -> "a"`/`+1 -> 1`, and it'll throw errors on non-JSON nodes, e.g. comments.

Closes #3671

cc @vjeux


azz:
I mean strict as in doesn't support comments, trailing commas, etc.

vjeux:
This sounds very complicated, why not have parse be JSON.parse and the printer be JSON.stringify?

azz:
Yeah agreed. We want the output to be exactly the same as npm, so `JSON.stringify(object, null, 2)`.

ikatyang:
I believe the output should be same as `JSON.stringify` now.

vjeux:
Makes sense! What do you think about naming it “json-stringify” instead of “json-legacy”?

ikatyang:
Good idea!

src/language-js/parser-json-stringify.js
```diff
@@ -0,0 +1,69 @@
...
+    default:
+      throw createJsonError(node.type);
+  }
+
+  // istanbul ignore next
+  function createJsonError(attribute) {
+    const name = !attribute
+      ? node.type
+      : `${node.type} with ${attribute}=${JSON.stringify(node[attribute])}`;
+    return createError(`${name} is not allowed in legacy JSON.`, {
```
suchipi: The word "legacy" should be changed/removed here

j-f1:
Is it intentional that `{ "foo": bar }` turns into `{ "foo": "bar" }`.

ikatyang:
No, it should throw.

ikatyang:
**Prettier pr-4450**
[Playground link](https://deploy-preview-4450--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEwAEAdEAzCEtJoBGAhgE5oC+IANCBAA4wCW0AzsqOWRAO4AK5BBxQkAbhGYATWiCJkSYANZwYAZQaLmUAObIYZAK5w62tnDIx+CnQFsSybCQA25ugCs2ADwBCC5apqJLZwADLacI4ubiCaZOZkyCCe0AC0bAbaOszYAJ6yDGTaMADq0jAAFsgAHAAMdIUQ5iUKDEmFcAlikXRkcACOhsx91iR2DkhOriYg5rbM+kYzbFnOcACKhhDwUdN0MCREZVKVyABM+wrMzlkAwhC29klQ0D0ghuYAKociU+aUlCAA)
```sh
--parser json-stringify
```

**Input:**
```json-stringify
{ "foo": bar }
```

**Output:**
```json-stringify
SyntaxError: Identifier is not allowed in JSON. (1:10)
> 1 | { "foo": bar }
    |          ^
```

scripts/build/rollup.docs.config.js
```diff
@@ -14,6 +14,6 @@ export default Object.assign(baseConfig, {
   format: "iife",
   plugins: [json(), resolve({ preferBuiltins: true }), commonjs(), globals()],
   useStrict: false,
-  moduleName: basename.replace(/.+-/, ""),
+  moduleName: basename.replace(/.+?-/, "").replace(/-/g, "_"),
```
suchipi: `.+?` can be written as `.*` (you don't have to change it, just an FYI)

vjeux: It's not --exactly-- the same. One is greedy and the other one is not. I don't think it matters in this case though :)

ikatyang: I just want to use its full name as module name so as to avoid conflicts, though it's probably not going to be happened. 😅

- basename: `parser-json-stringify`
- with `.+-`: `stringify`
- with `.+?-`: `json_stringify`

suchipi: @vjeux huh, TIL!

vjeux: @ikatyang `*` means `0 or more`

j-f1: @vjeux `+?` means match as few of the preceding specifier as possible instead of the most possible, while still getting a match.

lydell: Yeah, `?` has different meanings depending on the preceding thing.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-questionmark

PR #4451, opened 2018-05-09
Improve plugin resolution when path does not start with ./

kachkaev:
I noticed that passing `--plugin=path/to/plugin` rather than `--plugin=./path/to/plugin` crashes Prettier. This happens because it tries to resolve `path/to/plugin` as a node module rather than a relative path. The error says:

> "Cannot find module 'path/to/plugin' from '/path/to/cwd'"

This PR fixes that and thus makes usage of `--plugin` more predictable. ~~~It is unlikely that someone would want to refer to a local node module, because a plugin from there would be autoloaded anyway (as long as the instance of Prettier is local).~~~ The new behaviour supports both node module names and paths without leading `./`.

I also tweak a couple of minor things given the opportunity, please see the commit log.


azz:
What about if someone wants to use a plugin that doesn't match the auto-load pattern? e.g. `prettier --plugin=@company/prettier-plugin-bespoke`?

kachkaev:
Please note  that I'm keeping `requirePath: resolve.sync(pluginName, { basedir: resolvedPluginSearchDir })` for [plugins found in `pluginSearchDirs`](https://github.com/prettier/prettier/pull/4451/commits/fad9cbb062b33faa9253290f709b0f1ed14c6aa5#diff-9e0934c7a0ce8de1ef8abc92bfe3ad6dR64). That's because there we get a list of node modules there and we do want to resolve them as modules.

kachkaev:
@azz good point. I guess we should decide between allowing people to use `--plugin=@company/prettier-plugin-bespoke` instead of `--plugin=node_modules/@company/prettier-plugin-bespoke` and saving others from confusion when whey type `--plugin=src/index.js` instead of `--plugin=./src/index.js`. I personally struggled in the second scenario, which brought me to proposing this change.

Practically speaking, how likely would people use `@company/prettier-plugin-bespoke` for Prettier plugin names given that these are not auto-loadable and so more difficult to setup in projects?

azz:
Maybe we can just offer a better error message if you forget the `./`?

suchipi:
We could search locally if the provided string ends in `.js`, and assume node modules otherwise

suchipi:
We may also want to support the string ending in `.`:
`--plugin=.`

suchipi:
That's a clean solution, I like it :+1:

kachkaev:
@suchipi that's node now, just waiting for the tests to pass (Travis builds are a bit bumpy today).

PR #4453, opened 2018-05-09
Simplify `build-docs.js`

duailibe:
Leverage use of unpkg.com instead of manually copying stuff to `static/`

This is motivated by #4449 to simplify build processes

### DIFF
--- a/website/static/service-worker.js
+++ b/website/static/service-worker.js
@@ -3,7 +3,7 @@
 
 "use strict";
 
-importScripts("lib/sw-toolbox.js");
+importScripts("https://unpkg.com/sw-toolbox@3.6.0/sw-toolbox.js");
 
 toolbox.precache([
   // Scripts
@@ -16,7 +16,7 @@ toolbox.precache([
   "lib/parser-graphql.js",
   "lib/parser-markdown.js",
   "playground.js",
-  "lib/sw-toolbox.js",
+  "https://unpkg.com/sw-toolbox@3.6.0/sw-toolbox.js",
 
   // CodeMirror
   "https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.36.0/codemirror.css",
--- a/scripts/build/build-docs.js
+++ b/scripts/build/build-docs.js
@@ -68,15 +68,6 @@ shell.cd("website");
 shell.echo("Building website...");
 shell.exec("yarn install");
 
-shell.echo("Copy sw-toolbox.js to docs");
-shell.cp("node_modules/sw-toolbox/sw-toolbox.js", `${docs}/sw-toolbox.js`);
-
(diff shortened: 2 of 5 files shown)


website/static/service-worker.js
```diff
@@ -3,7 +3,7 @@
 
 "use strict";
 
-importScripts("lib/sw-toolbox.js");
+importScripts("https://unpkg.com/sw-toolbox@3.6.0/sw-toolbox.js");
```
j-f1: This might not work — the same-origin policy also applies to `importScripts`.

duailibe: I tested on Chrome, macOS

lipis: Does it work here? https://deploy-preview-4453--prettier.netlify.com/

duailibe: Yes of course

image

duailibe: Worst thing that can happen is it'll stop working offline

lipis: Seems legit! :D

j-f1:
@duailibe Why did you remove the PR from 1.13?

duailibe:
It's just a change in a build script.. why should it be in the milestone?

duailibe:
I could also ask: @lipis Why are you adding "internal" PRs to the milestone?

cc @j-f1

lipis:
Because I believe that every PR belong to this version and I thought it would be a good idea to have all of them in one place. It doesn't change much and it's just a label in the end of the day. 

Makes more sense for future releases like 2.0 to have only the main changes. But someone could also argue that even a complete redesign of the website or the docs or the build system could have a specific milestone.. it's all part of the project.

lipis:
Mostly for the history on what exactly it was merged into the project during that version.. the actual changelog of what happened to Prettier can be found on the blog anyways.

lipis:
https://github.com/prettier/prettier/milestone/16?closed=1 I think it's OK.. but then again it's just my opinion :)

Plus it's not like bunch of people going there or anything.. it's just a reference point mostly for us! No harm done :)

duailibe:
Ok. I've been using milestones as a way to track what's left to do for a new release. And to check what merged between versions I just compare tags: https://github.com/prettier/prettier/compare/1.12.1...master but whatever, I just removed because I thought it was oversight, feel free to continue to use them if you want to :)

lipis:
Sure what's left is still there in a very clear view and that's the main reason to use Milestones. 

I'm also using the compare tags, but there is more noise there compare to the way they show up in the [milestone](https://github.com/prettier/prettier/milestone/16?closed=1). In the compare tags view, in order to go the the actual PR you have to click on the PR number and not the title. 

Anyways.. all good! Let's stop annoying the rest with our _OCD nonsense_ :)

PR #4454, opened 2018-05-09
Fix playground

duailibe:
After #4341, because of `find-parent-dir`, the playground from master was broken

This is the kind of thing that makes me wish for #4449


duailibe:
https://deploy-preview-4454--prettier.netlify.com/playground/ works

Issue #812, opened 2017-02-25
Parentheses around single parameter arrow functions / Add support for arrow-parens property


DavidBabel:
Hey ! 

It will be great to add an option to disable auto remove of parenthesis when there is only one argument in an arrow function.
Most of professional projects configure their way since it's the default eslint behaviour.
Here is a Eslint Link : 
http://eslint.org/docs/rules/arrow-parens

```
// Bad
a => {}

// Good
(a) => {}
```

andreypopp:
I think this is needed — always presenting parens make refactoring easier: when you need to add another arg you don't need to add parens (cause they are present already).

DavidBabel:
Personnally i found the real problem is in the spec, where only the "one args" arrow function allow removing parenthesis. It's handy when you code very small applications, but it's not at all when you need modify your prototype later.

```
() => {}             // ok
   => {}             // not allowed
(arg1) => {}         // ok
 arg1  => {}         // ok
(arg1, arg2) => {}   // ok
 arg1, arg2  => {}   // not allowed
```

If it's not configurable in prettier and the default configuration of prettier is opposite with default behaviour of **ESlint**. It makes **prettier** unusable for existing projects unless changing all project **ESlint** rules,  and it's a huge problem on multi users projects.

vramana:
Except in extreme cases, we are not going to add extra configuration options. The question then is can we make parenthesis around arrow function arg the default. Here is @jlongster's reply in another thread in case you missed it.

> Single-arg arrow functions are also extremely common and I don't think it's worth adding parens just to make it easier to add args later. It's really nice to just have arr.map(x => ...).

https://github.com/prettier/prettier/issues/572#issuecomment-277049560

Therefore I am closing this issue.

jlongster:
I realize this is one of those things where you can't please both sides. :( Many people would not like this change. We try to make decisions based on what a majority of people want, not just based on our own personal styles, but sometimes "majority" is hard to define.

rhengles:
@DavidBabel I just added this option in https://github.com/arijs/prettier-with-tabs/commit/0f918465dc1d16f8826268c658f55a27af7eb963

DavidBabel:
@rhengles awesome

androa:
[...]

Issue #40, opened 2017-01-10
Resist adding configuration


gaearon:
You’re going to be under a lot of pressure to make X, Y, and Z, configurable.
Configuration is a perfect example of [tragedy of the commons](https://en.wikipedia.org/wiki/Tragedy_of_the_commons).

Everybody wants to add just a tiny piece of configuration to satisfy their use case. As a result we end up with inconsistently designed tools, with unexpected combinations that don’t work correctly, with bugs and regressions that, when fixed in one configuration, break something else with another configuration.

The requests for configuration are vocal. The requests to stay simple are often silent.

I just wanted to take a moment to say there are some of us who actually appreciate tools with limited scope that don’t attempt to solve everybody’s use case, and instead do a specific thing well.

sapegin:
If we as a community could find the most common JavaScript code style, I’d try to adopt it in my projects. I see a lot of benefits in that for me as a maintainer and for contributors of my projects.

gaearon:
The thing is, I don’t even care about “most common” here as much. I’m fine with @jlongster making common sense decisions as long as it’s something consistent and as bug-free as possible. These style differences don’t really matter, people care way too much about them.

sapegin:
The only reason for the most common style is to make less people adapt to a new one ;-)

gaelduplessix:
@gaearon Agree that style differences don't really matter and any crazy style that has auto-format support is better than one without.

However, for a project that **is about formatting** I think style questions should be taken seriously. If it's about defining a "standard" that would be used by many people, would be better to chose one that looks nice (to as many people as possible) :)

jlongster:
@gaearon I totally agree. Thanks for opening this issue because we need to discuss here what the real goal of the project is (and put it in the readme). I see three different goals:

1. Converge on generally a single style (with *very* few basic options to help align with existing projects, like single/double quotes)
2. Cater to a few large demographics (support tabs or spaces, semi or no-semis)
3. Make it completely configurable/pluggable, possibly reading in eslint config and trying to adhere to that

I definitely don't want to do #3. There's just too much complexity already in making sure everything is consistent and I guarantee I will burn out (probably others too) if we attempt that and this project won't work.

[...]

PR #4456, opened 2018-05-10
Introduce 'ternary' config and extend 'arrowParens' config with 'requireForBlockBody' option

netoctone:
Hello,

What do you think of adding these configs in order to make prettier more flexible?

By the way, it addresses some of the wishes of prettier users (an alternative to changing eslint rules):
https://github.com/prettier/prettier-atom/issues/329
https://github.com/prettier/prettier/issues/812

Cheers,

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -98,6 +98,19 @@ Put the `>` of a multi-line JSX element at the end of the last line instead of b
 | ------- | ------------------------- | ---------------------------- |
 | `false` | `--jsx-bracket-same-line` | `jsxBracketSameLine: <bool>` |
 
+## Ternary
+
+Control ternary expressions.
+
+Valid options:
+
+* `"newline"` - place both expressions on new lines
+* `"parens"` - wrap expressions in parentheses
+
+| Default     | CLI Override                                 | API Override                                  |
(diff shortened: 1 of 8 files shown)


netoctone:
@j-f1 it's definitely not an option request, since it's an option implementation

netoctone:
@evilebottnawi it's compatible with eslint rule
` 
'arrow-parens': ['error', 'as-needed', {
      requireForBlockBody: true,
    }]
`

evilebottnawi:
@netoctone You can use prettier first and when use `eslint` for fix it, why not use this workflow?

netoctone:
@evilebottnawi it's less performant and more convoluted. Obviously solves the issue but seems a temporary solution.

suchipi:
We aren't interested in adding more options for things like this. The code looks  great but it's unfortunately not in line with our project goals, so I'm gonna close this PR. However we would love to receive more PRs from you in the future, but opening an issue first to discuss the proposed change before implementing it is the best approach.

j-f1:
You might be interested in checking out #40 and [our option philosophy](https://prettier.io/docs/en/option-philosophy.html) for more information on why we try to avoid adding options.

PR #4392, opened 2018-04-27
Fix: Always preserve frontmatter, remove dependency

eliasmeire:
👋 

This fixes #4390, fixes #4162

PR #4457, opened 2018-05-10
Fix: lazy match the frontmatter in postcss parser

eliasmeire:
Small follow up PR for #4392.

Old regex would match eagerly up to the third `---`. Fixes following edge case:

**Input**

```
---
foo: bar
---

a {
color:blue
}

---

.b {
color:red
}
```

**Old output**

```
---
foo: bar
---

a {
color:blue
}

---

.b {
  color: red;
}
```

**New output**

```
---
foo: bar
---

a {
  color: blue;
}
--- .b {
  color: red;
}
```

Added this as a test case, thanks to @jonschlinkert for catching this in #4392.


duailibe:
Thanks!

PR #4462, opened 2018-05-11
New primitive for whether parent groups broke

helixbass:
I've been working on a plugin for Coffeescript and came across some formatting needs that I believe aren't achievable with the current set of formatting primitives

For background, Coffeescript allows omitting function call parens and object braces in a lot of instances. And an opinionated Coffeescript formatter should (in my opinion) omit them in most cases when it's allowed. They can usually be omitted at the end of a line, so when a construct like a call or array breaks its arguments/elements onto separate lines, those arguments/elements can safely omit braces/parens. To illustrate:
```
[a(b), c] # Need parens in a(b) to avoid parsing [a b, c] as [a(b, c)]
[
  a b # Omitting parens is fine here
  c
]
```
I didn't see a way to achieve this type of rule using existing primitives. Since eg in this example the inner call needs to know whether the enclosing array broke or not to decide whether it's safe to omit parens

So I added a concept of "visible" parent/ancestor groups. Here's the "API" I've added so far:
- a new `visible` option to `group()` that declares this group to be one that's "visible" to descendants (as far as whether it broke or not)
- a new `ifVisibleGroupBroke()` primitive that like `ifBreak()` takes a `breakContents` and `flatContents`. But instead of choosing based on whether the current group broke, it chooses based on whether an enclosing "visible" group broke (defaults to the first enclosing "visible" group but takes a `count` option that specifies the offset up the visible group ancestor stack - in most of the use cases so far for Coffeescript, I'm passing `{count: 1}` since eg for `[{a: 1}, {b: 2}]` the `ifVisibleGroupBroke()` needs to be inside the visible `group()` for the objects but needs to check if the array's visible group broke)
- a `shouldBreakIfVisibleGroupBroke` option to `group()` (that similarly takes an optional eg `{count: 1}` as its value, otherwise `count` defaults to `0` (ie immediately enclosing visible group) if you just pass `shouldBreakIfVisibleGroupBroke: true`. I'd naively thought I could do something like `ifVisibleGroupBroke(breakParent)` for the use case where I wanted an object argument to a call to itself break if the call args break ie do this:
```
f(
  a
  b: 1
  c: 2
)
```
instead of
```
f(
  a
  b: 1, c: 2
)
```
But found that `breakParent` isn't "dynamic" like that, so `shouldBreakIfVisibleGroupBroke` handles this use case
- a new `blockVisible` option to `concat()`, which "blocks"/hides visibility of outer visible groups for everything inside the `concat()` (possible that you could need this exposed on something other than `concat()` as well, or I guess you could model it as its own wrapping primitive eg `blockVisibleGroups(concat(...))`). Used for telling non-final methods of a method chain that they shouldn't omit parens if the chain doesn't break. It's possible this one isn't strictly necessary but I haven't fully wrapped my head around it. And seems like a useful concept

It's been working quite well so far for adding these types of rules to the Coffeescript printer. There may be some hairier eg nested cases that I haven't covered but visible groups has so far proven to be a useful and sufficient mechanism for smartly omitting parens/braces eg inside call args, arrays, objects and method chains. If you're curious to see examples (or of the overall progress of the Coffeescript plugin) check out my [prettier-plugin-coffeescript](https://github.com/helixbass/prettier-plugin-coffeescript) repo, which uses my [prettier](https://github.com/helixbass/copheescript/tree/prettier) Coffeescript branch to generate Coffeescript AST (original point of that branch was for the compiler backend to target the `babylon` AST and then be able to use Prettier as a toolchain backend to generate compiled JS, which actually works well but that's a separate topic - but now am leveraging that work to be able to generate Coffeescript AST nodes that are as close as possible to their corresponding `babylon` AST types, which has allowed me to effectively reuse/copy a lot of code from `printer-estree.js`)

I've tried to keep the implementation somewhat clean while avoiding incurring any real additional cost for docs that don't use this feature. The mechanism I've used for keeping track of visible outer groups is to add a `doc.visibleGroups` "stack" when necessary. I don't know if it seems weird/dirty to mutate `doc` inside `doc-printer.js` like that (especially with state that assumes that it starts from scratch and only gets used for a single pass over the doc), but it seemed easier (and less presumptuous) than eg adding new fields to the `cmds` items

Modeling `visibleGroups` as a stack has sufficed for all the cases I've encountered in Coffeescript so far, but conceptually you could picture tagging visible groups with "types" eg `group(..., {visible: {type: 'call'}})` and allowing `ifParentGroupBroke()` to specify a particular type of visible group that it's interested in

The code should be considered WIP. But if this seems viable and necessary for these formatting needs, I'd appreciate pointers on how I should go about adding tests for it and/or suggestions re the implementation details

### DIFF
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -68,6 +68,11 @@ function group(contents, opts) {
   if (opts.visible) {
     ret.visible = !!opts.visible;
   }
+  if (opts.shouldBreakIfVisibleGroupBroke) {
+    ret.breakIfVisibleGroupBroke = {
+      count: opts.shouldBreakIfVisibleGroupBroke.count || 0
+    };
+  }
   return ret;
   // blockVisible: !!opts.blockVisible
 }
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -52,7 +52,8 @@ function group(contents, opts) {
     type: "group",
     contents: contents,
     break: !!opts.shouldBreak,
-    expandedStates: opts.expandedStates
+    expandedStates: opts.expandedStates,
+    visible: !!opts.visible
   };
 }
 
@@ -94,6 +95,19 @@ function ifBreak(breakContents, flatContents) {
   }
 
   return { type: "if-break", breakContents, flatContents };
+}
+
+function ifVisibleGroupBroke(breakContents, flatContents, { count = 0 } = {}) {
+  if (process.env.NODE_ENV !== "production") {
+    if (breakContents) {
+      assertDoc(breakContents);
+    }
+    if (flatContents) {
+      assertDoc(flatContents);
+    }
+  }
+
+  return { type: "if-visible-group-broke", breakContents, flatContents, count };
 }
 
 function lineSuffix(contents) {
@@ -159,6 +173,7 @@ module.exports = {
   cursor,
   breakParent,
   ifBreak,
+  ifVisibleGroupBroke,
   indent,
   align,
   addAlignmentToDoc,
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -11,7 +11,9 @@ function assertDoc(val) {
   }
 }
 
-function concat(parts) {
+function concat(parts, opts) {
+  opts = opts || {};
+
   if (process.env.NODE_ENV !== "production") {
     parts.forEach(assertDoc);
   }
@@ -22,7 +24,14 @@ function concat(parts) {
   //   // If it's a single document, no need to concat it.
   //   return parts[0];
   // }
-  return { type: "concat", parts };
+  const ret = {
+    type: "concat",
+    parts
+  };
+  if (opts.blockVisible) {
+    ret.blockVisible = !!opts.blockVisible;
+  }
+  return ret;
 }
 
 function indent(contents) {
@@ -48,13 +57,19 @@ function group(contents, opts) {
     assertDoc(contents);
   }
 
-  return {
+  const ret = {
     type: "group",
     contents: contents,
-    break: !!opts.shouldBreak,
-    expandedStates: opts.expandedStates,
-    visible: !!opts.visible
+    break: !!opts.shouldBreak
   };
+  if (opts.expandedStates) {
+    ret.expandedStates = opts.expandedStates;
+  }
+  if (opts.visible) {
+    ret.visible = !!opts.visible;
+  }
+  return ret;
+  // blockVisible: !!opts.blockVisible
 }
 
 function dedentToRoot(contents) {
--- a/src/doc/doc-printer.js
+++ b/src/doc/doc-printer.js
@@ -277,6 +277,17 @@ function printDocToString(doc, options) {
 
           break;
         case "group": {
+          if (doc.breakIfVisibleGroupBroke) {
+            const { count } = doc.breakIfVisibleGroupBroke;
+            if (
+              doc.visibleGroups &&
+              doc.visibleGroups.length >= count + 1 &&
+              doc.visibleGroups[doc.visibleGroups.length - (1 + count)] ===
+                MODE_BREAK
+            ) {
+              doc.break = true;
+            }
+          }
           switch (mode) {
             case MODE_FLAT:
               if (!shouldRemeasure) {
--- a/src/doc/doc-printer.js
+++ b/src/doc/doc-printer.js
@@ -173,6 +173,24 @@ function fits(next, restCommands, width, options, mustBeFlat) {
           }
 
           break;
+        case "if-visible-group-broke":
+          const { count } = doc;
+          if (
+            doc.visibleGroups &&
+            doc.visibleGroups.length >= count + 1 &&
+            doc.visibleGroups[doc.visibleGroups.length - (1 + count)] ===
+              MODE_BREAK
+          ) {
+            if (doc.breakContents) {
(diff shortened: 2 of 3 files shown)


duailibe:
This is interesting. We do have a few places where this might be useful in JS.

I'll experiment with this in the next 3 or 4 days and give some feedback :)

suchipi:
Although this is perfect for Coffeescript, I feel like it's a bit too vague for most cases- usually you need to be very aware of which parents you're within to know what you're allowed to do.

It seems like you'd be able to achieve the formatting you want with conditionalGroups or willBreak, wouldn't you?

suchipi:
In this example:
```coffee
f(
  a
  b: 1
  c: 2
)
```
It seems like the function parameters could use willbreak to decide whether to join with commas or not, and the object literal could use ifbreak to omit its commas

helixbass:
@suchipi I'm almost positive this is unachievable using existing mechanisms eg `conditionalGroup()`/`willBreak`. For one thing, from my understanding, `willBreak` is "static" ie unaware of whether it breaks due to line length. And even if some very basic cases could be detected using existing mechanisms, I think it would be very painful and awkward as it'd have to be the "parent" (the thing that may or may not break, eg the call) that would have to try and control the formatting of its various types of children based on whether it breaks or not

As opposed to the children having the ability to render themselves differently based on whether the "visible" parents broke or not. This technique has proven itself able to handle rather sophisticated cases as I've been using it in the Coffeescript plugin

> usually you need to be very aware of which parents you're within to know what you're allowed to do.

Yes perhaps, so far in Coffeescript most of the cases simply need to know whether they've been broken out onto a separate line (regardless of the type of the parent). But I did mention that you could imagine tagging visible groups with eg "types". @duailibe I'm curious what are some potential use cases for JS?

helixbass:
@duailibe no worries, let me know if you have any questions when you get a chance to play around with it. Or if you can describe any JS use cases, I can try and see how visible groups could be applied. Eg when I started working on it, I thought perhaps there was already a mechanism because I saw this in the 1.12 release blog post re #4267:
> any nested non-leaf object or array patterns in a destructuring pattern will get multi-lined if any of their parents do

But really what it's doing is explicitly breaking nested patterns (forcing the parents to break). I believe that with visible groups, it'd be straightforward to still allow the whole thing not to break if it fits, eg:
```
const {a, b: {c, d}} = e
```
while enforcing the desired behavior that if the outer pattern breaks, the inner ones should as well

> I think @suchipi meant ifBreak() instead

The point is still the same, that trying to use `ifBreak()` (instead of `ifVisibleGroupBroke()`) you'd still have to try and do contortions to have the parent control the rendering of different children itself, rather than allowing children to flexibly set their own rules with an awareness of whether visible parent groups broke or not

rattrayalex:
My hunch is that conditionalGroup and ifBreak, perhaps with the contortions you mention, may be the way to go. 

Even in the coffeescript case, I'm not sure that doing this from the parent level is bad necessarily... Eg; should it ever really go without parens if the parent is a binop?

rattrayalex:
Hmm ok yeah, poking around that seems reasonable. I think I would have liked something more like this when working on jsx. Haven't looked at the implementation though. 

Others, thoughts?

j-f1:
Something seems to be badly broken.

helixbass:
@j-f1 broken how? `yarn test` seems ok against the updated (`master`-merged) version of this branch. And the Coffeescript plugin that relies on this functionality seems ok after upgrading its `prettier` dependency to include the latest version of this branch

j-f1:
@helixbass [Circle CI](https://circleci.com/gh/prettier/prettier/4119?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) seems to have lots of errors. Would you be able to push the working version to this PR?

rattrayalex:
Looks like the prettier plugin for coffeescript, which relies on this branch, just shipped (congrats @helixbass !)

Would be great for us to review and merge. I'll try to take a look this weekend but haven't been active in prettier lately, so if someone else is available that might be better.

eugenet8k:
It would be great to finally have first-class support for CoffeeScript in Prettier!

eugenet8k:
@helixbass can this be merged assuming the release of Prettier v2?

Kwaadpepper:
Please continue the good work, coffee script needs more support. :+1:

MentalGear:
I support this pull request to honor @helixbass tireless work to allow a hack-free coffeescript integration into prettier. Please accept the pull.

helixbass:
@eugenet8k it's not a breaking change so afaik it doesn't matter when it gets merged

The status is that noone on Prettier core ever really stepped up and reviewed this

My hunch is that noone since @jlongster has felt comfortable deeply reasoning about the core printing algorithm 🤷‍♀️ 

But I'd love to be proven wrong on that

Because I think what it deserves is a thoughtful conceptual examination. There's a deep idea there about having visibility of whether you're inside parent groups that broke or not. I came up with an API that worked for the use cases I encountered in Coffeescript. But I still have a hard time reasoning (on my own) about how it generalizes

eugenet8k:
@helixbass I guess I feel your pain because it's my pain as well. It seems surely it would be great for Prettier to have one more legendary language in the list of supported ones. Especially assuming that all the technical work is done (thank you so much!). I hope the Prettier's gods will look down and give their blessing to this PS 🤞.

aminya:
Is there any update on this?

4086606:
I'd really appreciate something like this in Prettier. I could compile JSONC to CSON format, enforcing a consistent quoting style for users of my future VS Code extension.

PR #4463, opened 2018-05-12
Support new node types for optionals

xixixao:
Once you want to bump babylon, you'll need something like this. The node types have changed in https://github.com/babel/babel/pull/7288 in Feb.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -90,12 +90,16 @@ function genericPrint(path, options, printPath, args) {
         node.type !== "ClassMethod" &&
         (decorator.type === "Identifier" ||
           decorator.type === "MemberExpression" ||
+          decorator.type === "OptionalMemberExpression" ||
           (decorator.type === "CallExpression" &&
+            decorator.type === "OptionalCallExpression" &&
             (decorator.arguments.length === 0 ||
               (decorator.arguments.length === 1 &&
                 (isStringLiteral(decorator.arguments[0]) ||
                   decorator.arguments[0].type === "Identifier" ||
-                  decorator.arguments[0].type === "MemberExpression")))))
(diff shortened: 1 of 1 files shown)


src/language-js/printer-estree.js
```diff
@@ -4122,12 +4135,9 @@ function printClass(path, options, print) {
 
 function printOptionalToken(path) {
   const node = path.getValue();
-  if (!node.optional) {
-    return "";
-  }
```
xixixao: I guess this should stay for other parsers

src/language-js/printer-estree.js
```diff
@@ -90,12 +90,16 @@ function genericPrint(path, options, printPath, args) {
         node.type !== "ClassMethod" &&
         (decorator.type === "Identifier" ||
           decorator.type === "MemberExpression" ||
+          decorator.type === "OptionalMemberExpression" ||
           (decorator.type === "CallExpression" &&
+            decorator.type === "OptionalCallExpression" &&
```
j-f1: This won’t work — the node can’t be both types at the same time.

src/language-js/printer-estree.js
duailibe: Yeah there are other types of optional nodes like optional class properties in TypeScript.

duailibe:
We'll need this when upgrade Flow too.

@j-f1 We can't add any tests for this until we upgrade Flow and/or Babylon (which we might no be able to upgrade because of Node 4 support).

xixixao:
I suggest you guys add the newer parsers as separate (not sure if real option or just for tests), if you want to support Node 4. I agree this should go out with exercised tests.

vjeux:
If babel now requires node 4, we should take the plunge... It's sad but I believe it now passes the threshold of being too hard to maintain.

PR #4469, opened 2018-05-12
Fix: Don't run prettier twice for --list-different

### LINKED ISSUE #4345: --list-different twice as slow as --write on large codebase
`--list-different` runs are twice as slow than `--write` runs on a very large codebase (~2650 files). This is counterintuitive to me since I would expect `--write` to have do more work. Is this a known issue? I can't find any related issues or docs explaining that `--list-different` takes a performance hit.

```
$ yarn prettier -v
1.12.1
```

```
$ prettier --write --loglevel silent "{apps,libs,tools}/**/*.js"
✨  Done in 43.48s.
```

```
$ prettier --list-different --loglevel silent "{apps,libs,tools}/**/*.js"
✨  Done in 85.87s.
```

eliasmeire:
fixes #4345.

Found out that if prettier is called with --list-different, prettier is formatting each file twice.
Once by calling `prettier.check` and once by calling `prettier.format`. Not sure if this is the best way to solve it but it works 👍 any feedback welcome.

Also maybe I should put this elsewhere, but this code is pretty confusing with all the checking on `--write` and `--list-different`, mainly because both flags can be set. Wouldn't it make more sense to have subcommands instead? e.g. `prettier check` and `prettier write`


eliasmeire:
Sure 👍

tests_integration/runPrettier.js
```diff
@@ -101,7 +104,8 @@ function runPrettier(dir, args, options) {
     jest.restoreAllMocks();
   }
 
-  const result = { status, stdout, stderr, write };
+  const formatCalls = formatSpy.mock.calls.length;
+  const result = { status, stdout, stderr, write, formatCalls };
```
duailibe: What about returning below:

```js
return Object.assign({ test: testResult, formatCalls }, result);
```

and then write one or more tests that assert specifically the number of format calls:

```js
const result = runPrettier(args);
expect(result.formatCalls).toEqual(0);
```

WDYT?

eliasmeire: Sounds good to me! I'll check it out tomorrow :)

eliasmeire:
So @duailibe it's working great in development but not in `NODE_ENV=production`. I think it's because in prod i'm spying on the exported `formatWithCursor` from dist/index.js but in prod the `formatWithCursor` that is being used is bundled with the cli in dist/bin-prettier.js and I don't think there is a way to spy on that. Any ideas?

duailibe:
That's strange we shouldn't be bundling `index.js` in `bin-prettier.js`. I can't debug it now but I'll try to find some time tomorrow to see what's happening.

j-f1:
Is this ready to merge?

duailibe:
I think we ended up fixing this on one of the latest refactors.. Sorry we took this long @eliasmeire, but I was focused on some big refactors..

Thanks for the contribution anyway.. It was great!

PR #4475, opened 2018-05-13
Wrap ?() => T in parens when needed.

### LINKED ISSUE #4461: Prettier changes precedence on flow arrow functions
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#.....)
Prettier removed the parens around this input code:
```jsx
type Foo = { arg: ?(() => void) | string };
```

causing this output:
```jsx
type Foo = { arg: ?() => void | string };
```

The output is actually a different behavior due to the precedence of the nullable operator. The output is equivalent to the following which demonstrates the difference in behavior from the original input.
```
type Foo = { arg: ?() => (void | string) };
```

nicolo-ribaudo:
Fixes #4461

### DIFF
--- a/tests/flow_function_parentheses/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow_function_parentheses/__snapshots__/jsfmt.spec.js.snap
@@ -97,6 +97,21 @@ interface F {
 }
 
 type ExtractType = <A>(B<C>) => D
+
+type T = ?(() => A);
+
+type T = ?(() => A) | B;
+
+type T = ?() => A | B;
+
+type T = (?() => A) | B;
+
+// https://github.com/babel/babel/issues/7924
+//type T = ??() => A;
+
+type T = ?(?(() => A));
+
+type T = ?(?() => A) | B;
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 type Banana = {
   eat: string => boolean
@@ -141,6 +156,21 @@ interface F {
 }
 
 type ExtractType = <A>(B<C>) => D;
+
+type T = ?() => A;
+
+type T = ?(() => A) | B;
+
+type T = ?() => A | B;
+
+type T = ?(() => A) | B;
+
+// https://github.com/babel/babel/issues/7924
+//type T = ??() => A;
+
+type T = ??(() => A);
+
+type T = ??(() => A) | B;
 
 `;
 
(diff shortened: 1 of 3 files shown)


tests/flow_function_parentheses/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -190,6 +220,21 @@ interface F {
 }
 
 type ExtractType = <A>(B<C>) => D
+
+type T = ?(() => A);
+
+type T = ?(() => A) | B;
+
+type T = ?() => A | B;
```
j-f1: Would it make sense to explicitly disambiguate this to `?() => (A | B)`?

tests/flow_function_parentheses/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -235,6 +280,21 @@ interface F {
 
 type ExtractType = <A>(B<C>) => D;
 
+type T = ?() => A;
+
+type T = ?(() => A) | B;
+
+type T = ?() => A | B;
```
j-f1: Would it make sense to explicitly disambiguate this to `?() => (A | B)`?

j-f1: It’s not clear to me, but I don’t use Flow so I’ll leave the decision up to those who do.

nicolo-ribaudo: If we end up deciding to wrap it in parentheses, I'd prefer to wrap the whole function. e.g.
```js
type T = ?(() => A | B)
```

This is consistent with how JS functions work:
```js
+(() => a || b);

+() => a || b; // SyntaxError
```

hawkrives: My personal anecdote: I can't remember the operator precedence for `|`, so I actually thought that type was `(?() => A) | (B)`. So parenthesis would be helpful for me.

nicolo-ribaudo: I have a local implementation which puts the parentheses around `A | B` in `() => A | B`, but some snapsot tests look a bit strange to me.

e.g.
```diff
  then<U>(
-    onFulfill?: ?(value: R) => Promise<U> | ?U,
-    onReject?: ?(error: any) => Promise<U> | ?U
+    onFulfill?: ?(value: R) => (Promise<U> | ?U),
+    onReject?: ?(error: any) => (Promise<U> | ?U)
  ): Promise<U>;
```

Can we merge this PR as-is (since it fixes an actual bug with prettier, which changes the meaning of the code) and discuss about `() => A | B` in a new issue?

duailibe: > discuss about `() => A | B` in a new issue?

Agreed

duailibe:
Thank you!

PR #4478, opened 2018-05-14
refactor: parser-postcss

evilebottnawi:
Description:
1. Use `*-unknown` for all what we can't parse and output this as is
2. `try/catch` in `parse*` functions

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -249,15 +249,9 @@ function genericPrint(path, options, print) {
       );
     }
     case "media-unknown": {
-      return adjustStrings(node.value, options);
+      return node.value;
     }
     // postcss-selector-parser
-    case "selector-root-invalid": {
-      // This is likely a SCSS nested property: `background: { color: red; }`.
-      return adjustNumbers(
-        adjustStrings(maybeToLowerCase(node.value), options)
-      );
-    }
     case "selector-root": {
       const atRuleAncestorNode = getAncestorNode(path, "css-atrule");
       const insideAtRuleNode =
@@ -367,6 +361,18 @@ function genericPrint(path, options, print) {
       ]);
     }
     case "selector-nesting": {
+      return node.value;
+    }
+    case "selector-unknown": {
+      const ruleAncestorNode = getAncestorNode(path, "css-rule");
+
+      // Nested SCSS property
+      if (ruleAncestorNode && ruleAncestorNode.isSCSSNesterProperty) {
+        return adjustNumbers(
+          adjustStrings(maybeToLowerCase(node.value), options)
+        );
+      }
+
       return node.value;
     }
     // postcss-values-parser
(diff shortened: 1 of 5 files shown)


src/language-css/printer-postcss.js
```diff
@@ -249,15 +249,9 @@ function genericPrint(path, options, print) {
       );
     }
     case "media-unknown": {
-      return adjustStrings(node.value, options);
+      return node.value;
```
evilebottnawi: We don't should change nothing in `unknown` nodes, because it is potential can break output. safe output as is

tests/css_comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -265,8 +265,8 @@ exports[`at-rules.css 1`] = `
 /* comment 20 */
 @import /* comment 21 */url("bluish.css") /* comment 22 */projection/* comment 23 */,
   /* comment 24 */tv/* comment 25 */;
-@import /* comment 26 */"custom.css"/* comment 27 */; /* comment 28 */
-@import /* comment 29 */ url("landscape.css") /* comment 30 */ screen /* comment 31 */ and /* comment 32 */ (/* comment 33 */orientation/* comment 34 */: /* comment 35 */landscape/* comment 36 */) /* comment 37 */;
+@import /* comment 26 */'custom.css'/* comment 27 */; /* comment 28 */
+@import /* comment 29 */ url('landscape.css') /* comment 30 */ screen /* comment 31 */ and /* comment 32 */ (/* comment 33 */orientation/* comment 34 */: /* comment 35 */landscape/* comment 36 */) /* comment 37 */;
```
evilebottnawi: Will be fixed after merge https://github.com/prettier/prettier/pull/4435

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -664,15 +664,15 @@ $KeepTopLevelVar: val;
     prop: val;
   }
 
-  background: {
+  BACKGROUND: {
```
evilebottnawi: WIP

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -343,7 +343,7 @@ $KeepScssVar: val;
 
   @{Keep}[@{Keep}][@{Keep}^=@{Keep-1A}]:@{Keep} {
     prop: val;
-  }
+  };
```
evilebottnawi: Because parse as Less `detached ruleset`, but it is invalid syntax for CSS/Less/SCSS, using `@{var}` in Less allowed only in string - i.e. `"@{var}"`

evilebottnawi:
Blocker https://github.com/prettier/prettier/pull/4435 (need first merge to avoid difficult rebase)

PR #4479, opened 2018-05-14
Keep parens around do-expressions inside ExpressionStatement

### LINKED ISSUE #4466: [bug] Expression statements containing only do expressions need to keep parens
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAKAJhABMAvgShABoQIAHGAS2gGdlQBDAJyYgHcAFZhOlBgNwiV0xEACMmDMAGs4MAMpkplKAHNkMJgFc4JFTThMYHSaoC2DZADMGAGwMkAVjQAeAIUky58hmbgAZFThrOwcQJSYDJmRxBjEAT1toUTJWAwB1STIY1Lgo-mCSJjgARy1KYpMGc0skG3tdEAMzSg1tRpoVVVs4AEUtCHgQhpIYOPThGAALZAAmUclKWy6AYQgzCxioaEKQLQMAFTjeerDUlRgJ9GnkAA4ABlxcIA)

**Input:**
```jsx
(do {})
```

**Output:**
```jsx
do {
};

```

**Expected behavior:**
```jsx
(do {
});

```

The prettier output won't parse at all (ie is invalid JS). Babylon has a test case for this, which is how I found it.

existentialism:
Fixes #4466.


nicolo-ribaudo:
Are parentheses required in this case?
```js
do {} + 1;
```

tests/do/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -16,4 +48,42 @@ const envSpecific = {
...
+  );
+}
+
+(do {
+});
+(do {
+} + 1);
+(1 +
+  do {
+  });
```
nicolo-ribaudo: Around `1 + do {}`, the parentheses aren't needed, because `do` isn't at the start of the statement.

j-f1: IMO we should keep them for consistency reasons. Also, this is a highly improbable case since the result of the expression is thrown away.

nicolo-ribaudo: Currently we don't for objects:
```js
// Input
({} + 1);
(1 + {});

// Output
({} + 1);
1 + {};
```

Since the rules for wrapping objects are the same of do expression, I think that they should behave the same.

> the result of the expression is thrown away.

Not with do expressions :stuck_out_tongue: 
```js
var a = do {
  1 + do { if (x) 1; else 2; };
};
```

bakkot:
This puts parentheses in `() => do {}` (I think, from reading the code). Personally I don't think it should, which means moving `case "DoExpression":`  up a line (and probably renaming the `forbidFunctionAndClass` parameter).

Either way, `() => do {}` should probably be added as a test.

tests/do/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -16,4 +49,45 @@ const envSpecific = {
...
+(do {
+});
+(do {
+} + 1);
+1 +
+  do {
+  };
+() =>
+  do {
+  };
```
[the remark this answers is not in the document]
existentialism: 👍 updated!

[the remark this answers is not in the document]
duailibe: This could've been another PR, unrelated to the issue this is fixing

PR #4482, opened 2018-05-14
Refactor frontmatter handling

duailibe:
We now have several methods for handling front-matter. This commit normalizes this code to be reused in different parsers.

When we introduce the pragma detection and insertion for CSS we'll need to use like Markdown does right now.


src/language-css/parser-postcss.js
```diff
@@ -438,32 +439,32 @@ function parseNestedCSS(node) {
 }
 
 function parseWithParser(parser, text) {
+  const frontmatter = (function() {
+    const parsed = parseFrontmatter(text);
+    text = parsed.content;
+    return parsed.frontmatter;
+  })();
```
j-f1: You should be able to remove the IIFE here.

src/language-css/parser-postcss.js
```diff
@@ -438,32 +439,32 @@ function parseNestedCSS(node) {
 }
 
 function parseWithParser(parser, text) {
+  const frontmatter = (function() {
```
j-f1: I personally would write this `frontMatter` and `front-matter`. What do you think?

src/language-css/parser-postcss.js
```diff
@@ -472,7 +473,7 @@ function requireParser(isSCSS) {
   }
 
   // TODO: Remove this hack when this issue is fixed:
-  // https://github.com/shellscape/postcss-less/issues/88
+  // https://github.com/shellscape/postcss-less/isses/88
```
j-f1: Typo?

src/utils/frontmatter.js
```diff
@@ -0,0 +1,28 @@
...
+    delimiter = "---";
+  }
+
+  if (text.indexOf("+++") === 0) {
+    delimiter = "+++";
+  }
+
+  let end = -1;
+
+  if (!delimiter || (end = text.indexOf(`\n${delimiter}`, 3)) === -1) {
```
j-f1: You shouldn’t need the start index — if the file starts with `---`, it by definition doesn’t start with `\n---` and we’ll instead capture the whole document even without the start index.

tests/css_yaml/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -31,6 +31,9 @@ a {
 }
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ---
+
+
+
```
j-f1: This looks like a regression to me.

duailibe: We'll print the front(-)matter as-is from now on

src/language-css/parser-postcss.js
duailibe: I know that, but why?

src/language-css/parser-postcss.js
duailibe: Makes no difference to me 🤷‍♂️

src/utils/frontmatter.js
duailibe: > and we’ll instead capture the whole document even without the start index.

I don't understand what you mean. You're right it doesn't need the start index but that won't change anything

src/language-css/parser-postcss.js
j-f1: It feels like it would run slower, And it makes the code more complex and harder to understand.

PR #4491, opened 2018-05-16
Set up CircleCI

duailibe:
Context: #3919

With CircleCI we can make a "pipeline" that first builds the production bundle in a modern Node environment, then uses the built artifact to run test suites in other node versions (in parallel).

This opens up some new interesting possibilities that I'm really excited to explore in the near future


duailibe:
It's working on my fork: https://circleci.com/workflow-run/8058899c-5a46-4dce-8e64-f016b3775bb9

PR #4241, opened 2018-04-02
Inline Angular async tests and beforeEach, etc.

ad1992:
Fixes #3791

PR #4495, opened 2018-05-17
Inline AngularJS tests that use `inject`

thorn0:
Like #4241, but for the older Angular.

A code example (from [here](https://docs.angularjs.org/guide/unit-testing#using-beforeall-)):
```js
  it("has calculated the answer correctly", inject(function(DeepThought) {
    expect(DeepThought.answer).toBe(42);
  }));
```

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5581,30 +5581,31 @@ const unitTestRe = /^(skip|[fx]?(it|describe|test))$/;
 
 // eg; `describe("some string", (done) => {})`
 function isTestCall(n, parent) {
-  if (n.type === "CallExpression") {
-    if (n.arguments.length === 1) {
-      if (isAngularTestWrapper(n) && parent && isTestCall(parent)) {
-        return isFunctionOrArrowExpression(n.arguments[0].type);
-      }
+  if (n.type !== "CallExpression") {
+    return false;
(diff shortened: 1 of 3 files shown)


src/language-js/printer-estree.js
```diff
@@ -5633,11 +5633,13 @@ function isTemplateLiteral(node) {
   return node.type === "TemplateLiteral";
 }
 
-function isIdentiferAsync(node) {
+// `inject` is used in AngularJS 1.x, `async` in Angular 2+
+// example: https://docs.angularjs.org/guide/unit-testing#using-beforeall-
+function isIdentiferInjectOrAsync(node) {
```
j-f1: What do you think about remaining this to something that specifies its purpose instead of just what it tests for?

thorn0: fixed, pls have a look

src/language-js/printer-estree.js
```diff
@@ -5580,45 +5577,40 @@ function isObjectType(n) {
   return n.type === "ObjectTypeAnnotation" || n.type === "TSTypeLiteral";
 }
 
+const unitTestRe = /^(skip|(f|x)?(it|describe|test))$/;
```
j-f1: `(f|x)` can be `[fx]`.

Issue #4492, opened 2018-05-17
Incorrect output for CSS level 4 color adjusters


taion:
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA6AZhCACYAdKbbSAGwgCcliIzyAKcuAE2xLoGpsAmABgFIAlAIDcBAL4gANCAgAHGAEtoAZ2SgAhuXIQA7gAVNCVSnUA3CAqZSQAI3LqwAazgwAyrIcKoAc2QxyAK5w0l7KcOQwevbeALbqyOjqJGHSAFbKAB4AQvZOLq7qMXAAMl5wCUkpIB7kYeTI4Mqq0rLkXjAA6pYwABbIABw8LdphHfayDa1wdabl0owAjgEKjFHqsfFIicnBIGExCn6Bu8pe3iRwAIoBEPAVO9Iw6jZdTL3IXI-2CiRnAMIQGJxBpQaBzEABMIAFWexm2YTEYiAA)
```sh
--parser css
```

**Input:**
```css
.foo {
  color: color(red l(+ 20%));
}
```

**Output:**
```css
.foo {
  color: color(red l(+20%));
}

```

**Expected behavior:**
```css
.foo {
  color: color(red l(+ 20%));
}

```

Per the spec, the space after the `+` is meaningful:

>Note: While scaling can be specified without any spaces, like lightness(*150%), adding/subtracting must be done with spaces after the +/-, or else the +/- will be interpreted as part of the number by the CSS parser.

https://www.w3.org/TR/css-color-4/#typedef-color-adjuster

evilebottnawi:
Oh, `css-color-4` contains breaking changes in syntax.

Changes from Colors 3
- rgb() and rgba() functions now accept <number> rather than <integer>.
- hsl() and hsla() functions now accept <angle> as well as <number> for hues.
- All uses of <alpha-value> now accept <percentage> as well as <number>.

It can't have compatibility with Less and SCSS, i don't have idea right now how we can handle this situation. Feedback welcome. Not in `1.13`, sorry

taion:
Also, it probably needs to be renamed to `color-mod` or something.

Currently the transpiler here is fine with the spaces being absent and interprets them the way a user would want, but this will start breaking at some point; there was a merged PR that was reverted because of how much it broke people, so this is a real problem: https://github.com/ianstormtaylor/css-color-function/pull/19.

[...]

PR #4501, opened 2018-05-18
refactor: `value-comma_group` node

evilebottnawi:
Second refactoring step

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -39,19 +39,30 @@ const isSCSSControlDirectiveNode = utils.isSCSSControlDirectiveNode;
 const isDetachedRulesetDeclarationNode = utils.isDetachedRulesetDeclarationNode;
 const isRelationalOperatorNode = utils.isRelationalOperatorNode;
 const isEqualityOperatorNode = utils.isEqualityOperatorNode;
+const isMultiplicationNode = utils.isMultiplicationNode;
+const isDivisionNode = utils.isDivisionNode;
+const isAdditionNode = utils.isAdditionNode;
 const isMathOperatorNode = utils.isMathOperatorNode;
 const isEachKeywordNode = utils.isEachKeywordNode;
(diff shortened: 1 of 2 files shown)


src/language-css/printer-postcss.js
```diff
@@ -461,166 +474,142 @@ function genericPrint(path, options, print) {
...
-          isNextForKeyword ||
-          isNextEachKeyword
-        ) {
-          parts.push(line);
+
+          continue;
         }
+
+        // Be default all values go through `line`
+        parts.push(line);
```
evilebottnawi: Just add own logic before this for formatting and it is all, easy read = easy maintain

evilebottnawi:
After merge solve https://github.com/prettier/prettier/issues/4492

src/language-css/printer-postcss.js
```diff
@@ -39,19 +39,30 @@ const isSCSSControlDirectiveNode = utils.isSCSSControlDirectiveNode;
...
 const isDetachedRulesetCallNode = utils.isDetachedRulesetCallNode;
 const isPostcssSimpleVarNode = utils.isPostcssSimpleVarNode;
 const isSCSSMapItemNode = utils.isSCSSMapItemNode;
+const isInlineValueCommentNode = utils.isInlineValueCommentNode;
+const isHashNode = utils.isHashNode;
+const isLeftCurlyBraceNode = utils.isLeftCurlyBraceNode;
+const isRightCurlyBraceNode = utils.isRightCurlyBraceNode;
+const isWordNode = utils.isWordNode;
+const isColonNode = utils.isColonNode;
+const isMediaAndSupportsKeywords = utils.isMediaAndSupportsKeywords;
```
j-f1: What do you think about (maybe in a future PR) getting rid of all of these declarations and just prefixing the variables with `utils.` when they’re used?

evilebottnawi: @j-f1 maybe, i think we can import `utils` as `u` for less breaking lines (example in `if`)

PR #4512, opened 2018-05-19
Don't add new lines after unless present in graphql

ad1992:
Fixes https://github.com/prettier/prettier/issues/4164


tests/multiparser_js_graphql/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -386,6 +388,8 @@ gql\`
     test
   }
 
+
+
```
lydell: This looks like a regression?

src/language-graphql/printer-graphql.js
```diff
@@ -24,10 +24,24 @@ function genericPrint(path, options, print) {
 
   switch (n.kind) {
     case "Document": {
-      return concat([
-        join(concat([hardline, hardline]), path.map(print, "definitions")),
-        hardline
-      ]);
+      const parts = [];
+      let i = 0;
+      path.map(pathChild => {
```
lydell: The index is provided as the second parameter to `.map` so no need to manually count.

tests/multiparser_js_graphql/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -17,6 +17,8 @@ gql\`
 
 `;
 
+
+
```
lydell: Just a curios Jest question: Does anybody know where these newlines came from?

ad1992: This got fixed when i merged with latest master.

ad1992:
@lydell @j-f1  can this be merged ?

j-f1:
Thanks for contributing!

PR #4513, opened 2018-05-20
added `--stdin` and `--stdin-filepath` to website

### LINKED ISSUE #4359: `--stdin` and `--stdin-filepath` are not shown on CLI section of the website
When visiting https://prettier.io/docs/en/cli.html, I expected to see `--stdin` and `--stdin-filepath` documented, but they are not. `--stdin-filepath` is mentioned elsewhere at https://prettier.io/docs/en/options.html#filepath, but that's a bit harder to find, and `--stdin` is still missing regardless.

Inspired by https://github.com/prettier/prettier/issues/4351, which was solved by using `--stdin-filepath`.

derekkramer:
Appended `--stdin` and `--stdin-filepath` to the CLI section of the website with examples.

resolves #4359


docs/cli.md
```diff
@@ -115,3 +115,35 @@ Change the level of logging for the CLI. Valid options are:
...
+A path to the file that the Prettier CLI will treat like stdin. For example:
+
+_abc.css_
+```css
+.name { display: none; }
+```
+
+_shell_
+```bash
+$ prettier --stdin-filepath abc.css
```
ikatyang: See https://github.com/prettier/prettier/issues/4514#issuecomment-390451393

ikatyang: We currently only treat `--stdin` as a signal to force using stdin.

https://github.com/prettier/prettier/blob/938d2f06be95012c0e512828958f2ee54a645958/src/cli/index.js#L54-L55

docs/cli.md
```diff
@@ -115,3 +115,18 @@ Change the level of logging for the CLI. Valid options are:
 * `log` (default)
 * `debug`
 * `silent`
+
+## `--stdin`
+
+Force the Prettier CLI to read input from stdin. For example:
+
+```bash
+$ prettier --stdin
```
lydell: This doesn't work with latest master.

```
$ ./bin/prettier.js --stdin
function sampleFunction (var1,
var2) {
  return  var1 *  var2
}
[error] No parser and no file path given, couldn't infer a parser.
```

lydell:
The PR title mentions both `--stdin` and `--stdin-filepath` but the diff only mentions `--stdin?`

derekkramer:
@lydell 
I may guess I misread the requested edits as wanting to remove `--stdin-filepath` from the CLI page. I had found documentation on both flags in other forms of Prettier docs and copied the explanations almost word for word from them (aside from the examples I gave). I'm not sure which flag is supported by the latest `master` or whether they should be added to `cli.md` at all.

lydell:
The `--stdin` flag seems to have been added in commit 9ffbd32723b7b9a5ad7bf454c113fe16436a862d (2017-01-11) as the only way to read from `--stdin`. commit 239521e97df824e477187c361c57cf9911fc6421 (2017-01-26) added stdin auto detection, but it looks like we've kept around the flag anyway since then.

I can't think of any use case for the flag either.

azz:
Moved that discussion to #4559.

derekkramer:
So I re-added the word-for-word documentation of the `--stdin-filepath` flag as per @azz 

I removed `--stdin` in anticipation of Prettier 2.0

And I resolved the remaining merge conflicts.

I think we're at an agreeable place for documentation, if I'm on the right page with the discussion at #4559 and the above requests.

j-f1:
I think you should add the code example back and replace the shell part with `cat abc.css | prettier --stdin-filepath abc.css`. What do you think?

derekkramer:
@j-f1 
I agree. Added the example with the change as requested and confirmed it's working with the most updated version of Prettier.

PR #4529, opened 2018-05-23
refactor: math operator

evilebottnawi:
I rewrote complex logic for a more readable. Also require for https://github.com/prettier/prettier/issues/4492

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -42,9 +42,9 @@ const isEqualityOperatorNode = utils.isEqualityOperatorNode;
 const isMultiplicationNode = utils.isMultiplicationNode;
 const isDivisionNode = utils.isDivisionNode;
 const isAdditionNode = utils.isAdditionNode;
+const isSubtractionNode = utils.isSubtractionNode;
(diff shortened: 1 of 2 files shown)


src/language-css/printer-postcss.js
```diff
@@ -507,60 +507,62 @@ function genericPrint(path, options, print) {
...
+          (iNextNextNode && isWordNode(iNextNextNode)) ||
+          iNode.type === "value-func" ||
+          isWordNode(iNode);
+        const requireSpaceAfterOperator =
+          iNextNode.type === "value-func" ||
+          isWordNode(iNextNode) ||
+          (iPrevNode && iPrevNode.type === "value-func") ||
+          (iPrevNode && isWordNode(iPrevNode));
+
+        // Formatting `/`, `+`, `-` sign
```
lydell: Isn't this `if` about _not_ formatting things? Could you add helpful examples like in the above two `if`s?

evilebottnawi: @lydell i will do second refactor and add message in second PR :+1:

PR #4533, opened 2018-05-24
Fix HTML entity escaping in Glimmer printer

omghax:
Previously, prettier would unescape `&lt;` and `&gt;` which would result in invalid HTML. This change escapes Glimmer TextNode nodes to ensure that `&`, `<`, and `>` remain escaped properly.


lydell:
```
&amp;
&nbsp;
&#160;
&#xa0;
```

src/language-handlebars/printer-glimmer.js
```diff
@@ -28,6 +28,21 @@ const voidTags = [
...
+// Escaping logic based on handlebars.js:
+// https://github.com/wycats/handlebars.js/blob/master/lib/handlebars/utils.js
+const escape = {
+  "&": "&amp;",
+  "<": "&lt;",
+  ">": "&gt;",
+  "\xa0": "&nbsp;"
+};
+
+const badChars = /[&<>\xa0]/g;
```
lydell: So you had to add yet another character here? Where will it end?

tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -426,6 +426,14 @@ exports[`else-if.hbs 1`] = `
 {{/if}}
 `;
 
+exports[`escaped-chars.hbs 1`] = `
+<p>Some escaped characters: &lt; &gt; &amp; &nbsp; &#160; &#xa0;</p>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<p>
+  Some escaped characters: &lt; &gt; &amp; &nbsp; &nbsp; &nbsp;
```
lydell: So all of those are turned into `&nbsp;`? Not sure if that’s a feature or a bug. Perhaps it’s out of scope for this PR.

lydell:
This doesn’t feel like the right thing to do overall.

Perhaps we could merge a quick-fix for only `<`, `>` and `&` so that at least the output code is valid, and then open an issue about properly handling escapes.

omghax:
@lydell Okay, I'll remove that commit from this PR and just leave it focused on `<`, `>`, and `&`.

rwjblue:
FWIW - `@glimmer/syntax` uses `simple-html-tokenizer` which ships with a file with all the HTML char references ([see here](https://github.com/tildeio/simple-html-tokenizer/blob/master/src/generated/html5-named-char-refs.ts)) which is actually _generated_ from https://github.com/w3c/html/blob/master/entities.json...

j-f1:
What’s the status of this? Should we merge or close?

lifeart:
@omghax @j-f1 sorry for pinging, is it possible to get it merged?

j-f1:
@lydell what do you think?

lydell:
I guess we could merge it.

j-f1:
Can you fix the merge conflicts @omghax?

billdami:
Just curious on the status of this issue? Also, it was a bit hard to judge from the discussion here, but will this fix result in `&nbsp;` _not_ being converted to literal spaces? We are seeing issues with these specifically, since `&nbsp;` behaves differently in the DOM then a normal space (e.g. collapsing multiple adjacent spaces).

Ideally, we'd love a way to disable prettier's entity conversion entirely (for example we prefer to see/edit `&raquo;` in the source rather than `»`), but non-breaking spaces was the more immediate concern.

evilebottnawi:
@billdami looks like PR was abandoned, feel free to send new PR

rwjblue:
FWIW, new changes were made to the latest @glimmer/syntax package to allow a "codemod" mode. In that mode, **no** entity escaping is done at all. If we migrate to that new mode, I think we will have a much better time.

See https://github.com/glimmerjs/glimmer-vm/pull/938 for the changes I'm referring to...

lydell:
@rwjblue Awesome! We should definitely try that out.

GavinJoyce:
I've created https://github.com/prettier/prettier/pull/6234 which uses the new "codemod" glimmer mode

GavinJoyce:
@evilebottnawi perhaps you meant to approve https://github.com/prettier/prettier/pull/6234 instead? I think this PR can then be closed

lydell:
Supersedes by #6234.

PR #4535, opened 2018-05-24
Fix printing of {{{raw}}} handlebars expressions

omghax:
Previously, `{{{raw}}}` would be output as `{{raw}}`.

It seems as though this logic was (mostly) in place [inside `printer-glimmer.js`](https://github.com/prettier/prettier/blob/d0f6aac0017a7e0697feaf71eba205740c46ed35/src/language-handlebars/printer-glimmer.js#L151-L154), but commented out for some reason.


src/language-handlebars/printer-glimmer.js
```diff
@@ -148,10 +148,10 @@ function print(path, options, print) {
       const isConcat = pp && pp.type === "ConcatStatement";
       return group(
         concat([
-          /*n.escaped ? "{{{" : */ "{{",
+          n.escaped === false ? "{{{" : "{{",
           printPathParams(path, print),
           isConcat ? "" : softline,
-          /*.escaped ? "}}}" :*/ "}}"
+          n.escaped === false ? "}}}" : "}}"
```
j-f1: This could just be `n.escaped ? "}}" : "}}}"`.

omghax: That's what I tried initially, but it resulted in test failures because prettier would output `{{{` where it should have output `{{`. My hunch is that `escaped` is only set when it's `false`, otherwise it's `undefined` or `null`.

PR #4536, opened 2018-05-24
Update babylon & flow

vjeux:
This updates babylon and flow. The motivation is that we want to start using `?.` and `??` at Facebook :)

The two big changes are:
- Babylon shipped the new decorator spec in master and you need to use the legacy one which has a slightly different API.
- The new `?.` spec adds "OptionalXXX" nodes, so we need to update all the call sites to support it. Thankfully @xixixao did most of the hard work in #4463

### DIFF
--- a/src/language-js/clean.js
+++ b/src/language-js/clean.js
@@ -128,7 +128,7 @@ function clean(ast, newObj, parent) {
   }
 
   // CSS template literals in Angular Component decorator
-  var expression = ast.expression || ast.callee;
+  const expression = ast.expression || ast.callee;
   if (
     ast.type === "Decorator" &&
     expression.type === "CallExpression" &&
--- a/src/language-js/clean.js
+++ b/src/language-js/clean.js
@@ -89,6 +89,11 @@ function clean(ast, newObj, parent) {
     delete newObj.key;
   }
 
+  if (ast.type === "OptionalMemberExpression" && ast.optional === false) {
+    newObj.type = "MemberExpression";
+    delete newObj.optional;
+  }
+
   // Remove raw and cooked values from TemplateElement when it's CSS
   // styled-jsx
   if (
@@ -123,12 +128,13 @@ function clean(ast, newObj, parent) {
   }
 
   // CSS template literals in Angular Component decorator
+  var expression = ast.expression || ast.callee;
   if (
     ast.type === "Decorator" &&
-    ast.expression.type === "CallExpression" &&
-    ast.expression.callee.name === "Component" &&
-    ast.expression.arguments.length === 1 &&
-    ast.expression.arguments[0].properties.some(
+    expression.type === "CallExpression" &&
+    expression.callee.name === "Component" &&
+    expression.arguments.length === 1 &&
+    expression.arguments[0].properties.some(
       prop =>
         prop.key.name === "styles" && prop.value.type === "ArrayExpression"
     )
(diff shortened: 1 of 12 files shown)


existentialism:
:+1:

I also have Prettier PRs ready for some new things in Flow 0.72+ (explicit type args is https://github.com/prettier/prettier/tree/issue4489, waiting for https://github.com/facebook/flow/pull/6320 for internal slot).

We're planning on a new Babel/Babylon release soon that should support them as well (Ref: https://github.com/babel/babel/pull/7973, https://github.com/babel/babel/pull/7934, https://github.com/babel/babel/pull/7947), so I'll follow up with updates to Prettier!

src/language-js/clean.js
```diff
@@ -89,6 +89,11 @@ function clean(ast, newObj, parent) {
     delete newObj.key;
   }
 
+  if (ast.type === "OptionalMemberExpression" && ast.optional === false) {
+    newObj.type = "MemberExpression";
+    delete newObj.optional;
+  }
```
vjeux: @existentialism you may know what's going on here

```js
(a?.b).c
a?.b.c
```

give slightly different ast. In the second one, the outer `.` is

```js
{
  type: "OptionalMemberExpression",
  optional: false,
}
```

This doesn't make sense to me, why doesn't it print `{type: "MemberExpression"}` instead?

existentialism: Think @jridgewell is going to explain it better than I can, but AFAIK it's due to https://github.com/tc39/proposal-optional-chaining#edge-case-grouping

jridgewell: This is intentional. The first example is explicitly breaking the optional chain by adding parenthesis. Imagine it were a sequence instead: `(0, a?.b).c`. Obviously in this case, I'm doing a regular property lookup on the result of the sequence.

The second case, however, is a continued optional chain. This shown by the ouptut: `a == null ? undefined : a.b.c`. The `.c` property access is "optional" in that it can be short-circuited. It's just not hard optional like the `a?.b` is.

vjeux: Is it safe (won't change behavior) to remove parenthesis there?

loganfsmyth: Removing the parens changes the runtime chaining behavior, so you wouldn't want to remove them.

vjeux:
We need to figure out what to do with the flow segfault :x

src/language-js/clean.js
jridgewell: No, they have different behavior: http://babeljs.io/repl/build/master#?babili=false&browsers=ie%2010&build=&builtIns=usage&code_lz=DYUwLgBAHgjBC8ECGB-AdAIzQYwFCkigCYEIAKVTAShyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=es2015%2Cstage-1&prettier=false&targets=&version=7.0.0-beta.48&envVersion=

However, there is no reason to every have parenthesis around an optional chain. The only observable difference between the `a?.b.c` and `a.b.c` expression (without the optional chain) is the error message. If you wanted it to throw, just use regular property chaining. If you wanted it to short-circuit, then you definitely didn't want it to throw.

vjeux: Thanks! I just opened an issue to track it: #4538. We didn't use to put parenthesis even before this update.

Issue #3457, opened 2017-12-11
Flow parser on Node v6 causes Segmentation Fault


flarnie:
Hi! This is an issue showing up in Draft.js CI runs.

We enabled Prettier as part of our Travis CI runs on Github recently - in https://github.com/facebook/draft-js/commit/cd20b992d4b054d70f756c258e5e75e518995e12

During CI we currently install the most recent compatible version of any dependency automatically, and so we started pulling in v1.9.0 and then v1.9.1 when those were released 6 days ago.

At that time we started seeing CI failing due to a segmentation fault when it runs the `eslint` check, which is the step that runs `prettier`.

I verified that disabling prettier or downgrading to v1.8.* does not lead to the segmentation fault - see https://github.com/facebook/draft-js/pull/1567

I have not dug into the Prettier codebase so I'm not sure what could be happening to cause a segmentation fault, but I'm happy to provide any info or help that I can. We'd like to keep using Prettier in Draft.js, and get the goodies in new releases.

Thanks for creating such a powerful and useful tool! 📐✨💕

duailibe:
Hi @flarnie! Thanks for the report! I have cloned the draft-js repo to try and reproduce this.

Is it happening on master? It runs find for me:

```
in draft-js/ on master 
› eslint --version
v4.13.0

in draft-js/ on master 
› prettier --version
1.9.1

in draft-js/ on master 
› eslint .

/Users/duailibe/Code/draft-js/src/model/transaction/__tests__/moveBlockInContentState-test.js
  102:11  warning  'data' is assigned a value but never used. Allowed unused vars must match /^_/           no-unused-vars
  103:11  warning  'characterList' is assigned a value but never used. Allowed unused vars must match /^_/  no-unused-vars
  104:11  warning  'depth' is assigned a value but never used. Allowed unused vars must match /^_/          no-unused-vars
  105:11  warning  'type' is assigned a value but never used. Allowed unused vars must match /^_/           no-unused-vars
  106:11  warning  'text' is assigned a value but never used. Allowed unused vars must match /^_/           no-unused-vars

✖ 5 problems (0 errors, 5 warnings)
```

I tried running both `npm install` and `yarn install` (after cleaning `node_modules/`) and got the same results. But I noticed you managed to run locally: https://github.com/facebook/draft-js/issues/1562#issuecomment-350770530

Do you have any other hints or info so that I can reproduce?

duailibe:
Oh I can reproduce it with node 6.12.2

[...]

PR #4539, opened 2018-05-24
Update to Jest 23

duailibe:
Changes are because of https://github.com/prettier/prettier/pull/3518#issue-158890517

We can do this now that we merged #4537


duailibe:
Guess not

SimenB:
Segfault? 😭

SimenB:
Oh, syntax error on node 4, gotcha

duailibe:
@SimenB Yes, either Jest or a dependency

The node 6 tests are failing ever since we enabled them but I don't think it's Jest fault (#3457)

SimenB:
Doesn't seems like you set `testEnvironment` in your jest config? jsdom is the default (and uses destructuring, so probably the culprit). `node` is also faster. Might be worth a go?

duailibe:
@SimenB Thanks, I'll try that

SimenB:
Docs, if it helps: https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string

We have been talking about changing the default to node, but that would break "zero config" for react. Easy enough to google `document is not defined`, I suppose...

@rickhanlonii ^

SimenB:
We discussed it at the summit this week. I do think there's an issue though, lemme see if I can dig it up.

duailibe:
@SimenB I think something else is failing
```
~/Code/prettier master*
❯ node -v
v4.9.1

~/Code/prettier master*
❯ yarn test -v
yarn run v1.6.0
$ jest -v
23.0.0
✨  Done in 0.37s.

~/Code/prettier master*
❯ yarn test
yarn run v1.6.0
$ jest
Test Suites: 0 of 838 total
Tests:       0 total
Snapshots:   0 total
Time:        0.395s, estimated 84s
Ran all test suites.
✨  Done in 1.96s.
```

That happens with `testEnvironment` node or unset

SimenB:
It's `@babel/code-frame` using invalid node 4 syntax. Not sure why it doesn't give any output, but since it's one of jest's dependencies, you can trigger transformation of it either

SimenB:
I'm looking into this now. We can try to do something clever ™ with the lockfile making sure we keep the babel deps on a version supported by node 4

SimenB:
See #4782 - there's syntax issues in Prettier's code base, though

PR #4543, opened 2018-05-25
Add support for flow inline interfaces

existentialism:
Flow commit: https://github.com/facebook/flow/commit/4f93f3adb7a5fedd3d572d62d718cf48c15a08ce

Babel commit: https://github.com/babel/babel/pull/7973

Will follow up w/ flow team on making `InterfaceType` -> `InterfaceTypeAnnotation`, since it seems a bit inconsistent.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -2485,8 +2485,11 @@ function printPathNoParens(path, options, print, args) {
         path.call(print, "id"),
         path.call(print, "typeParameters")
       ]);
+
     case "DeclareInterface":
-    case "InterfaceDeclaration": {
+    case "InterfaceDeclaration":
+    case "InterfaceType":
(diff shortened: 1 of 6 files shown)


src/language-js/printer-estree.js
```diff
@@ -2485,6 +2485,26 @@ function printPathNoParens(path, options, print, args) {
         path.call(print, "id"),
         path.call(print, "typeParameters")
       ]);
+
+    case "InterfaceType":
+    case "InterfaceTypeAnnotation": {
```
j-f1: Is there any way to combine this with the TS one?

existentialism: I was hesitant initially, since they can't have an `id` or `typeParameters`, but I went ahead and merged it with the logic for printing flow's `DeclareInterface` and `InterfaceDeclaration`.

PR #4540, opened 2018-05-24
Add support for flow explicit type arguments

existentialism:
Fixes #4489.

Babylon support landed in `beta.48` just now but requires a few other fixture changes, so I'm going to follow this up with another PR.

PR #4544, opened 2018-05-25
Switch to @babel/parser

existentialism:
We landed a change that added a new `InterpreterDirective` AST node type for hashbangs, and no longer add it as a comment/leadingComment.

Ref: https://github.com/babel/babel/pull/7928

I mimicked what we do in `@babel/generator` here, since I found it better than trying to add comments to the ast in `parser-include-shebang.js`). Definitely open to a better/cleaner option though!

Note: I'll follow this up with enabling tests for https://github.com/prettier/prettier/pull/4543 and https://github.com/prettier/prettier/pull/4540 once they land too.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -338,7 +338,21 @@ function printPathNoParens(path, options, print, args) {
   let parts = [];
   switch (n.type) {
     case "File":
-      return path.call(print, "program");
+      // Print @babel/parser's InterpreterDirective here so that
+      // leading comments on the `Program` node get printed after the hashbang.
+      if (n.program && n.program.interpreter) {
+        parts.push(
+          path.call(
+            programPath => programPath.call(print, "interpreter"),
+            "program"
+          )
+        );
+      }
+
+      parts.push(path.call(print, "program"));
+
+      return concat(parts);
+
     case "Program":
       // Babel 6
       if (n.directives) {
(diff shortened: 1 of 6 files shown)


src/language-js/printer-estree.js
```diff
@@ -3168,6 +3182,21 @@ function printPathNoParens(path, options, print, args) {
     case "TSInferType":
       return concat(["infer", " ", path.call(print, "typeParameter")]);
 
+    case "InterpreterDirective":
+      parts.push("#!", n.value, hardline);
+
+      if (
+        sharedUtil.isNextLineEmpty(
+          options.originalText,
+          path.getValue(),
```
duailibe: Can use `n` here

src/language-js/printer-estree.js
```diff
@@ -338,7 +338,21 @@ function printPathNoParens(path, options, print, args) {
...
+      // Print @babel/parser's InterpreterDirective here so that
+      // leading comments on the `Program` node get printed after the hashbang.
+      if (n.program && n.program.interpreter) {
+        parts.push(
+          path.call(
+            programPath => programPath.call(print, "interpreter"),
+            "program"
+          )
+        );
+      }
```
j-f1: Why can’t this be in the `Program` printer?

duailibe: Because the leading comments would get printed above the shebang

duailibe:
I'll see if I can get tests to pass on Node 4

duailibe:
@existentialism Ran this with #4548 and tests passed :)

existentialism:
@duailibe thanks for the rebase!

Did a quick bump to 49 that we just released which has some bug fixes. Will land after tests pass.

PR #4548, opened 2018-05-25
[build] Always transpile everything

duailibe:
Now that babel 7 beta 48 is out, we can do this.

This also means we can now start using Node 6 features in our own code now.


duailibe:
cc @j-f1

suchipi:
Nice!

scripts/build/build.js
```diff
@@ -49,16 +45,14 @@ function getBabelConfig(bundle) {
...
-      // From https://jamie.build/last-2-versions
-      targets.browsers = [">0.25%", "not ie 11", "not op_mini all"];
-    }
-    config.presets = [
-      [require.resolve("@babel/preset-env"), { targets, modules: false }]
-    ];
+  const targets = { node: 4 };
+  if (bundle.target === "universal") {
+    // From https://jamie.build/last-2-versions
+    targets.browsers = [">0.25%", "not ie 11", "not op_mini all"];
```
azz: Can we keep the IE11 support? People still use it.

lydell: Has it ever worked properly in IE11? Not sure anyone's tested that. Which usually means no.

PR #4550, opened 2018-05-25
Add standalone docs

duailibe:
cc @suchipi @azz @j-f1 for help :-)

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -13,7 +13,7 @@ See [Usage](browser.md#usage) below for examples.
 
 ## Usage
 
-### Global:
+### Global
 
 <!-- prettier-ignore -->
 ```html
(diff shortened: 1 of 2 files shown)


docs/browser.md
```diff
@@ -0,0 +1,58 @@
...
+## Usage
+
+### Global:
+
+<!-- prettier-ignore -->
+```html
+<script src="https://unpkg.me/prettier@1.13.0/standalone.js"></script>
+<script src="https://unpkg.me/prettier@1.13.0/parser-graphql.js"></script>
+<script type="text/javascript">
+prettier.format("query { }", { parser: "graphql", plugins: prettierPlugins });
```
duailibe: Thinking about it, maybe we could just read from `prettierPlugins` global and the `plugins` for additional plugins in case they are not added to the global?

docs/browser.md
```diff
@@ -0,0 +1,54 @@
...
+
+### CJS
+
+```js
+const prettier = require("prettier/standalone");
+const plugins = [require("prettier/parser-graphql")];
+prettier.format("query { }", { parser: "graphql", plugins });
+```
+
+This syntax doesn't necessarily works in the browser but this can be used when bundling the code with browserify or Rollup.
```
j-f1: > This syntax doesn't necessarily **work** in the browser**,** but **it** can be used when bundling the code with browserify or Rollup.

docs/browser.md
suchipi: It would only take effect when CJS/AMD is not used, which seems relatively uncommon...

suchipi: There might also be situations where you have multiple instances of prettier configured to run with different plugins, I guess? But 🤷

suchipi:
I'll take a look at this later tonight if I have time but don't wait for my review if it's ready to merge before I can look at it

docs/browser.md
duailibe: Makes sense!

docs/browser.md
```diff
@@ -0,0 +1,54 @@
+---
+id: browser
+title: Browser
+---
+
+Run Prettier in the browser with the `standalone.js` UMD bundle shipped in the NPM package (starting in version 1.13). The new UMD bundle only formats the code and has no support for config and ignore files, no CLI usage and doesn't load plugins automatically.
+
+### `prettier.format(code, options)`
+
+Differently than the `format` function from the [main API](api.md#prettierformatsource-options), this function does not load plugins automatically, so a `plugins` property is required if you want to load external plugins. This also means the parsers included in the Prettier package won't be loaded automatically, so you need to load them before using.
```
azz: Doesn't load both internal and external plugins, so maybe drop the word "external"?

docs/browser.md
```diff
@@ -0,0 +1,54 @@
+---
+id: browser
+title: Browser
+---
+
+Run Prettier in the browser with the `standalone.js` UMD bundle shipped in the NPM package (starting in version 1.13). The new UMD bundle only formats the code and has no support for config files, ignore files, CLI usage, or automatic loading of plugins.
```
suchipi: I just realized this, but we might also want to mention that `formatWithCursor` isn't present (unless that's unintended?)

duailibe: Probably an oversight on my side.. there’s no reason to leave it out!

suchipi: https://github.com/prettier/prettier/pull/4556

PR #4551, opened 2018-05-25
Add support for Flow's proto modifier syntax

existentialism:
Flow commit: https://github.com/facebook/flow/commit/eb815be907eeae9f4851ff154087ebf079510d18

Babel PR: https://github.com/babel/babel/pull/7978

### DIFF
--- a/tests/flow_proto_props/jsfmt.spec.js
+++ b/tests/flow_proto_props/jsfmt.spec.js
@@ -1,0 +1,1 @@
+run_spec(__dirname, ["flow"]);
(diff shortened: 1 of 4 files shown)


tests/flow_proto_props/jsfmt.spec.js
```diff
@@ -0,0 +1 @@
+run_spec(__dirname, ["flow"]);
```
suchipi: Since this is supported by babel, too, should we run this with babylon as well?

existentialism: We actually _just_ landed it, so I'll enable it after our next release!

suchipi: Cool :+1:

PR #4554, opened 2018-05-26
Use destructuring for imports

azz:
Now that #4548 is merged it is safe to use Node v6 features, such as destructuring.

I've refactored a large number of `require` statements to use destructuring, as I think it has a huge impact on legibility.

Also in this PR:
- In integration tests, replaced `require("../../../require_prettier")` calls with `require("prettier/local")`


src/language-css/printer-postcss.js
```diff
@@ -1,67 +1,71 @@
...
+  isPostcssSimpleVarNode,
+  isSCSSMapItemNode,
+  isInlineValueCommentNode,
+  isHashNode,
+  isLeftCurlyBraceNode,
+  isRightCurlyBraceNode,
+  isWordNode,
+  isColonNode,
+  isMediaAndSupportsKeywords
+} = require("./utils");
```
j-f1: What do you think about just doing `const utils = require("./utils")` here to decrease the number of local variables?

azz: Advantage of many locals is ESLint will catch typos. That isn't the case with `until.foo`.

j-f1: Good point, although it won’t catch typos if you make them in the initial destructuring. (We’d need type checking or ESM to get errors in that case)

duailibe:
Some of the files are imported in `tests_integration` tests (like `early-exit.js`)

azz:
Should we use `babel-jest` for the integration tests on Node 4, or clean them up to not use any prettier internals?

azz:
Going for the cleanup route for now.

azz:
Replaced all `require("../../../../require_prettier")` calls with `require("prettier/local")` with Jest's `moduleNameMapper` feature. There were plugins in the integration tests that were using the non-built version of Prettier.

azz:
Down to two tests failing:

```
 FAIL  tests_integration/__tests__/file-info.js
  ● API getFileInfo with no args

    expect.assertions(1)

    Expected one assertion to be called but received zero assertion calls.

      at extractExpectedAssertionsErrors (node_modules/expect/build/extract_expected_assertions_errors.js:46:19)

  ● API getFileInfo.sync with no args

    expect(function).toThrow(undefined)

    Expected the function to throw an error.
    But it didn't throw anything.

      at Object.<anonymous> (tests_integration/__tests__/file-info.js:116:45)
```

azz:
Looks like `master` is failing with the same `Segmentation Fault` as this PR. https://travis-ci.org/prettier/prettier/jobs/384227019#L693 (Travis Node 6)

azz:
I think this is good to merge, but I'll let someone else take a look first.

duailibe:
> Looks like `master` is failing with the same `Segmentation Fault` as this PR. https://travis-ci.org/prettier/prettier/jobs/384227019#L693 (Travis Node 6)

It's #3457 still

I'll take a look at the PR now

jest.config.js
```diff
@@ -19,7 +29,9 @@ module.exports = {
     // `graceful-fs` does `require('fs')`.
     // Ref: https://github.com/facebook/jest/issues/2179#issuecomment-355231418
     // If this is removed, see also scripts/build/build.js.
-    "graceful-fs": "<rootDir>/tests_config/fs.js"
+    "graceful-fs": "<rootDir>/tests_config/fs.js",
+
+    "prettier/local": "<rootDir>/tests_config/require_prettier.js"
```
duailibe: Why not `"prettier"`?

azz: Two reasons:
1. If moduleNameMapper is misconfogured, well be requiring the prettier from node_modules.
2. I was getting errors from jest due to something related to Haste.

duailibe: If moduleNameMapper is misconfigured, tests will fail either way, but I understand that a missing `local.js` will be more easy to know what's happening

PR #4557, opened 2018-05-27
Update docs for default parser

duailibe:
Context: #4528

I don't know how to communicate the change between versions.

Related: do you think we should start versioning docs (supported by docusaurus)? Would this be enough to communicate changed values?

(versioning docs are also useful for "hiding" unreleased features anyway. so we should keep the `since v1.XX.X` we use and use docs versioning)

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -174,9 +174,11 @@ Built-in parsers:
 
 [Custom parsers](api.md#custom-parser-api) are also supported. _Since v1.5.0_
 
-| Default   | CLI Override                                    | API Override                                               |
-| --------- | ----------------------------------------------- | ---------------------------------------------------------- |
-| `babylon` | `--parser <string>`<br />`--parser ./my-parser` | `parser: "<string>"`<br />`parser: require("./my-parser")` |
+| Default | CLI Override                                    | API Override                                               |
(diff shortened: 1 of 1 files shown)


j-f1:
It looks like something went wrong with the Netlify build — all of the pages 404 as far as I can tell.

duailibe:
I haven't finished yet..

duailibe:
@azz @j-f1 https://deploy-preview-4557--prettier.netlify.com/

https://deploy-preview-4557--prettier.netlify.com/versions.html is a generated page, we can customize that

duailibe:
Removed the docusaurus stuff from the PR, we can add that later if we want to

j-f1:
Cool! ✨

- The links on the `/versions` page don’t seem to do anything — they just reload the page.
- How do you feel about putting the descriptive text above the tables so that people don’t have to scroll all the way down to see the information about the last section?
- Is it possible to have a dropdown for available versions for the current doc page appear above the sidebar to allow switching versions without having to navigate back to the page?

duailibe:
Ooops sorry @j-f1 😔 

1) Yeah that was generated by Docusaurus, we need to add the links and stuff
2) Sure
3) I believe it's possible

PR #4562, opened 2018-05-27
Add a single test suite for blank files in all parsers

j-f1:
The parsers that don’t yet support blank files have been filtered out of the list until they do.


tests/empty/jsfmt.spec.js
```diff
@@ -1 +1,12 @@
-run_spec(__dirname, ["flow", "typescript"]);
+const core = require("../../src/main/core-options");
```
duailibe: See #4554 about importing code directly in tests

duailibe: Might try using `require('../tests_config/require_prettier')` (or `require('prettier/local')`, who lands first) and `prettier.getSupportInfo()`

tests/js_empty/jsfmt.spec.js
```diff
@@ -0,0 +1 @@
+run_spec(__dirname, ["flow", "typescript"]);
```
duailibe: babylon?

PR #4563, opened 2018-05-27
feat: support YAML

### LINKED ISSUE #3017: Add support for YAML files
I'm not sure how difficult/useful it will be but I'm guessing that it's less complicated than adding Java or Python :D

ikatyang:
Fixes #3017

- use [yaml](https://github.com/eemeli/yaml) + [yaml-unist-parser](https://github.com/ikatyang/yaml-unist-parser)
  - from [`yaml` README](https://github.com/eemeli/yaml/tree/master/src/ast#read-anything-as-yaml):
    > Does practically no error checking, and should never throw -- but if you feed it garbage, you'll likely get a garbage AST as well
  - since the AST from `yaml` is hard to use:
    - (flow)mapItem is just a collection of `{`, `,`, `}`, etc.
    - some comment are attached to a weird place
  - so I use `yaml-unist-parser` to transform the `yaml` AST and re-attach comments.

+ support all nodes, including comments
+ support `# pretter-ignore`
+ support requirePragma/insertPragma
+ support bracketSpacing for `flowSequence`
+ support empty line
+ passed tests from [yaml-test-suite](https://github.com/eemeli/yaml-test-suite/tree/fixed-data)

TODOs:

- [x] ~~better error message: https://github.com/prettier/prettier/pull/4563#issuecomment-392529476 -> ikatyang/yaml-unist-parser#14~~ (fixed in 7ca4e696b2e464afb469bc6691361c1c0798a83c)
- [x] ~~fix next empty line in [this case](https://deploy-preview-4563--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEwA6UCG2e7-gwo4vJAAgEYAmAZgBoyAjZl1t9jzr7t86+shgxg+tDAF8MIOiAgAHGAEtoAZ2ShMAJ00QA7gAUtCNSkwA3CIoAm0kI02YwAazgwAynMeKoAc2QxNAFc4GW8VOE0YfQcfAFtMZAAzTAAbcJkAKxUADwAhB2dXN0xYuAAZbzgk1PSQT01wzWQQAE8SlNs5TW8YAHVrGAALZAAOAAYZLohw3oc5Zq64RrMqmU04AEdAxXXozDiEpGS0kJBw2MV-INOVbx8UuABFQIh4apOZGExGfqsh5ConwcihSdwAwhBYvFmlBoKsQIFwgAVb4mY7hcTiIA)~~ (fixed in d0b19083eab4a99f634464021fb6935ddadbb895)
- [x] ~~`yaml-logo.svg` that fits our homepage style https://github.com/prettier/prettier/pull/4563#discussion_r191079846~~ (fixed in 277a03c642c3417a7f3c180c99fd7d79c09d10d7)
- [ ] ~~investigate which indent style for sequence in mapping is more popular https://github.com/prettier/prettier/pull/4563#issuecomment-392401797~~
~~-> indented sequence in mapping is more readable https://github.com/prettier/prettier/pull/4563#issuecomment-397812369 (fixed in 0892d1780afc7cebd51565040c2b8bbf2ff3bdf7)~~
-> we should seek wider consensus for this indent style https://github.com/prettier/prettier/pull/4563#issuecomment-397917375
- [x] ~~fix node 4 failure~~ (fixed in 4d2733c8b62d7dd65dca55c31548036b24c37e5b)
- [x] ~~support singleQuote for `quoteSingle`/`quoteDouble` (currently print as-is)~~ (fixed in 77116914f1e44da0d28f9c174966a7ea37edce28)

Future PRs:

- support proseWrap for `blockFolded` (currently always wrap)
- format `yaml` node in markdown

[Playground](https://deploy-preview-4563--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEIA0IIAcYEtoDOyoAhgE5kQDuACuQkSiQG4S4Am6IARmSWAGs4MAMpZ+uKAHNkMMgFc4GSQThkYNPlIC2JZADMSAG1UYAVgQAeAIT6DhIktrgAZSXAPHTIcWVVlkEABPJyMuLDJJGAB1DhgAC2QADgAGDAiIVWi+LECIuH9mDwwyOABHeVxSzRIdPSRDEyUQVW1cWQVmgkkpIzgARXkIeE8mjBgSblj2BOQAJnG+XCMegGEIbV1AqGhikHlVABVJxkbVAF9zoA)


tests/yaml_spec/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,2828 @@
...
+# comment
+
+`;
+
+exports[`block-sequence-in-block-mapping.yml 1`] = `
+key:
+ - item1
+ - item2
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+key:
```
azz: I usually don't indent these (I would write it like the input). Do you have opinions on this?

ikatyang: I forgot to not indent it. 😂

azz: Cool. It will probably come up again as people probably have _very_ strong opinions both ways. 😆

nevir:
Looks like the parser isn't happy with nested lists of objects/lists in some cases.

Example (also see [original config](https://gist.github.com/nevir/6457342c559ced8bbe65caa4bd5e0a2f) for more context):
```yaml
aliases:
  - docker:
    - image: circleci/node:8.11.2
  - key: repository-{{ .Revision }}
```

```
SyntaxError: Sequence items are not allowed on the same line with map keys (4:3)
  2 |   - docker:
  3 |     - image: circleci/node:8.11.2
> 4 |   - key: repository-{{ .Revision }}
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  5 | 
```

suchipi:
It's kinda weird to me that we don't indent these lists:

**Prettier pr-4563**
[Playground link](https://deploy-preview-4563--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBnMAnAlgBxkgHSgAJiBaYgHyNNKwDNiBtcxgkAEgAkB5AWQCiAegB0YMAEMwACzhDUMCQHMsUJQH0AJlgwjFugB4AvdsQC6AbmIxZUGrVJwZEYpAC2OLABtVS4vQwIN1cpWXsHNwkAazhyACsAZmIANQBeAHdiAmySB1oAMnz-VU1iACNVLJy8gqLJGHLVIX0lOBhUIQwJN1wOjBgEgA5BhPlpCQAmAFYANlQAVzdUcNo4L1Q4FYcnaRd3Tx81awgIda3aSJj4pJgT9aF971js84dC4kvYskSUjOPTmQSJqPLybXIOehYewUajg4gMZjEACErCynF4glE4lCcgUyl8Wh0egkhhMIHMVhsCHOOxcCgg2COILgpVup1Q8JIkhkYJq+nIYEYFDxKjUhIwVGIxgF5CSFAAfGjuPxhGJubJ5IpRRptLp9CJjOwVpCiEQQAAaEAQPBYaCoZCgEmBdIABRJCHtKAkADcIFhNBaQGUumAYjAAMo4KS+ZAwDDzOCW1QbfourpKSLIegSdaJkBxVAGABCIbD4e6cAAMqo4FmcxtLVGMCnkCAAJ7dLyBnCMmAAdX9NmQgwADI3Ahs+10cK2e3AU97a5aMHAAI7zHRwNPKTNIbO5y0bHqx+N51C+UEARXmEHgdYPIEUZQHmiHSAmlrjQMOSgAwkFdxAKBoCXEB5g2AAVCQyk9fcNgAX3goA)
```sh
--parser yaml
```

**Input:**
```yaml
script:
  - |
    if [ -f "$HOME/.ccache/staging_dir.tar.xz" ]; then
      echo compiling from cache
      make -j3 V=w \
        && find bin \
        && cat bin/targets/ramips/rt3883/sha256sums
      else
        echo compiling tools
        make -j3 tools/compile \
          && make -j3 V=w toolchain/compile
      fi
  - |
    if [ ! -f "$HOME/.ccache/staging_dir.tar.xz" ]; then
      echo storing compiled tools in cache
      tar -cf - staging_dir | xz -c -3 - > "$HOME/.ccache/staging_dir.tar.xz"
    fi


```

**Output:**
```yaml
script:
- |
  if [ -f "$HOME/.ccache/staging_dir.tar.xz" ]; then
    echo compiling from cache
    make -j3 V=w \
      && find bin \
      && cat bin/targets/ramips/rt3883/sha256sums
    else
      echo compiling tools
      make -j3 tools/compile \
        && make -j3 V=w toolchain/compile
    fi
- |
  if [ ! -f "$HOME/.ccache/staging_dir.tar.xz" ]; then
    echo storing compiled tools in cache
    tar -cf - staging_dir | xz -c -3 - > "$HOME/.ccache/staging_dir.tar.xz"
  fi

```

website/static/new-worker.js
```diff
@@ -62,6 +62,12 @@ var parsers = {
   get vue() {
```
duailibe: There's only `website/static/worker.js` now :)

ikatyang:
@nevir it looks like a bug in `yaml`, opened eemeli/yaml#2.

ikatyang:
@suchipi @azz

I don't have strong opinion on either style, do you know which one is more popular?

indent:

```yaml
a:
  - b
  - c
```

do not indent:

```yaml
a:
- b
- c
```

hawkrives:
@ikatyang Personally, I indent my YAML lists (and I find it quite difficult to read un-indented ones) but I have a suspicion that not indenting is the more common practice?

eemeli:
Heya. `yaml` developer here. Just wanted to point out that there are a number of YAML syntax errors that we catch at a level above the raw AST that's used here (YAML is complicated, and the AST parser tries not to be). In other words, the parser will make apparent sense out of input that isn't valid YAML.

To catch those errors as well, I'd recommend passing the AST to `new YAML.Document().parse(ast)`, and looking through the document's `errors` and `warnings` arrays. Each error should include a `source` AST node for the error or warning.

robert-john-small:
My preference is always to indent, it makes things more readable and seems to be the push from some of the YAML consumers at least in the infrastructure automation world.
```
thing:
  - other_thing
  - something:
    more: things
  - right: we like things
  - imagination: off
```

ascorbic:
Nice work. Are the fixes from eemeli/yaml#2 merged in? I'm still getting the same "Sequence items are not allowed on the same line with map keys" errors in the playground.

ikatyang:
Not yet, I'm still waiting for the patch release for eemeli/yaml#10.

lipis:
I think it was [released](https://github.com/eemeli/yaml/releases) :)

evilebottnawi:
```yaml
a:
  - b
  - c
```

is more readable, example:
```yaml
before_script:
  - command
  - command
  - command
stages:
  - test
  - test
  - test
other:
  - other
  - other
  - other
```
vs
```yaml
before_script:
- command
- command
- command
stages:
- test
- test
- test
other:
- other
- other
- other
```

lipis:
Yes, with indentation is much better!

ikatyang:
We now indent sequence in mapping since it's [more readable](https://github.com/prettier/prettier/pull/4563#issuecomment-397812369).

And also I rewrote the mappingItem logic so that it fully respects the `printWidth`, we previously hardcoded which node type should use explicit or implicit key, it now only uses explicit key for multiline key, for example:

**Prettier pr-4563**
[Playground link](https://deploy-preview-4563--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBtAhpr2e75gXSQAJ8zyLMAdKGjShxo0x18mutr7n8533jRAAaEBAAOMAJbQAzslDoATkogB3AArKE8lOgBuEKQBMRIAEZL0YANZwYAZXHWpUAObIYSgK5xRr2TglGA0rNwBbdGQAM3QAG0DRACtZAA8AIStbewd0cLgAGVc4GPjEkGclQKVkEABPPLizcSVXGAB1ExgAC2QADgAGURaIQParcVqWuGr9EtElOABHbylF0PQIqKRYhL8QQPCpTx992Vc3OLgARW8IeFK90Rh0c07jHuQAJmerKTiLgBhCDhSK1KDQeYgbyBAAqr10u0CAF9kUA)
```sh
--parser yaml
```

**Input:**
```yaml
[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa]: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa]: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa]: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

```

**Output:**
```yaml
[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa]: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa]:
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

? [
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
  ]
: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

```

ikatyang:
This PR should be ready for review, anyone wants to take a look?

azz:
> is more readable

I don't know about more readable, certainly more indented... I think we should seek wider consensus before shipping this. (Independent from merging this PR of course)

ikatyang:
Probably some twitter poll like the [markdown one](https://twitter.com/PrettierCode/status/981451100262936576)?

scripts/build/config.js
```diff
@@ -66,6 +67,14 @@ const parsers = [
   {
     input: "src/language-html/parser-parse5.js",
     target: "node"
+  },
+  {
+    input: "src/language-yaml/parser-yaml.js",
+    target: "universal",
+    bundler: "webpack",
```
duailibe: This doesn't work with rollup?

duailibe: Rollup didn't work?

scripts/build/config.js
```diff
@@ -66,6 +67,14 @@ const parsers = [
...
+    plugins: [
+      require.resolve("./babel-plugins/replace-array-includes-with-indexof")
```
duailibe: 1) I thought babel was including polyfills, looks like it isn't. Maybe we should set it up?

2) This should probably be named `babelPlugins` to be more specific about what kind of plugin this is (we could also have rollup or webpack plugins)

website/static/worker.js
```diff
@@ -70,6 +70,12 @@ var parsers = {
   get vue() {
     importScriptOnce("lib/parser-vue.js");
     return prettierPlugins.vue.parsers.vue;
+  },
+
+  // YAML
+  get yaml() {
+    importScripts("lib/parser-yaml.js");
```
duailibe: This should be `importScriptOnce` too

scripts/build/config.js
ikatyang:
```
TypeError: lines_and_columns_1.default is not a constructor
```

duailibe: Possibly related to this: https://github.com/rollup/rollup-plugin-commonjs/issues/160

`line-and-colum` is shipped with both CJS and ESM versions:

```js
// ESM
export default LineAndColumns;

// CJS
exports["default"] = LineAndColumns;
```
The distributed `yaml-unist-parser` version has a `require("line-and-columns")["default"]`. But Rollup uses the ESM version by default and by its semantics, when using ESM `require("line-and-columns") === LineAndColumns`.

I aliased `line-and-columns` in rollup bundle to point to the CJS version.

ascorbic:
Does this preserve each input nodes' quoting style rather than enforcing a common style?

ikatyang:
What do you mean quoting style?

ascorbic:
SIngle quote, double quote, plain. http://yaml.org/spec/1.2/spec.html#id2786942

ikatyang:
We use the minimal escape style then `options.singleQuote` for `quoteDouble`/`quoteSingle`, and we won't change `plain` into `quoteDouble`/`quoteSingle`.

mohsen1:
Can you extract out the YAML AST parser into a package so I can use it for my projects?
https://github.com/mohsen1/yawn-yaml/

Thanks in advance!

hawkrives:
@mohsen1 This PR just uses `yaml` and `yaml-unist-parser`, I believe, which are already separate packages. They're linked in the first post in this thread.

mohsen1:
There is a shortage of YAML AST parsers that actually works right now. I can copy [parser-yaml.js](https://github.com/ikatyang/prettier/blob/98a5e65691141cdf2bc92f21aa46b516fe7ceeb3/src/language-yaml/parser-yaml.js) into my project but for the sake of maintainability and ability to contribute to upstream it would be nice to have that file and related tests in a package. I'm ok if it's against your project goals. 

Very excited for this btw! Amazing work!

azz:
I'm not sure whether it's a guaranteed to be maintained API (I think it is since we supported standalone?), but you can do `require('prettier/parser-yaml')(yamlString)`

duailibe:
FTR it's actually something like `require('prettier/parser-yaml').parsers.yaml.parse(string)`

ikatyang:
I'll merge this PR tomorrow if there's no further comment for something needs to be addressed here.

lipis:
The most _readable_ version is with the indent.. and since we don't want to have options let's release it like that. My bet is that there are not going to be tooooo many complains. Just the usual numbers I would guess :)

ikatyang:
I'd like to avoid things like #4251 in the future, let's see how people think about it. Opened #4723 to vote for the indentation style for sequence in mapping.

JPinkney:
Is there a version of prettier with the yaml stuff released on npm? I'd like to use the formatting for https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml but am getting parser not found when trying to do prettier.format(text, { parser: "yaml" })

ikatyang:
We haven't released it yet (v1.14), but you can install it from GitHub for now:

```sh
npm install prettier/prettier
```

I'm not sure how vscode extension works but if you'd like to use the production build (bundled version), you could clone this repo then `yarn && yarn build`, the output will be in `./dist`.

dploeger:
I think, the support table inside the code is still missing the Yaml-feature, right? I noticed, because I couldn't get IntelliJ to run on YAML, even when configured the parser for all *.yaml files.

ikatyang:
If you're using the current master, you'll have to specify `--parser yaml` manually to get it worked since it's still an unreleased feature. (Any published v1.13.x does not support YAML.)

https://github.com/prettier/prettier/blob/dc362c212d068a2714dc3e39c33709e506a2560a/src/language-yaml/index.js#L7-L13

dploeger:
@ikatyang Yes, I know and I did that, but Webstorm/IntelliJ/etc. exclusively rely on the GetSupportInfo and show a "file (...) has an unsupported type" and don't even start prettier.

ikatyang:
Uh, then you probably have to modify the version in prettier's `package.json` to be something like `1.14.0-dev` to get it worked.

dploeger:
@ikatyang Ah, sorry. I mistakenly looked at the source code for an 1.9.2 version on my system. And there was a support table missing Yaml, but it is handled using plugins now. Hmmm... So GetSupportInfo should actually show Yaml? I'll check that out. Maybe it's an issue with the IntelliJ plugin after all.

ikatyang:
There's a `showUnreleased` option in `getSupportInfo`, you should be able to format YAML if IntelliJ enabled it, otherwise you'll still have to modify the version in prettier's `package.json`.

dploeger:
Hm. Changing the `package.json` version didn't work, but I simply hacked the Yaml index.js and modified the "since" parameter. 😆

PR #4565, opened 2018-05-27
Use ES6+ language features

j-f1:
[node.green](https://node.green) has helpful information for seeing what’s supported.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -23,6 +23,7 @@ rules:
   - never
   prefer-arrow-callback: error
   prefer-const: error
+  prefer-destructuring: error
   react/no-deprecated: off
   react/prop-types: off
   strict: error
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -25,6 +25,7 @@ rules:
   prefer-arrow-callback: error
   prefer-const: error
   prefer-destructuring: error
+  prefer-spread: error
   react/no-deprecated: off
   react/prop-types: off
   strict: error
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -18,6 +18,7 @@ rules:
   no-unneeded-ternary: error
   no-useless-return: error
   no-var: error
+  object-shorthand: error
   one-var:
   - error
   - never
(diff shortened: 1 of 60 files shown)


src/cli/util.js
```diff
@@ -567,7 +566,7 @@ function createOptionUsageType(option) {
 }
 
 function flattenArray(array) {
-  return [].concat.apply([], array);
+  return [].concat(...array);
```
azz: `array.smoosh()` 😆

tests_integration/plugins/extensions/plugin.js
```diff
@@ -1,7 +1,10 @@
 "use strict";
 
-const prettier = require("prettier/local");
-const concat = prettier.doc.builders.concat;
+const {
```
azz: Integration tests need to run on node 4, so this might not work.

website/playground/markdown.js
```diff
@@ -33,23 +33,22 @@ function formatMarkdown(
...
+      return parser;
   }
 }
 
 function formatCLIOptions(cliOptions) {
   return cliOptions
     .map(option => {
-      const name = option[0];
-      const value = option[1];
+      const [name, value] = option;
```
azz: Can inline this one into the `map(([name, value]) =>`

scripts/build/rollup-plugins/executable.js
```diff
@@ -20,7 +20,7 @@ module.exports = function() {
       const source = fs.readFileSync(id, "utf-8");
       const match = source.match(/^\s*(#!.*)/);
       if (match) {
-        banner = match[1];
+        [, banner] = match;
```
azz: Old one is shorter and (IMO) clearer

.eslintrc.yml
```diff
@@ -18,11 +18,14 @@ rules:
   no-unneeded-ternary: error
   no-useless-return: error
   no-var: error
+  object-shorthand: error
   one-var:
   - error
   - never
   prefer-arrow-callback: error
   prefer-const: error
+  prefer-destructuring: error
```
azz: Probably turn this one off as there are already 10 exceptions?

j-f1: I’ll change it to [stop looking at array accesses](https://eslint.org/docs/rules/prefer-destructuring#options).

j-f1:
@duailibe Yep, I checked that all of the features I used were supported on Node 6. If I missed anything, please point it out.

duailibe:
I honestly don't know which are supported or not, it was just a reminder =)

azz:
Maybe we should change AppVeyor to use Node 6 (currently Node 8). I wonder if it will segfault?

PR #4568, opened 2018-05-27
fix spelling

dnalborczyk:


duailibe:
I've noticed the whole sentence was weird.

Thanks!

PR #4586, opened 2018-05-29
Support specifying parser/printer mode via name of custom parser

### LINKED ISSUE #3150: Custom parser not compatible with writers
When following the [custom-parser-api](https://github.com/prettier/prettier#custom-parser-api) example from the docs with a non-js file format, it fails on selecting the correct writer:

```
prettier.format(
  '.foo { color: red; }',
  {
    // parser: 'scss'
    parser: (text, parsers) => {
      const ast = parsers.scss(text, parsers, { parser: 'scss' });
      return ast;
    }
  }
)
```

```
node_modules/prettier/index.js:8014
      throw new Error("unknown type: " + JSON.stringify(n.type));
      ^

Error: unknown type: "css-root"
    at genericPrintNoParens (node_modules/prettier/index.js:8014:13)
```

It checks `options.parser`, but since that's a function, it will fallback to the default writer:
https://github.com/prettier/prettier/blob/ec8deb6c0baaa3eb17a7236f9c1cd9e1a8063c44/src/printer.js#L53-L68

I believe the option to pass a function (#1783) was added before the custom languages, but I think this feature needs a fresh look. Some suggestions:
* Either go back to a separate option like #1710 (astTransform)
* Or add a `printer` option that can be set to a string (not needed when parser is a string or can be detected on filetype)
* Or allow the response of the parser function to return an object with the options that the writer can use.

xixixao:
Fixes #3150

This PR doesn't include required documentation (can someone pick that up?).

### DIFF
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -492,12 +492,11 @@ function printString(raw, options, isDirectiveLiteral) {
     canChangeDirectiveQuotes = true;
   }
 
-  const enclosingQuote =
-    options.parser === "json"
-      ? double.quote
-      : shouldUseAlternateQuote
-        ? alternate.quote
-        : preferred.quote;
+  const enclosingQuote = isParser(options, "json")
+    ? double.quote
+    : shouldUseAlternateQuote
+      ? alternate.quote
+      : preferred.quote;
 
   // Directives are exact code unit sequences, which means that you can't
   // change the escape sequences they use.
@@ -518,11 +517,15 @@ function printString(raw, options, isDirectiveLiteral) {
     rawContent,
     enclosingQuote,
     !(
-      options.parser === "css" ||
-      options.parser === "less" ||
-      options.parser === "scss"
+      isParser(options, "css") ||
+      isParser(options, "less") ||
+      isParser(options, "scss")
     )
   );
+}
+
+function isParser(options, name) {
+  return options.parser === name || options.parser.printer === name;
 }
 
 function makeString(rawContent, enclosingQuote, unescapeUnnecessaryEscapes) {
@@ -795,6 +798,7 @@ module.exports = {
   getIndentSize,
   printString,
   printNumber,
+  isParser,
   hasIgnoreComment,
   hasNodeIgnoreComment,
   makeString,
--- a/src/doc/doc-printer.js
+++ b/src/doc/doc-printer.js
@@ -468,7 +468,7 @@ function printDocToString(doc, options) {
                   if (
                     out.length &&
                     typeof out[out.length - 1] === "string" &&
-                    (options.parser !== "markdown" ||
+                    (!util.isParser(options, "markdown") ||
                       // preserve markdown's `break` node (two trailing spaces)
                       !/\S {2}$/.test(out[out.length - 1]))
                   ) {
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -404,8 +404,8 @@ function needsParens(path, options) {
         parent.type === "ExpressionStatement" &&
         // TypeScript workaround for eslint/typescript-eslint-parser#267
         // See corresponding workaround in printer.js case: "Literal"
-        ((options.parser !== "typescript" && !parent.directive) ||
-          (options.parser === "typescript" &&
+        ((!util.isParser(options, "typescript") && !parent.directive) ||
+          (util.isParser(options, "typescript") &&
             options.originalText.substr(options.locStart(node) - 1, 1) === "("))
       ) {
         // To avoid becoming a directive
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -14,6 +14,7 @@ const {
   getStringWidth,
   printString,
   printNumber,
+  isParser,
   hasIgnoreComment,
   skipWhitespace,
   hasNodeIgnoreComment,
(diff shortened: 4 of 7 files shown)


src/common/util.js
```diff
@@ -518,13 +517,17 @@ function printString(raw, options, isDirectiveLiteral) {
...
-      options.parser === "less" ||
-      options.parser === "scss"
+      isParser(options, "css") ||
+      isParser(options, "less") ||
+      isParser(options, "scss")
     )
   );
 }
 
+function isParser(options, name) {
```
azz: Can you rename this to `isLanguage`? It aligns with our plans to introduce that concept.

src/common/util.js
```diff
@@ -518,13 +517,17 @@ function printString(raw, options, isDirectiveLiteral) {
...
+  return options.parser === name || options.parser.printer === name;
```
azz: Can you call this `astFormat` instead?

src/common/util.js
azz: Ideally the call would be `isLanguage(options, 'JSON')` so we wouldn't need to check `'json-stringify', `json5`, `json`. But it is ok for now.

azz:
Thanks for doing this!

azz:
We can probably supersede this API with the plugins API, as you can now do this:

```js
const parserPlugin = {
  parsers: {
    "custom-css-parser": {
      parse(text, { postcss }, opts) {
        const ast = postcss(text);
        mutate(ast);
        return ast;
      },
      astFormat: "postcss"
    }
  }
};

prettier.format("code", {
  plugins: [parserPlugin],
  parser: "custom-css-parser"
})

```

azz:
Let me know if that comment works for your use case. If so we can document it and remove the custom parser API in V2.

src/common/util.js
azz: Or just check `astFormat`

xixixao:
So on the parser plugin we call it `astFormat` but in code we call it `isLanguage` - shouldn't the 2 be the same?

azz:
Sorry, was conflating two different concepts. Basically I think we should replace checks for `options.parser === "X"` with:

- `options.language === "X` when the printer needs context on the _language_ that is being printed. For example, the following would be `options.language === "json"` and `options.language === "typescript"` because it is a printer decision that JSON files and TypeScript class properties should never have quotes removed. Most checks will fall into this category.

  
  https://github.com/prettier/prettier/blob/c7d40b0ed958b0e345511857a55e9cac87f18131/src/language-js/printer-estree.js#L3298-L3304

- `options.astFormat === "X"` when the check is determining the type of AST. For example, this following could be `options.astFormat === "postcss"`:
    https://github.com/prettier/prettier/blob/c7d40b0ed958b0e345511857a55e9cac87f18131/src/common/util.js#L517-L525

Note that you'll need to add `rawOptions.language` here somewhere:
https://github.com/prettier/prettier/blob/c7d40b0ed958b0e345511857a55e9cac87f18131/src/main/options.js#L59 and to `hiddenDefaults`. (Maybe use `language.name.toLowerCase()`? from `getSupportInfo()`?)

xixixao:
Hmm, I'm not sure I have all the necessary context, the distinction between "postcss" and "typescript" or "json" is still fuzzy to me. Do you want to proceed with the changes I made here, and do the correct naming on top of them?

azz:
There's two core components to Prettier.

* A parser (in: code, out: ast):
  - Language being parsed: `language`: (e.g. `javascript`), inferred from extension.
  - Name of the parser: `parser` (e.g. `babylon`), selected by association with `language`.
  - Type of the AST: `astFormat` (e.g. `estree`)
* A printer (in: ast, out: doc):
  - Type of the AST: `astFormat` (e.g. `estree`), printers are selected based on this criteria.
  - Language being printed: `language`: (e.g. `javascript`)

ikatyang:
I'm thinking of if we could make language to be something like:

- `--language javascript` implies `--parser babylon` + `--printer estree` (the name of printer, not astFormat)
- `--language typescript` implies `--parser typescript` + `--printer estree`

where the output astFormat from parser should match the input astFormat from printer, otherwise throw an error, so that we could replace one of them by using a custom parser (or printer?), e.g. `--language javascript --parser ./my-custom-parser.js`.

And also, the `json-stringify` parser currently has to output a different `astFormat` (i.e. `estree-json`) other than `estree` since we currently match printers using the `astFormat`, but it's acutally just same as the `estree`, I think it'd be better to be something like:

- `--language json` implies `--parser babylon` + `--printer estree`
- `--language json5` implies `--parser babylon` + `--printer estree`
- `--language json-stringify` implies `--parser babylon` (output: estree) + `--printer json-stringify` (input: estree)

azz:
I think `json-stringify` should be a printer name, not a language.

azz:
So `--language json` infers `estree-json` printer, `--language json --printer json-stringify` would override.

azz:
But all that aside, the requested changes for this PR is to rename it from `isParser` to `isLanguage` and check the `options.language` property instead. `options.language` should be initialized when the parser inference is happening.

xixixao:
So then the API will be
```jsx
prettier.format({
  parser(text) {...},
  language: "json",
})
```
correct?

azz:
If you can get that working, that'd be great!

j-f1:
What’s the status of this? Should we fix the conflicts and merge, or close without merging?

xixixao:
Can someone pick this up? I won't have time in the near future, sorry.

lipis:
That's a bit outdated.. please ping back if it should remain open.

PR #4589, opened 2018-05-29
Set AppVeyor to Node 6

azz:
Let's see if it passes...


j-f1:
> `error Command failed with exit code 3221225477.`

I guess that’s the Windows version of a segfault :/

azz:
Apparently not. 😢

PR #4593, opened 2018-05-29
fix(markdown): better handling for trailing spaces

ikatyang:
- preserve trailing spaces in html, excluding html comment
- `doc-printer` is now unrelated to `options.parser`
- fix some cases that two trailing spaces are mis-considered as `break`


ikatyang:
Does anyone want to take a look at this?

duailibe:
Sorry this went unnoticed

PR #4625, opened 2018-06-03
feat(typescript): support unknown type

ikatyang:
Ref: Microsoft/TypeScript#24439


tests/typescript/compiler/badArrayIndex.ts
```diff
@@ -1 +0,0 @@
-var results = number[];
```
ikatyang: This is a syntax error now.

```
SyntaxError: An element access expression should take an argument. (1:22)
> 1 | var results = number[];
    |                      ^
  2 |
```

src/language-js/printer-estree.js
```diff
@@ -2833,6 +2833,8 @@ function printPathNoParens(path, options, print, args) {
       return "string";
     case "TSUndefinedKeyword":
       return "undefined";
+    case "TSUnknownKeyword":
+      return "unknown";
```
j-f1: Should we extract all of these “constant” nodes to an object to reduce the number of `case` statements needed?

PR #4631, opened 2018-06-04
Introduce Release script

duailibe:
This is an attempt to automate the steps in [the checklist](https://github.com/prettier/prettier/wiki/Release-Checklist) (or most of what can be automated).

It's heavily based on React's release script.

The commits can be reviewed separately to focus on each step if desired.


scripts/release/steps/bump-prettier.js
```diff
@@ -0,0 +1,28 @@
...
+const { exec } = require("child-process-promise");
+const { logPromise } = require("../utils");
+
+async function format() {
+  await exec("yarn lint --fix");
+  await exec("yarn lint-docs --fix");
+}
+
+async function commit(version) {
+  await exec(`git commit -am "Bump Prettier dependency to ${version}"`);
```
j-f1: Since you’re using `exec` here, someone could put a string like `" && rm -rf / && echo "` as `version` and run arbitrary code (not sure how dangerous this is).

scripts/release/steps/bump-prettier.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+module.exports = async function({ dry, version }) {
+  if (dry) {
+    return;
+  }
+
+  await logPromise(
+    "Installing Prettier",
+    exec(`yarn add --dev prettier@${version}`)
```
j-f1: Same as above.

scripts/release/steps/check-schema-store.js
```diff
@@ -0,0 +1,47 @@
...
+      "In a new session, run `node scripts/generate-schema.js` and copy the output.";
+  }
+
+  console.log(
+    dedent(chalk`
+      {yellow.bold A manual step is necessary.}
+
+      The schema in {yellow SchemaStore} needs an update.
+      - ${instruction}
+      - Go to {cyan.underline ${EDIT_URL}}
```
j-f1: If you have the new schema in your clipboard and your terminal doesn’t linkify links, you’d have to retype the URL into your browser. Maybe use [`opn`](https://www.npmjs.com/package/opn) to open the URL after a short delay or the user hits enter?

scripts/release/steps/publish-to-npm.js
```diff
@@ -0,0 +1,36 @@
...
+
+      {bold.underline Test the new releae}
+      - In a new session, run {yellow npm i prettier@latest} in another directory
+      - Test the API and CLI
+
+      After that, we can proceed to bump this repo's Prettier dependency.
+      Press any key to continue.
+    `)
+  );
+  await keypress();
```
j-f1: Is there a way to make the user press enter so they don’t accidentally continue the process?

scripts/release/steps/update-changelog.js
```diff
@@ -0,0 +1,88 @@
...
+      The script has updated the file {bold CHANGELOG.md} with a draft of ${changes}, you can make any changes now if necessary.
+      You don't need to commit the file, the script will take care of that.
+    `)
+  );
+  if (warning) {
+    console.log("\n" + warning);
+  }
+  console.log("\nPress any key to continue");
+
+  await keypress();
```
j-f1: Same here — make it require the enter key so the person doesn’t accidentally continue.

scripts/release/steps/validate-new-version.js
```diff
@@ -0,0 +1,14 @@
...
+const chalk = require("chalk");
+const semver = require("semver");
+
+module.exports = async function({ version, previousVersion }) {
+  if (!semver.valid(version)) {
+    throw Error(`Invalid version specified`);
+  }
+
+  if (!semver.gt(version, previousVersion)) {
+    throw Error(`Version ${chalk.white(version)} has already been published`);
```
j-f1: If the user has a light theme, the version number may not appear (unless the error message is shown with a colored background).

scripts/release/utils.js
```diff
@@ -0,0 +1,70 @@
...
+    })
+    .catch(err => {
+      process.stdout.write(`${FAIL}\n`);
+      throw err;
+    });
+}
+
+async function execYarn(command) {
+  try {
+    await exec(`yarn --silent ${command}`);
```
j-f1: Same possible injection issue here.

scripts/release/steps/publish-to-npm.js
```diff
@@ -0,0 +1,36 @@
...
+const dedent = require("dedent");
+const { exec } = require("child-process-promise");
+const { logPromise, keypress } = require("../utils");
+
+module.exports = async function({ dry, version }) {
+  if (dry) {
+    return;
+  }
+
+  await logPromise("Publishing to NPM", exec("npm publish", { cwd: "./dist" }));
```
azz: *`npm` 😉

scripts/release/steps/bump-prettier.js
duailibe: We're the ones who are gonna use this, so I guess it's on us to be careful...

In this specific case, the script checks if `version` is valid using semver.

scripts/release/steps/bump-prettier.js
duailibe: Same as above

scripts/release/steps/publish-to-npm.js
duailibe: Yeah it's possible

scripts/release/steps/validate-new-version.js
duailibe: Oh that was left when I was printing the error all in red.. gonna change this

azz:
💯 💯 💯

scripts/release/steps/bump-prettier.js
lydell: Why not use `spawn` instead?

```js
await spawn("git", ["commit", "-am", `Bump Prettier dependency to ${version}`])
```

lydell: Why not use `spawn` instead?

```js
await spawn("git", ["commit", "-am", `Bump Prettier dependency to ${version}`])
```

scripts/release/steps/update-changelog.js
```diff
@@ -0,0 +1,88 @@
...
+    const year = date.getFullYear();
+    const month = new String(date.getMonth() + 1).padStart(2, "0");
+    const day = new String(date.getDate()).padStart(2, "0");
+    const blogPostFile = `website/blog/${year}-${month}-${day}-${version}.md`;
+    const blogPostPath = `${year}/${month}/${day}/${version}.html`;
+
+    releaseNotes = `- [Release Notes](https://prettier.io/blog/${blogPostPath})`;
+
+    if (!fs.existsSync(blogPostFile)) {
+      warning = chalk`Make sure the blog post file {yellow ${blogPostFile}} will be created.`;
```
suchipi: This message is confusing me- is the user supposed to create a blog post before running the script? Or will the script create a skeleton that the user is supposed to edit before continuing?

suchipi:
Awesome!! 🎉

scripts/release/steps/update-changelog.js
duailibe: @suchipi Neither. The user just needs to make sure this file exists or will exist in the future, just so the link in CHANGELOG works.

How can I make it more clear?

I also added this because of the last blog post in which I committed with the wrong filename.

suchipi: We might want to die here instead and ask the user to create the file before running the script, because a blog post should be required for non-patch releases.

scripts/release/steps/update-changelog.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+  } else {
+    // Changelog should be the most important PRs, let's add everything
+    // and ask for the releaser to edit the file later.
+    changes = "merged PRs since last release";
+    const prs = await prsMergedSince({
```
suchipi: prs-merged-since makes a ton of GitHub API requests (when running against large repos like prettier), and GitHub limits the number of requests you can make per hour if you don't use an API Token. I think we can get through one or two runs of prs-merged-since against prettier without a token, but if someone had to run this script a couple times in a row because they were debugging stuff, they might hit the issue. prs-merged-since will read the token from the environment variable `GITHUB_API_TOKEN`; we could print a warning here if it's not set, maybe? Or maybe prs-merged-since should be the one to print the warning instead.

scripts/release/steps/update-changelog.js
duailibe: @suchipi I don't understand, why should the script die here?

I don't think the blog post should block the release. We'll either merge the blog post before or after editing `CHANGELOG.md` anyway. So the two scenarios are:

- post does not exist because PR is still open
  - runs the script
  - updates `CHANGELOG.md` and prints a warning for what the blog post path should be
  - publish to npm
  - verify blog post path in PR
  - merge blog post PR
  - tweet

- post does not exist because path is wrong
  - runs the script
  - updates `CHANGELOG.md` and prints a warning for what the blog post path should be
  - publish to npm
  - fix blog post path
  - commit
  - tweet

- post exists
  - runs the script
  - updates `CHANGELOG.md`
  - publish to npm
  - tweet

Also I initially want this script to run completely at once so we don't have to remember the order of the "script commands" we need to run. Then we can add a few `--skip-X` if we see an use case where it's useful

scripts/release/steps/update-changelog.js
duailibe: I thought about that, but patch releases aren't likely to have too many PRs so I *think* we're safe.

And yes, we can add a warning, let me know what you think.

suchipi: I remembered that when using the JS API, you pass the token as `githubApiToken`, not using env.

If I remember correctly, we still have to request every PR in the repo, even for patch releases, due to a limitation of the GitHub API.

I think we should pass `process.env.GITHUB_API_TOKEN` into `prsMergedSince` here if it's present, and if it isn't, print a warning.

scripts/release/steps/update-changelog.js
suchipi: Alright, I understand after reading through your examples. I agree then, it shouldn't die. Let's just change the message to:

```

scripts/release/steps/check-schema-store.js
duailibe: I decided to remove the clipboard part entirely and just print what the user should do.

There's still value in just checking if the schema changed or not, which we weren't doing at all

scripts/release/steps/update-changelog.js
duailibe: Oh I'm already passing the `process.env.GITHUB_API_TOKEN`, just didn't document it anywhere.

I included the warning

duailibe:
I think I addressed all the points raised so far

m1ck2:
Was this enough to fix problem in script so it will run yarn as well now.

suchipi:
@m1ck2 I don't understand, could you please open a new issue?

scripts/release/steps/bump-prettier.js
j-f1: If you’re interested, I [created a simple `spawn` template tag](https://runkit.com/j-f/5b155a812ac4a80011ca80e2) that you can use so you don’t have to split the arguments individually.

PR #4653, opened 2018-06-08
fix(playground): no unnecessary import

### LINKED ISSUE #4652: [bug] prettier playground compiles too slow when it came across syntax it doesn't know
**Prettier 1.13.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBDTBeTAAwB0oATASzQAcAbbATyUwDNa4APAblICNswAawDmAJwgBXcs1FwyPKADpsvMJmCZKNek0y9aEIQsx6BI8VLLMA7gAsK8BQF9ShLiAA0ICNRgV0yKDYouLWAArBCGjIINgAbhAUZJ4gvKJmcDAAytQCFFDCyDCiEnBe+WhwojBh6cIAttjILNi0lV4AVmgcAELpQplZ2PVwADL5cM2t7SC5opWiMWwQ1inUovkwAOpJMLbIABwADF7rEJVb6dQx63ALcZNesgCOEhSytdgNTUgtbWUgSr1ChFEoAtD5YTsACKEgg8Cm-y8MBUOzIe2QACZkekKLRIQBhCD1RoxKDQR4gCSVAAqKmiv2mcCcTiAA)
```sh
--parser flow
```

**Input:**
```jsx
const a = `
display: flex;
background: red;
.abc { display: block;
  background: white;
}
`;
```

**Output:**
```jsx
const a = `
display: flex;
background: red;
.abc { display: block;
  background: white;
}
`;

```

**Expected behavior:**

When I change the code at left, the right part compiles too slow....

beside, how can I format a code block using a specific syntax like less/scss

ikatyang:
Fixes #4652

### DIFF
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -2,7 +2,7 @@
 /* eslint no-var: off, strict: off */
 /* globals prettier prettierPlugins */
 
-var imported = {};
+var imported = Object.create(null);
 function importScriptOnce(url) {
   if (!imported[url]) {
     imported[url] = true;
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -2,65 +2,73 @@
 /* eslint no-var: off, strict: off */
 /* globals prettier prettierPlugins */
 
+var imported = {};
+function importScriptOnce(url) {
+  if (!imported[url]) {
+    imported[url] = true;
+    importScripts(url);
+  }
+}
+
 // this is required to only load parsers when we need them
 var parsers = {
   // JS - Babylon
   get babylon() {
-    importScripts("lib/parser-babylon.js");
+    importScriptOnce("lib/parser-babylon.js");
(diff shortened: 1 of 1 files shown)


website/static/worker.js
```diff
@@ -2,65 +2,73 @@
 /* eslint no-var: off, strict: off */
 /* globals prettier prettierPlugins */
 
+var imported = {};
+function importScriptOnce(url) {
+  if (!imported[url]) {
+    imported[url] = true;
```
j-f1: Perhaps use an array, `Object.create(null)`, or a set here to avoid bugs we’ve encountered with using an object as a map on the past?

ikatyang: I'll change it to use `Object.create(null)`, though I think it's almost impossible to encounter the same issue since we probably won't have some parser named something like `hasOwnProperty`.

PR #4663, opened 2018-06-10
Removed a "s"

yanishoss:
# Removed a "s"

| Location                              | Line | Issue                                                                                                       |
|:------------------------------|------|-----------------------------------------------------------------------------|
| *docs/option-philosophy.md* | 12 | A *"s"* after the 👍was **removed**, it was **alone** not should **not be there.** |


j-f1:
I feel like this is fine as-is — there are multiple 👍 reactions on that issue.

PR #4668, opened 2018-06-11
Run tests using the standalone build in an empty context

duailibe:
This uses the [`vm.Script`](https://nodejs.org/api/vm.html) API to run the standalone build in an empty context. This is mostly to validate that our build is stripping all the "native" dependencies (native modules and env vars).

It might not be necessary to run *all* `run_spec` tests, but I figured why not?

### DIFF
--- a/tests_config/require_standalone.js
+++ b/tests_config/require_standalone.js
@@ -18,7 +18,7 @@ const sources = [
 );
 
 const sandbox = vm.createContext();
-vm.runInContext(sources.join(""), sandbox);
+vm.runInContext(sources.join(";"), sandbox);
 
 module.exports = {
   formatWithCursor(input, options) {
--- a/tests_config/require_standalone.js
+++ b/tests_config/require_standalone.js
@@ -1,0 +1,52 @@
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+const vm = require("vm");
+
+const files = [
+  "standalone.js",
+  "parser-babylon.js",
+  "parser-flow.js",
+  "parser-typescript.js",
+  "parser-postcss.js",
+  "parser-graphql.js",
+  "parser-markdown.js",
+  "parser-vue.js"
+];
+const base = files
+  .map(filename =>
+    fs.readFileSync(path.join(process.env.PRETTIER_DIR, filename), "utf-8")
+  )
+  .join("\n");
+
+const formatScript = new vm.Script(
+  base +
+    `$$$g.output = (function(input, options) {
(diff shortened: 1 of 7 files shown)


tests_config/require_standalone.js
```diff
@@ -0,0 +1,59 @@
...
+    const g = { input, options, massage };
+    parseScript.runInNewContext({ $$$g: g });
+    return g.parsed;
+  }
+}
+
+if (process.env.NODE_ENV === "production") {
+  module.exports = makeStandaloneRunner();
+} else {
+  module.exports = null;
```
suchipi: Is this conditional still needed?

duailibe: @suchipi I guess not... might just let it break if you `require` this in a wrong "context"

duailibe:
Just found a bug! 😂

tests_config/require_standalone.js
```diff
@@ -20,33 +20,34 @@ const base = files
...
+        `output = prettier.__debug.parse(
+          ${JSON.stringify(input)},
+          Object.assign({ plugins: prettierPlugins }, ${JSON.stringify(
+            options
+          )}),
+          ${JSON.stringify(massage)}
+        );`,
+        sandbox
+      );
+      return sandbox.output;
```
suchipi: You can also get a return value from vm.runInContext, as long as the script you're running is an expression.

duailibe:
@suchipi it's working now! Can you take another look? =)

tests_config/require_standalone.js
```diff
@@ -0,0 +1,46 @@
...
+  "parser-postcss.js",
+  "parser-graphql.js",
+  "parser-markdown.js",
+  "parser-vue.js"
+].map(filename =>
+  fs.readFileSync(path.join(process.env.PRETTIER_DIR, filename), "utf-8")
+);
+
+const sandbox = vm.createContext();
+vm.runInContext(sources.join(""), sandbox);
```
suchipi: You might want to join them with `;` to avoid ASI issues. I don't think our build output has any problems with that in practice, but better safe than sorry

tests_config/require_standalone.js
```diff
@@ -0,0 +1,46 @@
...
+
+module.exports = {
+  formatWithCursor(input, options) {
```
suchipi: The real standalone build also has `format`, `utils`, etc. We should probably have those here, too.

duailibe: @suchipi but we have no tests for them

PR #4671, opened 2018-06-11
refactor: remove redundant call to normalizeOptions

tony19:
Removes redundant call to `normalizeOptions` (already done in preceding line)

https://github.com/prettier/prettier/blob/8ec543276838f66f0a84824a39c215a3039e9071/src/main/core.js#L260-L263


ikatyang:
Thanks!

Issue #4675, opened 2018-06-12
Suggestion: Publish a Prettier Babel plugin


jaydenseric:
Prettier is ✨awesome ✨for keeping source code tidy, but what about distribution code?

Babel output is messy and it would be nice for package consumers to see pretty code when debugging `node_modules`.

Currently, to publish pretty code you must:

1. Run a babel build script.
2. Run a Prettier script over the files Babel output.

This is a bummer for several reasons:

1. [It adds boilerplate to package scripts](https://github.com/jaydenseric/graphql-api-koa/blob/v0.3.1/package.json#L63).
2. Code is parsed to an AST and manipulated two times, synchronously.
3. If you use the Babel CLI `--watch` mode, you have a conundrum. Either you can simply not produce pretty output in development, or do what I do: [Avoid the Babel CLI `--watch` mode and setup your own watcher](https://github.com/jaydenseric/graphql-api-koa/blob/v0.3.1/package.json#L71).
4. It rules out using Babel source maps.

**What if Prettier provided a Babel plugin**, so that all you need to do is add it to you Babel config? In one parse the output is pretty and you can use the regular babel CLI `--watch` mode.

[I had the idea a while ago](https://twitter.com/jaydenseric/statuses/961121691303858176) and have had a few false starts trying to make a plugin; The necessary internals are not conveniently exposed.

josephfrazier:
> I had the idea a while ago and have had a few false starts trying to make a plugin; The necessary internals are not conveniently exposed.

I think one of the fundamental issues here is that Babel plugins only transform the AST, whereas prettier would need to operate in the code generation stage of Babel. See here for details: https://github.com/jamiebuilds/babel-handbook/blob/c6d34a64e212b1ffa63eca1e79e033c9b8f8bf54/translations/en/plugin-handbook.md#toc-stages-of-babel

It looks like Babel does allow you to specify a custom code generator, so that might be the place to use Prettier. See here: https://github.com/babel/babel/pull/3561. However, note that Prettier does use the original source code to make some of it's formatting decisions, so that would need to be passed by Babel into the code generator as well.

azz:
I've spoken about it before with @hzoo: Prettier as an alternative to `@babel/generator`. It's a cool idea and could be worth pursuing. 

We should definitely do #2068 first, it doesn't make sense to have to run `prettier(babelGenerator(ast))`.

Also not sure whether it should live in the Prettier org or the Babel monorepo?

hzoo:
[...]

PR #4677, opened 2018-06-12
Allow `originalText` to be passed to `formatAST()`

josephfrazier:
This allows `prettier.__debug.formatAST` to not crash when given the AST
for the below code (taken from https://astexplorer.net/), which is
useful when trying to use Prettier as a code generator for an AST
without corresponding source code (https://github.com/prettier/prettier/issues/4675)

```js
let tips = [
  "Click on any AST node with a '+' to expand it",

  "Hovering over a node highlights the \
   corresponding part in the source code",

  "Shift click on an AST node expands the whole substree"
];

function printTips() {
  tips.forEach((tip, i) => console.log(`Tip ${i}:` + tip));
}
```

### DIFF
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -13,9 +13,7 @@ const addAlignmentToDoc = docBuilders.addAlignmentToDoc;
 const docUtils = doc.utils;
 
 function printAstToDoc(ast, options, addAlignmentSize) {
-  options = Object.assign({}, options, {
-    originalText: options.originalText || ""
-  });
+  options = Object.assign({ originalText: "" }, options);
   addAlignmentSize = addAlignmentSize || 0;
 
   const printer = options.printer;
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -13,6 +13,9 @@ const addAlignmentToDoc = docBuilders.addAlignmentToDoc;
 const docUtils = doc.utils;
 
 function printAstToDoc(ast, options, addAlignmentSize) {
+  options = Object.assign({}, options, {
+    originalText: options.originalText || ""
+  });
   addAlignmentSize = addAlignmentSize || 0;
 
   const printer = options.printer;
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -13,9 +13,6 @@ const addAlignmentToDoc = docBuilders.addAlignmentToDoc;
 const docUtils = doc.utils;
 
 function printAstToDoc(ast, options, addAlignmentSize) {
-  options = Object.assign({}, options, {
-    originalText: options.originalText || ""
-  });
   addAlignmentSize = addAlignmentSize || 0;
 
   const printer = options.printer;
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -13,7 +13,9 @@ const addAlignmentToDoc = docBuilders.addAlignmentToDoc;
 const docUtils = doc.utils;
 
 function printAstToDoc(ast, options, addAlignmentSize) {
-  options = Object.assign({ originalText: "" }, options);
+  options = Object.assign({}, options, {
+    originalText: options.originalText || ""
+  });
   addAlignmentSize = addAlignmentSize || 0;
 
   const printer = options.printer;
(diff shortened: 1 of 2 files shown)


src/main/ast-to-doc.js
```diff
@@ -13,6 +13,9 @@ const addAlignmentToDoc = docBuilders.addAlignmentToDoc;
 const docUtils = doc.utils;
 
 function printAstToDoc(ast, options, addAlignmentSize) {
+  options = Object.assign({}, options, {
+    originalText: options.originalText || ""
+  });
```
j-f1: You can make this shorter (& clearer IMO) by doing

```js
options = Object.assign({ originalText: "" }, options)
```

josephfrazier: Ah, of course. Fixed in https://github.com/prettier/prettier/pull/4677/commits/752b71dceb32b8c3ac2482750cc8cc656a583731

azz:
It won't crash, but will it introduce bugs (such as placing comments in the wrong location?)

josephfrazier:
Well, it looks like comments get omitted entirely with this change, but I *think* it shouldn't introduce any non-comment-related bugs (if it does, we likely have bigger problems). On the other hand, `formatAST` seems not to work at all on `master`...

azz:
e.g. this will subtly break:

https://github.com/prettier/prettier/blob/889b7dab792de2afbbd69cced06dd6e5ba7ed1cc/src/language-css/printer-postcss.js#L110-L122

azz:
And `// prettier-ignore` will break (printing `''`)

josephfrazier:
:disappointed: Oh yeah, I guess `// prettier-ignore` is a relatively prominent case. Shame about the CSS bug (and any others like it).

Given that `formatAST` isn't part of the documented API (it's only exposed under `prettier.__debug`, right?), I don't think it's so bad for these to start subtly breaking (versus not working at all).... but I definitely understand if you/others do think so.

azz:
In my mind, failing early is better than creating subtle bugs. It would be great if Prettier worked without `originalText`, but it's used in a lot of places to work around parser bugs, and in valid cases like Vue support.

josephfrazier:
I see, thanks for the explanation. I guess another way to support the use case in https://github.com/prettier/prettier/issues/4675#issuecomment-396653791 is to provide a way for the user to pass `originalText` to `formatAST`, instead of just assuming it's empty. That way, it's a little more obvious to the user that they're responsible for what happens. I'll see if I can make that happen.

azz:
You already can: `prettier.__debug.formatAST(ast, {originalText: '...'})`

josephfrazier:
Hmm, when I try that, it looks like the option doesn't make it through:

```
No parser and no filepath given, using 'babylon' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.
Ignored unknown option `{ "originalText": "" }`.
/home/josephfrazier/workspace/prettier/src/common/util.js:232
  return text.charAt(
              ^

TypeError: Cannot read property 'charAt' of undefined
    at Object.getNextNonSpaceNonCommentCharacter (/home/josephfrazier/workspace/prettier/src/common/util.js:232:15)
    at hasClosureCompilerTypeCastComment (/home/josephfrazier/workspace/prettier/src/language-js/needs-parens.js:15:10)
    at needsParens (/home/josephfrazier/workspace/prettier/src/language-js/needs-parens.js:67:5)
    at Object.genericPrint [as print] (/home/josephfrazier/workspace/prettier/src/language-js/printer-estree.js:148:19)
    at genericPrint (/home/josephfrazier/workspace/prettier/src/main/ast-to-doc.js:108:18)
    at comments.printComments.p (/home/josephfrazier/workspace/prettier/src/main/ast-to-doc.js:37:14)
    at Object.printComments (/home/josephfrazier/workspace/prettier/src/main/comments.js:487:19)
    at printGenerically (/home/josephfrazier/workspace/prettier/src/main/ast-to-doc.js:35:22)
    at FastPath.call (/home/josephfrazier/workspace/prettier/src/common/fast-path.js:68:18)
    at printPathNoParens (/home/josephfrazier/workspace/prettier/src/language-js/printer-estree.js:1508:22)
```

azz:
Huh, I thought that worked 🤔

josephfrazier:
I dug through the option normalization code (it seems to have gotten very complicated...), and came up with a way to make it work: https://github.com/prettier/prettier/pull/4677/commits/7ea97cc109d1023f7b9fee6700b11f69cf1e8ae7

josephfrazier:
Thanks @azz! @j-f1, do you have any further thoughts on this?

PR #4681, opened 2018-06-13
Add require-pragma scss tests

sirreal:
Prettier does not appear to respect pragmas in scss files (https://github.com/Automattic/wp-calypso/pull/24906#issuecomment-396833896)

Add `require-pragma` tests demonstrating this behavior.

Note that the tests will pass as the snapshots are auto-generated. However, the `scss-without-pragma.scss 1` test output is different from the input, demonstrating that require-pragma is not respected.


ikatyang:
Tracked in #4684.

PR #4685, opened 2018-06-13
[WIP] Add --pure option to ignore source formatting

### LINKED ISSUE #2068: Option for ignoring original formatting 100%
Edit by @lydell: Prettier currently keeps multiline object literals multiline even if they fit in a single line, and does special things to blank lines: removes some, collapses consecutive blank lines into one and never adds blank lines. This is because the majority of people seem to want this level of control since we haven’t found a super good heuristic for printing those things. Some people prefer a 100% uniform printing, though, that _completely_ ignores original formatting, even if it sometimes results in “uglier” code. We should discuss this.

Original post:

Is it possible to have a strict option that disambiguates whitespace and always prints out the same so long as the AST is the same? I'm running into some weird issues where lines wrap and then don't unwrap when they are under the line-length limit. This would greatly help with some of my merge conflicts right now :)

P.S. I'm using typescript.

Thanks

azz:
This is an attempt to check the viability of implementing #2068 - that is to create an option that ignores the source formatting. We'll still need to use the source to workaround some parser bugs, and sometimes for correct comment placement, but all other places will have to have heuristics for when to break or not.

The only _major_ issue is when to break object literals, and the solution I implemented is always break unless _all_ properties are shorthand. This could be refined (e.g. perhaps shorthand or simple identifiers/literals).

I want to run this across a few codebases and see if it does anything stupid.

TODO also: bikeshed the name: I can came up with `pure` as reading from the original text is dirty. 😆

Closes #2068

cc. @SimenB as you've often expressed interest in this feature.

---

Here’s how you can install this now to test it out:

```shellsession
$ yarn add --dev azz/prettier#pure
$ # or:
$ npm i -D azz/prettier#pure
```
<sup>Versions of Prettier installed directly from GitHub (like this one) **don’t support Node 4.x/5.x**.</sup>

---


SimenB:
Yeeeeeees! :heart:

lydell:
Really cool! Have you had the time to start thinking about where to add blank lines yet?

azz:
Not really. For each case I basically picked one of _always break_ or _never break_, except for object literals, where I checked for shorthand properties.

Each check to `options.pure` might need additional heuristics.

azz:
Thanks for that, @j-f1!

I should have noted in the original post: `TODO: Break heuristics`. 😆

josephfrazier:
I'm super happy to see progress on this! Thanks for kicking it off, @azz, and for the detailed feedback, @j-f1.

azz:
(I haven't forgotten about this, hoping to get back to it soon)

rattrayalex:
This is interesting... could we see what happens if we run with this option on across the test suite?

josephfrazier:
@rattrayalex, if you change the default option value to `true` [here] and run `yarn test -u`, the diff shows you how the snapshots change. I wonder if we should temporarily do that in this PR, to make it easier to find tests that have bad outputs...

[here]: https://github.com/prettier/prettier/pull/4685/files#diff-e2dba4fd5a5a1685c0aa9a8d434f884aR19

azz:
Thanks so much for picking it up!

tests/json/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -241,7 +256,12 @@ trailingComma: "all"
 1e10,1e-3]}}
 
 =====================================output=====================================
-{ key1: [true, false, null], key2: { key3: [1, 2, "3", 1e10, 1e-3] } }
+{
+  key1: [true, false, null],
+  key2: {
+    key3: [1, 2, "3", 1e10, 1e-3],
+  },
+}
```
josephfrazier: It looks like the object breaking logic is currently being changed even when `--pure` isn't specified...

j-f1:
**Prettier pr-4685**
[Playground link](https://deploy-preview-4685--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBzATnOMCWVUB0AzgDY5hwAUABAAwA01AjNQJQEwQCqADj3OgDCAQyJVWIeiAg9c0IslDD06CAHcACsoQKUwgG4QcAE0kgARumFgA1tgDKPa3lTIY6AK5wpACxgBbEgB1Hxx4IicKex0wnH0wgE9kcCIFKTwxdBgNK1R-YWQAM2ESMSkAKyIADwAhK1sHYX84ABk8OCKSspBKqvsXEjgARQ8IeE7S7xAndEzk82FzBJJoMx50PBggkxgfZAAOBmnVMSCrHmT1uEz9DqlMAEcPHEwc4TyCpGLJqTF-HDcnimRAGw1G4y+XSmMEW22Mu2QACYpO5hDgyPhBBB-PlklBoHcQB4xAAVRa6b5iAC+VKAA)
```sh
--parser babylon
```

**Input:**
```jsx
greeting.slice( 0, 1 ).toUpperCase()
```

**Output:**
```jsx
greeting

  .slice(0, 1)

  .toUpperCase();

```

josephfrazier:
Thanks for keeping up with this, @j-f1! It's still very much a work in progress. I'm hoping https://github.com/prettier/prettier/pull/4685/commits/fed45a3a172fbeca69d877650151cd8707515c23 will fix the blank lines between chained calls.

josephfrazier:
It seems like a great deal (perhaps most?) of the current snapshot diff is due to the `pure` logic not putting blank lines after multiline statements. It seems pretty reasonable to me to put blank lines after multiline statements, so I need to figure out how to make that happen. I'm guessing I'll need to modify `printStatementSequence` again (see https://github.com/prettier/prettier/pull/4685/commits/7ac0dc28ca586826f9a1cea3ffeba3a701ef367d) to check whether each statement is multiline, but I'm not yet sure how.

tests/ternaries/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -466,69 +462,28 @@ aaaaaaaaaaaaaaa
...
-                  a: 0
-                }
-              : {
-                  a: 0
-                }
-          }
-    };
-
+  ? { a: 0 }
+  : { a: { a: 0 } ? { a: 0 } : { y: { a: 0 } ? { a: 0 } : { a: 0 } } };
```
josephfrazier: this isn't great

azz:
Yeah... Gonna call this one dead. Still think it's a neat idea.

fisker:
Want continue? :smile:

PR #4688, opened 2018-06-13
feat(css): support pragma

ikatyang:
Fixes #4684
Closes #4681

PR #4689, opened 2018-06-13
Transpile webpack generated bundles

duailibe:
Context: https://github.com/prettier/prettier/pull/4688#discussion_r195140218


scripts/build/bundler.js
```diff
@@ -142,22 +142,35 @@ function getRollupOutputOptions(bundle) {
...
+      ]
+    },
     output: {
       path: path.resolve(root, "dist"),
       filename: bundle.output,
-      library:
-        bundle.type === "plugin"
-          ? ["prettierPlugins", bundle.name]
-          : bundle.name,
+      library: ["prettierPlugins", bundle.name],
```
suchipi: Why'd this part change?

duailibe: We won't use webpack for "not plugins", so I'm throwing in line 145 if `bundle.type !== "plugin"`

suchipi: Ah, okay

PR #4690, opened 2018-06-13
Travis - Allow failures in Node 6

duailibe:
I tried not running files with "flow" in them, but it doesn't stop segfaults :(

I'm setting `allow_failures` now, but we might as well just stop running tests in node 6 :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 
 node_js:
   - node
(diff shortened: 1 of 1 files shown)


suchipi:
We might want a comment on that part of the yaml that points to the relevant flow parser issue, so if it is closed in the future, we can revert it.

PR #4693, opened 2018-06-14
[build] Cache build result

duailibe:
I over-engineered the build script again. I'm saving the checksums of all the files included in each bundle, and in subsequent builds, I'll skip the build entirely if nothing changes.

In practice, this will be most relevant to skip bundling `parser-*.js` files since we most frequently change stuff in printer (i.e. only rebuild `index.js` and `standalone.js`), or we only make changes to a single parser at a time and can skip rebuilding all the others.

To be safe, we'll bundle everything if `yarn.lock` changes.

It works, but there's no pressure to merge this. To put things in perspective, the build in CircleCi currently takes around 4min.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,18 @@ aliases:
         - node_modules
       key: v1-yarn-deps-{{ checksum "yarn.lock" }}
 
+  - &restore_build_cache
+    restore_cache:
+      keys:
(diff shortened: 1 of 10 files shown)


duailibe:
https://circleci.com/gh/prettier/prettier/1747 took only 14s 😍

PR #4699, opened 2018-06-15
feat(graphql): support pragma

### LINKED ISSUE #4682: Support pragma for graphql
- `parser#hasPragma(text: string): boolean`
- `printer#insertPragma(text: string): string`

ikatyang:
Fixes #4682

### DIFF
--- a/src/language-graphql/parser-graphql.js
+++ b/src/language-graphql/parser-graphql.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const createError = require("../common/parser-create-error");
+const { hasPragma } = require("./pragma");
 
 function parseComments(ast) {
   const comments = [];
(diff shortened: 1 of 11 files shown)


duailibe:
I think we currently only test: formats with a correct pragma and doesn't format without pragma. I think we're missing tests for: doesn't format with an incorrect pragma. No big deal though, we haven't got any issues yet for false positives.

ikatyang:
Everything other than correct pragma is considered an incorrect pragma, including no pragma, so I think it should be fine?

duailibe:
@ikatyang Sorry I wasn't clear.. I meant tests that check an incorrect pragma isn't considered correct (a false positive), but nevermind.. was just thinking out loud

PR #4704, opened 2018-06-16
Report test summaries to CircleCI

### LINKED ISSUE #4510: Set up test summaries
CircleCI lets you set up test summaries to easily see which tests are failing and provide insights into tests. Let’s set that up!

It looks like we need something that outputs JUnit files, such as [`jest-junit`](https://www.npmjs.com/package/jest-junit). We need to configure it to output to `$CIRCLE_TEST_REPORTS/jest/prettier.xml`.

j-f1:
Fixes #4510.


j-f1:
screen shot 2018-06-16 at 18 48 06

scripts/test-dist.js
```diff
@@ -20,7 +20,12 @@ shell.exec("npm init -y", { cwd: tmpDir });
 shell.exec(`npm install "${tarPath}"`, { cwd: tmpDir });
 shell.config.silent = false;
 
-const cmd = `yarn test --color --runInBand --reporters default --reporters jest-junit ${
+const reporters =
+  +process.version.match(/^v\d+\./)[1] >= 6
```
lydell: As far as I can tell this condition is always false. I think you forgot parentheses around `\d+`:

```js
+process.version.match(/^v(\d+)\./)[1] >= 6
```

azz: Or `require('semver').major(process.version) >= 6`

scripts/test-dist.js
```diff
@@ -20,7 +20,12 @@ shell.exec("npm init -y", { cwd: tmpDir });
 shell.exec(`npm install "${tarPath}"`, { cwd: tmpDir });
 shell.config.silent = false;
 
-const cmd = `yarn test --color --runInBand ${
+const reporters =
+  +process.version.match(/^v\d+\./)[1] >= 6
+    ? "--reporters default --reporters jest-junit"
+    : "";
+
+const cmd = `yarn test --color --runInBand ${reporters} ${
```
duailibe: Why not add another conditional?

```js
${process.env.JEST_JUNIT ? "--reporters ..." : ""}
```

And then just set this env in `test_prod_node9`

package.json
```diff
@@ -104,5 +105,8 @@
     "build": "node ./scripts/build/build.js",
     "build-docs": "node ./scripts/build-docs.js",
     "check-deps": "node ./scripts/check-deps.js"
+  },
+  "jest-junit": {
```
duailibe: Is it possible to move this to `jest.config.js`?

j-f1: [nope](https://www.npmjs.com/package/jest-junit#configuration). Is there a way to remove it before publishing?

duailibe: Yeah we already modify it in the build script: 
https://github.com/prettier/prettier/blob/3f91d6427d00bb0e247e80575f8c4034c34ca0fc/scripts/build/build.js#L77

j-f1: After re-reading the docs, it looks like `package.json` and environment variables are the only way to configure the package without explicitly enabling it. Putting the config in `jest.config.js` would cause the package to always be loaded, which would cause issues on older mode versions.

SimenB: @palmerj3 is that correct? If so, we should fix it

j-f1: @SimenB I meant that if I put the config in the `reporters` section, that would enable the reporter, which I only want to do on the one test run.

palmerj3: That's correct. You can have configuration live in jest.config.js if you want and use the reporter configuration. Most folks have jest-junit loaded as a reporter all the time and just gitignore the junit.xml file since it doesn't add very much overhead at all.

But if you REALLY don't want a junit.xml file generated locally then perhaps use the CI=true environment variable and update your jest.config.js to have a conditional for that.

SimenB:
```js
reporters: [
  "default",
  process.env.CI ? ["jest-junit", { ...config }] : null,
].filter(Boolean);
```

or something like that should work, shouldn't it?

palmerj3: Yep, exactly.

scripts/build/build.js
```diff
@@ -80,6 +80,7 @@ async function preparePackage() {
   pkg.engines.node = ">=4";
   delete pkg.dependencies;
   delete pkg.devDependencies;
+  delete pkg["jest-junit"];
```
SimenB: this isn't necessary, is it?

j-f1: #4859

PR #4715, opened 2018-06-18
Bump globby

duailibe:
It uses `fast-glob`, which is supposed to be a lot faster than `node-glob`. There shouldn't be any noticeable improvement, but there's no harm in updating either way.


duailibe:
Yeah both appveyor and travis are timing out so maybe it's not that faster :P

ikatyang:
Probably related to sindresorhus/globby#75?

Issue #4720, opened 2018-06-20
Using with prettier API


shreyas-a:
```
prettier.format('php code', {
   plugins: ['prettier/plugin-php'],
});
```

This throws an error - No parser and no filepath given, using 'babylon' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.

ikatyang:
```js
prettier.format('php code', {
   filepath: "something.php", // or parser: "php"
   plugins: ['@prettier/plugin-php'],
});
```

shreyas-a:
@ikatyang shows `Couldn't resolve parser "php"`

ikatyang:
Can you share some environment info and steps to reproduce? (e.g. did you install the php plugin? global or local? etc.)

shreyas-a:
Yeah, I've installed it locally.
```
"prettier": "^1.13.4",
"@prettier/plugin-php": "github:prettier/plugin-php",
```

ikatyang:
I still cannot reproduce the issue, here's how I tried:

```sh
mkdir prettier-issue-4720
cd prettier-issue-4720
yarn init -y
yarn add prettier@1.13.4 github:prettier/plugin-php
echo '
const prettier = require("prettier");
const example = `
<?php
                    echo "Hello World!";
?>
something
`
console.log(prettier.format(example, {
  filepath: "something.php", // or parser: "php"
  plugins: ["@prettier/plugin-php"]
}));
' > test.js
node test.js
```

```php
<?php echo "Hello World!"; ?>
something
```

shreyas-a:
When I pass `filepath: 'something.php'` is gives - `TypeError: path$8.extname is not a function`.

ikatyang:
It seems you're using the `standalone.js`, which is [supposed to be used in the browser](https://prettier.io/blog/2018/05/27/1.13.0.html#prettier-works-in-the-browser). If you'd like to use it in the browser, you'll have to request this feature in [prettier/plugin-php](https://github.com/prettier/plugin-php).

shreyas-a:
Thanks, as you mentioned I was using `standalone.js`, no errors or warnings as of now but the code is not getting formatted.

duailibe:
@ikatyang I'm not sure what we should do.. the only use case we thought was passing the `parser`, but it wouldn't be hard to shim `path.extname()`

ikatyang:
[...]

PR #4722, opened 2018-06-21
chore(ISSUE_TEMPLATE): add template for integration issue

ikatyang:
Ref: https://blog.github.com/2018-05-02-issue-template-improvements/

image

So that people would know what kind of info we need to know to understand issues like #4720.


j-f1:
How does this affect the “Report issue” button in the playground?

ikatyang:
It shouldn't affect the report button since we fill in all the content in the link:

https://github.com/prettier/prettier/blob/dcf44ffbdc2f403de02f12516b0c6d5d5813b16f/website/playground/Playground.js#L252-L256

PR #4727, opened 2018-06-21
Upgrade flow-parser to 0.75.0

rubennorte:
The main change in the parser is the support for optional catch bindings (which were a syntax error previously).

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "esutils": "2.0.2",
     "find-parent-dir": "0.3.0",
     "find-project-root": "1.1.1",
-    "flow-parser": "0.74.0",
+    "flow-parser": "0.75.0",
     "get-stream": "3.0.0",
     "globby": "6.1.0",
     "graphql": "0.13.2",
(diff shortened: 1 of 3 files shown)


existentialism:
@rubennorte let's enable flow in the optional catch binding tests (https://github.com/prettier/prettier/blob/master/tests/optional_catch_binding/jsfmt.spec.js)

vjeux:
@suchipi: would there be a way to release a dot release with just this change, so that we can start using it at Facebook? :)

duailibe:
@vjeux we haven't done any releases off a branch (instead of master), but sure we can do that

vjeux:
I've done some in the early days in order to move fast :p

duailibe:
I'll do it later tonight (UTC-3 here) if no one does before :) the release script doesn't support it though

vjeux:
Awesome, thanks a ton!

rubennorte:
@duailibe is there something we can do to help with this release?

duailibe:
@rubennorte Sorry I got distracted over the weekend, will release it right now

duailibe:
@rubennorte @vjeux Just released 1.13.6.. Sorry it took a while!

rubennorte:
@duailibe no problem at all. Thank you very much!

Issue #737, opened 2017-02-19
alternate ternary formatting


mvolkmann:
Here is how Prettier currently reformats nested ternaries:
````
const value = condition1
  ? value1
  : condition2
      ? value2
      : condition3
          ? value3
          : value4;
````
This isn't so bad when there are only two ternaries, but I sometimes have many more because I essentially use them as an alternative to a switch statement when each case only gets a single value.

Here is how I prefer to format this case:
````
const value =
  condition1 ? value1 :
  condition2 ? value2 :
  condition3 ? value3 :
  value4;
````
I like this because it clearly associates each condition with its corresponding value and the indentation doesn't increase regardless of the number of ternaries.

Perhaps some like the current formatting because it discourages use of nested ternaries, but I think my example shows that you can have any number without making the code hard to read.

Would you consider adding an option to support this alternate formatting? Maybe something like this? `--ternary=condval`

**Comment by ggoodman** [comment]
A minor tweak to the current ternary formatting that I favour is like this:

```js
const score = doc.thumbs
    ?   _.reduce(_.range(1, doc.thumbs), (score, n) => {
            return score + baseIncrement / Math.pow(decayFactor, n);
        }, doc.created_at.valueOf())
    :   doc.created_at.valueOf();
```

Specifically, I like to move to the next tab stop after the `?` and `:` portions of the operator. This means that multi-line 'values' are expressed at a consistent indent-level with other blocks.

As you can see above, the logic that is relevant to the situation where `doc.thumbs` is truthy is easy to identify by visual scanning.

As a reference point, here is the output of the same block of code via `prettier`:
```js
const score = doc.thumbs
    ? _.reduce(
          _.range(1, doc.thumbs),
          (score, n) => {
              return score + baseIncrement / Math.pow(decayFactor, n);
          },
          doc.created_at.valueOf()
      )
    : doc.created_at.valueOf();
```

In the latter, you can see that while the reducer function is at a consistent indentation level, it is out of sync with my configured 4-space indentation level (it is indented by 6 spaces). Further indentations within that block propagate this unusual level of indentation to 10 spaces, 14 spaces, etc...

nfroidure:
I agree that nested ternaries do not need indentation since it is clear that they are just acting on the same output value.

The only thing we need is to be able to distinguish conditions from actual values which is clearly visible when prefixing with the `?`/`:` tokens, this is why i find this formatting is the most readable:

```
const value = condition1
  ? value1
  : condition2
  ? value2
[...]

Issue #2729, opened 2017-09-01
JSX ternaries include parens


jasonLaster:
I just upgraded and noticed that JSX ternaries seem to always be including parens. I assume this is to support JSX. I think it would be nice to special case ternaries that have js function calls and avoid the parens.

![screen shot 2017-09-01 at 4 02 00 pm](https://user-images.githubusercontent.com/254562/29985901-08597db0-8f2f-11e7-8c6e-0e922fc394bf.png)

vjeux:
Thanks, we need to fix it. Cc @azz

vjeux:
Another issue: when using the pattern (a ? b ? c : d : e) instead of the usual (a ? b : c ? d : e), then the way it is printing is completely unreadable:

```js
class A {
  render() {
    return (
      <div>
        {properties.length > 1 ||
        (properties.length === 1 && properties[0].apps.size > 1)
          ? draggingApp == null || newPropertyName == null
            ? <MigrationPropertyListItem />
            : <MigrationPropertyListItem apps={Immutable.List()} />
          : null}
      </div>
    );
  }
}
```

outputs

```js
class A {
  render() {
    return (
      <div>
        {properties.length > 1 ||
        (properties.length === 1 &&
          properties[0].apps.size > 1) ? draggingApp == null ||
        newPropertyName == null ? (
          <MigrationPropertyListItem />
        ) : (
          <MigrationPropertyListItem apps={Immutable.List()} />
        ) : null}
      </div>
    );
  }
}
```

vjeux:
cc @suchipi

suchipi:
The formatted presented by @jasonLaster was intentional, but I've been seeing some weird stuff with it too lately, so I think it needs to be changed. I'm not sure what the best way of doing it would be, though.

The current rules for "should we print this ternary in JSX mode" are:
* The ternary is inside of a jsx expression container (the `{}`)
OR
* The ternary contains JSX elements

I'm thinking we should maybe drop the first check, and only care about it if JSX elements are inside the ternary. But the first check was there for this case:
```jsx
// Note: not necessarily how prettier would format it (the names aren't long
// enough to break), but I'm just trying to show the pattern

const Something = ({ answer }) => {
  const contentIfYes = (
    <div>
      {"...lots and lots of content"}
    </div>
  );

  const contentIfNo = (
    <div>
      {"...lots and lots and lots of content"}
    </div>
  );

  return (
    <div>
      {answer ? (
        contentIfYes
      ) : (
[...]

PR #4731, opened 2018-06-21
Ternary Operators in JSX Expression Containers

tyrerexus:
This pull request aims to fix the issue of excessive nesting brought up in the following comment: https://github.com/prettier/prettier/issues/737#issuecomment-392541493

This "fix" is very simple. It adds a new possible way to activate JSX mode for ternary operators. If the first non-conditional parent of the expression is a JSX expression container, then JSX mode gets activated.

This branch also contains a test case for said scenario.


suchipi:
Actually, I think we did this before but it surprised some people...

suchipi:
Sorry, reverted; I shouldn't have merged.

suchipi:
I'll try to find the issue that made us remove this before...

suchipi:
Yeah, see https://github.com/prettier/prettier/issues/2729. Sorry for the confusion 🙏

PR #4738, opened 2018-06-23
Upgrade Docusaurus to 1.3.1

yangshun:
## Summary

Docusaurus 1.3.1 brings about some [backwards-incompatible CSS changes](https://github.com/facebook/Docusaurus/blob/master/CHANGELOG.md#breaking-changes). So the team is helping the major users update 😄 

The header font weights in Docusuaurs have been changed. I left them alone in this PR but will be glad to help Prettier change it such that it looks better (according to my limited artistic skills). Let me know if you want that!

cc @JoelMarcey @endiliey

## Test plan

Ran the site locally and visited every page. The playground doesn't seem to be working locally though.

screen shot 2018-06-23 at 12 19 24 am

screen shot 2018-06-23 at 12 19 32 am

### DIFF
--- a/website/data/languages.yml
+++ b/website/data/languages.yml
@@ -38,7 +38,7 @@
   showName: true
   image: /images/languages/tools_wip.svg
   variants:
-  - "[`Elm`](https://github.com/gicentre/prettier-plugin-elm)` (via `[`elm-format`](https://github.com/avh4/elm-format))"
+  - "[`Elm`](https://github.com/gicentre/prettier-plugin-elm) (via [`elm-format`](https://github.com/avh4/elm-format))"
   - "[`Java`](https://github.com/thorbenvh8/prettier-java)"
   - "[`PHP`](https://github.com/prettier/prettier-php)"
   - "[`PostgreSQL`](https://github.com/benjie/prettier-plugin-pg-sql/)"
--- a/website/data/languages.yml
+++ b/website/data/languages.yml
@@ -2,46 +2,46 @@
   showName: false
   image: /images/languages/tools_js.svg
   variants:
-  - "[ES2017](https://github.com/tc39/proposals/blob/master/finished-proposals.md)"
-  - "[JSX](https://facebook.github.io/jsx/)"
-  - "[Flow](https://flow.org/)"
-  - "[TypeScript](https://www.typescriptlang.org/)"
-  - "[Vue](https://vuejs.org/)"
-  - "[JSON](http://json.org/)"
+  - "[`ES2017`](https://github.com/tc39/proposals/blob/master/finished-proposals.md)"
+  - "[`JSX`](https://facebook.github.io/jsx/)"
+  - "[`Flow`](https://flow.org/)"
+  - "[`TypeScript`](https://www.typescriptlang.org/)"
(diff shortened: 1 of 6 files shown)


duailibe:
Thanks for doing this!!

FYI you can check the changes in the Netlify preview: https://deploy-preview-4738--prettier.netlify.com/

I see the a lot changed in the main page (I tested on Chrome with macOS) but it looks better. Only thing I'd call a regression is the red "borders" are missing here:

screen shot 2018-06-24 at 22 37 53

cc @azz who worked on the landing page

ikatyang:
There's a blank block during animation, probably caused by the `padding` in `.mainContainer`.

image

yangshun:
> Only thing I'd call a regression is the red "borders" are missing here:

@duailibe We removed the borders from all code blocks styles, so naturally that border was removed. Maybe we can add a background to the code block to get it to be more obvious?

> There's a blank block during animation, probably caused by the padding in .mainContainer

@ikatyang I noticed that as well but thought it was only on my computer. I'll try to fix it. If you have any ideas how to fix it please let me know!

yangshun:
@duailibe @ikatyang How does the PR look now? Any more changes need to be made?

suchipi:
The red border-left on these code blocks is still missing:

| prettier.io | https://deploy-preview-4738--prettier.netlify.com/ |
|--|--|
|  |  |

ikatyang:
The `text-decoration: underline` for `a:focus` is missing:

| [prettier.io](https://prettier.io/) | [deploy-preview-4738--prettier.netlify.com/](https://deploy-preview-4738--prettier.netlify.com/) |
|-|-|
|image|image|

The alignment and the "border" for editor support is not suitable now, page width 1107px for example:

| [prettier.io](https://prettier.io/) | [deploy-preview-4738--prettier.netlify.com/](https://deploy-preview-4738--prettier.netlify.com/) |
|-|-|
|image|image|

orta:
Heyo 👋  - can confirm this looks and feels good WRT the index page

---

Personally speaking, I feel like the text is pretty big on the API references, could probably do with dropping down a bit. 

https://deploy-preview-4738--prettier.netlify.com/docs/en/api.html#prettierresolveconfigfilepath-options

screen shot 2018-06-30 at 5 38 10 pm

*However* I feel like that font size is 👍 for normal titles, [like here](https://deploy-preview-4738--prettier.netlify.com/docs/en/precommit.html)

I think this is less of a "font size" issue than a white spacing issue, as these titles could do with a larger `margin-top` ( if you try removing the `margin-top: 14px`from `.mainContainer .wrapper .post h2, .mainContainer .wrapper .post h3`) they start to feel right IMO

yangshun:
> Personally speaking, I feel like the text is pretty big on the API references, could probably do with dropping down a bit.

This is part of Prettier's CSS overrides which I'm hesitant to change. It would be faster to merge this PR, get the designer of the site looks through the entire site and tweak the overrides for any styling issues. At the very least, this version is functional.

orta:
Cool, that's 👍 with me

duailibe:
Thank you @yangshun for doing this, and @orta for taking a look ❤️

PR #4743, opened 2018-06-24
It should be said that filewatcher is a plugin

ehmkah:
File-Watcher is an extra plugin which needs to be installed.


suchipi:
Thanks!

PR #4753, opened 2018-06-26
[WIP] feat: html support

### LINKED ISSUE #1882: Supporting HTML/Mustache

evilebottnawi:
Partial #1882

Feedback welcome.

How to use:
```shell
yarn add --dev prettier/prettier#feat-html-support
```
```shell
npm i -D prettier/prettier#feat-html-support
```

NOTE: you need to pass `--parser parse5` to enable it.

Initial style preference questions:
1. Indent `head` and `body` tags inside `html` or not
2. Indent inside `script` and `style` tags or not
3. Insert newline after `script` and `style` tags or not


tests/html_basics/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,202 @@
...
+  <div class="form-group">
+    <label for="exampleInputEmail1">Email address</label>
+    <input type="email"
+      class="form-control"
+      id="exampleInputEmail1"
+      aria-describedby="emailHelp"
+      placeholder="Enter email">
+
+    <small id="emailHelp"
+      class="form-text text-muted">We'll never share your email with anyone else.
```
j-f1: This shouldn’t be broken onto multiple lines unless it goes over the print width.

evilebottnawi: @j-f1 I know that there are a lot of problems, so I will assign a review from the main developers when I finish this, now it's just for feedback

lydell:
What is the plan for handling whitespace? (I guess the code tells, but I'm too lazy.) In theory we can't change anything, because `white-space: pre;` and `element.textContent`. But that would be sad.

evilebottnawi:
@lydell looks like we should output this as is, it is impossible to solve, also i think we should release html asap, it is allow to get feedback from people and maybe we have new ideas how to solve this

gabrielmaldi:
Simple instructions on how to try this? I'd happily provide feedback after formatting my codebase.

j-f1:
@gabrielmaldi `yarn add --dev prettier/prettier#feat-html-support` or `npm i -D prettier/prettier#feat-html-support`.

gabrielmaldi:
@j-f1 thanks! I'm constantly getting ```[error] Invalid ``--write`` value. Expected a boolean, but received `[null,true]`.``` and I can't seem to get around it. This doesn't happen with prettier 1.13.6 (without this PR); could it be that something's wrong with `--write` or am I just running it wrong?

j-f1:
@evilebottnawi Would it be possible for you to avoid squashing your commits? It makes it harder to see what’s been changed since earlier and review those changes. Also, GitHub automatically squashed commits together when we merge the PR, so it doesn’t help keep the commit history any cleaner than it already is.

evilebottnawi:
@j-f1 :+1:, just squash PRs whats is look very bad (all first PRs look very bad :smile: ), no squash after last PR

evilebottnawi:
@j-f1 Can we rename `html_glimmer` and `html_handlebars` to `handlebars_glimmer` and `handlebars`, because it is not html also for easy testing using `node_modules/.bin/jest tests/html_*/ -u`?

lipis:
👍 for indenting `head` and `body` tag inside the `html` and 👎 in this post.

vjeux:
@j-f1 feel free to rename it to just glimmer, it’s no longer exactly compatible with handlebars

vjeux:
For whitespace, we should figure out a way to opt-out for specific places. The trick I used to do was:

```html
<ul><!—
  —><li>First</li><!—
  —><li>Second</li><!—
—></ul>
```

It’s not pretty but it let me indent things without introducing whitespace when needed.

The second thing we need to have a documentation page about it and put a lot of warnings saying that unlike js, pretty printing a whole html folder is not safe because of this issue.

vjeux:
One tool that would be useful is to print all the places where prettier would add/remove whitespace in an easily digestible format so that you can manually inspect them when doing the big codemod.

nunocastromartins:
@vjeux I don't know if it matters, but your trick introduces new comment nodes on the DOM tree, which may not be desired if you're performing DOM manipulations via JavaScript. An alternative, which I usually use, is:
```
<ul
  ><li>First</li
  ><li>Second</li
></ul>
```
Not the prettiest thing, but it shouldn't have any side-effect, afaik.

evilebottnawi:
@vjeux I agree about note in some case prettify is no safe, especially when you use `white-space: pre`. My solution:
1. By default do not format children nodes inside `pre` tag.
2. Add note about this.
3. Use `prettier-ignore` as workaround when you want ignore formatting inside tag.

Any objections?

evilebottnawi:
Vote +1 for indenting inside `script` and 'style` tag and -1 for not.

evilebottnawi:
Vote :+1: for insert newline after script and style tags and -1 for not or :heart: print as in source
```html
<script>
alert("test");
</script>
```
vs
```html
<script>alert("test");</script>
```

evilebottnawi:
Newline in empty tags
:+1: for first :-1: for second

```html
<article
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars"
></article>
```
vs
```html
<article
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars"
>
</article>
```

vjeux:
Let's avoid this voting game. Can you make decisions based on what you believe should be done and then we can revisit based on feedback we get and running in existing codebases.

evilebottnawi:
@vjeux ok

vjeux:
I forgot about prettier-ignore. This is indeed a good solution for whitespace in those cases.

evilebottnawi:
@vjeux should we respect xhtml?

vjeux:
What do you mean exactly by xhtml?

evilebottnawi:
@vjeux https://en.wikipedia.org/wiki/XHTML

vjeux:
I mean, which part of xhtml do you have in mind?

evilebottnawi:
@vjeux if doctype is `xhtml` use `<br />`, if not use `<br>`

lipis:
To answer the first question: [https://twitter.com/PrettierCode/status/1011743248145178625](Twitter poll). Let's indent them all.

vjeux:
Does `<br />` also work in normal html? If yes, let’s always output it?

evilebottnawi:
@vjeux `<br />` works for `html` and `xhtml`, and looks consistently with how we print `jsx`

evilebottnawi:
Blocker https://github.com/prettier/prettier/pull/4753

j-f1:
@evilebottnawi That’s the same PR as this one.

evilebottnawi:
@j-f1 fixed

lydell:
If `<br />` has broader compatibility, then why not?

ikatyang:
@gabrielmaldi

> - I need to pass `--parser parse5`: `./node_modules/.bin/prettier "file.html" --write --parser parse5`. Otherwise I get `[error] No parser could be inferred for file: file.html`. Shouldn't `.html` files be inferred as HTML by default?

This is intended, any unreleased language is not enabled by default, I've updated the original post to mention it.

https://github.com/prettier/prettier/blob/42cddd7b42010afcb5e52a76b87adaf1116f64ea/src/language-html/index.js#L6-L12

evilebottnawi:
https://github.com/inikulin/parse5/issues/261#issuecomment-401132306 
Aby ideas? Get from source code?

evilebottnawi:
@gabrielmaldi can you great gist, with small examples (actual and expected)?

michaelauderer:
What's left to do here?

evilebottnawi:
@michaelauderer some works :smile:

solshark:
Just tried this version myself for .vue files - works great so far (template section). Found one small issue, it always adds extra empty line before </template>
image

listepo-alterpost:
Does it already work for Angular?(inline templates too)

evilebottnawi:
@listepo-alterpost should, but in some cases can be ugly, feel free to send example where ugly

evilebottnawi:
@solshark known issue, will be fixed

StarpTech:
Some points from my side:

* Parse5 parse `<template>` nodes as subdocuments. The nodes are exposed under `node.content` with a new `root` element and `childNodes`.
* If you format embedded content like javascript please be aware of this strange edge case https://github.com/inikulin/parse5/issues/262. Prettier removes unneeded escape characters.
* If you need a fork of parse5 which is able to parse case-sensitive, custom self-closing tags have a look here https://github.com/StarpTech/prettyhtml/tree/master/packages/prettyhtml-parse we could also fix your issue https://github.com/inikulin/parse5/issues/261#issuecomment-401132306

StarpTech:
@solshark vue template won't work because custom self-closing elements are heavily used. Therefore camelCase tag names are allowed too.

@listepo-alterpost this is not possible because angular use case-sensitive attributes like `[(onChange)]="dede"` which are converted to lowercase at the parsing step.

@evilebottnawi correct?

src/language-html/parser-parse5.js
```diff
@@ -3,12 +3,15 @@
 function parse(text /*, parsers, opts*/) {
   // Inline the require to avoid loading all the JS if we don't use it
   const parse5 = require("parse5");
+  const htmlparser2TreeAdapter = require("parse5-htmlparser2-tree-adapter");
+
   try {
     const isFragment = !/^\s*<(!doctype|html|head|body|!--)/i.test(text);
```
StarpTech: This will fail `"<!-- comment --> <p></p>"` I recommend to use 

```js
var isFragment = !/<(!doctype|html|head|body)/i.test(doc)
```

src/language-html/embed.js
```diff
@@ -3,53 +3,89 @@
...
+          const langAttr = parentNode.attribs["lang"];
+
+          if (langAttr === "postcss") {
+            parser = "css";
+          } else if (langAttr === "scss") {
+            parser = "scss";
+          } else if (langAttr === "less") {
+            parser = "less";
+          } else {
+            parser = "css";
```
StarpTech: Use it as the default value for `let parser` and overwrite when needed.

src/language-html/embed.js
```diff
@@ -3,53 +3,89 @@
...
-        );
-        return concat([hardline, doc]);
+      if (parentNode.type === "script") {
+        if (Object.keys(parentNode.attribs).length === 0) {
+          parser = "babylon";
+        }
+
+        const typeAttr = parentNode.attribs["type"];
+
+        if (
```
StarpTech: Don't forget to add `parser:json` e.g for common `<script type="application/ld+json">` tags

src/language-html/embed.js
evilebottnawi: @StarpTech it is part of multiparser (there we should based on standard parsers), idea how we can solve this?

StarpTech: Could you rephrase? I mean to initial `parser` with `"css"` and update it when the type is different.

evilebottnawi:
@StarpTech you fork looks good, maybe we should use this instead `parse5`, can't find `CHANGELOG`, also I'm afraid that you can abandon him.

maxime1992:
With Angular it's updating from:

```
<ng-template #showFileUpload>
  <ng-container *ngIf="...">
```

To 

```
<ng-template #showfileupload>
  <ng-container *ngif="...">
```

Otherwise that's a good start :ok_hand: !

StarpTech:
> can't find CHANGELOG

@evilebottnawi As a standalone user, I didn't need to maintain it. The changelog is one front of the readme but we can maintain one when we do more changes.

> also I'm afraid that you can abandon him.
I don't abandon without any reasons. The fork will still exist and I'm willing to maintain it.

vjeux:
I wouldn't worry about vue and angular for now, just make it work well for pure html and we can look at the other ones separately.

StarpTech:
@evilebottnawi I added CI and updated the changelog. Feel free to start any tests.

maxime1992:
Well as far as I can see it seems to work ok with Angular (at least when the template is into an HTML file of course). The only annoying thing is the fact that attributes are lower cased which IMO shouldn't happen but I might be wrong.

karlhorky:
@evilebottnawi I think you meant @maxime1992 and I think @maxime1992 meant his example here:

https://github.com/prettier/prettier/pull/4753#issuecomment-403854338

ianks:
Would it be reasonable to support templated html? For example, we use `erb` templates in ruby, and it would be awesome to use prettier to autoformat them. Currently, prettier does a good job of formatting the HTML portion of the document, but struggles with recognizing the embedded template tags (i.e `<% ... %>`).

In my eyes, it would be unrealistic to attempt to prettify the embedded ruby. However, if prettier could recognize these embed tags and simply ignore them (do not tranform), I think that could work as a nice middle ground.

### Example

When given this template:

```erb
<!doctype html>
<html lang="en">
<body>
  <% if local(:message).nil? %>
    <span>Hello world</span>
  <% else %>
    <span><%= message %></span>
  <% end %>
</head>
</html>
```

Prettier outputs:

```erb
<!DOCTYPE html>
<html lang="en">
  <head></head>
  <body>
    <% if local(:message).nil? %><span>Hello world</span>
    <% else %><span><%= message %></span>
    <% end %>
  </body>
</html>
```

It would be nice if prettier could output something like:

```erb
<!DOCTYPE html>
<html lang="en">
  <head></head>
  <body>
    <% if local(:message).nil? %>
      <span>Hello world</span>
    <% else %>
      <span><%= message %></span>
    <% end %>
  </body>
</html>
```

ChristianMurphy:
@ianks erb support may fit well with https://github.com/iamsolankiamit/prettier-ruby, which would potentially have the ability to format the embedded ruby.

StarpTech:
@evilebottnawi it would be good to define the scope of the current implementation. Does it support HTML5 only? Does it consider specific templates like erb, razor, handlebars, angular vue?

@ianks this means that prettier has to be aware of a specific template language but currently the goal is to support HTML5 only.

michaelauderer:
I suggest that HTML5 support is rolled out ASAP, and template implementations can be added later.

evilebottnawi:
Sorry, don't have time to do this for `1.14`, will be done for `1.15`

omardelarosa:
@evilebottnawi Will the lower-casing issue for angular that @maxime1992 raised be present in the `1.14` version?  If not, a flag to ignore attribute casing would be useful at least.

evilebottnawi:
@omardelarosa we solve this without new options or flags

omardelarosa:
@evilebottnawi Great, as long as it's resolved in some form.   It needn't be with flags/options.

StarpTech:
@evilebottnawi during the work on my project I found a bad circumstance. When `parse5` parse a `<template>` tag or even nested `template` tags it creates a new `document-fragment` similar to a root element. The children are exposed under the `content` property not `children`. In order to avoid changing the algorithm and to preserve parent informations I did following:

```js
// Before we pass the doc to parse5. Replace template tags with custom tag e.g prettier-template
      doc = doc
        .replace(/<template(.*)>/g, '<' + 'prettier-template'+ '$1>')
        .replace(/<\/template\s*>/g, '</' + 'prettier-template' + '>')

// during printing we replace the name again
  var name = node.tagName === 'prettier-template' ? 'template' : node.tagName
```

evilebottnawi:
@StarpTech can you provide example actual and expected

StarpTech:
Nevermind LGTM I think it's due to the different treeAdapter.

Input
```html
<template><v-select>
<template slot="selection" scope="data">
<v-chip>{{getChipString(data.item)}}</v-chip> <span>    {{data.item.name}}</span>
</template><template slot="item" scope="data">
<v-chip>{{getChipString(data.item)}}</v-chip> {{data.item.name}}
</template>
</v-select>
</template>
```
output
```html
<template>
  <v-select>
    <template slot="selection" scope="data">
      <v-chip>{{getChipString(data.item)}}</v-chip>
      <span>{{data.item.name}}</span>

    </template>
    <template slot="item" scope="data">
      <v-chip>{{getChipString(data.item)}}</v-chip>
      {{data.item.name}}
    </template>
  </v-select>

</template>
```

paulreimer:
FYI `camelCase=` attribute names are encouraged in lit-html / LitElement, so that is another case where the lowercasing causes issues.

mollstam:
As a temp fix to the lowercasing issue I commented out the following inside parse5 tokenizer/index.js:1206 (in the `Attribute name state` block). 

```javascript
else if (isAsciiUpper(cp)) {
    this.currentAttr.name += toAsciiLowerChar(cp);
}
```

Can't wait for the real solution. :)

StarpTech:
@mollstam a fork of the parse5 with all the needed features is already there https://github.com/StarpTech/parse5

mollstam:
Ooh, sweet stuff! Cheers :tada:

jraoult:
@evilebottnawi many thanks for taking care of this. I don't know if it is in the scope of this PR, but I just tried it on an HTML page with some YAML frontmatter data (used by Jekyll in my case).

It removes the line breaks so that:
```
---
layout: layout-main
---
<style>
</style>
```

becomes 

```
--- layout: layout-main ---<style>
</style>
```

It should probably not try to format it.

StarpTech:
@jraoult this is definitely not covered here. Your example is interpreted as text nodes. In the current implementation, we don't respect technologies except html5.

heavybeard:
I suggest to create new different plugins for these specific purposes: angular, yaml, etc...

Include particular behaviour in an HTML5 parses it's out of scope, IMHO

StarpTech:
@heavybeard this could be one solution but I see no problem to extend the HTML5 parser in order to support common template syntaxes as long it's still archivable with minor tweaks. Nobody needs a formatter for just HTML these days.

We could instantly support a broad range of frameworks like Vue, Angular, Svelte, Ember... without to handle everything separately.

heavybeard:
If you think there are only minor tweaks for me it's ok obviously. I'm waiting for this feature like I'm waiting for the next World Cup in the Italian football history :D

j-f1:
Since we already support YAML front matter in Markdown and CSS, it seems reasonable for us to support it in HTML. The other parsers remove the front matter and print it separately, then pass the rest of the file to the actual parser.

pauldraper:
Issues:

(1) HTML entities are mishandled. ```&amp;amp;``` becomes ```&amp;```.

(2) The case of attributes is transformed. That *should* have been fine, except Angular uses case-sensitive attributes. Angular is doing the The Wrong Thing but seeing at it is one of the largest frameworks that uses HTML templates, case nomalization should probably be optional.

StarpTech:
@pauldraper this is due to the parse5 parser. 
This will fail as well but it's correct behaviour according to the specs.

```
<my-component [icon]="&#333"></my-component>
```
output
```
<my-component [icon]="ō"></my-component>
```

pauldraper:
> A formatter shouldn't change content in attributes.

I think it should. For example,

```html
<my-component [icon]='"'></my-component>
```

```html
<my-component [icon]="&quot;"></my-component>
```

In your example, I would suggest that a canoncial format to write "small o with macron" to be well within the responsibilities of a formatter. The problem is that the latter way requires some assumptions about character encoding (Unicode vs ASCII). And unfortunately, AFAIK prettier lacks an option for that.

StarpTech:
@pauldraper OK I got your point but I have a different view on it. I don't care about raw html. I see the need for a formatter for ui frameworks like vue, angular, svelte where content should stay as it is.

StarpTech:
@evilebottnawi Do you still working on this?

evilebottnawi:
@StarpTech sorry, right now now, don't have enough time right now, we can merge this without official support, because a lot of work already done

KrickRay:
@evilebottnawi Can you change this to a plugin?

evilebottnawi:
@KrickRay why?

KrickRay:
@evilebottnawi would allow us to start using it

evilebottnawi:
@KrickRay i rebase this PR and remove publish version, it is allow to continue work

evilebottnawi:
@ikatyang Can we merge this PR as is. I remove `since` for disable `html` by default. Don't have time on this right now, maybe other developers want to continue work?

evilebottnawi:
@ikatyang i am glad to fix your comments, but i really don't have time :disappointed: I think other developers can solve this and add more tests

ikatyang:
I think it should be fine to merge, @j-f1 any thoughts?

j-f1:
If we’re not putting it in the production version, let’s merge!

ikatyang:
@evilebottnawi Is there a list for what haven't been done here? It'd be helpful for future work.

danburzo:
I've tried using this to prettify some Ractive templates, but it seems it converts all elements to lowercase? e.g. `<ProfileCard>` becomes `<profilecard>` which does not have the same meaning.

src/language-html/printer-htmlparser2.js
```diff
@@ -95,25 +202,152 @@ function genericPrint(path, options, print) {
   }
 }
 
-function printAttributes(path, print) {
-  const node = path.getValue();
+function isBooleanAttributeNode(node) {
+  return (
+    node.type === "attribute" &&
+    [
+      "allowfullscreen",
```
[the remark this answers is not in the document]
thorn0: They're marked as "Boolean attribute" here:
https://html.spec.whatwg.org/multipage/indices.html#attributes-3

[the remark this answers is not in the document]
ikatyang: Thanks! Added in a5cecbdfee56db2409bd6829b292ed61748377c2.

ikatyang:
Opened #5098 to track HTML issues that should be done before the public release, please comment there if you find something unexpected.

PR #4756, opened 2018-06-27
Update Jest URL to jestjs.io

RhysBower:
Jest got a new domain! Let's point to it.

See facebook/jest#6549.


duailibe:
Nice, thanks!

Issue #4751, opened 2018-06-26
Functional composition heuristics false positives


princed:
**Prettier 1.13.6**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMAWBLAzgOkgWwAcIs4AKAEwEYAaAAgoCYBKAbhBpAkJg2i2SgAhgCcREAO4AFUQgEohANwgYKHEACMRQsAGs4MAMqEdGKAHNkMEQFc4nM6REwp28-iHIAZkIA2pTgArLAAPACFtPQNDIXw4ABkzOG8-AJATESdkTSENAE9faHVCETMYAHVVdGQADgAGThKSOHLtQmySuCdFZM4ROABHGwx+1yF3TyQff3sQUnwMK1tZrDNzXzgARRsIeBSZzhhcyopqpEZD7QxfNYBhCHwPbKhoXpAbUgAVXPlp0gBff5AA)

**Input:**
```js
this.compose(d1, d2);
```

**Output:**
```js
this.compose(
  d1,
  d2
);
```

**Expected behavior:**
Input code is not reformatted. 

As discussed in https://github.com/prettier/prettier/pull/4431#issuecomment-393747847 functional composition heuristics produce some unexpected false positives,  in part because `compose`, `connect`, `flow` terms are too broad and might relate not only to functional context.

Possible solutions:
* Make sure identifiers are originated from certain packages (probably complex and slow as it requires scope)
* Use code shape instead of identifiers list for heuristics 
* At least blacklist obvious edge case, i.e. when `callee` is `MemberExpression` and not `Identifier`

cc @suchipi

suchipi:
Thanks. Not sure how to refine this yet; open to ideas.

I don't think bullet 1 is feasible- we don't have the infrastructure in place to check that right now, and even if we implemented it, it would be very slow.

Bullet 2 was discussed in the original issue and seems like our best bet, but I want to be careful not to introduce regressions.

The MemberExpression thing is intentional, to support eg `import * as R from "ramda"`. But we could filter out `this`, etc.

j-f1:
@suchipi How about checking if the `object` of the MemberExpression is a specific identifier as part of the heuristic?

princed:
Raised https://github.com/prettier/prettier/pull/4758 to filter out at least some straightforward cases.

PR #4758, opened 2018-06-27
Improve functional composition heuristics

### LINKED ISSUE #4858: Codes format unexpected with function "pipe"
image

princed:
As discussed a bit in https://github.com/prettier/prettier/issues/4751 in this PR I'm adding a couple of improvements to functional composition heuristics:

* ~~Do not react when `MemberExpression` object type is `ThisExpression` or `Super`~~
Moved to https://github.com/prettier/prettier/pull/4836

* Do not react when all arguments are `Identifier`s or literals **and** parent node is **not** `CallExpression` with callee equal current node.
Filters out cases like 
```js
composes(a, b, c);
someObj.someMethod(this.field.compose(a, b));
```
 but allows
```js
const ArtistInput = connect(
  mapStateToProps,
  mapDispatchToProps,
  mergeProps
)(Component);
```

fixes #4858

cc @suchipi


src/language-js/printer-estree.js
```diff
@@ -3442,7 +3442,10 @@ function isFunctionCompositionFunction(node) {
   switch (node.type) {
     case "OptionalMemberExpression":
     case "MemberExpression": {
-      return isFunctionCompositionFunction(node.property);
+      return node.object.type !== "ThisExpression" &&
+        node.object.type !== "Super"
```
duailibe: This won't match `this.a.b.c.compose()`

src/language-js/printer-estree.js
```diff
@@ -3511,7 +3515,12 @@ function printArgumentsList(path, options, print) {
   //    )
   // here, but not
   //    process.stdout.pipe(socket)
-  if (isFunctionCompositionFunction(node.callee) && args.length > 1) {
+  if (
+    isFunctionCompositionFunction(node.callee) &&
+    args.length > 1 &&
+    (args.some(a => a.type !== "Identifier") ||
```
duailibe: I'm not completely sure about this. This is a valid pattern:

```js
const someFn = pipe(
  firstStep,
  secondStep,
  thirdStep
);
```

You might want to call the result of the composition several times so it's a valid reason for it not to be the callee of a call expression.

src/language-js/printer-estree.js
princed: Fixed, thank you!

princed:
Hi @duailibe @suchipi, thanks for being so responsive here!

### 1.

I agree that
```js
const someFn = pipe(
  firstStep,
  secondStep,
  thirdStep
);
```
is a valid pattern in a functional context, but it arguably doesn't improve readability much over

```js
const someFn = pipe(firstStep, secondStep, thirdStep);
```

I think leaving is as-is, provided it can fit on one line, is actually more in-line with the Prettier ethos and people's mental model of how Prettier works. Splitting it into multiple lines in the above case is [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

This was also mentioned in https://github.com/prettier/prettier/issues/4172. My aim here is to help improve the 'is this functional' heuristic by looking at the shape of the code.

### 2.

> You might want to call the result of the composition several times so it's a valid reason for it not to be the callee of a call expression.

Correct, but that condition is used to prefer

```js
const ArtistInput = connect(
  mapStateToProps,
  mapDispatchToProps,
  mergeProps
)(Component);
```

over


```js
const ArtistInput = connect(mapStateToProps, mapDispatchToProps, mergeProps)(
  Component
);
```

In the case of an intermediate variable, it would be just

```js
const connectComponent = connect(mapStateToProps, mapDispatchToProps, mergeProps);
const ArtistInput = connectComponent(Component);
```

Which leads us to the previous case.

I'm happy to add more specific cases like Redux's `connect` if that would be useful?

### 3. 

I've also just added literals to that condition, so things like:

```js
compose(1, 2, 3);
```

won't become

```js
compose(
  1,
  2,
  3
);
```
Please take a look and let me know what you think :)

suchipi:
@princed despite keeping it inline being more consistent with prettier's philosophy, users of RxJS in particular voiced that they want `pipe` to keep everything on multiple lines, so that's behavior I want to keep for identifiers. However, I'm fine with changing it for literals (not object literals though).

suchipi:
Maybe we should *not* do the line breaks with identifiers for `connect`, but do it for `pipe` and etc?

marcfallows:
> users of RxJS in particular voiced that they want `pipe` to keep everything on multiple lines

Couldn't those composition function names be configuration based, so we aren't making any assumptions about which method names a particular codebase considers functional?

I know prettier is meant to be light on configuration, but catering the formatting to users of a particular library seems a bit strange to me as well.

alextreppass:
+1 for this being surprising behaviour — it doesn’t seem consistent with how Prettier has worked until this point: which is to keep things on one line if it can fit into the print width.

> Maybe we should not do the line breaks with identifiers for connect, but do it for pipe and etc?

I think that’s still open to false-positives. What if you happened to have a method called ‘pipe’ in one of your classes?

I’d love to see a compromise here, something like:

- the identifier behaviour you mentioned above, or 

- just keeping it on one line if it fits

- a break method names’ list option like @marcfallows describes above, so that people can opt-in to this behaviour and it’s not a hard-coded set of method names for a mix of libraries that not everyone uses. 

This last one would also shift the burden of listing the names to users, and Prettier won’t need to be updated each time RxJs, Ramda, etc add or rename methods.

Thanks!

karlhorky:
I suppose I interpreted the 176 👍s and long discussion on #4172 as being less about a particular library such as RxJs or Ramda and more about the programming concept of functional composition, which very often is paired with function names of `compose`, `pipe`, etc.

Maybe it would be useful to bring some of the original commenters from #4172 in here for further input...

suchipi:
A lot of the :+1:s on #4172 appeared when Ben Lesh tweeted something to the effect of ":+1: here to support better formatting for RxJS in prettier" and linked to the issue.

marcfallows:
@suchipi I would love to know which part of [my comment](https://github.com/prettier/prettier/pull/4758#issuecomment-401084182) made you give the 👎, and why. Could you elaborate?

suchipi:
@marcfallows it wouldn't be appropriate to add a configuration option for something so small, especially when we've turned down configuration options for changes with much more impact. See also https://prettier.io/docs/en/option-philosophy.html

princed:
@duailibe  @suchipi 
`this` filtering is split out to https://github.com/prettier/prettier/pull/4836, PTAL there

alextreppass:
@suchipi for the other cases, would it be possible to apply a strategy of:
- leave it on one line, if the statement fits in the print width
- if it doesn't, _then_ apply the argument breaking heuristics

I think that would be less surprising and more in line with how prettier used to work before https://github.com/prettier/prettier/pull/4431

karlhorky:
@alextreppass in the first post of the original PR (https://github.com/prettier/prettier/issues/4172#issue-306223781) @yannickglt made the point that the print width is not enough for functional composition. Consider the following:

**Input**

```js
import { range } from 'rxjs/observable/range';
import { map, filter, scan } from 'rxjs/operators';

const source$ = range(0, 10);

source$.pipe(
  filter(x => x % 2 === 0),
  map(x => x + x),
  scan((acc, x) => acc + x, 0)
)
.subscribe(x => console.log(x))
```

**Output**

```js
import { range } from "rxjs/observable/range";
import { map, filter, scan } from "rxjs/operators";

const source$ = range(0, 10);

source$
  .pipe(filter(x => x % 2 === 0), map(x => x + x), scan((acc, x) => acc + x, 0))
  .subscribe(x => console.log(x));
```

The output line with `.pipe` here is difficult to visually parse and figure out what is going on.

Some other points were made in that issue too, probably more eloquently and comprehensively than I can.

alextreppass:
@karlhorky thanks for the context. I agree the line in that example is noisy.

I'm just frustrated that something as terse as 
```js
foo.compose(d1, d2)
``` 

gets transformed to:

```js
foo.compose(
  d1,
  d2
);
```

I'm hoping that there's a way forward that keeps the functional composition people happy, which doesn't affect simple cases that are already fine on one line.

suchipi:
Maybe we should scope down eg `foo.compose` to only `R.compose`, etc from popular functional libraries, instead of any member expression. We would need to do some research to find common usage patterns in the wild among popular libraries.

duailibe:
After 2nd thought I think we should merge this.. it's still an improvement for several cases asked in the original thread. It's a trade-off and I'm more comfortable not breaking for FP users than breaking for non-FP users.

@suchipi

suchipi:
@duailibe I don't want to introduce back-and-forth churn for the (silent) users that were excited for this change in 1.13. I think it might finally be time to start checking scope to see if stuff is imported from a named package... and we should probably find a way to make that work with some kind of plugin API.

suchipi:
(For what it's worth, I am also affected by these false positives at work, so I don't intend to ignore this forever- but I think we need to be smarter about how we do it.)

benlesh:
As I commented [here](https://github.com/prettier/prettier/pull/4431#issuecomment-406715331) it might be simpler do something a little different.

karlhorky:
@benlesh This approach of making commas significant to prettier's decision of multiline vs inline would add manual steps for achieving certain formatting during editing code, which feels like it's against prettier's philosophy.

I think @jlongster has argued for this well in https://github.com/prettier/prettier/issues/3503#issuecomment-352804687:

> I'm certainly biased; it gives me great pleasure if I have a function whose args have been broken up:
> 
> ```
> function compile(
>   parser,
>   shouldExtendScopes,
>   lazilyCompile,
>   otherOption
> ) {
>   // ...
> }
> ```
> 
> And I remove otherOption and it collapses it back:
> 
> ```
> function compile(parser, shouldExtendScopes, lazilyCompile) {
>   // ...
> }
> ```
> 
> Removing that functionality would be a big loss IMHO.

arackaf:
Like others, I’m finding this PR after seeing simple things like 

@connect(null, { ...actions })

suddenly get broken into multiple lines.  I really hope Prettier will revert back to the old behavior, and just let the fp crowd configure whichever methods they want aggressive breaking on.  Guessing where to do this seems impossible, and an ideal use case for userland configuration.

benlesh:
@karlhorky I'm sorry, I don't think the argument is effective, because I think how people want these things broken up is too subjective for an algorithm to make good choices. A human decision is going to be better here. @arackaf's problem just above sort of proves my point. This is literally never going to make people happy in all cases until they have some form of control over the behavior on a case-by-case basis.

princed:
I'm closing this one since it's unlikely to be merged.
Happy to bring it back to life if the need arises.

arackaf:
Any hope of getting an option added so decorator users will stop having code mangled? Between this, and another recent change, decorators are becoming almost unusable with Prettier.

arackaf:
@lydell - yes! That’s amazing. I was really really hoping some hero would come along and do that PR. 

Any ETA on a release???

suchipi:
I think https://github.com/prettier/prettier/pull/5259 is the biggest thing that needs to be merged before we can release.

arackaf:
That’s unfortunate to hear. There’s no way to just get a patch release pushed with these decorator fixes, with Angular/Vue enhancements coming sometime later?

suchipi:
Right now the code for angular/etc is already in master, so if we released the branch as-is, users would use it before it's ready, which would cause issues. We would have to branch to make a release that disables HTML support, but for a release this big (it's been months), we usually want to write detailed release notes so people understand what changed, and the HTML stuff would probably be done before we finished the release notes. So, just a little bit of patience, please- we are eager to get these fixes out, too 🙏

arackaf:
Ah - understood. Thank you for taking the time to explain. Will be worth the wait - thanks a ton for all you guys do!

Issue #3107, opened 2017-10-25
Allow inline method chaining if line length does not exceed `printWidth`


sharmilajesupaul:
Just to give you some context on this, I've been working with @ljharb to bring prettier to @airbnb. 
We noticed a few cases where Prettier’s formatting could be adjusted to more closely align with our style guide.

cc. @vjeux

**Prettier 1.7.4**

If a method chain has more than 1 method, Prettier creates a linebreak after each chained method, increasing the indentation in the method body.

[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAEnADYkQDqEATiQCYAUAlPsADpT74DuOUdEbgDoAYhAgBhEjgQwhAZzgwAagENq8hu05d85MKpJwk+AOZKxEADIQDRrfgAOq8-gC+TADQdd+VehgACxocAC9VXGgAFQgAawQTOgjVIRg4hG8dDyE+HBU4ahxUAE8GAHJUcQB9SFhVPgKypgBuDh98VDgAK1UADyFVHt6tdq50WhMygHoANwKinANIqCmyzN8kmFUo4sdjfDKu+Wg19uyghAZR-AZN1RYAXgA+VmuuIJx5BSUAZS34ByfGyqOh8UwmVCGRTuFpvfAfL65GAUHB0cwwW7JWE6Lhuda6TFbR4vbS+d6BT5CcimACSAFtHNQ4PJ5HgoOVKhAqp0YGBAlUCtQaGt8HdsWT8CISKp5IEhIKaAwaQBGAAcUFS5QgqFQ0igcCqqNkeWKQk5hpZ6DgTXFujx7WxbhAnhAEEcy3kyFAMpgyEhJEULoARtRVGB4jAfs4wGDkDBqFaXQIwH6oXAXUdegAhUPh36qOlwKwNVMB9OugKOAIAJlLgZAzg0BWQICDqiDxXIUGdDcKsBRdCCyGVAAYRy7FHScHGE+XWVBTEYAIroCDwGeJkBbIMDodIasu+P1PWmCQQOl01QtwwkHvoRRRduepD+xRuNxAA)
```sh
# Options (if any):
--single-quote
--print-width 100
--trailing-comma "all"
```

**Input:**
```jsx
function HelloWorld() {
  window.FooClient.setVars({
    locale: getFooLocale({ page }),
    authorizationToken: data.token,
  }).initVerify('foo_container');

  fejax.ajax({
[...]

Issue #1565, opened 2017-05-09
Break on multiple chained calls


SimenB:
I've mentioned this possibility in a few different issue (#1099, #1282, #1401), but I thought a separate issue might be appropriate. 😄

I'd like for prettier to break on multiple chained calls. I prefer a higher `printWidth` than 80 in most cases (such as strings, log statements, class declarations etc), but increasing it leads prettier to one-line a lot of chained expressions that were multi-lined.
Eslint has a rule for this called [`newline-per-chained-call`](http://eslint.org/docs/rules/newline-per-chained-call). It defaults to 3 chained calls should break, which sounds like a sane starting point. This rule is unfixable, unfortunately, but Prettier could fix it! 😀

Even for the default width of 80 some things are inlined which IMO are more readable expanded.

I don't think there's any doubt which of these two are the easiest to read  (especially considering the git diff scenario from eslint's docs): https://prettier.github.io/prettier/#%7B%22content%22%3A%22d3%5Cn%20%20.select(%5C%22body%5C%22)%5Cn%20%20.selectAll(%5C%22p%5C%22)%5Cn%20%20.data(%5B1%2C%202%5D)%5Cn%20%20.enter()%5Cn%20%20.style(%5C%22color%5C%22%2C%20%5C%22white%5C%22)%3B%5Cn%5Cnd3%5Cn%20%20.select(%5C%22body%5C%22)%5Cn%20%20.selectAll(%5C%22p%5C%22)%5Cn%20%20.data(%5B1%2C%202%2C%203%5D)%5Cn%20%20.enter()%5Cn%20%20.style(%5C%22color%5C%22%2C%20%5C%22white%5C%22)%3B%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%7D%7D

azz:
Prettier will generally try to fit content in one line if possible.

I don't think a simple solution like "break when 3 or more chains" can be reasonably applied, consider:

```js
point().x(4).y(3).z(6).plot();
```

```js
point()
  .x(4)
  .y(3)
  .z(6)
[...]

PR #4765, opened 2018-06-28
Don’t break chains if they fit on one line & they were written that way

j-f1:
Ref #1565 & #3107.


j-f1:
cc @stephenh and @vjeux

tests/flow/promises/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -587,12 +587,10 @@ Promise.reject(0)
   });
 
 // resolvedPromise<T> -> catch() -> then():?T
-Promise.resolve(0)
-  .catch(function(err) {})
-  .then(function(num) {
-    var a: ?number = num;
-    var b: string = num; // Error: string ~> number
-  });
+Promise.resolve(0).catch(function(err) {}).then(function(num) {
```
vjeux: I think that you should also check the body of the method. I don't think that we want to inline this.

vjeux:
This is interesting. I'm not sure yet what I think about it. Would be nice to make all the examples that were written in the issues go through this PR and see how they look like now.

tests/flow/promises/__snapshots__/jsfmt.spec.js.snap
j-f1: Yeah I’m not sure why it decided to do that. Do you happen to know what the `path` is that’s passed to `printMemberChain`? Is it the whole expression (`Promise.resolve(0).catch(...).then(...)`) or is it just the `Promise.resolve(0)`?

bakkot:
> Would be nice to make all the examples that were written in the issues go through this PR and see how they look like now.

And add them as snapshots! :D

j-f1:
@bakkot & @azz done 🔼

suchipi:
I expect this will fix a lot of the complaints we've had about breaking chains in enzyme tests, too

tests/method-chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -638,6 +638,433 @@ var jqxhr = $.ajax("example.php")
...
+);
+
+// https://github.com/prettier/prettier/issues/1565#issuecomment-399123593
+const a1 = b1.c1().d1();
+const a2 = b2
+  .c2()
+  .d2()
+  .e2();
+
+fetchApi()
```
duailibe: Why did this break?

src/language-js/printer-estree.js
```diff
@@ -4571,7 +4571,12 @@ function printMemberChain(path, options, print) {
   // If the last group is a function it's okay to inline if it fits.
   if (
     hasComment ||
-    callExpressionCount >= 3 ||
+    (callExpressionCount >= 3 &&
+      hasNewlineInRange(
+        options.originalText,
+        options.locStart(node),
+        options.locEnd(node)
+      ) !== -1) ||
```
duailibe: `hasNewLineInRange` returns a boolean

tests/method-chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -638,6 +630,394 @@ var jqxhr = $.ajax("example.php")
...
+    .iso()
+    .min(Joi.ref('start'))
+    .required()
+  })
+);
+
+// https://github.com/prettier/prettier/issues/1565#issuecomment-399123593
+const a1 = b1.c1().d1();
+const a2 = b2
+            .c2()
```
stephenh: Shouldn't prettier force this `.c2()` to use two-space indent?

duailibe:
@stephenh The snapshots also contain the inputs (original code). The formatted parts are below “~~~~~~~~”

stephenh:
@duailibe ah shoot, sorry, newbie mistake. Thanks!

You're right, I'd assumed the snapshot was only formatted/expected output.

Might be nice to have smaller `issue-1565-comment-X.js` files, so that before/after are right next to each other by the `~~~~~~`.

j-f1:
Any ideas why Travis and Appveyor can’t see the relevant snapshots?

abeMedia:
Is this PR still active? Would be really good to see this feature merged in! (I hate how prettier currently deals with chained methods...)

pixelchutes:
Excited to find this thread and see the great progress towards this feature request.

@j-f1: has @duailibe been able to answer your question re: the failing Travis/Appveyor checks?

duailibe:
Sorry I've been missing I'll review it this weekend

tests/flow/node_tests/crypto/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -70,10 +70,7 @@ let tests = [
     hmac.update(buf, "utf8"); // 1 error: no encoding when passing a buffer
 
     // it's also chainable
-    (hmac
```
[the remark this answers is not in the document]
j-f1: If you like it better on multiple lines, you can manually break it and it’ll stay that way.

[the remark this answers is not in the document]
azz: That's the thing though, If *I* was writing it I'd (hopefully) understand it and probably want it to be on one line. But the next person to read the code would have found it more readable when split onto multiple lines 😉. Part of the value of Prettier is making _other people's_ code easier to read, less so your own.

RomainLanz:
Looking forward for this feature to be merged 👍

ikatyang:
Travis/AppVeyor uses `git checkout -f` while Circle uses `git checkout`:

```
-f, --force
    When switching branches, proceed even if the index or the working
    tree differs from HEAD. This is used to throw away local changes.

    When checking out paths from the index, do not fail upon unmerged
    entries; instead, unmerged entries are ignored.
```

merlinnot:
Can we have an option to disable it? This [comment](https://github.com/prettier/prettier/pull/4765#discussion_r205958360) is so true...

karlhorky:
> This comment [by @azz] is so true...

Yeah it feels like a sub-optimal solution to me too. But in my opinion, adding more options to Prettier is also not great.

I would instead challenge this feature completely, based on it allowing for choice in the final formatting of a file. I feel similarly to what @vjeux mentioned in https://github.com/prettier/prettier/pull/4862#issuecomment-410300314:

> The solution proposed here is going to make code formatted in different ways based on the way you type it. So it makes a dent in the "just write everything in one line and save" feeling. And also now that you have a choice, people may ask you to do it a certain way in pull requests.

The important part here for me is the last sentence about choice and about allowing more bikeshedding in pull requests, which was one thing that was very refreshing for me about switching to Prettier.

---

However, I can imagine that there may need to be exceptions to this rule, if this is important enough to enough users.

Maybe an alternative would be it being implemented as a default-off option? `preserveSingleLineChains` or something?

karlhorky:
Also apart from this pull request, I see a lot of discussion around adding more configuration around how Prettier formats things (either config options or "hints" in the form of how you write code).

Depending on how important such options are to users in the future and how important the philosophy of zero configuration is to Prettier, this and future options which make Prettier less zero-config could also be prefixed under a namespace (or potentially even moved to an extension). This would allow the core to stay opinionated and simple, while still offering options for other flavors.

Not personally taking a position that this is the right way to go, but it may be an escape hatch for users with different opinions while still allowing the vanilla Prettier experience to be zero-config.

j-f1:
Tests seem to be fine now 🤷‍♂️

SimenB:
I'll be sad if this lands without #4685 also landing (I know that one's WIP, but still). Then the new behaviour introduced in this PR would not be noticeable to those who don't want "hints" to impact how prettier formats things

suchipi:
I dislike hinting for reasons I've explained before but I'm not willing to die on this particular hill 🤷‍♀️

vicky-blr:
> Is this PR still active? Would be really good to see this feature merged in! (I hate how prettier currently deals with chained methods...)

I'm trying to find out how to disable this behavior in ``.prettierrc`` file. If there a way? Can we disable it somehow right now?

lydell:
@vicky-blr The only current way is to use `// prettier-ignore` comments.

josephMichaelRyan:
Is there any plan to merge this PR? This would be really useful to have to for large codebases that will take considerable effort to refactor. I would rather not have to litter everything with `// prettier-ignore` comments. Think it would make a really nice feature for folks who want to use prettier but would need to take some time refactoring their codebase.

lydell:
@josephMichaelRyan Can you share a few examples where Prettier's current output is unaccepatable?

josephMichaelRyan:
@lydell currently running Prettier via eslint, which is failing with a total of 1400+ errors. About 40 of these are other issues, but the remainder are due to the fact the codebase has been written so as to chain functions on a single line.
We want to bring Prettier in because we like it, but essentially without the option to switch off the rule that breaks up chained commands, it would make it hard for us to also use the VSCode extension.

Sm00g15:
cant wait for this to make it through!

vielhuber:
What's the current status on this?

j-f1:
@suchipi Would you be OK with merging this.

suchipi:
@j-f1 fine with me but maybe check with @lydell

lydell:
Let me see if I understand how this works.

From the beginning we had one rule: printWidth.

Then we added another rule on top of that: If >=3 items, always break.

This PR would replace that rule with: If >=3 items, allow the user to break if they want.

Is that correct?

---

Now, it's funny because I stumbled into this myself today. I wrote this:

```ts

function featuresEqual(a: Array<string>, b: Array<string>): boolean {
    return a.slice().sort().join(',') === b.slice().sort().join(',');
}
```

Which was turned into this:

```ts
function featuresEqual(a: Array<string>, b: Array<string>): boolean {
    return (
        a
            .slice()
            .sort()
            .join(',') ===
        b
            .slice()
            .sort()
            .join(',')
    );
}
```

Yikes! Because of that, I refactored to:

```ts
function featuresEqual(a: Array<string>, b: Array<string>): boolean {
    const comparable = (features: Array<string>) =>
        features
            .slice()
            .sort()
            .join(',');
    return comparable(a) === comparable(b);
}
```

So Prettier kind of nudged me into writing better code? (Still, that `featuresEqual` function is kinda weird.)

On the other hand, it feels a little like method chaining is used for so many things in JavaScript that it's really difficult to find a heuristic that works everywhere, just like object literals and decorators. Perhaps we should follow the Principle of least astonishment and merge this.

So … yes, I'm onboard.

shellscape:
let's get this badboy merged! 🌮 🎉

shellscape:
> I know that the current heuristic isn't perfect (almost by definition)

It's about as far from perfect as it could get. To the point where code is littered with directives turning off prettier for an incredible amount of lines for any projects that use libs/components that are designed to be chained.

> azz requested changes 6 hours ago

What are the changes being requested? I couldn't find any new comments.

Speaking as a maintainer of many projects, I really don't understand why there's a personal ideology being pushed here to block this PR when there is very clear user support for the change. Google searches reveal mountains of frustration with prettier over this, the reactions in the thread are overwhelmingly in favor of the change, and there seems to be no evidence of harm from the change, only relief for your users.

j-f1:
Also, re: the object issue, you now just need to delete the newline after the `{` to collapse the object.

lydell:
@j-f1 While you’re here, is my understanding of the new rules correct? https://github.com/prettier/prettier/pull/4765#issuecomment-471034767

j-f1:
That’s correct. < 3 items will never break, and >= 3 items will break if there’s a newline somewhere in the chain.

aboyton:
I’ve probably missed the discussion elsewhere by why not <= 3 don’t break, > 3 always break? That’s similar to what Airbnb does and it appears to work well.

azz:
> It's about as far from perfect as it could get. [...]

Well that's not true, it handles quite a lot of cases well. I'm open to _improving_ the heuristics, but I'm against using the input text as a factor.

> What are the changes being requested? I couldn't find any new comments.

The changes request aren't implementation changes, I was marking that I don't think we should merge this as-is. 

> Speaking as a maintainer of many projects, I really don't understand why there's a personal ideology being pushed here to block this PR when there is very clear user support for the change.

It's not so much a personal ideology, but what I think is an understanding of how and _why_ Prettier works, and a fear that these type of changes muddy how people treat the tool.

karlhorky:
> I'm continually manually deleting spaces just to trick prettier to format an object on one line

+1 for avoiding introducing features encouraging this kind of user behavior. I see this as similar to configuration, which is avoided due to the philosophy.

Similar to how I [commented earlier in the thread](https://github.com/prettier/prettier/pull/4765#issuecomment-410649086), I think that being deterministic based on the AST, not on the input text, is a nicer, more predictable developer experience.

This was also voiced by @vjeux in another issue - maybe @vjeux would be open to taking another look here for an updated opinion?

> I'm open to improving the heuristics

Also +1, maybe we can find another option that achieves the goals without being sensitive to newlines?

lydell:
I also want to improve the heuristics – but after all this time we still haven't come up with something better.

ekropotin:
Any chance it will be merged someday?

berrylaw:
Next time I didn’t study for it

「David Vielhuber <notifications@github.com>」在 2019年3月8日 週五，下午9:10 寫道：

src/language-js/printer-estree.js
```diff
@@ -4982,7 +4982,12 @@ function printMemberChain(path, options, print) {
   // If the last group is a function it's okay to inline if it fits.
   if (
     hasComment ||
-    callExpressions.length >= 3 ||
+    (callExpressionCount >= 3 &&
```
ekropotin: Where is callExpressionCount defined?

j-f1: Oops, looks like that didn’t get updated when I merged the latest changes in.

nicoandmee:
#blessed godspeed to release

gustsu:
Would love to see this get merged

shamilovtim:
please allow one line chaining or at least provide a setting to disable chaining being split across multiple lines. it's very difficult to read simple method chains this way and in my opinion it's not idiomatic unless the chained methods have callbacks. **only if chained methods have callbacks should they be split across multiple lines.**

MarvinJWendt:
First of all I really like prettier. This implementation would fix my only concern about using prettier. For example:

I have this **very easy to read** if statement
```javascript
if ($(el).attr("href").includes("/wiki/")) {
  ...
}
```

And prettier turns it into
```javascript
if (
$(el)
 .attr("href")
 .includes("/wiki/")
) {
  ...
}
```

*I like the first one waaay more.*

Let's just say we are nesting some if statements:

```javascript
if (
  $(el)
  .attr("href")
  .includes("/wiki/")
) {
  if (
    $(el)
    .attr("xyz")
    .includes("/whatever/")
  ) {
    if (
      $(el)
      .attr("hello")
      .includes("/world/")
    ) {
      ...
    }
  }
}
```
.. could be turned into:

```javascript
if ($(el).attr("href").includes("/wiki/")) {
  if ($(el).attr("xyz").includes("/whatever/")) {
    if ($(el).attr("hello").includes("/world/")) {
      ...
    }
  }
}
```

I think that the majority would prefer the compact one-line version.

---

## But
The point that we use prettier on **other peoples** code is solid. Yes we do not want them to decide the style. We want to use prettier and reformat all of their code to a readable version of it.

I guess one way to implement this is with a config option. 
Something like: `"prefer-oneline: [true/false]"` for example.

> I’ve probably missed the discussion elsewhere by why not <= 3 don’t break, > 3 always break? That’s similar to what Airbnb does and it appears to work well.

I really like this approach. Prettier wouldn't decide on the users code then.
(This could also have a config key to increase/decrease the number of allowed chains to make absolutely everyone happy.)

As much as I'd like to have prettier format the code in **my way**, I agree with @azz that this is not how prettier should work. There are other formatters for that. Prettier should format every piece of code to the same result (if the same config is given).

## TLDR:
> <= 3 don’t break, > 3 always break

Seems like a good solution for me.

epidemian:
Hi :wave: 

Total Prettier newbie here. I really dig the "ctrl+s and let the machine take care of formatting" approach that Prettier endorses, and i want to incorporate it in my everyday usage :slightly_smiling_face:. But when integrating Prettier on some small projects, i noticed these sort of formatting choices related to either splitting single-line expressions (below the `print-width` value) into multiple lines or collapsing muti-line expressions and they caught me off guard.

I think the approach proposed in this PR would be better for Prettier than the current behavior.

As a simple example, i had this function on a little project:

```javascript
// Parses a string like '321' into [1, 2, 3]
const parseNumbers = s => s.split('').map(Number).sort()
```

Which Prettier formatted as:

```javascript
const parseNumbers = s =>
  s
    .split('')
    .map(Number)
    .sort()
```

This function is inside the scope of a bigger function, and i intended it to be just a simple detail. Therefore the implementation as a succinct one-liner. To me, Prettier's re-formatting of this function is problematic for various reasons:

- First, it was **very unintuitive**. The original code being below `print-width`, i had no idea why it was being formatted like that, and i thought i had some syntax error or something. I noticed that when i removed the `.sort()` or even the parentheses on that `.sort` call, the code was re-formatted to a one-liner. It took me quite some time to understand that this was an intended behavior.

- It feels arbitrary and inconsistent with the rest of formatting rules. For example, the same function could be implemented in terms of top-level functions as `s => sort(map(Number, split('', s)))` or with getters like: `s => s.chars.map(Number).sorted`, and none of those would be split into multiple lines, even though, semantically, the operations are the same. Why would chained method calls be special in this regard compared to property accesses or function calls?

- Lastly, and this is _very_ subjective. It doesn't respect the writer's intent of splitting a statement into multiple lines or not. In the example i mentioned, my intent was for the little function to be just a passing detail, nothing of importance, but Prettier's formatting makes it take a considerable chunk of the scope it's declared in, so it seems much more important.

Expanding a bit on the last point, which i think is the most important, deciding whether a single piece of code ("statement", "expression") should be split into multiple simpler ones or not is a core part of programming. It's not the same as a menial formatting  adding spaces between operators, which i think everyone using Prettier agrees should be automated :slightly_smiling_face: 

For example, a statement like:

```javascript
me.eat(new Cake(getCakeIngredients()))
```

Could safely be written as:

```javascript
const ingredients = getCakeIngredients()
const theWholeCake = new Cake(ingredients)
me.eat(theWholeCake)
```

For the computer, they behave exactly the same. But for human readers, i could imagine cases where either of these styles would be preferable. Each has its pros and cons. For example, if i wanted to debug line-by-line, or to emphasize that i'm _eating the whole cake_, i would prefer the second example. But if that `me.eat()` call is part of many other steps to do, maybe the second style would feel too verbose and the first one would be preferable.

What i'm really sure is that **nobody would expect an automatic tool deciding to transform the first snippet into the second one or vice-versa**. For now at least, it's part of a human programmer's job to make that decision, and maybe for tools to assist them in that if necessary.

On a similar vein, i wouldn't want a an automatic tool transforming this:

```javascript
me
  .grabCakeIngredients()
  .prepareCake()
  .eat()
```

Into this:

```javascript
me.grabCakeIngredients().prepareCake().eat()
```

Or vice-versa.

If the programmer wrote it in separate lines, it's because they wanted to communicate that each method is to be thought as a "separate step". Whereas if they decided to write it on a single line they want to communicate that this chain is to be considered a single operation. Leaving aside the eating cake example, it's easy to imagine examples for either case.

As a bad analogy, i think there's not so much controversy with other formatting rules of Prettier because they are like fixing punctuation errors or typos on prose. It's very useful to have a tool that fixes those automatically. But the rule of splitting method chains into multiple lines is like splitting longish sentences into many short ones. Maybe useful for a tool to notice, but not its decision to make :slightly_smiling_face: 

---

Or, putting this in another way, and quite more bluntly: do you think there would be as many requests to change the formatting of chained method calls if the original rule was the one proposed on this PR?

toschlog:
Please make this change.  I'm considering using Prettier because we're bringing on someone who uses it, but I can't stomach splitting chains into multiple lines.

VictorCamargo:
Can't wait for this...

hoshinokanade:
Splitting chaining calls are against to what Prettier suggests.
> Prettier, on the other hand, strives to fit the most code into every line. 
[source](https://prettier.io/docs/en/options.html)

When it comes to fluent api, Prettier is fitting almost no code into every line.
I do think Prettier should stay with the mentioned style at the same time to response to our voices.\
Please get it merged.

rikkertkoppes:
I may have overlooked it, but has no one proposed to make the number of chains configurable? That is:

- break into separate lines if we have more than _n_ components in the chain
- break into separate lines if the line length exceeds the max line length
- join into one line otherwise

This would also allow to configure only s specific part of the repo differently with file overrides. For example, test files tend differ in coding style from other files as mentioned before. If I can allow long chains in tests, that would already be a huge win.

vytautas-pranskunas-:
Any progress on this? And do not get thi swhy 3? Why cannot be just config propertybecause my team want to have all shained methods on separate line.

jonathas:
Please guys, this is much needed!

thorn0:
The problem here is really similar to the one from #6033, isn't it? So I think changing the current heuristic in the spirit of https://github.com/prettier/prettier/pull/6033 (like also mentioned in [this comment](https://github.com/prettier/prettier/pull/4765#issuecomment-502408355)) should do the trick. A bit more sophisticated option: use the `>= 3` rule if there are function literals in arguments, switch to the `> 3` rule otherwise. What do you all think about this?

lselbeck:
Pretty please, prettier please!

MarvinJWendt:
Is there any status update on this? It's the most upvoted issue on the repo here. I am wondering if this pull is unofficially declined now?

ltb-bossuyt:
Hello, any updates ?

farshed:
Any chances this'll ever get merged? Its been 16 months already 🙄

lydell:
@farshed We’re currently exploring #6685

thorn0:
Closing this since https://github.com/prettier/prettier/pull/6685 has been merged.

syabro:
@thorn0 it will be released with 2.0?

rvion:
IMO: it's now as annoying as before for a different set of use-cases, notably fluent APIs. 

As a library author with dozen builder types libs with fluent apis, I now need to add comments everywhere to ensure lines are broken correctly.

Is there at least a way to choose previous heuristics ?

here is an example I'm editing right now : 

```ts
domain.concept('Page').val('title', 'string').vals('widgets', 'Widget')
domain.concept('Widget').val('title', 'string').val('color', 'Color')
. . .

```

here is what it used to be

```ts
domain //
    .concept('Page')
    .val('title', 'string')
    .vals('widgets', 'Widget')

domain //
    .concept('Widget')
    .val('title', 'string')
    .val('color', 'Color')

```


some people warned about this before with upvotes.

- https://github.com/prettier/prettier/pull/4765#issuecomment-508851575
- https://github.com/prettier/prettier/pull/4765#issuecomment-514094929

example mentioned above:

```
me
  .grabCakeIngredients()
  .prepareCake()
  .eat()
```

vs

```
me.grabCakeIngredients().prepareCake().eat()
```

rvion:
@hwalinga @RIkJansen81 and others => discussion is happening here: https://github.com/prettier/prettier/issues/7884

your feedback there would be very much appreciated :)

Issue #4199, opened 2018-03-24
Prettier v1.11.1 is mixing spaces and tabs for alignment


kiruh:
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAhgAsBXKAa3wDMcJ18AdEHOAExIA8BaYs8pgNwMoGbHnzB8YFgEN4AGQgBzJXBz4AvtVr0mLdtwA2y1TkHDRuApOlw5cAMoxccADT4ZmTIYCeAWVRWVkM4AHcZFndILAgAZzhNbTpGZjZOcygLGPF9EjA1JN0QADoAelz8nFjSjOFIKFiCVFiAETgAN3wAXkIcEjghKHrG-GMlACU0jm7e-sHhghJ4xRU1AEEoVjb2gBUICENYmcxafNjY4oR24oA5AHkWgFEAfUebgDV8AEIunqZ4RpMfAAMmB+Ga20GdWgI0aLhmSzgK1MGy2HT2B1iwgYMAA-FJZPAnC4ABTYmA4nEVNSucmUmBjSYGOkU1n46LYeIkzzefyBYJhCJwEm2ezItQkgCU7l4FElkpZ9KQUjonOFPN8ASCIXCLBJsvI8pZCsyMGVoqJzj11Jw7g1fO1gr1BvlUKgcA4YgIrDgVBkJEMBDhLEGIFcIAgmBgqBhyFAEVooQACkKGsgQDJ2hBAmGM41kH7Dm4QAAjHAyMDkOAwByYCuoKBKZAwPrF1gQMAFmRF8MN+J4JPlpToGRdnsgABWsQ4ACFy5Xqw4ZOgkQ24GP4uGICQYJgdwAmDfFutVNTpksyEs+YxQXMnBswADqgWIyAAHAAGcMnOJwR-lzB0xOOB+3addwxYABHEhUBYQcZGHUckELTcQHidBUGbVtw1iBslBCABFEgIHgI9wxgS9n1YV8kAAFnI8tUEMPCAGE6BHdMoGgcCQERHZL1iLD+g0DQgA)
```sh
# Options
--use-tabs
--tab-width 4
```

**Input:**
```js
const store = useLoggerAndDevTools
	? createStore(
			reducer,
			logRedux
				? compose(applyMiddleware(createLogger(), thunk))
				: compose(applyMiddleware(thunk))
>		)
[...]

PR #4767, opened 2018-06-28
Adjust indentation of nested ternaries

zeorin:
See https://github.com/prettier/prettier/issues/737 for the issue that sparked the discussion.

Primarily, it changes ternaries that were of the form (nesting in alternate sub-expressions):
```js
aaaaaaaaaaaaaaa
  ? bbbbbbbbbbbbbbbbbb
  : ccccccccccccccc
    ? ddddddddddddddd
    : eeeeeeeeeeeeeee
      ? fffffffffffffff
      : gggggggggggggggg;
```
into:
```js
aaaaaaaaaaaaaaa ?
  bbbbbbbbbbbbbbbbbb :
ccccccccccccccc ?
  ddddddddddddddd :
eeeeeeeeeeeeeee ?
  fffffffffffffff :
gggggggggggggggg;
```
as suggested in https://github.com/prettier/prettier/issues/737#issuecomment-363314699.

It also changes ternaries that nested consequent sub-expressions slightly to be more consistent with the new alternate form.

See the changes to ` tests/ternaries/__snapshots__/jsfmt.spec.js.snap` for a great overview of the new format.

JSX-mode ternaries were not changed.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -223,7 +223,6 @@ function formatTernaryOperator(path, options, print, operatorOptions) {
   // See tests/jsx/conditional-expression.js for more info.
   let jsxMode = false;
   const parent = path.getParentNode();
-  let forceNoIndent = parent.type === operatorOpts.operatorName;
 
   // Find the outermost non-ConditionalExpression parent, and the outermost
   // ConditionalExpression parent. We'll use these to determine if we should
@@ -246,7 +245,6 @@ function formatTernaryOperator(path, options, print, operatorOptions) {
     conditionalExpressionChainContainsJSX(lastConditionalParent)
   ) {
     jsxMode = true;
-    forceNoIndent = true;
 
     // Even though they don't need parens, we wrap (almost) everything in
     // parens when using ?: within JSX, because the parens are analogous to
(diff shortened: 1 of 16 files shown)


hawkrives:
This also changes single-step ternaries, i.e.

```js
mySuperDuperExtraLongTernary
  ? "yes"
  : "no"
```

becomes

```js
mySuperDuperExtraLongTernary ?
  "yes" :
  "no"
```

and … I really really like the current placement of `?` / `:`.

I'm OK with this proposal for multi-step ternaries, but then that'd be inconsistent.

My 2¢.

zeorin:
@hawkrives that was intentionally done, since the convention for other operators is to keep the operator at the end of the line instead of the beginning when breaking. Ternaries were the odd one out here. I assumed that consistency would be desired.

It'd be easy enough to account for non-nested ternaries and keep the old behaviour of placing the operators at the beginning of the line, I don't mind doing it, but how do I know what the Prettier maintainers decide?

bakkot:
Personally not a fan of this style (in large part because it seems very much out of keeping with how most people write JS), especially for non-nested ternaries, but that aside, one thing:

```js
aaaaaaaaaaaaaaa ?
  bbbbbbbbbbbbbbbbbb :
ccccccccccccccc ?
  ddddddddddddddd :
eeeeeeeeeeeeeee ?
  fffffffffffffff :
gggggggggggggggg;
```

why is the `gggggggggggggggg` not indented? That seems really odd, to my eye, since `gggggggggggggggg` is a value like `b` / `d` / `f` and not a condition like `a` / `c` / `e`.

zeorin:
It is not indented because it is the form of the proposal suggested in https://github.com/prettier/prettier/issues/737#issuecomment-363314699. There was no clear direction or expressed preference for a specific proposal in that thread from Prettier maintainers, so I implemented the one that I felt had the most merit.

As to why I personally like the final value to be dedented, it's because it's like a `default` case in a switch statement. It's the value of the entire expression when none of the tests match. Not dedenting it makes it seem like it's related only to the last test when it's related to all the tests.

iclanzan:
I've never seen ternaries formatted like this before. At first glance I was disturbed but then I realized it makes sense, it's consistent and actually very readable. I think I like it! 
Splendid work @zeorin!

suchipi:
I think this style is too significant of a change. If nothing else, we shouldn't change the way non-nested ternaries format.

zeorin:
I do think this style has the most merit, in addition to the points raised by @0x24a537r9 in https://github.com/prettier/prettier/issues/737#issuecomment-363314699:
* operators at the end are easier to work with when maintaining the code, needing less nit-picky moving around of operators (which, admittedly, use of Prettier-on-save can obviate), and thus also leading to less noise in VCS diffs (much like trailing commas in multiline object & array literals).

Myself, I wasn't a fan of this proposal at first, either, but over time it grew on me, on the strengths of its merits. (Incidentally, I went through the same process when I first encountered trailing commas.)

I personally think that changing non-nested ternaries to be consistent with nested ternaries is more important than some one-time mental friction of getting used to the new style (especially as it would be consistent with operators in general, not just the nested ternaries).

_Having said that_, I mostly care about removing the "infinite stairs" for alterate sub-expression ternary nesting and the other details of the style don't matter nearly as much.

I'm happy to make changes to whatever style is settled on, but at this point this thread is suffering from the same bike-shedding as #737 was. I was hoping that shipping some actual code of an implementation would act as a sort of Occam's Razor and we could move on.

This means that I'm not sure what to change, as whatever I change it to, someone will have issue with. Who knew that ternary style preferences are so diverse?

Anyway, I think we need to make a decision on what style we go with, and to that end, @vjeux, please will you make a call?

lydell:
I think we should experiment with a style that causes as little change as possible to maximize the chance of moving forward. So don’t move the `?` and `:`. Just flatten the indentation (when it makes sense). Optimize for the common case.

```typescript
type TypeName<T> = T extends string
  ? "string"
  : T extends number
  ? "number"
  : T extends boolean
  ? "boolean"
  : T extends undefined
  ? "undefined"
  : T extends Function ? "function" : "object"
```

suchipi:
Removing my assignment because I intended to assign myself to the issue, not the PR.

thgreasi:
@zeorin many :+1:s on this.
How will this work for cases like #4199 #4874 #4979 ?

suchipi:
@thgreasi I don't intend to change that behavior as part of this. That is intended behavior; please comment on one of those issues if you have input.

ikatyang:
Fixed in #5039.

zeorin:
Whilst I'm glad that this issue is now fixed (:tada:), It's regrettable that there was no direction from the maintainers on this PR.

There was a lot of discussion, sure, by many people, but not one maintainer saying something like: "Please tweak foo, bar, and baz, and then it's good to merge".

The barrier to entry on making this PR was pretty high. I read the original paper to get a sense of how Prettier works, and spent a whole weekend coming up with the first draft of the PR, tweaking it a bunch over the following days.

That effort was wasted.

I urge the Prettier maintainers to be more forthcoming and transparent.

vjeux:
@zeorin I'm sorry that you didn't have the best experience here.

This particular issue has been long standing and is likely one of the oldest issues still opened. The reason is that we needed to decide on the way we wanted to print this piece of syntax. Your PR has actually been critical to get the discussion started again and it finally drove to a final decision.

About the actual implementation, it's a bit unfortunate that we created a new PR rather than asking you to update yours. There are two reasons for this:
- The eventual design was different than the one you suggested and it wasn't immediately obvious that code could be shared.
- The actual implementation has never been an issue in this case. This is a very simple change for anyone that has already contributed to prettier. So it only took something like 5-10min to spin up the PR and it would have likely taken an order magnitude more time to teach you how to morph your PR into the final one.

Thanks for raising this issue and I hope that this gives you some insight around why things happened this way. But please remember that your PR was not useless, this long standing bug has been closed now because of you.

Issue #3652, opened 2018-01-05
Make the playground example a placeholder


j-f1:
This would allow the user to enter their code right away without having to delete the example first. Plus, it would make the playground easier to use on mobile — it’s really hard to delete large amounts of content in the code editor on an iOS device.

Also, would it be possible to have different examples for each language? The JS one doesn’t work very well for CSS and Markdown.

(ref https://github.com/prettier/prettier/pull/3573#issuecomment-354602726)

lydell:
FYI: There is a "Clear" button.

PR #4770, opened 2018-06-29
feat(playground): use real code sample

### LINKED ISSUE #4769: Cant modify playground text
Yeah just cant modify ant text on the playground regardless of browser or extensions.

**Prettier 1.13.7**
[Playground link](https://prettier.io/playground/)
```sh
--parser babylon
```

**Input:**
```jsx
function HelloWorld({greeting = "hello", greeted = '"World"', silent = false, onMouseOver,}) {

  if(!greeting){return null};

     // TODO: Don't use random in render
  let num = Math.floor (Math.random() * 1E+7).toString().replace(/\.\d+/ig, "")

  return <div className='HelloWorld' title={`You are visitor number ${ num }`} onMouseOver={onMouseOver}>

    <strong>{ greeting.slice( 0, 1 ).toUpperCase() + greeting.slice(1).toLowerCase() }</strong>
    {greeting.endsWith(",") ? " " : <span style={{color: '\grey'}}>", "</span> }
    <em>
	{ greeted }
	</em>
    { (silent)
      ? "."
      : "!"}

    </div>;

}
```

**Output:**
```jsx
function HelloWorld({
  greeting = "hello",
  greeted = '"World"',
  silent = false,
  onMouseOver
}) {
  if (!greeting) {
    return null;
  }

  // TODO: Don't use random in render
  let num = Math.floor(Math.random() * 1e7)
    .toString()
    .replace(/\.\d+/gi, "");

  return (
    <div
      className="HelloWorld"
      title={`You are visitor number ${num}`}
      onMouseOver={onMouseOver}
    >
      <strong>
        {greeting.slice(0, 1).toUpperCase() + greeting.slice(1).toLowerCase()}
      </strong>
      {greeting.endsWith(",") ? (
        " "
      ) : (
        <span style={{ color: "grey" }}>", "</span>
      )}
      <em>{greeted}</em>
      {silent ? "." : "!"}
    </div>
  );
}

```

**Expected behavior:**

ikatyang:
Closes #4769

[preview playground](https://deploy-preview-4770--prettier.netlify.com/playground/)

### DIFF
--- a/website/playground/panels.js
+++ b/website/playground/panels.js
@@ -9,7 +9,6 @@ class CodeMirrorPanel extends React.Component {
     this._cached = "";
     this._overlay = null;
     this.handleChange = this.handleChange.bind(this);
-    this.handleDoubleClick = this.handleDoubleClick.bind(this);
   }
 
   componentDidMount() {
@@ -26,7 +25,6 @@ class CodeMirrorPanel extends React.Component {
       options
     );
     this._codeMirror.on("change", this.handleChange);
-    this._codeMirror.on("dblclick", this.handleDoubleClick);
 
     this.updateValue(this.props.value || "");
     this.updateOverlay();
@@ -40,6 +38,9 @@ class CodeMirrorPanel extends React.Component {
     if (this.props.readOnly && this.props.value !== this._cached) {
       this.updateValue(this.props.value);
     }
+    if (this.props.value !== prevProps.value) {
+      this.updateValue(this.props.value);
+    }
     if (
       this.props.overlayStart !== prevProps.overlayStart ||
       this.props.overlayEnd !== prevProps.overlayEnd
@@ -48,9 +49,6 @@ class CodeMirrorPanel extends React.Component {
     }
     if (this.props.mode !== prevProps.mode) {
       this._codeMirror.setOption("mode", this.props.mode);
-    }
-    if (this.props.placeholder !== prevProps.placeholder) {
-      this._codeMirror.setOption("placeholder", this.props.placeholder);
     }
     if (this.props.ruler !== prevProps.ruler) {
       this._codeMirror.setOption("rulers", [makeRuler(this.props)]);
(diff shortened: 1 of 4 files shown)


website/playground/panels.js
```diff
@@ -150,7 +159,7 @@ export function InputPanel(props) {
       autoCloseBrackets={true}
       matchBrackets={true}
       showCursorWhenSelecting={true}
-      tabWidth={2}
+      tabSize={4}
```
ikatyang: http://codemirror.net/doc/manual.html#option_tabSize

Not sure if it should be 2 or 4.

lydell:
People wouldn't know to double-click, though?

ikatyang:
I thought people would random click it if they think it's editable but not able to edit it. 🤔

hawkrives:
How about if it's just actual content by default? I'm always confused that it's a placeholder and I have to remind myself that I don't have to clear it out.

ikatyang:
Then there will be no code example for other languages, see #3652 for why placeholder exists.

duailibe:
> Then there will be no code example for other languages

I don't understand this. What does this have to do with the placeholder?

hawkrives:
@ikatyang yeah, and I've been confused by it ever since 😜

As an actual reason to switch to a code sample, it'll allow a potential user to experiment by editing the sample and watch how the format changes from the sample text.

lydell:
I think it's better to go back to real text again, as long as we have an easy to find Clear button.

When switching language we could do:

```
if content == oldLanguage.placeholder:
    content = newLanguage.placeholder
```

lydell:
I like it!

I'm just a little bit worried that people won't see the Clear button.

Does it make sense to automatically select all text when focusing the text area? (Perhaps only if it contains sample code?) That one can paste or start typing without having to clear first.

@j-f1 since you requested the placeholder thing in #3652 – what do you think?

j-f1:
The concept seems good, but anything I type comes out backwards :/ (on an iPhone)

duailibe:
Yeah it's prepending text when typing

website/playground/panels.js
```diff
@@ -38,6 +38,9 @@ class CodeMirrorPanel extends React.Component {
     if (this.props.readOnly && this.props.value !== this._cached) {
       this.updateValue(this.props.value);
     }
+    if (this.props.value !== prevProps.value) {
+      this.updateValue(this.props.value);
+    }
```
duailibe: I think you can just remove the `this.props.readOnly` above

lydell:
Thoughts on my auto-select-text idea?

ikatyang:
Feels not bad, though it doesn't work on mobile.

j-f1:
Sounds reasonable @lydell.

PR #4772, opened 2018-06-29
Add a new `trim` command to trim whitespaces in the current line (#4771).

### LINKED ISSUE #4771: Trim whitespaces without inserting a newline
While trying to support preprocessor directives (#4678), I found out that there is no way to explicitly trim whitespaces inserted by alignment. @ikatyang [mentioned](https://github.com/prettier/prettier/issues/4678#issuecomment-398799227) that it is possible to remove trailing spaces and use the root indentation with `dedentToRoot(hardline)` but it will insert a newline.

Is there a proper way to do it that I missed?

Otherwise, I'll link a MR that introduce a new command.

warrenseine:
I reused and factorized the pre-existing code to trim trailing spaces.

Note that this adds no test, and I'm not really confident about that. How would you test this change? I couldn't find any test that explicitly used `hardline` for instance.
Fixes #4771

### DIFF
--- a/commands.md
+++ b/commands.md
@@ -237,6 +237,14 @@ declare function dedentToRoot(doc: Doc): Doc;
 
 This will dedent the current indentation to the root marked by `markAsRoot`.
 
+### trim
+
+```ts
+declare var trim: Doc;
+```
+
+This will trim any whitespace or tab character on the current line. This is used for preprocessor directives.
+
 ### cursor
 
 ```ts
--- a/src/doc/doc-printer.js
+++ b/src/doc/doc-printer.js
@@ -109,9 +109,37 @@ function generateInd(ind, newPart, options) {
   }
 }
 
+function trim(out) {
+  if (out.length === 0) {
+    return 0;
+  }
+
+  let trimCount = 0;
+
+  // Trim whitespace at the end of line
+  while (
+    out.length > 0 &&
+    typeof out[out.length - 1] === "string" &&
+    out[out.length - 1].match(/^[ \t]*$/)
+  ) {
+    trimCount += out.pop().length;
+  }
+
+  if (out.length && typeof out[out.length - 1] === "string") {
+    const trimmed = out[out.length - 1].replace(/[ \t]*$/, "");
+    trimCount += out[out.length - 1].length - trimmed.length;
+    out[out.length - 1] = trimmed;
+  }
+
+  return trimCount;
+}
+
 function fits(next, restCommands, width, options, mustBeFlat) {
   let restIdx = restCommands.length;
   const cmds = [next];
+  // `out` is only used for width counting because `trim` requires to look
+  // backwards for space characters.
+  const out = [];
   while (width >= 0) {
     if (cmds.length === 0) {
       if (restIdx === 0) {
@@ -130,6 +158,8 @@ function fits(next, restCommands, width, options, mustBeFlat) {
     const doc = x[2];
 
     if (typeof doc === "string") {
+      out.push(doc);
+
       width -= getStringWidth(doc);
     } else {
       switch (doc.type) {
@@ -145,6 +175,10 @@ function fits(next, restCommands, width, options, mustBeFlat) {
           break;
         case "align":
           cmds.push([makeAlign(ind, doc.n, options), mode, doc.contents]);
+
+          break;
+        case "trim":
+          width += trim(out);
 
           break;
         case "group":
@@ -184,6 +218,8 @@ function fits(next, restCommands, width, options, mustBeFlat) {
             case MODE_FLAT:
               if (!doc.hard) {
                 if (!doc.soft) {
+                  out.push(" ");
+
                   width -= 1;
                 }
 
@@ -243,6 +279,10 @@ function printDocToString(doc, options) {
           break;
         case "align":
           cmds.push([makeAlign(ind, doc.n, options), mode, doc.contents]);
+
+          break;
+        case "trim":
+          pos -= trim(out);
 
           break;
         case "group":
@@ -471,24 +511,7 @@ function printDocToString(doc, options) {
                   pos = 0;
                 }
               } else {
-                if (out.length > 0) {
-                  // Trim whitespace at the end of line
-                  while (
-                    out.length > 0 &&
-                    typeof out[out.length - 1] === "string" &&
-                    out[out.length - 1].match(/^[ \t]*$/)
-                  ) {
-                    out.pop();
-                  }
-
-                  if (out.length && typeof out[out.length - 1] === "string") {
-                    out[out.length - 1] = out[out.length - 1].replace(
-                      /[ \t]*$/,
-                      ""
-                    );
-                  }
-                }
-
+                pos -= trim(out);
                 out.push(newLine + ind.value);
                 pos = ind.length;
               }
--- a/tests_integration/__tests__/doc-mark-as-root.js
+++ b/tests_integration/__tests__/doc-mark-as-root.js
@@ -1,0 +1,38 @@
+"use strict";
+
+const prettier = require("prettier/local");
+const docPrinter = prettier.doc.printer;
+const docBuilders = prettier.doc.builders;
+
+const printDocToString = docPrinter.printDocToString;
+const concat = docBuilders.concat;
+const hardline = docBuilders.hardline;
+const literalline = docBuilders.literalline;
+const trim = docBuilders.trim;
+const indent = docBuilders.indent;
+const markAsRoot = docBuilders.markAsRoot;
+
+describe("markAsRoot", () => {
+  test.each([
+    [
+      "with hardline will insert a newline with current indentation",
+      concat([indent(markAsRoot(indent(hardline))), "123"]),
+      "\n    123"
+    ],
+    [
+      "with literalline will insert a newline with root indentation",
+      concat([indent(markAsRoot(indent(literalline))), "123"]),
+      "\n  123"
+    ],
+    [
+      "followed by trim will trims up to the the first column, ignoring indented root",
+      concat([indent(markAsRoot(indent(literalline))), trim, "123"]),
+      "\n123"
+    ]
+  ])("%s", (_, doc, expected) => {
+    const result = printDocToString(doc, { printWidth: 80, tabWidth: 2 });
+
+    expect(result).toBeDefined();
+    expect(result.formatted).toEqual(expected);
+  });
+});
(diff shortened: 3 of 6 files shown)


src/doc/doc-printer.js
```diff
@@ -146,6 +166,10 @@ function fits(next, restCommands, width, options, mustBeFlat) {
         case "align":
           cmds.push([makeAlign(ind, doc.n, options), mode, doc.contents]);
 
+          break;
+        case "trim":
+          trim(out);
```
warrenseine: Hum, the bot caught a real bug. `out` is undefined here.

src/doc/doc-printer.js
```diff
@@ -244,6 +266,10 @@ function printDocToString(doc, options) {
...
```
ikatyang: I believe we'll have to update the `pos` (in `printDocToString`) and the `width` (in `fits`) so that the line length can be calculated correctly, we didn't need to do so previously since it's at the end of the line but it's not the case now.

And we should add some tests for it, though it seems we didn't have tests for it from the beginning.

I'm not too familiar with the algorithm here so I might miss something.

warrenseine: I've rebased the commit and fixed the trivial error.

I've trying to find a way to test this new `trim` command. All the tests appear to be integration tests (run Prettier on a file and compare snapshot), not unit tests (typical assertion-based unit tests), which means I can't really test the command because it's not used by any of the official formatters (only prettier-csharp). Any idea?

ikatyang: You could add a test file for `doc` in [`tests_integration/__tests__`](https://github.com/prettier/prettier/tree/master/tests_integration/__tests__) (see [`tests_integration/__tests__/load-toml.js`](https://github.com/prettier/prettier/blob/master/tests_integration/__tests__/load-toml.js) for example), though it's not ideal but we could polish our tests later in other PRs.

tests_integration/__tests__/trim.js
```diff
@@ -0,0 +1,62 @@
+"use strict";
```
ikatyang: Could you rename this file to `doc-trim.js`? so that it'd be clearer for what's this file used for.

tests_integration/__tests__/trim.js
```diff
@@ -0,0 +1,62 @@
...
+const line = docBuilders.line;
+const trim = docBuilders.trim;
+const group = docBuilders.group;
+const indent = docBuilders.indent;
+
+// These tests don't use `runPrettier` because `trim` is not used by any
+// bundled parser (only third-party plugins).
+
+describe("trim", () => {
+  test("trims the current line", () => {
```
ikatyang: [`test.each()`](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout) with `.toEqual()` should be better as the output won't be changed.

src/doc/doc-printer.js
```diff
@@ -109,9 +109,36 @@ function generateInd(ind, newPart, options) {
...
+    out[out.length - 1] = trimmed;
+  }
+
+  return trimCount;
+}
+
 function fits(next, restCommands, width, options, mustBeFlat) {
   let restIdx = restCommands.length;
   const cmds = [next];
+  const out = [];
```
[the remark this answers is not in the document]
warrenseine: Yes. Also, since we’re allocating and feeding this array linearly, it might impact performances. How can I ensure this is insignificant?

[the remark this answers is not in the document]
ikatyang: There's a performance section in the [CONTRIBUTING.md](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md#performance), I guess we could test it with the big file mentioned in https://github.com/prettier/prettier/issues/4776#issue-337167991.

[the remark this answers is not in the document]
warrenseine: Ok, I ran both `--debug-benchmark` and @vjeux's benchmark file and didn't notice any difference.

tests_integration/__tests__/doc-trim.js
```diff
@@ -0,0 +1,108 @@
+"use strict";
+
+const docPrinter = require("../../src/doc/doc-printer");
+const docBuilders = require("../../src/doc/doc-builders");
```
ikatyang:
```js
const prettier = require("prettier/local");
const docPrinter = prettier.doc.printer;
const docBuilders = prettier.doc.builders;
```

tests_integration/__tests__/doc-trim.js
```diff
@@ -0,0 +1,108 @@
...
+      "hello world!"
+    ]
+  ])("trims correctly", (doc, expected) => {
+    const result = printDocToString(doc, { printWidth: 12, tabWidth: 2 });
+
+    expect(result).toBeDefined();
+    expect(result.formatted).toEqual(expected);
+  });
+
+  // test("trims the current line", () => {
```
ikatyang: It seems you forgot to remove them.

tests_integration/__tests__/doc-trim.js
```diff
@@ -0,0 +1,108 @@
...
+#if DEBUG
+  alert(42);
+#endif
+}`
+    ],
+    [
+      // ignores trimmed characters when fitting the line
+      group(concat(["hello  ", "  ", trim, line, "world!"])),
+      "hello world!"
+    ]
```
ikatyang: Could you add a test for trimming to the root (markAsRoot)?

ikatyang:
And also you need to resolve merge conflicts.

tests_integration/__tests__/doc-trim.js
warrenseine: My bad, sorry.

tests_integration/__tests__/doc-trim.js
warrenseine: As is, the code will not respect `markAsRoot`, and will trim down to column 0. I confirmed this behavior with a test. I can commit that test, but I suppose your comment was about making the code respect `markAsRoot`.

Do you think it's worth dealing with this edge case? In which situation would I want to keep an indent of 4 spaces, but still want to trim the remaining spaces after?

ikatyang: You should update the docs then, and also add a test to ensure this behavior. I guess we probably don't need to deal with it since there's probably no use case for it.

https://github.com/prettier/prettier/blob/33bceb0cfb3ea26180f132383bee574577e33437/commands.md#L240-L246

tests_integration/__tests__/doc-trim.js
```diff
@@ -0,0 +1,94 @@
...
+}`
+    ],
+    [
+      // trims up to the the first column, even if there's indented root.
+      group(
+        concat([
+          "{",
+          indent(
+            concat([
+              markAsRoot(line),
```
ikatyang: It seems this `markAsRoot` has no effect, it won't affect anything.

I'd expect something like:

```js
[
  // hardline will insert a newline with current indentation
  concat([indent(markAsRoot(indent(hardline))), "123"]),
  "\n    123"
],
[
  // literalline will insert a newline with root indentation
  concat([indent(markAsRoot(indent(literalline))), "123"]),
  "\n  123"
],
[
  // trims up to the the first column, even if there's indented root.
  concat([indent(markAsRoot(indent(literalline))), trim, "123"]),
  "\n123"
],
```

tests_integration/__tests__/doc-trim.js
```diff
@@ -0,0 +1,94 @@
...
+    42
+  }
+}`
+    ],
+    [
+      // ignores trimmed characters when fitting the line
+      group(concat(["hello  ", "  ", trim, line, "world!"])),
+      "hello world!"
+    ]
+  ])("trims correctly", (doc, expected) => {
```
ikatyang: We could use [the `printf`-like parameters provided by jest](https://jestjs.io/docs/en/api#1-testeachtable-name-fn-timeout) here to improve our test names:

```js
test.each([
  [
    "test name", // change those comments into test names
    doc,
    expected,
  ]
])("%s", (_, doc, expected) => {
//  ^^    ^
})
```

```
 PASS  tests_integration/__tests__/doc-trim.js
    ✓ trims the current line
    ✓ trims existing indentation
```

instead of 

```
 PASS  tests_integration/__tests__/doc-trim.js
    ✓ trims correctly
    ✓ trims correctly
```

commands.md
```diff
@@ -237,6 +237,14 @@ declare function dedentToRoot(doc: Doc): Doc;
 
 This will dedent the current indentation to the root marked by `markAsRoot`.
 
+### trim
+
+```ts
+declare var trim: Doc;
+```
+
+This will trim any whitespace or tab character on the current line, without inserting a new line. This is used for preprocessor directives.
```
ikatyang: In this context, _"without inserting a new line"_ seems unnecessary.

tests_integration/__tests__/doc-trim.js
warrenseine: Ok, will do. Thanks for your time :)

warrenseine: These tests won’t pass as is. Did you mean that you expected them to pass and that `markAsRoot` seems broken?

Anyway, even if they passed, they should probably be part of another test file, say `doc-mark-as-root.js`. But in that case, I’m not sure what I should do. Maybe tests for `markAsRoot` should be added in a separate PR, where we clarify what is expected?

warrenseine: I pushed the tests above along with my changes to see if it breaks anything on a different environment. I've pushed these changes from a Windows machine, so that could be a line ending problem.

tests_integration/__tests__/doc-mark-as-root.js
```diff
@@ -0,0 +1,64 @@
...
+      concat([indent(markAsRoot(indent(literalline))), "123"]),
+      "\n  123"
+    ],
+    [
+      "followed by trim will trims up to the the first column, ignoring indented root",
+      concat([indent(markAsRoot(indent(literalline))), trim, "123"]),
+      "\n123"
+    ]
+  ])("%s", (_, doc, expected) => {
+    const result = printDocToString(doc, { printWidth: 80 });
```
ikatyang: There's no default value for `tabWidth` in `printDocToString`, adding `tabWidth: 2` should get it passed.

warrenseine: That's it. It passes on my machine.

tests_integration/__tests__/doc-mark-as-root.js
```diff
@@ -0,0 +1,64 @@
+// "use strict";
+
+// const prettier = require("prettier/local");
```
ikatyang: 😉

ikatyang:
Thanks!

Issue #4776, opened 2018-06-30
Performance on big files


vjeux:
I just found out that the performance metrics of prettier inside of Nuclide at Facebook:

- p50: 122ms
- p75: 203ms
- p90: 377ms

It's now awesome but totally reasonable. However, for big files like [this one (3k lines)](https://github.com/facebook/nuclide/blob/master/pkg/nuclide-vscode-language-service-rpc/lib/LspLanguageService.js) it consistently takes multiple seconds.

Now that prettier is in a good shape (not changing as often) and massively successful, it would be nice to start spending time on performance improvements. As far as I know, outside of fixing exponential complexity, we haven't spent any time trying to optimize it.

This PR is a call for help, I'm sure that there's a ton of optimizations that can be done to make prettier faster, now would be an awesome time to start working on them.

Here's the challenge, what can we do to make the following file take less time to print?
- https://github.com/facebook/nuclide/blob/master/pkg/nuclide-vscode-language-service-rpc/lib/LspLanguageService.js

Profiling what is taking time would be an awesome first step.

sompylasar:
One more thing I tried, maybe it's already been tried and is already optimized, please tell me if it is.

`fits` and `printDocToString` uses a 3-item array to store commands in the queue. This does not take advantage of hidden classes to reuse object memory layouts.

I'm a little confused which variant should theoretically be best-performing in this case: with arrays or with classes.

> objects created by the same constructor and have the same set of properties assigned in the same order
> [...]
> use literal initializer for Arrays with mixed values
>
> – http://thlorenz.com/talks/demystifying-v8/talk.pdf

### Proposed solution

Refactor to use a constructor function that initializes the properties in a predefined order. Use `new` to create the objects.
```js
function Cmd(ind, mode, doc) {
  this.ind = ind;
  this.mode = mode;
  this.doc = doc;
}
```
```js
cmds.push(new Cmd(makeIndent(ind, options), mode, doc.contents));
```
```js
const cmds = [new Cmd(rootIndent(), MODE_BREAK, doc)];
```

sompylasar:
### Problem

`printAstToDoc` uses `Map` for its cache keyed by nodes. 
On the test file, the stats are that it's 97.7% more misses than hits on this cache:
```js
{printAstToDocCacheHitCount: 206, printAstToDocCacheMissCount: 9083}
```

j-f1:
This is awesome @sompylasar! 😎

A few comments:

[...]

PR #4779, opened 2018-06-30
performance: cache NODE_ENV production check

JoelEinbinder:
The `concat` function is very hot, and it does the same string comparison every time. Moving the production check into its own variable gave a 2-3% speedup in a production run. I tried switching between two different concat functions, but that didn't give any meaningful speedup and comes at the cost of less readable code.

#4776

### DIFF
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const production = process.env.NODE_ENV === "production";
+
 function assertDoc(val) {
   /* istanbul ignore if */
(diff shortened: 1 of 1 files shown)


ikatyang:
I guess we could simply remove those `IfStatement`s from the production build using some custom babel plugins.

https://github.com/prettier/prettier/blob/50d242d0853d8294e3770bb1b502cdfbee65c03b/scripts/build/config.js#L16

duailibe:
They were being stripped before.

I have to check why it regressed

PR #4782, opened 2018-06-30
Upgrade jest to 22

SimenB:
New attempt at #4539. There's still some issues.

- Jest uses object shorthand (https://github.com/facebook/jest/pull/6582)
- `realpath-native` doesn't work on node 4 (https://github.com/SimenB/realpath-native/pull/8)
- Prettier code base uses destructuring (and have disabled babel transform in Jest): https://github.com/prettier/prettier/blob/50d242d0853d8294e3770bb1b502cdfbee65c03b/src/main/core.js#L12-L15 

So status (with fixes for the dependency issue applied to local `node_modules`) is:

image


duailibe:
@SimenB thanks for doing this! And congrats on those error messages, looking good!

We try to only run tests on the prod bundle in Node 4, but there are some tests that still require the source files. I'll look into this and rebase this PR after that

duailibe:
@SimenB sorry I got confused. Actually we don't run the tests on Node 4 using the source code. We first run the build script, then run the tests using the built artifacts.

That error shouldn't happen on Prettier.. in fact, the logs in CircleCI show different errors https://circleci.com/gh/prettier/prettier/2546

SimenB:
Hmm, in that case, we should be good to use Jest 22 in the mean time. If it passes, I can come back to you with a Jest 23 PR once those two linked issues are ironed out

package.json
```diff
@@ -37,7 +37,7 @@
     "graphql": "0.13.2",
     "html-tag-names": "1.1.2",
     "ignore": "3.3.7",
-    "jest-docblock": "22.2.2",
+    "jest-docblock": "23.2.0",
```
SimenB: not really related (and should be the same code)m but why not 🙂

SimenB:
@duailibe Node 4 passes (🎉) but "standalone" build fails (😭). What's that?

SimenB:
Ah, concating the snapshot name with test filename breaks your hack: https://github.com/prettier/prettier/blob/99c4e41fa1bf71a99a34a734c9e4f7ce2e3ad900/tests_config/run_spec.js#L20-L25

Lemme remove the filename from the snapshot (should reduce the diff as well). Why do you have it in the `toMatchSnapshot` call at all?

SimenB:
Bah, no, that still won't be enough. Do you think you can handle it? I think I'd have to dig a bit more into your test setup to fix it, while I think you'd be able to just quickly fix it

SimenB:
Incentive for getting 23 to work: https://github.com/jest-community/jest-watch-typeahead 😀

duailibe:
@SimenB the standalone tests run the "standalone" build (no outside dependencies, meant to run in an environment like a browser) but not all parsers are supported yet so we need to skip some tests.

Unfortunately, simply not running the snapshots tests would make Jest mark them as obsolete so I needed to have a skipped test with the exact name of the one that saved the snapshot.

I've just pushed a fix that worked locally for me, let's see if it passes CircleCi

duailibe:
> Incentive for getting 23 to work: https://github.com/jest-community/jest-watch-typeahead 😀

That looks awesome... I might stop writing random `aaaaaaaaaa.js` tests when working on Prettier 😂

SimenB:
Oh yeah, green :D

duailibe:
@SimenB thank you so much for pushing it forward :heart:

SimenB:
BTW, for skipping some tests in certain conditions, we use a `.only` (or `fit`) in Jest. Maybe applicable here? See https://github.com/facebook/jest/blob/e8a8f624d8cd6afc1475ddeb9ba9eef56daab90c/scripts/ConditionalTest.js

---

I'll try to remember to follow up with a Jest 23 version when we're ready 🙂

PR #4783, opened 2018-06-30
Do not require space after markdown block language name

kachkaev:
https://github.com/prettier/prettier/pull/4153 made it possible to detect fenced code block language when it is followed by arguments (e.g. ` ```js {something=something} `). This PR makes it also possible to detect language in cases cases like ` ```js{something=something} ` (no space).

The reason for this change is that Atom highlights code blocks regardless of a space after the language name, which makes the editor wonder why the correctly detected code block is not being formatted:

screen shot 2018-06-28 at 22 08 13

PR background: https://github.com/prettier/prettier/pull/4153#issuecomment-401174999

### DIFF
--- a/src/language-markdown/embed.js
+++ b/src/language-markdown/embed.js
@@ -12,7 +12,7 @@ function embed(path, print, textToDoc, options) {
 
   if (node.type === "code" && node.lang !== null) {
     // only look for the first string so as to support [markdown-preview-enhanced](https://shd101wyy.github.io/markdown-preview-enhanced/#/code-chunk)
-    const lang = node.lang.split(/\s/, 1)[0];
+    const lang = (node.lang.match(/^[A-Za-z0-9_-]+/) || [])[0];
     const parser = getParserName(lang);
     if (parser) {
       const styleUnit = options.__inJsTemplate ? "~" : "`";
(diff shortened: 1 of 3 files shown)


src/language-markdown/embed.js
```diff
@@ -12,7 +12,7 @@ function embed(path, print, textToDoc, options) {
 
   if (node.type === "code" && node.lang !== null) {
     // only look for the first string so as to support [markdown-preview-enhanced](https://shd101wyy.github.io/markdown-preview-enhanced/#/code-chunk)
-    const lang = node.lang.split(/\s/, 1)[0];
+    const lang = (node.lang.match(/^[A-Za-z0-9_-]+/) || [])[0];
```
ikatyang: `lang` should be a string, we could use `.match(/^[A-Za-z0-9_-]*/)` or `.split(/[^A-Za-z0-9_-]/, 1)` to avoid `undefined` in this case.

kachkaev: What about?

```js
const lang = (node.lang.match(/^[A-Za-z0-9_-]+/) || [])[0] || "";
```

or

```js
const langMatch = node.lang.match(/^[A-Za-z0-9_-]+/);
const lang = langMatch ? langMatch[0] : "";
```

Nevertheless, it's interesting that none of the tests fail despite that lang is undefined 🤔

ikatyang: Isn't the first one much more complex? 😂 The second one looks good to me.

PR #4785, opened 2018-06-30
feat(playground): add description for options

ikatyang:
image

### DIFF
--- a/website/playground/sidebar/options.js
+++ b/website/playground/sidebar/options.js
@@ -9,6 +9,7 @@ export function BooleanOption({ option, value, onChange }) {
   return (
     <Checkbox
       label={option.cliName}
+      title={option.description && option.description.replace(/\n/g, " ")}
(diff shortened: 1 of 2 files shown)


j-f1:
Looks like there’s a problem with Netlify.

lydell:
Nice!

The descriptions for `--no-bracket-spacing` and `--no-semi` seem to be backwards, though.

`--no-bracket-spacing`: Print spaces between brackets.
`--no-semi`: Print semicolons.

website/playground/sidebar/options.js
```diff
@@ -51,3 +54,10 @@ export default function(props) {
       throw new Error("unsupported type");
   }
 }
+
+function getDescription(option) {
+  const description = !option.inverted
```
j-f1: Can you invert this ternary so we don’t have a negated condition? (i.e. `inverted ? oppositeDescription : description`)

PR #4788, opened 2018-07-02
Shortcut getStringWidth for ASCII-only strings; add options to measure performance

sompylasar:
https://github.com/prettier/prettier/issues/4776#issuecomment-401639921

Refs #4776


vjeux:
Could you split this into two PRs? Trying to keep things standalone so it’s easier to bisect, revert...

PR #4792, opened 2018-07-02
Fix benchmark require

duailibe:
`benchmark` would either get bundled in the production build or unable to be required, this PR prevents both from happening.

cc @sompylasar

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -200,6 +200,7 @@ function format(context, input, opt) {
     return { formatted: pp, filepath: opt.filepath || "(stdin)\n" };
   }
 
+  /* istanbul ignore if */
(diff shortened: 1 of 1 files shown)


src/cli/util.js
```diff
@@ -201,10 +201,19 @@ function format(context, input, opt) {
   }
 
   if (context.argv["debug-benchmark"]) {
+    let benchmark;
+    try {
+      benchmark = eval("require")("benchmark");
+    } catch (err) {
+      context.logger.debug(
+        "'--debug-benchmark' requires the 'benchmark' to be installed."
```
j-f1: > requires the 'benchmark' **module** to be installed

?

vjeux:
Sorry I merged it quickly, I got so excited :)

PR #4797, opened 2018-07-02
[RFC] Use Flow for typechecking

duailibe:
I made this for learning purposes, to use Flow in a real project and learn about it.

I decided to open PR because why not.. if you think this is wanted, I can write the types for the rest of the code.


src/doc/doc-debug.js
```diff
@@ -32,18 +40,20 @@ function flattenDoc(doc) {
...
 
-function printDoc(doc) {
+function printDoc(doc /*: Doc */) {
   if (typeof doc === "string") {
     return JSON.stringify(doc);
   }
 
   if (doc.type === "line") {
-    if (doc.literalline) {
+    if (doc.literal) {
```
duailibe: This helped catch a bug

suchipi:
Haven't looked at the code much yet, but a couple thoughts:
* If we're gonna go all-in, we should probably use a babel plugin to remove the types, instead of using comment syntax.
* I'm a huge fan of flow, but it raises the bar for contribution significantly. Even though its error messages have gotten a lot better recently, some aspects of its behavior are unintuitive and could frustrate first-time contributors. I introduced flow at work about a year ago and my coworker today told me she just guesses stuff until the errors go away, which is indicative of the experience others will have, I think.

suchipi:
I'm curious what other maintainers think, I'm open to it

ikatyang:
FYI, TypeScript also supports comment syntax ([JSDoc](https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript)), which I already [wrote some here](https://github.com/prettier/prettier/blob/master/src/main/core-options.js#L13-L51) for intellisense.

If we're going to use some type system, I think it'd be better to have declaration files (e.g., `.js.flow`, `.d.ts`) for API consumers, but as far as I can tell:
  - Flow does not support generating a readable `.js.flow` file, the output from `flow gen-flow-files` is not readable. (I think there should be a way to generate a readable one but I couldn't find it.)
    - need to write it manually, which is easy to be out of sync.
  - TypeScript does not support generating declaration files from a JSDoc-typed file, only support TS files.
    - need to use TS file, which will lose the ability to install from GitHub, or we'll have to setup [`prepack` hook](https://docs.npmjs.com/misc/scripts) to build it:

      > prepack: run BEFORE a tarball is packed (on npm pack, npm publish, and **when installing git dependencies**)
    - [the declaration file is the easiest way to understand the API](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/prettier/index.d.ts).

I personally prefer TypeScript but I won't reject Flow, any type system is better than nothing.

(We currently do not support formatting those comment syntax 😂, e.g., `/*: Array   <  
 Doc  > */` -> `/*: Array<Doc> */`, or `/** @type { Doc []   } */` -> `/** @type {Doc[]} */`.)

j-f1:
We could add a separate entry point for installs from GitHub that requires `ts-node` (or its Flow equivalent) to compile the main bundle on-the-fly.

duailibe:
> FYI, TypeScript also supports comment syntax (JSDoc),

I'm aware, but I'm under the impression that is inferior to actual TypeScript. Flow's comment syntax lets you include *anything* between `/*::  */`

> I think it'd be better to have declaration files (e.g., `.js.flow`, `.d.ts`) for API consumers

I also think that'd be nice to have, but I'm less worried about that because we rarely change the API. I think we can come up with some mechanism to prevent them getting out of sync.

> We currently do not support formatting those comment syntax

We actually format inside Flow comments when used with the Flow parser: [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAyhALZwwAWOUA5gBSoQT4D0AVEvl1wM4wBO1Gt3xsWASgA6Udp3x9BtUS3zBpXdfk2QoffI2bt8nKOhIAjOP2X4AvPgBsAbmkBfEABoQEAA65oPMigAIb8-BAA7gAKoQiBKMEAbhA4ACaeIOb8wWAA1uSEPjlCyALocF7UPFYwUdk0JMHIqMEANtVeAFY8AB4AQtl5BcFkADLUcM1tHSBF-NX8yGitkRk+ijAA6mmUyAAcAAxe6xDVm9k+S+twC4mTXvxwAI7oOI91wQ1NSC3tFSDVEg4Ur8cpeHhCVpwACK6Ag8Cmfy8MGC5m2qV2SAATMjsjhWkIAMKkRpLKDQe4gdDVAAqqPiv2qrlcQA), with some caveats

suchipi:
We might get more quick wins with TypeScript since we already have jsdoc comments in a couple places, but I wouldn't base the decision off that alone

duailibe:
@azz @lydell @vjeux do you have an opinion on this?

lydell:
I like type checking (have worked quite a bit with Flow), but I don't have an opinion on what to choose.

vjeux:
Few reasons I didn’t put flow or typescript:

- The main thing that prettier does is to take an ast and turn it into a “doc”. There are no available type definitions for all the asts so this drastically limit the usefulness of type checking. And I was not really interested in being in the business of building one.

- It’s —really— nice to not have a build step. You can just do ./bin/prettier.js <...> and it just works. No webpack to configure...

- It’s one more thing people have to learn trying to work on the project. There’s been a ton of contributions + forks and I believe it’s in part due to the simple process.

- Our snapshot tests + running it on real codebases have been very effective at catching bugs. I don’t believe that we would have caught many bugs with it.

So my tldr is that it isn’t super needed. But I’m not going to stop you from doing it. You are working on the codebase more than I am nowadays so please shape it the way you can be more productive with it

azz:
I agree with @vjeux, and in my experience the amount of bugs that type systems catch is not worth the time it takes to keep the types up to date. I've refactored Prettier fairly significantly without feeling the need for types.

duailibe:
In my experience the type checking is less about catching bugs and more about confidence when refactoring, which I currently don't have, specially in cryptical parts like `src/cli/`.

suchipi:
I think types will bring a lot more value to the metaprogrammed options and cli stuff, rather than the printer.

aboyton:
Obviously the maintainers can do whatever they like.

I found when I tried contributing to Prettier in the early days it was much harder for me to get started without the types on functions as I had little idea what I could do with any values. I find types wonderfully self documenting for new people coming into the code. I know others have different experiences but find it much easier to jump into a new codebase with types than without.

As for type systems, personally I really like TypeScript. I think the community behind TypeScript is a lot bigger and the support across platforms appears to be better than Flow but I've never really used Flow much. The tooling behind TypeScript is also amazing.

In terms of making it harder for people to contribute, using TypeScript does raise the bar a little although given that Prettier is an implementation of an academic paper for pretty printing Haskell I personally don't think that adding TypeScript makes it particularly harder for people to contribute. I think anyone who can understand the algorithm behind Prettier should be able to pick up TypeScript very easily. There's nothing extra that they would need to install to get started, you can specify TypeScript as a dependancy and make the `npm` commands run `ts-node`.

So I think moving to TypeScript might restrict some people from contributing but actually might make it easier for others to.

One side benefit of using TypeScript is that you would get definition files "for free" for users of Prettier which is a nice bonus. Yes I realise that there's one already on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/prettier).

That said, that's just my experience.

duailibe:
I don't think TypeScript is an alternative unless there's a comment based solution. It's not just running `prettier` in the console (for which `ts-node` can be used but it's less optimal), but also calling `require("prettier")` without compiling.

I think we should move forward with this. If anyone could review this

duailibe:
This is ready for review

j-f1:
> I don't think TypeScript is an alternative unless there's a comment based solution

There is — JSDoc is supported.

To use Prettier with TS syntax features, we could dynamically compile the code (as I suggested above).

duailibe:
> There is — JSDoc is supported.

I'm aware, is it equivalent? With Flow, everything is supported. Also, with Flow we can format the comments

> we could dynamically compile the code

I didn't notice a `require` option was available, I thought you had to use `ts-node file.ts`. Either way I still prefer having a comment only solution

lydell:
I’m OK with this PR as-is, but as said I don’t have much of an opinion either way.

hawkrives:
@ikatyang IIRC, the Node types are built-in to Flow itself, e.g. at https://github.com/facebook/flow/blob/master/lib/node.js. Just FWIW.

ikatyang:
If I understand correctly, that means there's no version control for those built-in Node.js libraries and ES features? and Flow users are always considered using the latest version, looks weird to me. 🤔

thorn0:
>> There is — JSDoc is supported.

> I'm aware, is it equivalent?

There might be peculiarities, but overall - **yes**, it is.

Please have a look at this demo setup: https://github.com/thorn0/simple-js-ts-demo (basically, the whole 'setup' lives in one file - `tsconfig.json`).

In `src`, we have pure JavaScript with JSDoc comments that can be run standalone if needed.

In `typings`, we may have `.d.ts` files with additional type definitions where we can use all the sophisticated syntax for types TypeScript has.

We can open the project in something like VS Code or compile it with the `tsc` command to see type errors. Also we can use the compiler to compile JS to JS, e.g. ES2017 to ES5.

j-f1:
Expanding on @thorn0’s setup, you could get rid of the `index.d.ts` file by replacing the JSDoc comment in `index.js` with this:

```js
/**
 * @param {{ num: number }} a
 */
```

thorn0:
@j-f1 Of course, but my point was that it's possible to use the full richness of TypeScript's type system this way: generics, mapped types, conditional types and whatnot. There might exist JSDoc support for some of those, but definitely not for all.

thorn0:
So, the answer to the question '_Is it equivalent_?' is '_Yes, if you use the js + d.ts setup_'. Nevertheless, there is one important inequivalency already [mentioned above](https://github.com/prettier/prettier/pull/4797#issuecomment-402027830). Namely, the `--alowJs` compiler option isn't compatible with `--declaration` for now, that is TS can't generate type definitions out of JS files ([issue in the TS project](https://github.com/Microsoft/TypeScript/issues/7546)), which you guys will obviously want to have.

j-f1:
The PR for that feature is https://github.com/Microsoft/TypeScript/pull/21455.

duailibe:
Well I did this only for learning purposes and after I was done I felt that would be good adding this safety. 

I did try TypeScript though: https://github.com/duailibe/prettier/commit/7cfbe00ae0ff01cfebc61129f4a198775f9df109 I was able to type stuff with reasonable effort.

My impressions:

- `tsc` vs `flow`: error message are equally cryptical
- editor integration: the Atom plugin doesn't work with JS files so I didn't have any of the benefits
- TS has namespaces which is nice, but not a hard requirement
- Flow type inference *feels much* better.. I had to add more `/** @type {Foo} */` comments than I'd like
- JSDoc looks slightly better than inline comments; but no formatting, but we can format Flow comments (I've just started JSDoc formatting.. let's see how that goes)

I have no horse in this race so if anyone to write stuff in JSDoc than go ahead. I'm willing to write the rest of the types in Flow since that's what I'm currently using in other projects and at work, but 🤷‍♂️. 

After all, I just really wish we had types :)

duailibe:
> I'm a little worried about the external definitions

I'm not.. it seemed to me we use pretty simple stuff and wouldn't hurt writing minimal type definitions for what we need

thorn0:
@duailibe Try using the `strict` compiler option. It should improve the type inference.

thorn0:
Also error messages [might get a bit less cryptic](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) if you use the `interface` syntax in cases like this
```ts
type Indentation = {
    value: string;
    length: number;
    queue: Alignment[];
    root?: Indentation;
 };
```
vs
interface Indentation {
    value: string;
    length: number;
    queue: Alignment[];
    root?: Indentation;
 };
```

suchipi:
FWIW, TypeScript is amazing in VS Code, but flow is not as good. Kind of the opposite of Atom

j-f1:
@suchipi I’ve been using TS in Atom, and I’ve found the editor support is quite nice (I’m using `ide-typescript` and `atom-ide-ui`).

suchipi:
Ah, good to know. Do you get JS type inference?

j-f1:
If you hover over a symbol, you’ll see the inferred type (not at my 🖥right now so I can’t provide a screenshot).

duailibe:
> If you hover over a symbol, you’ll see the inferred type (not at my 🖥right now so I can’t provide a screenshot).

Doesn't support JSDoc though.. everything is treated as any 😔

azz:
I've never managed to get flow working properly in VS Code. It always seems to clash with the built-in stuff.

suchipi:
@azz You have to set `javascript.validate.enable` to false and then disable the built-in typescript extension. You lose syntax error detection unless you have eslint, though.

meyer:
The webpack folks use TypeScript with `.js` files and JSDoc comments and [they seem to dig it](https://twitter.com/TheLarkInn/status/1006861097729585154). Might be a good example project to check out.

See also: https://twitter.com/TheLarkInn/status/1011813941981294592

azz:
@suchipi The problem is some of the projects at work that I want to switch to Flow have some TypeScript files that I'd want to gradually move to Flow. So telling everyone to disable the TS extension isn't really practical. 😭

suchipi:
Ah :/

Issue #4801, opened 2018-07-03
formatWithCursor performance bottleneck


tmcw:
I've isolated a bottleneck from our production environment and here's a nifty self-contained benchmark for it: https://gist.github.com/tmcw/1a4e8ee47941454337dc5952dbf90180 (swap require('./') for require('prettier') if that's intended)

What I'm seeing in this and in testing different cases is that `formatWithCursor` has a huge performance penalty over format, and it's entirely rooted in the call to [diff.diffArrays](https://github.com/prettier/prettier/blob/432c9d7d04df2490905c55a277c1f70e65006233/src/main/core.js#L151-L154), which creates may intermediate objects and arrays. The further into the document the cursor is, the more of a problem it becomes. This testcase takes upwards of 14s to format about 300 lines of code.

Reading the code, it's a bit insidery - I suspect that an approach that doesn't require a char array, or does the diff progressively with the checking, or skips the creation of so many intermediate objects, may be ideal.

robwise:
This can cause my entire kernel to go over memory even when just calling from node, much less inside of prettier-atom (https://github.com/prettier/prettier-atom/issues/480). I'm getting a bunch of bug reports over there because of this code and my Atom debugger likewise points to this diff arrays code the OP mentions as causing an OOM error.

Is there any way we can improve the performance of this diff check? There are example HTML code files in that issue that can reproduce the problem.

You can reproduce with this simple harness, assuming you have a file called `test.html` with a huge amount of code to diff.

```js
prettier = require('prettier');
fs = require('fs');

fs.readFile('./test.html', 'utf8', function(err, contents) {
  prettier.formatWithCursor(contents, { cursorOffset: 2, parser: 'html' });
});

console.log('FINISHED');
```

An example of HTML that causes things to blow up is this guy's homepage copy and pasted into that test.html file: https://miloslav.website/

tmcw:
https://github.com/prettier/prettier/pull/6164 is the current outstanding PR. Unfortunately it's quite tricky to produce exactly the same behavior as the current diff-based implementation, and thus none of the PRs are landing.

lydell:
I don’t know at all how formatWithCursor works, and I’ve never used it. So I might not know what I’m talking about, but to me it sounds sensible to merge a PR that makes this feature usable at the expensive of not working 100% the same in all cases (and probably the current implementation isn’t perfect either so it probably matters even lesss).

robwise:
[...]

PR #4802, opened 2018-07-04
Lock closed issues and PRs after 3 months of inactivity

lipis:
Mostly to avoid random notifications from people "me too"ing on stuff that no one has touched in ages.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -7,4 +7,3 @@ lockComment: >
   any recent activity after it was closed. Please open a new issue for
   related bugs.
 setLockReason: true
-only: issues
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,4 +1,5 @@
 # Configuration for lock-threads - https://github.com/dessant/lock-threads
 daysUntilLock: 90
 lockLabel: locked-due-to-inactivity
+lockComment: false
 setLockReason: true
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,8 @@
+# Configuration for lock-threads - https://github.com/dessant/lock-threads
+daysUntilLock: 90
+exemptLabels: []
+lockLabel: false
+lockComment: >
+  This thread has been automatically locked since there has not been
+  any recent activity after it was closed. Please open a new issue for
+  related bugs.
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,5 +1,4 @@
 # Configuration for lock-threads - https://github.com/dessant/lock-threads
 daysUntilLock: 90
-exemptLabels: []
 lockLabel: locked-due-to-inactivity
 setLockReason: true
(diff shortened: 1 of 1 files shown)


j-f1:
Looks cool!

Let’s set `setLockReason: true` so it gets locked as “resolved” rather than without a reason, and limit it to issues with `only: issues`.

lipis:
Updated

duailibe:
I don’t think we should limit to issues. People are commenting on closed PRs when they should open new issues

lipis:
@duailibe makes sense.. WDYT @vjeux?

Feel free to merge it..

vjeux:
Hopefully it doesn’t spam me with locked emails (which may be worst that the occasional reply). But I’m in favor, go for it

duailibe:
I didn't check the file before, only the comments here.

I think we should use `lockComment: false` or we'll be spammed with bot comments

ikatyang:
I guess we could create a new label with description from `lockComment` and use `lockLabel` instead.

lipis:
Right I didn't think of the bot comment at all.. the descriptive lock label would be better in that case.. as we might get 4K emails.. :)

Some ideas for the label name? It could be long like: `locked-due-inactivity-since-it-was-closed`

j-f1:
`locked-due-to-age`? We could put a more detailed description in the label’s hover text (although we’re limited to 100 characters).

duailibe:
Won't this spam the Github feed as well? Like:

image

I'd start with just locking issues and moving on and see if people are getting confused by not being able to comment

ikatyang:
I tested it on my repo and it seems only `help wanted` and `good first issue` will appear in the feed:

image

image

j-f1:
Does me adding `type:perf` to #4801 show up on your timelines?

ikatyang:
No.

duailibe:
That's cool!

I should've noticed before though.. we label issues with `lang:*` everyday 🙈

lipis:
I'm out of the office these days.. so feel free to edit and merge with whatever you decide.

lipis:
Per instructions I set the `lockComment: false` to avoid any surprises :)

lipis:
Good luck!

lipis:
I created that Label.. https://github.com/prettier/prettier/labels and ready to press the button! Any last words?

j-f1:
“What could possibly go wrong?”

j-f1:
Description updated.
New:
> Please open a new issue and fill out the template instead of commenting.

Old:

> Issues that haven't been active after they were closed

lipis:
May your gods be with ya!

lipis:
https://github.com/prettier/prettier/issues?q=label%3Alocked-due-to-inactivity+is%3Aclosed

it works.. :)

suchipi:
Thanks everyone!!

PR #4803, opened 2018-07-04
Update precommit docs to be compatible with latest husky version

cshg:
The current docs are not compatible with the latest husky API, which changed with the version bump from 0.1.x to 1.x.x.
This updates the docs to provide valid examples with husky version 1.x.x.

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -19,8 +19,10 @@ and add this config to your `package.json`:
 
 ```json
 {
-  "scripts": {
-    "precommit": "lint-staged"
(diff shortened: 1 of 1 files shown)


typicode:
Hey thanks for the help 👍 

Yes, a few things, the latest version is still behind `husky@next` as things are getting posished for the 1.x.x release.

So if you just install with `yarn add husky --dev`, instead of `yarn add husky@next`, it will install 0.1.x and the new 1.x.x config in the examples won't work.

PR #4809, opened 2018-07-04
Upgrade to Jest 23

SimenB:
I have some small hope that this will pass 🙂

Includes typeahead.

image


tests_config/run_spec.js
```diff
@@ -72,22 +75,19 @@ function run_spec(dirname, parsers, options) {
...
-            prettyprint(input, path, compareOptions),
-            compareOptions
-          );
-        } catch (e) {
-          pperr = e.stack;
-        }
-
-        test(path + " parse", () => {
-          expect(pperr).toBe(null);
+        test(`${path} parse`, () => {
```
SimenB: Seemed better to stick it inside the test, so that if it throws it's in the context of a test

SimenB:
@duailibe it passed! 🙂 

@azz's inline snapshots is in this version too

duailibe:
@SimenB as always, thank you so much!! 💯 🎉

PR #4812, opened 2018-07-05
Fix example of husky pre-commit hook to include next version tag

igeligel:
Tried to install prettier with current installation description here: https://prettier.io/docs/en/precommit.html

Unfortunately `husky` was not updated on npm yet, so you need to install the pre-release version which is more or less fine I guess. Still should be added to the instructions there.


suchipi:
Thanks!

Issue #1722, opened 2017-05-25
1.4 Release


vjeux:
Starting this issue to track what needs to be done for the 1.4 release:

TypeScript support:
- [ ] Add TypeScript to the help
- [ ] Bundle the dependencies https://github.com/prettier/prettier/issues/1640

TypeScript formatting blockers:
- [ ] #1460 Comments inside of decoratos
- [ ] #1694 Comments inside of function types
- [x] #1769 generator functions
- [ ] eslint/typescript-eslint-parser#288
- [x] eslint/typescript-eslint-parser#289
- [x] eslint/typescript-eslint-parser#290
- [x] eslint/typescript-eslint-parser#297

Facebook blockers:
- [x] #1520 Do not break on extend
- [x] #774 !() indentation
- [x] Inline `import from` for a single one like we do for `require`
- [x] Update to flow 0.47.0
- [x] #1724 Extra space in flow type

Things that are easy we should knock down:
 - [x] #1607 Make multiple variable declarations on their own line

It's going to be pretty awesome :)

vjeux:
https://github.com/prettier/prettier/releases/tag/1.4.0

azz:
@vjeux Should we mention in the Readme/release that TypeScript 2.3 is required? (this is a requirement of typescript-eslint-parser)

vjeux:
I'm now bundling everything so prettier is now dependency-free. https://unpkg.com/prettier@1.4.0/src/

JamesHenry:
TypeScript is a peerDependency of the parser, so I think @azz's point still stands, @vjeux. The first step in the process is that the code is parsed using the TypeScript compiler, so it does have to be there and the version is important in various cases.

TypeScript 2.3 is the supported "major" version (based on how TS handles version numbers), `typescript-eslint-parser` will log a warning to the console if another version is being used.

vjeux:
@JamesHenry The npm version of prettier has 0 dependencies. I'm bundling all the dependencies in a single `parser-typescript.js` file that includes typescript 2.3, typescript-eslint-parser, all their dependencies... This way, you can just drop in prettier and "it just works".

JamesHenry:
Ah, I see, TypeScript is also included in that - that's awesome 😄 

Is the TS version that's bundled dictated by the devDependency of this project?

vjeux:
Yes, https://github.com/prettier/prettier/blob/master/package.json#L44

JamesHenry:
[...]

PR #4836, opened 2018-07-12
Blacklist this in functional composition heuristics

princed:
As filtering out `this` and `super` cases in #4758 didn't seem to get any objections, here's that part as a separate PR:

Do not react when `MemberExpression` object type is `ThisExpression` or `Super`
Filters out cases like 
```js
this.compose(a(), b);
super.compose(a(), b);
```

cc @suchipi


suchipi:
Thanks @princed!

lydell:
@princed #1722

princed:
@lydell Thanks! Looks like it's rather #4856

PR #4842, opened 2018-07-14
Use destructuring for readability and optimzations at 'for' loop length property…

prashant-andani:
Add destructuring for readability and optimzations at 'for' loop length property…

### DIFF
--- a/src/language-vue/parser-vue.js
+++ b/src/language-vue/parser-vue.js
@@ -147,7 +147,9 @@ function parseHTML(html, options) {
         }
       }
 
-      let text, rest, next;
+      let text;
+      let rest;
+      let next;
       
       if (textEnd >= 0) {
         rest = html.slice(textEnd);
--- a/src/language-vue/parser-vue.js
+++ b/src/language-vue/parser-vue.js
@@ -150,7 +150,7 @@ function parseHTML(html, options) {
       let text;
       let rest;
       let next;
-      
+
       if (textEnd >= 0) {
         rest = html.slice(textEnd);
         while (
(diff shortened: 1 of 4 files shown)


src/language-yaml/utils.js
```diff
@@ -81,8 +82,9 @@ function isLastDescendantNode(path) {
     case "nonSpecificTag":
       return false;
   }
+  const pathStackLength = path.stack.length - 1;
```
ikatyang: This one looks not equivalent to the original one.

src/language-vue/parser-vue.js
```diff
@@ -146,9 +147,8 @@ function parseHTML(html, options) {
         }
       }
 
-      let text;
-      let rest;
-      let next;
+      let text, rest, next;
```
ikatyang:
```
✘  one-var            Split 'let' declarations into multiple statements                                                                                               

src/language-vue/parser-vue.js:150:7
      let text, rest, next;
       ^
```

prashant-andani:
@ikatyang Thanks for catching it... I have fixed it.

ikatyang:
Thanks!

duailibe:
Did you measure the perf improvements of memoizing `list.length`? I don't think this is actually an improvement in perf but it's less readable IMO

ikatyang:
Agreed that's less readable, the balance between readability and performance is hard, not sure how to deal with it.

https://jsperf.com/for-vs-foreach

image

duailibe:
Funny I get different results.

image

Nevertheless, we should measure this. I've been thinking about adding benchmark tests to measure perf improvements/regressions but haven't had the time lately.

PR #4789, opened 2018-07-02
feat(benchmark): add options to measure performance (#4776)

sompylasar:
- `--debug-benchmark` uses `benchmark` module to produce statistically significant time measurements.
- `--debug-repeat` uses a naive loop and measures just the average run time, but useful for profiling to highlight hot functions.

Refs https://github.com/prettier/prettier/issues/4776

PR #4846, opened 2018-07-15
docs(perf): add performance section to CONTRIBUTING.md

sompylasar:
As requested here: https://github.com/prettier/prettier/pull/4789#issuecomment-401861558


CONTRIBUTING.md
```diff
@@ -22,3 +22,28 @@ Run `yarn lint --fix` to automatically format files.
...
+
+If you're contributing a performance improvement, the following Prettier CLI options can help:
+
+- `--debug-benchmark` uses [`benchmark`](https://npm.im/benchmark) module to produce statistically significant time measurements. The measurements are printed at the debug log level, use `--loglevel debug` to see them.
+- `--debug-repeat N` uses a naive loop to repeat the formatting `N` times and measures the average run time. It can be useful to highlight hot functions in the profiler. The measurements are printed at the debug log level, use `--loglevel debug` to see them.
+
+To run Prettier with the profiler, use the Node CLI `--inspect-brk` option and Chromium/Chrome browser's [`chrome://inspect`](chrome://inspect) page to activate the Chrome DevTools Profiler for Node:
+
+```
+$ cat FileToFormat.js | NODE_ENV=production node --inspect-brk ./bin/prettier.js --stdin-filepath FileToFormat.js --debug-repeat 1000 --loglevel debug > /dev/null
```
sompylasar: We might want to recommend building it and using `./dist/bin-prettier.js` instead of `./bin/prettier.js` to get more accurate profiles (e.g. because of these `process.env.NODE_ENV` accesses that slow down `concat`).

j-f1: Do you think it would make sense to have an npm script for this? Here’s my start:

```sh
cat $1 | NODE_ENV=production node --inspect-brk ./bin/prettier.js --stdin-filepath $1 --debug-repeat 1000 --loglevel debug > /dev/null
```

you could then do `yarn foo FileToFormat.js` and it would run the tests.

sompylasar: Yes, I can add an npm script, but the user interaction will still be required to open the inspector and start profiling.

sompylasar: Also, the 1000 probably deserves to be a CLI argument, and this opens a whole new can of worms because now there are two CLI arguments to parse.

j-f1: IMO even if user interaction is required, it’s still easier to run a script than to copy-and-paste that same long command each time. For the `--debug-repeat` value, you can do `${2:-1000}`.

sompylasar: Agree.

sompylasar:
Updated, @j-f1 please have a look.

sompylasar:
Is something preventing this from being merged?

j-f1:
Just waiting on another 👍.

Thanks for the contribution!

Issue #4856, opened 2018-07-17
1.14 Release


ikatyang:
It has been a long time since last release, it's time to release a new version.

I'll send a blog post draft later. --> #4857

ikatyang:
We should run 1.14 on some big codebases to see if there's any regression.

j-f1:
GitHub Desktop: https://github.com/desktop/desktop/pull/5182

j-f1:
1.14 seems to be [broken on VSTS](https://github.visualstudio.com/Desktop/Desktop%20Core%20Team/_build/results?buildId=2721&view=logs):

<details>

```diff
diff --git a/.github/config.yml b/.github/config.yml
index e8583ed8b..b8cbaa5e6 100644
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,20 +1,23 @@
-# Configuration for request-info - https://t-info
-
-# *Required* Comment to reply with
-requestInfoReplyComment: >
-  Thanks for reaching out!
-
-  We require the
-  [template](https://github.com/desktop/ONTRIBUTING.md#how-do-i-submit-a-good-bug--  to be filled out with all new issues. We do n we
-  have all the information we need to address This
-  allows the maintainers to spend more time -  enhancements, and reviewing and merging pull -
-  Thanks for understanding and meeting us -
-requestInfoLabelToAdd: more-information-needed
-
-requestInfoOn:
-  pullRequest: false
-  issue: true
+# Configuration for request-info - https://t-info
+
+# *Required* Comment to reply with
+h requestInfoReplyComm: >
+  Thanks for reaching out!
+
+  We require the
+  [template](https://github.com/desktop/ONTRIBUTING.md#how-do-i-submit-a-good-bug-+  to be filled out with all new issues. We do n we
+  have all the information we need to address This
+  allows the maintainers to spend more time +  enhancements, and reviewing and merging pull +
+  Thanks for understanding and meeting us
+
+?  halfway ≡ƒÿÇ
+
+  reques
+: nfoLabelToAdd: more-info
+
+mation-needed:
+  requestInfo: :
+  lRequ: t: f
diff --git a/.github/lock.yml b/.github/lock.yml
index 753d4c9e5..f80ce64f5 100644
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,7 +1,10 @@
+?
+  daysUntilLoc
 # Configuration for lock-threads - https://ds
-daysUntilLock: 60
[...]

PR #4863, opened 2018-07-19
fix(yaml): no wrong loc info for CRLF

ikatyang:
Context: https://github.com/prettier/prettier/issues/4856#issuecomment-405982539

Workaround for eemeli/yaml#20

### DIFF
--- a/src/language-yaml/parser-yaml.js
+++ b/src/language-yaml/parser-yaml.js
@@ -65,6 +65,8 @@ const parser = {
   hasPragma,
   locStart: node => node.position.start.offset,
   locEnd: node => node.position.end.offset,
+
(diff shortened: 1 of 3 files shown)


kachkaev:
Hi @ikatyang! Does this PR mean that until the upstream issue is fixed, Windows users who check out repos with `CRLF` will be getting `LF` in YAMLs after running Prettier? A team I'm working in has a mix of Windows / *nix laptops and `npm run lint` fails on Windows machines due to jumps from `CRLF` to `LF`. VSTS builds with Windows agents also fail for the same reason.

Shall I open an issue that describes this behaviour to make sure it is not forgotten?

PR #4877, opened 2018-07-23
feat(config): support TOML configuration files

### LINKED ISSUE #4845: Support TOML for configuration files
It would be fairly easy for Prettier to support `.prettierrc.toml` files as `cosmiconfig` supports [loaders](https://github.com/davidtheclark/cosmiconfig#loaders).

```ts
// Example loader using toml from npm (https://github.com/BinaryMuse/toml-node)
import toml from 'toml';

function loadToml(filePath: string, content: string): Object | null {
  try {
    return toml.parse(content);
  } catch (error) {
    error.message = `TOML Error in ${filePath}:\n${error.message}`;
    throw error;
  }
}
```

This [issue](https://github.com/davidtheclark/cosmiconfig/issues/150) discusses adding TOML to `cosmiconfig`'s defaults. If they do implement this, Prettier would get this feature for free.

Some reasons why TOML is better than YAML for configuration files: https://arp242.net/weblog/yaml_probably_not_so_great_after_all.html

Some arguments against JSON as a configuration language: https://www.lucidchart.com/techblog/2018/07/16/why-json-isnt-a-good-configuration-language/

jorgegonzalez:
Closes #4845

In order to support `.toml` configuration files, `cosmiconfig` must first be updated to `v5.0`. This update seems to break a bunch of tests. The following errors are thrown all over the place: 

image

image

**Update 29/7/2018**: This PR adds support for `.prettierrc.toml` config files.


jorgegonzalez:
Also, I recognize adding another dependency is not ideal for what is essentially a minor feature, but `toml` is [quite light](https://packagephobia.now.sh/result?p=toml).

foray1010:
@jorgegonzalez how about `@iarna/toml`? it is actively maintained, and it is the only js package support toml 0.5.0 at this time
https://github.com/iarna/iarna-toml

j-f1:
It’s also [much smaller](https://bundlephobia.com/result?p=@iarna/toml@2.0.0).

jorgegonzalez:
@foray1010 I'll use `@iarna/toml` if @j-f1 will allow the size increase 🙏

j-f1:
Should be fine 👍

jorgegonzalez:
Not sure why it's failing on `node4` 🤔

jorgegonzalez:
Ah I see.

j-f1:
I think you could use a snapshot in that test, which would save having to create an object to match against.

ikatyang:
I'll let other maintainers to decide if we should merge it as I hesitate about adding another config format. If someone wants to add support for other format later, what should we do?

I'm thinking of if we should make loaders be part of plugins so that people can use whatever format they want.

jorgegonzalez:
I understand the hesitance to add another format, and this is indeed an opinionated PR. Personally, I would love to see the JavaScript community to embrace TOML format configuration files more, as it is subjectively superior to both YAML and JSON configurations, and Prettier could lead the community in this. I suspect more users would also lean toward this format if they were given the chance to try it.

As linked in  #4845:

Some reasons why TOML is better than YAML for configuration files: https://arp242.net/weblog/yaml_probably_not_so_great_after_all.html

Some arguments against JSON as a configuration language: https://www.lucidchart.com/techblog/2018/07/16/why-json-isnt-a-good-configuration-language/

suchipi:
There's not a *huge* value-add imo since prettier config files are very small and you're unlikely to need comments in them, but since the work is already done let's merge it :+1:

j-f1:
Thank you for contributing to Prettier!

PR #4889, opened 2018-07-24
fix(glimmer): do not strip block params on <Component>

### LINKED ISSUE #4887: Prettier with glimmer parser strips block params from <Component>
**Prettier 1.13.7**
```sh
--parser glimmer
```

**Input:**
```handlebars
<MyComponent as |model|>
  {{model.name}}
</MyComponent>
```

**Output:**
```handlebars
<MyComponent>
  {{model.name}}
</MyComponent>
```

**Expected behavior:**
I expect block params to be preserved

tchak:
fix #4887


duailibe:
Thank you!

PR #4890, opened 2018-07-24
fix(glimmer): remove extra space from sub-expressions without params

tchak:
*before*
```handlebars
{{foo (bar )}}
```
*after*
```handlebars
{{foo (bar)}}
```

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -158,14 +158,13 @@ function print(path, options, print) {
       );
     }
     case "SubExpression": {
+      const params = getParams(path, print);
+      const printedParams =
+        params.length > 0
+          ? indent(concat([line, group(join(line, getParams(path, print)))]))
+          : "";
       return group(
-        concat([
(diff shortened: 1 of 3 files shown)


src/language-handlebars/printer-glimmer.js
```diff
@@ -157,11 +157,18 @@ function print(path, options, print) {
...
+          return indent(concat([line, group(join(line, params))]));
+        }
+        return "";
+      };
       return group(
         concat([
           "(",
           printPath(path, print),
-          indent(concat([line, group(join(line, getParams(path, print)))])),
+          printParams(path, print),
```
duailibe: I'd prefer not having this additional function, just declare another variable like:

```js
case "SubExpression": {
  const params = getParams(path, print);
  const printedParams = params.length > 0 ? indent(concat(/* ... */)) : "";
  return group(concat(["(", printPath(path, print), printedParams, /* ... */ ]));
}

duailibe:
Thanks!

PR #4891, opened 2018-07-24
fix(glimmer): better formatting of boolean attributes

tchak:
I hope this is not controversial...

*input*
```handlebars
<button disabled>Click</button>
<button disabled=disabled>Click</button>
<button ...attributes>Click</button>
```

*before*
```handlebars
<button disabled="">Click</button>
<button disabled="disabled">Click</button>
<button ...attributes="">Click</button>
```

*after*
```handlebars
<button disabled>Click</button>
<button disabled="disabled">Click</button>
<button ...attributes>Click</button>
```

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -168,7 +168,11 @@ function print(path, options, print) {
       );
     }
     case "AttrNode": {
-      const quote = n.value.type === "TextNode" ? '"' : "";
+      const isText = n.value.type === "TextNode";
+      if (isText && n.value.chars === "") {
+        return concat([n.name]);
+      }
+      const quote = isText ? '"' : "";
       return concat([n.name, "=", quote, path.call(print, "value"), quote]);
     }
     case "ConcatStatement": {
--- a/tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/html_glimmer/__snapshots__/jsfmt.spec.js.snap
@@ -112,6 +112,12 @@ exports[`component.hbs - glimmer-verify 1`] = `
 {{@greeting}}, {{@name}}!
 
 <button onclick={{action next}}>Next</button>
+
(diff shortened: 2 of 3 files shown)


tchak:
Not sure what's going on here. Local tests seems to pass. I will investigate...

j-f1:
Does running `AST_COMPARE=1 yarn test` fail on your machine? If so, you need to update the AST cleaner to avoid differentiating between the two syntaxes.

duailibe:
I'm not familiar with glimmer so I'm not sure this affects anything. Either way, that's not align with Prettier does things.

So I'd rather not do it, but I'll leave the decision to others more informed.

evilebottnawi:
@duailibe It is invalid code:
```
<button disabled="">Click</button>
<button ...attributes="">Click</button>
```

evilebottnawi:
@duailibe 
Input:
```
<button disabled>Click</button>
<button disabled=disabled>Click</button>
<button ...attributes>Click</button>
```
Output:
```
<button disabled="">Click</button>
<button disabled="disabled">Click</button>
<button ...attributes="">Click</button>
```

Prettier create invalid code in this cases

duailibe:
@evilebottnawi Ah ok :+1:

In that case I think only the `<button disabled=disabled>` => `<button disabled>` transformation looks weird

tchak:
I can see how `<button disabled=disabled>` => `<button disabled>` can be debatable, so I will remove it. I suppose this needs to be aligned with how prettier dose it for other html based dialects.

tchak:
I removed debatable formatting and the test pass now

tchak:
@duailibe it's fine. It's a stylistic concern and we can discuss it, but it is definitely not invalid. I need to look at other html dialects to see what choices they make.

duailibe:
@tchak Thanks! And thank you for the contributions!

tchak:
I fixed the mentioned case with #4897

PR #4902, opened 2018-07-28
fix command

yofriadi:


yofriadi:
Turns out it works if I install husky@next

PR #4945, opened 2018-08-07
Update install.md

ondine:
Adding Yarn install instructions to then install with yarn. Add `npm` homepage link.

- [x] I’ve added tests to confirm my change works. N/A
- [x] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory) N/A
- [x] I’ve read the [contributing guidelines] (https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -3,6 +3,8 @@ id: install
 title: Install
 ---
 
+To install using [Yarn](https://yarnpkg.com/), you'll first need to install [yarn](https://yarnpkg.com/lang/en/docs/install/) on your system.
+
 Install with `yarn`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


lipis:
I think it's already obvious..

docs/install.md
```diff
@@ -11,7 +13,7 @@ yarn add prettier --dev --exact
 yarn global add prettier
 ```
 
-_We're using `yarn` but you can use `npm` if you like:_
+You can also install with [npm](https://www.npmjs.com/get-npm) if you like:_
```
j-f1: You shouldn’t need to install npm — it comes with Node.

lipis:
We can just pass on this one I think..

Issue #4941, opened 2018-08-07
After upgrading to v1.14.0 formatted by prettier code in some cases looks ugly


yuliaHope:
**Prettier 1.14.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAPABAXnQHgCYCWAbgHwDKAhgDZwDO6MFGdMBAthfHugK60FQA5ugroATpzjoIAM3TBa1OgBUm6AFToAjAAYdAXwCkAOhwB6QqQA6UGxmz5iJG+nmsYNfQDIoAI1oADgDcNuaWJCAANCAQAazQtMigFGJiEADuAAopCIkoFEQQBHhRIL4SYADWcDBkARRgAoLIMGI8cNECtHBiMJkSghzIMtTd0QBWtKgAQhXVtRRscAAyAnDDox0g9WLdYshlFL4AnlTQpQFp3QDqEgEHl3Q9ROvRYnAAjjwE7-0UgxQNlQxiBumwCC02lt+EIaABFHgQeBAkGMXzXYowAAWyAATNFWhQCFQmgBhCBsIYoKDQV4gPhwVT+FFbS4CGAYvDY5C6PT6fRAA)
```sh
# Options (if any):
--parser babylon
--print-width 1000
```

**Input:**
```jsx
x = <div>Sales tax estimated using a rate of {salesTax * 100}%.</div>

x = <div>
  {title}&nbsp;
</div>
```

**Output:**
```jsx
x = (
  <div>
    Sales tax estimated using a rate of {salesTax * 100}
    %.
  </div>
);

x = (
  <div>
    {title}
    &nbsp;
  </div>
);

```

**Expected behavior:**
Don't split this jsx text and don't move it to new line.

gpbl:
Yeah we are waiting to upgrade to 1.14 for the same reason, and see what others are saying.

Examples:

```diff
 <span>
-    <strong>{name}</strong>’s{' '}
+    <strong>{pet.name}</strong>
+    ’s{' '}

- </strong>.<br />
+ </strong>
+ .<br />

- <Text>You {type}ed this shipment to</Text>
+ <Text>
+     You {type}
+     ed this shipment to
+ </Text>
```

The last one is the worst :)

dawnmist:
I was about to raise this as an issue, however I think it doesn't just look wrong, it changes the generated the string.

[...]

PR #4965, opened 2018-08-09
fix: splitting jsx text correctly

yuliaHope:
Fixes: [4941](https://github.com/prettier/prettier/issues/4941)


lydell:
Prettier's splitting is already correct. This is a change. Personally I don't think the snapshots look better in this PR.

lydell:
Prettier's splitting is already correct. This is a change. Personally I don't think the snapshots look better in this PR.

PR #4972, opened 2018-08-11
feat(yaml): prefer implicit key for empty value

### LINKED ISSUE #4971: How can I format my `volumes` section correctly in docker-compose.yaml?
I'm using prettier as my YAML file formatter, but I got some troubles today I edit my `docker-compose.yaml`.

the `volumes` section should looks like this, and I typed it like this.
```yaml
...
volumes:
  postgres:
  redis:
```

When I saving it, my vim call prettier automatically to format the code, and I got this
```yaml
...
volumes:
  ? postgres
  ? redis
```

How can I get a `right` and `formatted` `docker-compose.yaml` file?
Or, please fix it, if it is a bug.

Thanks all.

ikatyang:
Fixes #4971


tests/yaml_root/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -282,9 +282,9 @@ pairs:
 set:
   # Explicitly typed set.
   baseball players: !!set
-    ? Mark McGwire
-    ? Sammy Sosa
-    ? Ken Griffey
+    Mark McGwire:
+    Sammy Sosa:
+    Ken Griffey:
```
j-f1: IMO we should keep the `?` syntax in this case (`!!set`) since these keys can’t have values and they can be any type of object.

PR #4987, opened 2018-08-16
Add a label that exempts issues from auto-locking

### LINKED ISSUE #4986: breakBeforeElse – closing an issue prevents thumbs up
When you close an issue, I can no longer add a thumbs up to it. Are there any other ways of voting on various features?

Like the poster of #840, it is a deal-breaker for me when `else` statements are put on the same line as the previous close bracket. The argument, in my case, is the same as with most other rules that Prettier is doing right: It gives **better commit diffs**.

I want to be able to add any number of `else if` or `else` statements under each other, with comments and all, without it changing the previous block whatsoever.

This has been added in `prettier-miscellaneous` as a rule named `breakBeforeElse`, but I want to add my vote that it should be native to Prettier. It's just more correct.

j-f1:
This allows us to track 👍s on closed issues so we can perhaps reopen them if there’s strong enough support. (inspired by & closes #4986)

- [ ] After merging, unlock all [locked awaiting-response issues](https://github.com/prettier/prettier/issues?q=is%3Aissue+label%3A%22status%3Aawaiting+response%22+is%3Aclosed+label%3Alocked-due-to-inactivity).

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -5,3 +5,4 @@ lockComment: false
 setLockReason: true
 exemptLabels:
   - keep-unlocked
+  - status:awaiting response
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -3,3 +3,5 @@ daysUntilLock: 90
 lockLabel: locked-due-to-inactivity
 lockComment: false
 setLockReason: true
+exemptLabels:
+  - keep-unlocked


azz:
Can/should we auto-add this label if there is a lot of 👍👎?

j-f1:
> Should we auto-add this label if there is a lot of 👍👎?

Seems like a good idea 👍

> Can we auto-add this label if there is a lot of 👍👎?

There’s no way I know of to do this automatically.

.github/lock.yml
```diff
@@ -3,3 +3,6 @@ daysUntilLock: 90
 lockLabel: locked-due-to-inactivity
 lockComment: false
 setLockReason: true
+exemptLabels:
+  - keep-unlocked
+  - status:awaiting response
```
lipis: Good catch with the awaiting response..! 🍺

duailibe:
If we close an issue, we won't move forward with it, why let people add 👍/ 👎 ?

j-f1:
If there are a hundred or more 👍s, we would probably want to think about it again & reconsider — we are somewhat beholden to our users.

lydell:
Personally, when I look for something in a project and find a closed/rejected issue about it I feel so much better after +1:ing that issue (even if it doesn't make a difference in the end), rather than doing nothing.

j-f1:
I unlocked several of the issues & removed the “awaiting response” label in cases where the person had actually responded.

PR #4993, opened 2018-08-19
Update configuration to be clearer about js config

darvid7:
So in the list, it says to use `prettier.config.js` as your file name for configuring prettier in js. But in the example it has `//.prettierrc.js`. This confused me so I tried both file names. Both these file names work to configure prettier so I updated the docs to reflect that.

- [ ] I’ve added tests to confirm my change works. **n/a**
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory) **n/a**

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -5,7 +5,7 @@ title: Configuration File
 
 Prettier uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) for configuration file support. This means you can configure prettier via:
 
-- A `.prettierrc` file, written in YAML or JSON, with optional extensions: `.yaml/.yml/.json/.js`.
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -7,7 +7,7 @@ Prettier uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) for co
 
 - A `.prettierrc` file, written in YAML or JSON, with optional extensions: `.yaml/.yml/.json/.js`.
```
ikatyang: We should remove the `.js` here since we've mentioned it in the following item.

darvid7: Done!

ikatyang:
Thanks!

PR #5006, opened 2018-08-22
fix: splitting jsx text correctly

yuliaHope:
Fixes: #4941

### DIFF
--- a/tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
@@ -416,12 +416,6 @@ x =
 
 x = <div><strong>text here</strong>.<br /></div>
 
-x = <div><strong>{name}</strong>’s{' '}</div>
-
-x = <Text>You {type}ed this shipment to</Text>
-
-x = <HelpBlock>{parameter.Description}: {errorMsg}</HelpBlock>
-
 x = <div>Sales tax estimated using a rate of {salesTax * 100}%.</div>
 
 x = <div>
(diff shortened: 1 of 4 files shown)


tests/jsx-text-wrap/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -474,23 +513,9 @@ x = (
...
-    after
-    {stuff}
-    after
-    {stuff}
-    {stuff}
-    {stuff}
-    after
-    {stuff}
-    after
+    {stuff}after{stuff}after
```
lydell: Why is there one line with only a single `{stuff}` in it? This is a weird edge case so it doesn’t matter much, but I’d like to understand how things work.

yuliaHope: Because we have three jsx expressins one after another. The same [here](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAPABAXnQHgCYCWAbgHzByoAOATnAM50HQCMAvuVbQ01AEzsUa9RhADMrHAHpCpEABoQESjB51koAIbVqEAO4AFLQjUoNRCATzyQAI2oawAazgwAypQcEoAc2QxqAK5wCl50cNQw+vbeALYayABmGgA2YQoAVnSoAEL2Ti6uGjFwADJecIkpaSAe1GHUyLYaNgCeydDWNF4wAOqWMAAWyAAcAAwKNBBhPfaUjUL1RBUKtACOAQS0URqx8UhJqcEgYTEEfoFHjD7JcACKARDwlYcKMM19eIPIvK-2BMlebwAYQgMTijSg0GWIACYQAKs0TAcwqxWEA)

yuliaHope: We don't wrap text without space before/after jsx expression, but we wrap multiple jsx expressions, don't we?

lydell: Ah, now I get it. Makes sense! :+1:

lydell:
Thank you very much for doing this and sorry for the delay!

PR #5010, opened 2018-08-23
Code Refactor : Use destructures

prashant-andani:

### DIFF
--- a/src/language-vue/parser-vue.js
+++ b/src/language-vue/parser-vue.js
@@ -85,7 +85,7 @@ function decodeAttr(value, shouldDecodeNewlines) {
 
 function parseHTML(html, options) {
   const stack = [];
-  const expectHTML = options.expectHTML;
+  const { expectHTML } = options;
   const isUnaryTag = options.isUnaryTag || no;
   const canBeLeftOpenTag = options.canBeLeftOpenTag || no;
   let index = 0;
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -14,7 +14,7 @@ const docUtils = doc.utils;
 function printAstToDoc(ast, options, addAlignmentSize) {
   addAlignmentSize = addAlignmentSize || 0;
 
-  const printer = options.printer;
+  const { printer } = options;
   const cache = new Map();
 
   function printGenerically(path, args) {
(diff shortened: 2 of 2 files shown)


src/cli/minimist.js
```diff
@@ -1,14 +1,14 @@
-"use strict";
+'use strict';
 
-const minimist = require("minimist");
+const minimist = require('minimist');
```
j-f1: Can you switch these back to double quotes?

src/main/core.js
```diff
@@ -122,7 +121,6 @@ function coreFormat(text, opts, addAlignmentSize) {
       newCursorNodeStart = result.cursorNodeStart;
       newCursorNodeText = result.cursorNodeText;
     } else {
-      oldCursorNodeStart = 0;
```
j-f1: Can you explain why this isn’t needed anymore?

src/cli/minimist.js
```diff
@@ -1,14 +1,14 @@
...
 
 const PLACEHOLDER = null;
 
 /**
  * unspecified boolean flag without default value is parsed as `undefined` instead of `false`
  */
 module.exports = function(args, options) {
-  const boolean = options.boolean || [];
+  const { boolean } = options || [];
```
ikatyang: This should be `const { boolean = [] } = options`.

prashant-andani: Yes ... thanks... i will update that... i missed it...

src/main/core.js
prashant-andani: we have reassigned the variable two times...
 `oldCursorNodeStart = 0; ` and `oldCursorNodeText = text;` that means the first assignment doesnt make sense...

src/cli/minimist.js
prashant-andani: Yes...  I will ...

src/main/core.js
j-f1: I see, it’s never used outside of the `if` branch. Can you move the `let` to the assignment in the `if` branch?

prashant-andani: @j-f1 if I understood right what u mentioned, oldCursorNodeStart is used in "if", "else" and outside as well... hence it can be declared outside like the way it is...

j-f1: Where is it used outside of the `if` block? I can’t seem to find those uses.

Issue #5013, opened 2018-08-23
YAML formatting strips document begin


pbrisbin:
One thing I've noticed with the new YAML formatting (which is awesome btw!) is that it will strip a standalone (not part of frontmatter) document-separator (`---`). For example, you can see it happen [in this PR diff](https://github.com/pbrisbin/bugsnag-haskell/pull/40/files#diff-54eaec2f732ed7ea540163ee133a989fL1) (a commit made by Restyled, which runs prettier).

I don't have a definitive reference that it's good style to include these document delimiters when they're not strictly necessary, but I tend to include them because:

1. [`yamllint`](https://github.com/adrienverge/yamllint#readme) has a rule (by default) that requires them, and
2. I saw a tweet with this justification (paraphrased):

   If you concatenate multiple YAML documents into some process, e.g.

   ```console
   cat one.yaml two.yaml | some-process
   ```

   Having document-separators in the files will ensure they get processed correctly (as multiple documents). Without the separators, you may silently get something unexpected (smashing of objects) or it might just error.

Because of point (1), it's particularly annoying when ALE runs prettier as an auto-fixer, which strips the `---`, then it runs yamllint as a linter, which complains about its omission...

So, is it intentional to strip it? If so, can an option be added to preserve it?

ikatyang:
We did not have special case for it, we currently just print the necessary part so  any unnecessary parts are removed.

According to our [Option Philosophy](https://prettier.io/docs/en/option-philosophy.html), we avoid to add options if possible unless there're some critical reasons, so the only possible way to go here is to preserve them, but I'm not sure if we should do so since it may introduce inconsistency, let's see how people think about it, marked as needs discussions for now.

---

Regarding `yamllint`, we probably need something like [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) but for `yamllint`, we can refer to those rules but we're not intend to follow any linter's rules.

---

Regarding the `cat` use case, the leading [directives end (`---`)](http://yaml.org/spec/1.2/spec.html#marker/directives%20end/) only works for the case that does not contain [directives](http://yaml.org/spec/1.2/spec.html#id2781147), for example the following case won't work:

```yaml
---
content 1
```

```yaml
%YAML 1.2
---
content 2
```

unless we add the explicit document end as well:

```yaml
---
content 1
...
```

```yaml
%YAML 1.2
---
content 2
...
```

but I think nobody wants these unnecessary document ends, at least I didn't see someone wrote this before.

[...]

PR #5027, opened 2018-08-30
fix(yaml): update parsers

### LINKED ISSUE #4991: YAML: Prettier can't parse its own output
Prettier v1.14.2

I can't report this using the "Report issue" button because the code is too long, and the issue doesn't reproduce with shorter code.

Steps to reproduce:

1. Copy the content of [this file](https://raw.githubusercontent.com/AustinGreen/gatsby-starter-netlify-cms/master/static/admin/config.yml) to [Prettier Playground](https://prettier.io/playground).

2. Copy the result of formatting (Prettier's output).

3. Try to format it again by pasting it into the input pane of the Playground.

4. See this: `SyntaxError: The "undefine...ndefined" key is too long (76:13)`

ikatyang:
Fixes #4991

- upgrade to `yaml@1.0.0-rc.8` and `yaml-unist-parser@1.0.0-rc.4`
- refactor some logic since the AST has slightly changed (ikatyang/yaml-unist-parser#82)
- unmatched aliases are now errors since it may introduce invalid AST from `yaml`
- rewrite the document separator (`...`/`---`) logic, this fixes some cases where it can use `---` but we printed `...`
- removed some unnecessary duplicate trailing newline
- trailing comments on `document` (`... #comment`) and `documentHead` (`--- #comment`) are preserved (i.e. they won't be moved somewhere)

---


marcofranssen:
The comment approach mentioned above is a workarround. Many kubernetes deployment files generated by tools or available in OSS out there don't have those comments. Meaning if I need to do small change I have to add this unnecessary comment all over the place. Can't we just allow these type of document headers?

ikatyang:
Could you move your comment to #5013? as this PR is not aimed to fix that issue, I just found it _might_ be a solution. We should discuss it in #5013.

PR #5039, opened 2018-08-31
Print nested ternaries differently

### LINKED ISSUE #737: alternate ternary formatting

suchipi:
Changes formatting of switch-style nested ternaries to the suggestion in https://github.com/prettier/prettier/pull/4767#issuecomment-401764876.

Fixes #737

- [ ] (n/a) (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)


tests/ternaries/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -718,10 +718,10 @@ a
...
     ? bbbbbbbbbbbbbbbbbb
     : ccccccccccccccc
-        ? ddddddddddddddd
-        : eeeeeeeeeeeeeee
-            ? fffffffffffffff
-            : gggggggggggggggg;
+      ? ddddddddddddddd
+      : eeeeeeeeeeeeeee
+        ? fffffffffffffff
+        : gggggggggggggggg;
```
ikatyang: This one seems not working.

tests/ternaries/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1633,8 +2301,8 @@ a => a ? aasdasdasdasdasdasdaaasdasdasdasdasdasdasdasdasdasdasdasdasdaaaaaa : a
 debug ? (this.state.isVisible ? "partially visible" : "hidden") : null;
 debug
 	? this.state.isVisible && somethingComplex
-		? "partially visible"
-		: "hidden"
+	  ? "partially visible"
+	  : "hidden"
```
ikatyang: This shouldn't be changed, we have a special case for ternary in ternary to use tabs in `--use-tabs` mode, that's what the removed part in 3018dd1ff5c2669dff0b3feb593f4ca9ba526175 wanted to achieve.

suchipi: Ok, thanks. I will fix it

lydell:
@suchipi Do you think you’ll be able to finish this one up?

duailibe:
@ikatyang Can you check if 23cc0cc fixed it?

ikatyang:
`--use-tabs` is working, but spaces are not.

**Prettier pr-5039**
[Playground link](https://deploy-preview-5039--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBDABAfnQRiQJkx0wGYkAWAkAGhAgAcYBLaAZ2VFQCcuIB3AArcE7FKgBuEJgBMaIAEZdUYANZwYAZXrKmUAObIYXAK5xaACxgBbADYB1c03ittYOBpFOm4pwE9k4KzstLqscFwwAkp6VqjIAGaoNmG0AFasAB4AQkqq6hqoVnAAMrpwCUkpINpcYVwB8qjyvjbQcvRcujB2MjDmyDi0HRBhdkr0AR1wdeLltFxwAI7GTAtRqDFxSInJZiBhVkyGJnusuno2cACKxhDwFbu0ME090n3IAKxPSkw25wDCECssQCUGgcxAxjCABUmqIdmEAL6IoA)
```sh
--parser babylon
--print-width 1
--tab-width 5
```

**Input:**
```jsx
a ? 1:2?1?3:4:2
```

**Output:**
```jsx
a
     ? 1
     : 2
     ? 1
       ? 3
       : 4
     : 2;

```

**Expected Output**
```jsx
a
     ? 1
     : 2
     ? 1
          ? 3
          : 4
     : 2;

```

tests/typescript_conditional_types/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -25,6 +33,16 @@ type DeepReadonlyObject<T> = {
...
 
+type TypeName<T> = T extends string
+  ? "string"
+  : T extends number
+  ? "number"
+  : T extends boolean
+  ? "boolean"
+  : T extends undefined
+  ? "undefined"
+  : T extends Function ? "function" : "object";
```
ikatyang: Unrelated to this PR: We should force breaking nested TS type ternaries just like normal ternaries.

suchipi:
Thanks @ikatyang, my life has been really busy lately because I am getting a new job and my apartment got flooded, I really appreciate the help ❤️

duailibe:
Great work @suchipi and @ikatyang. Thank you!

davidmccabe:
This is a huge improvement, thanks for working on it.

Looking at the test file, it seems that one of the examples is still not intended, lines 5 – 11:

```
const value = condition1
? value1
: condition2
    ? value2
    : condition3
        ? value3
        : value4;
```

It's not clear why this example is formatted differently from the others. Is this intended?

suchipi:
@davidmccabe could you link the the file and line that you're looking at?

It's possible that you're looking at the unformatted input. Our snapshots look like this:

```
input
~~~~~~~~~
output
```

fbartho:
@suchipi / @ikatyang -- It looks like this PR implemented changes that led to this issue being opened: https://github.com/prettier/prettier/issues/5257 -- would somebody mind commenting in there to see if that issue should be instead marked as a "Won't-fix"/"Working as intended"?

If this is the new way, and that issue isn't a bug so much as a stylistic preference that doesn't match with the changes introduced here, then I'll just deal.

lydell:
@fbartho This PR intended to flatten the “tail” of ternaries – if it caused #5257 it is a bug.

PR #5041, opened 2018-09-01
feat: add printer.preprocess

ikatyang:
Sometimes we need to transform the ast to make it easier to print, but it's currently done in the parser (markdown and yaml), which leads the output ast harder to use for external users (custom parser), adding `printer.preprocess` can solve this issue.

And also this way we could move the trailing newline for json from `ast-to-doc.js` to `language-js`.

---


ikatyang:
text -> parser.preprocess -> parser (custom parser) -> printer.preprocess -> printer

Issue #4989, opened 2018-08-17
big ignore paths slow down cli


eps1lon:
**Environments:**
- Prettier Version: 1.14.2
- Running Prettier via: CLI
- Runtime: v10.8.0
- Operating System: Linux

**Steps to reproduce:**
```bash
mkdir repro-dir && cd repro-dir
npm init
npm add prettier
# add some dependencies to feel the weight on the prettier cli
npm add express typescript webpack
touch index.js
cp -r node_modules a_deeply_nested_structure
echo "a_deeply_nested_structure" > .prettierignore
node_modules/.bin/prettier --list-different "**/*.js"
```

**Expected behavior:**
Does not walk down any folder in `a_deeply_nested_structure`. Prettier finishes instantly.

**Actual behavior:**
`glob` does walk down `a_deeply_nested_structure`. They get ignored in the end by prettier but this can slow down runtime heavily if we want to ignore big folders. You can check this by inspecting https://github.com/prettier/prettier/blob/d480858f016c6b165e55195027226a612e97d47c/src/cli/util.js#L444-L452

Is there a reason why the ignore patterns are not passed to `node-glob`?

If I got this right this behavior was request by #3590 but in my opinion it should be the responsibility of the extension to get knowledge of `.prettierignore`. As of right now the cli is very slow when someone has to ignore paths other than node_modules.

j-f1:
Perhaps we can add a conditional before we call the ignorer and add the ignore file exclusions to the glob when `context.argv["debug-check"] || context.argv["write"] || context.argv["list-different"]`.

eps1lon:
I'm intersted in why there is the need for the ignorer to begin with. Isn't this a clear violation of the Separation of Concerns principle which imposes this performance penalty? 

You're also listing both `minimatch` and `ignore` as dependencies both of which have different matching specs. All of this makes it very hard to reason about when and why files are ignored.

ikatyang:
#3590 is about if we should output ignored file, which did not affect the glob searching.

Here's how we process files:

1. determine which files we should look for using `globby`
   - walk through the `fs` tree based the patterns from CLI
   - `--with-node-modules` affects this step
2. ignore files that match the patterns listed in `.prettierignore` using `ignore`
   - ignored files are printed to stdout if `--write` not provided, which is requested by #3590
[...]

PR #5053, opened 2018-09-04
Bump globby

duailibe:
Another attempt at #4715

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "find-project-root": "1.1.1",
     "flow-parser": "0.78.0",
     "get-stream": "3.0.0",
-    "globby": "6.1.0",
+    "globby": "8.0.1",
     "graphql": "0.13.2",
     "html-tag-names": "1.1.2",
     "ignore": "3.3.7",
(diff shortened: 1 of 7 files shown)


ikatyang:
I've tried that as well and it seems there're too many modules need to be loaded for `fast-glob`, which slows down the startup speed and we have a lot of test files.

We may want to switch to [`glob-gitignore`](https://github.com/kaelzhang/node-glob-gitignore) for ignoring folders based on `.ignore` file while traversing the file system (see https://github.com/prettier/prettier/issues/4989#issuecomment-414062231 for more info), which seems not supported in `globby@8` (`fast-glob`).

ikatyang:
The `.ignore` issue for `fast-glob` should be tracked in mrmlnc/fast-glob#106 (use a custom `readdir` to return empty for `.ignore`d directories), though it seems there's not much interest on it.

evilebottnawi:
@ikatyang He is working on C++ library (extension for `nodejs`) for reduce `fs` call and speedup glob

evilebottnawi:
But better don't update `globby` (a lot of other bugs in `fast-glob`, not full compatibility with original glob)

PR #5057, opened 2018-09-05
refactor: move cli leven suggestion to options-normalizer

ikatyang:
Option-related processing should be done in `options-normalizer`.

- unknown flag suggestions are colored
  image
- unknown flags with no suggestion are now errors:
  ```
  $ bin/prettier.js --help noseminosemi
  [error] Invalid `--help` value. Expected `a flag`, but received `"noseminosemi"`.
  ```

---

### DIFF
--- a/tests_integration/__tests__/__snapshots__/early-exit.js.snap
+++ b/tests_integration/__tests__/__snapshots__/early-exit.js.snap
@@ -146,21 +146,8 @@ exports[`show version with --version (stderr) 1`] = `""`;
 
 exports[`show version with --version (write) 1`] = `Array []`;
 
-exports[`show warning with --help not-found (stderr) 1`] = `
-"[warn] Unknown option name \\"not-found\\"
-"
-`;
-
-exports[`show warning with --help not-found (stdout) 1`] = `
-"-h, --help <flag>
-
-  Show CLI usage, or details about the given flag.
-  Example: --help write
-"
-`;
-
 exports[`show warning with --help not-found (typo) (stderr) 1`] = `
-"[warn] Unknown option name \\"parserr\\", did you mean \\"parser\\"?
+"[warn] Unknown flag \\"parserr\\", did you mean \\"parser\\"?
 "
 `;
 
@@ -189,8 +176,6 @@ Valid options:
 `;
 
 exports[`show warning with --help not-found (typo) (write) 1`] = `Array []`;
-
-exports[`show warning with --help not-found (write) 1`] = `Array []`;
 
 exports[`throw error and show usage with something unexpected (stderr) 1`] = `""`;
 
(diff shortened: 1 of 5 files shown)


tests_integration/__tests__/__snapshots__/early-exit.js.snap
```diff
@@ -311,6 +296,15 @@ exports[`throw error with --find-config-path + multiple files (stdout) 1`] = `""
 
 exports[`throw error with --find-config-path + multiple files (write) 1`] = `Array []`;
 
+exports[`throw error with --help not-found (stderr) 1`] = `
+"[error] Invalid --help value. Expected \\"arrow-parens\\", \\"bracket-spacing\\", \\"color\\", \\"config\\", \\"config-precedence\\", \\"cursor-offset\\", \\"editorconfig\\", \\"file-info\\", \\"find-config-path\\", \\"h\\", \\"help\\", \\"ignore-path\\", \\"insert-pragma\\", \\"jsx-bracket-same-line\\", \\"l\\", \\"list-different\\", \\"loglevel\\", \\"no-bracket-spacing\\", \\"no-color\\", \\"no-config\\", \\"no-editorconfig\\", \\"no-semi\\", \\"parser\\", \\"plugin\\", \\"plugin-search-dir\\", \\"print-width\\", \\"prose-wrap\\", \\"range-end\\", \\"range-start\\", \\"require-pragma\\", \\"semi\\", \\"single-quote\\", \\"stdin\\", \\"stdin-filepath\\", \\"support-info\\", \\"tab-width\\", \\"trailing-comma\\", \\"use-tabs\\", \\"v\\", \\"version\\", \\"with-node-modules\\" or \\"write\\", but received \\"not-found\\".
```
j-f1: Would it be easy to make this into a bulleted list? I feel like it would be easier to read that way.

ikatyang: Ref: ikatyang/vnopts#9

Something like this?

image

j-f1: That looks great!

duailibe: I personally wouldn't show every option available.. in this case, I'd just show "Invalid --help value"

j-f1: Perhaps “Invalid --help value, expected an option name but got X.”?

ikatyang: Since we called it `flag` in the usage, "expected `a flag`" should be better I think?

https://github.com/prettier/prettier/blob/4d488e8b81276e310b00c097b1ac9e385181c3cd/tests_integration/__tests__/__snapshots__/early-exit.js.snap#L124-L125

PR #5092, opened 2018-09-13
Prevent JSX from breaking a logical expression chain

### LINKED ISSUE #5089: JSX gets different indentation levels based on number of boolean conditions
**Prettier 1.14.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAVOGDCEAtgA7QJYC8mAFAJSaUB8mwAOlJpgE5wwCu3TjQ5cuAHiaixXYGACGMNJgBkKzETibla2tJkSAYhAiYARvO5Nx2bhAA2Dh5gD01l8YhTOBugF99MWAAE34wAGsddREfA0xxT3NLa1snR3tXd09vOP9A8TdpOgBuDj8QABoQCBIYAEt0ZFBLOwB3AAVLBDRkEHkANwg64MqQM255CL4AZRJJuqgAc2QYbn44KoW0OG4YdonFonlkADN5e22qgCs0AA8AIQmpmGn5TQAZBbhT88uQOe4224vQsZgAno4oKMSNwFjAAOrDGAAC2QAA4AAxVGEQbbwiYkXowvA7frfKq8ACO-DqvH28kOxyQZwuGxA2yIdRWazZaAWi3scAAivwIPAfqyqjB5GZEcEUcgACxSiZ1ez8whEI69KDkUb8bbYGU9Zm-OB+PxAA)
```sh
--parser babylon
--tab-width 4
```

**Input:**
```jsx
const TestComponent = () => {
  return (
    <>
      {cats && memes && (
        <Foo bar><Trololol /></Foo>
      )}
      {ducks && (
        <Foo bar><Trololol /></Foo>
      )}
    </>
  );
}
```

**Output:**
```jsx
const TestComponent = () => {
    return (
        <>
            {cats &&
                memes && (
                    <Foo bar>
                        <Trololol />
                    </Foo>
                )}
            {ducks && (
                <Foo bar>
                    <Trololol />
                </Foo>
            )}
        </>
    );
};

```

**Expected behavior:**

Having different levels of indentations based on the number conditionals makes the code harder to read. I'd suggest to wrap only if the line exceeds the line length like it's already the case for normal `if (...)` conditions outside JSX, and maybe when wrapping use a style that does not require an extra indentation level (e.g. using half the tab-width when wrapping conditionals in JSX)

duailibe:
Closes #5089 

Maybe we want to think about doing the same for objects and arrays 


tests/jsx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -823,8 +823,7 @@ exports[`logical-expression.js - flow-verify 1`] = `
 </div>;
 
 <div>
-  {a &&
-    b && (
+  {a && b && (
       <span>
```
suchipi: This indentation level looks too deep

duailibe: You're right, I missed that

duailibe:
I don't think there's a way to solve this without a new primitive, or using different (but worse IMO) heuristics..

What I'm currently doing is: if the last member is JSX, split in two groups: one for the JSX and another for everything else. In that case, we can either have:

```js
{a && b && (
    <Foo />
)}
{a &&
  b && (
    <Foo />
  )}
```

versus:

```js
{a && b && (
  <Foo />
)}
{a &&
  b && (
  <Foo />
  )}
```

suchipi:
@duailibe I *think* we could solve it by checking if the last thing in the ternary chain is JSX when we print every ternary child, but it's duplicative work (from each child, walk up to the parent, then walk back down to check if the chain's last child is JSX). I guess we could use some kind of cache if we wanted to reduce the scope, but it'd be nice to have a primitive that works like React context for this particular use-case; something like (pseudocode):
* when encountering ternary, walk down chain and check if last child is JSX
* if it is, wrap the whole thing in a `context("withinTernaryWithJSXAsLastChild", group(...))`
* when rendering each child, use something like `ifContext("withinTernaryWithJSXAsLastChild", dontGroup(...), group(...))`

suchipi:
The idea being that every doc node in the tree could use `ifContext` to ask for the current context and it would be the result of all `context(...)` nodes up the tree, eg with this structure:
```js
context("foo",
  context("bar",
    group(...)
  )
)
```
Inside the group, context is `{ foo: true, bar: true }`, so it can use `ifContext("foo"` to render one way or the other depending on if the context's `foo` property is true, etc.

duailibe:
@suchipi I was able to fix this using the `groupId` in `ifBreak` that @ikatyang created.

In that case, everything on the left of the chain is in a single group and I can indent or not depending if that group broke or not :-)

PR #5108, opened 2018-09-19
feat(js): move binary operators to beginning of line

### LINKED ISSUE #3806: Placing operators at the beginning of lines
**Prettier 1.10.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA3AhgJwATuwXmwHIBGAJgGYKBWAFmptpqOwGpjLayuKSGiA3CAA0ICAAcYAS2gBnZKCyYIAdwAKWBPJTpUEKQBMRIdLJjIAZugA2suKIBGmdGADWcGAGVxLqVADmyDCYAK72IAYQYJY2dqJ+dpgwas7+ALboMbbhAFayAB4AQs5uHp7oaXAAMn5wWXFiITDiTWT14T6YicggDugOAJ7W0MbimH4wAOqGMAAWyGRkomMQdpPO4j1jcImodaKYcACOIVKHKejpmUhW2aJ2aVJBoeGyfv7WcACKIRDw7aIYP1pgY5shaIDnFJrO8AMIQNIZHpQaD7EAhOwAFX62ludgAvvigA)
```sh
# Options (if any):
--print-width=22
--tab-width=4
```

**Input:**
```js
var a = '123354535435' + '2342423153';
```

**Output:**
```js
var a =
    "123354535435" +
    "2342423153";
```

**Expected behavior:**
```js
var a
    = "123354535435"
    + "2342423153";
```

btmills:
When binary expressions span multiple lines, this change moves the operator to the beginning of the second line instead of the end of the first. This fixes #3806.


vjeux:
Hey, I’m sorry but this is a core decision that affects almost every printed line of code and at this point we will not change the existing behavior. This would be a huge breaking change and be way too controversial to everyone using prettier today.

btmills:
No worries! The original issue was locked, so I couldn’t ask if there was still any interest there, but it was easy enough just to submit a PR. Is there any interest in doing this behind an option instead?

kachkaev:
Something to consider for v2?

suchipi:
I think at this point it would be more appropriate for this to be in a fork rather than v2 or behind an option, I don't want to invite the onslaught of opinions

Issue #4638, opened 2018-06-05
Forbid setting parser at toplevel in config file?


lydell:
Extracted from: https://github.com/prettier/prettier/issues/4633#issuecomment-394454392

We've had several issues where people have put `parser: something` at toplevel in their .prettierrc, causing _everything_ to be treated as JS instead of CSS, TS, etc.

Should we forbid setting parser at toplevel, since it's never useful and an easy mistake to do?

It would still be possible to use it in overrides (for specific files).

duailibe:
I think we should update the docs and emit a warning now, and see if we can forbid the config in the future

lydell:
We should also update the docs in https://prettier.io/docs/en/configuration.html to not show examples with toplevel parser.

PR #5129, opened 2018-09-22
Improve documentation for the parser option

lydell:
- Warn about setting parser at the top level. Refs. #4638.
- Remove top level parser from examples. Instead show the most common
options (I also removed printWidth, since we don't recommend changing
it). Refs. #4638.
- Show how to configure the flow parser for .js files. Refs. #4638.
- Update valid options for the parser option, including links and "since
version".
- Clarify babylon vs flow.


docs/options.md
```diff
@@ -160,17 +160,29 @@ These options cannot be used with `cursorOffset`.
...
+- `"json5"` (same parser as `"json"`, but outputs as [json5](https://json5.org/)) _Since v1.13.0_
+- `"json-stringify"` (same parser as `"json"`, but outputs like `JSON.stringify`) _Since v1.13.0_
+- `"graphql"` (via [graphql/language](https://github.com/graphql/graphql-js/tree/master/src/language)) _Since v1.5.0_
+- `"markdown"` (via [remark-parse](https://github.com/wooorm/remark/tree/master/packages/remark-parse)) _Since v1.8.0_
+- `"vue"` (uses several parsers) _Since 1.10.0_
+- `"yaml"` (via [yaml](https://github.com/eemeli/yaml) and [yaml-unist-parser](https://github.com/ikatyang/yaml-unist-parser)) _Since 1.14.0_
+
+<!-- TODO: Uncomment and move below "markdown" above when 1.15.0 is released.
+- `"mdx"` (same parser as `"markdown"`, with some custom overrides) _Since 1.15.0_
+-->
```
ikatyang: Since there's a `since` field, I guess we don't need to hide it?

(Actually, I forgot to hide the one on the homepage. 😅)

lydell: I think it is better to hide it. If I would stumble upon that in the docs, I'd try to upgrade Prettier in my project, get confused why I got 1.14.x, go hunt for 1.15.0 in the repo, and finally be sad when I found it's not released yet.

PR #5132, opened 2018-09-22
fix(html): do not format unknown script

ikatyang:
- do not format unknown script
- do not normalize spaces in script tags

---

### DIFF
--- a/tests/html_js/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/html_js/__snapshots__/jsfmt.spec.js.snap
@@ -96,6 +96,18 @@ document.getElementById("demo").innerHTML = "Hello JavaScript!";
 
 `;
 
+exports[`something-else.html - html-verify 1`] = `
+<script type="text/template">
+ <div>
+    </div>
+</script>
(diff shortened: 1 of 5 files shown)


tests/html_js/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -176,7 +189,31 @@ exports[`typescript.html - html-verify 1`] = `
   }
 </script>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-<script type="application/x-typescript">class Student { fullName: string; constructor(public firstName: string, public middleInitial: string, public lastName: string) { this.fullName = firstName + " " + middleInitial + " " + lastName; } } interface Person { firstName: string; lastName: string; } function greeter(person : Person) { return "Hello, " + person.firstName + " " + person.lastName; } let user = new Student("Jane", "M.", "User"); document.body.innerHTML = greeter(user);</script>
+<script type="application/x-typescript">
```
ikatyang: There're syntax errors in the following two cases, so we just preserve them. I was confused with these changes at the first glance, but it turns out it's just the result for the space normalization fix.

PR #5143, opened 2018-09-24
Escape Quotes in Glimmer String Literals

mahirshah:
Previously, escaped quotes inside a string literal passed to the glimmer
printer would become unescaped, causing parsing errors after printing.
Properly preserve escaped quotes when writing string literals in the
glimmer printer.


lydell:
Hi! Thanks for the PR?

I'm not super familiar with Glimmer. Could there be other characters that need escaping as well, such as backslashes and newlines? Is there a spec for this?

Also, a related question is: Why are the backslashes for the quotes lost in the first place? Are they removed by the parser? Does that mean that running Prettier on your file will turn your escapes into actual characters?

mahirshah:
Hi @lydell, thanks for taking a look. I updated the RB to follow the same pattern that the glimmer printer will use for StringLiterals (JSON.stringify), so this should be consistent with expected behavior. [https://github.com/glimmerjs/glimmer-vm/pull/835/commits/4c2ecde5ebb6261b3e3e4c5d96d67689dc8d4970]

The backslashes are removed during the glimmer parsing because it simply stores the raw string in the string literal node. When we print the string, we have to ensure that the raw string is properly escaped.

And yes, currently running Prettier on my file will turn my escapes into actual characters, which in turns makes it invalid hbs.

lydell:
Nice! Could you add tests for backslashes and newlines, too?

As a side note, ideally we'd want to preserve the escapes [just like we for strings in JS](https://prettier.io/docs/en/rationale.html#strings). If you're up for another PR doing that, it would be appreciated! But it probably requires some change in the glimmer parser, so I'm aware it might not be super simple.

lydell:
@mahirshah Nice! Everything you said seem correct, go for it! :+1:

lydell:
It's starting to look good!

Could you add tests where:

- A string goes from single quotes to double quotes because it leads to fewer escapes?
- Same thing, but the other way around.
- A test with `singleQuote: true` (check how the JS tests do it, I don't remember from the top of my mind).

lydell:
Maybe also add a comment before the `printStringLiteral` explaining how it differs from `printString` in `common/util` so people don't waste time trying to refactor / clean up in a way that won't work.

PR #5153, opened 2018-09-28
Add stylelint integration docs

BPScott:
[stylelint-prettier](https://github.com/BPScott/stylelint-prettier) exposes prettier changes a within stylelint, in the same way that eslint-plugin-prettier exposes prettier changes within eslint.

Prettier has docs on how to integrate with ESLint using eslint-plugin-prettier at https://prettier.io/docs/en/eslint.html.

This PR adds docs for how to integrate with Stylelint using the stylelint-prettier. Reading this may give you deja vu as the doc is almost identical to eslint.md, as the mechanism for integration is deliberately very similar. This PR was mostly a case of copying the existing doc and replacing instances of "eslint" with "stylelint".

---

This also removes a "just" from one sentence in the eslint doc as it minimizes prior knowledge.


azz:
I think we should make one linter integration/setup page. Would be nice to do a configurator like https://babeljs.io/setup, but a single doc with a ToC/jump links would suffice.

BPScott:
When I get a bit of free time I'll repurpose this PR to make a single "Integrating with linters" page with separate ESLint and Stylelint sections.

BPScott:
@lydell, @lipis, @azz: I've updated this to be a single "integrating with linters" page with ESLint and Stylelint sections.

I've done this in a 2nd commit so you can see what my original PR was in the 1st commit. Let me know if this is any good, and I'll rebase it all before merge (assuming it's good).

website/static/_redirects
```diff
@@ -1,6 +1,7 @@
 # Redirect removed doc links
 /docs/:lang/usage.html          /docs/:lang/install.html
 /docs/:lang/usage-options.html  /docs/:lang/options.html
+/docs/:lang/eslint.html         /docs/:lang/integrating-with-linters.html
```
lydell: Nice :+1:

docs/integrating-with-linters.md
```diff
@@ -0,0 +1,119 @@
...
+title: Integrating with Linters
+---
+
+Prettier can be integrated into workflows with existing linting tools.
+This allows you to use Prettier for code formatting concerns, while letting your linter focus on code-quality concerns as outlined in our [comparison with linters](comparison.md).
+
+Whatever linting tool you wish to integrate with, the steps are broadly similar.
+First disable any existing formatting rules in your linter that may conflict with how Prettier wishes to format your code. Then you can either add an extension to your linting tool to format your file with Prettier - so that you only need a single command for format a file, or run your linter then Prettier as separate steps.
+
+All these instructions assume you have already installed `prettier` as a developer dependency.
```
lydell:
```suggestion
All these instructions assume you have already installed `prettier` as in your `devDependencies`.
```

I know it's not a real word, but it makes it immediately clear where to put it. I've always thought of it as "development dependencies", so it didn't immedialy click for me. What do you think?

Also, there are a couple more occurrences of "developer dependencies" below.

j-f1:
How do you feel about adding TSLint and `prettier-{eslint,tslint,stylelint}` as documented [here](https://prettier.io/docs/en/related-projects.html)?

docs/integrating-with-linters.md
BPScott: Good call. You're right it is "development dependencies" and I just had a brainfart. `devDependencies` certainly offers clearer intent

BPScott:
I knew somebody would mention TSLint :D

I've added a section on TSLint. I have not touched upon `prettier-{eslint,tslint,stylelint}` yet as 
I need to have a think about the best way to introduce those tools and how to explain why you would want to use them over the plugin approach.


Somewhat unrelated: In the TSLint docs we link to the repos which all live outside the prettier org. If we consider TSLint 'official' enough to be mentioned in our usage docs should we think about moving them into the prettier organization?

j-f1:
Cc @prettier/core on the repo move. I can make the teams and invite the maintainers.

ikatyang:
I'll move `tslint-plugin-prettier` to this org tomorrow.

cc @alexjoverm, would you like to move `tslint-config-prettier` to this org?

alexjoverm:
@ikatyang yeah, that's fine for me. I think it totally makes sense to move it to the org. Just make sure I have access to it please

ikatyang:
Sure, I just sent the invitation.

ikatyang:
`tslint-config-prettier` and `tslint-plugin-prettier` are under the prettier org now.

docs/integrating-with-linters.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Recommended configuration
+
+`stylelint-prettier` exposes a "recommended" configuration that configures both `stylelint-prettier` and `stylelint-config-prettier` in a single step. Add both `stylelint-prettier` and `stylelint-config-prettier` as developer dependencies, then extend the recommended config:
+
+```bash
+yarn add --dev stylelint-config-prettier stylelint-prettier
+```
+
+Then in `.eslintrc.json`:
```
lougreenwood: Looks like a typo, should be:

```suggestion
Then in `.stylelintrc`:
```

lydell:
I stumbled upon this PR again today by chance. Why haven’t we merged this? Looks great to me. (I tried to commit the typo fix, but I didn’t have permission.)

azz:
Let's just merge and fix the typo afterwards?

PR #5157, opened 2018-09-28
fix(js): quote property name as es5 compatible

koba04:
Currently, Prettier quotes property names if the name isn't es6 compatible.
It means that Prettier might strip the quote for a property name that is es6 compatible but is not es5 compatible.
So this might break scripts on ES5 environments.

For example, the following example doesn't work on some ES5 environments.
You can try the script on Node v0.12.

**Prettier 1.14.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA3AhgJwAQQEYBW2AvNsADpTbkiAAG4OV7NS2A5BANYuUC+A3JZUhQAzhAA2cAHRiIAcwAU+AgEp+UEABoQEAA4wAltGHJQWTBADuABSwJjKdKgj6AJppB5M6MOzgwAyjre+lCyyDCYAK5wWiHCcJgwVl6yALboyABm6GLxWgTCAB4AQl4+fv7oqXAAMiFwWTl5IEGY8ZjIHuh4AJ4y6lo6mCEwAOquMAAWyAAcAAyD5vGjXjqdQ3DtqA1amHAAjpH6e8noaRlI2bkxIPGp+uFRN8IhshIAipEQ8I3XWjDdcYuKbIABM-y8+jErwAwhBUulOlBoDsQJF4gAVbr2K7xbjcIA)
```sh
--parser babylon
```

**Input:**
```jsx
var obj = {
 "𐊧": 'ok'
};

console.log(obj);

```

**Output:**
```jsx
var obj = {
  𐊧: "ok"
};

console.log(obj);

```

This PR is to fix this.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -28,7 +28,7 @@ const {
   isNextLineEmptyAfterIndex,
   getNextNonSpaceNonCommentCharacterIndex
 } = require("../common/util-shared");
-const isIdentifierName = require("esutils").keyword.isIdentifierNameES6;
+const isIdentifierName = require("esutils").keyword.isIdentifierNameES5;
 const embed = require("./embed");
 const clean = require("./clean");
 const insertPragma = require("./pragma").insertPragma;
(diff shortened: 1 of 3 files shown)


koba04:
Thank you!

PR #5176, opened 2018-10-02
Fix expect.resolves assert

ahutchings:
This adds a <del>currently</del> *no longer* failing test case for using getFileInfo with a filePath containing nested, relative globs.

When using prettier-atom, which makes use of getFileInfo, I found that I was unable to ignore nested paths like the one used by this new test case. I think the issue is due to getFileInfo requiring an absolute path to the file, but not stripping off the common part of the path between the .prettierignore path and the file path. This results in the relative globs in .prettierignore being compared against the absolute file path, which will never match.

NOTE: I also fixed a previous test to await a couple of promise-based assertions - otherwise Jest continues processing without failing the test if the assertions fail.

- [X] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)


ahutchings:
@duailibe Thanks - I've updated the test to replace multiple async assertions with Promise.all.

fisker:
I think this one is fixed by https://github.com/prettier/prettier/pull/8585

evilebottnawi:
Let's close

thorn0:
We indeed have a lot of incorrect tests that use [`expect(...).resolves`](https://jestjs.io/docs/en/asynchronous#resolves--rejects) without returning a promise in this file.

fisker:
I  prefer  `await  expect(...).resolves...`, easier to  do  something between asserts.

ahutchings:
Updated to use async/await syntax in the added and changed tests. It may be worth adding some new ESLint rules (in a separate PR) to enforce use of async/await in tests. These two rules would be a good start:

```yaml
jest/no-test-return-statement: error
jest/valid-expect: error
```

fisker:
Thanks for fixing this, I'm enabling `jest/valid-expect` in #8756

PR #5178, opened 2018-10-02
Update language support docs

timup:
This commit updates docs language to reflect the change in #4528 that removes the inferred default parser.


lydell:
Hmm, actually, does this file provide any useful information at all? I'd say we remove it.

timup:
Not especially, as the same information is available in [/docs/options.md](https://github.com/prettier/prettier/blob/master/docs/options.md)

j-f1:
Should we keep the part about the Flow test suite somewhere?

lydell:
No, I don't think so. It was important when Prettier was new an untested, but not so much anymore. It's not even entirely true anymore (Prettier supports all Flow features, but `tests/flow` does not contain the _entire_ Flow test suite anymore).

lipis:
So shall we remove it or what?

lipis:
I say we kill it with fire!

duailibe:
@timup Do you want to update the PR removing the file?

j-f1:
IMO the facts that we require a minimum TS version and support all JS proposed features should be kept somewhere.

duailibe:
Thank you @timup!

PR #5188, opened 2018-10-04
Keep decorators inline if they were written inline

### LINKED ISSUE #4924: Angular Output/Input decorators are moved to new line
**Prettier 1.14.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAwhAtgB2gmACmAB0oACcgZzgBs4wYIAnJcgciYAsBaLuZhB4AjAK4wmUdgBoyleLloBDeGwAGAHjEToAPmDBlwugF8TGgPTbJutWRMBKMnAAeeZjHJhlVKuQASAhAAQuKSmLj4sOSkFOSoAPLiOOKEDl5cSlAA5nDkALzkUHAA7uQAogBuBOVYAJYSAhpZAJ66aQDccvEAklApROlGdGxUMMx1OV1QJiDSIBA4MHXQVMigSsyCJQAKmwhrKEqVEHUAJnMgwsxKYADWcDAAyji3k9nI46Jw85M0HjsbtksEpkAAzJS0GjzABWVBcwRu90eTyUWDgABlJnBwZDoSBXsx-sgQDAWjg4FQwBMlpccBNYAB1c4wLjIAAcAAZ5vSIDRGTccCT6ZSBNVLsw4ABHUR1SWApTA0FICFQn4gGj1T7Mb7zKjvegARVEEHguLV8xgSmEzLOrOQACZLTc6rR3hEQSSoPhLqIaAAVa2HVU0MxAA)
```sh
--parser typescript
```

**Input:**
```tsx
@Component({
  selector: 'toh-hero-button',
  template: `<button>{{label}}</button>`
})
export class HeroButtonComponent {
  @Output() change = new EventEmitter<any>();
  @Input() label: string;
}
```

**Output:**
```tsx
@Component({
  selector: "toh-hero-button",
  template: `<button>{{label}}</button>`
})
export class HeroButtonComponent {
  @Output()
  change = new EventEmitter<any>();
  @Input()
  label: string;
}

```

**Expected behavior:**
Decorators are kept on the same line.

**More details:**
When using Angular, it's recommended to follow their [style guide](https://angular.io/guide/styleguide), [which specifies](https://angular.io/guide/styleguide#decorate-input-and-output-properties):
> Placing the decorator on the same line usually makes for shorter code and still easily identifies the property as an input or output. Put it on the line above when doing so is clearly more readable.

Angular classes are a valid (and common) use-case for Prettier, but with Prettier 1.14 ([relevant PR](https://github.com/prettier/prettier/pull/4830)) it deviates from the Angular style guide, causing unwanted changed for most people using Angular.
Note that the above code example (input) is taken directly from the Angular style guide.

The original PR to change this behavior mentions that mobx is the only library using same-line decorators, which is clearly not the case, with Angular being a (big) second one.

To avoid reverting the new behavior completely, which is probably still wanted by some people - I suggest adding an option to the config to keep the old behavior, or prefer the new one (default should be discussed further).

### LINKED ISSUE #5088: Decorators for member variables should be on the same line
**Prettier 1.14.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACMAGwEMBnUvAZTmOzAAtKYc49gAdKPPAAQgCNScbADdi-QqyG0GAFXT4AvHgDkKztz6DhYiawC2ccsQDmrZWo28BQ0eMl4A7sQCWMF1BN5lAM2KEhKy1bXQchSTB4ABMASSjvPCgAV0JCIJsde1Z0DGIoKLh4339AqCtsIxSYCmUMuz0AOlpsYgBPAAoASk4gyH0MJOirM3xwuEjCrrYrAF9OGZAAGhAIDHdoUmRQZohHAAVaBE2UYhEIFyilkH4WsABrOBhKXLAPE2QYbCS4ZY9bGD2LRM+mIyD8AR+IAAVqQ0AAhW4PJ7EQwAGQ8cDBJUhuWwtmQ13ErUI0CuGGwHhgAHULjB6MgABwABmW5IgQipLQwBPJRh0mOWFQAjkkXBVAaYQViIcshPoXB8vpDSG9JABFJIQeDSoTLGDiGlROnIABMepaLkIbwAwhB9FKUFBoAKQEkhLJxMdwUIZjMgA)
```sh
--parser babylon
```

**Input:**
```jsx
export class SearchStore {
  @observable searchText = ''
  @observable message = ''
  @observable waiting = false
  @observable selectedId = null
  @observable expanded = false

  results = observable.array()

  @computed
  get selected() {}
}
```

**Output:**
```jsx
export class SearchStore {
  @observable
  searchText = "";
  @observable
  message = "";
  @observable
  waiting = false;
  @observable
  selectedId = null;
  @observable
  expanded = false;

  results = observable.array();

  @computed
  get selected() {}
}

```

**Expected behavior:**
1.14 introduced "all decorators are on separate line" behaviour, which is a good thing when decorating functions. 

However, when used in front of member variables, they should be on the same line. This is an extremely common use case for MobX for example. 1.14 simply makes MobX code unreadable.

duailibe:
Implements @lydell's suggestions: https://github.com/prettier/prettier/issues/4924#issuecomment-425659254

Closes #4924
Closes #5088

What this PR does is select the separator for each decorator by checking:
  - is there a new line between the start of the first decorator and the start of the last decorator
  - is there a new line right after the end of the last decorator

If either of those are true, we should use a `hardline` separator, otherwise it's a `line`. If they are separated with `line` they will be the in the same line, unless they don't fit, then they'll break; `hardline` will force a break.

"Breaking", in this case, means each decorator is printed in a separate line and the thing they decorate too.

I'll navigate to the threads now to see if I can't find any case not covered in our tests.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5294,6 +5294,9 @@ function printAssignmentRight(leftNode, rightNode, printedRight, options) {
       isBinaryish(rightNode.test) &&
       !shouldInlineLogicalExpression(rightNode.test)) ||
     rightNode.type === "StringLiteralTypeAnnotation" ||
+    (rightNode.type === "ClassExpression" &&
+      rightNode.decorators &&
+      rightNode.decorators.length) ||
     ((leftNode.type === "Identifier" ||
       isStringLiteral(leftNode) ||
       leftNode.type === "MemberExpression") &&
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -96,10 +96,11 @@ function genericPrint(path, options, printPath, args) {
     !getParentExportDeclaration(path)
   ) {
     const shouldBreak =
+      node.type === "ClassDeclaration" ||
       hasNewlineInRange(
         options.originalText,
         options.locStart(node.decorators[0]),
-        options.locStart(getLast(node.decorators))
+        options.locEnd(getLast(node.decorators))
       ) ||
       hasNewline(
         options.originalText,
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -20,8 +20,7 @@ const {
   getPenultimate,
   startsWithNoLookaheadToken,
   getIndentSize,
-  matchAncestorTypes,
-  isWithinParentArrayProperty
+  matchAncestorTypes
 } = require("../common/util");
 const {
   isNextLineEmpty,
@@ -96,11 +95,18 @@ function genericPrint(path, options, printPath, args) {
     // responsible for printing node.decorators.
     !getParentExportDeclaration(path)
   ) {
-    const separator =
-      node.decorators.length === 1 &&
-      isWithinParentArrayProperty(path, "params")
-        ? line
-        : hardline;
+    const shouldBreak =
+      hasNewlineInRange(
+        options.originalText,
+        options.locStart(node.decorators[0]),
+        options.locStart(getLast(node.decorators))
+      ) ||
+      hasNewline(
+        options.originalText,
+        options.locEnd(getLast(node.decorators))
+      );
+
+    const separator = shouldBreak ? hardline : line;
 
     path.each(decoratorPath => {
       let decorator = decoratorPath.getValue();
--- a/tests/decorators-ts/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/decorators-ts/__snapshots__/jsfmt.spec.js.snap
@@ -161,8 +161,7 @@ class Greeter {
 }
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 class Greeter {
-  @format("Hello, %s")
-  greeting: string;
+  @format("Hello, %s") greeting: string;
 
   constructor(message: string) {
     this.greeting = message;
--- a/tests/decorators-ts/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/decorators-ts/__snapshots__/jsfmt.spec.js.snap
@@ -41,6 +41,27 @@ class Point {
 
 `;
 
+exports[`angular.ts - typescript-verify 1`] = `
+@Component({
+  selector: 'toh-hero-button',
+  template: \`<button>{{label}}</button>\`
+})
+export class HeroButtonComponent {
+  @Output() change = new EventEmitter<any>();
+  @Input() label: string;
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+@Component({
+  selector: "toh-hero-button",
+  template: \`<button>{{label}}</button>\`
+})
+export class HeroButtonComponent {
+  @Output() change = new EventEmitter<any>();
+  @Input() label: string;
+}
+
+`;
+
 exports[`class-decorator.ts - typescript-verify 1`] = `
 @sealed
 class Greeter {
--- a/tests/decorators-ts/angular.ts
+++ b/tests/decorators-ts/angular.ts
@@ -1,0 +1,8 @@
+@Component({
+  selector: 'toh-hero-button',
+  template: `<button>{{label}}</button>`
+})
+export class HeroButtonComponent {
+  @Output() change = new EventEmitter<any>();
+  @Input() label: string;
+}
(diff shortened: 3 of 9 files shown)


duailibe:
I believe our current test suite covers all the cases, this is ready for review

lydell:
Looks good!

I played around in the playground and … I'm not sure what I found.

**Prettier pr-5188**
[Playground link](https://deploy-preview-5188--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEB6VACAkgZwzACwEs8ioAbMuDCAJwwFsBXcmIyqOAfgB0oABAGYQIACj4YMARj4BKDGHIBDHHgCCGYAF8+fdBgDKBCCwAm+AnBzUARtUhQcJeLF4DhEDPxtL6cAB4ADnQwGKZwgkosoYoqeABCmjruIl4+fkEhCsqqGADCSXxCqd6+2XEYACJJGPrC9JAMgb4k0LpQ+kYm5OYA7tRwAG4IGPVg1Mys7FSjdOW5SlDmgkxQYGzQGABU4bG0ShuOW3qYojZMoVAQoQxwhBCmOAA0GIG0EIFwtGxWGIvmzX2DBwsi4ICeIA+hxwyFAvnevQACr4EDCUEpBhAiKZwSAbPswABrO4GZpgMgAc2QMFoTDgELI1m+iP2FIYSmQkXI1ghACscP54gTiTADEpbgAZKicpTc+kgQFM5B4pQ2ACe5GguLeZBgAHVsYRkAAOAAMELeEGsev2gWVbysX2GuNocAAjkwiK6WUo2RykFyeSBrAwiNTafKnFAKeQ4ABFJjXOAyuUQmCqg2mI1IABMaf202jeQgDHZyqunFxTGsABVVWjA3AtFogA)
```sh
--parser babylon
```

**Input:**
```jsx
// Is this inline or multiline?
@foo(
  1
) class A {}

// Should these be consistent?
@foo @bar export default class B {}
@foo @bar export class C {}
@foo @bar class D {} // for comparison

// Should we even force multiline for class and function *declarations*
// (but not methods, properties and params)?
```

**Output:**
```jsx
// Is this inline or multiline?
@foo(1) class A {}

// Should these be consistent?
@foo
@bar
export default class B {}
@foo
@bar
export class C {}
@foo @bar class D {} // for comparison

// Should we even force multiline for class and function *declarations*
// (but not methods, properties and params)?

```

duailibe:
```js
// Is this inline or multiline?
@foo(
  1
) class A {}
```
My intuition is that it should be multiline (i.e. that's a bug in the implementation), but I'm not sure.

```js
// Should these be consistent?
@foo @bar export default class B {}
@foo @bar export class C {}
@foo @bar class D {} // for comparison
```
I think so, but I'm not sure.

```js
// Should we even force multiline for class and function *declarations*
// (but not methods, properties and params)?
```
That sounds reasonable to me. And this fixes the above case, as well.

---

I'll change it 👍

lydell:
Nice! What about _function_ declarations in addition to _class_ declarations?

duailibe:
@lydell I don't think function declarations can be decorated. https://tc39.github.io/proposal-decorators/#sec-updated-syntax

Am I missing something?

lydell:
> I don't think function declarations can be decorated.

Oops! Sorry, you're right.

> Also, what about class expressions?

Cool, I didn't know that was valid. Currently it looks a bit weird when multiline:

**Prettier pr-5188**
[Playground link](https://deploy-preview-5188--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzCFMC8mAAgCZyQA6UYANgIZpqbAC+IANCBAA4wCW6ZKHoAnURADuABTEI0yEPQBuEfqU4gARqPpgA1nBgBlHnv5QA5shiiArnC4W0cUTGm7LAW3rJs9WhcuACs0AA8AIV0DI2N6LzgAGQs4PwCgkDNRF1FFLXotAE9aaE0eUQsYAHV1GAALZAAOAAYucogXKt0eRXK4HOVUrlE4AEc7fhGPem9fJH9AxxAXL34beyW0C0taOABFOwh4NMWuGAKa0nrkACYz3X5abYBhCC8fRShoIZA7FwAVAoKebpOCsVhAA)
```sh
--parser babylon
```

**Input:**
```jsx
const foo = @deco
class {}
```

**Output:**
```jsx
const foo = @deco
class {};

```

Maybe we should _not_ look at the source for those?

duailibe:
We could not look but then we could have a `@deco(verylongname)` that would break either way. I guess we just need to print it better when breaking

```js
const foo = 
  @deco
  class {}
```

What about this?

lydell:
@duailibe Sounds good :+1:

Issue #5170, opened 2018-09-30
Compress objects


danielreichhart:
**Prettier 1.14.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEATAhjXTAXkwG0AdKTTYS66gW2hgAslNyQApXKTgGjr0AKhADWATwjsA7ADoADIKr0AMtGzR2AZjkBOIQF9l1WisbM2HEADE4AIwFDqoydMwA2fSbUatmABY5ACYjHzN6TCZYK04AWVwAJydzTFcpdj05AFYfanUoTSh2bLkZI0oAXQBuShB+EAgABxgAS3RkUCTEiAB3AAUkhDRkEFwANwhW7HqQe0TcMDE4GABlJsXWqABzZBhEgFc4Bq20OESYfoXthlxkADNcABszhoArNAAPACEFpZXVrgGHBVFs4A9nq8QBtEmdEqN7Lh7BIntBZk1ElsYAB1aasZAADiU0J6Z2xCyaowxcDh43BDUScAAjgdWoyrrgbnckI8XscQGcGK09od+WgttsnnAAIoHCDwCF8hoEey47D4pDBZULVpPCUAYQgDFuoyg0HpIAOZ2ESJGPMhcEMhiAA)
```sh
--parser babylon
```

**Input:**
```jsx
const data = [
  {
    month: "Jan",
    Tokyo: 7.0,
    London: 3.9
  },
  {
    month: "Feb",
    Tokyo: 6.9,
    London: 4.2
  },
  {
    month: "Mar",
    Tokyo: 9.5,
    London: 5.7
  }
];

```

**Output:**
```jsx
const data = [
  {
    month: "Jan",
    Tokyo: 7.0,
    London: 3.9
  },
  {
    month: "Feb",
    Tokyo: 6.9,
    London: 4.2
  },
  {
    month: "Mar",
    Tokyo: 9.5,
    London: 5.7
  }
];

```

**Expected behavior:**
```jsx
const data = [
  { month: "Jan", Tokyo: 7.0, London: 3.9 },
  { month: "Feb", Tokyo: 6.9, London: 4 },
  { month: "Mar", Tokyo: 9.5, London: 5 }
];


```

lydell:
If there's a newline anywhere inside an object literal, it will stay multiline even if it would fit in a single line. That's not ideal, but we couldn't find any good heuristic that works better. See:

[...]

PR #5205, opened 2018-10-08
Improved object-break heuristic

j-f1:
This changes the heuristic from breaking the object if it contains a newline somewhere inside it to only breaking if the newline is between the `{` and the first key. This allows users to easily collapse objects by deleting that linebreak without having to manually move the object onto one line.


j-f1:
Ref #5170.

lydell:
I kind of like the shortcut aspect of it, but I'm afraid this will be harder to explain/teach.

I think it's pretty easy to select the object and press the "join lines" keyboard shortcut most editors have (such as `ctrl+j`, or `J` in Vim).

suchipi:
I'm in favor of this change, better DX for sure :+1:

zackkrida:
Could this heuristic also be used for _destructuring_ objects? Makes sense to have that consistent to me. 

Would help deal with issues like the discussion at #4267

j-f1:
@zackkrida since we don’t have any heuristic for destructured objects, it’s unlikely ATM.

j-f1:
Actually, this might happen: https://github.com/prettier/prettier/issues/2550#issuecomment-437397749

PR #2630, opened 2017-08-16
Switch to updated decorators plugin in Babylon

existentialism:
Fixes https://github.com/prettier/prettier/issues/2628

This probably needs to be a breaking change as users using the old syntax w/ Babylon will now see errors, primarily:

```js
@foo
export default class {}
```

Now needs to be:

```js
export default @foo class{}
```

PR #5207, opened 2018-10-08
Respect original text decorator order

### LINKED ISSUE #4795: export default is moved after decorator
**Prettier 1.13.7**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAdKcAeAHCAnGAAgBM4AzAQwFcAbIgAUigzBgApgBfASnTBooBnQYQCimCgFtsNOISzwoxEQCU4FVgDoAwhGnQERYOkKE8CUnkIBeQm27WAfISi0a6TiAA0ICNhgAltCCyKAUeHgQAO4ACuEIISgUAG4QAcTeIABGeBoA1nAwAMrYGgFQAObIMHhUcD7lgnAEMbkVkhTIlDRNPgBWgpgAQrlgBcVScAAy5XBdFD31IKV4TXjI2RRZAJ400JnYeOUwAOrpMAAWyAAcAAw+hxBNJ7nYG4dwa8lzPuYAjlQAuZWhR2p0kN1eiAmpIAtVaktBOUKrIAIpUCDweaLHwwLZnYiXZAAJlxuQCNGRukkHQ2UAMmSoTQAKltEpC4JxOEA)
```sh
--parser babylon
```

**Input:**
```jsx

export default @connect({})
class Example extends React.Component {
  render = ()=> null
}
```

**Output:**
```jsx
@connect({})
export default class Example extends React.Component {
  render = () => null;
}

```

**Expected behavior:**
It should keep export default where I put it.

### LINKED ISSUE #5045: Support new decorators proposal
https://github.com/babel/babel-eslint/releases/tag/v9.0.0

**Input:**
```jsx
import React, { PureComponent } from 'react';

import { connect } from 'dva';

export default
@connect(({ list }) => ({
  list,
}))
class Center extends PureComponent {
  render() {
    return <div>123</div>
  }
}

```

**Output:**
```jsx
import React, { PureComponent } from 'react';

import { connect } from 'dva';

@connect(({ list }) => ({
  list,
}))
export default class Center extends PureComponent {
  render() {
    return <div>123</div>;
  }
}

```

**Expected behavior:**
Using the export keyword between a decorator and a class is not allowed. Please use `export @dec class` instead.

prettier/prettier-vscode#572

duailibe:
Closes #5045
Closes #4795
Supersedes #2630

Hopefully this will enable to continue using Prettier without too much problems until the spec is standardized

### DIFF
--- a/src/language-js/parser-babylon.js
+++ b/src/language-js/parser-babylon.js
@@ -44,28 +44,22 @@ function parse(text, parsers, opts) {
   const babylon = require("@babel/parser");
 
   const combinations = [
-    babylonOptions({}, [["decorators", { decoratorsBeforeExport: false }]]),
-    babylonOptions({ strictMode: false }, [
+    babylonOptions({ strictMode: true }, ["decorators-legacy"]),
+    babylonOptions({ strictMode: false }, ["decorators-legacy"]),
+    babylonOptions({ strictMode: true }, [
       ["decorators", { decoratorsBeforeExport: false }]
     ]),
-    babylonOptions({}, ["decorators", { decoratorsBeforeExport: true }]),
     babylonOptions({ strictMode: false }, [
-      ["decorators", { decoratorsBeforeExport: true }]
-    ]),
-    babylonOptions({}, ["decorators-legacy"]),
-    babylonOptions({ strictMode: false }, ["decorators-legacy"])
+      ["decorators", { decoratorsBeforeExport: false }]
+    ])
   ];
 
   const parseMethod =
     !opts || opts.parser === "babylon" ? "parse" : "parseExpression";
 
-  function parseFn(options) {
-    return babylon[parseMethod](text, options);
-  }
-
   let ast;
   try {
-    ast = tryCombinations(parseFn, combinations);
+    ast = tryCombinations(babylon[parseMethod].bind(null, text), combinations);
   } catch (error) {
     throw createError(
       // babel error prints (l:c) with cols that are zero indexed
--- a/src/language-js/parser-babylon.js
+++ b/src/language-js/parser-babylon.js
@@ -4,69 +4,97 @@ const createError = require("../common/parser-create-error");
 const hasPragma = require("./pragma").hasPragma;
 const locFns = require("./loc");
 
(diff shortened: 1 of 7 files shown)


vjeux:
I don't have the time to do a proper review but I like what it does.

src/language-js/parser-babylon.js
```diff
@@ -4,71 +4,99 @@ const createError = require("../common/parser-create-error");
...
-      "optionalCatchBinding",
-      "optionalChaining",
-      "classPrivateProperties",
-      ["pipelineOperator", { proposal: "minimal" }],
-      "nullishCoalescingOperator",
-      "bigInt",
-      "throwExpressions"
-    ]
-  };
+  const combinations = [
```
ikatyang: We probably need a `guess` function (similar to `isProbablyJsx` in `parser-typescript.js`) to try not to do unnecessary parsing.

tests/typescript_decorators/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -49,8 +49,7 @@ export class TabCompletionController {}
   selector: "angular-component"
 })
 class AngularComponent {
-  @Input()
-  myInput: string;
+  @Input() myInput: string;
```
ikatyang: This looks like the changes from #5188, rebasing issue?

src/language-js/parser-babylon.js
duailibe: I thought about it, trying to find a `@deco` before/after an export

tests/typescript_decorators/__snapshots__/jsfmt.spec.js.snap
duailibe: Oh probably

duailibe:
I'm making the call and merging this. We can fix any bugs that show up after the release. Will try a solution for the "guess decorator before or after export" in a follow-up PR if nobody beats me to it

existentialism:
@duailibe sorry for the delay, been an insane week. Agree with rolling with this and fixing any bugs.

PR #5261, opened 2018-10-13
[ImgBot] optimizes images

boidolr:
Reduce the image size by adding the app ImageBot:
Total -- 1,155.42kb -> 742.94kb (35.7%)
These are similar to the existing optimizer for SVG, only for PNG. If you are interested one could also either add a module so this task can be handled as a build step, or via adding [ImageBot](https://imgbot.net/) to the repository directly.


duailibe:
Hi @boidolr, thanks for doing this. I think this would be better as a step of the site build. Are you interested in doing that? :-)

duailibe:
Either way, this looks good already. Thanks

boidolr:
@duailibe if you can give me some guidance on how it should be integrated, sure.
Executing these steps for all images at once is rather slow, so not something one would want to do on every commit for example. Even for single images it can be annoying to have to wait for the optimization to finish.
That is why I think a solution such as the bot is preferable.

Issue #4937, opened 2018-08-03
prettier v1.4 does not work on styled-components


sevenryze:
**Prettier 1.14.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAgCp8BDAZ3wCU4SwCAzHCdfAHRBxrvYG5WoM2PEVIVqtGABEA8gFl8jZmw5cYAWgAmzXv0G4CAQUyYFTFuwB0AehLGdArPvzB8qKACs4dAOIAbCABGJL74AL6mSuxkMACevnAaapCOUAgwZPa6Hl4wfoHBAAb8+CWlARAaMc7FpWxQJawwjc1NMOgkOADmbkj4AAx8UPw1mCQaGm6dvQPDQ-UK0Or0JOiovjG9ZCRQZGpkcDio9IOh-AWD-OJ0MrIWnFAaBwAUADxGJlYAfAA0+FpgAFd0GkLJ04DAAKLxYGwABCMQAkhonuwmBAmiAAJSYwYgb4gCCYGCoaBkZCgDpMADuAAUOggySgSAA3CCoDR4kABHC0ADW4IAyqMwJNkDAcAC4Pi3Ps8DSeZ12shlr59vj3GQAB6wnlgfkwAUrOAAGTccGVwTVIFGOFlyC5JACcWgnMwh1gAHV2TAABbIAAcfXxbog+w9PMw9rdcFlzPN+M4AEcAahOPKSIqSBbVVKQPtVmKJbmyJN4gBFAHo81IFVWmCOr0aX3IABM+PFJDWkwAwswlSgoNB4yAAfsACqOxm1uChUJAA)
```sh
--parser babylon
```

**Input:**
```jsx
import * as React from "react";
import * as ReactDOM from "react-dom";
import App from "./app";
import { injectGlobal } from "styled-components";

injectGlobal`
      body {
     
  					margin: 0;


  padding: 0;


  font-family: sans-serif;
}
`;

ReactDOM.render(<App />, document.getElementById("root"));

```

**Output:**
```jsx
import * as React from "react";
import * as ReactDOM from "react-dom";
import App from "./app";
import { injectGlobal } from "styled-components";

injectGlobal`
      body {
     
  					margin: 0;


  padding: 0;


  font-family: sans-serif;
}
`;

ReactDOM.render(<App />, document.getElementById("root"));

```

**Expected behavior:**
Format the above tagged template.

[...]

Issue #2970, opened 2017-10-03
Support injectGlobal as css templateLiteral


brikou:
injectGlobal is a templateLiteral used by emotion (and maybe more css in js lib?). Prettier should treat it as css template literal.

Great work, can't imagine not using prettier anymore... :+1: 

**Prettier 1.7.2**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22export%20default%20()%20%3D%3E%20injectGlobal%60%5Cn*%2C%20*%3A%3Abefore%2C%20*%3A%3Aafter%20%7B%5Cnbox-sizing%3A%20inherit%3B%20%7D%5Cn%5Cn%40-ms-viewport%20%7B%20width%3A%20device-width%3B%20%7D%5Cn%60%3B%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Afalse%7D%7D)

**Input:**
```jsx
export default () => injectGlobal`
*, *::before, *::after {
box-sizing: inherit; }

@-ms-viewport { width: device-width; }
`;
```

**Output:**
```jsx
export default () => injectGlobal`
*, *::before, *::after {
box-sizing: inherit; }

@-ms-viewport { width: device-width; }
`;

```

**Expected behavior:**

```jsx
export default () => injectGlobal`
  *,
  *::before,
  *::after {
    box-sizing: inherit;
  }

  @-ms-viewport {
    width: device-width;
  }
`;
```

azz:
We actually briefly supported `injectGlobal` but it was removed because it is specific to one library, see https://github.com/prettier/prettier/pull/2337#issuecomment-311803127

We _might_ look at making this mapping configurable in the future.

kripod:
Are there any news about supporting this behavior? Making the mapping configurable would definitely help.

azz:
Still needs a proposal.

PR #2337, opened 2017-06-28
Add support for css, keyframes and injectGlobal keywords

zanza00:
This is my fix for #2330 

I also added the keyword `keyframes` that is used in the [same way](https://www.styled-components.com/docs/api#helpers) as `css`

The AST for them is slightly different since they are not `MemberExpression` but simple `Identifier`

PR #5264, opened 2018-10-14
fix: detect more styled template literals

SimenB:
Styled components has a couple more template literals than just `css` (see https://www.styled-components.com/docs/api). This detects them


lydell:
Hi!

It appears as if Prettier used to support more tags, but removed them:

- https://github.com/prettier/prettier/pull/2337#issuecomment-311801382
- #2342
- #2970
- #4937 

So I guess we should not merge this? Or is this PR bringing anything new to the table?

PR #5266, opened 2018-10-14
Update husky integration code for husky v1.0.0

GillesDebunne:
Latest husky (version 1.0.0) moved its hooks config from the `scripts` field to a new `husky` field in `package.json` ([ChangeLog](https://github.com/typicode/husky/blob/master/CHANGELOG.md)).

The `scripts` syntax described here, although shorter, will be deprecated soon. This change replaces it with the new husky syntax.

The new syntax is already used in [precommit.md](https://github.com/prettier/prettier/blob/master/docs/precommit.md).

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -301,8 +301,8 @@ class GetStartedSection extends React.Component {
                   Then add this config to <code>package.json</code>:
                   <MarkdownBlock>
                     {json({
-                      "husky": {
-                        "hooks": {
+                      husky: {
+                        hooks: {
                           "pre-commit": "pretty-quick --staged"
                         }
                       }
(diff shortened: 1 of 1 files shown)


website/pages/en/index.js
```diff
@@ -301,8 +301,8 @@ class GetStartedSection extends React.Component {
                   Then add this config to <code>package.json</code>:
                   <MarkdownBlock>
                     {json({
-                      "husky": {
-                        "hooks": {
+                      husky: {
+                        hooks: {
```
lydell: > No longer a valid JSON?

This file is written in JavaScript, not JSON, so no worries.

GillesDebunne: Indeed, but this is supposed to be readily copy-pasted in `package.json`.

Omitting the double quotes leads to an error message on my machine:
`JSONError: Unexpected token h in JSON`

GillesDebunne: You're right. I now realize there's a formatter involved.

This solves this format discussion.

Shall I squash my commits?

lydell: No, we use “Squash and merge” for all PRs here so there’s no need to.

ikatyang:
Thanks!

Issue #5098, opened 2018-09-15
HTML TODOs


ikatyang:
This issue tracks HTML issues that should be done before the public release.

---

# Please take a look at #5259, which should address this issue.

---

Try out the current experimental HTML support:

```
yarn add prettier/prettier
yarn prettier filename.html 
```

[...]

PR #5274, opened 2018-10-15
Remove HTML from pre-commit hooks until available

acdha:
The pre-commit hook registers for HTML files but that will always fail until #5098 is completed. This means that anyone following [the instructions](https://prettier.io/docs/en/precommit.html#option-3-pre-commit-https-githubcom-pre-commit-pre-commit) will have the hook fail any time they attempt to commit an HTML file.


ikatyang:
Thanks!

Issue #719, opened 2017-02-15
Flow Typecast Comments


jacobp100:
In flow, you can typecast with the following,

```js
(obj: Class).property
```

You can also do this with comments,

```js
(obj /*: Class */).property
```

However, this gets transformed into,

```js
obj /*: Class */.property
```

Both are syntactically valid, but only one is recognised by flow. Maybe we'd just want to check comments that start with `:`?

Happy to help if you decide what direction we should go!

# Edit

Also,

```js
const x = (input /*: string */) /*: string */ => {};
// to
const x = (input /*: string */ /*: string */) => {};
```

vjeux:
Thanks for the report, we need to fix it!

vjeux:
@jacobp100 https://github.com/jlongster/prettier/pull/722

Do you mind running this pull request and try to find crazy annotations that wouldn't work with it? Thanks!

jacobp100:
Seemed to work!

vjeux:
Another idea: use the flow parser to parse comments and have a way to output flow types as comments instead of the ast version.

jokeyrhyme:
Related to #204 ?

jokeyrhyme:
I've just encountered the Flow comment syntax issues with functions

Here are a few more test cases, I hope these are helpful

https://prettier.github.io v0.22.0

```js
// input
function fnNoArgs () /* : Promise<void> */ {
  return Promise.resolve()
}

// output
function fnNoArgs(/* : Promise<void> */) {
  return Promise.resolve();
}
```

```js
// input
function fnOneArg (a /* : string */) /* : Promise<string> */ {
  return Promise.resolve(a)
}

// output
function fnOneArg(a /* : string */) /* : Promise<string> */ {
  return Promise.resolve(a);
}
```

```js
// input
const arrowNoArgs = () /* : Promise<void> */ => {
  return Promise.resolve()
}

// output
const arrowNoArgs = (/* : Promise<void> */) => {
  return Promise.resolve();
};
```

```js
// input
const arrowOneArge = (a /* : string */) /* : Promise<string> */ => {
  return Promise.resolve(a)
}

// output
const arrowOneArge = (a /* : string */ /* : Promise<string> */) => {
  return Promise.resolve(a);
};
```

jokeyrhyme:
Awesome work! It looks like most of the above has been resolved.

[...]

PR #5280, opened 2018-10-19
Add support for flow typecast comments

swac:
This attempts to fix the bug described in #719. Previously, Prettier was removing parentheses in expressions like: 

```
new (window.Request /*: Class<Request> */)("")
```

turning them into:

```
new window.Request /*: Class<Request> */("")
```

which no longer satisfies Flow's syntax. With this change, we detect these typecast expressions and attempt to keep the parentheses intact in such cases.

This is my first time contributing to Prettier, so I'm probably making some egregious mistakes!

### DIFF
--- a/src/language-js/flow-comments.js
+++ b/src/language-js/flow-comments.js
@@ -1,0 +1,15 @@
+"use strict";
+
+function hasFlowShorthandAnnotationComment(node) {
+  // https://flow.org/en/docs/types/comments/
+  // Syntax example: const r = new (window.Request /*: Class<Request> */)("");
+
+  return (
+    node.extra &&
+    node.extra.parenthesized &&
+    !!node.trailingComments &&
+    node.trailingComments[0].value.startsWith(":")
+  );
+}
+
+module.exports = hasFlowShorthandAnnotationComment;
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -158,10 +158,9 @@ function genericPrint(path, options, printPath, args) {
       parts.push(node.trailingComments[0].value.trimLeft());
       parts.push("*/");
       node.trailingComments[0].printed = true;
-      parts.push(")");
-    } else {
-      parts.push(")");
     }
+
+    parts.push(")");
   }
 
   if (decorators.length > 0) {
@@ -2834,9 +2833,11 @@ function printPathNoParens(path, options, print, args) {
       return concat([
         "(",
         path.call(print, "expression"),
-        commentSyntax ? " /*: " : ": ",
+        commentSyntax ? " /*" : "",
+        ": ",
         path.call(print, "typeAnnotation"),
-        commentSyntax ? " */)" : ")"
+        commentSyntax ? " */" : "",
+        ")",
       ]);
     }
 
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -32,9 +32,9 @@ const embed = require("./embed");
 const clean = require("./clean");
 const insertPragma = require("./pragma").insertPragma;
 const handleComments = require("./comments");
-const hasFlowShorthandAnnotationComment = require("./flow-comments");
 const pathNeedsParens = require("./needs-parens");
 const preprocess = require("./preprocess");
+const { hasFlowShorthandAnnotationComment } = require("./utils");
 
 const {
   builders: {
@@ -2823,7 +2823,9 @@ function printPathNoParens(path, options, print, args) {
         value &&
         value.typeAnnotation &&
         value.typeAnnotation.range &&
-        options.originalText.substr(value.typeAnnotation.range[0], 3) === "/*:";
+        options.originalText
+          .substring(value.typeAnnotation.range[0])
+          .match(/^\/\*\s*:/);
       return concat([
         "(",
         path.call(print, "expression"),
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -32,6 +32,7 @@ const embed = require("./embed");
 const clean = require("./clean");
 const insertPragma = require("./pragma").insertPragma;
 const handleComments = require("./comments");
+const hasFlowShorthandAnnotationComment = require("./flow-comments");
 const pathNeedsParens = require("./needs-parens");
 const preprocess = require("./preprocess");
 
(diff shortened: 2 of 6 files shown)


src/language-js/flow-comments.js
```diff
@@ -0,0 +1,15 @@
...
+
+  return (
+    node.extra &&
+    node.extra.parenthesized &&
+    !!node.trailingComments &&
+    node.trailingComments[0].value.startsWith(":")
+  );
+}
+
+module.exports = hasFlowShorthandAnnotationComment;
```
j-f1: Can you rename this file to `utils.js` and export this as an object (i.e. `{ hasFlowShorthandAnnotationComment }`)?

src/language-js/flow-comments.js
```diff
@@ -0,0 +1,15 @@
...
+
+function hasFlowShorthandAnnotationComment(node) {
+  // https://flow.org/en/docs/types/comments/
+  // Syntax example: const r = new (window.Request /*: Class<Request> */)("");
+
+  return (
+    node.extra &&
+    node.extra.parenthesized &&
+    !!node.trailingComments &&
+    node.trailingComments[0].value.startsWith(":")
```
j-f1: Can the comment be formatted as `/* : foo */`? If so, perhaps use `.value.trimLeft()`?

src/language-js/flow-comments.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+function hasFlowShorthandAnnotationComment(node) {
+  // https://flow.org/en/docs/types/comments/
+  // Syntax example: const r = new (window.Request /*: Class<Request> */)("");
+
+  return (
+    node.extra &&
+    node.extra.parenthesized &&
+    !!node.trailingComments &&
```
j-f1: Can this just be `node.trailingComments`?

src/language-js/printer-estree.js
```diff
@@ -2802,14 +2812,27 @@ function printPathNoParens(path, options, print, args) {
...
+      // Flow supports a comment syntax for specifying type annotations: https://flow.org/en/docs/types/comments/.
+      // Unfortunately, its parser doesn't differentiate between comment annotations and regular
+      // annotations when producing an AST. So to preserve parentheses around type casts that use
+      // the comment syntax, we need to hackily read the source itself to see if the code contains
+      // a type annotation comment.
+      const commentSyntax =
+        value &&
+        value.typeAnnotation &&
+        value.typeAnnotation.range &&
+        options.originalText.substr(value.typeAnnotation.range[0], 3) === "/*:";
```
j-f1: Similar question to the one in `flow-comments.js` — should this be checking against the regex `/^/\*\s*:/`?

src/language-js/flow-comments.js
swac: Good point, it seems whitespace between `/*` and `:` is indeed allowed.

lydell:
We need to be careful with that whitespace trimming though, because newlines do not seem to be supported. A goal of Prettier is to never break your code. Here's a case where it could break: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVxhgIIGMAuArgIYxj4CeADgKZjEB2Dc+x+AlnAwFyoBuxAE70wobmAaEAtgCMawlGAC8YAKzpMYAHIsRlWvSYs2nHvyFgZokKnEB+SbPnIsK9ZoAqgimEE0pcHzsDADm5AAWdAw0CDDBdHJQcH4RdLhw8AxggMgEYBRwhADkMGQhNPh61HTygskAhOgCwrjKIgA8lkA

src/language-js/utils.js
```diff
@@ -8,7 +8,15 @@ function hasFlowShorthandAnnotationComment(node) {
...
-    node.trailingComments[0].value.trimLeft().startsWith(":")
+    // We intentionally only match spaces and tabs here instead of any whitespace because
+    // Flow annotation comments cannot be split across lines. For example:
+    //
+    // (this /*
+    // : any */).foo = 5;
+    //
+    // is not picked up by Flow, so removing the newline would create a type annotation
+    // that the user did not intend to create.
+    node.trailingComments[0].value.match(/^[ \t]*:/)
```
lydell: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) `\s` is equivalent to:

```
 \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff
```

I just tried it, and Flow allows any of those except the JS line terminators `\r`, `\n`, `\u2028` and `\u2029`.

This regex might do the trick: `/^(?:(?=.)\s)*:/`

src/language-js/utils.js
```diff
@@ -0,0 +1,25 @@
...
+
+function hasFlowShorthandAnnotationComment(node) {
+  // https://flow.org/en/docs/types/comments/
+  // Syntax example: const r = new (window.Request /*: Class<Request> */)("");
+
+  return (
+    node.extra &&
+    node.extra.parenthesized &&
+    node.trailingComments &&
+    // We intentionally only match spaces and tabs here instead of any whitespace because
```
lydell: `We match any whitespace except line terminators because`

src/language-js/printer-estree.js
```diff
@@ -151,7 +152,16 @@ function genericPrint(path, options, printPath, args) {
...
-    parts.push(")");
+    const node = path.getValue();
+    if (hasFlowShorthandAnnotationComment(node)) {
+      parts.push(" /*");
+      parts.push(node.trailingComments[0].value.trimLeft());
+      parts.push("*/");
+      node.trailingComments[0].printed = true;
+      parts.push(")");
+    } else {
+      parts.push(")");
```
j-f1: Can this line be pulled out of both blocks and placed after the `if` statement?

src/language-js/printer-estree.js
```diff
@@ -2802,14 +2812,34 @@ function printPathNoParens(path, options, print, args) {
...
+          .substring(value.typeAnnotation.range[0])
+          .match(/^\/\*\s*:/);
       return concat([
         "(",
         path.call(print, "expression"),
-        ": ",
+        commentSyntax ? " /*: " : ": ",
         path.call(print, "typeAnnotation"),
-        ")"
+        commentSyntax ? " */)" : ")"
```
j-f1: Similar thing here: how about splitting these two lines into an optional comment character and then the required `:` and `)`?

src/language-js/utils.js
```diff
@@ -0,0 +1,25 @@
...
+    node.extra.parenthesized &&
+    node.trailingComments &&
+    // We match any whitespace except line terminators because
+    // Flow annotation comments cannot be split across lines. For example:
+    //
+    // (this /*
+    // : any */).foo = 5;
+    //
+    // is not picked up by Flow, so removing the newline would create a type annotation
+    // that the user did not intend to create.
```
j-f1: Do you want to reference https://github.com/facebook/flow/issues/7050 here?

j-f1:
Looks like you’ve just got one lint error left.

swac:
Thanks so much for reviewing! It looks like the commits in master are squashed, so I don't need to squash my commits in this branch, correct?

j-f1:
That’s correct! I’ll merge tomorrow if nobody has any objections.

PR #5290, opened 2018-10-24
Fix Flow generic comment positioning

swac:
This fixes a bug in Prettier where Flow generic comments get repositioned, breaking Flow's syntax. An example:

```
foo /*:: <bar> */ (baz);
```

Currently, Prettier will rewrite this ([source](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzCEAEB6AVEpTAHgCMBDAJwD5NdtMAKcgLwEoBuEAGhAgAcYAS2gBnZKEoUIAdwAKlBGJRkAbhEEATbiBIUyYANZwYAZT77BUAObIYFAK5welkXAoxZeqwFsyyVGQANq48AFYiAB4AQnqGxiZk3nAAMpZw-kEhIOYUrhTIOmQkAJ6B0Np8FJYwAOqaMAAWyAAcAAw8lRCuNXp8BZVweSrpPBRwAI72gmOeZD5+SAHBTiCu3oK2DisillaBcACK9hDwGcs8MEV1Go3IAEwXeoKBuwDCEN6+BVDQIyD2rgAKkUlEtXABfcFAA)) as:

```
foo(/*:: <bar> */ baz);
```

This PR attempts to fix this behavior. One issue that I ran into was dealing with the Flow parser's AST that it generates from the comment syntax. Code like this:

```
/*:: type Props = {
  foo?: ?string,
  bar: number,
}; */
```

gets turned into this ([source](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6AVEpACGBPABzmwAUAnCAgZ2wF5tgAdKbbAMwggH4cuqYyASygBzADTNWAIwCGZHFACuAWylwyEqAF8A3NnSoQYkJRiDoVZKDkUA7iTkJLKGQDcIggCZGQUsjLAAazgYAGUCAOERZAFFOGNhKnUYchkRZRlkNhkAGyTjACsqAA8AIX8gkNCZZTgAGWE4LNz8kAiyJLJkEDYciFsfAiFYAHUvGAALZAAOAAZjIYgkkf8CbqG4Ttcm4zI4AEdFQT3U9MykbLz4kCTlQRiyOOMqKJy4AEVFCHhmq+MYGRSMaeSbIABM-38ghyUQAwhBlBlulBoDsQIokgAVQHOS5JLRaIA)):

```
type Props = {
  foo?: ?string,
  bar: number
};
```

I wasn't able to find an easy way to have Prettier reliably traverse the code backward to see if the annotation is inside a comment block. Since it's not in the AST at all ([source](https://astexplorer.net/#/gist/9fca72f6e760daf48cde1d63c6b829f6/efcab9538dd88de8ea1ac8b7980f374a2cae3a54)), the options seem limited, particularly since the comment block could have begun many lines above the type definition. For this reason, this PR only ensures the correct behavior for non-inline type definitions like these for the Babylon parser. The Flow parser's behavior here remains the same as before.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -34,7 +34,10 @@ const insertPragma = require("./pragma").insertPragma;
 const handleComments = require("./comments");
 const pathNeedsParens = require("./needs-parens");
 const preprocess = require("./preprocess");
-const { hasFlowShorthandAnnotationComment } = require("./utils");
+const {
+  hasFlowAnnotationComment,
+  hasFlowShorthandAnnotationComment
+} = require("./utils");
 
 const {
   builders: {
@@ -1115,6 +1118,19 @@ function printPathNoParens(path, options, print, args) {
         ]);
       }
 
+      // Inline Flow annotation comments following Identifiers in Call nodes need to
+      // stay with the Identifier. For example:
+      //
+      // foo /*:: <SomeGeneric> */(bar);
+      //
+      // Here, we ensure that such comments stay between the Identifier and the Callee.
+      const isIdentifierWithFlowAnnotation =
+        n.callee.type === "Identifier" &&
+        hasFlowAnnotationComment(n.callee.trailingComments);
+      if (isIdentifierWithFlowAnnotation) {
+        n.callee.trailingComments[0].printed = true;
+      }
+
       // We detect calls on member lookups and possibly print them in a
       // special chain format. See `printMemberChain` for more info.
       if (!isNew && isMemberish(n.callee)) {
@@ -1125,6 +1141,9 @@ function printPathNoParens(path, options, print, args) {
         isNew ? "new " : "",
         path.call(print, "callee"),
         optional,
+        isIdentifierWithFlowAnnotation
+          ? `/*:: ${n.callee.trailingComments[0].value.substring(2).trim()} */`
+          : "",
         printFunctionTypeParameters(path, options, print),
         printArgumentsList(path, options, print)
       ]);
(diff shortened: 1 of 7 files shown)


tests/flow_comments/generics.js
```diff
@@ -0,0 +1,9 @@
+const Component = branch/*::     <Props, ExternalProps> */(
+  ({ src }) => !src,
+  // $FlowFixMe
+  renderNothing,
+)(BaseComponent);
+
+const C = b/*:: <A> */(foo) + c/*:: <B> */(bar);
+
+foo/*::<bar>*/(baz);
```
j-f1: Here’s one of the fancy new suggestions ✨

```suggestion
foo/*::<bar>*/(baz);

foo/*::<bar>*/();
```

src/language-js/printer-estree.js
```diff
@@ -2876,7 +2895,27 @@ function printPathNoParens(path, options, print, args) {
...
+    case "TypeParameterInstantiation": {
+      const value = path.getValue();
+      const commentStart = value.range
+        ? options.originalText.substring(0, value.range[0]).lastIndexOf("/*")
+        : -1;
+      // As noted in the TypeCastExpression comments above, we're able to use a normal whitespace regex here
+      // because we know for sure that this is a type definition.
+      const commentSyntax =
+        commentStart >= 0 &&
+        options.originalText.substring(commentStart).match(/^\/\*\s*::/);
```
lydell: Are newlines allowed between `/*` and `::`? I'm thinking of https://github.com/facebook/flow/issues/7050. Maybe something similar applies here?

swac: Doesn't seem like it: https://flow.org/try/#0PQKgUAXBAEAuCeAHAptAYgew9AvNAzrAE4CWAdgOYDcYIwYYAxhmYdAGZagybZ27QA5IKpA

That said, we should be safe to use `\s*` as our whitespace match here since we know we're looking at a `TypeParameterDeclaration` or `TypeParameterInstantiation` AST node.

j-f1:
Thanks for contributing!

Issue #5320, opened 2018-10-31
An option to enforce certain line endings (LF / CRLF)


kachkaev:
My issue is opposite to #92, which was closed by #472.

Our team uses Macs and Windows machines to work on the same code base and we struggle to enforce Unix-style line endings in the central repo. Time and again a person on Windows commits a file with CRLF instead of LF despite that we're using multiple methods to prevent that. These include:

* `git config --global core.autocrlf false` in contributing instructions (this does not help if the repo has already been checked-out with CRLF)
* `* text=auto eol=lf` in the repo's `.gitattributes` (not sure how this works)
* `.vscode/settings.json` that includes `"files.eol": "\n"` (not everyone uses vscode)
* `husky` + `lint-staged` as in [Prettier instructions](https://prettier.io/docs/en/precommit.html) (this helps commit clean code, but does not fix line endings)
* A CI/CD task that runs [`prettier-check`](https://www.npmjs.com/package/prettier-check) for the same file pattern as we have in `lint-staged` (it passes even if some files use lf and some crlf)

All the above measures do not stop a contributor from creating a new file with CRLF and committing it as is. Not setting `git config --global core.autocrlf false` moves the issue to another place: now a Windows user can check out the files in cmd.exe or SourceTree (LF → CRLF on checkout) and commit them via Ubuntu terminal (CRLF are preserved).

Prettier could really help us here if it had an option to enforce a certain flavour of line endings. I know that new options are not welcomed in general, but I can't really see a good workaround. The only thing that comes to my mind is to tweak our `package.json` somewhat like this:

```diff
   "lint-staged": {
     "complex-file-pattern-for-files-that-get-prettified": [
       "prettier --write",
+      "some-npm-module-that-converts-crlf-to-lf-for-a-given-file",
       "git add"
     ]
   },
```

This will serve as an additional barrier to not committing CRLF, however, it's not 100% robust. Someone can disable git hooks once and thus get CRLFs trough. `prettier-check` task in CI will not catch that. CRLFs will live in the central repo until someone check outs the latest commit, searches for `\r` in all code and makes another commit.

It feels like setting the flavour of new lines is a part of making the code prettier. So what do you think folks?

```bash
prettier --eol=lf path/to/file.js
```
```yaml
# .prettierrc or .prettierrc.yaml
eol: lf
```

Accepted `eol` values: `lf`, `crlf`, `auto` (default).

[...]

PR #5327, opened 2018-11-02
Add option to enforce certain line endings

kachkaev:
This PR contains a minimal viable implementation of https://github.com/prettier/prettier/issues/5320.

There is also a [sidecar PR in another repo](https://github.com/josephfrazier/editorconfig-to-prettier/pull/4) to read `.editorconfig` option as suggested by @evilebottnawi. I believe it is ready to be reviewed and can be merged after we close the first item in the TODO list below. We need to release a new version of `editorconfig-to-prettier` before completing this feature.

## TODO (based on reviews)

- [x] Agree on the option name:
    * `eol` **(originally proposed)** – a commonly known abbreviation, but may be confusing for newcomers
    * ✅ `endOfLine` / `--end-of-line` **(suggested by two maintainers)** – easier for newcomers, but takes longer to write; similar to `.editorconfig`'s `end_of_line`
    *  `lineEnding(s?)` / `--line-ending(s?)`  – a commonly used term in human-to-human communication

- [x] Decide if we want to support `--endOfLine=cr` at all. I'm not sure if Prettier can successfully parse files without `LF` at the moment. In any case, only `CRLF` and `LF` are prevalent (I haven’t seen anyone using `CR` only). Discussion: 4th conversation in https://github.com/prettier/prettier/pull/5327#pullrequestreview-171391657 (marked as resolved by @ikatyang). Solution: keep `cr` amongst values for consistency with what `.editorconfig` supports.

- [x] Merge https://github.com/josephfrazier/editorconfig-to-prettier/pull/4 and replace a temporary used fork with the original package.

Looking forward for your thoughts folks 🙌 

## Future work

- Discuss if it's possible to apply an explicitly defined `eol` for files that have a syntax error don't have a known grammar (e.g. `.gitignore`). This should make it possible to `prettier --write '**/*'` or `prettier-check '**/*'`. We can do this in follow-up issues and PRs.


package.json
```diff
@@ -29,7 +29,7 @@
     "dedent": "0.7.0",
     "diff": "3.2.0",
     "editorconfig": "0.15.2",
-    "editorconfig-to-prettier": "0.0.6",
+    "editorconfig-to-prettier": "https://github.com/kachkaev/editorconfig-to-prettier#eol-option",
```
lipis: Needs to be a version (as already mentioned)

lipis:
So this is ready to be merged? Maybe we should update the TODO list then..

lipis:
Also if this is part of the 1.15 milestone we will have to update the blog post as well!

kachkaev:
@lipis the PR is RFC (Ready for comments). I'm waiting for feedback from folks participating in https://github.com/prettier/prettier/issues/5320 to update TODOs. To be honest, I'd love `--eol` to become a part of 1.15, because dealing with EOLs is a big pain for my team 🙏

kachkaev:
FYI: I’ve replaced `\r` with `/*CR*/` in unit and integration tests; otherwise the distinction between different line endings vanishes. If this is considered controversial, I'm happy to change my workaround with something else.

suchipi:
I think this is fine to merge but I don't want to hold up the 1.15 release for it if it gets to that point.

src/main/core-options.js
```diff
@@ -62,6 +62,34 @@ const options = {
     `,
     cliCategory: CATEGORY_EDITOR
   },
+  eol: {
```
ikatyang: `endOfLine` should be better I think.

ikatyang:
> Discuss if it's possible to apply an explicitly defined `eol` for files that have a syntax error don't have a known grammar (e.g. `.gitignore`). This should make it possible to `prettier --write '**/*'` or `prettier-check '**/*'`. We can do this in follow-up issues and PRs.

I don't think we should format unsupported files, but you could create a fake plugin that matches everything to achieve the same goal.

kachkaev:
Thank you for your prompt reviews @suchipi  and @ikatyang 🙌 I will now:

* [ ] rename `eol` to `endOfLine`
* [ ] Add a couple of tests for markdown files

src/main/core-options.js
```diff
@@ -62,6 +62,34 @@ const options = {
...
+    type: "choice",
+    default: "auto",
+    description: "Which end of line characters to apply.",
+    exception: value => typeof value === "string",
+    choices: [
+      {
+        value: "auto",
+        description: dedent`
+          Maintain existing
+          (mixed values within one file are normalised by looking at what's used after the first line)
```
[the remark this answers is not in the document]
kachkaev: I'd probably keep it if that does not harm. Otherwise, it is not clear if Prettier keeps mixed line endings or combs them out. Mixed line endings can end up in a file if a text editor is not smart enough.

src/doc/doc-printer.js
```diff
@@ -241,7 +242,7 @@ function printDocToString(doc, options) {
   groupModeMap = {};
 
   const width = options.printWidth;
-  const newLine = options.newLine || "\n";
+  const newLine = convertEolToChars(options.eol);
```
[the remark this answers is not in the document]
kachkaev: Do you suggest setting `options.newLine` based on `options.endOfLine`? I thought of doing this in my first attempt, but then did not like seeing two possibly conflicting options for one thing in a single object. Although `options.newLine` has been exposed, it has not been documented anywhere.

src/main/core-options.js
```diff
@@ -62,6 +62,34 @@ const options = {
...
+        `
+      },
+      { value: "cr", description: "Carriage Return character only (\\r)" },
```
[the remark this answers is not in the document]
kachkaev: I agree that this choice is ancient, so probably not including it will not harm anyone. However, `cr` is one of accepted values for `end_of_line` in [`.editorconfig`](https://editorconfig.org/), so removing it will break consistency. If you think we should get rid of `cr` despite that, what shall we do if someone uses it in `.editorconfig`?

src/main/core-options.js
kachkaev: Done. Renamed.

src/doc/doc-printer.js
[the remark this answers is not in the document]
ikatyang: I'm not sure what should we do in this case (is it considered a breaking change?), let's see how other people think.

lipis:
Let's deal with this in the next release, as we might need more time for testing and investigating. Added a 1.16 milestone

kachkaev:
I hope all is good now, the investigation is complete. 🙂 The only thing left to decide is whether we want to keep `cr` as an option value or not (see the first comment and https://github.com/prettier/prettier/pull/5327#discussion_r230616681). If we agree to remove it, I'll make necessary changes right away no matter if it's day or night in order to make this PR ready for merging! 😅 

I agree with the rationale for adding this feature to 1.16 milestone – version 1.15 is about to be out. However, I'd be super grateful if we can manage to squeeze it in! As I've mentioned before, my cross-OS team has been struggling with CRLF/LF for a while, so waiting for another 2-3 months or sticking with Prettier fork will be a bit inconvenient to us. Lots of other teams have the same issue and it's very often unknown of. So the earlier they can leverage `endOfLine` in Prettier, the higher number of clean `git diffs` and sane `git blame` there'll be in the world!

Let the work on 1.15 take its pace and let's not hold the release for several days because of this feature. However, if the remaining tasks in this PR resolves quickly, let's let the magic happen ✨

lipis:
Regardless if this is going to be in 1.15 or not.. it won't take 2-3 months for the next release. We will try to release more often these kind of changes. HTML is a very special case.. and it's far from done with so many flavours.. :)

src/main/core-options.js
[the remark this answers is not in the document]
kachkaev: Given that you've marked this conversation as resolved, I put a checkbox next _Decide if we want to support `--endOfLine=cr` at all_ in the first comment.

kachkaev:
Looks like the only thing left to do is to wait for the new version of `editorconfig-to-prettier` following the merge of https://github.com/josephfrazier/editorconfig-to-prettier/pull/4! 🚀

src/doc/doc-printer.js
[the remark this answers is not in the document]
kachkaev: I'd say it is not a breaking change given that it was never exposed in the documentation, tests or examples.

josephfrazier:
> Looks like the only thing left to do is to wait for the new version of editorconfig-to-prettier following the merge of josephfrazier/editorconfig-to-prettier#4! 🚀

Done, you can upgrade to version `0.1.0` now!

kachkaev:
Many thanks for doing this so quickly @josephfrazier! 💯

This PR should be ready for a final review and can be merged after that! @lipis can we squeeze it into 1.15 given that all is done? 😅

duailibe:
@ikatyang You're pretty much managing the 1.15 release, I think you can decide to include in this release or not.

ikatyang:
I'd like to wait for a day to see if there's any objection, it will be included in 1.15 if everything goes well.

docs/options.md
```diff
@@ -271,3 +271,41 @@ Valid options:
...
+   Alternatively, you can add `* text=auto eol=lf` to the repo's `.gitattributes` file to achieve this.
+
+All modern text editors in all operating systems are able to correctly display line endings when `\n` (`LF`) is used.
+However, old versions of Notepad for Windows will visually squash such lines into one.
+
+Valid options:
+
+- `"auto"` - Maintain existing line endings
+  (mixed values within one file are normalised by looking at what's used after the first line)
+- `"cr"` - Carriage Return character only (`\r`), used very rarely
```
lydell: Is there a reason to support this value? I'd say we only support auto, crlf and lf.

ikatyang: --> https://github.com/prettier/prettier/pull/5327#discussion_r230617150

Basically, it's for `.editorconfig` compatibility.

kachkaev: The link no longer works because the discussion was marked as resolved. It's 4th in this review: https://github.com/prettier/prettier/pull/5327#pullrequestreview-171391657

lydell: Thanks. It's sad but it looks we don't have a choice then.

vjeux: Can you put it at the end of the bulletted list since no one will likely want to use it?

kachkaev: I'm tending to keep values ordered alphabetically both in the docs and in the code for consistency. Moving `cr` in the markdown would probably mean moving it in in `core-options.js` as well.

Perhaps we can re-order _all_ values except auto in the reverse alphabetical order, thus turning the list from `auto|cr|crlf|cr` into `auto|lf|crlf|cr` 🤔 That might be better actually! Shall I change?

kachkaev:
After some more thinking, I changed the order of option values both in docs and cli (`auto|cr|crlf|lf` → `auto|lf|crlf|cr`). It used to be alphabetic and now the values go in the order of prevalence. See a resolved discussion in @lydell's review for details.

@ikatyang if you don't agree with this change, simply chop off my last commit before merging. Let me know if you need help with anything else!

docs/options.md
kachkaev: Marking this as resolved. See https://github.com/prettier/prettier/pull/5327#issuecomment-436055858 below.

kachkaev:
Any final comments or suggestions folks? If not, shall we... 



? 😅

ikatyang:
Don't worry, I'll merge it later. I was fixing regressions.

ikatyang:
Thanks!

kachkaev:
Amazing!!! Many thanks for all your prompt comments and very useful reviews! I'm soooooo glad we could squeeze this feature into 1.15 that I'm jumping through the roof! 😅

PR #5331, opened 2018-11-03
Support config files that have `.prettierrc` extension with `--config` option

moznion:
Problem:

- when passing a configuration file that has a name with `.prettierrc` extension (e.g. `foo.prettierrc`), the runtime gives up with `[error] Invalid configuration file: No sync loader specified for extension ".prettierrc"`

This commit fixes this problem.

example: `prettier --config foo.prettierrc index.js`

--


lipis:
What's the usecase?

PR #5332, opened 2018-11-03
Update rationale for the 1.15 changes

lydell:
To be merged when 1.15 has been released and we've updated our own Prettier version.


suchipi:
Great work, thanks!

PR #5346, opened 2018-11-05
Make ast-to-doc.js a little easier to understand

twbartel:
Rename genericPrint function to callPluginPrintFunction to avoid similar name to printGenerically
Add comments that explain the descent/recursion
Rename "addAlignmentSize" parameter to "alignmentSize" to make it sound less like a function

No functionality has been changed. Test still run. No changes to API or CLI.


vjeux:
Thanks!

PR #5355, opened 2018-11-06
fix(html-in-js): do not break empty content

ikatyang:
This one should be the final fix in 1.15.


lipis:
🐑 it :)

j-f1:
IMO ```html`` ``` should never break. Perhaps check if the content is empty and print the backticks together in that case?

Issue #5360, opened 2018-11-07
@decorators on class are still being placed on separate line


desbrowne:
**Prettier 1.15.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAmHwEMwZVoAafCAIwGc4cA3Y2gGznwF98AzHBHT4A5OjoAPEQG4AOlAzY8hGgybMmPfoOFjJAWhxxSMGfMW4CAKhL18AJWNltQ0UZNmFWSyvsQArvAAwkLYUAgwAAqCmHa8Aq4i7mSGAfA4nuawTHykXCH+sAASxFAAJpw40RCx+HAS8OV2foFwId7hsNWxADzA3AB8KtzyAAJ0jCya9UoEYOzE9HYFxaUV0w0IZc1OMAB07WERPSswJeWV3fRDwPKyMONqLGyc+JCFBAC8+AAMclD3IzlJgACgAlIR5PhoTDIFB6AQiGViDBiFpvjAABaoej-GEwowwfw4KD4EFQ-EwnplVDMAYUykw05IQhYnF7d6wEakxlU2iBGDQGhQILsVBgADWn2AbPoe0xa04AEkoGB3IxReKJYMVWrjIwegB6fkwQVQek83n4Homs3CzWS6UKi5wRyMGAO7UDN1wGBG23QC2M+5Gml0hkwsH-e7c+6jEwUUnO9a69VtMWS-DfcFZoayjkBWAAaiL0YeCaFyc4Po9GYlWbJEM+Nwj0PznK+vzxMO53BAlBANXI0HoyFAxBwggA7pEJwhRyhiMwIKgyv2QLQcKQJb6AMqYUioKAAc2QMBw-jgA8xMHQ7AA6tj4PQD2A4Lv56hyMwvwBPZDgEs65HpMURbse6DEMguTsIwA4AFb0BIABCW6SnuxDoHAAAyR5wNBxCwVeICIRIu5HsenAAIr+BA8AEURA4HjgkwATAv6YHA9BqqgmAwOumA4EeMD3quWLIAAHD8TGCIw95bpgAGCVx6j4QORgAI7+KgRjRMQEFQUgMFwSAjDoKgZ4XsR9AUdRtH0UZhEmaitCiWU4lIAATAO57EKgYonu0kEAVA0BqSA-iMAAKmwC7GXA3DcEAA)
```sh
--parser typescript
```

**Input:**
```tsx
[...]

PR #5423, opened 2018-11-09
fix(javascript): inline property decorator should stay inline (part 2)

ikatyang:
Fixes cases mentioned in https://github.com/prettier/prettier/issues/5360#issuecomment-436842209

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -99,6 +99,24 @@ class HeroButtonComponent {
 }
 ```
 
+There's one exception: classes. We don't think it ever makes sense to inline the decorators for them, so they are always moved to their own line.
+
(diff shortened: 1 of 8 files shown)


docs/rationale.md
```diff
@@ -99,24 +99,6 @@ class HeroButtonComponent {
 }
 ```
 
-There's one exception: classes. We don't think it ever makes sense to inline the decorators for them, so they are always moved to their own line.
```
ikatyang: --> https://github.com/prettier/prettier/issues/5360#issuecomment-436516768

ikatyang:
OK, let's bisect them from this PR and see how people think about it.

PR #5430, opened 2018-11-09
Handle reselect's createSelector as function composition

### LINKED ISSUE #5285: Prettier should have a better opinion on createSelector() calls
**Prettier 1.14.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBzOMDKcAhgE5gAWASnGgK4A2MamAvJmCcfIfXGDBBIAKADpRMEnHkKkK1OowCSAEzQAaMZKkwAgvXqUIEALbrNkoSSPGV6zEX2GTaAJSsAfJisnbAOmNEAA5CAJbKHvaO1mgA2mEAui5iLgDcIGogEIEwIejIoKRWAO4ACqQIaMggRABuEGHpIABGJERgANbSgW0hUNjIMCS0cBm9aHAkMCWt2AHIAGYO4xkAVmgAHgBCrR3SRMZwADK9cAtLIyDdJOMkVU1ETQCe9NCNgSS9MADqYTDkyABGABMAAYMu8IOMvq1AlV3jQJjVThlOABHWghTjTIizIhnejLEDjYwhAZDC5oXrYXgARVoEHg+MJMAeP2Uf2QABYMoMiCF6FSAMImOYoKDQZEgWjjAAqD0qSEWBLgAF8VUA)
```sh
--parser babylon
--print-width 120
--tab-width 4
```

**Input:**
```jsx
const getSearchResults = createSelector(
    getSearchResultIds,
    getAllRooms,
    (roomIds, allRooms) => roomIds.map(id => allRooms[id])
);
```

**Output:**
```jsx
const getSearchResults = createSelector(getSearchResultIds, getAllRooms, (roomIds, allRooms) =>
    roomIds.map(id => allRooms[id])
);

```

**Expected behavior:**
```jsx
const getSearchResults = createSelector(
    getSearchResultIds,
    getAllRooms,
    (roomIds, allRooms) => roomIds.map(id => allRooms[id])
);
```

I understand that currently prettier applies its rules regardless of context, but in case of reselect createSelector() calls this really reduces readability. The style that can be found in pretty much all examples of reselect and projects using it is very straightforward: One input selector per line, and then the callback to combine the input.

However, due to prettier's preference of wrapping a callback argument if everything else fits in the same line (or even fitting everything in the same line), the resulting code is harder to read. It's also inconsistent if you have multiple selectors in the same file and depending on the length of their arguments they sometimes get wrapped and sometimes they don't.

One possible option for this could be to have some default overrides for certain function names, e.g. "one argument per line" for createSelector. I realize this might get close to "too much configuration" territory but it could always use a hardcoded list (or sane defaults) - also, easy-to-read/understand code should be a top priority IMHO, and that's not the case with the current formatting here.

ThiefMaster:
This fixes #5285; initially I thought this was a feature but after seeing that #4431 already added special handling for similar function composition functions, including one from redux, I think adding support for [reselect's `createSelector`](https://github.com/reduxjs/reselect) could be considered a fix (especially since it's an official library in the same github org as redux)

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -3780,7 +3780,8 @@ const functionCompositionFunctionNames = new Set([
   "composeK", // Ramda
   "flow", // Lodash
   "flowRight", // Lodash
-  "connect" // Redux
+  "connect", // Redux
+  "createSelector" // Reselect
 ]);
 
 function isFunctionCompositionFunction(node) {
--- a/tests/functional_composition/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/functional_composition/__snapshots__/jsfmt.spec.js.snap
@@ -390,6 +390,35 @@ const ArtistInput = connect(
 
 `;
 
+exports[`reselect_createselector.js - flow-verify 1`] = `
+import { createSelector } from 'reselect';
+
+const resolve = createSelector(
+  getIds,
+  getObjects,
+  (ids, objects) => ids.map(id => objects[id])
+);
+
+const resolve = createSelector(
+  [getIds, getObjects],
+  (ids, objects) => ids.map(id => objects[id])
+);
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+import { createSelector } from "reselect";
+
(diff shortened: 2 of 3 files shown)


ThiefMaster:
CI failures seem unrelated

suchipi:
Looks great, thank you!

PR #5493, opened 2018-11-15
demonstrate bug with `css` parsers and tests

alexander-akait:
Some tests are invalid because used invalid parser. 

Tests not respected extension of file. It is expected? Because using this through CLI we have right value in `opts.parser`.

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -478,7 +478,8 @@ function parseNestedCSS(node) {
   return node;
 }
 
-function parseWithParser(parser, text) {
+function parseWithParser(isSCSSParser, text, opts) {
+  const parser = requireParser(isSCSSParser);
   const parsed = parseFrontMatter(text);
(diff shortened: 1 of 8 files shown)


ikatyang:
Since we explicitly specify which parser to test in `jsfmt.spec.js`, this should be the expected behavior. But I think we should tweak our tests to be more explicit:

- before
  ```
  input
  ~~~~~
  output
  ```
- after
  ```
  parser: babylon, typescript
  singleQuote: true
  ~~~~~
  input
  ~~~~~
  output
  ```

evilebottnawi:
@ikatyang it is required to restructure a lot of tests on new directories :disappointed: maybe we handle this edge case with `css`, infer right parser based on extension

lipis:
What's the status with this one?

evilebottnawi:
@lipis we need refactor tests for css to avoid this problem - `.sass` should be in own directory with `parser: scss`, same for `.less` and `.css`

PR #5494, opened 2018-11-16
[WIP] fix: CRLF

ikatyang:
It looks like we did not test CRLF in most cases.

### DIFF
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -38,7 +38,6 @@ function run_spec(dirname, parsers, options) {
       let rangeEnd = Infinity;
       let cursorOffset;
       const source = read(path)
-        .replace(/\r\n/g, "\n")
         .replace("<<<PRETTIER_RANGE_START>>>", (match, offset) => {
           rangeStart = offset;
           return "";
@@ -60,9 +59,18 @@ function run_spec(dirname, parsers, options) {
         cursorOffset
       });
       const output = prettyprint(input, path, mergedOptions);
+      const visualizedOutput = visualizeEndOfLine(output);
       test(`${filename} - ${mergedOptions.parser}-verify`, () => {
+        expect(visualizedOutput).toEqual(
+          visualizeEndOfLine(consistentEndOfLine(output))
+        );
         expect(
-          raw(source + "~".repeat(mergedOptions.printWidth) + "\n" + output)
+          raw(
+            normalizeEndOfLine(source) +
+              "~".repeat(mergedOptions.printWidth) +
+              "\n" +
+              normalizeEndOfLine(output)
+          )
         ).toMatchSnapshot();
       });
 
@@ -70,7 +78,7 @@ function run_spec(dirname, parsers, options) {
         const verifyOptions = Object.assign({}, mergedOptions, { parser });
         test(`${filename} - ${parser}-verify`, () => {
           const verifyOutput = prettyprint(input, path, verifyOptions);
-          expect(output).toEqual(verifyOutput);
+          expect(visualizedOutput).toEqual(visualizeEndOfLine(verifyOutput));
         });
       });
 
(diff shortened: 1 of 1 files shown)


ikatyang:
Some initial thoughts:

- Replace all CRLF with LF in this repo
- Set `* text=auto` in `.gitattributes`. If I understand correctly, it allows us to have consistent LF on Unix and consistent CRLF on Windows.
  > (https://stackoverflow.com/a/38017715)
  > `text=auto`: when someone commits a file, Git guesses whether that file is a text file or not, and if it is, it will commit a version of the file where all CR + LF bytes are replaced with LF bytes. It doesn't directly affect what files look like in the working tree, there are other settings that will convert LF bytes to CR + LF bytes when checking out a file.
- Visualize whitespaces for comparison (i.e., `expect(visualize(a)).toEqual(visualize(b))`), we could observe their diffs if the test failed.
- Jest normalizes line endings for snapshots so we don't need to handle it. (no visualization)

j-f1:
Seems like Azure Pipelines didn’t run for some reason.

ikatyang:
@j-f1 Its message looks confusing but it actually works, you need to click it to see the detail logs.

---

@kachkaev

> Jest simply kills all CRs in snapshots – see its [`normalizeNewlines()` function](https://github.com/facebook/jest/blob/8eefa9690e167e9f056ec5d9532a9df465018c67/packages/jest-snapshot/src/utils.js#L157). We can't really check CR/CRLF with this, and this is why `replace(/\r/g, "/*CR*/")` was necessary. Perhaps, we could use `visualizeEndOfLine()` in snapshots too?

I meant, the CRLF/LF checks are done in `expect(visualize(a)).toEqual(visualize(b))`, snapshots are used for visible records. And that's also the reason why I don't want to use `eol=lf`, I'd like to use CRLF on our Windows CI to test CRLF.

ikatyang:
Or maybe we should use `eol=lf` and add another test to replace LF with CRLF before formatting?

kachkaev:
IMO we can use `eol=lf` in the whole repo except a couple of test files where we're checking `endOfLine` option.

azure-pipelines.yml
```diff
@@ -43,6 +43,18 @@ jobs:
     steps:
       - template: .azure-pipelines/dev.yml
 
+  # There're issues for unicode on Windows build
+  # https://github.com/Microsoft/azure-pipelines-tasks/issues/8534
```
ikatyang: Windows:

image

Linux:

image

ikatyang:
Based on https://github.com/prettier/prettier/pull/5498#issuecomment-439600771, it seems we should revert 38bfa050b2f265729dd93c247277053585ecd2d2, but I'm not sure how to deal with those bad unicode. Should we use `<CRLF>` instead of `↵`?

kachkaev:
_copied a couple of comments from https://github.com/prettier/prettier/pull/5498 as asked by @ikatyang_

> We shouldn't need to configure global settings and we should be able to test CRLF locally even if it's on LF system (via TEST_CRLF).

Testing on Windows machines with globally configured CRLF and LF in git is what I meant. This can be useful because Windows folks have different preferencess and all of them should be able to contribute to Prettier without problems of running local tests. By default, git’s core.autocrlf is true, but some devs set it to false, e.g. create-react-app contribution guidelines [recommend doing so](https://github.com/facebook/create-react-app/blob/master/CONTRIBUTING.md#line-endings).

I did not know Azure pipelines had a flag for CRLF, interesting!

----

Running tests on a Windows machine with LF ia different than doing this on Linux, because on Windows [`os.EOL`](
https://nodejs.org/api/os.html#os_os_eol) is CRLF regardless of how git is configured. Some modules refer to this value and this can affect what Prettier generates in the end.

ikatyang:
Updated the OP to reflect the changes, ready for review.
(Most of the diffs are line ending normalization, just ignore them.)

PR #5497, opened 2018-11-16
Fix `hasFlowShorthandAnnotationComment` check

ukyo:
Check node.trailingComments.length in hasFlowShorthandAnnotationComment.


j-f1:
Thank you for opening a PR, @ukyo! Can you add a test for this scenario to make sure there isn’t a regression for this in the future?

ukyo:
Hi @j-f1 , I would like to add the following test. Where should I put it? I can't find [custom parser api](https://prettier.io/docs/en/api.html#custom-parser-api) tests.

```js
const prettier = require("prettier/local");
const traverse = require("@babel/traverse").default;

test("replace CallExpression", () => {
  expect(
    prettier.format("a(x)", {
      parser(text, { babylon }) {
        const ast = babylon(text);
        traverse(ast, {
          CallExpression(path) {
            if (path.node.callee.name === "a") {
              path.replaceWithSourceString("b(y)");
            }
          }
        });
        return ast;
      }
    })
  ).toEqual("b(y);\n");
});

```

duailibe:
@ukyo there are a couple of tests already: https://github.com/prettier/prettier/blob/master/tests_integration/__tests__/parser-api.js

fisker (approved):
I'm sure you already handled your case, but I think we can add this condition to prevent it happen to our users use the custom parser api.

Tests added, condition changed a little bit to make linter happy.

PR #5498, opened 2018-11-16
chore: tweak azure-pipelines.yml

ikatyang:
- add missing AST_COMPARE
- test on Node 4/10

---


j-f1:
Can you move the new files to a `.azure-pipelines` directory?

kachkaev:
Can there be two azure pipelines with tests? One with `git config --global core.autocrlf true` (checks out files with CRLF by default) and `git config --global core.autocrlf false` (checks out files as is).

If gitattributes are configured correctly, that should not affect tests.

j-f1:
Netlify is segfaulting :/

ikatyang:
@kachkaev I'd like to use `eol=lf` in #5494 and add another test flag `TEST_CRLF` to test CRLF by replacing LF with CRLF before formatting. We shouldn't need to configure global settings and we should be able to test CRLF locally even if it's on LF system (via `TEST_CRLF`).

@j-f1 It looks like a random error, I've rebuilt it and it's green now.

I'm going to merge this PR now since it's only for tweaking the current settings, CRLF should be handled in #5494.

kachkaev:
> We shouldn't need to configure global settings and we should be able to test CRLF locally even if it's on LF system (via TEST_CRLF).

Testing on Windows machines with globally configured CRLF and LF in git is what I meant. This can be useful because Windows folks have different preferencess and all of them should be able to contribute to Prettier without problems of running local tests. By default, git’s core.autocrlf is true, but some devs set it to false, e.g. create-react-app contribution guidelines [recommend doing so](https://github.com/facebook/create-react-app/blob/master/CONTRIBUTING.md#line-endings).

I did not know Azure pipelines had a flag for that, interesting!

kachkaev:
Running tests on a Windows machine with LF ia different than doing this on Linux, because on Windows [`os.EOL`](
https://nodejs.org/api/os.html#os_os_eol) is CRLF regardless of how git is configured. Some modules refer to this value and this can affect what Prettier generates in the end.

ikatyang:
@kachkaev Can you move your comments to #5494? It should be the better place for CRLF discussion.

ikatyang:
(`TEST_CRLF` is a new flag I added in our `run_spec.js`. 😅)

PR #1453, opened 2017-04-28
Break inline object first in function arguments

vjeux:
This is getting subtle where the groups need to be in a precise position but that works :)

Fixes #1409

PR #5500, opened 2018-11-18
Use visible groups mechanism for "hugged" params

helixbass:
I started looking for places in the JS formatter where I could apply the visible groups mechanism introduced in #4462 (ideally I could get suggestions of places it might be useful from contributors with a strong knowledge of the JS formatter - @duailibe and @rattrayalex both mentioned in #4462 that they might have ideas of places it could come in handy but haven't given me any specific examples yet)

The one place I already knew might make sense to try and use visible groups is in nested object destructuring since the release notes for 1.12 sort of imply its existence in the description of #4267: "any nested non-leaf object or array patterns in a destructuring pattern will get multi-lined if any of their parents do"

In looking at using visible groups there, I ran into the fact that sometimes the `group()` around an `ObjectPattern` gets omitted, specifically when a function should "hug its parameters" - it was introduced in #1453 as a way of trying to make sure that hugged params break before the function's return type. This "inlining" (non-grouping) of the hugged param feels like a hack, and was going to cause me difficulty trying to use visible groups for nested object destructuring

But lo, visible groups can be used to get rid of this hack! It provides a cleaner/more declarative/more decoupled way of saying "always break the hugged params if the enclosing group (which contains the params and return type) breaks"

So this PR restores the symmetry of always having a `group()` enclosing an `ObjectPattern` and uses visible groups to make sure it breaks in the hugged-param case

(I haven't included the nested object destructuring use of visible groups in this PR)

Based on `visible-group` (#4462), [here](https://github.com/helixbass/prettier/compare/visible-group...helixbass:hug-visible-group) is just the diff against that branch


tests/function_single_destructuring/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -330,6 +330,8 @@ type T = ({
   items: number
 }) => void;
 
-const X = (props: { a: boolean }) => <A />;
+const X = (props: {
+  a: boolean
```
helixbass: This is interesting evidence that using visible groups is a cleaner approach here:

The "respect newlines in `ObjectTypeAnnotation`s" heuristic was being violated in the hugged-param case because there was no `group()` to force-break

PR #5505, opened 2018-11-19
feat: add mjml extension to html language

### LINKED ISSUE #5471: Feature Request: Add MJML Formatting
Can we add support for formatting MJML templates?

[Official Website](https://mjml.io/)

Shouldn't be too hard to add support, since it's essentially HTML. I'm willing to contribute to a PR for this but also wouldn't mind some guidance/advice. Thanks!

n1ru4l:
Fixes #5471

This pull request adds a parser for mjml files (https://mjml.io/documentation/) which is a variation of html using custom tags. As suggested in the feature request issue (#5471). I used the same parser which is used for vue.

Also I two basic tests, overall there should not be much to test since there is no difference to html expect the self-closing tags and the custom tag names.

I did not update `docs/options.md` yet since I do not know under which version number the parser will be published (if it will be published).


j-f1:
cc @ikatyang 
Should we just enable self-closing tags for plain HTML?

j-f1:
Can you change that option in the HTML parser, remove the MJML parser, and add `.mjml` to the list of extensions for HTML @n1ru4l?

n1ru4l:
@ikatyang Had some time and updated the pull request, rebased on master and adjusted the changes as you requested 😊

j-f1:
Can you add MJML to the homepage?

j-f1:
The data file is here: `website/data/languages.yml`. The HTML section would work best.

docs/index.md
```diff
@@ -17,6 +17,7 @@ Prettier is an opinionated code formatter with support for:
 - [GraphQL](http://graphql.org/)
 - [Markdown](http://commonmark.org/), including [GFM](https://github.github.com/gfm/) and [MDX](https://mdxjs.com/)
 - [YAML](http://yaml.org/)
+- [MJML](https://github.com/mjmlio/mjml)
```
ikatyang: I don’t think we should add it to the docs explicitly, it’s just the same as the HTML formatting, we don’t need to add every file extension to the support list.

n1ru4l:
@j-f1 @ikatyang done!

ikatyang:
Sorry, I might not describe it well but what I meant in https://github.com/prettier/prettier/pull/5505#discussion_r239284054 is that we didn't do any special handling for it so we should not add it to the docs (including homepage).

ikatyang:
Thanks!

PR #5521, opened 2018-11-22
test: improve snapshots

ikatyang:
Context: https://github.com/prettier/prettier/pull/5493#issuecomment-439341433

- before
  ```
  ${input}${"~".repeat(printWidth)}
  ${output}
  ```
- after
  ```
  ===options===
  ${options}
  ${" ".repeat(printWidth)}| printWidth
  ====input====
  ${input}
  ===output====
  ${output}
  =============
  ```
---


tests/angular_component_examples/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,6 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`test.component.ts - typescript-verify 1`] = `
+====================================options=====================================
+parsers: ["typescript"]
```
j-f1: Should this specify the exact parser instead?

ikatyang: Did you mean to use exact parser if there's only 1 parser?

j-f1: Are the tests run separately for each parser or are they combined somehow?

ikatyang: They combined in a sense:

https://github.com/prettier/prettier/blob/6ee2f464ac5e0bb05f0418a303bdfad2e7710bb7/tests_config/run_spec.js#L62-L75

j-f1: I see, so there aren’t snapshot tests created for them. Could you remove the `- foo-verify` from the test name for the snapshot case to clarify?

Issue #5311, opened 2018-10-29
NodeJS API: Changes to .js config not being picked up due to NodeJS module caching


undeadcat:
**Environments:**
- Prettier Version: **1.14.1 and above**
- Running Prettier via: Node.js API 
- Runtime: Any Node.js version
- Operating System: Any OS

**Steps to reproduce:**
This is reproducible via any Prettier editor integration, e.g. [prettier-vscode](https://github.com/prettier/prettier-vscode) or [Intellij prettier plugin](https://plugins.jetbrains.com/plugin/10456-prettier):

* Create a .prettierrc.js config file with, for example, a specified tab size:
```
module.exports = {
  tabWidth: 7
}
```
* Invoke 'Reformat Code' from the editor. File is reformatted according to .prettierrc correctly
* Modify .prettierrc.js, for example, changing `tabWidth` to 2
* Make sure the editor persists changes to the config file to disk
* Invoke 'Reformat Code' again

**Expected behavior:**
File is reformatted with the new tabWidth (2).
**Actual behavior:**
File is reformatted with the previous tabWidth (7), until the editor is restarted. After that the file is reformatted with new settings.

With `prettier` version **1.13.7** the changes to the file are picked up correctly.
Same with .json or .yaml config files - changes are picked up immediately after changing the config file.

This seems to be caused by the fact that Node.js `require` [caches the loaded modules](https://nodejs.org/api/modules.html#modules_caching) for the duration of the process. Editor integrations usually create a long-running process that then invokes 'reformat' multiple times, so if the config is read via `require`, subsequent attempts to read it will return the same result.
Prettier updated `cosmiconfig` to version 5 [here, in 1.14.1](https://github.com/prettier/prettier/commit/440ab4f60b7e7c4fe80565b65929a1a9534de86d#diff-b9cfc7f2cdf78a7f4b91a753d10865a2). 
Cosmiconfig replaced `require-from-string` with plain `require` [here](https://github.com/davidtheclark/cosmiconfig/commit/0143e84934d0f425af81327c12ebd1fb2eaad6c7#diff-b9cfc7f2cdf78a7f4b91a753d10865a2). 
`require-from-string` [had the side effect of not caching the loaded modules](https://github.com/floatdrop/require-from-string/issues/13#issuecomment-360775277).
ESLint had a similar issue in the past: https://github.com/eslint/eslint/issues/5067 and addressed it by using `require-uncached`. 

[...]

PR #5558, opened 2018-11-27
Updating cosmiconfig to not cache .js config files when told not to

wuweiweiwu:
fixes: #5311 


ikatyang:
It seems Jest uses its own `require` (facebook/jest#5120) which we're not able to test `require`'s cache using Jest. I guess we could just update `cosmiconfig` without adding a test in this case.

SimenB:
You can run `jest.resetModules()`, but I'm now sure how cosmiconfig caches. If it's just `delete require.cache[thing]` then it's always empty when running through Jest, so a `resetModules` should work

SimenB:
Ah, indeed.

I have some nefarious plans at some point to add support for `require.cache` properly (including `delete`, `Proxy` makes it quite straight forward), but it's super low pri at the moment

wuweiweiwu:
removed the test!

ikatyang:
Thanks!

PR #5562, opened 2018-11-27
feat: Upgrade dependencies

lipis:
It's been a while and I updated only the minor versions of them:

screenshot 2018-11-27 at 14 27 06


lipis:
Not so easy after all :/

j-f1:
Seems like `no-control-regex` should be removed and the other issue should be fixed.

package.json
```diff
@@ -61,62 +61,62 @@
...
-    "shelljs": "0.8.1",
-    "snapshot-diff": "0.4.0",
+    "shelljs": "0.8.3",
+    "snapshot-diff": "0.4.1",
     "strip-ansi": "4.0.0",
     "tempy": "0.2.1",
     "webpack": "3.12.0"
   },
   "resolutions": {
-    "@babel/code-frame": "7.0.0-beta.46"
```
SimenB: this will probably break your node 4 CI unless you transpile everything (maybe you do?)

package.json
```diff
@@ -61,62 +61,62 @@
...
-    "eslint-plugin-import": "2.9.0",
-    "eslint-plugin-prettier": "2.6.0",
-    "eslint-plugin-react": "7.7.0",
-    "execa": "0.10.0",
-    "jest": "23.3.0",
-    "jest-junit": "5.0.0",
+    "eslint-plugin-import": "2.14.0",
+    "eslint-plugin-prettier": "2.7.0",
+    "eslint-plugin-react": "7.11.1",
+    "execa": "0.11.0",
```
SimenB: should be `1.0.0`: https://github.com/sindresorhus/execa/releases/tag/v1.0.0

j-f1:
Whoa Checks lets you rerun failed builds 😯

lipis:
Checks are nice..! Any clue what on what we need to downgrade in order to fix the problem? Or actually fix the problem?

lipis:
Feel free to edit directly in the branch

package.json
```diff
@@ -61,56 +61,56 @@
...
     "jest-snapshot-serializer-ansi": "1.0.0",
     "jest-snapshot-serializer-raw": "1.1.0",
-    "jest-watch-typeahead": "0.1.0",
+    "jest-watch-typeahead": "0.2.0",
     "mkdirp": "0.5.1",
     "prettier": "1.15.2",
     "prettylint": "1.0.0",
     "rimraf": "2.6.2",
-    "rollup": "0.47.6",
+    "rollup": "0.67.3",
```
j-f1: This is probably the issue. In `0.x.y` releases, the `x` is the major version. Rollup probably made breaking changes to their config.

yarn.lock
```diff
@@ -3443,7 +3443,7 @@ jest-changed-files@^23.4.2:
   dependencies:
     throat "^4.0.0"
 
-jest-cli@^23.6.0:
+jest-cli@^23.3.0:
   version "23.6.0"
```
SimenB: the downgrade has to be more manual due to semver ranges, jest is still the same version

ikatyang:
We should probably upgrade patches/prereleases first, then upgrade minors one-by-one. It's quite hard to revert a single dependency and its sub-dependency locks at the same time.

duailibe:
Is this updating just for the sake of it or were there actual relevant updates to them?

PR #5612, opened 2018-12-09
Remove the dynamic `require()` call in the standalone bundle

### LINKED ISSUE #4959: webpack warning: "Critical dependency: the request of a dependency is an expression"
**Environments:**
- Prettier Version: 1.14.2
- Running Prettier via: Angular 6, WebPack, TypeScript
- Runtime: Node v8.9.4, TypeScript 2.7.2
- Operating System: macOS

**Steps to reproduce:**

`npm install --save --save-exact prettier`

in some component 

```
import * as prettier from "prettier/standalone";
import * as parserYaml from "prettier/parser-yaml";

...

//somewhere within 
const formatted = prettier.format(..., { semi: false, parser: "yaml", plugins:[parserYaml] } );

```
**Expected behavior:**

No compilation warnings 

**Actual behavior:**

WARNING in ./node_modules/prettier/standalone.js
7397:15-64 Critical dependency: the request of a dependency is an expression

(everything seems to work well, the issue is just the compilation warning)

j-f1:
Prettier will now throw an error if it can’t find the parser in the `parsers` object.

Fixes #4959.

- I’ve added tests to confirm my change works. ❓ How should I test this (if at all)?


scripts/build/babel-plugins/strip-target-env.js
```diff
@@ -0,0 +1,36 @@
...
+// BEFORE:
+//   if (!process.env.PRETTIER_TARGET_UNIVERSAL) {
+//     foo();
+//   }
+//
+// AFTER:
+//   // in standalone mode: no output
+//   // in regular mode:
+//   foo();
+//
```
ikatyang: This looks like the `NODE_ENV` replacing magic to me, we should be able to use the same magic (dead code elimination) in this case, and limit it to the `standalone.js` bundling.

https://github.com/prettier/prettier/blob/952bc0cc039c034c5fe651864db528dcd9800e9e/scripts/build/bundler.js#L97-L99

https://github.com/prettier/prettier/blob/952bc0cc039c034c5fe651864db528dcd9800e9e/src/doc/doc-builders.js#L15-L17

j-f1: Are those statements removed from the outputted file or does the if statement stay there?

ikatyang: They'll be removed by the dead code elimination if it replaced correctly.

```js
// source
if (process.env.NODE_ENV) { doSomething() }

// input (replaced)
if (false) { doSomething() }

// output (bundled)
/* nothing */
```

src/main/parser.js
```diff
@@ -42,14 +42,19 @@ function resolveParser(opts, parsers) {
...
     }
-    try {
-      return {
-        parse: eval("require")(path.resolve(process.cwd(), opts.parser)),
-        astFormat: "estree",
-        locStart,
-        locEnd
-      };
-    } catch (err) {
+    if (process.env.PRETTIER_TARGET !== "universal") {
```
ikatyang: How about an early exit with `BUNDLE_TARGET !== "node"`?

```js
if (process.env.BUNDLE_TARGET !== "node") {
  throw ...
}

try { ... }
catch (...) { ... }
```

src/main/parser.js
```diff
@@ -42,14 +42,19 @@ function resolveParser(opts, parsers) {
...
+          locStart,
+          locEnd
+        };
+      } catch (err) {
+        /* istanbul ignore next */
+        throw new ConfigError(`Couldn't resolve parser "${opts.parser}"`);
+      }
+    } else {
       /* istanbul ignore next */
       throw new ConfigError(`Couldn't resolve parser "${opts.parser}"`);
```
ikatyang: Something like _"Resolving parser is not supported in non-Node environment"_ should be better I guess.

PR #5613, opened 2018-12-09
Don’t run tests on Node 6

j-f1:
They’re currently segfaulting so we ignore them. Therefore there’s no point in even running this job until #3457 is fixed since there isn’t a way for it to signal that something has gone wrong on Node 6 other than the segfault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
     - JOB=test AST_COMPARE=1
 
 matrix:
+  fast_finish: true
   include:
     - node_js: "node"
       env: JOB=lint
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,15 +17,9 @@ env:
     - JOB=test AST_COMPARE=1
 
 matrix:
-  fast_finish: true
```
ikatyang: Ref: https://blog.travis-ci.com/2013-11-27-fast-finishing-builds

> With fast finishing enabled, Travis CI will mark your build as finished as soon as one of two conditions are met:
>
> - The only remaining jobs are allowed to fail, or 
> - **a job has already failed.**
> 
> In these cases, the status of the build can already be determined, so there’s no need to wait around until the other jobs finish.

We should leave it enabled.

PR #5626, opened 2018-12-11
Fix incorrect grammar

gnapse:
A sentence that did not read well. Fixed it with the smallest amount of new words I could think of.

- [ ] (N/A) I’ve added tests to confirm my change works.
- [ ] (N/A) (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -222,7 +222,7 @@ describe("NodeRegistry", () => {
 });
 ```
 
-Prettier special cases work with common test framework functions such as `describe`, `it` and `test`.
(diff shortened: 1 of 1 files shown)


docs/rationale.md
```diff
@@ -222,7 +222,7 @@ describe("NodeRegistry", () => {
 });
 ```
 
-Prettier special cases common test framework functions such as `describe`, `it` and `test`.
+Prettier special cases work with common test framework functions such as `describe`, `it` and `test`.
```
j-f1:
```suggestion
Prettier has special cases for common testing framework functions such as `describe`, `it` and `test`.
```

Issue #5520, opened 2018-11-21
prettier --lint / --check mode


kachkaev:
Prettier CLI has an interesting option called [`--list-different`](https://prettier.io/docs/en/cli.html#list-different), which is designed to fail the CI if not all files are formatted. That's a nice perk for those who want to stick with good practices, so it's great that Prettier has such an execution mode built into the core!

In my opinion, one minor problem with `--list-different` as means to check the code base is that it's not quite ‘humane’. The output is just a list of unformatted files, which is fine for piping them to another command but probably a bit too ‘dry’ for a human to interpret. A repo contributor who's CI pipeline has failed will see something like:

```
$ yarn install
...
✨  Done in 23.41s.

$ yarn prettier --list-different "src/**/*.js"
src/fileA.js
src/fileB.js
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

It's a bit hard to say what's behind the failure and what to do with this, isn't it? To solve a probable confusion, there exist two npm modules that I know of: [`prettier-check`](https://www.npmjs.com/package/prettier-check) by @hexacta and [`prettylint`](https://github.com/ikatyang/prettylint) by @ikatyang. Frankly speaking, their sole goal is to provide alternatives to `--list-different` that would simply output some text for humans in addition to the very same file list. Here's what I see in CI when utilizing `prettier-check` instead of `prettier --list-different`:

```
$ yarn prettier-check "src/**/*.js"
Forgot to run prettier? There are files without correct code style:
src/fileA.js
src/fileB.js

error Command failed with exit code 3.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

or if I'm lucky:

```
yarn prettier-check "src/**/*.js"
All files using prettier code style.
✨  Done in 1.89s.
```

The output is much better as it makes perfect sense!

Given that the distinction between `--list-different` and an external npm module is purely cosmetic, would not it be great to install just one npm dependency instead of two? This would not only ease the project setup process, but also make the dependency tree much lighter:

```bash
cd new-project-with-prettier
yarn init --yes
yarn add --dev prettier
## node_modules: 8.5MB, 8 lines in yarn.lock listing 1 package
```

```bash
cd new-project-with-prettier-check
yarn init --yes
yarn add --dev prettier prettier-check
## node_modules: 8.9MB, 116 lines in yarn.lock listing 18 packages
```

```bash
cd new-project-with-prettylint
[...]

Issue #4612, opened 2018-05-31
Add diff in output of --list-different


pronebird:
Hi,

Currently `--list-different` only prints the filenames of files that aren't formatted. It would be great to add an option to print out a diff between current and formatted source.

Maybe an extra option like `--list-different --diff` that would affect the way `--list-different` works. I believe in any case `--list-different` runs formatter on the source code, so it shouldn't be too hard to run it through some `diff` tool and output a nicely colored code to console.

ad1992:
@j-f1 can I take this up ?

j-f1:
Sure, go ahead!

martinsik:
I just encountered the same thing. It would be very helpful if I could see what exactly is different in the source and the re-formatted file.

arnavb:
Any update on this issue by any chance? I'd also like to see this implemented, similar to how rustfmt does it.

alex-arriaga:
This would be a great option to have, is there any news about it?
Thanks in advance.

thorn0:
closing in favor of #6885

PR #5629, opened 2018-12-12
Add --check option

kachkaev:
Closes https://github.com/prettier/prettier/issues/5520

- [x] I’ve implemented the functionality.
- [x] I’ve documented the changes I’ve made (see [preview](https://deploy-preview-5629--prettier.netlify.com/docs/en/cli.html#check)).

WDYT folks?


ikatyang:
I think it'd be better to use a reporter-like solution:

```
prettier --reporter my-reporter.js
```

```js
// my-reporter.js

module.exports = function ({ filename, source, formatted }, { logger }) {
  // ...
}
```

(It looks like we can even rewrite `--write` and `--list-different` as reporters.)

j-f1:
I’m wondering if `--check` should display a diff, as requested in #4612.

ikatyang:
I meant that the reporter could be a sharable module just like `eslint --format <my-format>`. And of course there should be some built-in reporters.

> Optionally, we can park the concept of reporters till 2.0 as suggested by @lydell in https://github.com/prettier/prettier/issues/5520#issuecomment-441087896 and simply focus on giving users a simple working solution.

I'm fine with either way as I think we probably need to rewrite our API/CLI in 2.0, and it does not look like it'll happen soon.

kachkaev:
@j-f1 good point about the `--diff` – @ikatyang's reporter suggestion makes even more sense with it!

> @ikatyang:  
> I'm fine with either way as I think we probably need to rewrite our API/CLI in 2.0

2.0 feels like a good time to reconsider the API design, so perhaps the introduction of custom / built-in reporters can be done later. Showing diffs fits the idea of reporters quite well, but I'm not sure that it should be enabled by default. I was recently helping to switch one rather large project into Prettier, which involved adding [`prettier-check`](https://www.npmjs.com/package/prettier-check) to the pipeline. If we were shown Prettier diffs during the transition period, a message saying _Forgot to run Prettier?_  would turn into a needle in a massive haystack 😅

The main use case for `--check` as I see it is in CI to avoid merging unformatted code into master. Making a simple variation of `--list-different` should add enough value for now IMO as it will propagate the right habit across the community. In 2.0, `--check` and `--list-different` can be deprecated due to a shift to the reporters API, but either way this won't be a painful shift.

Let's wait for objections to what I wrote in the docs for a few more days and I'll go for implementing a minimal change, which will make this feature real 😉

docs/cli.md
```diff
@@ -21,6 +21,36 @@ Don't forget the quotes around the globs! The quotes make sure that Prettier exp
...
+Console output if some of the files require re-formatting:
+
+```
+Forgot to run Prettier? There are files without correct code style:
+src/fileA.js
+src/fileB.js
+```
+
+The command will return exit code 1 in the second case, which is helpful inside the CI pipelines.
+Human-friendly status messages make help project contributors react on possible problems.
```
lydell:
```suggestion
Human-friendly status messages help project contributors react on possible problems.
```

lydell:
I'm :+1: on the stuff described in the docs.

kachkaev:
RFC 🕺 A few notes:

I had to change the proposed output from

```
Forgot to run Prettier? There are files without correct code style:
src/fileA.js
src/fileB.js
```

to

```
Checking formatting...
src/fileA.js
src/fileB.js
Code style issues found in the above file(s). Forgot to run Prettier?
```

This because the file list is constructed asynchronously and it would be challenging to "hold" it until all formatting done in order to print the summary message above. The docs are updated accordingly.

---

The implementation of `--check` is rather simple and I would even say boring. No massive refactoring or new abstractions, only minimal necessary changes to do what the docs promise. Hope that this is fine, especially given a possibility of switching to reporters in 2.0.

Tests are also quite boring – I just duplicated the ones for `--list-different`. There is just one new unique test – it makes sure that `--check` and `--list-different` cannot be used together.

---

I noticed a tiny glitch in [`output with --list-different + unformatted differs when piped`](https://github.com/prettier/prettier/blob/e12cd17be2d80585d7b9fc9710e0fdbfd6bc3c2e/tests_integration/__tests__/piped-output.js#L5) – the test was returning [`unformatted.jsunformatted.js`](https://github.com/prettier/prettier/blob/6c482e2e6c4d771407302fcc999bb3e2b491cbd3/tests_integration/__tests__/__snapshots__/piped-output.js.snap#L20). This is now fixed.

kachkaev:
Would it be reasonable to assign this feature to [1.16 milestone](https://github.com/prettier/prettier/milestone/22)? 🤔 It'd be great if it was released together with the next batch of perks! ✌️

kachkaev:
Good point @j-f1! I've implemented this.

tests_integration/__tests__/__snapshots__/piped-output.js.snap
```diff
@@ -12,6 +32,44 @@ exports[`no file diffs with --list-different + formatted file (write) 1`] = `Arr
 
 exports[`no file diffs with --list-different + formatted file (write) 2`] = `Array []`;
 
+exports[`output with --check + unformatted differs when piped (stderr) 1`] = `""`;
+
+exports[`output with --check + unformatted differs when piped (stderr) 2`] = `""`;
+
+exports[`output with --check + unformatted differs when piped (stdout) 1`] = `
+"Checking formatting...
+unformatted.jsunformatted.js
```
kachkaev: One thing I don't quite understand is why `{ stdoutIsTTY: true }` does not produce a truly TTY behaviour (there are three cases similar to this). Here, the first instance of `unformatted.js` should have been deleted by `readline.clearLine(process.stdout, 0); readline.cursorTo(process.stdout, 0, null);`, but this does not happen. When using the same options in a macOS terminal, all works fine. 🤔

tests_integration/__tests__/__snapshots__/piped-output.js.snap
```diff
@@ -1,5 +1,25 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`no file diffs with --check + formatted file (stderr) 1`] = `""`;
+
+exports[`no file diffs with --check + formatted file (stderr) 2`] = `""`;
+
+exports[`no file diffs with --check + formatted file (stdout) 1`] = `
+"Checking formatting...
+formatted.jsAll matched files use Prettier code style!
```
kachkaev: one more case

tests_integration/__tests__/__snapshots__/piped-output.js.snap
j-f1: Looks like the mocked stdout [only supports text being appended](https://github.com/kachkaev/prettier/blob/215ff43a4713878146936e30fa25cd25f3d54844/tests_integration/runPrettier.js#L32).

kachkaev: OK, thanks! Looks like we can leave things as is then.

kachkaev:
We can probably remove status:wip if this helps with getting more reviews or with merging 😉

src/cli/util.js
```diff
@@ -535,10 +533,38 @@ function formatFiles(context) {
...
+    );
+  }
+
+  // Reset exitCode to 0 when using check/list-different is combined with write
+  if (
+    (context.argv["check"] || context.argv["list-different"]) &&
+    context.argv["write"] &&
+    process.exitCode === 1
+  ) {
+    process.exitCode = 0;
```
ikatyang: Using `process.exitCode` to identify if it's formatted does not look like a good idea (especially we did not define what's the meaning of every exit code, and it's not a single-way mutation), can you change it to use a non-global variable?

kachkaev: Hi @ikatyang 👋 OK, will do. Just curious: how many days / weeks / months do I roughly have before 1.16 doors a closing? Need to plan stuff or the NY break.

ikatyang: At least 3 weeks I guess. (We'll spend probably a week writing/reviewing the blog post so you'll get the notification one week before the release.)

kachkaev: @ikatyang this should be done now (I actually managed to implement a quick patch during lunch 🖥🍴) WDYT?

BTW: Thanks to adding `numberOfUnformattedFilesFound`, it is now possible to write  
  `Code style issues found in the above file. Forgot to run Prettier?`  
  `Code style issues found in the above files. Forgot to run Prettier?`  
instead of  
  `Code style issues found in the above file(s). Forgot to run Prettier?`  

Despite that this would look nicer and would not require much effort to implement, I've still decided to keep things as is. The reason is that there is a chance that someone might copy the plural form into another tool and check only against it. This can work fine in tests, but may fail in some real word cases. Prettier's CLI output is facing developers and developer tools, not the end-users, so this linguistic compromise looks justified to me. Let me know if you have a different opinion 😉

src/cli/util.js
```diff
@@ -146,8 +146,8 @@ function listDifferent(context, input, options, filename) {
     if (!prettier.check(input, options)) {
       if (!context.argv["write"]) {
         context.logger.log(filename);
-        process.exitCode = 1;
       }
+      process.exitCode = 1;
```
ikatyang: This one looks like an unnecessary change.

kachkaev: Agree. Fixed!

src/cli/util.js
j-f1: IMO the `--check` option is more for end-users rather than tooling.

kachkaev: I agree that `--check` is designed more for users than for the tools, but we can't be 100% certain. Having `(s)` in the end of a word is not that uncommon for messages that are coming from various dev tools, so I don't see that making the language look 100% fluent is worth the effort and risks. If we wanted to go for it, we'd need extra tests and probably an extra example in docs, but this would not add any measurable value to the end users after all.

I could find `(s)` in a [recent blog post](https://github.com/prettier/prettier/blame/master/website/blog/2018-11-07-1.15.0.md#L332), where I'd expect the language bar to be higher. If that was considered acceptable, I believe we can apply the same trick in stdout too 😉

ikatyang:
Thanks!

daKmoR:
this is not released yet right?

it's a little confusing as it's already shown at https://prettier.io/docs/en/cli.html - I just wanted to set this up and found out that it's not working

lydell:
That’s true, we should comment out the docs.

kachkaev:
@daKmoR thanks for reporting this problem with the docs – it's truly confusing. There is a temporary fix in https://github.com/prettier/prettier/pull/5671 as well as a discussion of how to prevent the problem in future.

PR #5633, opened 2018-12-13
Implement `jsxBracketSpacing` option

### LINKED ISSUE #2651: Feature request - bracket spacing inside JSX
Hi,
I just set up prettier and it's amazing, thanks for the work!

I noticed that bracket spacing only works on object literals.
Is it possible to add bracket spacing inside JSX as well? right now this
`<div>{ foo }</div>` would result in `<div>{foo}</div>`

Thanks

pugnascotia:
Closes #2651. This PR adds the `jsxBracketSpacing` option, which pads JSX attributes and expression containers with either spaces or newlines, depending on whether the content breaks. It does this by switching between `line` and `softline`, depending on the option value.

Note that in the handling for `JSXExpressionContainer`, the decision whether to inline the expression also considers this option, and does not inline if `jsxBracketSpacing` is true. I did this after examining the output, and deciding what would be prettier. The term "inline" in this context is bit misleading, since the content can still break.

In order to update the tests, I defaulted the option to true, noted the failing tests, then turned the option off again and added extra cases to the specs with the option enabled. Repeating this process may be the simplest way of reviewing the impact of this switch.

I did not update the Flow tests. I did run the sync script but it resulted in a vast number of changes.


duailibe:
Hi @pugnascotia!

Your code changes look great to me, and could be merged right away, but I vote against adding this feature to Prettier.

And since this is not a big change, it looks like it wouldn't be hard to maintain a fork with these changes, which is what I recommend.

I'm sorry your first contribution to Prettier might not be merged, but I hope this won't stop you in the future.

pugnascotia:
OK, I understand the reasons. I'm disappointed, because I hoped to address the current inconsistency with `bracketSpacing`, because I can see that this has been requested several times, and because I don't want to also maintain forks of linting plugins and TypeScript definitions.

If the community decides to reverse this decision, I'll be happy to provide the PR.

pugnascotia:
I've published `@pugnascotia/prettier` in case anyone's interested.

PR #5643, opened 2018-12-15
Tests the right directory

arcanis:
This is just a simple fix to ensure that prettier tests that the actual `node_modules` folder is a folder, rather than the parent folder. It doesn't change anything in 99.99% of the cases, but it doesn't hurt 🙂

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -53,7 +53,7 @@ function loadPlugins(plugins, pluginSearchDirs) {
 
       if (!isDirectory(nodeModulesDir)) {
         throw new Error(
-          `${pluginSearchDir} does not exist or is not a directory`
+          `${nodeModulesDir} does not exist or is not a directory`
         );
       }
 
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -46,16 +46,16 @@ function loadPlugins(plugins, pluginSearchDirs) {
         pluginSearchDir
       );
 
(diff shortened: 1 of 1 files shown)


arcanis:
Thinking more, this change will cause issues when the `node_modules` doesn't exist at all ...

kachkaev:
This change will break `--plugin-search-dir` option. What is the use case you are trying to solve?

duailibe:
If I had to guess, it's because of https://github.com/yarnpkg/rfcs/pull/101 :-)

arcanis:
I have a fringe case where a the parent folder of a folder is not a folder (I know) 😅

But yeah this diff isn't good enough yet, I'll think about something better (in the meantime I can just manually pass `/tmp` to pluginSearchDirs)

arcanis:
@kachkaev @duailibe I got around giving it another look and made the following change (it was pretty simple in the end):

https://github.com/prettier/prettier/compare/master...arcanis:patch-2

Does that look ok with you?

Basically with Yarn v2 we'll make it possible to load files directly from within the zip archive. For now this will be implemented by patching the `fs` module so that Node is able to see the files within the archive. The problem is that in order to work well with `rimraf` and similar crawlers, the archive itself mustn't be stat'd as a directory. This is problematic in Prettier because you actually check it by default.

With this change the node_modules isDirectory check will succeed, and will shortcut the one to check whether its parent directory is a directory.

duailibe:
@arcanis Yes, I think that solution makes sense! If you'd like to reopen, I can merge it

PR #5668, opened 2018-12-20
fix(angular): Format <pre> in template strings properly.

ajcrites:
The current logic doesn't take into account <pre> in a template string. This updates the logic to add additional base indentation to compare the alignment of the opening and closing pre tags if it's in a template string since it should be indented twice (once for the indentation of the template property itself, and once for the indentation of the string contents.

Because of the way <pre> works, this will only work properly if the closing </pre> tag was already indented to where it was supposed to be. Otherwise, the closing caret is dropped to the newline to align with the opening tag in order to keep whitespace consistent.

### DIFF
--- a/src/language-html/printer-html.js
+++ b/src/language-html/printer-html.js
@@ -244,10 +244,11 @@ function genericPrint(path, options, print) {
                           node.isIndentationSensitive)) &&
                       new RegExp(
                         `\\n\\s{${options.tabWidth *
-                          countParents(
+                          (countParents(
                             path,
                             n => n.parent && n.parent.type !== "root"
-                          )}}$`
+                          ) +
+                            (options.parentParser ? 2 : 0))}}$`
                       ).test(node.lastChild.value)
                     ? /**
                        *     <div>
--- a/tests/angular_component_examples/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/angular_component_examples/__snapshots__/jsfmt.spec.js.snap
@@ -10,6 +10,13 @@ printWidth: 80
        selector: 'app-test',
   template: \`<ul>   <li>test</li>
   </ul>
+
+  <div>
+    <pre>
+        test
+      </pre>
+    ~
+</div>
   \`,
   styles: [   \`
   
@@ -31,6 +38,13 @@ class     TestComponent {}
     <ul>
       <li>test</li>
     </ul>
+
+    <div>
+      <pre>
+        test
+      </pre>
+      ~
+    </div>
   \`,
   styles: [
     \`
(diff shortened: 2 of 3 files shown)


src/language-html/printer-html.js
```diff
@@ -244,10 +244,11 @@ function genericPrint(path, options, print) {
                           node.isIndentationSensitive)) &&
                       new RegExp(
                         `\\n\\s{${options.tabWidth *
-                          countParents(
+                          (countParents(
                             path,
                             n => n.parent && n.parent.type !== "root"
-                          )}}$`
+                          ) +
+                            (options.parentParser ? 2 : 0))}}$`
```
ikatyang: We need to figure out a way to access the indentation information from the parent context since the parent context can be different, for example:

**Prettier pr-5668**
[Playground link](https://deploy-preview-5668--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEALGBbANgAwDpQAEhAPAA4BOcAfAccQB52kD0lNB+UBAlgGaEAFDAoBXOAEpCwZumxd6pdrSKLCTVcRJsqK4lwC+BXgOFjJ05vyEjxUmZsJzczReV2u1Gta2WeFhEZQBiAANCAQZDA80ADOyKAAhhQUEADuAArJCPEoiQBuEDwAJmEgAEYUiWAA1nAwAMpk1TxQAObItnDhzgDqqDzwsc1gcA05gzz5gwCeyOCx8eGtsXAUMBlVbRiJyHyJWKvhAFaxDABCVbX1DYkYcAAyrXB7B0cgpwwNrW1YcACKogg8Feh26IGaFFWFHmMBmZDgsTAFB4UTKlFaMF6JRgqGQAA4AAzhSgQVa9KpkebsaH5F7hKgAR1EPCom0S212SH2YPCqwwPE65j5Pz+gOBL25b3BMES5WxxVxyAATOERIkeFgfgBhCAYHbzKDQekgUSrAAqctyPNWBgMQA)
```sh
--parser typescript
```

**Input:**
```tsx
html`
  <pre>
    x
  </pre>
`

if (true) {
  html`
    <pre>
      x
    </pre>
  `
}

if (true) {
  if (true) {
    html`
      <pre>
        x
      </pre>
    `
  }
}
```

**Output:**
```tsx
html`
  <pre>
    x
  </pre
  >
`;

if (true) {
  html`
    <pre>
      x
    </pre>
  `;
}

if (true) {
  if (true) {
    html`
      <pre>
        x
      </pre
      >
    `;
  }
}

```

lipis: Any ideas?

PR #5671, opened 2018-12-21
Temporary remove --check docs until 1.16 is released to avoid confusion

kachkaev:
Context: https://github.com/prettier/prettier/pull/5629#issuecomment-448933238

It'd be great if the current docs were based on the current version of prettier to avoid confusion. How about using `master` branch for patches / current docs and `canary` for the future stuff? Next.js folks do that (https://github.com/zeit/next.js/).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,7 +18,7 @@ prettier.format("foo ( );", { semi: false, parser: "babylon" });
 
 ## `prettier.check(source [, options])`
 
-`check` checks to see if the file has been formatted with Prettier given those options and returns a `Boolean`. This is similar to the `--check` or `--list-different` parameter in the CLI and is useful for running Prettier in CI scenarios.
+`check` checks to see if the file has been formatted with Prettier given those options and returns a `Boolean`. This is similar to the `--list-different` parameter in the CLI and is useful for running Prettier in CI scenarios.
 
 ## `prettier.formatWithCursor(source [, options])`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -21,38 +21,6 @@ Don't forget the quotes around the globs! The quotes make sure that Prettier exp
 
 Prettier CLI will ignore files located in `node_modules` directory. To opt-out from this behavior use `--with-node-modules` flag.
 
-## `--check`
-
-When you want to check if your files are formatted, you can run Prettier with the `--check` flag (or `-c`).
-This will output a human-friendly message and a list of unformatted files, if any.
-
-```bash
-prettier --check "src/**/*.js"
-```
-
-Console output if all files are formatted:
-
-```
-Checking formatting...
-All matched files use Prettier code style!
-```
-
-Console output if some of the files require re-formatting:
-
-```
-Checking formatting...
-src/fileA.js
-src/fileB.js
-Code style issues found in the above file(s). Forgot to run Prettier?
-```
-
-The command will return exit code 1 in the second case, which is helpful inside the CI pipelines.
-Human-friendly status messages help project contributors react on possible problems.
-To minimise the number of times `prettier --check` finds unformatted files, you may be interested in configuring a [pre-commit hook](precommit.md) in your repo.
-Applying this practice will minimise the number of times the CI fails because of code formatting problems.
-
-If you need to pipe the list of unformatted files to another command, you can use [`--list-different`](./cli.md#list-different) flag instead of `--check`.
-
 ## `--debug-check`
 
 If you're worried that Prettier will change the correctness of your code, add `--debug-check` to the command. This will cause Prettier to print an error message if it detects that code correctness might have changed. Note that `--write` cannot be used with `--debug-check`.
@@ -103,8 +71,6 @@ Another useful flag is `--list-different` (or `-l`) which prints the filenames o
 ```bash
 prettier --single-quote --list-different "src/**/*.js"
 ```
-
-You can also use [`--check`](./cli.md#check) flag, which works the same way as `--list-different`, but also prints a human-friendly summary message to stdout.
 
 ## `--no-config`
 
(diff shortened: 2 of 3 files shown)


kachkaev:
There must be various ways of doing so I believe. E.g. there can be a `canary` branch, into which all PRs get merged. This branch is merged into `master` when a new feature release is baked. Meanwhile, if there is any patch that applies to the current version of Prettier, a PR commit is cherry-picked by a maintainer from `canary` into `master`. This updates the docs immediately. Alternatively, 'patch' PRs can be created against `master` instead of `canary` (maintainers can change the target branch if it's wrong). In this case, `master` needs to be merged into `canary` from time to time to keep them sync.

With a `canary` branch that represents upcoming features, there is a possibility to have https://canary.prettier.io/ with its Playground and other perks.

ikatyang:
We could use [versioned docs](https://docusaurus.io/docs/en/versioning) to achieve this goal, see [here](https://docusaurus.io/en/versions) for example. (We don't need to add past versions, _stable_ and _next_ should be enough.)

daKmoR:
I would totally agree that a separate branch for each major version + dev/canary is a very good idea.

I would even go on step further and say every merge you do on master should be auto-released - e.g. you merge a bugfix => get's released.
you merge canary/or any ready feature branch => get's released.

canary would probably be too noisy for that => so merge without release and schedule a release every night.

I think something like this would be the ideal release system.

ikatyang:
@daKmoR Installing from GitHub is already supported, which means you can simply `npm install prettier/prettier` to get the "nightly" release.

kachkaev:
@ikatyang if master plays the role of `canary`, then there is a need for another branch (say, `stable` or `current`) for bugfixes and and minor doc improvements. This is because ongoing pull requests have two distinct destinations.

I'm more in favour of `canary` / `master` name pair instead, but that would require people to `npm install prettier/prettier#canary` if they want to try out bleeding edge versions from github. What is cool about `canary` as name is that it can be also used on `npm` for automatic nightly releases (`npm install prettier@1.16.0-canary.20181221` / `npm install prettier@canary`). The third usage for this name would be: https://canary.prettier.io/. `master` instead of `canary` would look slightly less self-explanatory in the above cases.

If `canary` does not feel like the right word to use, `dev` or `next` can work as a replacement synonym.

Here is an example of a different approach: https://github.com/containous/traefik. They use `master` for the 'future' version and for each of the `x.x` versions they use their own branch. If the current latest version of Prettier is `1.15.3`, patches would go to branch `v1.15`and new features would go to `master`. Time and again commits from `v1.15` would be cherry-picked to `master` ([example](https://github.com/containous/traefik/pull/4303)) and one `v1.16.0` is released, branch `v1.16` is created from `master`. This is more complex than just two branches, but makes it possible to patch multiple versions if needed (not sure this is valuable for Prettier though).

One way or another, it would be great to see the development and bugfix work to be split between branches to avoid rewinding upcoming or keeping feature pull requests open till the last minute. The new workflow will also make patch releases much easier. Imagine someone discovering an urgent bug tomorrow, which would require releasing `1.15.4` immediately. `master` already contains https://github.com/prettier/prettier/pull/5629, so it will be necessary to manually go through commits after `1.15.3`, cherry-pick them into a temporary branch and release from it. This feels somewhat convoluted and ad hoc.

Meanwhile, I'll prepare a PR that reverts this PR that reverts docs in PR #5629 that confused @daKmoR 😅

ikatyang:
@kachkaev 

> if master plays the role of canary, then there is a need for another branch (say, stable or current) for bugfixes and and minor doc improvements. This is because ongoing pull requests have two distinct destinations.

While it sounds like a good idea, our website is currently deployed from master via Netlify and I'm not sure if deploying from branches is supported. And I think it'd be better to use [versioned docs](https://docusaurus.io/docs/en/versioning) since we could modify both the current and the `next` docs in the same PR.

Regarding bugfixes, it's already the case, though we only did it if it's necessary, see [`release-1.14`](https://github.com/prettier/prettier/tree/release-1.14) branch for example.

> it can be also used on npm for automatic nightly releases (npm install prettier@1.16.0-canary.20181221 / npm install prettier@canary)

While it's a nice thing to have, disabling OTP (which is required for nightly build) does not look like a good idea.

> One way or another, it would be great to see the development and bugfix work to be split between...

Our current versioning is not that strict, we could try a stricter one in the next release, that is, patch releases are only used for significant bugfixes (which usually requires `cherry-pick`), everything else will go to the next minor version.

daKmoR:
> While it sounds like a good idea, our website is currently deployed from master via Netlify and I'm not sure if deploying from branches is supported.

it is supported... it's actually what happens with every merge request - when you get the preview URL.

> And I think it'd be better to use versioned docs since we could modify both the current and the next docs in the same PR.

true, but what would be the usecase? updating the docs of 1.x and 2.x at the same time will mean also releasing at the same time and having only one release branch... makes it kinda tough.

Most bigger projects have branches for each major version so I guess you will need that at some point anyways and having a "next" branch (seems to be currently the most common name for it) would be just one other release branch. You could use it for nightly builds but as said before you can skip that and say install via `npm install prettier/prettier#next`.

I would even say that's easier to understand... 
- `npm i prettier` => you get the latest released version from npm (might not have all files because of .npmignore)
- `npm i prettier/prettier` => get latest master version from git (same version as npm but all files)
- `npm i prettier/prettier#next` => get next version from git (possibly unstable)

Overall I would still say the benefits for multiple release branches are worth the extra effort.

PS: not every PR needs to go next -> master; you can go to master directly and then rebase
PPS: Not a problem for prettier but if you have a monorepo then you sort of need nightly builds as you can't just install from a git repository (as you can't install only subdirectories)

ikatyang:
> it is supported... it's actually what happens with every merge request - when you get the preview URL.

Thanks, I just found it on Netlify:

image

> true, but what would be the usecase? updating the docs of 1.x and 2.x at the same time will mean also releasing at the same time and having only one release branch... makes it kinda tough.

I meant the next minor version, not the major one. Of course the next major version should be in a separate branch.

Issue #5654, opened 2018-12-18
Prettier format url path with number incorrectly


jasonxia23:
**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAjCATAngAmAHShx1QEMwBrAcwCcIBXKDJHemgGwAoALGGAByQB6IQEYATAGYAdHAAepALb92caZEUBKANyEAviAA0ICPxgBLaAGdkoUjToB3AAr2ENlKQBuEcxiNoNOQUcDAAyvzk5lBUyDA09HDGvIrsAOrc5vBWkWBwYe5Z5l5ZWMjgVjbG0VZwNDDOQVSKpMgAZqTstcYAVlZyAEJBlKFhSnAAMtFw7Z3dIH1yYdFUqgCK9BDws11JIJE0tTTlYJUB-DTRMGl+MNzIABwADMYXELVpQfzlF3BHXjNjDQ4ABHejmYGNUjNVpIDq7Yy1RTmOIJPZWFbrTbbOFzPYwUioG4YO7IcTGeKkczsFYAYQgiha5Sg0EBIHotQAKoSPPDano9EA)
```sh
--parser css
```

**Input:**
```css
body {
  background: url(http://123.example.com);
}
```

**Output:**
```css
body {
  background: url(http://123example.com);
}

```

**Expected behavior:**

ikatyang:
As a workaround, you could use quoted url for now:

**Prettier 1.15.3**
[...]

PR #5692, opened 2018-12-28
fix: point followed number in css url function be collapsed

jasonxia23:
https://github.com/prettier/prettier/issues/5654

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -40,8 +40,17 @@ function parseValueNodes(nodes) {
       node.group.groups[0].groups[0].value === "data" &&
       node.group.groups[0].groups[1].type === "colon" &&
       node.group.groups[0].groups[1].value === ":";
+    const isUnquotedURLWithNumber =
+      node.type === "func" &&
+      node.value === "url" &&
+      node.group &&
+      node.group.groups &&
+      node.group.groups[0] &&
+      node.group.groups[0].groups &&
+      node.group.groups[0].groups.length > 2 &&
+      node.group.groups[0].groups.find(group => group.type === "number");
 
-    if (isUnquotedDataURLCall) {
+    if (isUnquotedDataURLCall || isUnquotedURLWithNumber) {
       node.group.groups = [stringifyGroup(node)];
     }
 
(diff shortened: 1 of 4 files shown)


src/language-css/parser-postcss.js
```diff
@@ -40,8 +40,17 @@ function parseValueNodes(nodes) {
...
       node.group.groups[0].groups[1].value === ":";
+    const isUnquotedURLWithNumber =
+      node.type === "func" &&
+      node.value === "url" &&
+      node.group &&
+      node.group.groups &&
+      node.group.groups[0] &&
+      node.group.groups[0].groups &&
+      node.group.groups[0].groups.length > 2 &&
+      node.group.groups[0].groups.find(group => group.type === "number");
```
evilebottnawi: I think better implement utils `isValidURL` and print all valid url in one line, it is solve many problem, not only with numbers, anyway good work!

jasonxia23: Good point. Could you please figure out any other cases?

jasonxia23: @evilebottnawi You mean extract `isUnquotedDataURLCall` and `isUnquotedURLWithNumber` to `isValidURL ` util? Maybe I haven't get your point.

evilebottnawi: @jasonxia23 i thinks we need get logic from here https://nodejs.org/docs/latest/api/url.html

> A TypeError will be thrown if the input or base are not valid URLs. Note that an effort will be made to coerce the given values into strings. For instance:

```js
const myURL = new URL({ toString: () => 'https://example.org/' });
// https://example.org/
```

And it is valid url we stringify group

jasonxia23: Well, I think CSS `<url>` is not same as Javascript `URL`.

For example, I can use `url(foo.png)` but `foo.png` is not "VALID" for `URL`

lydell: Technically, at the parsing level the stuff inside `url()` does not have to be a valid URL: https://www.w3.org/TR/css-syntax-3/#url-unquoted-diagram

evilebottnawi: @jasonxia23 let's use logic for check what @lydell write above, it is good solution

jasonxia23: @evilebottnawi Maybe I should turn to `postcss-values-parser`. Prettier uses `postcss-values-parser` other than `postcss-value-parser`.

fisker:
Thank you for your contribution, the original issue is fixed by #7592.

Your tests added in #8482.

Issue #5603, opened 2018-12-07
SCSS single-line comments are formatted incorrectly


d0x:
**Important Note:**
Prettier has nice support for Grid-Template-Areas . See [here](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA6A5gJwJYBMAEwAOlPvlngLTwC2ADgDYCG8lTmcTAzkiWWURAALTrjiYyIpmIn4pMwX36CueOACN2+SLAQxt0eLEWllIAGYQI8WZeviydm4IDcJAL4gANCAh0Y2NBcyKDsmBAA7gAK7AjBKEwAbhB43iDqmExgANZwMADKdFnYUOjIMJgArnA+QjA0DADqQtjwXEVgcPlxrdiJrQCeyOBcwT4lXOIwUZnoNEzI5kwMkz4AVlwAHgBCmTl5+Uw0cAAyJXCLy6sgG5v5JegMcACKlfaXKzUgRZiTmMNcMCjNJ0HCwRp4GBCZAADgADD5QRBJo1MnRhqC4H9EhcfBwAI6VbAcGZMOYLJBLT4+SY0bDlKpfVSlJ6vd6Uq5fGBMdQQ3BQ5AAFh8FSY2AYDwAwhAaPNhlBoLiQJVJgAVHnxKnXbgwD7XXAQMB6r4QSowOhmgBMxvc7iAA )

The example shows only how to reproduce the issue with single-line comments!

**Prettier 1.15.3**
[...]

PR #5710, opened 2019-01-03
Fix formatting of lists in SCSS property/variable values

jsnajdr:
Fixes the following bug:

**Input:**
```
$my-list2:
  a // a
  b
  c;
```

**Actual Output:**
```
$my-list2: a// a
  bc;
```
(no space before comment, values after the comment glued together)

**Expected Output:**
```
$my-list2: a // a
  b c;
```

The cause of the bug is that the printer doesn't add a space after a node if the previous or
the next node is an inline comment. There's no reason to do that. The only case that warrants
a special behavior is when the current node is an inline comment. Then the printer should add
a hardline after it.


jsnajdr:
Hi @evilebottnawi, the formatting of your example is broken:
```
$my-list3: // b
  // extra a // a
  // extra
  b // b
  c;
```
The `a` token becomes a part of inline comment. That's already reported in #5603 and this PR is not related to the issue: it doesn't fix it and it doesn't introduce any new issues either.

I think it's not a good idea to add tests for things that are broken. Is there a reason why you'd want to add the test anyway?

ikatyang:
Thanks!

Issue #12, opened 2017-01-10
Add option to omit semicolons in output


iki:
E.g. `semicolons: true` to use semicolons at the end of statements (default), and `semicolons: false` to not use semicolons.

rogeliog:
@jlongster If you are both accepting PRs and want this feature, I would be more than happy to help with a PR for this one.

jlongster:
I'm open to it, but it gets problematic fast. When you don't use semicolons, you run into things like this where you are forced to put a semicolon:

```js
var x = 5

;[1, 2, 3].forEach(...)
```

The first time this would be parsed correctly, but we would output:

```
var x = 5

[1, 2, 3].forEach(...)
```

which is invalid JS. (at least, not what you want.) This means that in order to support semicolon-less JS we need to code in all the ASI rules so we can see where semicolons are important. I'm not sure I'm willing to take on that complexity just yet.

Feel free to play around with it. Code is powerful and you might convince me. My preference for evolving projects is to take it slowly though, and it's hard to remove things later than to add them, so let's play with it over time and see what we can do. I'd like to keep the number of options low, but this is such a divisive issue that it could be worthwhile.

gaelduplessix:
Have two suggestions:
- what about making this the default (and only) behavior ? I agree [configuration options should be avoided](https://github.com/jlongster/prettier/issues/40) so why not considering making this the standard ?
- how about auto-detecting depending on the existing code ? If it has semi-colons, output semi-colons, if not, do not output them ?

swansontec:
I use [standard.js](standardjs.com), which is a popular no-configuration style guide & linter. They insist on a semicolon-free style, and the only rule needed to make this work is to [never start a line with `(`, `[`, or `](http://standardjs.com/rules.html#semicolons).

K-Leon:
I would love seeing this integrated - is currently the only show stopper for me.

bakkot:
It's worth mentioning that "the only rule needed to make this work" will get progressively more complex as the language evolves. For example, under popular [the public fields proposals](https://github.com/tc39/proposal-class-public-fields) these two classes differ:

```js
class A {
  set
  foo(x){} // setter for 'foo'
}

class B {
  set; // uninitialized field named 'set'
  foo(x){} // method named 'foo'
}
```

(Note that babylon currently gets these edge cases wrong [in a number of exciting ways](https://github.com/babel/babylon/pull/158).)

Presumably a no-semi style would have you writing

```js
class B {
  set
[...]

PR #5723, opened 2019-01-05
Introduce standard.js compatibility flag

sheerun:

Hello prettier,

I am the author of [prettier-standard](https://github.com/sheerun/prettier-standard) library that tries to be similar to prettier while maintaining compatibility with [standard](https://standardjs.com/) that has been present for some time before prettier. Standard is still used by many developers and its rules are preferred by many developers (equivalents of --no-semi --single-quote --jsx-single-quote like + not implemented --space-before-function-paren).

While I enjoy github stars \s prettier-standard is not a perfect solution because it needs to pass every file first through prettier and then though eslint with standard whitespace rules. This makes this process very slow. Ideally this functionality would be embedded within prettier itself so developers can just write `prettier --standard` instead of `prettier-standard`. I would gladly deprecate my package and point to new prettier version if you agreed to merge this PR.

[There's](https://github.com/prettier/prettier/issues/3845) [discussion](https://github.com/prettier/prettier/issues/3847) [whether](https://github.com/prettier/prettier/issues/1139) introduce extra flag for inserting space before function name or anonymous function, this won't make prettier compatible with standard as there are many more nuances like spaces around generator stars. This PR introduces `--standard` flag that has single purpose: make prettier format code as such running `standard` after it won't complain about formatting of code. I think it's good measure to not introduce many flags into prettier and at the same time satisfy people who even start to [fork](https://github.com/brodybits/prettierx) [this](https://github.com/arijs/prettier-miscellaneous) project.

I've gathered all whitespace-related rules of standard and added tests for each of them from straight from eslint docs (both correct ones and incorrect ones). I've also included compatibility with the way standard wants to format typescript and flow. You can verify that with this flag formatting is compatible with standard by running following commands or visually inspecting new tests:

```bash
# verifies that es6 example is compatible with standard (nothing changes)
standard --parser babel-eslint --fix tests/standard/correct.js # nothing changes
diff tests/standard/correct.js <(node bin/prettier.js --standard tests/standard/correct.js)
```

```bash
# verifies that typescript example is compatible with standard (nothing changes)
standard --parser typescript-eslint-parser --plugin typescript --fix tests/standard/typescript/correct.ts 
diff tests/standard/typescript/correct.ts <(node bin/prettier.js --standard tests/standard/typescript/correct.ts)
```

```bash
# verifies that flow example is compatible with standard (nothing changes)
standard --parser babel-eslint --plugin flowtype --fix tests/standard/flow/correct.js
diff tests/standard/flow/correct.js <(node bin/prettier.js --standard tests/standard/flow/correct.js)
```

You can find more information how to configure standard for flow & typescript on their home page.

What do you think? This could potentially be last and final flag to satisfy standard.js camp.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -77,6 +77,14 @@ Use single quotes instead of double quotes in JSX.
 | ------- | -------------------- | ------------------------ |
 | `false` | `--jsx-single-quote` | `jsxSingleQuote: <bool>` |
 
+## Standard.js compatibility
+
+Configure prettier to be compatible with Standard.js
+
+| Default | CLI Override | API Override       |
+| ------- | ------------ | ------------------ |
+| `false` | `--standard` | `standard: <bool>` |
+
 ## Trailing Commas
 
 Print trailing commas wherever possible when multi-line. (A single-line array, for example, never gets trailing commas.)
--- a/src/language-js/options.js
+++ b/src/language-js/options.js
@@ -72,5 +72,12 @@ module.exports = {
       { value: true, deprecated: "0.19.0", redirect: "es5" },
       { value: false, deprecated: "0.19.0", redirect: "none" }
     ]
+  },
+  standard: {
+    since: "1.16.0",
+    category: CATEGORY_JAVASCRIPT,
+    type: "boolean",
+    default: false,
+    description: "Use standard rules for formatting."
   }
 };
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -916,6 +916,9 @@ function printPathNoParens(path, options, print, args) {
       parts.push("yield");
 
       if (n.delegate) {
+        if (options.standard) {
+          parts.push(" ");
+        }
         parts.push("*");
       }
       if (n.argument) {
@@ -3723,6 +3726,9 @@ function printMethod(path, options, print) {
   if (!kind || kind === "init" || kind === "method" || kind === "constructor") {
     if (node.value.generator) {
       parts.push("*");
+      if (options.standard) {
+        parts.push(" ");
+      }
     }
   } else {
     assert.ok(kind === "get" || kind === "set");
@@ -3744,6 +3750,7 @@ function printMethod(path, options, print) {
           printFunctionTypeParameters(valuePath, options, print),
           group(
             concat([
+              options.standard ? " " : "",
               printFunctionParams(valuePath, print, options),
               printReturnType(valuePath, print, options)
             ])
@@ -4224,6 +4231,9 @@ function printFunctionDeclaration(path, print, options) {
   parts.push("function");
 
   if (n.generator) {
+    if (options.standard) {
+      parts.push(" ");
+    }
     parts.push("*");
   }
   if (n.id) {
@@ -4234,6 +4244,7 @@ function printFunctionDeclaration(path, print, options) {
     printFunctionTypeParameters(path, options, print),
     group(
       concat([
+        options.standard ? " " : "",
         printFunctionParams(path, print, options),
         printReturnType(path, print, options)
       ])
@@ -4254,6 +4265,9 @@ function printObjectMethod(path, options, print) {
   }
   if (objMethod.generator) {
     parts.push("*");
+    if (options.standard) {
+      parts.push(" ");
+    }
   }
   if (
     objMethod.method ||
@@ -4269,6 +4283,9 @@ function printObjectMethod(path, options, print) {
     parts.push("[", key, "]");
   } else {
     parts.push(key);
+    if (options.standard) {
+      parts.push(" ");
+    }
   }
 
   parts.push(
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -34,6 +34,12 @@ const docUtils = doc.utils;
  * the path to the current node through the Abstract Syntax Tree.
  */
 function printAstToDoc(ast, options, alignmentSize = 0) {
+  if (options.standard) {
+    options.semi = false;
+    options.singleQuote = true;
+    options.jsxSingleQuote = true;
+  }
+
   const printer = options.printer;
 
   if (printer.preprocess) {
--- a/tests/standard/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/standard/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,1415 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`correct.js 1`] = `
+====================================options=====================================
+parsers: ["babel"]
+printWidth: 80
+standard: true
+                                                                                | printWidth
+=====================================input======================================
+// Needs to be at the top but belongs to object-curly-spacing
+import { foo } from 'bar'
+
+// Needs to be at the top but belongs to indent rule
+import {
+  foosjdfhalsjkdhflasjdhfajashdlfjahsdlfjashdlfjaksdsadfaso,
+  barsjdfhalsjkdhflasjdhfajashdlfjahsdlfjashdlfjaksdsadfaso,
+  bazsjdfhalsjkdhflasjdhfajashdlfjahsdlfjashdlfjaksdsadfaso
+} from 'qux'
+
+// "arrow-spacing": ["error", { "before": true, "after": true }]
+a => {}
+;() => {}
+a => a
+;() => {
+  '\\n'
+}
+
+// "block-spacing": [ "error", "always" ],
+function foo () {
+  return true
+}
+if (foo) {
+  bar = 0
+}
+
+// "brace-style": [ "error", "1tbs", { "allowSingleLine": true } ],
+function foo () {
+  return true
+}
+if (foo) {
+  bar()
+}
+if (foo) {
+  bar()
+} else {
+  baz()
+}
+try {
+  somethingRisky()
+} catch (e) {
+  handleError()
+}
+
+// "comma-spacing": [ "error", { "before": false, "after": true } ],
+var foo = 1
+var baz = 3
+var arr = [1, 2]
+var arr = [1, , 3]
+var obj = { foo: 'bar', baz: 'qur' }
+foo(a, b)
+new Foo(a, b)
+function foo (a, b) {}
+a, b
+
+// "eol-last": "error",
+// cannot be tested here, but true (Unix-style new lines)
+
+// "func-call-spacing": [ "error", "never" ],
+fn()
+
+// "generator-star-spacing": [ "error", { "before": true, "after": true } ],
+function * generator () {}
+var anonymous = function * () {}
+var shorthand = { * generator () {} }
+class Example {
+  async * foo () {}
+}
+
+// "indent": [
+//   "error",
+//   2,
+//   {
+//     "SwitchCase": 1,
+//     "VariableDeclarator": 1,
+//     "outerIIFEBody": 1,
+//     "MemberExpression": 1,
+//     "FunctionDeclaration": { "parameters": 1, "body": 1 },
+//     "FunctionExpression": { "parameters": 1, "body": 1 },
+//     "CallExpression": { "arguments": 1 },
+//     "ArrayExpression": 1,
+//     "ObjectExpression": 1,
+//     "ImportDeclaration": 1,
+//     "flatTernaryExpressions": false,
+//     "ignoreComments": false
+//   }
+// ],
+
+switch (a) {
+  case 'a':
+    break
+  case 'b':
+    break
+}
+
+var aajshdfaljskdhflakjshdflkjashdlfjkhasldkjfhlasjkdfhlaskjdhfalsjkdfha,
+  bajshdfaljskdhflakjshdflkjashdlfjkhasldkjfhlasjkdfhlaskjdhfalsjkdfha,
+  bajshdfaljskdhflakjshdflkjashdlfjkhasldkjfhlasjkdfhlaskjdhfalsjkdfha
+let aajshdfaljskdhflakjshdflkjashdlfjkhasldkjfhlasjkdfhlaskjdhfalsjkdfha,
+  bajshdfaljskdhflakjshdflkjashdlfjkhasldkjfhlasjkdfhlaskjdhfalsjkdfha,
+  bajshdfaljskdhflakjshdflkjashdlfjkhasldkjfhlasjkdfhlaskjdhfalsjkdfha
+const a = 1
+const b = 2
+const c = 3
+;(function () {
+  function foo (x) {
+    return x + 1
+  }
+})()
+if (y) {
+  console.log('foo')
+}
+function foo (
+  barverylongverylongverylongverylongverylongverylongverylongverylong,
+  bazverylongverylongverylongverylongverylongverylongverylongverylong,
+  quxverylongverylongverylongverylongverylongverylongverylongverylong
+) {
+  qux()
+}
+foo(
+  barverylongverylongverylongverylongverylongverylongverylongverylong,
+  bazverylongverylongverylongverylongverylongverylongverylongverylong,
+  quxverylongverylongverylongverylongverylongverylongverylongverylong
+)
+var foo = [
+  barverylongverylongverylongverylongverylongverylongverylongverylong,
+  bazverylongverylongverylongverylongverylongverylongverylongverylong,
+  quxverylongverylongverylongverylongverylongverylongverylongverylong
+]
+var foo = {
+  barverylongverylongverylongverylongverylongverylongverylongverylong: 1,
+  bazverylongverylongverylongverylongverylongverylongverylongverylong: 2,
+  quxverylongverylongverylongverylongverylongverylongverylongverylong: 3
+}
+var a = barverylongverylongverylongverylongverylongverylongverylongverylong
+  ? bar
+  : barverylongverylongverylongverylongverylongverylongverylongverylong
+
+// "key-spacing": [ "error", { "beforeColon": false, "afterColon": true } ],
+var obj = { foo: 42 }
+
+// "keyword-spacing": [ "error", { "before": true, "after": true } ],
+if (foo) {
+  // ...
+} else if (bar) {
+  // ...
+} else {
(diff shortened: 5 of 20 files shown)


sheerun:
I fixed the tests

Also, one notable project that needs to run [`standard --fix`](https://github.com/zeit/next.js/blob/canary/package.json#L17) for fixing formatting so it's 100% compatible with standard [because](https://github.com/zeit/next.js/pull/5850) prettier doesn't support it is [next.js](https://github.com/zeit/next.js)

sheerun:
Thank you for response.

First of all let me kindly dismiss your argument "look too ugly to me.". As you noticed later it's subjective matter plus provided cases are very rare in real code. Prettier states about itself that it's "opinionated code formatter", it's no different. Standard is just a name. I'm not here to argue which style is better but provide solution for these developers who find standard's syntax better.

I've read this article you provided and it focuses on something that in my long experience is never a problem. It states problem but immediately provides solution (semicolons when line starts with `[`, `(` or tilde, very rare case. It also asks "Moreover, where is code style consistency?", but this style actually has more consistent rules than rules where to put semicolons. 

Again I don't want to argue about which style is better so let's focus on more objective reasons.

> Just curious: what stats could support this claim?

I've [downloaded](https://gist.github.com/sheerun/7af754ab0e73b0b05deaf435abe392a3) 1000 most popular [npm packages](https://gist.githubusercontent.com/anvaka/8e8fa57c7ee1350e3491/raw/27c81e27e0ebd7331db1fd1ecf14f2179530c083/01.most-dependent-upon.md) dependent upon and 154 of them use standard while 125 of them use prettier, which is not ultimate stat but I hope it shows standard isn't a niche and your statement that "majority of popular repos I've looked into do use semicolons" might be true but is not relevant.

Also in my experience many projects use prettier but also use `--no-semi` and `--single-quote` which is closer to standard formatting than prettier's default one. Sorry but I don't have time to provide analytics for this one, do you? 

> Given that Prettier is not just for JS and --standard is JS-specific, I'm not sure how the new option can benefit Prettier users in a general case

Actually most of current prettier options apply only to javascript.

> some of its decisions are based on the reasoning coming from quite a few years ago and so may be suboptimal in ES6 / TypeScript

Not sure what's the reasoning you're referring to, or what may be suboptimal. What's sure that many project exist with codebases in standard style, a well as many developers who would like to start projects in standard style and they would also love to use prettier for nice whitespace. This PR allows all of them for easy migration to prettier without having whitespace removed after every function name.

The same way, dont' consider my words criticism of your likings. I realise many prefer semicolons and I don't intend to argue about it. I hope prettier realises this is not matter of one team fighting the other but rather respecting and acknowledging each other.

brodybits:
Thanks @sheerun for your work on this idea. But I have a major concern that it mixes several things together:
* spacing in functions and generators
* no semicolons
* single quotes

I would really favor keeping these items in separate options. This would really help people using "semistandard", for example.

sheerun:
"semistandard" is much less popular than standard (7 packages of top 1000 use it). Ultimately this decision is up to prettier, what I care only about it that I can use prettier to format according to standard rules. It doesn't really matter to me if I use one flag or 5. I would need official stance of @prettier in form of "if you change this and that, we'll merge this PR"

lydell:
So … I might be wrong, but here goes: I've been under the impression that people use standard because it's easy to use, helpful and ends many code style debates (no configuration). Not because it uses the exact code style people have been looking for (no semicolons, spaces in some places). If so … does it really matter if Prettier's code style is slightly different?

brodybits:
I integrated this proposal into the [`prettierx`](https://www.npmjs.com/package/prettierx) fork, behind several options as documented [here](https://github.com/brodybits/prettierx#standard-like-formatting). I would love to see these formatting capabilities integrated into [`prettier`](https://www.npmjs.com/package/prettier) someday.

I think we still need some kind of a higher-level tool like [`prettier-standard`](https://www.npmjs.com/package/prettier-standard) for easy integration between [`standard`](https://www.npmjs.com/package/standard) semantic rules and the improved formatting capabilities (could be part of standard/standard#811). Maybe based on Babel?

An alternative solution for the improved formatting capabilities may be some kind of plugin support.

lydell:
@brodybits Have you considered using ESLint directly instead of through standard, with eslint-config-standard + eslint-config-prettier?

sheerun:
> So … I might be wrong, but here goes: I've been under the impression that people use standard because it's easy to use, helpful and ends many code style debates (no configuration). Not because it uses the exact code style people have been looking for (no semicolons, spaces in some places). If so … does it really matter if Prettier's code style is slightly different?

No semicolons is main reason, but it's more than that: Imagine you've created your library in 2015. Because you don't like semicolons you've chosen `standard` for formatting and linting your codebase. Your project is successful and your codebase grows huge. Then two years later tool called `prettier` is released that can format whitespace in your project even better, great! Then you realise that you cannot switch to it without reformatting most of your codebase because it doesn't support `--jsx-single-quote` and `--space-before-function-paren`. You wait two years for first one to be added and even then prettier cannot decide on adding the second one. This is what's happening.

> I integrated this proposal into the prettierx fork, behind several options as documented here. I would love to see these formatting capabilities integrated into prettier someday.

I could do the same for `prettier-standard` but I'm convinced adding `standard` or `space-before-function-paren` functionality directly to prettier would be superior to forking this project.

brodybits:
> Have you considered using ESLint directly instead of through standard, with eslint-config-standard + eslint-config-prettier?

Yes. I personally like to use "semistandard", which is standard with semicolons, through configuration of eslint.

But prettier seems to be missing some formatting capabilities that are proposed here:

- space before function paren (see eslint [`space-before-function-paren`](https://eslint.org/docs/rules/space-before-function-paren) rule)
- generator star spacing (see eslint [`generator-star-spacing`](https://eslint.org/docs/rules/generator-star-spacing) rule)
- yield star spacing (see eslint [`yield-star-spacing`](https://eslint.org/docs/rules/yield-star-spacing) rule)

> I'm convinced adding `standard` or `space-before-function-paren` functionality directly to prettier would be superior to forking this project.

👍

j-f1:
PR for the spacing is here: #3903.

sheerun:
Maybe prettier could introduce `--preset` flag that could be used like so:

- `--preset standard`
- `--preset semistandard`

kachkaev:
> Maybe prettier could introduce --preset flag

In accordance to [Prettier's options philosophy](https://prettier.io/docs/en/option-philosophy.html), there exist [>400 votes](https://github.com/prettier/prettier/issues/40) against adding `--preset` or `--standard`.

sheerun:
It's not like there are thousands of formatting standards, there is prettier, standard, and maybe semistandard. What @gaearon said was probably a reference to hyge number of eslint config options.

brodybits:
Any ideas for moving forward with the formatting improvements?

Anything better than forking?

sheerun:
prettier is similar to standard / semistandard / airbnb / google that all are collections of formatting rules, they don't allow you to choose exact formatting options the same way prettier doesn't want to

prettier is at the same something more: it allows for precise whitespace formatting, previous tools were very limited at this. many projects that use other formatting standard like `standard` would love to have similar feature without passing their code first through prettier and then though other tool

Ideally prettier would allow to format code according to preset of rules and leave linting for other tools.

Please realise that every option prettier introduced so far and that is requested is not because people want to configure this particular option but because they want standard and semistandard formatting:

- no-semi: https://github.com/prettier/prettier/issues/12#issuecomment-271702377
- single quotes: unfortunately cannot point to anything, it was available from the beginning
- single jsx quotes: https://github.com/prettier/prettier/issues/1080#issuecomment-390363232
- space-before-function-paren: https://github.com/prettier/prettier/issues/1139#issuecomment-294177756 followed by https://github.com/prettier/prettier/issues/3847 https://github.com/prettier/prettier/issues/3845 https://github.com/prettier/prettier/issues/3845#issuecomment-362128025

So it a way it was possible for prettier to have single `--preset` option but (maybe because of constant pushback) prettier ends implementing these options one by one, contradicting its own rules.

lydell:
@brodybits How do you motivate space before function paren, generator star spacing and yield star spacing?

sheerun:
These are rules used by standard whitespace formatting

It doesn't need motivation as it's arbitrary decision made by `standard` the same way `prettier` made arbitrary decisions about formatting. It's about following once set rules of alternative formatting rules that closely popular to prettier's own.

This is why this PR doesn't try to introduce these options one by one, it actually follows your "resisting adding configuration" rule more than any single option you'd add (though `--space-before-function-paren` is last remaining significant one to be added for satisfactory compatibility with standard).

brodybits:
> How do you motivate space before function paren, generator star spacing and yield star spacing?

* Better consistency with "standard" & "semistandard" formatting that is in place in so many major projects
* Looks cleaner IMHO

brodybits:
As a side point I have the following idea: split prettier, in a similar fashion to standard, maybe in a mono repo:
* prettier-engine which has the logic, with more options
* prettier CLI with limited options
* Something like prettier-miscellaneous or prettierx which provides more options from the command like, in a similar fashion to standardx (I would be happy to surrender the prettierx package to prettier, if needed)

Even further may be to improve the integration with babel or eslint.

SimenB:
>It doesn't need motivation as it's arbitrary decision made by `standard` the same way `prettier` made arbitrary decisions about formatting. It's about following once set alternative formatting rules that are as popular as prettier's own.

Not that I much care here (since I won't be the one maintaining yet another option), but I think saying they're equal in popularity, when it's based on number of dependants and not e.g. number of contributors to those projects (or something else) is a bit odd. For more numbers (if they matter) you can also compare `standard` and `prettier` more directly (albeit that will be skewed the other way since prettier is used as a tool in a lot of IDE plugins and code generators).
`prettier` has 1 771 538 downloads the last week, compared to `standard`'s 96 622 (18x in `prettier`'s favor) and 1 838 direct dependents while `standard` has 387 (5x in `prettier`'s favor).
Another data point might be the stats of `eslint-config-standard`. They probably make more sense, as that encompasses tools being "`standard`-compliant", `semistandard` etc. `eslint-config-standard` has 539 727 downloads (way closer, still 1/3 of prettier) and 831 dependants (again, way closer but still less than half).

---

Again, I won't be maintaining (or using) this option, so feel free to completely ignore me! 😀 But the fact a bunch of popular projects went with `standard` isn't the same thing as saying `standard`'s formatting is as widespread/popular as `prettier`'s. The code formatting in peoples `node_modules` is not as interesting as the code formatting in peoples source code.

j-f1:
Similarly, GitHub’s dependent listing shows shows Prettier [has](https://github.com/prettier/prettier/network/dependents) 375 710 dependents, whereas `eslint-config-standard` [has](https://github.com/standard/eslint-config-standard/network/dependents) 219 193 dependent repos.

bakkot:
I think it's also interesting to look at trends over time. It [looks to me](https://www.npmtrends.com/prettier-vs-standard) like standard is rarely getting adopted in new projects.

sheerun:
I mentioned many times it's not only about supporting new projects but existing ones.

Also both standard and prettier are increasing in popularity it is around 24% of prettier's now:

a) https://www.npmtrends.com/prettier-vs-eslint-config-standard
b) https://hackernoon.com/what-javascript-code-style-is-the-most-popular-5a3f5bec1f6f

I'm not trying to prove that standard is more popular than prettier but that it is relevant for many

lydell:
> I mentioned many times it's not only about supporting new projects but existing ones.

But why? Why can't existing projects change? Just trying to understand all the details.

sheerun:
- https://github.com/prettier/prettier/pull/5723#issuecomment-451836295

While changing formatting standard with few developers is fairly easy, it is a big challenge for bug projects with many pending pull requests to merge which would cause many conflicts and coordination and disorder of whole team for some time.

I don't like how discussion goes because everyone tells me to prove that standard is better or why somebody would not like to switch to prettier. It is not better or worse, it's different. If prettier was black-hair only hairdresser I suggested to support blonde hair I wouldn't need to explain how popular blonde hair is, hear that it's "ugly", or why blonde people just won't color their hair black, because it's so easy.

sheerun:
I don't want to argue anymore so I'd like to ask maintainers: what is the solution you propose?

j-f1:
> Instead, research how other big code bases migrated to Prettier.

I’m wondering if we should make a docs page with some information from people who have done this, since the docs don’t have much information on adding Prettier to an existing project.

duailibe:
@sheerun We're not asking you to prove why one is better than the other. We just really hate supporting multiple code styles. And in order to keep a single code style, we ask since day one "what's the most common way people write this code by hand" for every piece of syntax. That's also what drives our decisions to add new options or not: we won't add new options unless we believe it will "unblock" lots of people to start using Prettier. That's why you're seeing "pushback" in this thread, we really don't like adding options.

I'd like to remind that while we do want new users and try to support them, we won't do that by supporting other code styles and that's why I think Prettier is not for everyone. We support this tool that formats code automatically as consistent as we can (which is good for a number of reasons), and the users just need to let go of their previous formatting preferences. If they can't do that, maybe Prettier is not for them, but that's fine.

sheerun:
Could you as compromise agree to accept PR with `--space-before-function-paren`?

KidkArolis:
At the risk of going slightly off topic, I wanted to share my perspective.

@lydell one thing I didn't see being discussed in this thread is the non-style related aspects of standard. I happily switched from standard to prettier, the auto formatting is great. I don't mind the specifics of each tool's styling choices, like many, I like to not have to make those decisions. The issue I quickly ran into after switching from standard to prettier is how prettier only formats the code, it doesn't lint against very common issues (aka stuff in eslint:recommended, things like unused variables). I was used to `standard`'s opinionatedness in helping with both **style** and **function** (can't think of a better word).

So now I have a few options (back to having to make decisions with linters). Use `prettier-standard and standard` (less configuration needed, but frustrating that both tools have different interface) or `prettier and eslint` (fidly configuration required, not clear which tool to run when and which order). 

Ideally, I just want a single tool to do both formatting and linting, but maybe I'm in the minority? I guess it's simply not prettier's concern to add `--lint` flag. But if that, or something equivalent (plugin, separate "official prettier tool", etc.) existed, I wonder if you'd get higher adoption of prettier as people would have less reason to stick with standard.

And just for completeness sake, here's the possible scenarios I see that would solve this problem for me and hopefully many other long time standard users:

1. `prettier-standard` adds `--lint` flag as discussed in https://github.com/sheerun/prettier-standard/issues/35 and https://github.com/sheerun/prettier-standard/issues/17
2. `prettier` adds `--lint` flag or an equivalent method for running `plugin:prettier/recommended`
3. `standard` adds `--format` that's powered by prettier
4. we switch to another tool that combines both formatting and linting, e.g. https://github.com/jorgegonzalez/zoe (just found this)

Again, apologies if this is a bit off topic. I understand the thread was more about making the interop between prettier and standard better. Especially for reasons @sheerun mentioned in this thread, where long time codebases want to benefit from prettier without having to reformat existing code. But I think there's this other overlooked dimension to the question of why people might want to combine the two tools – which is that both formatting and linting are very valuable and that convention based easy to use tools are also valuable.

kachkaev:
@KidkArolis the fact that Prettier does not dive into the code semantics its actually a benefit rather than a disadvantage. It leaves you with an option to check the semantics with ESLint, TypeScript compiler, TSLint, StyleLint, and many others. There are lots of languages that Prettier can work with, including those available through [plugins](https://prettier.io/docs/en/plugins.html) so if Prettier goes outside the scope of spaces and tabs, the tool will become too bulky.

I see your point about standard.js caring both about the style and the semantics, this is handy indeed. However, this leaves you with a tool that is JS-only, which is somewhat too narrow these days.

KidkArolis:
@kachkaev I think that makes sense, which is why my suggested options 1, 3 or 4 are probably better options. I'm just trying to see which one is the most viable.

@lydell if you use ESLint + eslint-config-standard + eslint-config-prettier + eslint-plugin-prettier, don't the prettier rules conflict with standard rules? Or else this GH issue (about `--space-before-function-paren` and all that) wouldn't exist?

lydell:
> if you use ESLint + eslint-config-standard + eslint-config-prettier + eslint-plugin-prettier, don't the prettier rules conflict with standard rules?

They don't! eslint-config-prettier disables all the conflicting ones.

KidkArolis:
And now going really off topic.. should look into this myself, but which one is more comprehensive: eslint-config-standard or eslint-config-recommended? If I'm jumping standard ship into the prettier one, might as well reconsider which eslint set I'm using.

lydell:
> which one is more comprehensive: eslint-config-standard or eslint-config-recommended

If you mean the built-in `eslint:recommended`, `eslint-config-standard` is much more comprehensive. I've heard eslint-config-react-app is good, too.

sheerun:
I guess I for now I won't convince anyone from prettier to support flags that make formatting closer to standard, so for now I've released [prettier-standard](https://github.com/sheerun/prettier-standard) that uses [prettierx](https://github.com/brodybits/prettierx) fork instead, so anyone wants even better standard.js compatibility than prettier-standard@9 provides, please send pull request to it.

User486375:
> > if you use ESLint + eslint-config-standard + eslint-config-prettier + eslint-plugin-prettier, don't the prettier rules conflict with standard rules?
> 
> They don't! eslint-config-prettier disables all the conflicting ones.

@lydell Have you actually tested and verified this yourself? Because it doesn't in mine and many others experiences when it comes to `--space-before-function-paren`. Multiple people mention it in both https://github.com/prettier/prettier/issues/3847 and https://github.com/prettier/prettier/issues/1139. https://github.com/prettier/prettier-vscode/issues/494 is a more specfic example.

bakkot:
@User486375 Can you give a minimal reproduction? I have a `.eslintrc.json` containing

```json
{
  "extends": ["standard", "prettier"],
  "plugins": ["prettier"],
  "rules": {
    "prettier/prettier": "error"
  }
}
```

and a `.prettierrc.json` containing

```json
{
  "semi": false
}
```

and I do not get any errors when linting a file containing

```js
function f() {
  return 0
}

console.log(f())
```

So it looks to me like extending `"prettier"` does, indeed, disable the conflicting rule from `"standard"`. Is that not what you observe?

Issue #5740, opened 2019-01-10
Docs: broken relative links


nmussy:
There are currently several broken links in a few different pages. They all reference other pages relatively (e.g. `[api](./api.md)`.

Here are a few examples:
* https://github.com/prettier/prettier/blame/master/docs/cli.md#L133
* https://github.com/prettier/prettier/blame/master/docs/options.md#L313

Links without the redundant `./` work fine:
* https://github.com/prettier/prettier/blame/master/docs/cli.md#L133

I can make a quick PR removing them if needed.

j-f1:
Awesome, please do!

PR #5741, opened 2019-01-10
docs: remove redundant "./" from relative links (#5740)

nmussy:
Removed the redundant relative paths to fix broken links on build. This was done with the following command:
```sh
sed -i "s#(\./#(#" docs/*.md website/versioned_docs/version-stable/*.md
```

This commit also removed the following warnings thrown by the `yarn run build-docs` script:

```
 [WARN] unresolved links in file 'version-stable/cli.md' > [ './api.md' ]
 [WARN] unresolved links in file 'version-stable/options.md' > [ './precommit.md' ]
 [WARN] unresolved links in file 'version-stable/api.md' > [ './plugins.md' ]
 [WARN] unresolved links in file 'version-stable/cli.md' > [ './api.md' ]
```

Maybe a `build-docs` job should be added to the CI to catch these warnings in the future?

- [ ] ~~I’ve added tests to confirm my change works.~~
- [ ] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~~
- [x] I’ve built the site locally to make sure the links now work

~~**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**~~

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -86,7 +86,7 @@ As you repeatedly call `resolveConfig`, the file system structure will be cached
 
 Setting `options.ignorePath` (`string`) and `options.withNodeModules` (`boolean`) influence the value of `ignored` (`false` by default).
 
-Providing [plugin](./plugins.md) paths in `options.plugins` (`string[]`) helps extract `inferredParser` for files that are not supported by Prettier core.
+Providing [plugin](plugins.md) paths in `options.plugins` (`string[]`) helps extract `inferredParser` for files that are not supported by Prettier core.
 
 Use `prettier.getFileInfo.sync(filePath [, options])` if you'd like to use sync version.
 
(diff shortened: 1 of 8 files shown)


duailibe:
> Maybe a `build-docs` job should be added to the CI to catch these warnings in the future?

It already runs in CI.

nmussy:
> It already runs in CI.

Sorry, I was only looking at the Azure jobs. Should an exception be thrown to prevent netlify from completing the build then?

duailibe:
@nmussy Yeah it could break the build. We could suggest that to https://github.com/facebook/docusaurus

nmussy:
I was thinking about something simpler, like watching the [website `yarn build` stdout in build-docs.js](https://github.com/prettier/prettier/blob/master/scripts/build-docs.js#L40) and throw if a warning is found.

But indeed, we could open an issue asking for a CLI option to abort the build if any warnings are triggered.

PR #5749, opened 2019-01-12
fix(markdown): handle abbr attribute correctly

### LINKED ISSUE #5686: Prettier disables abbreviation (abbr tag) for markdown
**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBBARugTnAbgSwEMZ9oACeAZxiTIBUALOMgCToFkAZMygBzjD4AZvjDFSUMvkpkAtoXywFUOABMy6AJ4UmZAOoBmAMIA6ADpQLAKgDabLgF1aLTfyz04ADxhl2hLADWAK68ZJyEUADmQYSRcBYgADQgELwk0JTIoP5YEADuAAr+CJkohLgQ+KpJINiEYAFwMADKvPWKkcgwWEFwyQwwsgA2egz4VG1gcM0l4-gEMJrI4JSZyYqUcFgwBVix8shChEObyQBWlJ4AQnsNTc2EsnCcinCHx6cgF57NHUNwAEUghB4O8Tn0QG0sJssMt5IFVPkoDVeFhFDA9FUYAxkAAOAAMyVREE2ej2vGWqLgMNwb2SOAAjkF8DhdvtCGDPptZPguj0IZQ-oDgaCkEdwckYIR0JjVNjkAAmSV7fBDDpGCCyA4oKDQOkgIKbOjS0rizYAX3NQA)
```sh
--parser markdown
```

**Input:**
```markdown
Abbreviation test: The HTML specification is maintained by the W3C.

*[HTML]: Hyper Text Markup Language

```

**Output:**
```markdown
Abbreviation test: The HTML specification is maintained by the W3C.

\*[HTML]: Hyper Text Markup Language

```

**Expected behavior:**
```markdown
Abbreviation test: The HTML specification is maintained by the W3C.

*[HTML]: Hyper Text Markup Language

```

Reference `abbr` tag implementation for `markdown-it`: https://github.com/markdown-it/markdown-it-abbr

nmussy:
Fixes #5686. 
I'm still having trouble with new lines. I've left the formater over break, but formatted them correctly in the Snapshots.

Some help wrapping this up would be greatly appreciated!

- [ ] Should this be added to the docs? ~~I’ve documented the changes I’ve made (in the `docs/` directory)~~

### DIFF
--- a/src/language-markdown/parser-markdown.js
+++ b/src/language-markdown/parser-markdown.js
@@ -39,6 +39,7 @@ function createParse({ isMDX }) {
       .use(remarkMath)
       .use(isMDX ? mdx.esSyntax : identity)
       .use(liquid)
+      .use(abbr)
       .use(isMDX ? htmlToJsx : identity);
     return processor.runSync(processor.parse(text));
   };
@@ -99,6 +100,28 @@ function liquid() {
   }
   tokenizer.locator = function(value, fromIndex) {
     return value.indexOf("{", fromIndex);
+  };
+}
+
+function abbr() {
+  const proto = this.Parser.prototype;
+  const methods = proto.inlineMethods;
+  methods.splice(0, 0, "abbr");
+  proto.inlineTokenizers.abbr = tokenizer;
+
+  function tokenizer(eat, value) {
+    const match = value.match(/^\*\[\s*([^\]]+)\s*]\s*:\s*([^\n]+)/);
+
+    if (match) {
+      return eat(match[0])({
+        type: "abbr",
+        abbr: match[1],
+        title: match[2]
+      });
+    }
+  }
+  tokenizer.locator = function(value, fromIndex) {
+    return value.indexOf("*[", fromIndex);
   };
 }
 
--- a/src/language-markdown/printer-markdown.js
+++ b/src/language-markdown/printer-markdown.js
@@ -427,6 +427,9 @@ function genericPrint(path, options, print) {
         options.locEnd(node)
       );
     }
+    case "abbr": {
+      return concat(["*[", node.abbr, "]: ", node.title.trim(), hardline]);
+    }
 
     case "tableRow": // handled in "table"
     case "listItem": // handled in "list"
--- a/tests/markdown_abbr/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/markdown_abbr/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,151 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`conflicts.md 1`] = `
+====================================options=====================================
+parsers: ["markdown"]
+printWidth: 80
+proseWrap: "always"
+                                                                                | printWidth
+=====================================input======================================
+*[*em*]: *that's en emphasis*
+*[star]: and that's not an empasis, that's just a star*
+*[that's an empasis*
+**[and that's just bold]: right?**
+*[link](:that's still a link though)
+
+* and
+* that's
+* [a]: list
+
+*Not here though*
+*[raw]:*Actually, any kind  of formatting is ignored here*
+*nor here*
+
+=====================================output=====================================
+*[*em*]: *that's en emphasis*
+
+
+*[star]: and that's not an empasis, that's just a star*
+
+
+*[that's an empasis*
+**[and that's just bold]: right?**
+ \\*[link](:that's still a link though)
+
+- and
+- that's
+- [a]: list
+
+_Not here though_
+
+*[raw]: *Actually, any kind  of formatting is ignored here*
+ _nor here_
+
+================================================================================
+`;
+
+exports[`even.md 1`] = `
+====================================options=====================================
+parsers: ["markdown"]
+printWidth: 80
+proseWrap: "always"
+                                                                                | printWidth
+=====================================input======================================
+An even test to make sure *we don't get caught by the emphasis*, or even _these ones_
+
+*[MD]:Markdown
+*[CSS]: Cascading Style Sheets
+=====================================output=====================================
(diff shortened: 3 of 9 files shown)


src/language-markdown/parser-markdown.js
```diff
@@ -102,6 +103,28 @@ function liquid() {
   };
 }
 
+function abbr() {
```
ikatyang: Since there's already a [`remark-abbr`](https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-abbr) plugin mentioned in https://github.com/prettier/prettier/issues/5686#issuecomment-450272405, any reason why you don't want to use it here? or is this what you meant "_The contents are left untouched_"? If so, could you open an issue in that repo to suggest this change?

src/language-markdown/printer-markdown.js
```diff
@@ -427,6 +427,9 @@ function genericPrint(path, options, print) {
         options.locEnd(node)
       );
     }
+    case "abbr": {
+      return concat(["*[", node.abbr, "]: ", node.title.trim(), hardline]);
```
ikatyang: The `hardline` between nodes are controlled by `shouldNotPrePrintHardline`, `shouldPrePrintDoubleHardline`, and `shouldPrePrintTripleHardline`, so you don't need to add it here. But in this case, I guess you could simply reuse the logic from `definition`, which looks like the same syntax with a prefix `*` added.

src/language-markdown/parser-markdown.js
nmussy: I didn't use [remark-abbr](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-abbr/) because it included a [transformer](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-abbr/src/index.js#L34) that replaced the `abbr` in the paragraphs on the fly and I didn't find a clean way to disable it.

*The contents are left untouched* comment was about the fact that there wasn't any additional linting applied to the parsed `node.title`

ikatyang: Got it. Could you move those `remark-abbr` code to another file just like what we did for `mdx.js`? though I still think we should ask them to expose the parsing API but this should fine as a temporary workaround.

> _The contents are left untouched_ comment was about the fact that there wasn't any additional linting applied to the parsed `node.title`

I'm not sure what did you mean but if it passed the AST alignment test on CI, it should be fine.

tests/markdown_abbr/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,151 @@
...
+=====================================input======================================
+*[LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGABBR] :  Long abbr
+
+*[AMLAEIKOUR]:A much longer abbreviation explanation, it's kind of unneeded really
+*[LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGABBR]         :                A much longer abbreviation explanation, it's kind of unneeded really
+
+=====================================output=====================================
+*[LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGABBR]: Long abbr
+*[AMLAEIKOUR]: A much longer abbreviation explanation, it's kind of unneeded really
+*[LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGABBR]: A much longer abbreviation explanation, it's kind of unneeded really
```
j-f1: Is it allowable for these to break?

```markdown
*[LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGABBR]:
  A much longer abbreviation explanation, it's kind of unneeded really
```

tests/markdown_abbr/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,151 @@
...
+*[raw]:*Actually, any kind  of formatting is ignored here*
+*nor here*
+
+=====================================output=====================================
+*[*em*]: *that's en emphasis*
+
+
+*[star]: and that's not an empasis, that's just a star*
+
+
```
j-f1: This should be a single blank line.

tests/markdown_abbr/__snapshots__/jsfmt.spec.js.snap
nmussy: [CommonMark](https://spec.commonmark.org/0.28/) doesn't have a specification for `abbr`. It seems this notation originates from [PHP Markdown](https://michelf.ca/projects/php-markdown/extra/#abbr), and it seems they define broken `abbr` as empty:

>  An abbreviation may also have an empty definition, in which case <abbr> tags will be added in the text but the title attribute will be omitted.

Lookign at the implementations, neither [remark-abbr](https://github.com/zestedesavoir/zmarkdown/blob/master/packages/remark-abbr/src/index.js#L9) nor [markdown-it-abbr](https://github.com/markdown-it/markdown-it-abbr/blob/master/index.js#L43) support breakable `abbr`

src/language-markdown/printer-markdown.js
nmussy: I gave it another shot, but wasn't successful. I couldn't make the `definiton` logic work. I'll give it another shot later, but if you have any insight it'd be greatly appreciated.

ethernal:
Is there any progress on this feature? Any way to help? (and info how is appreciated)

j-f1:
Feel free to open a new PR based on this one @ethernal.

PR #5752, opened 2019-01-13
docs(blog): 1.16 release

ikatyang:
Feel free to edit this PR directly.

⭐️[Blog Preview](https://deploy-preview-5752--prettier.netlify.com/blog/2019/01/20/1.16.0.html) 🌟


website/blog/2019-01-00-1.16.0.md
```diff
@@ -12,61 +12,213 @@ This release adds TODO.
...
+```
 
 ### General
 
 #### Better CRLF handling ([#5494] by [@ikatyang])
 
-TODO
+You might have noticed a few weird things happened only on Windows, which is
+usually caused by the bad CRLF handling. In this release, we completely get rid
+of them by normalizing linebreaks before/after formatting.
```
lydell: Is this an internal change, or will it make the CRLF story better for users? Since I'm confused, maybe we shouldn't mention this one if it's an internal change?

website/blog/2019-01-00-1.16.0.md
```diff
@@ -12,61 +12,213 @@ This release adds TODO.
...
+    // do something
+  }, [props.value]);
+}
+```
 
 #### Rename `babylon` with `babel` ([#5647] by [@wuweiweiwu])
 
-TODO
+`babylon` was renamed with `@babel/parser` in Babel 7, we reflected this change
+in 1.16 so that it should be easier to understand for newcomers.
```
lydell: Should we mention that `babylon` is deprecated but still works?

website/blog/2019-01-00-1.16.0.md
j-f1: #5494 is the PR in question here.

ikatyang: This PR did affect the output so I think it should be mentioned somewhere, though the issue it fixed were never reported (e.g., the mixed line endings caused by `// prettier-ignore`).

website/blog/2019-01-00-1.16.0.md
```diff
@@ -0,0 +1,370 @@
...
+class Foo {
+  bar: () => boolean;
+}
+```
+
+### JavaScript
+
+#### TODO(title): fix(javascript): skip .connect() method when composing fun ([#5739] by [@makepost])
+
+TODO
```
ikatyang: Not quite sure what's the reason behind it, why do we need to add a special special case for `connect`?

cc @j-f1 @suchipi

duailibe: The reason is that even though Redux and friends are very common, there are lots of other just as common and valid cases, and they can be summed up as `anything_that_might_need_a_connection.connect()` (sockets, DBs, etc.)

ikatyang:
I'd like to release 1.16 tomorrow if everything goes well.

website/blog/2019-01-00-1.16.0.md
```diff
@@ -0,0 +1,607 @@
...
+
+MJML is a markup language that uses the same syntax as HTML. We added the `.mjml`
+file extension to the list of extensions that are recognized as HTML, so Prettier
+will format it.
+
+#### Smart quote for attributes ([#5590] by [@ikatyang])
+
+Previously, the quotes around HTML attribute values were always printed as double
+quotes. To improve readability, they are now printed as the quote type that results
+in fewer produce fewer escape characters in the string, like we do for other strings.
```
thorn0: "results in fewer produce fewer" ☝

j-f1: Yikes. Fixed!

PR #5778, opened 2019-01-20
fix: do not format functions with arguments as react hooks

SimenB:
See discussion in https://github.com/prettier/prettier/pull/5608#issuecomment-455860396

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -3904,6 +3904,7 @@ function printArgumentsList(path, options, print) {
   if (
     args.length === 2 &&
     args[0].type === "ArrowFunctionExpression" &&
+    args[0].params.length === 0 &&
     args[0].body.type === "BlockStatement" &&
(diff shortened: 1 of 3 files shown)


ikatyang:
We just added a `CHANGELOG.unreleased.md` in #5775, could you add a changelog item in that file?

(I plan to release 1.16.1 tonight (UTC+8) if there's no more issue reported.)

SimenB:
Sure, have it a go!

ikatyang:
👉 https://github.com/prettier/prettier/releases/tag/1.16.1

SimenB:
Thank you! Landed the upgrade in Jest as well - no changes from 1.15 🎉

aleclarson:
I don't think presence of arguments is the best heuristic here.

Why not detect non-method calls with a name matching `/^use[A-Z]\w*$/`?

j-f1:
Maybe it should use the Hook style if `foo((...) => {` fits on one line, and break the arguments if it doesn.

PR #5793, opened 2019-01-23
JavaScript: fix edge cases in type cast comment detection regex

yangsu:
Fix edge cases involving {...} ([record @type](http://usejsdoc.org/tags-type.html))  and newlines in prettier's detection code for type cast comments


src/language-js/needs-parens.js
```diff
@@ -34,7 +34,7 @@ function hasClosureCompilerTypeCastComment(text, path, locStart, locEnd) {
         comment =>
           comment.leading &&
           comments.isBlockComment(comment) &&
-          comment.value.match(/^\*\s*@type\s*{[^}]+}\s*$/) &&
+          comment.value.match(/^\*\s*@type\s*{(.|\n)+}\s*$/) &&
```
ikatyang: This also matches 

```
* @type {A} 
* @something {B}
```

which does not look right. It should be better to write a simple function to match nested brackets.

CHANGELOG.unreleased.md
```diff
@@ -41,3 +41,5 @@ Examples:
   ```
 
 -->
+
+JavaScript: fix edge cases for type cast comment detection ([#5793] by [@yangsu])
```
ikatyang: Could you fill in the description and code samples? It's not clear what do edge cases refer to.

src/language-js/needs-parens.js
```diff
@@ -34,11 +34,37 @@ function hasClosureCompilerTypeCastComment(text, path, locStart, locEnd) {
...
-          comment.value.match(/^\*\s*@type\s*{[^}]+}\s*$/) &&
+          isTypeCastComment(comment.value) &&
           util.getNextNonSpaceNonCommentCharacter(text, comment, locEnd) === "("
       )
     );
   }
+
+  function isTypeCastComment(comment) {
+    const trimmed = comment.trim();
+    if (!/^\*\s*@type\s*\{/.test(trimmed)) {
```
ikatyang: It should be closed correctly as well:

```suggestion
    if (!/^\*\s*@type\s*\{[^]+\}\s*$/.test(trimmed)) {
```

src/language-js/needs-parens.js
```diff
@@ -34,11 +34,37 @@ function hasClosureCompilerTypeCastComment(text, path, locStart, locEnd) {
...
+      return false;
+    }
+    const brackets = [];
```
ikatyang: It'd be clearer this way:

```js
let isCompletelyClosed = false;
let unpairedBracketCount = 0;
for (const char of trimmed) {
  if (char === "{") {
    if (isCompletelyClosed) {
      return false;
    }
    unpairedBracketCount++;
  } else if (char === "}") {
    if (unpairedBracketCount === 0) {
      return false;
    }
    unpairedBracketCount--;
    if (unpairedBracketCount === 0) {
      isCompletelyClosed = true;
    }
  }
}
return unpairedBracketCount === 0;
```

yangsu: This logic doesn't require the {} to contain any content so technically `@type {}` is also valid. i guess in order to fully validate the syntax, we'd need doctrine or a mini parser for the syntax, which is probably unnecessary for this.

src/language-js/needs-parens.js
yangsu: \s* is unnecessary since trim() already removes it

PR #5800, opened 2019-01-25
Adds LWC Parser to support unquoted interop attributes

### LINKED ISSUE #5627: HTML Templating requires unquoted attributes
HTML supports unquoted attributes, we use a template language that uses unquoted attributes for templating and requires template attributes to NOT have templates. It would be ideal to support an option to not surround attributes with quotes.

Alternatively, having a "sub" HTML syntax like is done with Vue would work as well. We'd prefer to avoid forking the entire HTML plugin just to change this single format if possible.

**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4D4A6UABEagA5GboC8wAhAGoCCAMgKoCiAvrgBJwAbARFQB6MgShiseEABoQEMjEzQAzslABDAE46IAdwAKuhBpRbsEKvJAAjHVrABrODADKZJ5igBzZDA6AK5wCgAWMAC2AgDqYZjwal5gcO5mCTgJAJ7I4GoaCj5qcDowRo6+kVrIAGZaAsUKAFZqAB4AQo4ubu5akXAsPnC19Y0gLa3uPr4CcACKQRDwIw2hIF46xTq5EdG2ZDo+MDFUMGHIABwADAoHEMUxjmS5B3Bb2MMKOnAAjkGY33KWkq1SQdVWCmKkUwAWCazU01mCyWwzBozWMC0dhO6DOyAATApAlpMAJpgBhCCRKq5KDQT4gILFAAqWPM4OKXC4QA)
```sh
--parser html
```

**Input:**
```html
<div>
  <p id={VALUE}>Hello</p>
</div>
```

**Output:**
```html
<div><p id="{VALUE}">Hello</p></div>

```

**Expected behavior:**
The attribute should not be surrounded by quotes.

```html
<div><p id={VALUE}>Hello</p></div>

```

ntotten:
Fixes #5627

This pull requests adds a new parser option `lwc`. This is the same as the HTML parser, but it adds support for unquoted HTML attributes per the needs of LWC. See: #5627


src/language-html/printer-html.js
```diff
@@ -123,6 +123,21 @@ function embed(path, print, textToDoc, options) {
         return concat([node.rawName, "=", node.value]);
       }
 
+      // lwc: html`<my-element data-for={value}></my-elememt>`
```
ikatyang: Though the current implementation is fine, what do you think if we add a more advanced support that tries to parse and print the content in `{here}`? we could normalize quotes (`{a+'str'}` -> `{a+"str"}`) for example.

ntotten: Ah yeah, that would certainly be nice. I will look into this.

ikatyang: It should be just a matter to use `textToDoc` to get the `Doc` from regular JavaScript printer, remove all whitespaces and force printing it in flat mode. Something like:

```js
try {
  const jsDoc = textToDoc(/* key={code} */ code, {
    parser: "__js_expression"
  });
  const jsDocWithoutWhitespaces = mapDoc(jsDoc, doc => 
    doc.type === "line" || doc.type === "break-parent"
      ? ""
      : doc.type === "if-break"
      ? doc.flatContents
      : typeof doc === "string"
      ? doc.replace(/\s/g, "")
      : doc
  });
  return concat([node.rawName, "={", jsDocWithoutWhitespaces, "}"]);
} catch (e) {
  // fallback to the current logic if it cannot be parsed
  return ...
}
```

ntotten: @ikatyang I tried this out and it seems that any time I put a quote or space inside the attribute value it breaks the parser completely because it isn't valid HTML. For example, I can't use:

```
<div data-one={foo["baz"]} data-two={foo[ "baz" ]}></div>
```
Neither of those options is successfully parsed. I think at this point the current implementation is fine, but I'm open to suggestions.

ikatyang: It's indeed an invalid HTML if there're spaces, which is expected. But it should be valid for quotes, this looks like a bug in the HTML parser, we probably need to open another issue to track this bug and add this feature in other PR.

ntotten: I opened two bugs to track this. #5815 and #5816 to track these issues.

ikatyang:
Thanks!

PR #5804, opened 2019-01-27
Fix CI detection to avoid unwanted TTY behavior

### LINKED ISSUE #5801: list-different (and check) log every file in CI, instead of just changed files
**Environments:**
- Prettier Version: 1.16.1
- Running Prettier via: CLI
- Runtime: Node.js 10 (and probably every version)
- Operating System: Linux

**Steps to reproduce:**

- Use [this reproduction repo](https://github.com/DSchau/prettier-check-repro) and specifically [this PR](https://github.com/DSchau/prettier-check-repro/pull/1) for an illustration of the issue

**Expected behavior:**

`prettier --list-different` (and `prettier --check`) should _only_ log the different files in CI

**Actual behavior:**

`prettier --list-different` logs _every_ file in a CI server, tested with TravisCI and CircleCI.

**More information**

Prettier v1.16.1 tweaked the behavior of `--list-different` (and additionally `--check`). It appears to now log all files that are being checked, rather than just the files that differ. In a local, non-CI environment this works well because the shell is wiped with each new file, and the end result ends up being _just_ the files that are changed. In CI, this behavior is **not** present, and so the output becomes incredibly challenging to read and it's not clear which files are different and should be updated.

It appears that [this PR](https://github.com/prettier/prettier/pull/5629) may have been the cause of the regression, specifically with relying on only `process.stdout.isTTY` which seems to log _every_ file being accessed in CI.

kachkaev:
Fixes #5801

- [ ] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~~

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -3,6 +3,7 @@
 const path = require("path");
 const camelCase = require("camelcase");
 const dashify = require("dashify");
+const isCI = require("is-ci");
 const fs = require("fs");
 const globby = require("globby");
 const chalk = require("chalk");
@@ -54,7 +55,7 @@ function diff(a, b) {
 
 function handleError(context, filename, error) {
   if (error instanceof errors.UndefinedParserError) {
-    if (context.argv["write"] && process.stdout.isTTY) {
+    if (context.argv["write"] && process.stdout.isTTY && !isCI) {
       readline.clearLine(process.stdout, 0);
       readline.cursorTo(process.stdout, 0, null);
     }
@@ -459,7 +460,7 @@ function formatFiles(context) {
       return;
     }
 
-    if (process.stdout.isTTY) {
+    if (process.stdout.isTTY && !isCI) {
       // Don't use `console.log` here since we need to replace this line.
       context.logger.log(filename, { newline: false });
     }
@@ -503,7 +504,7 @@ function formatFiles(context) {
 
     const isDifferent = output !== input;
 
-    if (process.stdout.isTTY) {
+    if (process.stdout.isTTY && !isCI) {
       // Remove previously printed filename to log it with duration.
       readline.clearLine(process.stdout, 0);
       readline.cursorTo(process.stdout, 0, null);
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -3,7 +3,6 @@
 const path = require("path");
 const camelCase = require("camelcase");
 const dashify = require("dashify");
-const isCI = require("is-ci");
 const fs = require("fs");
 const globby = require("globby");
 const chalk = require("chalk");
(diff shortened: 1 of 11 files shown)


kachkaev:
What can we do with the code coverage?

SimenB:
`jest.mock('is-ci', () => true)`?

kachkaev:
`is-ci` returns a value, not a function, which makes its mocking problematic. Besides, all mocked modules have to be done via [third-party.js](https://github.com/prettier/prettier/blob/7345a38e645193926d66ef09ce5e7d754265c8bb/src/common/third-party.js) – otherwise, they can't be mocked in prod tests.

https://github.com/prettier/prettier/blob/18b03a31cc76209a2810c30819e04d659cc940f0/tests_integration/runPrettier.js#L71-L73

SimenB:
Ah, didn't know about the issue when bundling

SimenB:
Won't make a difference, `require('ci-info')` will still give you an object

lydell:
Yeah, I didn’t mean it would help the tests. I guess I was mostly wondering if anybody else was surprised by the seemingly pointless wrapper package :)

src/cli/util.js
```diff
@@ -54,7 +54,7 @@ function diff(a, b) {
 
 function handleError(context, filename, error) {
   if (error instanceof errors.UndefinedParserError) {
-    if (context.argv["write"] && process.stdout.isTTY) {
+    if (context.argv["write"] && process.stdout.isTTY && !thirdParty.isCI()) {
```
ikatyang: Since `process.stdout.isTTY` and `!thirdParty.isCI()` are always used together, it'd be better to extract them to be a new function.

kachkaev: Sounds like a reasonable idea, but how would you call it? I could not come up with a good name.

ikatyang: How about calling it `isTTY` directly?

lydell: Yes – calling `isTTY` plus adding a comment explaining it would be nice

src/utils/is-tty.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
+const thirdParty = require("../common/third-party");
+
+// Some CI pipelines incorrectly process.stdout.report isTTY status,
```
lydell: Looks like some words have been scrambled out of order?

kachkaev: 😅 true

ikatyang:
@kachkaev Thanks!

@DSchau Today or tomorrow if everything goes well.

tests_integration/runPrettier.js
```diff
@@ -74,6 +76,9 @@ function runPrettier(dir, args, options) {
   jest.spyOn(require(thirdParty), "getStream").mockImplementation(() => ({
     then: handler => handler(options.input || "")
   }));
+  jest
+    .spyOn(require(thirdParty), "isCI")
+    .mockImplementation(() => process.env.CI);
```
j-f1: Should this be mocked to a specific value so results are consistent?

kachkaev: We want to test both CI and non-CI modes and I’m setting env: { CI: true } in a couple of integration tests to make sure the behavior changes.

PR #5819, opened 2019-01-31
Tweaks the plugin directory search

arcanis:
Basically with Yarn v2 we'll make it possible to load files directly from within the zip archive. For now this will be implemented by patching the fs module so that Node is able to see the files within the archive. The problem is that in order to work well with rimraf and similar crawlers, the archive itself mustn't be stat'd as a directory. This is problematic in Prettier because you actually check it by default.

With this change the node_modules isDirectory check will succeed, and will shortcut the one to check whether its parent directory is a directory.

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -52,16 +52,19 @@ function loadPlugins(plugins, pluginSearchDirs) {
         pluginSearchDir
       );
 
-      if (!isDirectory(resolvedPluginSearchDir)) {
-        throw new Error(
-          `${pluginSearchDir} does not exist or is not a directory`
-        );
-      }
-
       const nodeModulesDir = path.resolve(
         resolvedPluginSearchDir,
         "node_modules"
       );
+
+      // In some fringe cases (ex: files "mounted" as virtual directories), the
+      // isDirectory(resolvedPluginSearchDir) check might be false even though
+      // the node_modules actually exists.
+      if (!isDirectory(nodeModulesDir) && !isDirectory(resolvedPluginSearchDir)) {
+        throw new Error(
+          `${pluginSearchDir} does not exist or is not a directory`
+        );
+      }
 
       return findPluginsInNodeModules(nodeModulesDir).map(pluginName => ({
         name: pluginName,
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -60,7 +60,10 @@ function loadPlugins(plugins, pluginSearchDirs) {
       // In some fringe cases (ex: files "mounted" as virtual directories), the
       // isDirectory(resolvedPluginSearchDir) check might be false even though
       // the node_modules actually exists.
-      if (!isDirectory(nodeModulesDir) && !isDirectory(resolvedPluginSearchDir)) {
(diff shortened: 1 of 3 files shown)


kachkaev:
I'm wondering if this new behaviour could be tested somehow 🤔 Otherwise, it won't take long for someone to break it 😅

src/common/load-plugins.js
```diff
@@ -52,17 +52,20 @@ function loadPlugins(plugins, pluginSearchDirs) {
...
-
       const nodeModulesDir = path.resolve(
         resolvedPluginSearchDir,
         "node_modules"
       );
 
+      // In some fringe cases (ex: files "mounted" as virtual directories), the
+      // isDirectory(resolvedPluginSearchDir) check might be false even though
+      // the node_modules actually exists.
+      if (!isDirectory(nodeModulesDir) && !isDirectory(resolvedPluginSearchDir)) {
```
kachkaev: From [CI](https://dev.azure.com/prettier/prettier/_build/results?buildId=512):

```
✘  prettier/prettier  Replace `!isDirectory(nodeModulesDir)·&&·!isDirectory(resolvedPluginSearchDir)` with `⏎········!isDirectory(nodeModulesDir)·&&⏎········!isDirectory(resolvedPluginSearchDir)⏎······`  
   src/common/load-plugins.js:63:11
         if (!isDirectory(nodeModulesDir) && !isDirectory(resolvedPluginSearchDir)) {
              ^
 
✘ 1 problem (1 error, 0 warnings)
```

arcanis:
Maybe by adding a mock that would call the real `stat`? Like here:

https://github.com/prettier/prettier/blob/master/tests_integration/runPrettier.js#L54-L56

arcanis:
Added a test 😃

kachkaev:
Linting is [failing again in CI](https://dev.azure.com/prettier/prettier/_build/results?buildId=514):

```
cross-env EFF_NO_LINK_RULES=true eslint . --format node_modules/eslint-friendly-formatter

  ✘  prettier/prettier  Delete `,`                         
  tests_integration/__tests__/plugin-virtual-directory.js:14:14
      write: [],
                ^

  ✘  no-else-return     Unnecessary 'else' after 'return'  
  tests_integration/runPrettier.js:63:12
      } else {
              ^

✘ 2 problems (2 errors, 0 warnings)
```

The first error should be fixed by running this:
```
./node_modules/.bin/prettier --write tests_integration/**/*.*
```

In the second case, the answer seems to be:

```js
    if (path.basename(filename) === `virtualDirectory`) {
      return origStatSync(path.join(__dirname, __filename));
    }
    return origStatSync(filename);
```

arcanis:
Woops, forgot to commit the fix, here it is!

arcanis:
Ping? 🙂

j-f1:
Thanks for contributing!

arcanis:
Any chance this could be released? It would fix a problem that block us from using prettier in the Yarn repo :D

j-f1:
I’m not sure, but in the meantime you can install `prettier/prettier` to pull the latest code from GitHub.

PR #5854, opened 2019-02-12
feat: export `createLanguage` for plugins

evilebottnawi:

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -225,6 +225,7 @@ defaultOptions: {
 A `util` module from Prettier core is considered a private API and is not meant to be consumed by plugins. Instead, the `util-shared` module provides the following limited set of utility functions for plugins:
 
 ```ts
+createLanguage(linguistData: object, { extend: object, override: object }): object;
 getMaxContinuousCount(str: string, target: string): number;
 getStringWidth(text: string): number;
 getAlignmentSize(value: string, tabWidth: number, startIndex: number): number;
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -103,3 +103,5 @@ Examples:
   (a?.b[c]).c();
   let value = /** @type {string} */ (this.members[0]).functionCall();
   ```
+
+- API: export `createLanguage` util for plugins. (by [@evilebottnawi])
--- a/src/common/util-shared.js
+++ b/src/common/util-shared.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const createLanguage = require("../utils/create-language");
 const util = require("./util");
 const { mapDoc } = require("../doc").utils;
 
@@ -20,6 +21,7 @@ function getNextNonSpaceNonCommentCharacterIndex(text, node, options) {
 }
 
 module.exports = {
+  createLanguage,
   getMaxContinuousCount: util.getMaxContinuousCount,
   getStringWidth: util.getStringWidth,
   getAlignmentSize: util.getAlignmentSize,
--- a/tests_integration/__tests__/util-shared.js
+++ b/tests_integration/__tests__/util-shared.js
@@ -3,6 +3,7 @@
 const sharedUtil = require("../../src/common/util-shared");
 
 test("shared util has correct structure", () => {
+  expect(typeof sharedUtil.createLanguage).toEqual("function");
   expect(typeof sharedUtil.getMaxContinuousCount).toEqual("function");
   expect(typeof sharedUtil.getStringWidth).toEqual("function");
   expect(typeof sharedUtil.getAlignmentSize).toEqual("function");


docs/plugins.md
```diff
@@ -225,6 +225,7 @@ defaultOptions: {
 A `util` module from Prettier core is considered a private API and is not meant to be consumed by plugins. Instead, the `util-shared` module provides the following limited set of utility functions for plugins:
 
 ```ts
+createLanguage(linguistData: object, { extend: object, override: object }): object;
```
azz: If we're exporting it I think we should document better what `linguistData` is and where to get it from, as well as what `override` and `extend` mean.

azz:
@ikatyang should we add it to `linguist-languages`?

Could be as-is or something like this?
```ts
import {extend, override} from 'linguist-languages'
import json from 'linguist-languages/data/json'

let overrides = {
  since: "1.5.0",
  parsers: ["json"],
  vscodeLanguageIds: ["json"]
}
let extensions = {
  filenames: [".prettierrc"]
}

let language = override(extend(json, extensions), overrides)
```

ikatyang:
`linguist-languages` is a data-only package so it may not be suitable. After a quick search, I found [`deepmerge`](https://github.com/TehShrike/deepmerge), which can be used for `extend`. (`override` is just an alias for `Object.assign`.)

docs/plugins.md
evilebottnawi: @azz i am bad in documentation on English, can you help me with this?

evilebottnawi:
Example usage for plugin https://github.com/prettier/plugin-php/blob/c20c8d402da20104bccf6c0338e56e6c4c1b425b/src/index.js

duailibe:
> Example usage for plugin https://github.com/prettier/plugin-php/blob/c20c8d402da20104bccf6c0338e56e6c4c1b425b/src/index.js

If you're only using it once, there's no need for `createLanguage` at all:

```js
const languages = [
  Object.assign({}, require("linguist-languages/data/php"), {
    parsers: ["php"],
    vscodeLanguageIds: ["php"]
  })
];
```

It makes sense for Prettier because we have lots of languages.. I really can't see the need to export this for 3rd-party plugins, since they mostly have a single language

evilebottnawi:
@duailibe PHP have two language: `php` and `html+php` https://github.com/prettier/plugin-php/blob/f5ee9cf377b5c210d31b67cd541f99b8c8bf476f/src/index.js#L33, also exists other languages which have more then 2 language

evilebottnawi:
@duailibe what is problem to export this util?

duailibe:
@evilebottnawi I never said there was a problem. But if we're gonna add stuff just because there's no problem our API will grow indefinitely.

IMO we should only add stuff if necessary and I don't think it's necessary.

evilebottnawi:
@duailibe better api - less copy/paste, we already (a lot of PRs was done by me) export many utilities what save our code base more clean, same problem will have other plugins, we can avoid this problem in future for our plugin system

duailibe:
@evilebottnawi I agree, I just don't think this feature is worthy of being added to the API, and I don't think the plugins need to copy/paste `createLanguage`. In your example, you only need `Object.assign()`.

That's just my opinion, if the other maintainers disagree with me, then go ahead and merge.

duailibe:
FWIW if `createLanguage` did anything more than just `Object.assign()` (like helping build the language object), I'd consider adding it as an API. But as you're proposing, even if we exported `createLanguage()`, plugin authors would still need to know they need to add `parser` key to the object.

evilebottnawi:
@duailibe it is help to `extend` extensions (maybe override some properties for other extensions what can contains php), I’m almost 99.99% sure that more extensions are being added for `PHP` and we need copy/paste this function in our code base (what we already do) :disappointed: 

I can't understand what is problem, function is very simple and help have consistent code as in prettier core also in plugins.

fisker:
@evilebottnawi Should we close this?

evilebottnawi:
@fisker I still think it would be nice to have it, now we ship own version of `linguist-languages` (https://github.com/prettier/plugin-php/blob/master/package.json#L14), so we have two packages when you use prettier (bundled) and plugin, I don't think it is good

/cc @thorn0

fisker:
But our bundled `linguist-languages` only contains data we used, not whole package.

We [require this json file](https://github.com/prettier/prettier/blob/master/src/language-js/index.js#L9)

thorn0:
@evilebottnawi What Fisker wrote. Besides, I don't see how what you wrote is related to exposing this simple function. It doesn't expose `linguist-languages`. It only adds properties to an object passed to it, something that can be done with an object spread instead. The plugin API doesn't have functions like `createPrinter` or even `createPlugin`. I don't see why it should have `createLanguage`.

evilebottnawi:
Yes, I see, let's close

PR #5878, opened 2019-02-17
Honor stdin-filepath when outputting error messages.

ju1i3nm:
Honor stdin-filepath when outputting error messages.


j-f1:
Thanks for contributing to Prettier!

PR #5890, opened 2019-02-20
refactor(utils): rewrite partition

fisker:
use Array#reduce, make partition generic


fisker:
why no .editorconfig file in this project

src/utils/partition.js
```diff
@@ -1,14 +1,14 @@
...
-      a.push(item);
-    } else {
-      b.push(item);
-    }
-  });
-  return [a, b];
+  return reduce.call(
+    array,
+    (result, item) => {
+      result[fn(item) ? 0 : 1].push(item);
```
azz: I personally find this harder to read.

lydell:
Thank you for your contribution! I think we’re happy with the way the function is already implemented, though. Since this does not fix any bug, let’s not spend time on it. If you’re looking for ways to contribute, there are plenty of issues labeled [bug](https://github.com/prettier/prettier/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Atype%3Abug) – find one that looks interesting and see if you can fix it!

src/utils/partition.js
fisker: reduce is much reasonable,  and `lodash` has the exactly same implementation

https://github.com/lodash/lodash/blob/d5ef31929a1262abbc75b8dadc0b6ae6e9558b5f/partition.js#L26

PR #5905, opened 2019-02-24
Change --check output to be warnings

cliedeman:
Changed --checkout to be warnings instead of standard logs so that they are easier to spot during CI failures.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -42,6 +42,8 @@ Examples:
 
 -->
 
+- CLI: Output warnings instead of log statements for `--check`.
+
 - CLI: Honor stdin-filepath when outputting error messages.
 
 - Markdown: Do not align table contents if it exceeds the print width and `--prose-wrap never` is set ([#5701] by [@chenshuai2144])
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -25,6 +25,7 @@ Prettier CLI will ignore files located in `node_modules` directory. To opt-out f
 
 When you want to check if your files are formatted, you can run Prettier with the `--check` flag (or `-c`).
 This will output a human-friendly message and a list of unformatted files, if any.
+As this is meant to be human readable the output will be warnings in stderr. If you want to process the files please use `--list-different`
 
 ```bash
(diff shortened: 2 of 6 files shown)


kachkaev:
Hi @cliedeman! In which situation would  `console.warn` help you better than `console.log`?

When `--check` finds unformatted files, Prettier exits with a non-zero exit code, thus stopping the CI. The entire output (except the first and the last line) consists of file names that are unformatted. This looks visible enough to me both in the local terminal and in a CI, but a understand that there may other environments, where this is not the case.

cliedeman:
Hi @kachkaev,

The use case is a mono repo build which can get quite long

Here is a small snippet (granted this code does not have the --check flag enabled yet because its on an older version of prettier but it's hard to see nonetheless. )
screenshot 2019-02-24 at 16 00 20

The idea was to mirror eslint where the problems are easier to spot

image

kachkaev:
I guess ESlint would print without colours in most of the CIs, because that's often not supported.

Unlike Prettier, ESLint can indeed exit with 0 (all OK), but print warnings. In this case, the warnings can appear somewhere in the middle of a long output and colour can help. Prettier would always exit with non-zero if at least a single formatting inconsistency has been found. This means that your build script would terminate and not print anything further, so finding a list of files should be easy even without introducing coloured output.

cliedeman:
So the warnings image was a bad choice. I the only recent I selected warn for `--check` was that the error level seems to be reserved for fatal parsing errors and the like.

I am also using `FORCE_COLOR=1` to display colours in the CI. So the reasoning is more I can get eslint to display colorful output on failures but not prettier.

screenshot 2019-02-24 at 17 01 05

Not a massively important use case but it seemed easy enough to add to me :)

fisker:
Hi @cliedeman, I'm deeply sorry that our team seems missed this one, changes in this PR already merged as #8703.

However, I found only difference between this and #8703 is if `--check --write` this one use 

```js
context.logger.log("Code style issues fixed in the above file(s).");
```

#8703 use

```js
context.logger.warn
```

I think `warn` makes sense.

I'm closing this one., sorry again.

PR #5910, opened 2019-02-26
Add --only-changed flag to CLI

### LINKED ISSUE #5853: Add cache option to avoid processing unchanged files
**Environments:**
- Prettier Version: 1.16.4
- Running Prettier via: CLI (package.json script)
- Runtime: Node.js 8.11.3 
- Operating System: MacOS 10.13.6

**Steps to reproduce:**

N/A

**Expected behavior:**
That when provided with an option such as `--cache`, prettier only process new or changed files. See [eslint](https://eslint.org/docs/user-guide/command-line-interface#--cache) for example.

**Actual behavior:**
Prettier processes all files, whether they have changed or not.

g-harel:
closes #5853 

Add an `--only-changed` flag to be used in conjunction with `--write` to avoid re-checking files that were not changed since they were last written (with the same formatting configuration). These results are stored in the `.prettiercache` file by default, but this can be configured using the `PRETTIER_CACHE_LOCATION` environment variable.

```sh
prettier --write --only-changed ...
```

Here are some rough manual benchmarks using

```sh
time ./bin/prettier.js --write --only-changed "**/*.js"
```

|                      | prettier source | ~300 file project |
| -------------------- | ---------------:| -----------------:|
| write                |             31s |                8s |
| only-changed initial |             31s |                8s |
| only-changed 2nd run |              8s |                2s |

---


src/cli/changed-cache.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Writes the in-memory cache data to the configured file.
+// Previous file contents are overwritten.
+function close(context) {
+  let contents;
+  try {
+    contents = JSON.stringify(cache);
+  } catch (err) {
+    context.logger.error(`Could serialize cache: ${err}`);
```
lydell:
```suggestion
    context.logger.error(`Could not serialize cache: ${err}`);
```

g-harel:
@lydell 

I'm still missing some coverage % for the [error handling in the changed cache implementation](https://codecov.io/gh/prettier/prettier/compare/fa24bacec4a1bb82762bb6d4fcbba5602a240710...56f724eb7897da32f6db257d2c57587e91e6a63d/diff#diff-6). I'd like to add some unit tests, but I can't find an appropriate place to put them!

mbrookes:
> These results are stored in the `.prettiercache` file by default

@g-harel Going back to [the discussion you linked to](https://github.com/avajs/find-cache-dir/issues/1) in the related issue... Even though it seems no consensus was reached, what do you think about using `.cache/prettier`? Prettier is ubiquitous enough that it could help set the standard, and if not, it's the same number of entries (one!) to add to `.gitignore`. 

(Similarly I'd love do see tools adopt `.config/toolname` as the default configuration file location, rather than littering the root dir, but one step at a time!)

g-harel:
@mbrookes

Absolutely, I only used `.prettiercache` to match the pattern from eslint.

Do you know who would need to make the final call?

lydell:
`.cache/` would not be a good idea, because it's in this Node.js .gitignore template:

https://github.com/github/gitignore/blob/37abd930a6893fd3a716c48934e704795bc2a45c/Node.gitignore#L61-L62

Apparently Parcel already uses it.

mbrookes:
>  apparently it's used by Parcel:

@lydell Great! The whole idea is that it's a single ubiquitous directory to be shared by any tool or utility that wishes to cache some data, rather than polluting the root directory with multiple `.xyzcache` files.

g-harel:
@mbrookes From what I remember using parcel, it doesn't use the folder in the way you outlined. It uses the entire folder as a cache and creates [sub-directories with hex filenames](https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/src/FSCache.js#L31)

mbrookes:
Personally I think ut’s a missed opportunity; but I’m not going to argue with you if you’ve already made up your mind.

azz:
Just playing devil's advocate on this one, if this is a CLI-only change, wouldn't it be better to have an external CLI tool that conditionally operates on files?

e.g. say we had a CLI tool called `smart-exec`, that looked something like:

```bash
$ smart-exec "**/*.js" -- prettier --write
```

This would check a local cache before invoking the command after `--` for each file.

You could even wire it up to `.prettierignore`

```bash
$ smart-exec "**/*.js" --ignore-path=.prettierignore -- prettier --write
```

There's nothing Prettier-specific about the implementation, so it should be perfectly re-usable.

**_Alternatively_**, if you wanted to do a bespoke prettier-cache tool, you could go **zero-config** ✨:

```bash
$ prettier-changed
```

This would behave exactly the same as the previous command. 

I worry about adding too much frills to prettier-core (it's why [`pretty-quick`](https://github.com/azz/pretty-quick) is a standalone tool).

g-harel:
I think this feature is important enough on large projects to be included with the "core" CLI.

The current approach (in this PR) can abstracted to a different tool, but it might make sense to eventually extend caching behavior (ex. `--check`).

I feel `prettier-quick`'s dependency on external tools (`git`, `svn`) puts it in a different league because of the increased complexity and required environment configuration.

g-harel:
Any updates on this? What is the decision making process?

src/cli/util.js
```diff
@@ -448,6 +449,16 @@ function formatFiles(context) {
     context.logger.log("Checking formatting...");
   }
 
+  let changedCache = null;
+  if (context.argv["only-changed"]) {
+    changedCache = new ChangedCache(
+      fs,
+      process.env.PRETTIER_CACHE_LOCATION || ".prettiercache",
```
evilebottnawi: I think better use `https://www.npmjs.com/package/find-cache-dir` for location

src/cli/index.js
```diff
@@ -22,6 +22,11 @@ function run(args) {
       process.exit(1);
     }
 
+    if (context.argv["only-changed"] && !context.argv["write"]) {
+      context.logger.error("Cannot use --only-changed without --write.");
+      process.exit(1);
+    }
```
evilebottnawi: Maybe we should enable `cache` by default and better add option `--no-cache` (better DX)

src/cli/changed-cache.js
```diff
@@ -0,0 +1,78 @@
...
+    let contents;
+    try {
+      contents = JSON.stringify(this.cache);
+    } catch (err) {
+      this.context.logger.error(`Could not serialize cache: ${err}`);
+      return;
+    }
+
+    try {
+      this.fs.writeFileSync(this.location, contents, "utf8");
```
evilebottnawi: https://www.npmjs.com/package/write-file-atomic or we can get files with part of content

src/cli/changed-cache.js
```diff
@@ -0,0 +1,78 @@
+"use strict";
+
+const crypto = require("crypto");
+
+// Generates a hash of the input string.
+function hash(data) {
+  return crypto
+    .createHash("sha1")
+    .update(data)
+    .digest("base64");
```
evilebottnawi: Use `md4` (better use `xxHash`, but `xxHash` doesn't built-in in node :disappointed: ), `sha1` is very slow, also we don't need strong here

g-harel: We had a bit of a discussion about this on the issue and ended up finding that hashing is not a bottleneck in this case. I'm open to changing it, but I'm not sure it's worth adding a dependency.

https://github.com/prettier/prettier/issues/5853#issuecomment-469463458

src/cli/util.js
g-harel: I considered that option, but it looked [https://github.com/avajs/find-cache-dir/issues/1](controversial) so I opted to use a different approach.

src/cli/index.js
g-harel: I thought it might be best to avoid changing existing behavior silently, especially since this flag writes a file that needs to gitignored. I'll defer to the maintainers, y'all know best.

src/cli/util.js
evilebottnawi: /cc @prettier/core

src/cli/index.js
evilebottnawi: /cc @prettier/core

evilebottnawi: Maybe we can do this for `v2`

src/cli/util.js
evilebottnawi: hm, 99% tools use `node_modules/.cache`, only `eslint` doesn't do this, i think better store cache as do other tools, because it is require many developers change cache location and it is bad DX

src/cli/changed-cache.js
evilebottnawi: Okey, let's wait other feedback, for me `md4` is best solution here, when you have really big codebase, it is increase perf around 5%

src/cli/index.js
lydell: Yeah, probably better to do it for v2

evilebottnawi: But what is problem implement `--no-cache` by default?

JelteF:
This looks like a really nice addition. Would this also work when using `--list-different` instead of `--write`? i.e. it would only check the files not checked by the last `--write`

g-harel:
@evilebottnawi

Sorry about the delay, I was in exam season.

I've changed the cache directory and switched to `write-file-atomic`. Let me know what you think of the changes!

chdsbd:
This is great addition to prettier. Thank you for taking the time to build this feature.

I found myself wondering why [python/black](https://github.com/python/black) was so much faster than Prettier and it's because Prettier doesn't have this feature yet.

What needs to be done before this PR can get merged?

g-harel:
@chdsbd 

There is a [discussion](https://github.com/prettier/prettier/pull/5910#discussion_r273888932) left unresolved about potentially having this behavior be enabled by default, which would need to be introduced in the next major version bump.

@evilebottnawi @lydell

Anything I can do to move this along?

mbrookes:
> Anything I can do to move this along?

Storm prettier HQ and merge it yourself? 😆

tests_integration/__tests__/only-changed.js
```diff
@@ -0,0 +1,162 @@
...
+    write: [{ filename: "unformatted.js" }, { filename: cachePath }],
+    status: 0
+  });
+
+  const cacheContentsAfter = resAfter.write[1].content;
+
+  expect(cacheContentsAfter).not.toBe(cacheContentsBefore);
+});
+
+describe("ChangedCache", () => {
```
duailibe: We don't have much of those unit tests in the repo, but this is fine

duailibe:
@g-harel Thanks for pushing this through. This will go out on the next minor release, so I'll wait a bit until 1.18 gets more stable, so we don't have to push any other patches

src/cli/util.js
```diff
@@ -441,6 +443,18 @@ function formatFiles(context) {
     context.logger.log("Checking formatting...");
   }
 
+  let changedCache = null;
+  if (context.argv["only-changed"]) {
+    const cacheDir = findCacheDir({ name: "prettier", create: true });
```
evilebottnawi: /cc @g-harel need use `|| os.tmp()` https://github.com/avajs/find-cache-dir#findcachediroptions

> It returns a string containing the absolute path to the cache directory, or undefined if package.json was never found.

In theory you can install `prettier` globally and doesn't have `package.json`

g-harel: Nice catch! I will open up a PR with a fix.

I'm thinking it might be easier to leave `changedCache` null if `findCacheDir` fails.

PR #5930, opened 2019-03-01
Fix trailing comment output when the printer returns a doc instead of a string

warrenseine:
When a printer exposes its own `printComment` function and returns a doc (eg. `concat([])`) instead of a string, and a comments turns out to be trailing, Prettier printed `[Object object]`.

I spent a considerable amount of time trying to add a test for this bug fix, but it requires a test case with a custom parser supporting comments. I see that there are some test plugin in `tests_integration/plugins`, but nothing as fancy. Please tell me if you require new tests for this bug fix, and I'll try to come up with something.

### DIFF
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -443,7 +443,10 @@ function printTrailingComment(commentPath, print, options) {
     return concat([" ", contents]);
   }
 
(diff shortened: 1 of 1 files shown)


vjeux:
Thanks! Let's just merge it without tests, hopefully no one will end up breaking it in the future for you.

Issue #3368, opened 2017-12-01
Linebreaks weirdly in long ES6 template strings with expressions


Swizec:
**Prettier 1.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgIoCuE8OwAOlDlTmABboBO6Y8DOAvKRdTzlOgFs4SHAHIAIhAYBLdFFHdqAXwqKq8AB4wOOAAYBNOJgA0OACTAYtaZgB0dRs1a3+QpaYCeEAjgDu01FQcdFRfdA9cACM4HAAzaAATHAhYnCFbHABJHAY4AAdczARtGAgcLx8QoKsYzGkoXAqcegA3GKg4NrZ6JJqabyYAcxjS-oEBaRhbXQolEGMQCDyYaWhMZFAsGGRYkKKFyKYwAGs4GABlPOZ6weQYBgI4BYSIMB29p5B6ooYYAAUhgJ0O9UPsQAArTAaABCR1OF0EcAAMvU4CCwd4YHkCDAAEzoz5XBg-ZAgSLoSIeVDQeYgAr1GAAdWkCSsyAAHAAGBYFCBFRlMPJ3B6fXIARwI0lyAPQgyBBIWRQmwseipuqDghGIaKQu1BnxgFOZrNoyFxC3u6ACNwAwhBxsCUFBoGiFgQigAVCnrXUfJRKIA)
```sh
# Options (if any):
--print-width=80
```

**Input:**
```jsx
class Quote {
    character = {
        name: 'Dorian'
    }

    text = `Yes, ${this.character.name}, you will always be fond of me. I represent to you all the sins you have never had the courage to commit.`
}
```

**Output:**
```jsx
class Quote {
  character = {
    name: "Dorian"
  };

  text = `Yes, ${
    this.character.name
  }, you will always be fond of me. I represent to you all the sins you have never had the courage to commit.`;
}

```

**Expected behavior:**

I would prefer a long string to _not_ get split into multiple lines, even if it contains expressions. When there are no expressions, Prettier correctly lets a long string go beyond the `print-width` limit.

I can't decide if this is a bug or a personal preference, but it's been creating weird hard to read code in our codebase. We can clean most of it up by simplifying our expressions, but there's only so short we can go.

ikatyang:
Ref: previous discussion https://github.com/prettier/prettier/issues/3315#issuecomment-346897564

duailibe:
[...]

PR #5979, opened 2019-03-16
Don't break simple template literals

jwbay:
A simple template literal is initially defined as a literal wherein all the expressions are identifiers or member access expressions where all parts are identifiers. We print these expressions into strings with infinite print width before printing the containing template string. Adding a function to determine whether a template is "simple" allows us to expand the heuristic in a central location over time.

This handles some of #3368, but I'm not sure it should be closed as there's a lot of discussion over heuristics there.


src/language-js/printer-estree.js
```diff
@@ -3864,6 +3770,168 @@ function isSimpleFlowType(node) {
...
+      return true;
+    }
+
+    // Allow `a.b.c`, `a.b[c], and`this.x.y`
+    if (
+      expr.type === "MemberExpression" &&
+      (expr.property.type === "Identifier" || expr.property.type === "Literal")
+    ) {
+      let ancestor = expr;
+      while (ancestor.type === "MemberExpression") {
```
duailibe: I think we need to include `OptionalMemberExpression` here and above

jwbay:
> Just a quick question, why did you decide to check if all expressions are simple, instead of checking each one separately?

If we're going to break at all, I think it's better to let the normal break rules apply. This avoids a situation like

```js
`asdf asdf asdf ${will.not.break} asdf asdf asdf ${will.not.break} asdf asdf asdf ${will.not.break} asdf ${some.complex.thing({ x: 42 }, y, z)} asdf`
```
printing like this (note we're surpassing line length but still breaking)
```js
`asdf asdf asdf ${will.not.break} asdf asdf asdf ${will.not.break} asdf asdf asdf ${will.not.break} asdf ${some.complex.thing(
  { x: 42 },
  y,
  z
)} asdf`
```
instead of this
```js
`asdf asdf asdf ${will.not.break} asdf asdf asdf ${
  will.not.break
} asdf asdf asdf ${will.not.break} asdf ${some.complex.thing(
  { x: 42 },
  y,
  z
)} asdf`;
```

jwbay:
Anything that could move this forward?

evilebottnawi:
Need rebase

jwbay:
Rebased 👍

donaldpipowitch:
@jwbay, thank you so much for this PR. This could make our code incredible more readable in several places. @duailibe do you still need to approve this or is the approval from @evilebottnawi enough? @jwbay can you solve the conflict? That would be so useful. ❤️ I hope this gets merged.

duailibe:
Thank you @jwbay.

And sorry this took a while, thanks for being patient.

graue:
Is this meant to keep expressions on one line even when they're over the line width? If so, any way we can get the 1.17 behavior back? Upgrading to 1.18 has forced a lot of horizontal scrolling in the codebase I'm working on, and to me, it is a big readability downgrade.

j-f1:
> Is this meant to keep expressions on one line even when they're over the line width?

I believe so. You could enable soft wrapping in your editor to make this easier to read, though.

graue:
@j-f1, thanks for the response! For me, soft wrapping is also very bad for readability.

What is Prettier's philosophy on configuration? If I'm in the minority in wanting template strings to wrap, would it be reasonable to ask for an option to restore that behavior? Would a diff adding that configuration option be likely to be accepted?

I won't belabor the point if most people disagree, but to explain where I'm coming from, I thought the point of Prettier was to wrap lines for readability, so this seems to be going against the core reason for the package. It's even in the second sentence of the readme:

>[Prettier] enforces a consistent style by parsing your code and re-printing it with its own rules that *take the maximum line length into account, wrapping code when necessary*.

That's why I'm surprised by this patch.

duailibe:
@graue Yes this is the core heuristic to decide when to add new lines in the code but isn't the only one. We do break code in multiple lines even when it doesn't exceed print width, and prevent breaking the code even if it exceeds the print width.

This doesn't mean we're not open to discussions and improvements (as long as it doesn't include an option, as much as possible); but this one specifically was one of the most requested changes in the last months of Prettier.

FrederickEngelhardt:
@duailibe would it be possible to have a configuration to enable/disable this feature?

duailibe:
@FrederickEngelhardt I don't think so

jLouzado:
> > Is this meant to keep expressions on one line even when they're over the line width?
> 
> I believe so. You could enable soft wrapping in your editor to make this easier to read, though.

I'm having a hard time understanding the rationale for this change (even after reading all the threads). This change is _literally_ breaking readability if we now need to enable an IDE option to fix it.

evilebottnawi:
@jLouzado Can you provide examples?

j-f1:
@evilebottnawi
**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEADAhgIzAAgCZwBmOA5gBYCWOAJMADpQ44BWA1gDYMC+OAtlBBwAHAI4AnHAGcYAVxr1GOAG4B3AB4A6AJ4AvbjnYQxcXjgpDJM03giGJtBk0kUYG9Lzgx9n3D5x-DGEl5Rxx0KDwNQMk3CKiIIP0IYhUjPGCyOGNUEAAaEAghGApoSWRQdDExCBUABUqEMpR0JQgKPDyQTDF0MFZPAGUhXoooEmQYMRk4fLIYXnYAdUp4SWGwOAHGlwolFy1kcEky-NHJLJhanpJedGRCdHZz-OZJNQAhHr7B9zgAGVGcHuj2eIFeagGoxI7DgAEUZAkgUgHk8ZiBhmJzmJDpgsHB2J0hGJRjBFu0YGRkAAOAAM+SJEHOix6QkORLgWKUQPyIgR8CuhSaIHQkgAtFA4HACB18sZeRRjFd0Dc7siQWjzrwKBMphqoTD4Yjgaj8jAsGS8BTkAAmU09CjsKEAYQgvFuhwEEs6MnOABUsE0UecuFwgA)
```sh
--parser babel
```

**Input:**
```jsx
`abc def ghi ${
  jkl
} mno pqr stu ${
  vwx.yz
} lorem ipsum dolor ${
  sit.amet
} etc etc etc lots ${
  and.lots.and.lots
} of words here`
```

**Output:**
```jsx
`abc def ghi ${jkl} mno pqr stu ${vwx.yz} lorem ipsum dolor ${sit.amet} etc etc etc lots ${and.lots.and.lots} of words here`;

```

PR #5991, opened 2019-03-19
Add option neverIndentImports to skip import statements formatting

DmitryKoterov:
People sometimes prefer to use VSCode's built-in source.organizeImports feature which, among other things, places all imports on the same line. Adding neverIndentImports to Prettier to optionally disable import statements indentation.

- [x] I’ve added my changes to the `CHANGELOG.unreleased.md` file following the template.


evilebottnawi:
Sorry we don't want new option (please look https://prettier.io/docs/en/option-philosophy.html) and don't want to change current output, anyway thanks for using prettier

PR #5993, opened 2019-03-20
feat: format JS template literals in JS

### LINKED ISSUE #5908: Feature request: format JS in template literals like HTML is formatted
**Prettier 1.16.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA3AhgJwATuwXmwHoAqbACQBUBZAGWxKOwAMAeAB2y4D4BGAJgDMrIu27MA3AB0oGHACMCxMgCkAygybMZOqOgA2cTDAAUXbLwCUzEABoQEdjACW0AM7JQWTBADuABSwEDxR0VAhnABM7EHlMdDAAazgYNXYE5ygAc2QYTABXOHsACxgAW30AdWLneDd0sDg1YNrnVFqAT2RwNw97TLcjGH94rLL0ZAAzA0H7ACs3AA8AIXiklLV0MrhaTLgpmaKQBcW1TKzDAEV8iHgD-VmQdMxBzG6YDvY4NzBMZycYuw-rBKlEYMVkAAOAAM9iBEEGlXi7G6QO+RlQ+3smDgAEd8s4cSN0GMJkhpg8joMys5cgUqecrjc7uTDvYYOh5KDIuDkPx2fFnPpzgBhCBlcbdKDQLEgfKDSickIUwYAX1VQA)
```sh
--parser typescript
```

**Input:**
```tsx
var a = /* HTML */ `<p   >123</p>`;
var b = /* JS */ `


alert(   1)`
```

**Output:**
```tsx
var a = /* HTML */ `
  <p>123</p>
`;
var b = /* JS */ `


alert(   1)`;

```

**Expected behavior:**
```tsx
var a = /* HTML */ `
  <p>123</p>
`;
var b = /* JS */ `
  alert(1)
`;

```

thorn0:
Closes #5908

### DIFF
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -137,7 +137,7 @@ function needsParens(path, options) {
     if (
       node.extra &&
       node.extra.parenthesized &&
-      /^PRETTIER_HTML_PLACEHOLDER_\d+_\d+_IN_JS$/.test(node.name)
+      /^PRETTIER_[A-Z]+_PLACEHOLDER_\d+_\d+_IN_JS$/.test(node.name)
     ) {
       return true;
     }
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -433,7 +433,26 @@ f[(a::b)];
 f[a::b];
 ```
 
+### JavaScript: Format JS embedded in template literals ([#5993] by [@thorn0])
+
+Template literals preceded by a `/* JS */` or `/* JavaScript */` comment get a special treatment: their content is formatted as JavaScript.
+
+<!-- prettier-ignore -->
+```js
+// Input
+const code = /* JS */ `    alert(   1)`;
+
+// Output (Prettier stable)
+const code = /* JS */ `    alert(   1)`;
+
+// Output (Prettier master)
+const code = /* JS */ `
+  alert(1);
+`;
+```
+
 [#5979]: https://github.com/prettier/prettier/pull/5979
+[#6159]: https://github.com/prettier/prettier/pull/5993
 [#6086]: https://github.com/prettier/prettier/pull/6086
 [#6088]: https://github.com/prettier/prettier/pull/6088
 [#6089]: https://github.com/prettier/prettier/pull/6089
--- a/src/language-js/clean.js
+++ b/src/language-js/clean.js
@@ -185,7 +185,7 @@ function clean(ast, newObj, parent) {
       ast.leadingComments.some(
         comment =>
           comment.type === "CommentBlock" &&
-          ["GraphQL", "HTML"].some(
+          ["GraphQL", "HTML", "JS", "JavaScript"].some(
             languageName => comment.value === ` ${languageName} `
           )
       );
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -135,18 +135,24 @@ function embed(path, print, textToDoc, options) {
         ]);
       }
 
-      const htmlParser = isHtml(path)
+      const embeddedParser = isHtml(path)
         ? "html"
         : isAngularComponentTemplate(path)
         ? "angular"
+        : isJavaScript(path)
+        ? "babel"
         : undefined;
 
-      if (htmlParser) {
+      if (embeddedParser) {
+        const language = embeddedParser === "babel" ? "JS" : "HTML";
+        const uniqueId = options.nestingLevel || 0;
         return printHtmlTemplateLiteral(
           path,
           print,
           textToDoc,
-          htmlParser,
+          embeddedParser,
+          index =>
+            `PRETTIER_${language}_PLACEHOLDER_${index}_${uniqueId}_IN_JS`,
           options.embeddedInHtml
         );
       }
@@ -584,23 +590,24 @@ function isHtml(path) {
   );
 }
 
-// The counter is needed to distinguish nested embeds.
-let htmlTemplateLiteralCounter = 0;
+//     - /* JS */ `...`
+//     - /* JavaScript */ `...`
+function isJavaScript(path) {
+  const node = path.getValue();
+  return (
+    hasLanguageComment(node, "JS") || hasLanguageComment(node, "JavaScript")
+  );
+}
 
 function printHtmlTemplateLiteral(
   path,
   print,
   textToDoc,
   parser,
+  composePlaceholder,
   escapeClosingScriptTag
 ) {
   const node = path.getValue();
-
-  const counter = htmlTemplateLiteralCounter;
-  htmlTemplateLiteralCounter = (htmlTemplateLiteralCounter + 1) >>> 0;
-
-  const composePlaceholder = index =>
-    `PRETTIER_HTML_PLACEHOLDER_${index}_${counter}_IN_JS`;
 
   const text = node.quasis
     .map((quasi, index, quasis) =>
--- a/src/main/multiparser.js
+++ b/src/main/multiparser.js
@@ -26,6 +26,7 @@ function textToDoc(text, partialNextOptions, parentOptions, printAstToDoc) {
         parentOptions.parser === "angular" ||
         parentOptions.parser === "lwc"
       ),
+      nestingLevel: (parentOptions.nestingLevel || 0) + 1,
       originalText: text
     }),
     { passThrough: true }
--- a/tests/multiparser_js_js/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/multiparser_js_js/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,51 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`js-in-js.js 1`] = `
+====================================options=====================================
+parsers: ["babel", "flow", "typescript"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+function exprToFn(expr) {
+  return Function(/* JS */ \` return ( \${expr} ) / 2 \`);
+}
+
+const someJsCode1 = /* JS */ \`    alert(   1)\`;
+const someJsCode2 = /* JavaScript */ \`    alert(   \${1})\`;
+const someJsCode3 = /* JS */ \`\${a}( \${b} - 1 + "\${c} - 1" )+(\${d}).foo\`;
+const invalidJsStaysAsIs = /* JS */ \`console . log(\`;
+const testEscaping = /* JS */ \`var a=\\\`\\\` +'\\\${ 1}'\`;
+const nestedMess = /* JS */ \`var x = /* JS */ \\\`var x = /* JS */ \\\\\\\`\${1}\\\\\\\`;\\\`;\`;
+
+=====================================output=====================================
+function exprToFn(expr) {
+  return Function(
+    /* JS */ \`
+      return (\${expr}) / 2;
+    \`
+  );
+}
+
+const someJsCode1 = /* JS */ \`
+  alert(1);
+\`;
+const someJsCode2 = /* JavaScript */ \`
+  alert(\${1});
+\`;
+const someJsCode3 = /* JS */ \`
+  \${a}(\${b} - 1 + "\${c} - 1") + (\${d}).foo;
+\`;
+const invalidJsStaysAsIs = /* JS */ \`console . log(\`;
+const testEscaping = /* JS */ \`
+  var a = \\\`\\\` + "\\\${ 1}";
+\`;
+const nestedMess = /* JS */ \`
+  var x = /* JS */ \\\`
+    var x = /* JS */ \\\\\\\`
+      \${1};
+    \\\\\\\`;
+  \\\`;
+\`;
+
+================================================================================
+`;
--- a/tests/multiparser_js_js/js-in-js.js
+++ b/tests/multiparser_js_js/js-in-js.js
@@ -1,0 +1,10 @@
+function exprToFn(expr) {
+  return Function(/* JS */ ` return ( ${expr} ) / 2 `);
+}
+
+const someJsCode1 = /* JS */ `    alert(   1)`;
+const someJsCode2 = /* JavaScript */ `    alert(   ${1})`;
+const someJsCode3 = /* JS */ `${a}( ${b} - 1 + "${c} - 1" )+(${d}).foo`;
+const invalidJsStaysAsIs = /* JS */ `console . log(`;
+const testEscaping = /* JS */ `var a=\`\` +'\${ 1}'`;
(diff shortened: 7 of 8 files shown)


src/language-html/printer-html.js
```diff
@@ -113,7 +113,7 @@ function embed(path, print, textToDoc, options) {
 
       // lit-html: html`<my-element obj=${obj}></my-element>`
       if (
-        /^PRETTIER_HTML_PLACEHOLDER_\d+_IN_JS$/.test(
+        /^PRETTIER_PLACEHOLDER_\d+_IN_JS$/.test(
```
evilebottnawi: I think it is not good avoiding `HTML`, for `css` we also use `CSS`

src/language-js/needs-parens.js
```diff
@@ -121,6 +121,16 @@ function needsParens(path, options) {
     return true;
   }
 
+  // Placeholders used for formatting JS embedded in JS might be substituted by complex
+  // expressions, so the parens around them should not be dropped.
```
j-f1: IMO the `${}` is enough to wrap the expression and extra parentheses are unnecessary.

thorn0:
```js
function exprToFn(expr) {
    return Function(/* JS */ ` return ( ${expr} ) + 1 `);
}
```
If Prettier removes the parentheses around `${expr}` (which it would do by default as placeholders are identifiers), this function will break.

src/language-html/printer-html.js
thorn0: Ok, I put it back.

src/language-js/needs-parens.js
j-f1: I see — this is referring to the actual interpolated expression and not the interpolation.

src/language-html/printer-html.js
thorn0: However, these placeholders are long, which causes unneeded line breaks. Can we make them shorter?

thorn0:
[JS in JS in HTML in JS](https://deploy-preview-5993--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAzAFAegFQAIBSAyrtprgAYA6UlMAJMLrgEYA2AhrgL7W0YAWMALatKlcgB5OuAHwTMnGbgkBnMACcAlgAcYM6kyaQoKiKzgA6VhADmWPERKYxL8e3PqY6SiAAScVmtcVzEGJl4YWgBXFTh1cJoYLgBCHwBKEPE0+TUtXX0s6nI0gG4QABoQCF1NaBVkUHZ1dQgAdwAFJoR6lHYANwhNABMKkGZ1djAAazgYQm1JzSgbZBh1KLhKwREAdX5NeBUFsDhCboPNPoOAT2RwFXrKpdjPdombIXZkVHdYyoArFQADwAQhNprNCOwhHAADJLODfX6bECAoGEJY2cwARSiEHgSNYfxAC3ULzuzHYzGu1igo20WlgO2GMH4yAAHAAGSoMiCxHYTbR3BlwF59RGVACOePgb2qPRA7BUAFooHA4EMNaN1HBpZodW92B8vkgfkSUbEhJpVusLZicTLEabkZUYFTmUNWcgAEyuiaaViYgDCECEnzuUGgEpAMTgABUqT0zbEuFwgA) 🤣

duailibe:
Hi.. Thanks for the PR! But I don't think we should add this.

First of all, I don't see the point. Second, we've already problems assuming we can format template literals and breaking people's code. Yes, this is only enabled with a comment before but still.

thorn0:
@duailibe 

> First of all, I don't see the point.

For example, I have configuration data with code (functions). Now, I need to unit-test the logic that consumes that configuration. So in my tests I have quite a few template literals containing JS. And I would be really glad if Prettier could format them for me.

> Second, we've already problems assuming we can format template literals and breaking people's code. Yes, this is only enabled with a comment before but still.

Still what? Isn't the comment a great solution for this problem?

Besides, this PR fixes bugs with [escaping](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEACVA3AhgJ1QCxgFsAbVAXlQHoAqVACQBUBZAGVRqtQAMAeAZzA4AlgAcYAPmx4s5ADpzuCvlUEjxE7gG4QAGhARxw6P2ShcOCAHcACrgSmUWDBGEATPSABGOLGADWcDAAyqJ+wlAA5sgwOACucPqEpADq+MLw-GFgcMEOGcIYGQCeyOD8pvoR-HA4MDa+kURYyABmWCQ1+gBW-AAeAEK+AUHBWERwrBFwbR1dIL19wRGRJHAAinEQ8LOdiSBhODU4ZV5YXnAknqIisCnuMPjIABwADPo3EDUpvqJlN3BjhgZvocHAAI5xYRghpYJotJDtPb6GpEYQxeL7fgrNabbYzRFzfYwc73NyPZAAJn0sSwwhIKwAwhAiM0ylBoCCQHEaoxzo4kTUAL5CoA) and nested embeds (this one can't be shown without fixing the first one).

duailibe:
> Still what? Isn't the comment a great solution for this problem?

It's *a* solution, I wouldn't call it great.

> this one can't be shown without fixing the first one

Why not?

thorn0:
> It's a solution, I wouldn't call it great.

What's wrong with it then?

> Why not?

Because it involves template literals inside JS-in-HTML-in-JS, which are broken because of the first bug.

src/language-js/needs-parens.js
```diff
@@ -137,7 +137,7 @@ function needsParens(path, options) {
     if (
       node.extra &&
       node.extra.parenthesized &&
-      /^PRETTIER_HTML_PLACEHOLDER_\d+_\d+_IN_JS$/.test(node.name)
+      /^PRETTIER_[A-Z]+_PLACEHOLDER_\d+_\d+_IN_JS$/.test(node.name)
```
fisker:
```suggestion
      /^PRETTIER_[A-Z]+?_PLACEHOLDER_\d+_\d+_IN_JS$/.test(node.name)
```

fisker:
let's put some css in the js [CSS IN JS in JS in HTML in JS](https://deploy-preview-5993--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAzAFAegFQAIBSAyrtprgAYA6UlMAJMLrgEYA2AhrgL7W0YAWMALatKlcgB5OuAHwTMnGbgkBnMACcAlgAcYM6kyaQoKiKzgA6VhADmWPERKYxL8e3PqY6SiAAScVmtcVzEGJl4YWgBXFTh1cJoYLgBCHwBKAG5qYxUYXEhreIBeXABydTgAE1LslRUQhpdySs0AN1xgAoh1JFwGLvUuDJ4aRobyLNHXcjT5NS1dfXE06hmMkAAaEAhdTWgVZFB2dXUIAHcABWOEA5R2VohNSs2QZnV2MABrOBhCbQ-NFAbMgYOoonAtoIRAB1fiaeAqf5gOCEG7wtrwgCeyHAdRegNingu7xsQnYyFQ7liWwAVioAB4AIXeXx+hHYQjgABlAXAKVSISA6fTCICbOYAIpRCDwfmsakgf7qQk45jsZiY6xQF7aLSwaFPGD8ZAADgADFtdRBYtD3tocbq4ITWnytgBHaXwYk7W4gdgqAC0UDgVSqLwqHs0FWJ7FJ5KQlPlgtiQk0ILBybFks9fITAq2MHVBsqRuQACYC+9NKwxQBhCBCMk4qDQV0gGJwAAq6tuidiXC4QA)

fisker:
Can you rebase this? I think this is useful for case like this https://github.com/prettier/prettier/pull/8787/files#diff-8d6a050fa5dbb29680ec972a3550ae8bR31, this js code could be formatted automaticly.

fisker:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEqEATgLYCGMA6jjABYDC6xAziQBQ5QAO6MANPgjdc0FgEp8wADpR88-MTgxmcgG6kAdMUwBJKADk4AdwbR4ADxjtZC+QHoAVPgBSAZXyP7+AAa3IUCwwQiJ4gfgAvFL4mrEAJAnCooH83AA26ADmXCxI+ADatnbF8rGaAPIARgBWcNiaauQZcCzs3EowuHDEAAoZ2YHi-EUldmXsCXFJYSya6Vk5+AA+SwUAukMjo2vDcqMAvgDcW3btyl3EmkRklDT0TKwck1y8AiHJEsd7dj67JcD4Z48Ph5F58QSTaZiPJQlIxWIschQAAmlQgFnw+y24i+8n2IH4IFhLGQoHIxGIEGMPXJCBJKCaxnIAE8SYTKsRyGAANbKNzcLlcTLIGA6OCEugwUhpKh0WgtAVgOBuOm0HBqWjM5DgFhskA5bowHqczIUZCoJoscUgaosCwAIU5PL55FIcAAMlw4ObLdbbRY3EK0nAAIroCDwH1pK2EgWsbrayrkSpwNIEkDtLjUHDI+jIAAcAAZY5SrVROdxtWcrcQ1N7CQBHcPwY3CekgcgsAC0UDgcGR-fTSibOCUxvIpvIUZjICtpBwIrFhJYQdDze9SAt0etMGTNFzdGQACZCaLyDg0kKzKQzSgWgBWdPoK0AFWT9K3M7U6Dg+gHsDcMBiBwEQAEEUTcGBmWDac4H2fYgA)
```sh
--parser babel
```

**Input:**
```jsx
function formatWithCursor(input, options) {
    return vm.runInNewContext(
     /* JS */ `
 const options = { ...$$$options,plugins: [
            ...Object.values(prettierPlugins),
            ...($$$options.plugins || []),
          ],
        };
        prettier.formatWithCursor($$$input, options);
      `,
      { $$$input: input, $$$options: options, ...sandbox }
    );
  }
```

**Output:**
```jsx
function formatWithCursor(input, options) {
  return vm.runInNewContext(
    /* JS */ `
 const options = { ...$$$options,plugins: [
            ...Object.values(prettierPlugins),
            ...($$$options.plugins || []),
          ],
        };
        prettier.formatWithCursor($$$input, options);
      `,
    { $$$input: input, $$$options: options, ...sandbox }
  );
}

```

**Prettier pr-5993**
[Playground link](https://deploy-preview-5993--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEqEATgLYCGMA6jjABYDC6xAziQBQ5QAO6MANPgjdc0FgEp8wADpR88-MTgxmcgG6kAdMUwBJKADk4AdwbR4ADxjtZC+QHoAVPgBSAZXyP7+AAa3IUCwwQiJ4gfgAvFL4mrEAJAnCooH83AA26ADmXCxI+ADatnbF8rGaAPIARgBWcNiaauQZcCzs3EowuHDEAAoZ2YHi-EUldmXsCXFJYSya6Vk5+AA+SwUAukMjo2vDcqMAvgDcW3btyl3EmkRklDT0TKwck1y8AiHJEsd7dj67JcD4Z48Ph5F58QSTaZiPJQlIxWIschQAAmlQgFnw+y24i+8n2IH4IFhLGQoHIxGIEGMPXJCBJKHIaggOGRBJAlWI5DAAGtlG5uFyuJlkDAdHBCXQYKQ0lQ6LQWgKwHA3HTaDg1LQAJ7IcAsEmEnLdGA9TmZCjIVBNFjikDVFgWABCnJ5fPIpDgABkuHALVabXaLG4hWk4ABFdAQeC+tLWwkC1jdHWVciVTVpaBs9pcags+jIAAcAAY45TrVROdwdWdrcQ1D7CQBHCPwE3CekgcgsAC0UDgcGR-bZSibOCUJvIZvI0djIGtpBwIrFhJYwbDzZ9SEtMZtMBTNGReaQACZCaLyDg0kKzKRzSgoNB6yB0NaACop+lb637fZAA)
```sh
--parser babylon
```

**Input:**
```jsx
function formatWithCursor(input, options) {
    return vm.runInNewContext(
     /* JS */ `
 const options = { ...$$$options,plugins: [
            ...Object.values(prettierPlugins),
            ...($$$options.plugins || []),
          ],
        };
        prettier.formatWithCursor($$$input, options);
      `,
      { $$$input: input, $$$options: options, ...sandbox }
    );
  }
```

**Output:**
```jsx
function formatWithCursor(input, options) {
  return vm.runInNewContext(
    /* JS */ `
      const options = {
        ...$$$options,
        plugins: [
          ...Object.values(prettierPlugins),
          ...($$$options.plugins || [])
        ]
      };
      prettier.formatWithCursor($$$input, options);
    `,
    { $$$input: input, $$$options: options, ...sandbox }
  );
}

```

thorn0:
I thought the consensus was that we wouldn't add such things without configurability.

mfulton26:
> I thought the consensus was that we wouldn't add such things without configurability.

Is `/* JS */` or `/* JavaScript */` not a sufficient form of configurability?

thorn0:
@mfulton26 No. A new option `--embedded-language-formatting` has been added in #7875 (for context, see also the issues that PR links to). The option has only two simple values for now: `auto` - the current (default, legacy) behavior, `off` - disable embedded formatting completely. The idea is to design a way to use this option to specify the correspondence between template tags, special comments, etc. and embedded syntaxes whereas the default behavior won't be extended anymore.

mfulton26:
@thorn0 that is great context. Thank you for sharing!

fisker:
Like this

```js
{
  overrides: [
    file: "*.js",
    options: {
		"embedded-language-formatting": [
			{
				comments: "js"
				parser: "babel", // Use `language` when we support
			},
			{
				callee: "html",
				parser: "html"
			}
		]
    }
  ]
}
```

?

thorn0:
@fisker Yes, what is also needed is a way to specify whether these definitions are an addition to the default behavior or a replacement for it.

mfulton26:
I'd like to mention that something I've always really liked about Prettier is exactly what's on the homepage:

> ## What is Prettier?
> - An opinionated code formatter
> - Supports many languages
> - Integrates with most editors
> - Has few options
> ## Why?
> - You press save and code is formatted
> - No need to discuss style in code review
> - Saves you time and energy
> - And more

I'd rather not need to configure anything in order to get embedded JS support, etc. Today I get embedded GQL, CSS, etc. support without any configuration. Things just work out-of-the-box.

> I thought the consensus was that we wouldn't add such things without configurability.

I think it might be great to share the reasons for this and possibly reconsider. If prettier can ship default configuration that just works for most people then it reduces the need that anyone will need to configure it each time they set up a new project, etc. and it increases the odds that code formatting between projects with prettier will more be consistent/similar.

thorn0:
@mfulton26 I answered in #5588.

thorn0:
closing for now

PR #6004, opened 2019-03-23
Added salesforce & heroku as users

ntotten:
Added Salesforce and Heroku icons to users page.


j-f1:
Thank you for using Prettier!

Issue #6008, opened 2019-03-24
Add `doc` output bundle to allow `require('prettier/doc')`


rattrayalex:
I'm interested in working with the prettier `doc` IR directly, and printing it, in the browser, to do pretty code generation on the fly of small snippets of example code in a variety of languages. 

The full package checks in at around a megabyte, which is a bit heavy for my purpose.

Currently, the `prettier` npm package includes several builds, including standalone parsers, but does not have a separate build for [`doc`](https://github.com/prettier/prettier/tree/master/src/doc). 

I'd be happy to add this, but wanted to check first.

PR #6009, opened 2019-03-25
Add doc build

rattrayalex:
Fixes https://github.com/prettier/prettier/issues/6008

I ran `yarn build` locally, and was then able to do this: 
```
> const {builders: b, printer: p} = require('./dist/doc.js')
undefined
> p.printDocToString(b.concat(['hello', 'world']), {})
{ formatted: 'helloworld' }
```
The file is 58kb 🎉

I wasn't sure if this warranted docs or tests; please advise. 

### DIFF
--- a/scripts/build/config.js
+++ b/scripts/build/config.js
@@ -112,6 +112,13 @@ const coreBundles = [
     external: [path.resolve("src/common/third-party.js")]
   },
   {
+    input: "src/doc/index.js",
+    name: "doc",
+    type: "core",
+    output: "doc.js",
+    target: "universal"
+  },
+  {
     input: "standalone.js",
     name: "prettier",
     type: "core",


lydell:
How does this affect thinking about backwards compatibility and versioning?

j-f1:
Isn’t this already part of the public API @lydell?

lydell:
It is? Sorry, didn't know!

j-f1:
Yep, it’s available as `require('prettier').doc`.

rattrayalex:
Thanks @j-f1 and @evilebottnawi !

Looks like the only place `doc` is mentioned in the docs is in [the plugins docs](https://github.com/prettier/prettier/blob/0d0e88cfdf886f8915d25c64391f27f1a931e32e/docs/plugins.md#printers), so I think we're good to leave this undocumented. It's pretty niche. 

I think I might be cooking up something interesting with this, though, so hope to have something fun to show before too long 😺

PR #6015, opened 2019-03-29
Enable using custom parser with languages other than JS

adek05:
Currently using custom parser API implies using ASTFormatter "estree".
 
This does not allow for using prettier for simple GraphQL codemodding or using custom GraphQL parser.


fisker:
I like the idea.

fisker:
I'm going to rebase this, and accept it.

thorn0:
> The --parser CLI option may be a path to a node.js module exporting a parse function.

What about this way to specify custom parsers?

fisker:
> What about this way to specify custom parsers? How can be astFormat passed in that case?

You are right, to do this, we have to add `astFormat` option, and I think we won't add it. 

Another problem is if we drop `astFormat` in `normalizeOptions`, this function won't get it.

I didn't think through.

thorn0:
As a probably-good-enough solution we can allow attaching the `astFormat` property directly to the function:
```js
const astFormat = opts.astFormat || opts.parser.astFormat || "estree";
```

thorn0:
Overall, the plugin API can do everything custom parsers can. Custom parsers should have been deprecated long time ago.
See https://github.com/prettier/prettier/issues/3150#issuecomment-392622776

fisker:
I agree, let's close this then.

PR #6017, opened 2019-03-30
WIP: Add option to choose between 1TBS and Stroustrup brace style

### LINKED ISSUE #840: Feature Request: breakBeforeElse
I know the goal of prettier is to standardize on a style, but it is also difficult since this did not come out at the same time as the language. 

Our team uses the exact same style as prettier except placement for `else` statements. It is effectively blocking us from adopting this awesome library.

We prefer break before `else` to allow for cleaner comments.

```js
// This does something
if (a) {
}
// This does something else if
else if {

}
// This does something else
else {

}
```

Which is nicer than.

```js
// This does something
if (a) {

// This does something else if
} else if {

// This does something else
} else {

}
```

However this is preference and would be awesome if this was added as an option to prettier.

@vjeux  showed how easy it is to do https://github.com/prettier/prettier/pull/837 and apparently @jlongster likes it :) https://twitter.com/jlongster/status/834407714965057540

phaux:
Add option for Stroustrup brace style:
```js
if (foo) {
  bar()
}
else {
  baz()
}
```

Fixes #840 

### DIFF
--- a/src/language-js/options.js
+++ b/src/language-js/options.js
@@ -70,6 +70,23 @@ module.exports = {
       }
     ]
   },
+  braceStyle: {
+    since: "1.17.0",
+    category: CATEGORY_JAVASCRIPT,
+    type: "choice",
+    default: "1tbs",
+    description: "Choose style of braced statements",
+    choices: [
+      {
+        value: "1tbs",
+        description: "Put all braces on the same line"
+      },
+      {
+        value: "stroustrup",
+        description: "Put closing brace on it's own line."
+      }
+    ]
+  },
   trailingComma: {
     since: "0.0.0",
     category: CATEGORY_JAVASCRIPT,
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1764,7 +1764,9 @@ function printPathNoParens(path, options, print, args) {
             )) ||
           needsHardlineAfterDanglingComment(n);
         const elseOnSameLine =
-          n.consequent.type === "BlockStatement" && !commentOnOwnLine;
+          options.braceStyle === "1tbs" &&
+          n.consequent.type === "BlockStatement" &&
+          !commentOnOwnLine;
         parts.push(elseOnSameLine ? " " : hardline);
 
         if (hasDanglingComments(n)) {
@@ -1886,7 +1888,7 @@ function printPathNoParens(path, options, print, args) {
       const doBody = group(concat(["do", clause]));
       parts = [doBody];
 
-      if (n.body.type === "BlockStatement") {
+      if (n.body.type === "BlockStatement" && options.braceStyle === "1tbs") {
         parts.push(" ");
       } else {
         parts.push(hardline);
@@ -1943,7 +1945,13 @@ function printPathNoParens(path, options, print, args) {
         "try ",
         path.call(print, "block"),
         n.handler ? concat([" ", path.call(print, "handler")]) : "",
-        n.finalizer ? concat([" finally ", path.call(print, "finalizer")]) : ""
+        n.finalizer
+          ? concat([
+              options.braceStyle !== "1tbs" ? hardline : " ",
+              "finally ",
+              path.call(print, "finalizer")
+            ])
+          : ""
       ]);
     case "CatchClause":
       if (n.param) {
@@ -1962,6 +1970,7 @@ function printPathNoParens(path, options, print, args) {
         const param = path.call(print, "param");
 
         return concat([
+          options.braceStyle !== "1tbs" ? hardline : "",
           "catch ",
           hasComments
             ? concat(["(", indent(concat([softline, param])), softline, ") "])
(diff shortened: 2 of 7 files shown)


lydell:
Hi! Thank you for your pull request!

At the moment we’re not looking too add an option for this. See our [option philosophy](https://prettier.io/docs/en/option-philosophy.html).

The patch looks good, but it is missing tests for the new formatting and docs. However, since the PR might not be merged it might not be worth working on it.

phaux:
I'm gonna keep my fork around only so I can npm install it in my projects. If you don't want this option that's fine.

But! You should consider it. I don't wanna repeat all the justification that was said in previous issues, but srsly, people with good taste use Strustrup style 😏

JounQin:
Focus on your code instead.

lydell:
@JounQin I’ve noticed you’ve made lots of these short comments on various issues/PRs about style changes and options. Could you please stop that? I don’t think it helps. I actually think it upsets people for no reason. Usually, these discussions don’t have much more to say than what’s already been discussed and what’s mentioned in the [Option philosophy](https://prettier.io/docs/en/option-philosophy.html). Thanks!

JounQin:
@Iydell So why we're not closing this then?

JounQin:
> I’ve noticed you’ve made lots of these short comments on various issues/PRs about style changes and options. Could you please stop that? I don’t think it helps.

Are you saying I can not express my dissatisfaction for these issues or options which ask for new options? The [Option philosophy](https://prettier.io/docs/en/option-philosophy.html) is great, but I think the issue reporters or PR authors ignored it, that's the point.

And can you mention which issue or PR that I didn't provide necessary comment?

You're so rude to me here.

lydell:
Sorry, I did not mean to sound rude at all! That’s the thing with the Internet – it’s so hard to get your intents across.

Let’s take your comment above as an example: https://github.com/prettier/prettier/pull/6017#issuecomment-554698148:

> Focus on your code instead.

Now, _that’s_ something that can easily be interpreted as rude. When I first saw that my mind automatically read it like: “Focus on your code instead, you don’t know what you’re talking about. The problem isn’t Prettier – it’s you.” I don’t know why that’s my immediate interpretation, but that’s how my brain did it.

My point here is – what value does that comment bring? How does it improve the discussion? If you don’t like something – react to that issue/PR/comment with the :-1: emoji.

I’ve run into issues with tools so many times only to open or find an issue where the maintainers are reluctant but to me it feels _obvious_ that I’m right and the issues should be “fixed”. It’s so easy to be angry in these situations! I’ve posted several comments in the past that I regret today.

So if somebody comes up with a “Focus on your code instead” in such a thread, you’re kind of just making an already upset crowd even more upset.

Here are some other comments I was thinking about:

- https://github.com/prettier/prettier/pull/3903#issuecomment-554649658
- https://github.com/prettier/prettier/pull/6644#issuecomment-554685330

There was maybe one more, but when re-reading the above two I realize my interpretation of them was colored by the “Focus on your code instead”.

JounQin:
@lydell I'm asking maintainers for closing this PR instead.

https://github.com/prettier/prettier/pull/3903#issuecomment-554649658
I've replied at following that I was just going to bed, and I've explained the detail today.

https://github.com/prettier/prettier/pull/6644#issuecomment-554685330
Is there anything wrong with this comment? I was suggesting to use same policy like object literal. I can't agree with your conclusion definitely.

JounQin:
Again, please close this PR if we are not going to merge it at all.

lydell:
> I'm asking maintainers for closing this PR instead.

Sorry, I won’t close this because I see no reason to. We won’t agree on this so let’s not discuss it further.

> #3903 (comment)
> I've replied at following that I was just going to bed, and I've explained the detail today.

Great – thanks!

> #6644 (comment)
> Is there anything wrong with this comment? I was suggesting to use same policy like object literal. I can't agree with your conclusion definitely.

In “we "hate" options” it’s unclear what “we” refers to.

Why not keep it like literal object? Maybe we should. Maybe we shouldn’t. It’s like the blank lines discussion – if we keep newlines based on input it becomes a manual formatting task.

It’s fine to disagree!

---

I think the conclusion here is that both of us should think twice before pressing “Comment” when tired!

JounQin:
> Sorry, I won’t close this because I see no reason to. We won’t agree on this so let’s not discuss it further.

No one else is working on this PR for half one year, and no more comments except us. I cann't see the worth for not closing it.

Or let more maintainers to make the decision here.

lydell:
We’ve probably spammed the other maintainers enough here with notifications now so they’ll show up if the want to.

j-f1:
Personally, (although I haven’t worked on many proprietary codebases), I’ve never seen this style in JS as far as I can remember. Most of the existing options (bracket spacing, quotes, semi/no-semi, arrow parens, etc) have varied across codebases but 1TBS seems to be a de facto standard.

+1 for closing from me.

azz:
Yeah let's not do this. I appreciate (as will all formatting changes) some will prefer this style, but it is indeed against the option philosophy.

n1ghtmare:
This PR is dope! Please re-consider merging it (pretty-please)! 🤤

PR #6019, opened 2019-03-31
Upgrade graphql-js to 14.2.0

adek05:
14.0 added support for variable directives. I am planning to put a separate PR after this one is merged which will add support for printing them.

I run existing test suite and nothing breaks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "flow-parser": "0.84.0",
     "get-stream": "3.0.0",
     "globby": "6.1.0",
-    "graphql": "0.13.2",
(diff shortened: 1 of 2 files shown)


adek05:
I don't have write access to the repo. Will you be able to merge this for me?

j-f1:
I want to wait until tomorrow so maintainers from different time zones or who don’t do OSS on weekends can take a look at it.

duailibe:
I was very confused between this and #6020.

Since the other PR also has the dependency bump, let's keep that one instead.

Issue #6030, opened 2019-04-04
Add a way to disable functional composition rule


leonidasv:
The change introduced by #4431 isn't disableable. Since we upgraded to 1.13+, lots of our files are being changed by this rule (mostly Redux related) but this goes against the previous stilying rule (same line for Redux `connect` invocations).

As this rule isn't disableable in configs (afaik), we had to downgrade to 1.12 in order to avoid the warnings on those files and breaking the previously stablished rule.

Could this rule be made an option like `singleQuote`, `bracketSpacing` and others?

evilebottnawi:
Please provide examples

leonidasv:
@evilebottnawi 
```
export default connect(mapStateToProps, mapDispatchToProps)(Component);
```
Was a valid code until 1.13. Now, this line is turned into:
```
export default connect(
  mapStateToProps,
  mapDispatchToProps
)(Component);
```

Since this is common line code in a lot of React files, enabling this rule changes a lot of files (for example, in our codebase, it changed 200+ files in just one react project).

evilebottnawi:
Yep, looks ugly, it is only in `export`? Or you can have other examples?

brainkim:
related: #5969 :trollface:

leonidasv:
@evilebottnawi 
#5969 #4935 #4751 https://github.com/prettier/prettier/issues/4935#issuecomment-422843356

evilebottnawi:
@leonidasv let's close this in favor https://github.com/prettier/prettier/issues/5969 (feel free to feedback)

Issue #5898, opened 2019-02-22
Inconsitent formatting for one specific line (always the same line)


jisuo:
**Prettier 1.16.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACSKAZ33hIGEIBbLKBfAXj2AB0o894aAbAQ3gA0bDoRjYI3bnGxICfIkTwAVOBWq16zYRwLQS2AK5gYOABQASKAHMASnAAmBtAEot7HRwDk1gJJQAVnDGngDcbNoeMAAWAJZEAHQGxAYARkRg2DEpcHhMlrYOTvGEdMam0XHxVLwYAMow-HACzAC+zuWxRM5h7hwtEXgD1XUN8KYkja6svTrYcDAG2OzTHpGqMLIT8PFkMEIzfT0e-e4nLSACIBAYMDF6yKC82OIA7gAKTwhEyCC8AG4QGL2C4gFLYXhgADW81qGAhMWsyDEBiaICiMCo3AA6rEyHCwHBal8Yrc-iSAJ4-MAKEEIojSGBvcFWarIABmvG49Mu-iIaAAQuCoTDeFQ4AAZBFwdmc7kgXloWoIqxSACKBgg8BlXNRcOw9OwPxSvBS5O40BBGEysCxQOiyAAHAAGS5WiD0rHgjA-K2qaR-aWXOYARwMMTmTN4LN42rl9KoMSRhlRRGVao1WqQyNRDRStvs9qQACZLmJeDFuMrKFRWShVABWEEGelKE3fJAcnUtFpAA)
```sh
# Options (if any):
{
  "arrowParens": "avoid",
  "bracketSpacing": true,
  "htmlWhitespaceSensitivity": "css",
  "insertPragma": false,
  "jsxBracketSameLine": false,
  "jsxSingleQuote": false,
  "parser": "babylon",
  "printWidth": 80,
  "proseWrap": "preserve",
  "requirePragma": false,
  "semi": true,
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "es5",
  "useTabs": false
}
```

**Input:**
```jsx
export const testComponent = {
  template,
  controller: class TestComponent {
    constructor($ngRedux) {
      'ngInject';

      this.unsubscribe = $ngRedux.connect(this.mapState, {})(this);
    }
    
    mapState(state) {
      return {
        test: state.test,
      };
    }
  }
}
```

**Output:**
[...]

PR #5739, opened 2019-01-10
fix(javascript): skip .connect() method when composing fun

makepost:
Fixes #4935. Checks a member expression property against an ordinary method name blacklist, after the property has found a match in the existing set of function composer names.

- [x] I’ve added tests to confirm my change works.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

Issue #4935, opened 2018-08-03
Strange behavior with word `connect`


arch-mage:
**Prettier 1.14.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc05hgCgBYQM4wA0ABAA4QBOMAlANwA6UkUU6WuBJ5VdjzLbHPiJlKNBlBCEQEUjACW0PMlABDChQgB3AArqEylKoBuEeQBMpIAEYVVYANZwYAZVL35UAObIYFAK5w0p54cFQ6dl4AtqrIAGaqADah0gBWeAAeAEJ2js4uqlFwADKecPFJKSDuFKEUyDaq1gCeidBWpBSeMADqFjDYyAAcAAzSnfhwPXakDZ1wdcbl0hRwAI7+8qsRqtGxSAnJQSChUfK+Acd4nl6JcACK-hDwFUfSME195gPIAEzvdnkiRuAGEIFEYg0oGgrP5QgAVJqGQ6hAC+qKAA)

**Input:**
```jsx
conect(host, port);
connect(host, port);
connnect(host, port);
```

**Output:**
```jsx
conect(host, port);
connect(
  host,
  port
);
connnect(host, port);
```

**Expected output:**
```jsx
conect(host, port);
connect(host, port); // should not be split to multiple lines
connnect(host, port);
```

suchipi:
This is caused by https://github.com/prettier/prettier/pull/4431; see https://github.com/prettier/prettier/issues/1420 and https://github.com/prettier/prettier/issues/4172

linorabolini:
Same thing is happening to me.
My problem is that from this: 

```javascript
g.connect("A", "B")
g.connect("D", "B")
g.connect("B", "C")
g.connect("D", "C")
```

I get this:
```javascript
g.connect(
    "A",
    "B"
)
g.connect(
    "D",
    "B"
)
g.connect(
    "B",
    "C"
)
g.connect(
    "D",
    "C"
)
```

The quick solution is to change my API, but I would love to have a way to override this behaviour.

linorabolini:
Hi ! ping again, will there be a way to disable this ? I'm not even using redux :(

lydell:
@linorabolini Sorry, there's no way to turn it off other than using a `// prettier-ignore` comment.

makepost:
[...]

PR #6033, opened 2019-04-04
Javascript: Use function literals in arguments to detect function composition

### LINKED ISSUE #4751: Functional composition heuristics false positives

### LINKED ISSUE #4858: Codes format unexpected with function "pipe"

### LINKED ISSUE #5060: Functional Composition Heuristic Is Too Broad
**Prettier 1.14.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0pzDAFAMwhAGgAIAjAQwCcBKEAkCABxgEtoBnZUSiiAdwAVKCDijIA3CMwAmtECQpkwAazgwAyg0XMoAc2QwKAVzh1tbOBRj8FOgLZlkOMgBtzdAFZsAHgCEFy1TUyWzgAGW04Rxc3EE0KcwpkOTISAE9naFkGCm0YAHVpGAALZAAOAAY6bIhzPIUGJOy4BLFIugo4AEdDZg7rMjsHJCdXExBzW2Z9IzG2bR1nOABFQwh4KNG6GBSCqWLkACYthWZneYBhCFt7JKhoNpBDcwAVFJER8wBfT6A)
```sh
--parser babylon
```

**Input:**
```jsx
connect(foo, bar)
```

**Output:**
```jsx
connect(
  foo,
  bar
);

```

**Expected behavior:**
The functional composition heuristic discussed in #4172 is too broad. If I am not using Redux, the above styling is ugly and annoying. And even if I am using redux, if I don't define `mapStateToProps` and `mapDispatchToProps` inline, it's ugly and annoying to split that on multiple lines.

Rather than have prettier assume you are using all libraries, could I not list the libraries I am using in my config file? I'd rather opt in to redux specific styles than have them applied to my non-redux project.

In general though, it seems odd to hard code certain libraries anyway. It will leave prettier in the position of always judging which libraries are popular enough to support, and limits its support for less popular libraries. Why not consider an approach like the one mentioned in [this comment](https://github.com/prettier/prettier/issues/4172#issuecomment-377726296)? It seems like the eventual solution was proposed in the very next comment and no one really ever replied to this one. I think it would be much better than picking and choosing which libraries get full support.

### LINKED ISSUE #5769: Prettier formats a sensible one-liner into seven lines
**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcBuCYGUYQE5wB0ArgA4AmAhvABQCW5ANAAQD6hAZgDYQDuNZKvDws6UMAQC2mAGpw8AZzrQAlCoA6UEIxARSMZVAXJQlPHj4AFMwmMpKaCA20gARnkpgA1nGylPYgDmyDB4xHA6ABYwklwA6pF08Ar+YHBYtkl0aEkAnsjgCsY6YgryMJYegZKUyByUXGU6AFYKAB4AQh7evliU0gAyYnB1DU0grW1YQVxwAIrEEPCjjREg-oryBa6Urrk8WjqkeGIwcQwwkcgAHAAMRxZlcR6kBcdwZXgYLgQAjsR0AiVSjVWpIeqrHRlSR0EJhNZKKCBWYLJYjcFjNYwXbnciXZAAJh0oUodC4QQAwhBJDUClBoCMdMQygAVXZ2CFlAC+XKAA)
```sh
--parser babylon
```

**Input:**
```jsx
eventStore.update(id, _.flow(updater, incrementVersion))

```

**Output:**
```jsx
eventStore.update(
  id,
  _.flow(
    updater,
    incrementVersion
  )
);

```

**Expected behavior:**
Unchanged:
```jsx
eventStore.update(id, _.flow(updater, incrementVersion))

```

### LINKED ISSUE #5928: Ugly object destructing and function
I updated to latest version of prettier today and prettier added these type of changes

```diff
-    reducer(state, { payload: { id, aspects } }) {
+    reducer(
+      state,
+      {
+        payload: { id, aspects },
+      }
+    ) {
```

```diff
-export default connect(mapStateToProps, mapDispatchToProps)(Account);
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(Account);
```

Which looks ugly, makes it harder to read (first one especially) and IMHO is kinda unnecessary.

Can we add an option to disable this if it's within printWidth?

___

Prettier RC:

```
---
singleQuote: true
trailingComma: es5
```

### LINKED ISSUE #5969: Calls to functions or methods named “compose” are always split into multiple lines.
**Prettier 1.16.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucEC2AHCBnOAKAQwBoACAI1LAEoBuAHSgHpGS7VMc5yCwBrAdwIAnACbY2DONggY0EMIVIUS1elCwyAdJA65F5SrQYaMmqTLkLiBlbRBEQMmAEto2ZKGFCI-AArCEdxQCADcIZxF7EDIhHl44GABlDB5nKABzZBghAFc4BwALGDQAGwB1Aud4bBSwOETAqucQqoBPZHBsdwc03CEYX1j0tAJkADMCEtwHACtsAA8AIVi+BMSCNDgAGTS4ccnpkDn5xLT0krgARRyIeH2p-JAUoT6OsZKfKIwhNJgyiJgBWQAA4AAwOb6cMqxDAdb5SOBCEJ7BxCOAARxyzjRgwIw1GSAmDwcuDQziyuUe2DOF2utz2hIOjxgBDI-xEgOQACYHNkCM4SmcAMLoEYdKDQFEgHK4AAqrKCRNwAF9lUA)
```sh
--parser flow
```

**Input:**
```jsx
compose(a, b, c);
// "compose backwards"
esopmoc(a, b, c);
poop.compose(a, b, c);
poop.esopmoc(a, b, c);
```

**Output:**
```jsx
compose(
  a,
  b,
  c
);
// "compose backwards"
esopmoc(a, b, c);
poop.compose(
  a,
  b,
  c
);
poop.esopmoc(a, b, c);

```

**Expected behavior:**
```jsx
compose(a, b, c);
// "compose backwards"
esopmoc(a, b, c);
poop.compose(a, b, c);
poop.esopmoc(a, b, c);
```
🤔🧐🤪

brainkim:
This PR removes the hard-coded strings used to detect function composition in favor of a heuristic which counts the number of function literals passed as arguments.

In my opinion, the counter-examples are all places where it would be unreasonable to expect line-breaks (e.g. redux connect functions) or where line-breaks should occur anyways (the most common example is calling `Promise.prototype.then` with two short functions).

Related issues:
- Long curried function calls #1420
- RxJS pipe chaining get formatted on a single line #4172
- Add heuristic to format functional composition more nicely #4431
- Functional composition heuristics false positives #4751
- Improve functional composition heuristics #4758
- Codes format unexpected with function "pipe" #4858
- Composed functions not put on own line anymore after 1.14.0 update #4925
- Strange behavior with word `connect` #4935
- Functional Composition Heuristic Is Too Broad #5060
- fix(javascript): skip .connect() method when composing fun #5739
- Prettier formats a sensible one-liner into seven lines #5769
- Inconsitent formatting for one specific line (always the same line) #5898
- Ugly object destructing and function #5928
- Calls to functions or methods named “compose” are always split into multiple lines. #5969
- Add a way to disable functional composition rule #6030

fixes #4751 fixes #4858 fixes #5060 fixes #5769 fixes #5928 fixes #5969

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -286,52 +286,26 @@ foo;
 
 ### Javascript: Use function literals in arguments to detect function composition ([#6033] by [@brainkim])
 
-Previously, we used function/method names to detect function composition, so
-that we could split the arguments onto multiple lines. We now use the presence
-of function literals in arguments.
+Previously, we used names of functions and methods to detect function
+composition, so that we could split their arguments onto multiple lines. We now
+use the presence of function literals in arguments instead.
 
 <!-- prettier-ignore -->
 ```js
-const source$ = range(0, 10);
-source$.pipe(filter(x => x % 2 === 0), map(x => x + x), scan((acc, x) => acc + x, 0)).subscribe(x => console.log(x));
-const isEven = x => x % 2 === 0;
-const double = x => x + x;
-const add = (acc, x) => acc + x;
-source$.pipe(filter(isEven), map(double), scan(add, 0)).subscribe(x => console.log(x))
+// Input
+eventStore.update(id, _.flow(updater, incrementVersion));
+
 // Output (Prettier stable)
-const source$ = range(0, 10);
-source$
-  .pipe(
-    filter(x => x % 2 === 0),
-    map(x => x + x),
-    scan((acc, x) => acc + x, 0)
+eventStore.update(
+  id,
+  _.flow(
+    updater,
+    incrementVersion
   )
-  .subscribe(x => console.log(x));
-const isEven = x => x % 2 === 0;
-const double = x => x + x;
-const add = (acc, x) => acc + x;
-source$
-  .pipe(
-    filter(isEven),
-    map(double),
-    scan(add, 0)
-  )
-  .subscribe(x => console.log(x));
+);
+
 // Output (Prettier master)
-const source$ = range(0, 10);
-source$
-  .pipe(
-    filter(x => x % 2 === 0),
-    map(x => x + x),
-    scan((acc, x) => acc + x, 0)
-  )
-  .subscribe(x => console.log(x));
-const isEven = x => x % 2 === 0;
-const double = x => x + x;
-const add = (acc, x) => acc + x;
-source$
-  .pipe(filter(isEven), map(double), scan(add, 0))
-  .subscribe(x => console.log(x));
+eventStore.update(id, _.flow(updater, incrementVersion));
 ```
 
 ### Handlebars: Improve comment formatting ([#6234] by [@gavinjoyce])
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -284,6 +284,56 @@ foo;
 );
 ```
 
+### Javascript: Use function literals in arguments to detect function composition ([#6033] by [@brainkim])
+
+Previously, we used function/method names to detect function composition, so
+that we could split the arguments onto multiple lines. We now use the presence
+of function literals in arguments.
+
+<!-- prettier-ignore -->
+```js
+const source$ = range(0, 10);
+source$.pipe(filter(x => x % 2 === 0), map(x => x + x), scan((acc, x) => acc + x, 0)).subscribe(x => console.log(x));
+const isEven = x => x % 2 === 0;
+const double = x => x + x;
+const add = (acc, x) => acc + x;
+source$.pipe(filter(isEven), map(double), scan(add, 0)).subscribe(x => console.log(x))
+// Output (Prettier stable)
+const source$ = range(0, 10);
+source$
+  .pipe(
+    filter(x => x % 2 === 0),
+    map(x => x + x),
+    scan((acc, x) => acc + x, 0)
+  )
+  .subscribe(x => console.log(x));
+const isEven = x => x % 2 === 0;
+const double = x => x + x;
+const add = (acc, x) => acc + x;
+source$
+  .pipe(
+    filter(isEven),
+    map(double),
+    scan(add, 0)
+  )
+  .subscribe(x => console.log(x));
+// Output (Prettier master)
+const source$ = range(0, 10);
+source$
+  .pipe(
+    filter(x => x % 2 === 0),
+    map(x => x + x),
+    scan((acc, x) => acc + x, 0)
+  )
+  .subscribe(x => console.log(x));
+const isEven = x => x % 2 === 0;
+const double = x => x + x;
+const add = (acc, x) => acc + x;
+source$
+  .pipe(filter(isEven), map(double), scan(add, 0))
+  .subscribe(x => console.log(x));
+```
+
 ### Handlebars: Improve comment formatting ([#6234] by [@gavinjoyce])
 
 Previously, Prettier would incorrectly decode HTML entiites.
@@ -696,6 +746,7 @@ Previously, the flag was not applied on html attributes.
 ````
 
 [#5910]: https://github.com/prettier/prettier/pull/5910
+[#6033]: https://github.com/prettier/prettier/pull/6033
 [#6186]: https://github.com/prettier/prettier/pull/6186
 [#6206]: https://github.com/prettier/prettier/pull/6206
 [#6209]: https://github.com/prettier/prettier/pull/6209
@@ -704,10 +755,10 @@ Previously, the flag was not applied on html attributes.
 [#6236]: https://github.com/prettier/prettier/pull/6236
 [#6270]: https://github.com/prettier/prettier/pull/6270
 [#6289]: https://github.com/prettier/prettier/pull/6289
-[#6332]: https://github.com/prettier/prettier/pull/6332
 [#6284]: https://github.com/prettier/prettier/pull/6284
 [#6301]: https://github.com/prettier/prettier/pull/6301
 [#6307]: https://github.com/prettier/prettier/pull/6307
+[#6332]: https://github.com/prettier/prettier/pull/6332
 [#6340]: https://github.com/prettier/prettier/pull/6340
 [#6412]: https://github.com/prettier/prettier/pull/6412
 [#6423]: https://github.com/prettier/prettier/pull/6423
@@ -720,6 +771,7 @@ Previously, the flag was not applied on html attributes.
 [#6514]: https://github.com/prettier/prettier/pull/6514
 [#6467]: https://github.com/prettier/prettier/pull/6467
 [#6377]: https://github.com/prettier/prettier/pull/6377
+[@brainkim]: https://github.com/brainkim
 [@duailibe]: https://github.com/duailibe
 [@gavinjoyce]: https://github.com/gavinjoyce
 [@sosukesuzuki]: https://github.com/sosukesuzuki
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -286,9 +286,36 @@ foo;
 
 ### Javascript: Use function literals in arguments to detect function composition ([#6033] by [@brainkim])
 
-Previously, we used names of functions and methods to detect function
-composition, so that we could split their arguments onto multiple lines. We now
-use the presence of function literals in arguments instead.
+Previously, we used a set of hard-coded names related to functional programming
+(`compose`, `flow`, `pipe`, etc.) to detect function composition and chaining
+patterns in code. This was done so that prettier would not put code like the
+following call to `pipe` on the same line even if it fit within the allotted
+column budget:
+
+<!-- prettier-ignore -->
+```js
+source$
+  .pipe(
+    filter(x => x % 2 === 0),
+    map(x => x + x),
+    scan((acc, x) => acc + x, 0),
+  )
+  .subscribe(x => console.log(x));
+```
+
+However, this heuristic caused people to complain because of false positives
+where calls to functions or methods matching the hard-coded names would always
+be split on multiple lines, even if the calls did not contain function
+arguments (https://github.com/prettier/prettier/issues/5769,
+https://github.com/prettier/prettier/issues/5969). For many, this blanket
+decision to split functions based on name was both surprising and sub-optimal.
+
+We now use a refined heuristic which uses the presence of function literals to
+detect function composition. This heuristic preserves the line-splitting
+behavior above and eliminates many if not all of the false positives caused by
+the older heuristic.
+
+We encourage prettier users to try out the new heuristic and provide feedback.
 
 <!-- prettier-ignore -->
 ```js
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1210,7 +1210,7 @@ function printPathNoParens(path, options, print, args) {
         return printMemberChain(path, options, print);
       }
 
-      return concat([
+      const contents = concat([
         isNew ? "new " : "",
         path.call(print, "callee"),
         optional,
@@ -1220,6 +1220,14 @@ function printPathNoParens(path, options, print, args) {
         printFunctionTypeParameters(path, options, print),
         printArgumentsList(path, options, print)
       ]);
+
+      // We group here when the callee is itself a call expression.
+      // See `isLongCurriedCallExpression` for more info.
+      if (isCallOrOptionalCallExpression(n.callee)) {
+        return group(contents);
+      }
+
+      return contents;
     }
     case "TSInterfaceDeclaration":
       if (isNodeStartingWithDeclare(n, options)) {
@@ -4044,8 +4052,8 @@ function isFunctionCompositionArgs(args) {
 //
 // `connect(a, b, c)(d)`
 // In the above call expression, the second call is the parent node and the
-// first call is the current node whose arguments we're currently inspecting.
-function isLongCurriedCall(path) {
+// first call is the current node.
+function isLongCurriedCallExpression(path) {
   const node = path.getValue();
   const parent = path.getParentNode();
   return (
@@ -4225,28 +4233,22 @@ function printArgumentsList(path, options, print) {
     ]);
   }
 
-  if (isLongCurriedCall(path)) {
+  const contents = concat([
+    "(",
+    indent(concat([softline, concat(printedArguments)])),
+    ifBreak(maybeTrailingComma),
+    softline,
+    ")"
+  ]);
+  if (isLongCurriedCallExpression(path)) {
     // By not wrapping the arguments in a group, the printer prioritizes
-    // breaking up these arguments rather than the caller's args.
-    return concat([
-      "(",
-      indent(concat([softline, concat(printedArguments)])),
-      ifBreak(maybeTrailingComma),
-      softline,
-      ")"
-    ]);
+    // breaking up these arguments rather than the args of the parent call.
+    return contents;
   }
 
-  return group(
-    concat([
-      "(",
-      indent(concat([softline, concat(printedArguments)])),
-      ifBreak(maybeTrailingComma),
-      softline,
-      ")"
-    ]),
-    { shouldBreak: printedArguments.some(willBreak) || anyArgEmptyLine }
-  );
+  return group(contents, {
+    shouldBreak: printedArguments.some(willBreak) || anyArgEmptyLine
+  });
 }
 
 function printTypeAnnotation(path, options, print) {
@@ -5219,7 +5221,7 @@ function printMemberChain(path, options, print) {
   // If we only have a single `.`, we shouldn't do anything fancy and just
   // render everything concatenated together.
   if (groups.length <= cutoff && !hasComment) {
-    if (isLongCurriedCall(path)) {
+    if (isLongCurriedCallExpression(path)) {
       return oneLine;
     }
     return group(oneLine);
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -4049,8 +4049,8 @@ function isLongCurriedCall(path) {
   const node = path.getValue();
   const parent = path.getParentNode();
   return (
-    node.type === "CallExpression" &&
-    parent.type === "CallExpression" &&
+    isCallOrOptionalCallExpression(node) &&
+    isCallOrOptionalCallExpression(parent) &&
     parent.callee === node &&
     node.arguments.length > parent.arguments.length &&
     parent.arguments.length > 0
@@ -5219,6 +5219,9 @@ function printMemberChain(path, options, print) {
   // If we only have a single `.`, we shouldn't do anything fancy and just
   // render everything concatenated together.
   if (groups.length <= cutoff && !hasComment) {
+    if (isLongCurriedCall(path)) {
+      return oneLine;
+    }
     return group(oneLine);
   }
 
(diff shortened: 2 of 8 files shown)


brainkim:
The motivating intuition behind this heuristic is that inlined function composition only becomes unreadable when there are multiple anonymous functions within the same call.

Consider the following example from one of the earliest issues (https://github.com/prettier/prettier/issues/4172):

```js
import { range } from 'rxjs/observable/range';
import { map, filter, scan } from 'rxjs/operators';

const source$ = range(0, 10);

source$.pipe(
  filter(x => x % 2 === 0),
  map(x => x + x),
  scan((acc, x) => acc + x, 0)
)
.subscribe(x => console.log(x))
```

If we named the functions, it wouldn’t look so ugly for the function composition to happen on the same line:
```js
import { range } from 'rxjs/observable/range';
import { map, filter, scan } from 'rxjs/operators';

const source$ = range(0, 10);
const even = x => x % 2 === 0;
const double = x => x + x;
const add = (acc, x) => acc + x, 0;
// easy to understand, isn’t split up using new heuristic
source$
  .pipe(filter(even), map(double), scan(add))
  .subscribe(x => console.log(x))
```

Rather than forcing authors to name their functions we just check for function literals.

brainkim:
@duailibe Who do I ping to get this PR considered?
@suchipi original PR author?

tests/functional_composition/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -512,15 +476,12 @@ const resolve = createSelector(
...
-  (ids, objects) => ids.map(id => objects[id])
+const resolve = createSelector(getIds, getObjects, (ids, objects) =>
+  ids.map(id => objects[id])
 );
 
-const resolve = createSelector(
-  [getIds, getObjects],
-  (ids, objects) => ids.map(id => objects[id])
+const resolve = createSelector([getIds, getObjects], (ids, objects) =>
+  ids.map(id => objects[id])
```
evilebottnawi: Previously looks better

brainkim: 🤷‍♂️ can’t win every battle. Lemme just mention that all the ramda examples are now formatted [exactly as they are in the documentation](https://ramdajs.com/docs/#compose) 🙂🙃🙂🙃

evilebottnawi:
Also CI failed, need fix it

brainkim:
@evilebottnawi my bad didn’t realize docs needed to be linted

lydell:
- I like the approach of looking for function literals instead of using that list of names.
- I think most of the changed tests look better.
- I have not reviewed the implementation code. (I've just glanced at those `for` loops and made a thinking face.)
- @brainkim I'm pretty confident that if we make a change here there will be follow-up issues, so be ready for making follow-up PRs after release :)

brainkim:
@lydell

> * I have not reviewed the implementation code. (I've just glanced at those for loops and made a thinking face.)

I have converted to using iterators (`for…of`) in my personal code but couldn’t find any usage of iterators in prettier so stuck with old-fashioned incrementing for loops.

> * @brainkim I'm pretty confident that if we make a change here there will be follow-up issues, so be ready for making follow-up PRs after release :)

Sure thing.

lipis:
> * I think most of the changed tests look better.

I second that!

src/language-js/printer-estree.js
```diff
@@ -3863,40 +3863,34 @@ function isSimpleFlowType(node) {
...
+    if (
+      arg.type === "FunctionExpression" ||
+      arg.type === "ArrowFunctionExpression"
+    ) {
+      functionCount += 1;
+      if (functionCount > 1) {
+        return true;
+      }
+    } else if (arg.type === "CallExpression") {
+      for (let j = 0; j < arg.arguments.length; j++) {
```
j-f1: Should this recurse?

```js
if (looksLikeFunctionCompositionArgs(arg.arguments)) {
  return true;
}
```

brainkim: I initially thought I would have to write a recursive function as well, but after thinking about it, I decided against it because:
1. I was worried about the additional performance cost of walking the AST while walking the AST
2. checking at a depth of 1 is sufficient for all of the examples of function composition that I’ve found
3. in the course of walking the AST, any nested call expressions with function literals will split on multiple lines, which will automatically trigger the logic for splitting the parent call onto multiple lines.


If you can come up with counter-examples which would be uglier because we don’t check deeper in the tree for function literals I’ll be glad to change it, but honestly I can’t imagine a single example which wouldn’t trigger a line-split, if only because we’d blow the 80 character line budget.

j-f1: Thanks, that makes sense!

brainkim:
@evilebottnawi pretty much all of the open issues in the list in the first comment. Would you like me to explicitly reference these issues in the changelog or a commit?

#4751
#4858
#5060
#5769
#5928 <- this issue has an unrelated complaint related to object destructuring in arguments
#5969

evilebottnawi:
@brainkim just clarify what we can close after merge

brainkim:
Closes #4751 #4858 #5060 #5769 #5928 #5969

duailibe:
@brainkim 

Thanks for this, sorry I took a while. I think we did more harm than good with that change. But we'll never get everything right.

Personally, as much as I don't like going back and forth, I think this one is worth it. This will certainly raise some issues from the people who asked for the "only identifiers as arguments" case.

I'll tag @vjeux and @suchipi who had some opinions on the original change. What do you think?

src/language-js/printer-estree.js
```diff
@@ -3979,7 +3972,7 @@ function printArgumentsList(path, options, print) {
   //    )
   // here, but not
   //    process.stdout.pipe(socket)
```
duailibe: We might want to change this comment now

src/language-js/printer-estree.js
```diff
@@ -3863,40 +3863,33 @@ function isSimpleFlowType(node) {
...
+    return false;
+  }
+  let functionCount = 0;
+  for (let i = 0; i < args.length; i++) {
+    const arg = args[i];
+    if (
+      arg.type === "FunctionExpression" ||
+      arg.type === "ArrowFunctionExpression"
+    ) {
+      if (++functionCount > 1) {
```
duailibe: It's just a nit, but can you separate the increase from the `if` check? I personally like that, but we try to avoid those in codebase

src/language-js/printer-estree.js
brainkim: Moved and updated comment (let me know if you think no comment is necessary)

src/language-js/printer-estree.js
brainkim: My bad, I had it split up at first, bu noticed y’all had prefix increment operators in a couple places and changed it.

brainkim:
> This will certainly raise some issues from the people who asked for the "only identifiers as arguments" case.

Can you clarify what you mean by this case? I could take a look at it.

duailibe:
@brainkim 

```js
pipe(
  generate,
  mapToSomething,
  filter,
  reduce
);
```

I think it's reasonable to inline given the trade-offs, but it was requested and there's a chance people will complain it was reverted.

brainkim:
@duailibe

Yeah, I don’t think it’s necessary or “prettier” to deal with that case as explained in my comment [here](https://github.com/prettier/prettier/pull/6033#issuecomment-480093231). Isn’t one of the features of higher-order functions like `compose` and `flow` that you can write pretty one-liners? 🤷‍♂️

Almost all docs for higher-order functions put named functions on the same line. https://lodash.com/docs/#flow for instance

duailibe:
@brainkim We agree on this

brainkim:
@duailibe Feel free to ping if you need me to (destructively) rebase against the main branch I’m glad to help!

suchipi:
I'm anxious about changing this; last time we touched it, a lot of issues got opened where people who wanted the current formatting were asking why it got reverted. But I don't really have enough energy to defend the current implementation, so all I'm really saying is: be careful

brainkim:
@suchipi I’m fine with putting a hold on this until a major release. I’m not waiting for this PR to be merged for any reasons besides wanting to clear pending PRs in a timely fashion.

brainkim:
@duailibe Can we put a hold on this? I was looking over the issues for a second time and found an additional possible heuristic we can use from #1420, which is that if a call expression is immediately invoked (i.e. a curried function) we might consider breaking out the curried args onto multiple lines. Doing this will make connect look like this:

```js
const ArtistInput = connect(
  mapStateToProps,
  mapDispatchToProps,
  mergeProps
)(Component);
```

brainkim:
Here’s an initial pass at that logic: https://github.com/brainkim/prettier/commit/0d4c1a5a1667aa0b6d8d75e6dbe0179b4c82519f

The thing I’m struggling with is that I need to check if code would usually break in the second argument list and break in the first instead, otherwise, we’ll end up breaking a lot of tests, and I don’t yet know how to implement that because it’s logic involving multiple levels of the tree and finding out if the result before modification would normally cause a linebreak.

Lemme know if you think this should be done in a separate PR.

brainkim:
@duailibe @lydell @j-f1 @suchipi @evilebottnawi
I took a second pass at this pr last night when I couldn’t sleep. Sorry for letting it go stale just didn’t get around to working on this. I figured out a way to make curried calls break on the first group of arguments after reading [this comment about prioritizing line-breaks in sibling groups](https://github.com/prettier/prettier/issues/3014#issuecomment-376672416). The heuristic I use to detect long curried arguments is somewhat simple but it requires no changes to the test examples and covers most of the cases of ugly argument splitting described here: https://github.com/prettier/prettier/issues/1420.

brainkim:
Can someone review the new code though? I have literally no idea how the printer/builder commands work and I just sorta did guess and check until it printed what I wanted it to.

brainkim:
I added another bit of ungrouping logic for member expressions.

This reverts the breaking change where
```js
R.compose(
  Math.abs,
  R.add(1),
  R.multiply(2)
)(-4); //=> 7
```
gets printed as 

```js
R.compose(Math.abs, R.add(1), R.multiply(2))(-4) //=> 7
```

(https://github.com/prettier/prettier/pull/6033/commits/2e61e36ed2464b0903eae280d0eae666cb0c6073#diff-3bcf58dd9a3467170995691f1c4d741bL312). Now that I look at it, splitting curried calls across multiple lines feels like a good idea anyways because it’s not always obvious that you’re calling a call expression when it’s on a single line.

I also added another test case for the following example from #1420:
```
const mapStateToProps = state => ({
  users: R.compose(
    R.filter(R.propEq("status", "active")),
    R.values
  )(state.users)
});
```

At this point, I’m confident in saying almost all users won’t notice the change and we can finally get rid of using hard-coded strings (🤮🤮🤮🤮🤮)  to detect function composition.

lipis:
Is there anything else left in this PR?

brainkim:
@lipis not that I can tell. Maybe a review from someone who is familiar with the printer/builder code.

lipis:
There are bunch of approval.. I'll merge tomorrow if there are no objections.

brainkim:
Merge at your leisure! Getting it merged before the next minor release would be pretty chill though 😎😜👌

duailibe:
I was away on vacation for the last couple of weeks, sorry for the delay @brainkim

src/language-js/printer-estree.js
```diff
@@ -3976,40 +3984,55 @@ function isSimpleTemplateLiteral(node) {
...
+  for (let i = 0; i < args.length; i++) {
+    const arg = args[i];
+    if (
+      arg.type === "FunctionExpression" ||
+      arg.type === "ArrowFunctionExpression"
+    ) {
+      functionCount++;
+      if (functionCount > 1) {
+        return true;
+      }
```
duailibe: I believe you can skip the incremention and just return `true` directly :-)

And thus remove the `functionCount` entirely

src/language-js/printer-estree.js
```diff
@@ -3976,40 +3984,55 @@ function isSimpleTemplateLiteral(node) {
...
-      return functionCompositionFunctionNames.has(node.value);
+// Logic to check for args with multiple anonymous functions. For instance,
+// the following call should be split on multiple lines for readability:
+// source.pipe(map((x) => x + x), filter((x) => x % 2 === 0))
+function isFunctionCompositionArgs(args) {
+  if (args.length <= 1) {
+    return false;
+  }
+  let functionCount = 0;
+  for (let i = 0; i < args.length; i++) {
```
duailibe:
```suggestion
  for (const arg of args) {
```

duailibe:
I think it's a fair compromise, but I think at this point it's pretty obvious we won't please everyone.

evilebottnawi:
>  I think at this point it's pretty obvious we won't please everyone.
Agree, it is looks very good

src/language-js/printer-estree.js
brainkim: The count is > 1 meaning we don’t trigger function composition rules when there’s only one function (e.g. `setTimeout(resolve, 100);`).

duailibe: Oops my bad :-)

lipis:
What's the status with this one?

suchipi:
I still feel the same way as I did in https://github.com/prettier/prettier/pull/6033#issuecomment-492349702

That said, if you are fairly confident with the implementation, let's go for it

duailibe:
@lipis I'm aware of this PR, don't worry

brainkim:
I guess this is not being considered for 1.19? 😭😔😌

brainkim:
To be clear, I’m fine with this not being in the next release but maybe it could take merge priority over refactorings like https://github.com/prettier/prettier/pull/6562/files given the age of the PR. 🥺

evilebottnawi:
Added to `1.19` Milestone, need more view from prettier team

lydell:
I think we should be brave and merge this and see what feedback we get.

But.

The example in the changelog needs to better. It needs to show the worst output of the current behavior, and the best output of this PR. Currently, the example in the changelog is a little confusing, and kind of makes it look like this PR makes thing worse.

Also – have we explored looking at if arguments to a function are function calls? For example we could format like this:

```js
 .pipe(
    filter(isEven),
    map(double),
    scan(add, 0)
  )
```

Because the `.pipe` call has … at least 2 (3?) arguments and all (some?) are function calls? This might have unintended side effects though, but maybe our tests will call them out.

lipis:
> I think we should be brave and merge this

Let's do this!

tests/arrows/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -908,7 +908,10 @@ foo(c, a => b, d)
 foo(a => b, d)
 
 =====================================output=====================================
-promise.then(result => result, err => err);
+promise.then(
+  result => result,
+  err => err
+);
```
evilebottnawi: I afraid it will be annoy many

lydell: Hehe, this is one of the cases I think became better with this PR 😆

brainkim: Typically, most `promise.then` calls are long enough that they’ll be split on multiple lines anyways.

brainkim:
@lydell
I replaced the example in the CHANGELOG to a starker, simpler example (from https://github.com/prettier/prettier/issues/5769).

```js
// Input
eventStore.update(id, _.flow(updater, incrementVersion))

// Output (Prettier stable)
eventStore.update(
  id,
  _.flow(
    updater,
    incrementVersion
  )
);

// Output (Prettier master)
eventStore.update(id, _.flow(updater, incrementVersion))
```

> Also – have we explored looking at if arguments to a function are function calls?

Wouldn’t this require the type information of identifiers? I was unaware of any such functionality in the prettier code base or rules which allowed us to do this. Additionally, as I argued elsewhere I think putting the arguments to pipe on the same line looks better when the arguments consist of identifiers and not function literals.

```js
  .pipe(filter(isEven), map(double), scan(add, 0))
```

lydell:
> Wouldn’t this require the type information of identifiers?

I meant CallExpression AST nodes.

brainkim:
> I meant CallExpression AST nodes.

I think the problem with that approach is that there would be too many false positives, where a statement like:

```js
multiply(add(1, 3), add(-2, -2));
```

would be split up on multiple lines. That would be a broader, potentially more disruptive change to prettier.

evilebottnawi:
Will be great if somebody run this on big code base, but this is just a wish.
Anyway looks good and i think we can shipped this in next version.

CHANGELOG.unreleased.md
```diff
@@ -284,6 +284,30 @@ foo;
 );
 ```
 
+### Javascript: Use function literals in arguments to detect function composition ([#6033] by [@brainkim])
+
+Previously, we used names of functions and methods to detect function
```
lydell: We should mention the history here. That we used a hardcoded list of "fp" function names, and that users of non-fp code witht he same name were confused why their code broke on multiple lines. And that we’re now trying out a new heuristic that we’re excited about, and that people are welcome to provide feedback, etc. Kind of.

brainkim: I gave it a shot. I appreciate the idea that the CHANGELOG should be used to sell a specific feature and motivate users to upgrade. If you have further thoughts on the prose or language we should use lemme know.

brainkim:
Feel free to ping me for any issues caused by this PR. I promise to triage and attempt to fix any issues related to this change within a week for the next six months.

ThiefMaster:
This is screwing up the format of `createSelector` calls and making it much less readable. For example:

```diff
-export const isOngoingBooking = createSelector(
-  getDetailsWithRoom,
-  ({startDt, endDt}) => {
-    return moment().isBetween(startDt, endDt, 'day');
-  }
-);
+export const isOngoingBooking = createSelector(getDetailsWithRoom, ({startDt, endDt}) => {
+  return moment().isBetween(startDt, endDt, 'day');
+});
```

Issue #3888, opened 2018-02-04
Indent script and style tags content in *.vue files


GrahamCampbell:
**Prettier 1.10.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAzmATgSwA4wB8AOlAAQVlwAeeEWMZAJnAGYCGArgDaPCmVBZALYROsOEwAUASjL9yQpZCjoI3OADpuEAOZSA5AGEIwulASNR4+EwCEBmQKVkAvs4ruoqAPSZcBCRQIAA0IBAEONDoyKDsWFgQAO4ACvEIMSjsAG4QOEyhIOzoMMgc3OhwYQBGWOxgANZwMADKePU4ULrIMFicVSBMEGBl7BUDnZUMKXW6wuyj42EAVujUAEJ1jc0t7MJwADKdcIuVYWIweJwwAEynA+1YU8gg2f2FeLiwAOr5MAAWyAAHAAGMKfCCVb51PAvT5wKbZE5hLBwACOnBwqJm7DmCyQ5TOIEqwhwPT6A3QnV0GgAipwIPB7mEYOxqr8mADkAAWFl1HDcakmYTzF5QaDIkCcSoAFTZmUJcFcriAA)

When I have a Vue.js file:

```vue
<script>
    export default {
        mounted() {
            console.log('Component mounted!')
        }
    }
</script>
```

How do I stop prettier changing the indentation to:

```vue
<script>
export default {
    mounted() {
        console.log("Component mounted!");
    }
};
</script>
```

lipis:
Same with `<style>`

**Prettier 1.10.2**
[...]

PR #5574, opened 2018-11-28
Indent <script> and <style> tags’ content in *.vue files 

nakatanakatana:
Fixes #3888

- [x] I’ve added tests to confirm my change works.
-  (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://deploy-preview-5574--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeeBbADgGwIbwB8AOlAARmpaEAqc2+8AzmXgE5xlQQxkBmENhgLwAJmQCecXgDo5pChVQB6aqRWZcBOCSjqmYNgEssMXTmlkAChxgwjcNmQC8-QcN7PCZYJIgBXNhkAKyYACgByGAALAIBzaIiASjIAX3VlA2NTXX0YCQtdGTwocV8wJiYkMhgICDIAQiNsQRgS3nSoFSZ8wpAAGhAIUyNoJmRQdjYIAHcrdgRxlDwANwgjUQGQACM2PDAAa2kAZSx9oyg45Bg2fzhB6JgMHAB1aKNmM7A4Y8WPoxWHwkyHAlS2FyYjhgNjwcWEyD4eBwkMGoQAHgAhPaHE54DBwAAyFzgCKRKJA6OOFziFgAiv4eCSkIjkfcQGc2JC2CCVnctlhjLAXhsYsgABwABkGAogkJeeywIIFcC5KxJgw4AEd-EYODC4XhSazBpCMEZrrc2UxqXSGfAjeS2tthaJRUgAEyDG54Iw4akAYQgGHhKG4UHVIH8kJoeG2SxZkNSqSAA)✨**

PR #6077, opened 2019-04-23
feat: add new option `html-top-level-indent`

Justineo:
Added a new option: `html-top-level-indent`, which is described in https://github.com/prettier/prettier/issues/3888#issuecomment-473973916.

The new option has the following possible values:

- `"always"` - Always apply top-level indent for templates, scripts and styles.
- `"never"` - Avoid top-level indent for templates, scripts and styles.
- `"auto"` - Avoid top-level indent for scripts and styles inside Vue files.


lydell:
I like it! :+1:

Considering the last couple of comments in #3888 I'm also open to only doing `--vue-indent true/false` (should be an easy change). What to think, @azz (re https://github.com/prettier/prettier/pull/5574#pullrequestreview-212594873)?

Justineo:
@lydell Should I update the website code as well? (Not sure how to do it and may need some help.)

lydell:
I'd say we should apply it to all levels:

```html
<script>
let foo
</script>
<body>
  <script>
  let bar
  </script>
</body>
```

But this feels like another argument to making the option specific to Vue SFCs.

> Should I update the website code as well?

Do you mean the playground? I haven't touch it since the React rewrite, so I'm a bit unsure myself.

ForsakenHarmony:
To me indenting template but not script and style (in vue) by default seems like an inconsistency, should at least not be the default

Justineo:
@ForsakenHarmony Changing the default behavior will be a breaking change.

ForsakenHarmony:
I know, but maybe it should be

berrylaw:
Can u wait me ?


Berry Law

azz:
I'm not sure I like the name for `auto`, since it's vue-specific it should something like `not-vue`, no?

Justineo:
`auto` is for those who do not care about this option. We shouldn’t expose implementation details in the option value to those users IMO. It’s not Vue-specific and will apply to any embedded language inside an HTML context.

tszulc:
Is there any word on when this will go in?

lydell:
The code of this PR looks really good! I tried it out locally and realized something.

The reason I thought https://github.com/prettier/prettier/pull/5574#pullrequestreview-212594873 sounded like a good idea was because it sounded like there would be nothing Vue-specific about the option. But there is.

```
~/forks/prettier implement-3888 %  
❯ ./bin/prettier.js test.vue --html-top-level-indent auto
<template>
  <div>test</div>
</template>

<script>
test;
</script>

~/forks/prettier implement-3888 %  
❯ ./bin/prettier.js test.vue --html-top-level-indent always
<template>
  <div>test</div>
</template>

<script>
  test;
</script>

~/forks/prettier implement-3888 %  
❯ ./bin/prettier.js test.vue --html-top-level-indent never
<template>
<div>test</div>
</template>

<script>
test;
</script>
```

Notice the indentation of `<template>` with the different option values. This _does_ work as described by the help text and PR description, but didn't realize this until now.

Now I'm a little confused. What is it that users want really?

It makes sense from a logical point of view not to indent `<template>` if you're not indenting `<script>` and `<style>`. But nobody has complained about that? It feels more like there are two camps in #3888: Those who want to keep things as-is, and those who want to indent everything.

So maybe a `--vue-indent-script-and-style` boolean flag would make more sense? (On the other hand, we could just as well end up in the situation with `--jsx-bracket-same-line` where people want it in HTML as well: #5377) 

Sorry for all the back and forth! I really wanted to press the "Approve" button on this PR this morning, but now I'm not sure. Prettier's goal is a low amount of options. When we add a new one we must make sure that it is _really_ making people happier. And I'm not sure that `auto` users will be happy with `never` (if we remove `auto` in 2.0) since that also dedents `<template>`. Or maybe the would? Who knows? :cry:

bytesnz:
My understanding is the auto value will keep the indentation as it currently is with the HTML and vue formatters (the indenting `template` but not `script` and `style` seems to be one of the standards in vue) and the `always` and `never` values will change to be all indented or not indented. The PR works as I was hoping it would.

Justineo:
@lydell

> Sorry for all the back and forth! I really wanted to press the "Approve" button on this PR this morning, but now I'm not sure.

No worries. 

What I had in my mind when making this PR is:

1. provide a language-agnostic option API for all markup languages
2. keep current behavior as default to prevent a breaking change
3. having `always` but no `never` seems a little weird to me

But as you say, there's not yet huge demand for `never` yet so I'm totally okay with dropping `never` (though I personally prefer `never` but `auto` is just fine as well).

> So maybe a `--vue-indent-script-and-style` boolean flag would make more sense? (On the other hand, we could just as well end up in the situation with `--jsx-bracket-same-line` where people want it in HTML as well: #5377)

I see #5377 has a lot of upvotes, and I think it doesn't make much sense that we only allow it in JSX but not HTML (I'd say people waiting for that feature can't be very happy about it). So in this case I agree with your earliest opinion (https://github.com/prettier/prettier/issues/3888#issuecomment-471307248) to make it work beyond Vue templates.

> And I'm not sure that `auto` users will be happy with `never` (if we remove `auto` in 2.0) since that also dedents `<template>`.

If you are dropping the current behavior, I'd say it will be very likely that some users will get unhappy, whether you have `never` or not. I think we should consider whether to apply the option beyond Vue files first, then talk about possible option values and wording.

ForsakenHarmony:
`auto` to me seems like a very misleading name if it only changes something for Vue, I'd really prefer for it to not be merged in the current form.

If not making this very specific thing which most people don't like according to the polls a breaking change is important, the default value should have some kind of special name

Justineo:
@ForsakenHarmony `auto` actually means you don't care about how Prettier handles this by conventions we can find (in this case: the recommendation of the Vue.js official style guide).  And since it applies to all HTML-base language, I don't think we should add a Vue-specific option value. Maybe I should change the help text.

lydell:
- always – the easy to understand one
- auto – the Vue one
- never – the one you won't use

:thinking:

ForsakenHarmony:
@Justineo thing is it currently is Vue specific?

Justineo:
@ForsakenHarmony Yes.

kamilic:
Any plans to merge these changes in next version?

lydell:
I think the conclusion from what we’ve learned from this PR is that the `--html-top-level-indent` option sounded good on paper, but reality calls for a Vue-specific option `--vue-indent-script-and-style`.

kamilic:
I vote `--vue-indent-script-and-style`.
I think that no or few people will use this as html-specific option.
Maybe we can move this feature to internal, providing ability for`--vue-indent-script-and-style`  implementing.

Justineo:
As mentiond earlier, this PR aims to:

1. Answer the needs of Vue/Prettier users.
2. Respect the opinion of the Prettier team, to make it available for all HTML-base languages: https://github.com/prettier/prettier/issues/3888#issuecomment-471307248, https://github.com/prettier/prettier/issues/3888#issuecomment-473979489. And as @lydell said it's more future-proof and help us avoid situations like #5377.
3. Avoid introducing breaking changes.

Unfortunately, it's not ideal at last. Let's wait for a `vue-indent-script-and-style` implementation then.

lydell:
@Justineo Thank you for your hard work on this PR! I’m really sorry it did not end up merged, but sometimes a PR of exploration is what it takes to find the best solution to a problem. :bowing_man:

Justineo:
To be clear, I don't think `vue-indent-script-and-style` is a better solution. :confused:

But I respect your choice.

ForsakenHarmony:
I like this, but I just don't like `auto` as the name for the current behaviour

PR #6080, opened 2019-04-25
fix(glimmer): {{else}}{{#if}} into {{else if}} merging

dcyriller:
## Description
The first commit highlights the issue. The second fixes it.

The following code should be left untouched:
```
{{#if a}}
  a
{{else}}
  {{#if c}}
    c
  {{/if}}
  e
{{/if}}
```
Instead, it was turned into:
```
{{#if a}}
  a
{{else if c}}
  c
e
{{/if}}
```

## Checks

## Playground


tests/glimmer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -872,6 +872,16 @@ printWidth: 80
...
+
+{{#if a}}
+  a
+{{else}}
+  {{#if c}}
+    c
+  {{/if}}
+  e
+{{/if}}
+
```
evilebottnawi: Please add tests where in `if`/`else` more then one character, like:
```
{{#if a}}
  a
  b
  c
{{else}}
  {{#if c}}
    a
    b
    c
  {{/if}}
  a
  b
  c
{{/if}}
```

dcyriller: This is done 👍

dcyriller:
@evilebottnawi done, sorry for that

dcyriller:
@evilebottnawi `CHANGELOG.unreleased.md` is up to date.

evilebottnawi:
@dcyriller Thanks, can you fix CI?

dcyriller:
@evilebottnawi CI is now 🍏

I see that you squash the commits, so I don't bother to remove the last one (that was empty).

PR #6089, opened 2019-04-26
Add support for styled-jsx external styles

hongrich:
Supports all 3 tags in `styled-jsx/css`: `css`, `css.global`, and `css.resolve`.
https://github.com/zeit/styled-jsx/#external-css-and-styles-outside-of-the-component

Fixes https://github.com/zeit/styled-jsx/issues/548

Formatting template literal with `css` was already working because it
happens to be the same as what styled-components was using.
https://www.styled-components.com/docs/api#css

So this change just adds support for `css.global` and `css.resolve`.

### DIFF
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -335,22 +335,35 @@ function printGraphqlComments(lines) {
 }
 
 /**
- * Template literal in this context:
+ * Template literal in these contexts:
  * <style jsx>{`div{color:red}`}</style>
+ * css``
+ * css.global``
+ * css.resolve``
  */
 function isStyledJsx(path) {
   const node = path.getValue();
   const parent = path.getParentNode();
   const parentParent = path.getParentNode(1);
   return (
-    parentParent &&
-    node.quasis &&
-    parent.type === "JSXExpressionContainer" &&
-    parentParent.type === "JSXElement" &&
-    parentParent.openingElement.name.name === "style" &&
-    parentParent.openingElement.attributes.some(
-      attribute => attribute.name.name === "jsx"
-    )
+    (parentParent &&
+      node.quasis &&
+      parent.type === "JSXExpressionContainer" &&
+      parentParent.type === "JSXElement" &&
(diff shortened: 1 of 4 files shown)


hongrich:
I rebased this change on top the latest and added a change log entry. Let me know if there's anything else that should be done before this can be merged.

Thank you!

duailibe:
This *seems* safe, but I'm more and more afraid of formatting template literals

evilebottnawi:
@duailibe i think we should disable this by default in 2 version

hongrich:
Just rebased on top the latest, so should be mergeable again. This seem like a small enough change that should hopefully be pretty safe. It'd certainly help the developer experience in working with [`styled-jsx`](https://github.com/zeit/styled-jsx/issues/548).

I agree that it'd be good to separately take a closer look at formatting template literals as a whole for Prettier 2.0

duailibe:
@lydell @evilebottnawi Since this only applies to `css.global` and `css.resolve` it seems specific enough to be "safe". I'm willing to merge this and will start to think a solution for this problem for future versions.

WDYT?

lydell:
Let's do it!

src/language-js/embed.js
```diff
@@ -335,22 +335,35 @@ function printGraphqlComments(lines) {
...
+      parent.type === "JSXExpressionContainer" &&
+      parentParent.type === "JSXElement" &&
+      parentParent.openingElement.name.name === "style" &&
+      parentParent.openingElement.attributes.some(
+        attribute => attribute.name.name === "jsx"
+      )) ||
+    (parent &&
+      parent.type === "TaggedTemplateExpression" &&
+      parent.tag.type === "Identifier" &&
+      parent.tag.name === "css") ||
```
duailibe: FTR ``` css`` ``` was already supported because of styled components

duailibe:
The ``` css`` ``` check is now duplicated... I'll merge this as-is and refactor

Issue #6035, opened 2019-04-05
Backticks are stripped incorrectly on variable count of nested backticks


belochub:
**Prettier 1.16.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEADdACA7hATgEw3RwIwEsoN8IBXAIwBs4M6BDMAaxjM4GdNUJQhQy8KAcyYt2XHh35F0AHSggANCAgAHbtF7JQrXLghYACkYT6UrAG4Qy+dSDq4ZcGAGUt7Cchi4NHAaABYwALYMAOohZPC8PmBwnlZxZLZxAJ7I4Lz6GhS8cLgwZm7i4azIAGasDEUaAFa8AB4AQm6cHp6s4XAAMhRwNXUNIM0tnhJMAIo0EPAj9cEgPrhFuDmVuBzUWKoaWrgUMFGOMCHIABwADIcmRVFuWjlHcBu2wxq4cACONGQfmVWBUqkhassNEVwmR-IEVmIoJI4HMFsNwaMVjBWHQzvgLsgAEwaAKsMgMCQAYQg4UqOSg0C+IBoRQAKjjrBCigBfblAA)
```sh
--parser markdown
```

**Input:**
````markdown
``` word ``word in double backticks`` `word in single backticks` ```

````

**Output:**
```markdown
` word ``word in double backticks`` `word in single backticks` `

```

**Expected behavior:**

Triple backticks around the complete expression must not be replaced with single backticks since there are both single and double backticks inside the expression.

**GitHub parsing:**

``` word ``word in double backticks`` `word in single backticks` ```
` word ``word in double backticks`` `word in single backticks` `

PR #6110, opened 2019-05-10
fix(markdown): correctly determine count of backticks in inline code

belochub:
By the markdown spec, it is required to 'choose a string of n backtick
characters as delimiters, where the code does not contain any strings of
exactly n backtick characters.'
This changes the method of finding the required count of backticks from
using 2 backticks, when there is a backtick string of length 1 inside
the inline code block, and using 1 backtick in all other cases, to
finding a minimum length backtick string that can correctly be used as a
delimiter.

Fixes https://github.com/prettier/prettier/issues/6035


lydell:
Couldn’t the number of needed backticks be calculated by doing this:

```js
Math.max(0, ...(node.value.match(/`+/g) || []).map(match => match.length)) + 1
```

Or am I missing something?

belochub:
@lydell it wouldn't be the minimum length backtick string possible, consider the following example from the [existing tests](https://github.com/prettier/prettier/blob/6cee80b5b57257844fcca9310679eb019c1db04c/tests/markdown_inlineCode/__snapshots__/jsfmt.spec.js.snap#L18):
```md
`` ```123``` ``
```

In this example `node.value` is `" ```123``` "` and the snippet you provided above will return 4, but the number of needed backticks is actually 1, because there is no backtick string of length 1 inside the code span.

lydell:
Ooh, I had no idea it worked that way! I thought that the fences had to have _more_ backticks than the longest sequence. TIL.

Issue #4040, opened 2018-02-25
Idea: sort package.json keys


simonepri:
It would be really cool to integrate something like [this](https://www.npmjs.com/package/prettier-package-json) directly inside Prettier.

suchipi:
@azz @ikatyang does the current plugin API surface area make it possible to have a plugin that only formats files named exactly package.json?

ikatyang:
`filenames` field should work.

https://github.com/prettier/prettier/blob/b6c27893cc32f0ad8800b045d4e88d14e04057ba/src/language-markdown/index.js#L30

https://github.com/prettier/prettier/blob/b6c27893cc32f0ad8800b045d4e88d14e04057ba/src/main/options.js#L75-L91

suchipi:
If I'm reading that right, it looks like extensions have higher precedence over filenames, so I don't think you could make something that formats `package.json`, since the builtin `json` language would handle the json extension.

ikatyang:
Oops, looks like we should give `filenames` or later plugins higher priority.

simonepri:
What's the best way to integrate it?
We should use the [plugin API](https://prettier.io/docs/en/plugins.html) and release it as separate package?

suchipi:
Yeah, that was my idea, but it looks like we need to make a small change in order for that to be possible first.

simonepri:
Yeah I got this.
Should we discuss also about which will be the preferred order for the keys for this plugin?

j-f1:
> Should we discuss also about which will be the preferred order for the keys for this plugin?

I’d recommend just using `prettier-package-json`.

lipis:
It's under [MIT license](https://github.com/cameronhunter/prettier-package-json/blob/master/package.json#L5) but we could still ask @cameronhunter :)

j-f1:
I was suggesting using it directly (i.e. via `require`) 😀

simonepri:
@j-f1 actually there are a lots of 'orders' within the community, I believe that a short discussion on which one we want to choose will be appropriate.

Some of them:
[sort-package-json](https://github.com/keithamus/sort-package-json/blob/HEAD/index.js#L49) _(247,160 downloads in the last month)_
[fixpack](https://github.com/henrikjoreteg/fixpack#fixpack) _(9,384 downloads in the last month)_
[prettier-package-json](https://github.com/cameronhunter/prettier-package-json/blob/HEAD/src/defaultOptions.js) _(5,451 downloads in the last month)_
[...]

PR #6113, opened 2019-05-11
Allow external plugins to override internal ones

### LINKED ISSUE #5919: Plugins: allow overriding internal plugins, e.g., printer for JS or CSS
**Steps to reproduce:**
Create a plugin that exports a custom printer for JS and/or CSS:
```js
module.exports = {
  printers: {
    estree: myJSPrinter,
    postcss: myCSSPrinter
  }
}
```

Then run Prettier with this plugin on a JS or CSS file.

**Expected behavior:**
The plugin's printer is the one that's used for formatting.

**Actual behavior:**
The default printer is used, no matter what the plugin offers. That's because the array of plugins has the [internal plugins first, the external plugins last](https://github.com/prettier/prettier/blob/aec3978a134a01e62324db51985b7047cf57d71c/src/common/load-plugins.js#L93), and then the printer to use is the [first one that matches the AST format](https://github.com/prettier/prettier/blob/aec3978a134a01e62324db51985b7047cf57d71c/src/main/options.js#L109-L111).

**Why would anyone need it?**
The WordPress community uses a formatting style with a lot of spaces inside parens:
```jsx
function Label( { date } ) {
  const [ translate ] = useTranslate();
  return ( <span>{ translate( 'Updated %s', moment( date ).fromNow() ) }</span> );
}
```

We've been maintaining a [fork that adds a `--paren-spacing` option](https://github.com/automattic/wp-prettier) for the last two years. Now we would like to turn it into a plugin and publish it to NPM as `prettier-plugin-wordpress`.

Then users could install the official `prettier` package and configure it to use the `wordpress` plugin. All editor integrations that [search for](https://github.com/prettier/prettier-atom/blob/8e8be934712882a8902242891e80e3df6f793f6d/src/helpers/getPrettierPath.js#L8) `node_modules/prettier/index.js` and other packages that have `prettier` as peer dependency would use the right version.

Thoughts? Would the team be interested in a PR that swaps the order of the plugin array or searches in it using `findLast`?

remcohaszing:
This allows external plugins to override behaviour of internal prettier parsers
/ printers. This is useful for example if a plugin should override default
formatting by prettier.

Closes #5919


j-f1:
Does this work for e.g. overriding the JS parser and formatting

```html
<!-- HTML -->
<script>foo()</script>
```

remcohaszing:
I added an override for a custom JSON parser as well. @j-f1 I’m not sure what you meant with the HTML code snippet.

j-f1:
Prettier will allow you to embed languages within one another, and it will format them properly. Here’s an example:

**Prettier 1.17.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEADdArAzgHSgMwggAIAKYigSmIG5aa91UQAaECABxgEtotlQAQwBOwiAHcACiIT8UggG4RuAE1YgARsMFgA1nBgBlDju5QA5shjCArnDYALGAFsANgHUH3eFhNg4hrLe3AreAJ7I4Fj8bGZYcMIwktrmzoLI+IKu8WzYAB4AQtp6BoaCznAAMmZwGVk5IPmGZuaucACKNhDwddn2ICbC8cKRacK6KhJQ6hzCZjDuqjAOyAAcAAxssxDx7tockbNwwwq1bACOXfDJnHIgglgAtFBwcCpv6sJwl9xfyYKpdJITJ9NjxZzcKy2fpYFptTrdWrA+r9GCCDSLFTLZAAJjY1kE3FcLQAwhBnGlIlBoGcQDZ4gAVdFyEHxAC+bKAA)
```sh
--parser markdown
```

**Input:**
````markdown
```js
foo (   ) ; ;;
```
````

**Output:**
````markdown
```js
foo();
```

````

Could you add a test that verifies that replacing the inner parser (in this case the JS parser) will work properly?

duailibe:
cc @azz

remcohaszing:
@j-f1 I checked if it works for replacing inner parsers, and it doesn’t. I tried to fix it, but I’m really not familiar with the code base. In various places there appear to be hardcoded mappings of contexts to parser names.

azz:
Ideally there'd be a way to choose which plugin (or internal) to use. What if there's multiple plugins that override the parser or printer. Should they somehow chain like express middleware?

remcohaszing:
> What if there's multiple plugins that override the parser or printer?

This issue exists regardless of this PR.

> Should they somehow chain like express middleware?

_Thoughts regarding chaining:_

A parser parses code as a string into an AST. A Printer converts an AST to a string. I don’t see how they could be chained.

What could be done is something like _parser(code: string) => ast0_ → _formatter[1](ast0) => ast1_ → _formatter[0](ast1) => ast2_ → _formatter[n](ast n-1) => ast n_ → _printer(ast n) => string_.

This is basically also how Babel / PostCSS / Remark works.

This would allow plugins to make changes, such as:

- stripping any formatting, which basically implements strict mode as suggested in #2068
- sorting JSON keys, which would be the most appropriate solution for to create a plugin for #4040

This functionality would require a more complex configuration file, and give prettier plugins more options to act as a conventional linter. This sounds cool, but IMO conflicts with what prettier tries to achieve.

remcohaszing:
I’m a bit confused about the state of this PR.

This just changes some undefined behaviour and tests the new defined behaviour. Some questions arose, of which some are not directly related to this PR.

I’m unable to fix the replacing of nested parsers, as this involved a lot more in depth knowledge and fixing some hardcoded parsers.

Is this PR good enough as-is or does it need more work?

---

My goal to create a prettier plugin that runs [sort-package-json](https://github.com/keithamus/sort-package-json). This requires me to override an internal plugin, but not necessarily a nested language.

duailibe:
Hi @remcohaszing 

While this seems "simple", this is quite core to how the plugins work. We did not plan for people to write their own plugins for the languages Prettier supports "natively", and while this PR is good "as is", it's not necessarily the behavior we want to support.

I am starting to think there's value in letting users write code to change how Prettier works without having to fork, and this PR is a solution but I'm not sure it's the solution we want, it's probably something that needs more thought. It doesn't mean this PR needs more work though.

That's why @azz asked questions that aren't directly related to this PR but to this problem in general.

> This just changes some undefined behaviour and tests the new defined behaviour.

I just disagree with this. The old behaviour was defined (maybe not tested and not documented), but it wasn't an accident.

I hope you can understand it and thanks for your patience.

kachkaev:
> @remcohaszing:  
> My goal to create a prettier plugin that runs [sort-package-json](https://github.com/keithamus/sort-package-json). This requires me to override an internal plugin, but not necessarily a nested language.

Does [prettier-plugin-packagejson](https://github.com/matzkoh/prettier-plugin-packagejson) appear like what you are looking for? It worked for me, but there is an issue related to this topic: https://github.com/matzkoh/prettier-plugin-packagejson/issues/4.

remcohaszing:
> > @remcohaszing:
> > My goal to create a prettier plugin that runs [sort-package-json](https://github.com/keithamus/sort-package-json). This requires me to override an internal plugin, but not necessarily a nested language.
> 
> Does [prettier-plugin-packagejson](https://github.com/matzkoh/prettier-plugin-packagejson) appear like what you are looking for? It worked for me, but there is an issue related to this topic: [matzkoh/prettier-plugin-packagejson#4](https://github.com/matzkoh/prettier-plugin-packagejson/issues/4).

It’s very close. I’ll consider using it. Thanks!

Acidic9:
Is there any update to this? I'm attempting to develop a prettier plugin to sort CSS classes, but this issue is blocking me from further development.

tats-u:
Can someone tell me what prevents this PR from merged?
I am developing a plugin that overrides the printer of markdown to fix #6385.

thorn0:
@tats-u Did you read the comments above? Why ask then? Don't override the printer. Add a parser with a different name instead and format files with `--parser my-markdown-parser` (or rather add a config override for that).

tats-u:
@thorn0 I read azz's and duailibe's intensively.  And thank you for suggesting a plan to implement.
I am not convinced why this PR has been left without being merged, closed, or marked as "abandoned" or "WIP", though.

PR #6125, opened 2019-05-15
Add empty .prettierrc to project root

### LINKED ISSUE #6065: Unit tests fail when there is a global .prettierrc
Steps to reproduce:
1. Create a global `.prettierrc` in your home directory. Useful e.g. for editor integration plugins.
2. Go to the directory with your cloned `prettier` repo and run `yarn test`
3. Watch the following two tests fail:
```
  ● API resolveConfig with no args
  ● API resolveConfig.sync with no args
```
The resolved config is expected to be `null`, but it's an object loaded from your global `.prettierrc` file.

duailibe:
Both tests below are affected by the presence of a config file in the developer's environment:

https://github.com/prettier/prettier/blob/012b7a653e8c50a07791213d3083ad65a615ff58/tests_integration/__tests__/config-resolution.js#L82-L90

As I understand it, both tests are asserting that calling `resolveConfig` with no args follows the search rules starting from the current working directory, and tests below them assert that when specifying a path, Prettier will follow the search rules starting from another path. I think adding a `.prettierrc` in project's root to force Prettier to stop the search doesn't change what's being tested.

(The above was edited)

Fixes #6065


duailibe:
FTR this `.prettierrc` doesn't need to be empty, as I understand it this test is not about the resulting config being `null` or not. We could include a config in `.prettierrc` and assert the config

duailibe:
Oops, this doesn't work... it starts from the current working directory so it will only work if the cwd is `__tests__`. Had to move the file to the project's root

jsnajdr:
Thanks for the fix 💙

PR #6127, opened 2019-05-15
Stop converting empty JSX elements to self-closing elements

### LINKED ISSUE #5442: Auto-closed JSX tags are really annoying during development. I would like to propose a one-way option that could provide "development" and "commit" stages
**Prettier 1.15.2**

I started getting annoyed by this issue, and I found [this discussion on Twitter](https://twitter.com/wesbos/status/983717331699490818): 

> [@wesbos](https://twitter.com/wesbos/status/983717331699490818): Is there a prettier setting that will stop it from self-closing HTML elements? I’m too trigger happy with my saves and it converts \<div\>\</div> to \<div/> before I put anything inside of it

> [@tvanantwerp](https://twitter.com/tvanantwerp/status/983721724024934400): I hope someone knows. How many minutes (hours?) have I lost changing \<div/> back to \<div>\</div>...

> [@kiran_abburi](https://twitter.com/kiran_abburi/status/983721533330702337): I need this one.

> [@vigzmv](https://twitter.com/vigzmv/status/983786848114765825): Yes...! it's super annoying. I have to save to see the changes but then I got to add children inside it.

> [@AndyWynyard](https://twitter.com/AndyWynyard/status/983733360831926273): Omg this sooooo much this!!! I’ve started slowing down with the Cmd + S consciously now because of...

> [@NicklausRoach](https://twitter.com/NicklausRoach/status/983720970367193088): This would be a wonderful thing to have. I hate having to go back and “reopen” tags. Let us know if you find a solution!

### Scenario

1. Add a new `div` and get ready to add some content:

```
<div>
</div>
```
2. Switch to the browser to check something

3. Come back to VS Code and see that the div is now closed:

```
<div />
```

This happens because I use `formatOnSave`, and Auto Save with `onFocusChange`. (I know I can press ⌘+Z to undo the format, but it's still really annoying.) 

I've experimented with turning off `formatOnSave` and just using a keyboard shortcut and precommit hook. But now I really miss the auto-formatted code.

> [@wesbos](https://twitter.com/wesbos/status/983718155498508288): I love having it on every save - it saves me tons of time reformatting and indenting code. I don’t have to think about style anymore with it enabled

I would like to propose a new one-way boolean option called `closeEmptyJsxElements`. The default value would be `true`, which is the current behavior. Setting this to `false` would prevent any new empty JSX tags from being closed, but it would *not* re-open any self-closing tags. There are a lot of React components and HTML elements that *should* be self-closing, so re-opening these would also be annoying.

The idea is that a developer could set `"closeEmptyJsxElements": false` in their `.prettierrc`, which they use during development. Then they could pass a command-line argument in their git precommit hook: `--close-empty-jsx-elements true`. This would close any of the empty JSX tags where they don't need to add any more content. (They could also set up a new keyboard shortcut to run prettier with this option. It might also be known as "commit mode", versus "development mode".)

I can think of a few other options that share a similar philosophy:

* `trailingComma`
  * It's easier to add a new argument or property when you have trailing commas.
* `arrowParens`
  * It's easier to add a new argument if the parentheses are already there.
* `jsxBracketSameLine`
  * It's easier to add a new prop when the bracket is on a new line.

It might be interesting to have one-way versions of these options these during development, and then tidy up the code before you commit your changes. I'm actually not too excited about those, and I just wanted to come up with some more examples.  But the self-closing JSX tags have been really annoying, so I think this could be a good solution.

Thanks!

duailibe:
Closes #5442


CHANGELOG.unreleased.md
```diff
@@ -216,6 +216,30 @@ This changes the method of finding the required count of backticks from using 2
...
+<!-- prettier-ignore -->
+```js
+// Input
+function Foo() {
+  return <div></div>;
+}
+
+// Output (Prettier stable)
+function Foo() {
+  return </div>;
```
lydell:
```suggestion
  return <div/>;
```

duailibe: Thanks! ❤️

sompylasar:
This PR is said to be merged to `prettier:master` but `prettier:master` does not contain these commits and this logic. What happened?

duailibe:
@sompylasar I don't understand 

https://github.com/prettier/prettier/commit/eca23e111f1082b29bf1a1e4376f21b186bdc87d

https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L5530-L5535

sompylasar:
@duailibe 😮 I don't understand either! I should have uploaded screenshots, it wasn't there when I looked. 🤔 Maybe I used search that is supposed to find in latest master, maybe its index was not at the latest master, but I also checked the CHANGELOG.md and have not found any mention of this quite significant change in JSX handling, and it's not in CHANGELOG.unreleased.md either. Which version was it released in?

sompylasar:
OK I think I might have clicked "View file" on the `prettier/prettier` PR and did not notice that I landed on `duailibe/prettier` where I replaced the commit hash in the browser address bar with `master` and landed on `duailibe:master` which  is currently behind `prettier:master`.
Screen Shot 2019-07-15 at 10 43 07 AM
Screen Shot 2019-07-15 at 10 42 15 AM

sompylasar:
Found it, according to https://prettier.io/blog/2019/06/06/1.18.0.html#stop-converting-empty-jsx-elements-to-self-closing-elements-6127-by-duailibe it was released in `1.18.0`.

The `Stop converting empty JSX elements to self-closing elements` changelog entry was added to the website blog but not to the `CHANGELOG.md` file in the repo. 
- https://github.com/prettier/prettier/blob/644b419328fce055d9def7de2cecdb5731968e99/website/blog/2019-06-06-1.18.0.md#stop-converting-empty-jsx-elements-to-self-closing-elements-6127-by-duailibe 
- https://github.com/prettier/prettier/blob/644b419328fce055d9def7de2cecdb5731968e99/CHANGELOG.md#1180 the Release Notes is just a link to the blog, that's why I could not find it in the changelog itself when I looked.

duailibe:
@sompylasar Changelog for minor releases are blog posts

Alexzanderk:
Holly!!! why??? How turn on auto-close tag?

lydell:
@Alexzanderk You can use the [react/self-closing-comp](https://github.com/yannickcr/eslint-plugin-react/blob/97a9f397d2a8a3ce3b2af893bdbb86bb2c1d4480/docs/rules/self-closing-comp.md) ESLint rule to achieve that. It has autofix!

PR #90, opened 2017-01-11
Fix #79: omit unnecessary quotes in property names

bakkot:
This omits quotes for property names which don't need them, fixing #79. It doesn't touch property names in flow type annotations, since I'm not sure which quotes are semantic there.

There's currently no tests, and no effect on the snapshots. As an example of its effect, currently the below passes through prettier without change:

```js
function f({ "a": b = c }) {}
({ "0": null });
class A {
  ["d"];
  "e";
  "f"() {}
  get "g"() {}
  +"0";
}
```

~~With this PR, it becomes~~
```js
function f({ a: b = c }) {}
({ 0: null });
class A {
  ["d"];
  e;
  f() {}
  get g() {}
  +0;
}
```

Please take an especially careful look `printPropertyKey`, since I'm not entirely sure how it ought to be implemented.

Edit: per review, updated to not apply to numbers. Now it outputs

```js
function f({ a: b = c }) {}
({ "0": null });
class A {
  ["d"];
  e;
  f() {}
  get g() {}
  +"0";
}
```

PR #6132, opened 2019-05-17
Consistently add quotes around numeric literal for object properties

### LINKED ISSUE #6064: --quote-props=consistent doesn't add quotes around number but it is required for consistency
**Prettier 1.17.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAPTBeTYAHSk0wHIBGAOgAZylMYAnAVzgBoSyAmRl9l1KYiICDAAWcZqP5tOJAL4gOYgA4wAlumSgAhs2YQA7gAUDCNMhB6AbhE0ATFSABGzPWADWcGAGU1T00oAHNkAU4QCRgAWwAbAHUJTXg0QLA4P0sUzVsUgE9rMDQrVWC0aRhTDxCYvWQAMz04itUAKzRsACEPb18-PRi4ABlguEbm1pAO7D9gkLi4AEVWcXGkJpbIwOYK5mtXPVc4OJc1ZmCYBKdJZAAOWlVziAqEjzVrc7g923HVAEdVvBqhA1FYUJAoGhNBgEDAXMw4IDNIjqnpavUNpNIhUYppwvJVNDQosVmsJltVDAjtdHLckDwqR5NHF5gBhCAxOrWKDQP4gVgVAAqR3BmwqikUQA)
```sh
--parser babel
--quote-props consistent
```

**Input:**
```jsx
const x = {
  '1.0': true,
  2: true,
  "other": true,
}
```

**Output:**
```jsx
const x = {
  "1.0": true,
  2: true,
  "other": true
};

```

**Expected behavior:**

Given that we're using `--quote-props consistent`, I would expect the "2" property to be quoted but it is not.

If I [add quotes to the 2](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAPTBeTYAHSk0wHIBGAOgAZylMYAnAVzgBoSzyAmBpm07dMREBBgALOMzGMW7LlAC+IDuIAOMAJbpkoAIbNmEAO4AFIwjTIQBgG4RtAEzUgARswNgA1nBgAyhre2lAA5sgKnCCSMAC2ADYA6pLa8GjBYHAB1mna9mkAnrZgaDbqoWgyMOZeYXEGyABmBglV6gBWaNgAQl6+-gEGcXAAMqFwza3tIF3YAaFhCXAAiqwSk0gtbdHBzFXMtu4G7nAJbhrMoTBJLlLIABy06pcQVUleGraXcAf2k+oAI7reC1CAaGwoSBQNDaDAIGBuZhwYHaZG1Az1RpbabRKpxbSRITqWHhZZrDZTHbqGAnW7Oe5IXg0rzaBKLADCEDiDVsUGgAJArCqABUTpDtlVlMogA), then Prettier preserves them, but it doesn't add them in the case when they are missing.

duailibe:
Closes #6064


lydell:
I don't know if this is a feature or unexpected:

**Prettier pr-6132**
[Playground link](https://deploy-preview-6132--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAPABAXncAOlddAThNLPIqXVxAEMb8BfEAGhAgAcYBLaAZ2ShaAJ2EQA7gAURCASloA3CNwAmrEACNhtMAGs4MAMocd3KAHNkMYQFc4bABYwAtgBsA6g+7w+JsHENZb24FbwBPZHA+ATYzPjhhGEltc2daZAAzWld4tgArPlQAIW09A0NaZzgAGTM4TOzckALUQzNzVzgARRsIeAac+xATYXjhSI1aDTDXaHUOYTMYd1UYB2QADgAGNgWIePdtDkiFuDGFerYAR174ZM45cH5uPnhENmE4G+5P5NpU9JILKDNjxZzcKy2IZ8dqdHp9epAxpDGBTFYqNbIABMbGstG4rnaAGEIM40pEoNBLiAbPEACpTOTA+KMRhAA)
```sh
--parser babylon
--quote-props consistent
```

**Input:**
```jsx
x = {
  999999999999999999999: "a"
}
```

**Output:**
```jsx
x = {
  "1e+21": "a"
};

```

Side note and probably not related to this PR – with `--quote-props consistent`, why is it that adding `1: "x"` to an object literal causes all props to be quoted?

thorn0:
@lydell  It's expected:

```console
Object.keys({999999999999999999999:'a'})
["1e+21"]
```

lydell:
Yes, I know that JS works that way. But I wondered if it is expected that _Prettier_ does that change. It could just as well just add the quote: `"999999999999999999999"`. But most likely it doesn't matter – I suspect nobody has such big numbers as keys in their code.

lydell:
I'm wrong. It's correct the way it is.

```
> Object.keys({999999999999999999999: 1, "999999999999999999999": 1})
[ "1e+21", "999999999999999999999" ]
```

duailibe:
@lydell wow that was unexpected.

I started to wonder if we should do this at all, but maybe this falls in the "Prettier makes quirks in your code more explicit" bucket

duailibe:
> Side note and probably not related to this PR – with `--quote-props consistent`, why is it that adding `1: "x"` to an object literal causes all props to be quoted?

It wasn't related. It's similar to #6119.. the heuristics of what "needs" props is wrong. I'll fix in another PR, since that's more straightforward than this PR

lydell:
I think it might make sense taking a step back and fixing all the quote-props stuff in one branch.

One of the basic rules of Prettier is: We don't transform the AST.

But we've always turned string literals used as keys and that contain only a valid identifier into identifiers. On the other hand, we've never turned number-looking strings into numbers.

Then we added `--quote-props`.

- `preserve` is easy. Simply do not transform the AST.
- `consistent` now asks to quote numbers to be ... consistent.
- If so, `as-needed` should _unquote_ numbers as well, I think. Something like this: `const shouldUnquoteNumber = String(Number(key)) === key`.

I would like to see this test case:

```js
x = {
  a: 1,
  a() {},
  "a": 1,
  "a"() {},
  NaN: 1,
  "NaN": 1,
  NaN() {},
  "NaN"() {},
  1: 1,
  1() {},
  "1": 1,
  "1"() {},
  "1": 1,
  "1"() {},
  "1.0": 1,
  "1.0"() {},
  1e3: 1,
  1e3() {},
  "1e3": 1,
  "1e3"() {},
  0xa: 1,
  0xa() {},
  "0xa": 1,
  "0xa"() {},
  0b10: 1,
  0b10() {},
  "0b10": 1,
  "0b10"() {},
  [a]: 1,
  [a]() {},
}

x = {
  "needs-quotes": 1,
  a: 1,
  a() {},
  "a": 1,
  "a"() {},
  1: 1,
  1() {},
  "1": 1,
  "1"() {},
  "1": 1,
  "1"() {},
  "1.0": 1,
  "1.0"() {},
  1e3: 1,
  1e3() {},
  "1e3": 1,
  "1e3"() {},
  0xa: 1,
  0xa() {},
  "0xa": 1,
  "0xa"() {},
  0b10: 1,
  0b10() {},
  "0b10": 1,
  "0b10"() {},
  [a]: 1,
  [a]() {},
}
```

duailibe:
> I think it might make sense taking a step back and fixing all the quote-props stuff in one branch.

@lydell I disagree because I'm not yet convinced we should quote/unquote numeric literals, and having "small" commits helps if we decide to revert this in the future.

The other PR is a necessary bug fix, regardless of the decision on this PR.

duailibe:
This one is interesting too:

**Prettier pr-6132**
[Playground link](https://deploy-preview-6132--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAPABAXncAOlddXEXXAVwEYqLil0KAafAX33w2zwKJPOpuT0m3YlDhwAJgGcAtAEcyEeFNr0WIBiAgAHGAEtoU5KACGAJzMQA7gAVzCIyhMA3CHokaQAIzMmwAazgYAGVtPz0oAHNkGDMyOE0ACxgAWwAbAHVEvWUwsDhghxy9ZxyAT2RwKSNNCKk4MxgbX0iUk2QAMxM0+s0AKylUACFfAKDgkxS4ABkIuE7u3pAB1GCIyLS4AEVFeAWehJAws3qzSq8TLzK06E9tMwiYDPcYROQADgAGTXuIeozfNpKvc4KdnPNNAolHBmjpHOBDHopPBEJozHAFHp0c0TK12kgugdNPUUnoYnFDlJ1psdtD9ksYJdnhJXsgAEyaWImPRpdYAYQgKTalSg0AhIDI9QAKpdHIT6sxmEA)
```sh
--parser babylon
--quote-props consistent
```

**Input:**
```jsx
x = {
  "\u1111": 1,
}

x = {
  "\u1111": 1,
  "needs-quotes": 1
}
```

**Output:**
```jsx
x = {
  ᄑ: 1
};

x = {
  "\u1111": 1,
  "needs-quotes": 1
};

```

lydell:
I don't think we should turn strings with escapes into identifiers.

If we do quote numbers, I think we have to unquote them as well. Otherwise adding a line and then removing it again won't take you back to your initial code.

```
x = {
  1: 1,
}
```
edit →
```diff
 x = {
   1: 1,
+  "-": 1,
 }
```
prettier →
```
x = {
  "1": 1,
  "-": 1,
}
```
edit →
```diff
 x = {
   "1": 1,
-  "-": 1,
 }
```
prettier →
```
x = {
  "1": 1,
}
```

duailibe:
> If we do quote numbers, I think we have to unquote them as well.

I agree. That's why I want to keep this PR specifically about quote/unquote numbers

bakkot:
> I started to wonder if we should do this at all, but maybe this falls in the "Prettier makes quirks in your code more explicit" bucket

I think you should, for that reason.

I would propose the following heuristics:

- If the property is a quoted number which would mean exactly the same thing if unquoted, prettier should unquote it (unless an option says not to). So if you write `{ "1": x }`, prettier should change it to `{ 1: x }`.
- If the property is an unquoted number which would _not_ mean exactly the same thing if quoted, prettier should change it so that it matches what it will be interpreted as. In other words, if you write something like `{ 1.0: x }`, prettier should change it to `{ 1: x }`. I think it should do this regardless of options (except then also add quotes if necessary for `consistent`).

I implemented the first of these in #90, but backed out the change to numbers before landing. But the code is [still correct](https://github.com/bakkot/prettier/blob/be8f65d23eb0fc93f40e58a9d138a50c647752c8/src/printer.js#L1622-L1626), and still probably the right way to do it if you want to do that.

---

Some examples of what this would mean:

```js
// input
({
  "1": x,
  "2.0": x,
  3: x,
  4.0: x,
  0xDEADBEEF: x,
});
({
  5: x,
  6.0: x,
});
```
```js
// --quote-props as-needed
({
  1: x,
  "2.0": x,
  3: x,
  4: x,
  3735928559: x,
});
({
  5: x,
  6: x,
});
```
```js
// --quote-props consistent
({
  "1": x,
  "2.0": x,
  "3": x,
  "4": x,
  "3735928559": x,
});
({
  5: x,
  6: x,
});
```
```js
// --quote-props preserve
({
  "1": x,
  "2.0": x,
  3: x,
  4: x,
  3735928559: x,
});
({
  5: x,
  6: x,
});
```


This makes it clearest to the reader what's actually going on.

The only downside is that if you are for some reason using an alternative representation of a number on purpose, for example because you are going to access it in a similar way (think `let colorNames = { 0xf0f8ff: 'aliceblue' }; let name = opts[0xf0f8ff]`), then the definition and use site will no longer match.

But prettier tell can't if the author (or reader) of the code knows that unquoted object properties do this kind of weird thing where they mean something other than what they appear as, so I think it's better to represent them so their representation matches the engine's interpretation. For rare cases like in the previous paragraph, you can use a computed property name instead - `let colorNames = { [0xf0f8ff]: 'aliceblue' };`.

Pencl:
Hi there! 👋 
It seems like this branch is dead? ⚰️ 💀

Telokis:
Hi everyone!
What's the status of this PR?

evilebottnawi:
Abandoned, PR welcome

PR #6138, opened 2019-05-19
Prevent adding quotes if there's a numeric literal as key

duailibe:
```js
// input
foo = {
  a: 1,
  1: 1,
};

// output (before)
foo = {
  "a": 1,
  1: 1,
};
```


duailibe:
Note: this still happens

```js
// input
foo = {
  a: 1,
  "1": 1,
};

// output
foo = {
  "a": 1,
  "1": 1,
};
```

PR #6140, opened 2019-05-21
Keep parens around non-null assertions for new-expressions only

thorn0:
Correction for #6136.

### DIFF
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -725,14 +725,11 @@ function needsParens(path, options) {
         return false;
       }
       return true;
+
     case "TSNonNullExpression": {
-      if (
-        node.expression.type !== "Identifier" &&
-        (parent.type === "CallExpression" || parent.type === "NewExpression")
-      ) {
-        return true;
-      }
-      return false;
+      return (
+        node.expression.type !== "Identifier" && parent.type === "NewExpression"
+      );
     }
   }
 
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -728,7 +728,8 @@ function needsParens(path, options) {
 
     case "TSNonNullExpression": {
       return (
-        node.expression.type !== "Identifier" && parent.type === "NewExpression"
+        node.expression.type === "CallExpression" &&
+        parent.type === "NewExpression"
       );
     }
   }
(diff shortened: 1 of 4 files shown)


thorn0:
This solution looks good enough, but it isn't best because it adds superfluous parens e.g. in this case:

**Prettier pr-6140**
[Playground link](https://deploy-preview-6140--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEA3AhgJwATuwXmyjgHdsBPAOgA8BCACgEoBuEAGhAgAcYBLaAM7JQWTBBIAFLAiEp0qCLwAm7EACNM6MAGs4MAMpctvKAHNkMTAFc4HABYwAtgBsA6nd7wBRsHH0zPXlRPcmRwASEOEwE4TBgJTVNHdGQAM3RnGI4AKwFqACFNHT19dEc4ABkTODSMrJBc6n0TU2c4AEUrCHhazNsQI0wYzDCYci44ATBMXh5VLhnYV2UYO2QADgAGDgWIGNdNLjCFydjUGo4ARy74BO5ZEHQBAFpiOCV31Uw4a95vhPQSRSSHSfQ4MUcvAs1n6AhabU63RqILq-Rg6DUyyUq2QACYOJZ0LxnC0AMIQRzJMJQaAXEBWGIAFQxslBMQAvuygA)
```sh
--parser typescript
```

**Input:**
```tsx
var a = new y.x!();
```

**Output:**
```tsx
var a = new (y.x!)();

```

Whereas the parens are needed only to prevent `new` from accidentally hijacking a call expression. If the callee of the `new` expression doesn't start with a call expression, the parens aren't needed as there is nothing to hijack.

src/language-js/needs-parens.js
```diff
@@ -725,14 +725,11 @@ function needsParens(path, options) {
...
     case "TSNonNullExpression": {
-      if (
-        node.expression.type !== "Identifier" &&
-        (parent.type === "CallExpression" || parent.type === "NewExpression")
-      ) {
-        return true;
-      }
-      return false;
+      return (
+        node.expression.type !== "Identifier" && parent.type === "NewExpression"
```
duailibe: Why not?

```suggestion
        node.expression.type === "CallExpression" && parent.type === "NewExpression"
```

duailibe:
Thanks @thorn0 !

PR #6164, opened 2019-05-28
refactor: Switch to diff-sequences for performance

tmcw:
The successor to #4816, still chipping away at this performance bottleneck.

This switches to `diff-sequences` instead of `fast-diff`. diff-sequences is part of the Jest ecosystem, and is pretty high-quality.

Which means that it uses the same stated algorithm under the hood as before. Unfortunately, as far as I can tell, there's no module with identical results to `diff`, so the choice is either a slight change in behavior or 10s+ formatting times when you run formatWithCursor.

This improves on the previous PR because it keeps the use of a Symbol, so that behavior is also the same. And the behavior changes are all much more subtle.


tests/cursor/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -251,8 +251,8 @@ const y = 5
 <|>
   
 =====================================output=====================================
-const y = 5;
-<|>
+const y = 5<|>;
+
```
evilebottnawi: Looks like broken

tmcw: Yes, per the description, I can't find any way to fix performance without slightly changing the behavior. Given that formatWithCursor can currently churn in the `diff` method for >5s, I think it's worth the tradeoff, but it's up to the maintainers to decide.

j-f1: IMO as long as it works in most cases, it’s fine. Especially since the user can easily click back to where they were before.

duailibe:
Hi @tmcw 

Just to let you know I haven't forgotten about this 👍
I'll get to it in a few weeks

fisker:
Run a simple test, didn't see improvement


```text
[stable] 1 run: 1:08.285 (m:ss.mmm)
[next] 1 run: 1:44.659 (m:ss.mmm)
[nextBundled] 1 run: 1:06.527 (m:ss.mmm)
[stable] 2 run: 1:00.093 (m:ss.mmm)
[next] 2 run: 1:34.847 (m:ss.mmm)
[nextBundled] 2 run: 1:09.965 (m:ss.mmm)
```

<details>
<summary>Test script</summary>
```

const stable = require("prettier");
const next = require(".");
const nextBundled = require("./dist");

const text = require("fs").readFileSync(
  "./node_modules/prettier/index.js",
  "utf8"
);

for (let time = 1; time < 3; time++) {
  for (const [name, prettier] of Object.entries({
    stable,
    next,
    nextBundled,
  })) {
    const timer = `[${name}] ${time} run`;
    console.time(timer);
    for (let i = 0; i < 5; i++) {
      assert.ok(
        prettier.formatWithCursor(text, { cursorOffset: 1000, parser: "babel" })
          .cursorOffset === 1007
      );
    }
    console.timeEnd(timer);
  }
}
```
</details>

PR #6188, opened 2019-06-06
Add keywords

bromso:
For better marketing purposes I've added keywords to the package.json file. Hopefully those changes will make your wonderful tool more accessible/searchable thru eg.: https://www.npmjs.com/.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,5 +130,33 @@
     "build": "node --max-old-space-size=2048 ./scripts/build/build.js",
     "build-docs": "node ./scripts/build-docs.js",
     "check-deps": "node ./scripts/check-deps.js"
-  }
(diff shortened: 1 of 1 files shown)


thorn0:
Looks more like keyword spam to me (except for "formatter" and "code review"). Can't really think of a scenario when someone does a search by a keyword like "Angular" or "TypeScript" looking for something like Prettier. There are better keywords. E.g., "code style", "code standards", etc.

bromso:
As long as repo gets more exposure in more channels that can hopefully can result in attracting more contributors with higher knowledge in programming than me,  I'm happy because we will all benefit from that. :)

lipis:
No need

PR #6200, opened 2019-06-07
Rollup update

### LINKED ISSUE #6154: Update Rollup
We need to update it because the version we use doesn't parse some expressions like `...` which causes an error when bundling some of our dependencies.

This update isn't simple, I've tried a few times and gave up.

Opening this issue for me to keep track.

fisker:

fix #6154 

FYI:

1. `uglify` is replaced with `terser`, some es6 code come up in bundle
2. third-party external solution seem a little bit weird, but working for now, can't find a better way
3. thirdParty in `tests_integration` is using different export, hope it's not a problem


fisker:
@duailibe

duailibe:
@fisker should be. Maybe it's because of merge conflicts in `yarn.lock`?

brodybits:
> [...] some es6 code come up in bundle

Does not sound like an acceptable change to me unless Prettier does this in a new major release, unfortunately. (And I think there is some reluctance to make a new major release unless absolutely necessary at this point.)

I wonder if we should use some kind of tool that automatically verifies ES5 compliance, and look for some kind of simple solution for whatever cases may come up?

fisker:
@duailibe I tried to update from master, still conflicts

fisker:
@brainkim I understand we should ES5, just too much work for this update, I temp use `terse` to ignore some error. I will check the build and locate the problem

duailibe:
@fisker Did you update master?

duailibe:
> I wonder if we should use some kind of tool that automatically verifies ES5 compliance

@brodybits This isn't necessary, we already run tests in Node 4

scripts/build/config.js
```diff
@@ -109,7 +109,9 @@ const coreBundles = [
     input: "index.js",
     type: "core",
     target: "node",
-    external: [path.resolve("src/common/third-party.js")],
+    internalReplace: {
+      [path.resolve("src/common/third-party.js")]: "./third-party"
+    },
```
duailibe: This isn't necessary, the `external` works

fisker: No, it's not, you can try, I don't know what went wrong, a error throws, and I wasted lots of time trying to figure why

duailibe: Indeed. It's because of the `rollup-plugin-node-resolve` update

fisker:
@duailibe I saw you changed the sources, so external can't work, right?

duailibe:
@fisker You deleted some code in b08614402ba647f7f49dbe328d27dbc8d311b58e

duailibe:
> Iook like `rollup-plugin-babel` is not working as expected

?

fisker:
> > Iook like `rollup-plugin-babel` is not working as expected
> 
> ?

As I checked, before `uglify` works, code should be ES5 by babel, but it's not,  this is why `uglify` can't work

fisker:
seems `terser` for webpack has some problem, maybe minify postcss later.

sorry, not well tested

duailibe:
Yeah, we can't minify postcss-parser. I'm not sure why

duailibe:
> As I checked, before uglify works, code should be ES5 by babel, but it's not, this is why uglify can't work

I'm not sure why it didn't work, but the code is ES5 and is tested in Node 4.

duailibe:
Thank you @fisker! Great job!

fisker:
@duailibe the `third-party` code looks ugly in `src/`, also if later we want export another separate file, we have to change source again. I think we should keep my implemention

fisker:
> I'm not sure why it didn't work, but the code is ES5 and is tested in Node 4.

yes, this is why I start this pr, looks code is right, so I let you decide

duailibe:
> the third-party code looks ugly in src/

Why?

> if later we want export another separate file

That's unlikely.. ideally we wouldn't even need the `third-party.js`.. the only reason we need it is to mock it later.

fisker:
about ES5 check @brodybits mentioned before, I think there is a `eslint-plugin-es` may do the work

thorn0:
Node 4 supports a lot of ES6 stuff. See https://node.green/

duailibe:
@thorn0 You're right. It's not supposed to be ES5, it only needs to work in Node 4.

thorn0:
@duailibe Which means we can't expect uglify to parse it.

fisker:
@thorn0 @duailibe is `standalone.js` for browsers? we don't need support old browsers?

duailibe:
@fisker Yes, it's for browser. We don't necessarily need to support old browsers.

We use whatever is matched by `">0.25%", "not ie 11", "not op_mini all"` in browserslist.

fisker:
why node 4? for node I think 8 is good enough

duailibe:
@thorn0 Makes sense. But it seems terser should be able to minify anything uglify used to.

fisker:
@duailibe problem is babel seem not working right.  need more test before release

duailibe:
@fisker Ok, I didn't see any problem with babel. When you find out, let me know.

PR #6201, opened 2019-06-08
Bump `eslint`

fisker:


website/playground/VersionLink.js
```diff
@@ -16,6 +16,7 @@ export default function({ version }) {
   return ReactDOM.createPortal(
     <a
       href={`https://github.com/prettier/prettier/${href}`}
+      // eslint-disable-next-line react/jsx-no-target-blank
```
lydell: Could you change `rel="noopener"` below to `rel='noreferrer noopener'` instead of disabling the rule?

website/playground/buttons.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import ClipboardJS from "clipboard";
 
+// eslint-disable-next-line react/display-name
```
lydell: I think we can disable this rule entirely in .eslintrc instead.

website/playground/VersionLink.js
fisker: will do

fisker:
I think there is one more thing we should do, how about change `trailingComma` to `es5`?

@lydell @duailibe

package.json
```diff
@@ -19,7 +19,7 @@
     "@babel/parser": "7.2.0",
     "@glimmer/syntax": "0.38.4",
     "@iarna/toml": "2.2.3",
-    "@typescript-eslint/typescript-estree": "1.6.0",
+    "@typescript-eslint/typescript-estree": "1.9.0",
```
duailibe: this version has changed in master

PR #6202, opened 2019-06-08
minify parser-postcss bundle

fisker:

`parser-postcss.js` is now minified


fisker:
seem still not working, close for now

Issue #4081, opened 2018-03-01
Feature: handle .prettierignore location like .gitignore and .npmignore


bebbi:
Suggested feature: handle .prettierignore files like .gitignore and .npmignore files, allowing them at any level of the project, not only root.

Use-case:
Using prettier in sub-projects and their "container" projects, where they are editable both from within a sub-projects repository, as well as the container repository.
Examples: git submodules, git subtree, git subrepo, and mechanisms to sync mono- and sub-repos such as [splitsh](https://github.com/splitsh/lite)

In this case, we may need a .prettierignore file that is specific to a sub-project and is located at the sub-tree root rather than the project root. Otherwise, we need duplicate .prettierignore declarations with different paths in the sub-project repo and in the container repo.

From [git docs](https://git-scm.com/docs/gitignore):

```
Patterns read from a .gitignore file in the same directory as the path, or in any parent directory,
with patterns in the higher level files (up to the toplevel of the work tree) being overridden
by those in lower level files down to the directory containing the file.
These patterns match relative to the location of the .gitignore file.
A project normally includes such .gitignore files in its repository,
containing patterns for files generated as part of the project build.
```

From [npm docs](https://docs.npmjs.com/misc/developers):
```
Like git, npm looks for .npmignore and .gitignore files in all subdirectories of your package,
not only the root directory.
```

lipis:
Great idea..

ikatyang:
This looks like a duplicate of #3274.

duailibe:
It's not exactly a duplicate since #3274 is more of a question than a feature request..

I guess it's worth discussing if we should implement this

lipis:
Especially useful in monorepos.

duailibe:
Usually with monorepos you have a single `.prettierignore` and Prettier is installed in the root of the packages. So it's not really useful in those cases.

The use case the OP posted though (sub-repos may be edited "isolated") it might be useful but it's more rare.

I'm not convinced we need this feature but I don't oppose.

bebbi:
FWIW, we use git subtrees and similar in monorepos when externals collaborate on isolated sub-projects.

ericanderson:
Hmm, an extension of this might be to find the .prettierrc per file location. Thus:
```
child/
  src/
    file.js
  .prettierrc
src/
  file.js
.prettierrc
```

lipis:
@ericanderson  I think that one works.. (not sure though)

suchipi:
[...]

PR #6203, opened 2019-06-09
.prettierignore: resolve upwards like .gitignore

bmalehorn:
Closes https://github.com/prettier/prettier/issues/4081.

When deciding to ignore a file, allow `.prettierignore` files at any directory, not just the current directory.
For example, in this setup:

    .prettignore => *.js
    src/
        .prettierignore => !regular-module.js
        regular-module.js

`src/regular-module.js` _will_ be formatted, because it was un-ignored by `src/.prettierignore`.

This replaces the default behavior, where previously only the `.prettierignore` in the current directory would take effect.
After this PR, you can run `prettier` from any directory since the resolution is based on the directory of the file.

If passed the `--ignore-path` argument, fall back on the old implementation: only apply rules in the given file, don't resolve upwards.

In other words, you can get the old default behavior by running:

    prettier --ignore-path .prettierignore <arguments>

----


bmalehorn:
Some notes:

This call synchronous file APIs.
I should make a truly asynchronous version for `getFileInfo` API.

This could have a performance impact.
I should measure CLI performance, then cache file system reads / stats.

This installs both `ignore@4.0.6` and `ignore@5.1.2`.
`ignore@4.0.6` handles the `--ignore-path` case, since `../` paths throws exceptions on `ignore@5.1.2`.
`ignore@5.1.2` handles the new behavior, since it can tell if a file has been un-ignored.
Fortunately it's easy to install both side-by-side.

lucasconstantino:
Any news on this? It would be really useful for lots of people.

Also, some people mentioned monorepos aren't really affected by this issue, for you would have a single `.prettierignore` on root: not only this is really strict thinking, but you aren't considering monorepos with other languages - which is my current case. I don't want to pollute root with Prettier configs, as much I wouldn't like python or Java or event eslint config hanging in the root for no reason. Even `.gitignore` files should probably be specific to each inner repo's nuance.

suchipi:
Hey sorry I haven't reviewed this yet; I have been too busy to follow prettier lately and I missed the notification.

Could someone from @prettier/core who knows the history behind `--ignore-path` pick this up? cc @kachkaev

lipis:
@bmalehorn Could you also resolve the conflicts please?

bmalehorn:
@lipis I've rebased, fixing the conflicts. If more conflicts in `CHANGELOG.unreleased.md` and `yarn.lock` show up, they should be safe to ignore and just fix up at the end.

@suchipi @kachkaev Great to see this moving along! Let me know if you have any questions about the PR.

bmalehorn:
@lipis Thoughts on this PR? I realize it's a kind of long PR, but it's something a lot of people would find useful.

spitfjre:
Is there a timeline, when this can/should be reviewed?

spitfjre:
@lipis Hey, just want to know, if this is something that can/will be considered or if we need to fork, to get this kind of behaviour.

bmalehorn:
> For me it is breaking change

What specifically don't you like about this PR?

There was already some discussion in https://github.com/prettier/prettier/issues/4081, and it sounded like this approach had decent support. If you have suggestions for this PR though, I'd be happy to make some changes.

spitfjre:
Bump. Breaking changes could go in to v2.

bmalehorn:
@evilebottnawi @suchipi Any way this could be considered for 2.0? https://github.com/prettier/prettier/issues/6888

If this is considered a breaking change, I'd hope to get it in before 2.0 is solidified. This seems like an issue a lot of people run into when they try using `.prettierignore`.

evilebottnawi:
I am :-1: on this change

lipis:
I like the idea..

lydell:
@bmalehorn Do you know how ESLint handles .eslintignore?

spitfjre:
Maybe this should be done the other way and have this proposed behaviour behind a flag, so that the behaviour be a non-breaking change.

bmalehorn:
> @bmalehorn Do you know how ESLint handles .eslintignore?

OK, you got me here - ESLint only reads a single `.eslintignore` file in the current directory. But that doesn't mean that's the way it should be for prettier!

ESLint operates at the language level, so it needs to be tightly bound to a root package.json and babel compilation. Meanwhile, prettier operates at the file level, with no real "root" directory for a clump of js/css/html/md files. For these less centralized files, it makes sense to have less centralized ignore files. To me, prettier is closer to `git` - operating on files without interpreting them - so it should have a similar file-based ignore setup.

> Maybe this should be done the other way and have this proposed behaviour behind a flag, so that the behaviour be a non-breaking change.

Yeah, I'd be fine with that, since this doesn't seem likely to be merged as-is. Anyone interested in having this behind a flag, `--ignore-auto`?

bmalehorn:
This pull request is over 6 months old, I'm closing it due to lack of interest. There's no consensus from the maintainers that prettier should have this feature in the first place.

lipis:
@bmalehorn I'm sorry you feel that way, but sometimes it takes longer that usual. I added a Milestone v2.0 for this.

lipis:
I updated the target branch to `next`

bmalehorn:
@lipis Alright then, I've rebased it on `next` and fixed the conflicts.

lipis:
Linter is still complaining..

lipis:
How do you suggest implementing it, because the functionality is nice to have.

evilebottnawi:
I will look at that deeply tomorrow and give a feedback :+1:

bmalehorn:
> Linter is still complaining..

Ah yep, finally fixed them all now.

> also it is used the old dependency what can be problems in future for updating (that package can contains security problems and it will be big problem for us)

The old `ignore@4.0.6` install will be a problem regardless of this PR, because:

1. people have ignore files with `../` in them
2. `ignore@5.1.2` throws exceptions on `../`

> 👎 on that, it is break logic how `--ignore-path` works in eslint/stylelint/other similar tools

Yeah it was sort of an arbitrary decision I made during implementation. I figure there are few people who are relying on *not* loading .prettierignore files in parent directories, but there are probably some.

Instead, we could add another flag called `--ignore-auto` to enable the new behavior. This would be an alternative mode to `--ignore-path <file>` - the default behavior would be unchanged. Does that sound good, or were you thinking of something else @evilebottnawi?

evilebottnawi:
> Instead, we could add another flag called --ignore-auto to enable the new behavior. 

We prefer maximum 0CJS, so it is bad idea.

Using old version of `ignore` is really bad idea, in future we still have to remove it or migrate to another solution (do not forget about transitive dependencies which can have problems and bugs too).

Honestly, I believe that we should not change this behavior. We have node API for edge cases and I think that in this case it is better to use it, unfortunately we can’t make everything for everyone, this is the world of programming.

Perhaps in the future we will revisit the idea, so I suggest leaving the issue open. 

Honestly, I would also like to know an example of use, and why the current implementation does not solve the problem? Maybe you can create simple reproducible repo and clarify a problem in README?

danyim:
Could we merge this PR?

toplenboren:
Could we please merge this? Maybe behind a flag. Would be usable to monorepos

bmalehorn:
I'd be happy to rebase and fix up this PR, but there's still no consensus on the base premise: **do we even want this feature?**

@evilebottnawi says no, @lipis says yes. Could you two discuss this topic - what problems it might cause and why users want it?

Once you come to an agreement, we can start talking about the specifics of this PR (or close it). But after a year of this PR sitting around I really don't feel like arguing the point.

toplenboren:
> @evilebottnawi says no, @lipis says yes. Could you two discuss this topic - what problems it might cause and why users want it?

I can provide you with an example, it might come handy: 

------

We have got a monorepo with a few applications (`./apps` in https://github.com/8iq/nodejs-hackathon-boilerplate-starter-kit)

We want to ensure consistent code style around or codebase, thus we are going to add prettier.
We might have some **application-specific** folders or subfolders to be ignored by prettier

_For example_, we don't want to include some application-specific build folders (like `./out` or `./.next`) they may not be ignored by default

We have a global project-wide `.prettierignore` which tells us which apps to ignore

We want to have a application-wide `.prettierignore` which would tell us which application-specific files and folders to ignore

ArmandFrvr:
I came across this issue because we have a project that uses submodules and I would like to be able to ignore files in the submodules without having to also specify those same files in the main project.

toniopelo:
This is definitely needed in my case, and I'm sure in many other's.
There are pretty simple use case that would benefit from this IMO, this one from @toplenboren is an example : https://github.com/prettier/prettier/pull/6203#issuecomment-703645750

Furthermore, everybody using git submodules or monorepos would probably benefit from this.
I also think it's consistent with the way prettier already handle `.prettierrc` config.

Big +1 on my side, I would love to see this thread come to life again! Having this as default or behind a flag to avoid breaking changes is the same for me.

---

### Details on my situation

My company does have a monorepo with something like that: 
```
docker-compose.yml
package.json
...

./client
    .prettierrc
    .prettierignore
    package.json
    ...
./server
    .prettierrc
    .prettierignore
    package.json
    ...
```

Opening the root folder in vscode (which is the default use case of the team) make vscode prettier execute in the root folder (formatOnSave enabled) where no `.prettierignore` is present, thus formating files on save event if ignored in subfolders.
This creates inconsistent behaviors depending from where vscode is opened and make it difficult to enforce our config on every developer setup.

Having a `.prettierignore` in the root folder would mean to copy and merge the two `.prettierignore` files config in a single one and would be difficult to keep in sync. Especially in our case where we are gradually expending prettier coverage and these files are changed in almost every PR.

cc @lipis @bmalehorn

thislooksfun:
Is there any movement on this? It's been open for a really long time and it would be great to see this merged.

thislooksfun:
@bmalehorn why was this closed?

bmalehorn:
@thislooksfun I closed this due to lack of activity. While some of the prettier maintainers want this, others don't, and I don't think anyone thinks this is an essential enough feature to argue it out all the way & come up with solution that satisfies everyone.

There are still some questions that haven't been resolved:
- should prettier stop searching for .prettierignore at the root of the git repo, or keep going all the way to your home directory? How about other version control, like mercurial or svn?
- how much of a performance hit is worth adding this feature?
- what will we do about the fact that this will change behavior for users who already have nested `.prettierignore` files?

Over the past two years the only activity in this PR and https://github.com/prettier/prettier/issues/4081 have been from random people asking for this feature. So it's basically a dead PR, but still generates notifications & mentions.

I've long since worked around the issue myself by adding all rules to a root `.prettierignore` and only running prettier from the root. I'd recommend others do the same.

PR #6211, opened 2019-06-10
Use ES5-compatible syntax in `.prettierrc.js`

ryami333:
VSCode-Prettier is not able to parse the configuration example because it uses the spread operator. I presume that this is because VSCode uses an old version of Node under the hood.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -114,10 +114,11 @@ An example configuration repository is available [here](https://github.com/azz/p
 > Note: This method does **not** offer a way to _extend_ the configuration to overwrite some properties from the shared configuration. If you need to do that, import the file in a `.prettierrc.js` file and export the modifications, e.g:
 >
 > ```js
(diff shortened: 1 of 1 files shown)


lydell:
The spread seems to work for me. Here's my VSCode version:

```
Version: 1.34.0
Commit: a622c65b2c713c890fcf4fbf07cf34049d5fe758
Date: 2019-05-15T21:55:35.507Z
Electron: 3.1.8
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Linux x64 5.0.0-15-generic
```

ryami333:
Turns out the person who I was trying to assist with this issue had a year-old version of VSCode, so yeah this is probably not essential (but it's also not harmful, so I'll leave it up to you to close/merge if you want).

lydell:
Thanks! I think this is good the way it is, then.

PR #6224, opened 2019-06-15
Update many dependencies

fisker:
## incompatible dependencies

1. ~~`linguist-languages` ~~6.2.1-dev.20180706~~ -> **7.5.1**~~
   
  ~~source updated, new `TSX` language is ignored, and `.tsx` is merged into `TypeScript`, to keep support info as close as before~~

this has been moved to #6367

2. ~~`tempy`~~ ~~0.2.1~~

  ~~replaced with `temp-dir@2.0.0`~~

  ~~the new version of `tempy` is incompatible with node@4, and actually, we only use `tempy.directory()`, which is `temp-dir`~~

3. ~~`webpack` ~~3.12.0~~ -> **4.34.0** -> **4.39.1**~~
   
  ~~build script updated~~

this has been to #6366

## compatible dependencies

1. `camelcase` ~~4.1.0~~ -> **5.3.1**
2. `chalk` ~~2.1.0~~ -> **2.4.2**
3. `cosmiconfig` ~~5.0.7~~ -> **5.2.1**
4. `dashify` ~~0.2.2~~ -> **2.0.0**
5. `diff` ~~3.5.0~~ -> **4.0.1**
6. `editorconfig` ~~0.15.2~~ -> **0.15.3**
7. `escape-string-regexp` ~~1.0.5~~ -> **2.0.0**
8. `get-stream` ~~3.0.0~~ -> **5.1.0**
9.  `graphql` ~~14.2.0~~ -> **14.3.1** ->  **14.4.2**
10. `html-element-attributes` ~~2.0.0~~ -> **2.1.0**
11. `html-tag-names` ~~1.1.2~~ -> **1.1.4**
12. `jest-docblock` ~~23.2.0~~ -> **24.3.0**
13. `leven` ~~2.1.0~~ -> **3.1.0**
14. `mem` ~~1.1.0~~ -> **5.1.0**  -> **5.1.1**
15. `remark-math` ~~1.0.4~~ -> **1.0.6**
16. `resolve` from ~~1.5.0~~ -> **1.11.0** -> **1.12.0**
17. `semver` ~~5.4.1~~ -> **6.1.1** -> **6.3.0**
18. `string-width` ~~3.0.0~~ -> **4.1.0**
19. `unified` ~~6.1.6~~ -> **7.1.0** -> **7.3.2**
20. ~~`@babel/core` ~~7.2.0~~ -> **7.4.5**~~ this has been moved to #6368
21. `builtin-modules` ~~2.0.0~~ -> **3.1.0**
22. `cross-env` ~~5.0.5~~ -> **5.2.0**
23. `execa` ~~0.10.0~~ -> **1.0.0** -> **2.0.3**
24. `jest-junit` ~~5.0.0~~ -> **6.4.0** -> **7.0.0**
25. `jest-watch-typeahead` ~~0.1.0~~ -> **0.3.1**
26. `rimraf` ~~2.6.2~~ -> **2.6.3**
27. ~~`rollup` ~~1.14.3~~ -> **1.15.6**~~ this has been moved to #6365
28. ~~`rollup-plugin-node-resolve` ~~5.0.1~~ -> **5.0.3**~~ this has been moved to #6365
29. `shelljs` ~~0.8.1~~ -> **0.8.3**
30. `strip-ansi` ~~4.0.0~~ -> **5.2.0**
31.  ~~`eslint-config-prettier` ~4.3.0~ -> **5.0.0**~~ this has been moved to #6364

## Also, some dependencies I tried to update but failed

1. `globby`
2. `ignore`
3. `snapshot-diff`


scripts/build/bundler.js
```diff
@@ -176,13 +176,19 @@ function getWebpackConfig(bundle) {
       path: path.resolve(root, "dist"),
       filename: bundle.output,
       library: ["prettierPlugins", bundle.name],
-      libraryTarget: "umd"
+      libraryTarget: "umd",
+      // https://github.com/webpack/webpack/issues/6642
+      globalObject: "new Function('return this')()"
     },
     plugins: [
```
smuszel: In webpack 4 node env is [defined automatically](https://www.bennadel.com/blog/3598-webpack-4-automatically-makes-process-env-node-env-available-in-your-javascript.htm). Can remove the define plugin.

fisker: removed

package.json
```diff
@@ -22,35 +22,35 @@
...
+    "editorconfig": "0.15.3",
     "editorconfig-to-prettier": "0.1.1",
-    "escape-string-regexp": "1.0.5",
+    "escape-string-regexp": "2.0.0",
     "esutils": "2.0.2",
     "find-parent-dir": "0.3.0",
     "find-project-root": "1.1.1",
     "flow-parser": "0.84.0",
-    "get-stream": "3.0.0",
+    "get-stream": "5.1.0",
```
brodybits: I think get-stream version `5.1.0` requires Node.js >= 8: https://github.com/sindresorhus/get-stream/blob/v5.1.0/package.json#L12-L14

scripts/test-dist.js
```diff
@@ -4,7 +4,7 @@
 
 const path = require("path");
 const shell = require("shelljs");
-const tempy = require("tempy");
+const tmpDir = require("temp-dir");
```
SimenB: this is different - before you created a new directory inside of the OS temp dir - this just returns the path of existent temp dir without creating a new directory inside of it. So you run `npm init` the root temp dir, which is probably not what you want

package.json
fisker: This is not a  problem, cause this package willbe transformed in  finale bundle

package.json
```diff
@@ -22,35 +22,35 @@
...
+    "jest-docblock": "24.3.0",
     "json-stable-stringify": "1.0.1",
-    "leven": "2.1.0",
+    "leven": "3.1.0",
     "lines-and-columns": "1.1.6",
-    "linguist-languages": "6.2.1-dev.20180706",
+    "linguist-languages": "7.5.1",
     "lodash.uniqby": "4.7.0",
-    "mem": "1.1.0",
+    "mem": "5.0.0",
```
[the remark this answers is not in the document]
fisker: same as get-stream, I will test 5.1.0 later

scripts/test-dist.js
fisker: you are right, I didn't think through. I will revert that update when I back on desktop. thanks

fisker: I have revert that change

fisker:
@brodybits 

The reason I did't do separate PRs is because there is really lots of deps, and I update and fix code in the same time

fisker:
This PR is ready for review again, hope this can get merged.  these deps are only the easy part, some other outdated deps need some hard work, this is the real problem.

duailibe:
@fisker the dependencies need to work on Node >= 6, because that's the minimum version we support when installing directly from Github

lipis:
We should install dependabot :)

evilebottnawi:
@lipis agree, a lot of deps outdated :disappointed:

fisker:
@evilebottnawi I can do that, but I have a feeling, this won't get merged

fisker:
I support that bundle should work on old version of node, but I don't understand why we need devDependencies work on node 4/6.

I can't image, someone install from github and still working on node 4/6.

fisker:
All dependencies requires source code changes has been moved to separate PRs

lipis:
All tests are green! Let's merge it ;)

fisker:
> All tests are green! Let's merge it ;)

How about node 4 & 6 ? 
Are we going to drop support for installing from github on node 4?

lydell:
I thought the we run the tests in Node.js 4 in CI?

fisker:
@lydell test is runing on bundle, not source, check reviews above

brodybits:
> How about node 4 & 6 ?

I am able to install and run Prettier from GitHub on Node.js 6 but not on Node.js 4. I think this is working as intended.

If I would try the following on Node.js 6:

    npm i -g github:fisker/prettier#update-dependencies

then `prettier --version` gives me the following output:

```
/Users/brodybits/dev/nvs/node/6.17.1/x64/lib/node_modules/prettier/node_modules/get-stream/index.js:12
async function getStream(inputStream, options) {
      ^^^^^^^^

SyntaxError: Unexpected token function
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:549:28)
    at Object.Module._extensions..js (module.js:586:10)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
    at Module.require (module.js:504:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/brodybits/dev/nvs/node/6.17.1/x64/lib/node_modules/prettier/src/common/third-party.js:6:14)
```

I had already raised experimental WIP PR #6232 that includes dev integration testing on Node.js 6, which could have helped us avoid this kind of a regression.

I would also strongly favor dropping support for the old Node.js versions asap, as discussed in #3919.

I would really like to thanks @fisker for all of the efforts contributed to make these updates and make the separate PRs.

brodybits:
@fisker you can see in brodybits/prettier#1 (PR on my own fork) a test for running from git on Node.js version 6. It would fail in case of a breaking change like the get-stream@5 update (you can see the failure in brodybits/prettier#2).

My attempts to add full dev test to Azure Pipelines on Node.js 6 did not succeed on Linux, mac OS, or Windows (see brodybits/prettier#3).

duailibe:
> Are we going to drop support for installing from github on node 4?

@fisker We already did that. `package.json` has `node >= 6` and the bundle has `node >= 4`

fisker:
We should not merge this before we drop node 6 support, see previous discuss

evilebottnawi:
@fisker can we update only deps what support node@6

evilebottnawi:
@fisker we have many tests, so we catch problem for node@6 on CLI for some deps

fisker:
no, we can't catch it  get-stream@5 not support node 6 but test passed

fisker:
production will work , but not directly install from repo

evilebottnawi:
@fisker maybey `get-stream` support node@6, only `engines` will updated

fisker:
but npm install will fail unless ignore version requirement, this is not suppose to install in this way

evilebottnawi:
@fisker so no dep what we can update? I think no

fisker:
Thats not what I mean, just because too many deps in this PR, its hard to check one by one 

I made this PR, of course I want it merged, but I want you know what is the problem

evilebottnawi:
@fisker we can update some in deps in multiple PRs, not in one, what is problem update only deps support node@6?

evilebottnawi:
I think we can update some dependencies too https://david-dm.org/prettier/prettier, like `@glimmer/syntax`, `yaml`, maybe other

fisker:
Since we've drop node 6 support for direct install from github, I'm checking this deps, and sending separate PRs. then I'll close this one

fisker:
most of those dependencies updated, closing this

PR #6225, opened 2019-06-15
Rewrite FastPath class

fisker:
Rewrite FastPath to ES class, and use rest parameters

### DIFF
--- a/src/common/fast-path.js
+++ b/src/common/fast-path.js
@@ -1,33 +1,5 @@
 "use strict";
 
-const assert = require("assert");
-
-function FastPath(value) {
-  assert.ok(this instanceof FastPath);
-  this.stack = [value];
-}
-
-// The name of the current property is always the penultimate element of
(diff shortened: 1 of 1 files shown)


duailibe:
@fisker This is a hot path and babel transpiles [like this](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUAjAhgJwDSIB0xuA5gM4CUiA3gFCKIQLlwA2ApoW3KWmVQDcdAL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.4.5&externalPlugins=), which would be a performance regression.

Thanks for the intention, but the code was fine originally, we don't **need** to use every new ES feature.

PR #6232, opened 2019-06-17
[EXPERIMENTAL WIP] Azure Pipelines testing on Node.js 6, 8, and 12

brodybits:
UPDATED - EXPERIMENTAL WIP: Add Azure Pipelines testing on Node.js versions 6, 8, and 12

- do dev integration test (`yarn test-integration`) on Node.js versions 6 & 10
- add normal dev test on Node.js 8 & 12
- try get-stream@5.1.0 update, which fails the dev integration test on Node.js version 6 due to a breaking change
- try `yarn install --production=true` on Node.js 6 & 10 - fails on Node.js 6 due to <https://github.com/yarnpkg/yarn/issues/6323>

Note that dev integration testing is needed on Node.js version 6 since `yarn test` on Node.js 6 seemed to crash on Azure Pipelines.

I think these test updates are needed to ensure that dependency updates do not cause failures on other Node.js versions such as 6. For example: <https://github.com/prettier/prettier/pull/6224#issuecomment-502858408>

### DIFF
--- a/.azure-pipelines/jobs/dev-integration-test.yml
+++ b/.azure-pipelines/jobs/dev-integration-test.yml
@@ -1,0 +1,9 @@
+steps:
+  - template: ../steps/install-nodejs.yml
+  - template: ../steps/install-dependencies.yml
+    parameters:
+      flags: --ignore-engines # needed for yarn install to succeed on Node.js 6
+  - script: yarn test-integration
+    displayName: "Run tests"
+  - template: ../steps/publish-test-results.yml
+  - template: ../steps/publish-code-coverage.yml
--- a/.azure-pipelines/jobs/dev-test.yml
+++ b/.azure-pipelines/jobs/dev-test.yml
@@ -1,6 +1,8 @@
 steps:
   - template: ../steps/install-nodejs.yml
   - template: ../steps/install-dependencies.yml
+    parameters:
+      flags: --ignore-engines # needed for yarn install to succeed on Node.js 6
   - script: yarn test
     displayName: "Run tests"
   - template: ../steps/publish-test-results.yml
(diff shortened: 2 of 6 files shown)


duailibe:
This is testing the prod build, which we already test in Node 4 and 10, so it's really unlikely that it would break in 6, 8 or 12.

brodybits:
I still think we should do some kind of testing on all recent and all supported versions of Node.js.

For the issue discussed in <https://github.com/prettier/prettier/pull/6224#issuecomment-502858408>, I think we should have a step that ensures installation from GitHub does not break on Node.js 6. It should be possible to do this by the following command: `yarn --prod` but it doesn't seem to work due to <https://github.com/yarnpkg/yarn/issues/6323>.

I think this is a major problem with supporting old Node.js versions. I would really favor a new major release that drops support for the old Node.js versions 4 & 6.

brodybits:
I reworked the changes to do dev integration testing on Node.js versions 6 & 10, and add normal dev testing on Node.js versions 8 & 12. I was able to get a green build after a few tries. I then tried get-stream@5.1.0 update, which fails as expected on dev integration testing on Node.js version 6.

I had to do dev integration testing on Node.js 6 since the normal `yarn test` on Node.js 6 seemed to crash on Azure Pipelines.

I also tried adding a step to do `yarn install --production=true` on Node.js 6 & 10, fails on Node.js 6 due to <https://github.com/yarnpkg/yarn/issues/6323>. The Yarn bug is very unfortunate since `yarn install --production=true` would have been another good test on incoming PRs for Node.js 6.

Here are a few conclusions so far:

- I really think we need some additional testing on intermediate Node.js versions such as version 6, to avoid breaking changes such as get-stream@5.1.0 update (see <https://github.com/prettier/prettier/pull/6224#issuecomment-502858408>)
- We need to test that installation from GitHub does not break on intermediate Node.js versions such as 6, which could have been done by `yarn install --production=true` if not for the Yarn bug
- Maintaining support for old Node.js versions seems to cost us lots of times and block what I think are some badly needed updates. I think it would be much better if Prettier would make a new major release that drops the support for old Node.js versions.

PR #6240, opened 2019-06-20
Add `jsxSelfClosing` option

meriadec:
provide a way to put back the behavior removed in [#6127](https://github.com/prettier/prettier/pull/6127)
btw, this project is dope! :rocket: 

nb: pretty huge diff in `tests/jsx/__snapshots__/jsfmt.spec.js.snap` :man_shrugging: is there a way to make snapshots update more "diffable"? seems that lot of code just moved, and I'm not rly used to snapshot testing

### DIFF
--- a/src/language-js/options.js
+++ b/src/language-js/options.js
@@ -48,6 +48,13 @@ module.exports = {
     default: false,
     description: "Use single quotes in JSX."
   },
+  jsxSelfClosing: {
+    since: "1.19.0",
+    category: CATEGORY_JAVASCRIPT,
+    type: "boolean",
+    default: false,
+    description: "Self-close empty JSX elements"
+  },
   quoteProps: {
     since: "1.17.0",
     category: CATEGORY_JAVASCRIPT,
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5514,6 +5514,10 @@ function printJSXElement(path, options, print) {
   const n = path.getValue();
 
   if (n.type === "JSXElement" && isEmptyJSXElement(n)) {
+    if (options.jsxSelfClosing) {
+      n.openingElement.selfClosing = true;
+      return path.call(print, "openingElement");
+    }
     return concat([
       path.call(print, "openingElement"),
       path.call(print, "closingElement")
--- a/tests/jsx/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/jsx/__snapshots__/jsfmt.spec.js.snap
@@ -312,6 +312,85 @@ const TodoList = ({ todos }) => (
 ================================================================================
 `;
 
+exports[`array-iter.js 5`] = `
+====================================options=====================================
+jsxSelfClosing: true
+jsxSingleQuote: false
+parsers: ["flow", "babel", "typescript"]
+printWidth: 80
+singleQuote: false
+                                                                                | printWidth
+=====================================input======================================
+const UsersList = ({ users }) => (
+  <section>
+    <h2>Users list</h2>
+    <ul>
+      {users.map(user => (
+        <li key={user.id}>{user.name}</li>
+      ))}
+    </ul>
+  </section>
+)
+
+const TodoList = ({ todos }) => (
+  <div className="TodoList">
+    <ul>{_.map(todos, (todo, i) => <TodoItem key={i} {...todo} />)}</ul>
+  </div>
+);
+
+<div className="search-filter-chips">
+    {scopes
+        .filter(scope => scope.value !== '')
+        .map((scope, i) => (
+            <FilterChip
+                query={this.props.query}
+                onFilterChosen={this.onSearchScopeClicked}
+                key={i}
+                value={scope.value}
+                name={scope.name}
+            />
+        ))}
+</div>
+
+=====================================output=====================================
+const UsersList = ({ users }) => (
+  <section>
+    <h2>Users list</h2>
+    <ul>
+      {users.map(user => (
+        <li key={user.id}>{user.name}</li>
+      ))}
+    </ul>
+  </section>
+);
+
+const TodoList = ({ todos }) => (
+  <div className="TodoList">
+    <ul>
+      {_.map(todos, (todo, i) => (
+        <TodoItem key={i} {...todo} />
+      ))}
+    </ul>
+  </div>
+);
+
+<div className="search-filter-chips">
+  {scopes
+    .filter(scope => scope.value !== "")
+    .map((scope, i) => (
+      <FilterChip
+        query={this.props.query}
+        onFilterChosen={this.onSearchScopeClicked}
+        key={i}
+        value={scope.value}
+        name={scope.name}
+      />
+    ))}
+</div>;
+
+================================================================================
+`;
+
 exports[`attr-comments.js 1`] = `
 ====================================options=====================================
 jsxSingleQuote: false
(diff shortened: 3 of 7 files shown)


j-f1:
Why is this option necessary?

meriadec:
multiple reasons:

- enforce code convention and consistency (allow having `<Foo />` and `<Bar></Bar>` elements instead of unifying them can be seen as a regression, at least for some devs including me). Enforcing same code formatting is the prettier purpose.
- automatically fix the `react/self-closing-comp` eslint rule, which is btw enabled in [airbnb's eslint config](https://github.com/airbnb/javascript/blob/4539dbcf17b960ab650665bc95421dc8d0d3276e/packages/eslint-config-airbnb/rules/react.js#L237)
- the [initial issue](https://github.com/prettier/prettier/issues/5442) was ppl complaining about self-closing _during development_ because in _some cases_ they wanted to add content later inside :thinking_face:. Sounds like a fragile reason to remove a handy automatic formatting :) 
- having an option to control that was even [mentioned](https://github.com/prettier/prettier/issues/5442#issuecomment-487062764) in the original issue

IMHO the option should even be reversed: always enforce closing empty JSX elements except you pass a `jsxPreventSelfClosing`, **but** I don't want to bargain on that, I'm fine with putting an additional setting on all my projects.

j-f1:
Fortunately, that rule [is fixable](https://github.com/yannickcr/eslint-plugin-react/blob/7d449a9e4a1330ab950df4b55348d6eff5d025af/lib/rules/self-closing-comp.js#L25), so you can run `eslint --fix` and Prettier won’t mess with the result.

meriadec:
I don't really get your answer. Yeah eslint can "fix" few stuff but it feels weird to rely on a linter to do code formatter job. I mean, eslint can also fix replacing single quotes by double quotes, so why is it in prettier also? :smile:

Again, please understand that it can be seen as regression to not anymore have this:

```
<Foo></Foo>
<Bar />
```

transformed to

```
<Foo />
<Bar />
```

the PR provides a non-intrusive option to have this legit behavior back. What could be a blocker to merge?

j-f1:
https://prettier.io/docs/en/option-philosophy.html

meriadec:
I think I've detailed in multiple points why the concern is legit, so I don't see how it's relevant to link this :thinking: ? And furthermore, it's not a code style option, it's meant to unify empty JSX elements formatting.

If you track carefully [the original issue](https://github.com/prettier/prettier/issues/5442#issue-379553445) the whole thing originated because of a **text editor configuration** (combination of `formatOnSave` + auto save on `onFocusChange`) was causing trouble in combination with prettier.

One could have answered

> Fortunately, this is fixable by disabling auto save on focus change on your editor

instead of removing a whole feature from prettier.

So, I personally would totally :100: agree with just reverting https://github.com/prettier/prettier/pull/6127, and so: no additional prettier option (even better)!

My only purpose with adding an option was to be as unobtrusive as possible.

duailibe:
@meriadec Thank you for the suggestion and PR, but I don't think this behavior will be changed, nor an option should be added. We generally *try* to avoid changing ASTs in Prettier and only change whitespaces and whatnot. We do have some notable exceptions, and this was one of the exceptions that we decided it wasn't necessary.

For example, we could change:

```js
<Component foo={true} bar={"spam"} />

// to

<Component foo bar="spam" />
```

But we don't do that, and the change of preserving `<Foo></Foo>` is more aligned with the Prettier philosophy. I'm sorry you relied on the previous behavior, but I think a linter is a more suitable tool to do this kind of transformations.

Hope you can understand the rationale.

j-f1:
Sorry to shoot down your first PR to Prettier like this @meriadec 😞

meriadec:
First, thanks @duailibe for taking the time to answer :)

Don't want to argue indefinitely on that, I'm disappointed to see this feature leaving Prettier because of a text editor setting problem for few people :writing_hand: :book: :thinking:.

Still, Prettier is converting:

```js
function foo(
) {
}

// to

function foo() {}
```

which produce same kind of problem for the same "auto save on blur" editor configuration :man_shrugging: , why keeping it?

TL;DR I honestly think this is a mistake, discussed the subject with team and we all share the same surprise on the PR reaction (don't count that as an argument). Seems that once something is merged in codebase it is hard to discuss it :no_mouth:.

But no salt here, it's not a big deal, we will just manually self-close all those empty elements or more likely use a previous prettier version / a fork.

Good continuation.

j-f1:
The difference is that with the function example, you’re still able to easily add parameters by simply typing between the parens, whereas to add children to a self-closing JSX element, you have to manually add the closing tag, which can be long if the tag has a long name. Additionally, as I said earlier, you can add an ESLint rule that will automatically make empty JSX tags self-closing, and running `eslint --fix` on save will result in the same behavior as older versions of Prettier in this respect.

name-k:
Had to downgrade to 1.17.1, 100% agree with @meriadec. 
A fork would be nice to have with optional jsx tags self-closing.

joefiorini:
Not to throw more fuel on this, since it seems pretty decided, but I feel I have a decent argument to the productivity claims here..

One of the things I love the most about prettier is that I can get my desired formatting without having to worry about setting up a linter as well. I love it so much I have actually changed my preferences to work better with prettier's defaults, so I am very much not against worthwhile changes.

That said, I often write components that have self-closing tags (ie. routes, some data table libraries, Formik's `Field` component) and having them automatically self-close has been a time-saver for me since I started using prettier in the 0.1/0.2 days.

> you have to manually add the closing tag, which can be long if the tag has a long name.

Maybe my setup in vscode is different, but when I was ready to add children to auto-closed tags all I had to do was remove the `/>` on the end and retype the `>`, then my editor would auto insert the closing tag for me. Now instead I have to backspace out the cloning tag, which can be long if the tag has a long name.

> https://prettier.io/docs/en/option-philosophy.html

I've loved prettier's opinionated stance from the very beginning. However, it has made some of the opinions flexible because of demand from the community, ("Prettier has to strike a balance between ideal goals and listening to the community." – directly from the above link) so I don't understand why this can't be one of those cases.

> We generally try not to change ASTs in Prettier and only change whitespaces and other things. We do have some notable exceptions, and this was one of the exceptions, but we decided it wasn't necessary as it didn't bring much benefit.

That's an interesting point, and as a developer who hasn't worked with ASTs much, I can't really appreciate the finer points of why this is important here (perhaps an ELI5 on one of the "About" pages would help). As a user of prettier it's all about how much of my desired formatting I can without having to setup additional tooling like eslint. I have never felt the need for anything additional, but this is one case where I think prettier was doing the right before and could be a little more flexible.

Like I said, I just wanted to provide my viewpoint on this a long-time member of the community who tends to agree with prettier's opinions. I appreciate all the hard work from everyone involved in this project that I can no longer code without. If the final decision is that this is one place prettier must remain inflexible, then I'll adjust. But I am curious to see how others feel about this.

simonhaenisch:
I don't get this. Prettier stops prettifying our code and we can't even have an option to make it do its job again? IMO turning this **off** (which is taking away a feature) should have been behind an option in the first place (which sounds like a legit option)!  
@duailibe @j-f1 I assume we should rather argument against #6127? (which is too late now)

> * the [initial issue](https://github.com/prettier/prettier/issues/5442) was ppl complaining about self-closing _during development_ because in _some cases_ they wanted to add content later inside :thinking_face:. Sounds like a fragile reason to remove a handy automatic formatting :)

☝️exactly this. If you want to add content later, just add a dummy text or sth into it. Or you just hit Cmd+Z which undoes solely the auto-formatting.

Editors auto-close html tags, so now I have a lot of `<br></br>` and the likes that I need to manually take care of... and no, I don't want to add linter auto-fixing into my format-on-save toolchain :roll_eyes:.

> And furthermore, it's not a code style option

Exactly that as well. How is "just in case I want to put sth in there" a code style?

> we decided it wasn't necessary as it didn't bring much benefit

How much do you work with JSX?

I guess we'll have to wait two more months until enough people have updated and this becomes one of the "great demand" cases again.

BTW thanks @wesbos for kicking this off 😂 (we all love your courses 🤙)

nextriot:
+1 Add this back in as an option please. More times than not in JSX, I'm importing a component that doesn't wrap children. I assume many others also write self-closing components that simply take props.

duailibe:
Sorry I took so long to respond, I was on vacation.

@meriadec 

> I'm disappointed to see this feature leaving Prettier because of a text editor setting problem for few people

Yes, the annoyance during development was what caused people to raise an issue but that's not the reason this was changed, only the fact the Prettier shouldn't change your code too much, only whitespace. That's the main argument, don't get hung up on the "text editor setting problem" argument.

> Seems that once something is merged in codebase it is hard to discuss it

That's true and not true. It's not about "when something is merged" but a philosophy: we're opinionated and we won't add options. So once we think something was worth a change, we'll only change it back if someone comes up with a very good reason for it (probably something we missed when we made the decision in the 1st place).

And not true because we have churned features in the past after we figured out we made mistakes.

duailibe:
@simonhaenisch 

> How much do you work with JSX?

8h/day, 5days/week, more or less.

yyynnn:
Bring option back pls

ralmo:
Found this thread after starting a search because I was unable to find a corresponding option (which I pretty much took for granted) in the settings.

I understand, that this behavior might be annoying for some people but that's why it should be configurable. Personally, I loved it.

PR #6243, opened 2019-06-21
fix(glimmer): Add newline at the end of files

dcyriller:
This PR ensures a trailing newline is added in hbs files.

```hbs
// Input
Hello

// Output (Prettier stable)
Hello
// Output (Prettier master)
Hello

```


dcyriller:
Just to be clear: why this change? Because Prettier formats every other language this way.

Turbo87:
@dcyriller this goes against https://github.com/ember-cli/ember-new-output/blob/master/.editorconfig#L17

The reason for not having a newline at the end is that it will create an extra DOM text node which in most cases would be unnecessary.

dcyriller:
I've hoped for a moment that https://github.com/josephfrazier/editorconfig-to-prettier would have supported the editorconfig option: `insert_final_newline`. It would have allowed me to append / remove a final newline depending on this option. But the package doesn't have the option (on purpose I guess), so I can't make the final newline configurable through `.editorconfig` file.

An alternative would be to submit an RFC in ember community to actually add a newline at the end of hbs files. We could strip trailing whitespace in template compilation as suggested here: https://github.com/ember-cli/ember-cli/pull/3440.

Issue #6245, opened 2019-06-23
Introducing Plugin for Pug


Shinigami92:
referencing issues #5767 and #4686

@ispyhumanfly, @j-f1, @rdhar, @danielpza, @lydell
Hello at all 😃 
I'm happy to introduce the plugin for prettier to support pug! 🎉 
https://github.com/Shinigami92/prettier-plugin-pug

I'm currently in heavily development phase and I am not completely finished with the base yet.
But I would like to let you know that I would like to publish a first alpha in the next few days.
In addition, I am currently developing so that no options can be passed. I'll do that later if the base is _stable_.

If you want to test something, you're welcome to help me.

rdhar:
Brilliant, that's great to hear! 🎉

Do keep us in the loop with how you're getting along and we'd be more than happy to test/help out wherever possible.

And a huge thanks for taking on this challenge in the first place!

evilebottnawi:
Feel free to send a PR in docs :+1:

Shinigami92:
I'm happy to announce the first alpha :D

Please create issues for any bugs

https://github.com/Shinigami92/prettier-plugin-pug/releases/tag/1.0.0-alpha.1

https://www.npmjs.com/package/prettier-plugin-pug

Shinigami92:
I have published an [alpha.2](https://www.npmjs.com/package/prettier-plugin-pug/v/1.0.0-alpha.2)
I used `npm publish --tag aplha` and I think the `--tag` was wrong, so it's not seen as the _latest_ version 🤔 
Nevermind, [Changelog](https://github.com/Shinigami92/prettier-plugin-pug/blob/master/CHANGELOG.md#100-alpha2)

Thanks to @rdhar, I used your [repo](https://github.com/rdhar/template-pug) for some internal testing 😄

rdhar:
It's great to hear that some use came of that repo! I'd meant to keep it updated as a standard baseline for my smaller web projects in the future. But, would you believe it, due to an issue with [Parcel's CSS-loader](https://github.com/parcel-bundler/parcel/issues/2936), template-pug was left behind and forgotten(!)

Will be sure to pick this up shortly. Is there any additional steps on getting up and running with it on VS Code?

Shinigami92:
Yes, follow this PR: https://github.com/prettier/prettier-vscode/pull/757

lipis:
As already mentioned.. please add it to the this list: https://prettier.io/docs/en/plugins.html#community-plugins

[...]

PR #6246, opened 2019-06-24
Add Pug plugin to the list of plugins

Shinigami92:
ref #6245

I will release the plugin in an alpha state around next weekend?!
Then it can be found at https://www.npmjs.com/package/prettier-plugin-pug


j-f1:
Thank you for contributing to the Prettier ecosystem!

Shinigami92:
I think we have to run `yarn build-docs` 🤔
Or is there anything left to do to see the result on the page?

PR #6249, opened 2019-06-25
[Glimmer/Handlebars] Forces multilines print when almost at max length

jjaffeux:
Hi,

Before:

```
{{my-component  foo="bar"  bar="baz" action=(action "almostTheMaximumLengthxxxx")
}}
```

After:

```
{{my-component
  foo="bar"
  bar="baz"
  action=(action "almostTheMaximumLengthxxxx")
}}
```

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -174,7 +174,7 @@ function print(path, options, print) {
       return group(
         concat([
           n.escaped === false ? "{{{" : "{{",
-          printPathParams(path, print),
+          printPathParams(path, print, { group: false }),
           isConcat ? "" : softline,
           n.escaped === false ? "}}}" : "}}"
         ])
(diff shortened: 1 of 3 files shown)


jjaffeux:
I think:

```
{{my-component
  foo="bar"
  bar="baz"
  action=(action "almostTheMaximumLengthxxxx")}}
```

is superior to:

```
{{my-component
  foo="bar"
  bar="baz"
  action=(action "almostTheMaximumLengthxxxx")
}}
```

The spec is currently failing because of this, do you want me to try to fix the spec or force the previous the second output in every cases?

This `}}` is always taking a line for almost no value.

j-f1:
We do it the latter way in JS, but if Handlebars has a different convention, then that’s fine too. Mostly the rationale is that by putting the closing delimiter on its own line, the parameters can be safely moved around/out of the block.

evilebottnawi:
hm, maybe handlebars have basic/popular style examples to avoid problems with community

lougreenwood:
I prefer the latter way too. I find it easier to visually scan a doc when the tag opens and closes at the same indentation

jjaffeux:
@evilebottnawi I have changed my code to follow the preferred formatting, that should be all good now!

jjaffeux:
Anything I can do to help this move forward ?

duailibe:
Also, @GavinJoyce, will Glimmer PRs affect your work?

GavinJoyce:
> Also, @GavinJoyce, will Glimmer PRs affect your work?

I don't believe so, anything that doesn't break existing tests seems fine.

duailibe:
@GavinJoyce In that case let's just hope we don't have duplicate fixes :-)

jjaffeux:
@duailibe @evilebottnawi I tried to make the code simpler, thanks for feedback.

PR #6251, opened 2019-06-26
Swap bases in changelog

Shinigami92:
😸


Shinigami92:
In my opinion, we can safely merge this because it does not touch any code

PR #6293, opened 2019-07-11
Add JSON script types

ascorbic:
Prettier currently recognises `<script type="application/ld+json">` as JSON. This PR adds support for other `application/foo+json` types, as well as plain `application/json`. It also adds support for [import maps](https://github.com/WICG/import-maps).

### DIFF
--- a/src/language-html/utils.js
+++ b/src/language-html/utils.js
@@ -377,8 +377,8 @@ function inferScriptParser(node) {
     }
 
     if (
-      node.attrMap.type.substr(-4) === "json" ||
-      node.attrMap.type.substr(-9) === "importmap"
+      node.attrMap.type.endsWith("json") ||
+      node.attrMap.type.endsWith("importmap")
     ) {
       return "json";
     }
--- a/src/language-html/utils.js
+++ b/src/language-html/utils.js
@@ -376,7 +376,10 @@ function inferScriptParser(node) {
       return "markdown";
     }
 
-    if (node.attrMap.type === "application/ld+json") {
+    if (
+      node.attrMap.type.substr(-4) === "json" ||
+      node.attrMap.type.substr(-9) === "importmap"
+    ) {
       return "json";
     }
   }
--- a/tests/html_script/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/html_script/__snapshots__/jsfmt.spec.js.snap
@@ -7,12 +7,36 @@ printWidth: 80
                                                                                 | printWidth
 =====================================input======================================
 <script type="application/ld+json">
+  {   "json": true }
+</script>
+<script type="application/json">
   {   "json":true  }
+</script>
+<script type="importmap">
+  {   "json":true  }
+</script>
+<script type="systemjs-importmap">
+  {   "json":true  }
+</script>
+<script type="invalid">
+  {   "json":false  }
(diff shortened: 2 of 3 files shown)


src/language-html/utils.js
```diff
@@ -376,7 +376,10 @@ function inferScriptParser(node) {
       return "markdown";
     }
 
-    if (node.attrMap.type === "application/ld+json") {
+    if (
+      /^([a-z]+-)?importmap$/.test(node.attrMap.type) ||
+      /^application\/([a-z]+\+)?json$/.test(node.attrMap.type)
```
evilebottnawi: I think better use list of possible types than using regexp, just my opinion

ascorbic: The reason I didn't go with a list of types is so that it can support arbitrary "application/__foo__+json" and "__foo__-importmap" types. It's on a rarely-used path, so shouldn't cause significant performance issues.

ascorbic: I was curious whether substr would be faster. On my machine the regexp is faster in Firefox, but substr is faster in Chrome. https://jsperf.com/mime-matching

j-f1: The fever should also take into account hyphens since they’re used by e.g. GitHub.

ascorbic: @j-f1 Where would the hyphens be?

j-f1: Here’s [an example](https://developer.github.com/v3/reactions/) in the wild: `application/vnd.github.squirrel-girl-preview+json`

ascorbic:
I've simplified the test. It now just checks if the type ends with "importmap" or "json". I reckon was over-thinking it. False positives aren't really an issue here. There aren't loads of `<script type`s in the wild where we'll hit one that ends with "json" but isn't json. I've used substr because `endsWith` [isn't available](https://node.green/#ES2015-built-ins-well-known-symbols-Symbol-match--String-prototype-endsWith) in Node 6.4.

src/language-html/utils.js
```diff
@@ -376,7 +376,10 @@ function inferScriptParser(node) {
       return "markdown";
     }
 
-    if (node.attrMap.type === "application/ld+json") {
+    if (
+      node.attrMap.type.substr(-4) === "json" ||
```
lydell:
```suggestion
      node.attrMap.type.endsWith("json") ||
```

src/language-html/utils.js
```diff
@@ -376,7 +376,10 @@ function inferScriptParser(node) {
...
+      node.attrMap.type.substr(-9) === "importmap"
```
lydell:
```suggestion
      node.attrMap.type.endsWith("importmap")
```

src/language-html/utils.js
ascorbic: As I said in my comment above, I didn't use `endsWith` because it's not available in Node 6.4.

lydell: Am I missing something? :thinking: 

```
❯ node --version
v6.4.0

❯ node
> "".endsWith
[Function: endsWith]
> 
```

ascorbic: Looks like I misread the node.green table. I could've sworn it said it was unsupported! I used `endsWith` initially but changed it after checking.

PR #6313, opened 2019-07-18
Provide internal TypeScript support to improve code quality

Shinigami92:
### Add support for TypeScript

**This PR will NOT CHANGE any runtime code!**

We will follow this tutorial:
https://devblogs.microsoft.com/typescript/how-to-upgrade-to-typescript-without-anybody-noticing-part-1/
https://devblogs.microsoft.com/typescript/how-to-upgrade-to-typescript-without-anybody-noticing-part-2/

This will provide TypeScript support for projects like plugins that uses prettier as a dependency

ref #6286

---


scripts/run-external-tests.js
```diff
@@ -20,6 +20,10 @@ function tryFormat(file) {
   return null;
 }
 
+/**
+ *
```
j-f1: Can you delete this line?

scripts/sync-flow-tests.js
```diff
@@ -28,6 +28,10 @@ function tryParse(file, content) {
   return null;
 }
 
+/**
+ *
```
j-f1: Same here.

scripts/run-external-tests.js
Shinigami92: You mean it should be
```js
/**
 * @param {string | string[]} patterns
 */
```

or

```js
/** @param {string | string[]} patterns */
```

How do you want it?

Shinigami92:
There is no @types/string-width 3.0.0 only 4.0.1
This doesn't match and therefore can currently mot be used

scripts/run-external-tests.js
j-f1:
```ts
/**
 * @param {string | string[]} patterns
 */
```

kachkaev:
> This will provide TypeScript support for projects like plugins that uses prettier as a dependency

Not sure I understand how. If a project uses a non-CLI version of Prettier, they would need type definitions for Prettier API. These definitions are currently available at [`@types/prettier`](https://www.npmjs.com/package/@types/prettier) ([source](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/prettier)).

Shinigami92:
> Not sure I understand how. If a project uses a non-CLI version of Prettier, they would need type definitions for Prettier API. These definitions are currently available at [`@types/prettier`](https://www.npmjs.com/package/@types/prettier) ([source](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/prettier)).

That's the reason: there are only outdated types on DefinitelyTyped.
This will also provide better development experience in prettier itself and always will provide the newest and not outdated typedefs

Shinigami92:
There are no @types/chalk for 2.1.0

Shinigami92:
There are no types for `n-readlines` at all

src/common/fast-path.js
```diff
@@ -111,6 +142,9 @@ FastPath.prototype.each = function each(callback /*, name1, name2, ... */) {
 // Similar to FastPath.prototype.each, except that the results of the
 // callback function invocations are stored in an array and returned at
 // the end of the iteration.
+/**
+ * @param {(path: FastPath, index: number) => any} callback
```
Shinigami92: I'm currently learning JSDoc and don't know how to type this in the following way:

```ts
map<U>(callback: (path: this, index: number) => U, ...names: PropertyKey[]): U[];
// ~~~                  ~~~~                    ~                            ~
```

src/common/fast-path.js
```diff
@@ -23,16 +33,29 @@ FastPath.prototype.getName = function getName() {
...
+ */
 FastPath.prototype.getValue = function getValue() {
   const s = this.stack;
   return s[s.length - 1];
 };
 
+/**
+ * @param {FastPath} path
+ * @param {number} count
+ * @returns {string | null}
```
Shinigami92: Because this is only a function, I loose the scope of `@template {string} T`

Shinigami92:
How can we provide `.d.ts` files in `dist` folder?
So we can make `@types/prettier` obsolete

duailibe:
> This will provide TypeScript support for projects like plugins that uses prettier as a dependency

That's not the point. The point of #6286 is to improve code quality and onboarding of new contributors.

Shinigami92:
@duailibe OK, so this PR will only be responsible for improving the quality within prettier
Not to distribute TypeScript definitions to the outside

Did I understand you correctly?

duailibe:
@Shinigami92 Yes

Shinigami92:
OK
Because the majority is now done

1. creating a working `tsconfig.json`
2. adding the `devDependencies` `@types/...`

I think that can now be merged safely

I can provide more and more JSDocs in the future, but other or internal developers can improve the JSDocs as well

Shinigami92:
Oh how cool is that! https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37003 was also just merged 🎉

src/common/fast-path.js
j-f1: I think there’s an `@template U` or `@template {U}` tag.

Shinigami92: Yes, but I don't know how to apply it here
I have tested some combinations

thorn0:
@Shinigami92 

> and always will provide the newest and not outdated typedefs

> How can we provide .d.ts files in dist folder?

TS doesn't support this yet. See https://github.com/microsoft/TypeScript/issues/7546

src/common/fast-path.js
```diff
@@ -2,13 +2,23 @@
 
 const assert = require("assert");
 
+/**
+ * @class
+ * @constructor
+ * @template {string} T
+ * @property {T[]} stack
```
thorn0: Why make this class generic? Isn't `value` always a string?

Shinigami92: I used this as a start base:

https://github.com/ffflorian/DefinitelyTyped/blob/14f6f073a73f28fcdee4d899b50254eac30d4aac/types/prettier/index.d.ts#L13

thorn0: Oh, so `T` is an AST node, not a string. `@template {string} T` means `<T exends string>`.

Shinigami92: Yes

thorn0: So, this constraint is incorrect and also the type for `stack` is incorrect.

Shinigami92: You think the definition of `DefinitelyTyped` is wrong or `string` is wrong?
Can you be more precisely, what is wrong?

thorn0: Those definitions look fine. But if you are simply copying the types from there, the parts `@template {string} T` and `@property {T[]} stack` don't correspond to the source.

Shinigami92: So should it be `string` everywhere instead of `T`?

thorn0: You should either 1) read and understand the code and add types based on your analysis, or 2) at least copy the types from those definitions precisely without adding random bits. Sorry if i'm harsh, but adding types without understanding the code is a really questionable endeavor.

Shinigami92: 2) But I cant just copy those definitions because there are not JSDoc compatible :(

So I tried using some of both worlds
But it seams that this causes problems, so I will rework this and only use `string` at this point

(Only need to be at home for this)

thorn0: "and only use string at this point" - that's what I called 'random bits', unfortunately.

If you have read the code, what type are the elements of the `stack` array?

Shinigami92: I think it should be something like `Array<AstNode>`
_Whatever AstNode is_

thorn0: How did you come to such a conclusion?

Shinigami92: Here is one example:

https://github.com/prettier/prettier/blob/master/src/common/util.js#L621-L628

thorn0: Here is a [counterexample](https://github.com/prettier/prettier/blob/master/src/common/fast-path.js#L10-L16). And there are comments in the code of the class that describe the structure of the stack.

Also you seem to have not read the [docs](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html) on JSDoc support in TS. Otherwise you wouldn't use the `@property` tag the way you did. It works only with [typedefs](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#typedef-callback-and-param). "Properties are inferred from assignments in class bodies" as [it is said in the very beginning](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#properties-are-inferred-from-assignments-in-class-bodies) of that doc page. To override the inference, you can use `@type`. Also there is no need to add both `@class` and `@constructor`, they are synonyms... I could go on, but I don't really have time for that.

I've got mixed feelings about this PR. On the one hand, it might become a good enough starting point for a subsequent gradual migration to TS. On the other, a migration like that requires a deep understanding of both TS and the code base. Sloppy types will quickly become a nuisance like it happened with Redux, whose maintainers now [regret](https://twitter.com/acemarke/status/1149795712206594048) they have messed with TS without having a good expertise in it.

Shinigami92: I don't know how old and experienced you are, but I'm only 26 years old and have 3 years of JavaScript experience (2 of these years in TypeScript)
All my experiences are based on es6+

So instead of telling me all the time that I'm doing it wrong, you could be nice and help me get it right
That would provide a much more enjoyable community experience

That said,
sure, I can set it to `string`, but I'm confused by the `DefinitelyTyped` definitions

And also yes, I know TypeScript and I'm not very into JSDoc, because I don't need it my self (for types) because I'm working with `.ts`-native files in all my projects. So I have approximately 10 days experience in JSDoc

Shinigami92:
Now I'm a bit confused about `i in value`, where value is of type `string`

https://github.com/prettier/prettier/blob/22a633c654b5975e9572108af34b876c24a38a21/src/common/fast-path.js#L128

https://github.com/prettier/prettier/blob/22a633c654b5975e9572108af34b876c24a38a21/src/common/fast-path.js#L161

Error is:
```
The right-hand side of an 'in' expression must be of type 'any', an object type or a type parameter. ts(2361)
```

But I don't want to touch any code in this branch

brodybits:
What is the status of this proposal?

My one minor concern is that I find the title a little bit misleading. This proposal seems to be using the TypeScript tool to do type checking of JavaScript code with JSDoc comments, which I think is the right thing to do.

/cc @rbiggs

Shinigami92:
@brodybits Thank you for your interest
After I messaged @thorn0 my last message (https://github.com/prettier/prettier/pull/6313#discussion_r307395922) I didn't get any new response
I also think that we did come  a bit into a conflict here and then I stopped working on this.
But the current state of this PR is fully mergable (maybe I could revert any code-document changes or they can cherry-pick some of my changes)

From there (after it is merged) everyone can contribute and help to improve the documentation

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "noEmit": true,
+    "target": "esnext",
```
brodybits: I would personally favor target: "es5" to specify that we expect to generate ES5 code (using a JavaScript transpiler).

brodybits:
Any chance we can continue with this proposal, with the evidently controversial fast-path.js update separated into its own PR?

tsconfig.json
Shinigami92: Sure 
Just let me wakeup ^^ I will change it in few hours

brodybits: Thanks for the super-fast response, no rush on my part. Here is the suggested change in case you want to use it:

```suggestion
    "target": "es5",
```

Shinigami92:
> When I tried doing `npx tsc` to check the files, [...]
> 
> I started working on some of the errors in `src/language-js/printer-estree.js`, can raise a PR if anyone is interested.

Yes, all new work (Code-Changes / JSDoc-ing) should be in another dedicated PR

Currently the `@types` and the `tsconfig` is helping developers in IDEs like VSCode
It is currently not in scope to build it with tsc
IMO, this can be done later in another PR

----

Also looking forward into https://github.com/Microsoft/TypeScript/wiki/Roadmap#37-november-2019
TS 3.7 -> `Allow allowJs and declaration to be used together`
https://github.com/microsoft/TypeScript/pull/32372

Shinigami92:
> Should I raise a separate PR for tsconfig.json to support checkJs?

YES, **AND** I will try my best to help you 🚀

brodybits:
> YES, **AND** I will try my best to help you 🚀

PR with tsconfig.json, with checkJs enabled and working, is now on the way.

thorn0:
@Shinigami92 Unlike the type checking with tsc, "helping developers in IDEs like VSCode" is not something verifiable. We can't write tests to check that it doesn't break when the code is updated. That's why we can't use it as a starting point. On the other hand, when we get type checking, we get the IDE support automatically.

brodybits:
> So it seems #6705 is going to supersede this PR, right?

I will let @Shinigami92 answer that one. I am definitely with @thorn0 to test the changes on the command line (and on the build server, which is still a TODO item in #6705) before we make any statements about IDE behavior, yes.

The one thing about #6705 is that it is blocked by over 100 type errors that I think we should solve if possible. I have already raised a couple PRs to start solving a few of them, still many to go. I could also really use some help with this part. While it would be possible to just use `// @ts-ignore` comments, I would like to avoid it is possible.

I also saw some really nice updates and fixes that we should not throw away. I would be happy to raise separate PRs to get them in.

I will need to break soon, will come back in 6-12 hours or so. Thanks!

Shinigami92:
I dont know what `@glimmer` is
In my PR there are `@types/*` packages that provides typescript definitions for used packages/dependencies

Also in @brodybits PR, `strict` is not enabled (we could enable it later)

Also some `exclude`s are setup a bit different 🤷‍♂️

---

In the end, it does not matter to me which PR is merged or not
For me, it's important that the support for TypeScript is improved in the long term, and the definitions for plugins and so forth are more native and much better than what is in `@types/prettier`
And it would be nice if problems were found and fixed by using type checking and the quality of the code is improved

brodybits:
I will respond later today. We may want to discuss the approach in a new issue.

brodybits:
At this point, I would really favor resolving most of the JSDoc type issues that I reported in #6703 before we consider adding tsconfig.json. I have already started raising PRs to resolve some of these issues, looks like it will take quite a bit of work to get through the rest. (I could use some help if anyone has the time for it. Reviews and other feedback would also be really helpful.)

I have also raised PR #6711 to capture the "@types" packages and JSDoc updates that were originally done in this PR. I would appreciate some feedback if we want to consider integrating any of those changes or not.

thorn0:
superseded by #6770

brodybits:
> superseded by #6770

with some other enhancements captured in PR #6711

PR #6317, opened 2019-07-19
Use Array spread (convert es2015)

Shinigami92:
Convert to es2015 using array spread


kachkaev:
I guess this creates a [problem](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUAjAhgJwDSIB0xuA5gM4CUiA3gFCKIQLlwA2ApoW3KWmVQDcdAL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.4.5&externalPlugins=) mentioned by @duailibe in https://github.com/prettier/prettier/pull/6225#issuecomment-502392011

Shinigami92:
Ok, why is this set to es6?
https://github.com/prettier/prettier/blob/9c8d920c173534e7c65111f9310f763f80de55a3/.eslintrc.yml#L3

Are these just parts of the code or is it not allowed at all?
Should we move some files so we can create a sub-eslintrc file that denies `es2015+`?

Is there any chance to modernize the code?

Shinigami92:
Oh and I did not touched `FastPath` because I tried this here: https://github.com/prettier/prettier/pull/6316

brodybits:
> I guess this creates a [problem](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUAjAhgJwDSIB0xuA5gM4CUiA3gFCKIQLlwA2ApoW3KWmVQDcdAL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.4.5&externalPlugins=) mentioned by @duailibe in [#6225 (comment)](https://github.com/prettier/prettier/pull/6225#issuecomment-502392011)

Looks like that was some kind of a “performance regression” in a “hot path”. I am putting this info here for the sake of clarity.

> Ok, why is this set to es6?

I think that is needed to tell eslint that there is at least *some* es6-styled code in this project.

> […]
> Should we move some files so we can create a sub-eslintrc file that denies `es2015+`?

I would favor a different kind of solution that somehow disables some ES6 features such as array spreads if they can cause potential performance issues. A kind of opposite to the eslint [`prefer-spread`](https://eslint.org/docs/rules/prefer-spread) rule (not sure if one may exist or not).

> Is there any chance to modernize the code?

I would definitely favor modernizing the code in general, but with some discretion. As @duailibe said in <https://github.com/prettier/prettier/pull/6225#issuecomment-502392011>, using the Array spread and letting Babel transpile it seems to be not so good for performance. I am also *personally* biased against using ES6 classes unless absolutely necessary.

As a side point, I think the ultimate solution would be to drop support for old Node.js versions and use JavaScript that works on recent Node.js versions with minimal transpiling, if any. This should help avoid hidden performance surprises, increase debuggability, etc.

Shinigami92:
Currently, node 6 is the least supported, if I understand the following correctly:

https://github.com/prettier/prettier/blob/9c8d920c173534e7c65111f9310f763f80de55a3/package.json#L14

If I'm right, node 6 supports `es2015` completely https://node.green
So maybe we should only transpile to `es2015` (`es6`) instead of `es5`

Shinigami92:
Currently, node 6 is the least supported, if I understand the following correctly:

https://github.com/prettier/prettier/blob/9c8d920c173534e7c65111f9310f763f80de55a3/package.json#L14

If I'm right, node 6 supports `es2016` well enough https://node.green/#ES2016
So maybe we should **try** to transpile to `es2016` instead of `es2015`

That should also produce less code after the build process

lydell:
We even support Node.js 4

thorn0:
Thanks for contributing!

PR #6327, opened 2019-07-23
Handle missing cache directory for changedCache

g-harel:
Small patch to check that the cache directory was found before initializing the `changedCache`.

https://github.com/prettier/prettier/pull/5910#discussion_r305870825

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -446,13 +446,17 @@ function formatFiles(context) {
   let changedCache = null;
   if (context.argv["only-changed"]) {
     const cacheDir = findCacheDir({ name: "prettier", create: true });
-    changedCache = new ChangedCache({
-      location: path.join(cacheDir, "changed"),
-      readFile: fs.readFileSync,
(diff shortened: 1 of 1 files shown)


src/cli/util.js
```diff
@@ -446,13 +446,17 @@ function formatFiles(context) {
   let changedCache = null;
   if (context.argv["only-changed"]) {
     const cacheDir = findCacheDir({ name: "prettier", create: true });
```
evilebottnawi: Better add `const cacheDir = findCacheDir({ name: "prettier", create: true }) || os.tmp();` here otherwise cache doesn't work

g-harel: If I'm reading this correctly, this would write the cache to a unique temporary directory every time it runs. This would defeat the purpose of the writing the cache since it will not be ready the next time it runs.

evilebottnawi: @g-harel no, if we can't find cache dir, `/tmp/.cache/prettier` will be used and in next run cache was read from this location again, so all works find

PR #6333, opened 2019-07-27
Link to `fast-glob` instead of `glob` in docs

shroudedcode:
While debugging a glob-related issue (caused by me not quoting the glob string as recommended in the docs) I noticed a small mistake on [the **CLI** page in the docs](https://prettier.io/docs/en/cli.html) which states that:

> The [glob syntax from the glob module](https://github.com/isaacs/node-glob/blob/master/README.md#glob-primer) is used.

After taking a look at the source code I found out that's not actually true and [the `globby` module](https://github.com/sindresorhus/globby) is used instead. As a result, [the syntax from the `fast-glob` module](https://github.com/mrmlnc/fast-glob/blob/master/README.md#pattern-syntax) (on which `globby` depends) is used.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -17,7 +17,7 @@ In practice, this may look something like:
 prettier --single-quote --trailing-comma es5 --write "{app,__{tests,mocks}__}/**/*.js"
 ```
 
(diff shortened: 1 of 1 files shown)


lipis:
Thanks 🍺 🍻

PR #6348, opened 2019-08-01
Plugins can have falsy defaultOptions

jablko:
Otherwise `export const defaultOptions = { bracketSpacing: false };` doesn't work.


jablko:
> Can we add tests for this case?

Done: I've now added `bracketSpacing: false` to `tests_integration/plugins/defaultOptions/plugin.js`.

jablko:
> Please remove `.config.json.swp` from PR

Done!

PR #6366, opened 2019-08-08
Update webpack to v4

fisker:
Update webpack to v4


scripts/build/bundler.js
```diff
@@ -176,13 +176,14 @@ function getWebpackConfig(bundle) {
...
+      globalObject: 'new Function("return this")()'
     },
-    plugins: [
-      new webpack.DefinePlugin({
-        "process.env.NODE_ENV": JSON.stringify("production")
-      })
-    ]
+    optimization: {
+      // disable terser for `parser-postcss.js`
+      minimize: false
```
evilebottnawi: Why we disable this?

fisker: this module can't be minified, unknown reason, we can figure later

https://github.com/prettier/prettier/pull/6200#issuecomment-500018106

fisker: may be I should add a comment to this link or a TODO

evilebottnawi: @fisker yes, please leave comment in source code `// Todo description`

fisker:
I just resolve the conflicts, anything else need to be done?

lipis:
Is this good to go?

evilebottnawi:
I think all good and we can safely merge

evilebottnawi:
hm, minimizer: false output big js file, so it can slow site

fisker:
@lipis @evilebottnawi this minimizer: false only effect `parser-postcss.js` (it's not also minified before this PR), see comment above. and I already made a seperate PR to fix it, https://github.com/prettier/prettier/pull/6395 so  it's good to go

PR #6367, opened 2019-08-08
Update linguist-languages to v7

fisker:
Update linguist-languages to v7

### DIFF
--- a/src/language-js/index.js
+++ b/src/language-js/index.js
@@ -6,64 +6,70 @@ const options = require("./options");
 const createLanguage = require("../utils/create-language");
 
 const languages = [
-  createLanguage(require("linguist-languages/data/javascript"), {
+  createLanguage(require("linguist-languages/data/JavaScript"), {
     override: {
       since: "0.0.0",
       parsers: ["babel", "flow"],
-      vscodeLanguageIds: ["javascript"]
-    },
-    extend: {
-      interpreters: ["nodejs"]
+      vscodeLanguageIds: ["javascript"],
+      interpreters: ["node", "nodejs"]
     }
   }),
-  createLanguage(require("linguist-languages/data/javascript"), {
+  createLanguage(require("linguist-languages/data/JavaScript"), {
     override: {
       name: "Flow",
       since: "0.0.0",
       parsers: ["babel", "flow"],
       vscodeLanguageIds: ["javascript"],
-
       aliases: [],
       filenames: [],
-      extensions: [".js.flow"]
+      extensions: [".js.flow"],
+      interpreters: ["node"]
     }
   }),
-  createLanguage(require("linguist-languages/data/jsx"), {
+  createLanguage(require("linguist-languages/data/JSX"), {
     override: {
       since: "0.0.0",
       parsers: ["babel", "flow"],
       vscodeLanguageIds: ["javascriptreact"]
     }
   }),
-  createLanguage(require("linguist-languages/data/typescript"), {
+  createLanguage(require("linguist-languages/data/TypeScript"), {
     override: {
       since: "1.4.0",
       parsers: ["typescript"],
-      vscodeLanguageIds: ["typescript", "typescriptreact"]
+      vscodeLanguageIds: ["typescript"]
     }
   }),
-  createLanguage(require("linguist-languages/data/json"), {
+  createLanguage(require("linguist-languages/data/TSX"), {
+    override: {
+      since: "1.4.0",
+      parsers: ["typescript"],
+      vscodeLanguageIds: ["typescriptreact"]
+    }
+  }),
+  createLanguage(require("linguist-languages/data/JSON"), {
     override: {
       name: "JSON.stringify",
       since: "1.13.0",
       parsers: ["json-stringify"],
       vscodeLanguageIds: ["json"],
-
+      group: "JavaScript",
       extensions: [], // .json file defaults to json instead of json-stringify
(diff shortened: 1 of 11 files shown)


evilebottnawi:
Here issue about this https://github.com/prettier/prettier/pull/6362#issuecomment-519345148

fisker:
> Here issue about this [#6362 (comment)](https://github.com/prettier/prettier/pull/6362#issuecomment-519345148)

I have checked that, did you mean `exclude` for markdown? I think everything else is covered

evilebottnawi:
@fisker yes, and please do not send similar PRs in future, it is not good, because now we have 2 similar PRs, and whose work is done in vain (excluding abandoned PRs)

fisker:
Should I close this one, and you can continue work on that?

evilebottnawi:
@fisker it doesn't make sense anymore because you fix some notes from review, so you can continue to work, just try to avoid this in the future

src/language-js/index.js
```diff
@@ -6,64 +6,70 @@ const options = require("./options");
...
     }
   }),
-  createLanguage(require("linguist-languages/data/json"), {
+  createLanguage(require("linguist-languages/data/JSON"), {
     override: {
       since: "1.5.0",
       parsers: ["json"],
-      vscodeLanguageIds: ["json"]
+      vscodeLanguageIds: ["json"],
+      group: "JavaScript"
```
evilebottnawi: `group: "JavaScript"` is unnecessary now

src/language-js/index.js
```diff
@@ -6,64 +6,70 @@ const options = require("./options");
...
+    }
+  }),
+  createLanguage(require("linguist-languages/data/JSON"), {
     override: {
       name: "JSON.stringify",
       since: "1.13.0",
       parsers: ["json-stringify"],
       vscodeLanguageIds: ["json"],
-
+      group: "JavaScript",
```
evilebottnawi: Unnecessary

fisker:
OK, I will add exclude now

src/language-js/index.js
```diff
@@ -6,64 +6,70 @@ const options = require("./options");
...
       name: "Flow",
       since: "0.0.0",
       parsers: ["babel", "flow"],
       vscodeLanguageIds: ["javascript"],
-
       aliases: [],
       filenames: [],
-      extensions: [".js.flow"]
+      extensions: [".js.flow"],
+      interpreters: ["node"]
```
evilebottnawi: Don't think `node` can interpreted flow language

src/language-js/index.js
```diff
@@ -6,64 +6,70 @@ const options = require("./options");
...
+  createLanguage(require("linguist-languages/data/JavaScript"), {
     override: {
       since: "0.0.0",
       parsers: ["babel", "flow"],
-      vscodeLanguageIds: ["javascript"]
-    },
-    extend: {
-      interpreters: ["nodejs"]
+      vscodeLanguageIds: ["javascript"],
+      interpreters: ["node", "nodejs"]
```
evilebottnawi: Why?, original source contains all `interpreters`

fisker: I was trying to make the result close as original as possible

evilebottnawi: @fisker need to do it with understanding :smile:

src/language-js/index.js
fisker: not sure how this works, the original result has this, what is you suggestion? exclude `node` ?

evilebottnawi: @fisker looks in original `linguist-languages` data and try to not lose `filenames` and `extensions`, other fields should be get from `linguist-languages`

fisker: linguist-languages data for flow has `node` for interpreters, keep it?

evilebottnawi: if `linguist-languages` has this field, don't put it in our code because it is already expanding

src/language-js/index.js
fisker: removed

src/language-js/index.js
fisker: removed

src/language-js/index.js
fisker: this field is removed

src/language-markdown/index.js
```diff
@@ -9,21 +9,13 @@ const languages = [
...
-        ".mkdown",
-        ".ronn",
-        ".workbook"
-      ]
+      vscodeLanguageIds: ["markdown"]
     },
     extend: {
       filenames: ["README"]
+    },
+    exclude: {
```
evilebottnawi: `createLanguage` is not support `exclude`, we need implement this, please look in code, before adding somethings

src/language-js/index.js
```diff
@@ -6,54 +6,56 @@ const options = require("./options");
...
 const languages = [
-  createLanguage(require("linguist-languages/data/javascript"), {
+  createLanguage(require("linguist-languages/data/JavaScript"), {
     override: {
       since: "0.0.0",
       parsers: ["babel", "flow"],
       vscodeLanguageIds: ["javascript"]
-    },
-    extend: {
-      interpreters: ["nodejs"]
```
ikatyang: Removing it would be a breaking change.

fisker: restored

fisker:
@lipis can this one get merged in 1.19?

PR #6370, opened 2019-08-08
test on old node.js

fisker:
test on old node.js

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,11 +38,19 @@ jobs:
       - template: .azure-pipelines/jobs/dev-test.yml
 
   - job: Dev_Test_Linux
-    displayName: Dev Test on Linux Node v10
+    displayName: Dev Test on Linux Node.js
     pool:
       vmImage: "Ubuntu 16.04"
-    variables:
-      node_version: 10
+    strategy:
+      matrix:
+        Node_v6:
+          node_version: 6
+        Node_v8:
+          node_version: 8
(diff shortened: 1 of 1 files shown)


brodybits:
I think you need to do `yarn install --ignore-engines` for the dev test to succeed on Node.js 6, like I did in d2a3ad309168e58c1d4426c73d8e7660e6498bdc.

It also looks like this proposal was based on some changes that I had proposed in PR #6232. I would appreciate credit in a `Co-authored-by` comment in the git commit message (see [[1]](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors)).

[1] <https://help.github.com/en/articles/creating-a-commit-with-multiple-authors>

azure-pipelines.yml
```diff
@@ -38,11 +38,19 @@ jobs:
       - template: .azure-pipelines/jobs/dev-test.yml
 
   - job: Dev_Test_Linux
-    displayName: Dev Test on Linux Node v10
+    displayName: Dev Test on Linux Node.js
```
brodybits:
```suggestion
    displayName: Dev Test on Linux
```

fisker: I know you did last time, I'm thinking another way to test, maybe later

fisker: we can't do `yarn install --ignore-engines`, this requires people install with this argument, no one wan't that

fisker:
will try on fork

PR #6373, opened 2019-08-09
Codebase install test

fisker:
This PR tests directly install from source code.

many thanks to @brodybits


fisker:
/cc @backus 

https://dev.azure.com/prettier/prettier/_build/results?buildId=1356

This test supposed to fail on install, any idea why it's not fail?

fisker:
maybe test failure is good enough?

brodybits:
It looks like your test with the new `get-stream` failed with the syntax error on Node.js 6, as would be expected on the source install, and older `get-stream` resolves the failure. Definitely looks nice. I cannot promise when I will get a chance to review these changes more carefully.

fisker:
It should fail on npm install, I will check the npm docs

brodybits:
From my experience, `npm install` does not seem to actually fail in case of the engines mismatch. I think `npm install` would just show a warning message, instead.

fisker:
you are right, just warn

fisker:
Good news `engine-strict` works

fisker:
Need help here, handlebars test fails for unknown reason

fisker:
I figured out why test fails, deps in installed dir version is not equal to the one in repo, this leave us some options:

1. skip test, only check install 

2. do less tests, such as only integration test

3. add `postinstall` script to make sure installed on using the same version as they are in repo

option 1 and 2 can't guarantee install from source code works
option 3 should be the best in theory, people install from source code should use same version of deps as they are in source code, when testing we can skip installing, use source code as fake installed package

/cc @prettier/core

lydell:
Hi! Can you explain to me with an example what kinds of errors we can catch with this setup?

fisker:
@lydell this PR run direct-install-from-codebase , should catch all errors that people run `npm install prettier/prettier`

evilebottnawi:
Not sure we need this test

fisker:
I've rebased this PR, run tests by `direct-install-from-codebase` with `npm`/`yarn`/`npm`, set schedule to run once a week on monday.

We support this feature, I think we should keep a eye on it.

@brodybits @evilebottnawi @lydell

lydell:
I think it might make sense to test this, yes. But I can’t help with reviewing, unfortunately.

Issue #6376, opened 2019-08-09
Handlebars: quotes option not respected on html attributes


dcyriller:
**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#.....)
```sh
# Options (if any):
--single-quote
```

**Input:**
```hbs
<div class="a-class-name"></div>
```

**Output:**
```hbs
<div class="a-class-name"></div>
```

**Expected output:**
```hbs
<div class='a-class-name'></div>
```

evilebottnawi:
Thanks for issue, duplicate https://github.com/prettier/prettier/issues/5828

dcyriller:
Actually not 😞. It is the same bug in two different printers.

It makes me sad that we're not able to preprocess hbs files with html.

evilebottnawi:
@dcyriller it should be not hard to implement, feel free to send a PR :+1:

dcyriller:
Sure, I have it almost ready :)

dcyriller:
It's ready there ^

PR #6377, opened 2019-08-09
fix(handlebars): Use quotes from config in AttrNode

dcyriller:
Closes https://github.com/prettier/prettier/issues/6376

### DIFF
--- a/tests/handlebars-attr-node/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/handlebars-attr-node/__snapshots__/jsfmt.spec.js.snap
@@ -6,11 +6,16 @@ parsers: ["glimmer"]
 printWidth: 80
                                                                                 | printWidth
 =====================================input======================================
+<div title='My title'></div>
+<div title="My title"></div>
 <div title='My "title"'></div>
 <div title="My other 'title'"></div>
 
 =====================================output=====================================
-<div title='My "title"'></div><div title="My other 'title'"></div>
+<div title="My title"></div>
+<div title="My title"></div>
+<div title='My "title"'></div>
+<div title="My other 'title'"></div>
 ================================================================================
 `;
 
@@ -21,10 +26,15 @@ printWidth: 80
 singleQuote: true
                                                                                 | printWidth
 =====================================input======================================
+<div title='My title'></div>
+<div title="My title"></div>
 <div title='My "title"'></div>
 <div title="My other 'title'"></div>
 
 =====================================output=====================================
-<div title='My "title"'></div><div title="My other 'title'"></div>
+<div title='My title'></div>
+<div title='My title'></div>
+<div title='My "title"'></div>
+<div title="My other 'title'"></div>
 ================================================================================
 `;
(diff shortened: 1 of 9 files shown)


dcyriller:
@duailibe it’s rebased now! (I was on vacations.)

dcyriller:
> /cc @dcyriller need rebase again, thanks!

done, let me add the test

dcyriller:
I trembled a bit, but it has worked out great: https://github.com/prettier/prettier/pull/6377/commits/8ac337d7a05a3a393bdb818cb1ca75c80d0c956b 😄

tests/handlebars-attr-node/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,30 @@
...
+parsers: ["glimmer"]
+printWidth: 80
+singleQuote: true
+                                                                                | printWidth
+=====================================input======================================
+<div title='My "title"'></div>
+<div title="My other 'title'"></div>
+
+=====================================output=====================================
+<div title='My "title"'></div><div title="My other 'title'"></div>
```
evilebottnawi: I think we need quotes here like do prettier for js

dcyriller: You mean, we should have for the second div?
```
<div title='My other \'title\''></div>
```

On master, hbs printer produces the following output (the two divs wrong, the first being very wrong):
```
<div title="My "title""></div><div title="My other 'title'"></div>
```

evilebottnawi: hm, all fine, my mistake

tests/handlebars-basics/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -51,12 +51,12 @@ singleQuote: true
 <div class="{{hello}} {{world}}"></div>
 
 =====================================output=====================================
-<script id="entry-template" type="text/x-handlebars-template">
-  <div class="entry">
+<script id='entry-template' type='text/x-handlebars-template'>
```
evilebottnawi: Problem what we change default `"` on `'`. We should keep `"` quotes for html like syntax (it is preferable developers).

dcyriller: No, no, I don't change the default quote in this PR. In that specific test, the `singleQuote` option is true.

tests/handlebars-attr-node/brackets.hbs
```diff
@@ -0,0 +1,2 @@
+<div title='My "title"'></div>
+<div title="My other 'title'"></div>
```
evilebottnawi: Let's add here:
- `<div title='My title'></div>`
- `<div title="My title"></div>`

for consistently testing and we can merge :+1:

dcyriller: 👍
I've added these two cases in a last commit.

PR #6378, opened 2019-08-09
chore(Handlebars): Refactor test suite

dcyriller:
This PR has **only** impacts on handlebars / glimmer tests.

1. It merges `tests/handlebars/` and `tests/glimmer/` (as both test the same thing: `language-handlebars`).

2. It splits `tests/handlebars` into: `tests/handlebars-basics`, `tests/handlebars-block-statement`, `tests/handlebars-comment`, `tests/handlebars-concat-statement`, `tests/handlebars-element-node`, `tests/handlebars-escape`, `tests/handlebars-last-line`, `tests/handlebars-mustache-statement`, `tests/handlebars-newline`, `tests/handlebars-sub-expression`, `tests/handlebars-text-wrap`, `tests/handlebars-whitespace`. These directories test specific glimmer nodes (`element-node`, `sub-expression` for instance). They also test specific complex issues (such as `basic`, `whitespace`, `text-wrap`). The split is similar to what is done for other languages.

3. It adds whitespaces / newlines / eol-last tests. These tests are taken from html and jsx. It makes sense to have similar tests as Prettier has different printers for these languages. I hope in the future we'll be able to leverage `html-printer` as a preprocessor, but it's not done yet (and I'm not sure this is possible).

4. It merges a few duplicated tests (in this commit: https://github.com/prettier/prettier/commit/ca5ba233e51207dfff192a21681286e82af67146).

Also...

5. It highlights a few bugs! Like here: https://github.com/prettier/prettier/commit/c49dc81d02b4f33dc0f559eec20427edb2955e66#diff-fc35fa87ea555cd658d5fe1450a0e617R58. It should be `</button><button>` instead of `</button>\n\s\s<button>`.

Sorry for the upcoming conflicts!


dcyriller:
> It is always bad when we move tests because it is really hard to review due a lot of git diffs, I hope that we did not miss anything 😄

ahah yes indeed 😉

I felt it would be easier to review in one batch. But if you prefer, I can open multiple PRs. Do not hesitate to let me know.

evilebottnawi:
@dcyriller in one PR is good, no need multiple PRs, just double check what nothing was missed :smile:

dcyriller:
> just double check what nothing was missed 😄

double checking at the moment 😀

Please note that some tests were failing: those run with `AST_COMPARE=true yarn test` command.

Indeed, I've added quite a few tests in this PR. With some of them, I've uncovered a bug that is reproductible with the above mentioned command. You can have a look to the bug in [this diff](https://github.com/prettier/prettier/pull/6378/commits/7bc9596bd6e17eb396336137faf2e746d3804501#diff-f4342a000be92bb26a93b8546932dcd5L74). Issue is: `<span />` is transformed into `<span></span>` making the AST comparison fail.

I suspect it will have to be fixed in Glimmer parser. Not sure for now. This is why I've removed the tests that were introducing self-closing html elements. I propose to open an issue or a PR with the failing tests. Does it sound good to you @evilebottnawi?

evilebottnawi:
:+1:

duailibe:
@dcyriller Now the other PRs have merge conflicts. You can ping me once they're fixed, and I'll merge them

Great job 👍

PR #6384, opened 2019-08-12
docs: webstorm info on how to watch partials

schuer:
Since file watchers ignore partials like `_component.scss` (underscore!) when imported in some `main.scss`, they won’t be prettified. Use the `COMPILE_PARTIAL=true` environment variable to also watch and prettify partials.


duailibe:
Thanks!

PR #6398, opened 2019-08-18
fix: format JSX Fragment correctly, consider Fragment as block

JounQin:

### DIFF
--- a/changelog_unreleased/mdx/pr-6398.md
+++ b/changelog_unreleased/mdx/pr-6398.md
@@ -1,0 +1,21 @@
+#### MDX: fix JSX Fragment format incorrectly ([#6398](https://github.com/prettier/prettier/pull/6398) by [@JounQin](https://github.com/JounQin))
+
+Previous versions format JSX Fragment incorrectly in mdx, this has been fixed in this version.
+
+<!-- prettier-ignore -->
+```md
+<!-- Input -->
+<>
+    test   <World        />   test
+</>       123
+
+<!-- Output (Prettier stable) -->
+<>
+test <World        /> test
+</> 123
+
+<!-- Output (Prettier master) -->
+<>
+  test <World /> test
+</> 123
+```
--- a/src/language-markdown/mdx.js
+++ b/src/language-markdown/mdx.js
@@ -26,7 +26,7 @@
 
 const IMPORT_REGEX = /^import\s/;
 const EXPORT_REGEX = /^export\s/;
-const BLOCKS_REGEX = "[a-z\\.]*(\\.){0,1}[a-z][a-z0-9\\.]*";
+const BLOCKS_REGEX = "[a-z][a-z0-9]*(\\.[a-z][a-z0-9]*)*|";
 const COMMENT_REGEX = "<!---->|<!--(?:-?[^>-])(?:-?[^-])*-->";
 const EMPTY_NEWLINE = "\n\n";
 
--- a/tests/mdx/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/mdx/__snapshots__/jsfmt.spec.js.snap
@@ -265,6 +265,12 @@ printWidth: 80
 
 ---
 
+<>
+    test   <World        />   test
+</>       123
+
+---
+
 | Column 1 | Column 2 |
 |---|---|
 | Text | <Hello>Text</Hello> |
@@ -296,6 +302,12 @@ printWidth: 80
 <Hello>
   test <World /> test
 </Hello> 234
+
+---
+
+<>
+  test <World /> test
+</> 123
 
 ---
 
@@ -342,6 +354,12 @@ semi: false
 
 ---
 
+<>
+    test   <World        />   test
+</>       123
+
+---
+
 | Column 1 | Column 2 |
 |---|---|
 | Text | <Hello>Text</Hello> |
@@ -373,6 +391,12 @@ semi: false
 <Hello>
   test <World /> test
 </Hello> 234
+
+---
+
+<>
+  test <World /> test
+</> 123
 
 ---
 
--- a/tests/mdx/jsx.mdx
+++ b/tests/mdx/jsx.mdx
@@ -27,6 +27,12 @@
 
 ---
 
+<>
+    test   <World        />   test
+</>       123
+
+---
+
 | Column 1 | Column 2 |
 |---|---|
 | Text | <Hello>Text</Hello> |


JounQin:
@evilebottnawi I have no idea why babel parser throws, and `loc` is undefined only on Node 4, would you like to help?

j-f1:
You could try changing that section to this:

```js
if (error.loc) {
  throw createError(
    // babel error prints (l:c) with cols that are zero indexed
    // so we need our custom error
    error.message.replace(/ \(.*\)/, ""),
    {
      start: {
        line: error.loc.line,
        column: error.loc.column + 1
      }
    }
  );
} else {
  throw error;
}

**Comment by JounQin** [comment]
It seems `Array#includes` is not polyfilled on Node 4.

**Comment by JounQin** [comment]
@j-f1 The tests have been fixed, would you like to review again?

**Comment by JounQin** [comment]
@evilebottnawi @ikatyang Please help to review.

File: src/language-markdown/parser-markdown.js [hunk 1]
```diff
@@ -7,7 +7,8 @@ const parseFrontMatter = require("../utils/front-matter");
 const { mapAst, INLINE_NODE_WRAPPER_TYPES } = require("./utils");
 const mdx = require("./mdx");
 const remarkMath = require("remark-math");
-const htmlParser = require("../language-html/parser-html").parsers.html;
+const jsParser = require("../language-js/parser-babylon").parsers
```
ikatyang: Just noticed that you added `htmlParser` in #6332 and you want to replace it with `jsParser` here. I'm wondering why you need to add those parsers given that it should already be handled by `language-markdown/embed.js`:

https://github.com/prettier/prettier/blob/23146404850011972f695fb6bc2b8113c3cffbfc/src/language-markdown/embed.js#L56-L61

JounQin: It can not handle adjacent JSX nodes as I described in the first PR.

ikatyang: I guess we could simply always wrap it in a JSX fragment and add an internal option to handle this case in the babylon printer.

Something like:

```js
// language-markdown/embed.js

switch (node.type) { 
  // ...

  case "jsx": 
    return textToDoc(
      `<>${node.value}</>`,
      {parser: "__js_expression", __maybe_mdx_adjacent: true}
  ); 
  
  // ...
}
```

```js
// language-js/printer-estree.js

switch (node.type) {
  // ...
 
  case "JSXFragment":
    if (
      options.__maybe_mdx_adjacent &&
      node.children.length === 1 &&
      node.children[0].type === "JSXFragment" &&
      path.getParentNode().type === "JsExpressionRoot"
    ) {
      return outputDocWithoutOpenCloseFragment;
    }
    return originalOutputDoc;

  // ...
}
```

This way we don't need to parse the same code twice.

JounQin: Great advice, although I prefer `<$></$>` which is not considered as `html/jsx` in `mdx`, should we?

ikatyang: It shouldn't matter which symbol you want to use as long as it works.

JounQin: @ikatyang I tried to implement it locally and found out that we may add a lot of mdx specific logics in babel parser, of course maybe this is related to that I'm not so familiar with the code base, but it seems not reasonable to do it to me.

And also,

```jsx
// input
<div>1</div>   123

// output by embed
<div>1</div>{" "}
123

// output by wrapper
<div>1</div> 123
```

I prefer current output for mdx and want to keep it as is, what do you think?

ikatyang: Not quite sure why `123` is parsed as part of the JSX node, but the current output for MDX shouldn't be hard to implement in this way, we could use some regexes to prevent the non JSX part from being passed into the babel parser:

```js
// language-markdown/parser-markdown.js

const nonJsxPart = jsxText.match(/^[^]+<\/\w+>([^]*)$/)[1];
const jsxPart = !nonJsxPart ? jsxText : jsxPart.slice(0, -nonJsxPart.length);
```

ikatyang: FYI you could put those MDX-specific logic in another file, just like what we did for Vue-specific logic.

https://github.com/prettier/prettier/blob/23146404850011972f695fb6bc2b8113c3cffbfc/src/language-js/printer-estree.js#L443-L446
https://github.com/prettier/prettier/blob/23146404850011972f695fb6bc2b8113c3cffbfc/src/language-js/html-binding.js#L7

JounQin: @ikatyang Sorry about this, but can we keep it as is temporarily? I don't have much time on it recently, I will keep the refactor soon as I'm not too busy. Of course, it would be great if you could offer the help.

ikatyang: No worry, there's no time limit for this PR, you can work on it whenever you have time. Let me know if you need some help.

thorn0: > Not quite sure why 123 is parsed as part of the JSX node

This behavior is correct. It should be parsed JSX. See the interactive playground on https://mdxjs.com/:

image

JounQin: @thorn0 It's not true, see https://github.com/prettier/prettier/pull/6949#discussion_r346211287

thorn0: What exactly is not true? Is that playground broken or what?

thorn0: image

JounQin: OK, sorry I misunderstand the code, jsx with text prefix should be considered as html, without text prefix it is jsx. That's why I was in a dilemma at https://github.com/prettier/prettier/pull/6398#discussion_r315261475 .

JounQin: @ikatyang I don't think your regex `/^[^]+<\/\w+>([^]*)$/` is always correct, I prefer parsing the jsx value by `babel`, regex is not a good idea for code parser here.

The following examples are not considered as `jsx` part in your regex.

```jsx
<></>

<li/>
```

So I think we should keep it as by using `__js_expression` here. Any further idea?

JounQin: @evilebottnawi @ikatyang @thorn0 

Please review this PR again, although I didn't change the code according to @ikatyang's advice, see my explain above.

And I'm so sorry that my first PR affects so many people, so I hope we can fix is quickly. If more refactoring are needed, let's do it later.

evilebottnawi:
Sorry, i am not familiar with mdx, i can't review

thorn0:
I'm not really familiar with MDX too albeit I learnt a lot about it today. =)

JounQin:
@evilebottnawi @ikatyang @thorn0

This PR becomes much cleaner and easier for reviewing after #6949 been merged, please help to review it.

src/language-js/parser-babylon.js
```diff
@@ -58,17 +58,21 @@ function createParse(parseMethod, extraPlugins) {
...
     } catch (error) {
-      throw createError(
-        // babel error prints (l:c) with cols that are zero indexed
-        // so we need our custom error
-        error.message.replace(/ \(.*\)/, ""),
-        {
-          start: {
-            line: error.loc.line,
-            column: error.loc.column + 1
+      if (error.loc) {
```
thorn0: Can Babel throw an error without `loc`?

src/language-markdown/mdx.js
```diff
@@ -26,7 +26,7 @@
 
 const IMPORT_REGEX = /^import\s/;
 const EXPORT_REGEX = /^export\s/;
-const BLOCKS_REGEX = "[a-z\\.]*(\\.){0,1}[a-z][a-z0-9\\.]*";
+const BLOCKS_REGEX = "[a-z\\.]*(\\.){0,1}[a-z]?[a-z0-9\\.]*";
```
thorn0:
```suggestion
const BLOCKS_REGEX = "[a-z\\.]*(\\.){0,1}[a-z][a-z0-9\\.]*|";
```
We want "a valid JSX element name OR the empty string". Your regexp allows invalid element names like `2019` to slip through. I still don't like the resulting regexp because it seems to allow strange things like `<aa...vv...>`, but for this PR my suggestion might suffice.

src/language-js/parser-babylon.js
JounQin: It didn't, babel itself targets newer node version than prettier, it thrown `includes is not a function` error, you can check the discussion flow at two months ago when I tried to included `js_expression` parser.

src/language-markdown/mdx.js
JounQin: @thorn0
How about this one:
```js
const BLOCKS_REGEX = "(?:[a-z])[a-z\\.]*(\\.){0,1}[a-z]?[a-z0-9\\.]*$";
```

JounQin: @thorn0 How about this one:

```js
const BLOCKS_REGEX = "([a-z]([a-z0-9]*\\.?[a-z]+[a-z0-9]*)*|)";
```

src/language-js/parser-babylon.js
JounQin: Of course, I'm fine to revert it if you want.

thorn0: Yes, let's try reverting it. It doesn't seem to belong to this PR.

src/language-markdown/mdx.js
thorn0: Yes, this one is better. The outer parens aren't needed though.

JounQin: I'd like to keep it as is which seems clearer to me.

src/language-js/parser-babylon.js
JounQin: @thorn0 Done, shall we merge this PR then?

changelog_unreleased/mdx/pr-6398.md
```diff
@@ -0,0 +1,24 @@
+#### MDX: fix JSX Fragment format incorrectly ([#6398] by [@JounQin])
```
thorn0: We don't write links like this anymore in the changelog. Please change to this syntax: `[...](https://...)`

src/language-markdown/mdx.js
thorn0: The regexp on the next line does not have them.

changelog_unreleased/mdx/pr-6398.md
JounQin: Done.

PR #6404, opened 2019-08-20
Update ?? precedence to match stage 3 proposal

### LINKED ISSUE #6403: Parentheses around nullish coalescing operator (??) within logical expressions
in the latest stage 3 draft, nullish coalescing can no longer immediately contain, or be contained within, `&&` and `||`: https://tc39.es/proposal-nullish-coalescing/

so, `foo ?? bar || baz` is illegal, but you can write `(foo ?? bar) || baz`.

[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAKVAzCEAEB+PHAIwEMAnAShwB9riSAvCkAGhAgAcYBLaAZ2ShyZCAHcACuQQCUJAG4RuAE1YgiZEmADWcGAGUOm7lADmyGGQCucNgAsYAWwA2AdVvd4fQ2Dh7pH7jkPAE9kcD4BNmM+ODIYcQ0TBxJkDBInGLYAKz4ADwAhDW1dPRIHOAAZYzhU9MyQHNy9YxMnOABFSwh4WoybEEMyGLIw0iJgp2hVDjJjGBdlGFtkAA4ABjYZiBiXDQ4wmbhhuRq2AEcu+ATOGRASPgBaKDg4JVfVMjgL7k+EkiSUkg0n02DEHNxzFZ+nwWm1Ot0akC6v0YCQiAslEtkAAmNgWEjcJwtADCEAcyTCUGgpxAlhiABU0TJgTEAL6soA)

**Input:**
```jsx
(foo ?? bar) || baz
```

**Output:**
```jsx
foo ?? bar || baz
```

**Expected behavior:**
```jsx
(foo ?? bar) || baz
```

Flow v0.106.0 implements the change to the spec, making this a parse error. So, prettier will turn valid code into invalid code by removing the parens.

vjeux:
@mroch told me that the spec was updated ( https://tc39.es/proposal-nullish-coalescing/ ) with the following two changes:
- ?? has lower precedence than ||. (previously equal)
- ?? cannot immediately contain, or be contained within, an && or || operation. (parens are now required)

The safe way to upgrade is to run prettier with just this change so that parenthesis are properly added. Then to upgrade the parser (only flow supports the new precedence in the next release, babel hasn't been updated yet) in prettier so that the new precedence is being used. Failure to do so will result in parse errors because of missing parenthesis so we will not have silent errors.

There was only 44 callsites mixing || and ?? in a confusing way in the Facebook codebase so it shouldn't be a huge deal in practice.

Fixed #6403

### DIFF
--- a/tests/nullish_coalescing/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/nullish_coalescing/__snapshots__/jsfmt.spec.js.snap
@@ -14,6 +14,11 @@ foo ? bar ?? foo : baz;
 
 foo ?? (bar ?? baz);
 
+(foo ?? baz) || baz;
+
+// Note: this will trigger a syntax error once the parsers have been
+// updated to the latest specification. If you are doing the upgrade,
+// please remove the following line (and this comment).
 foo ?? baz || baz;
 
 (foo && baz) ?? baz;
@@ -28,7 +33,12 @@ foo ? bar ?? foo : baz;
 
 foo ?? (bar ?? baz);
 
-foo ?? baz || baz;
+(foo ?? baz) || baz;
+
+// Note: this will trigger a syntax error once the parsers have been
+// updated to the latest specification. If you are doing the upgrade,
+// please remove the following line (and this comment).
+(foo ?? baz) || baz;
 
 (foo && baz) ?? baz;
 foo && (baz ?? baz);
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -239,7 +239,8 @@ function setLocEnd(node, index) {
 const PRECEDENCE = {};
 [
   ["|>"],
-  ["||", "??"],
+  ["??"],
+  ["||"],
   ["&&"],
   ["|"],
   ["^"],
--- a/tests/nullish_coalescing/nullish_coalesing_operator.js
+++ b/tests/nullish_coalescing/nullish_coalesing_operator.js
@@ -6,6 +6,11 @@ foo ? bar ?? foo : baz;
 
 foo ?? (bar ?? baz);
 
+(foo ?? baz) || baz;
+
+// Note: this will trigger a syntax error once the parsers have been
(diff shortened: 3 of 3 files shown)


tests/nullish_coalescing/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -28,7 +28,7 @@ foo ? bar ?? foo : baz;
 
 foo ?? (bar ?? baz);
 
-foo ?? baz || baz;
+(foo ?? baz) || baz;
```
mroch: when you start using a fixed parser, the `foo ?? bar || baz` above will start to error, so might want to remove that case?

can you add a test for `(foo ?? bar) || baz` to make sure it doesn't strip the parens anymore?

mroch:
some background:

https://github.com/tc39/proposal-nullish-coalescing/pull/39 (PR that changed the draft spec)
https://github.com/tc39/proposal-nullish-coalescing/issues/15
https://github.com/tc39/proposal-nullish-coalescing/issues/26

tests/nullish_coalescing/__snapshots__/jsfmt.spec.js.snap
vjeux: Good points

tests/nullish_coalescing/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -28,7 +33,12 @@ foo ? bar ?? foo : baz;
...
+
+// Note: this will trigger a syntax error once the parsers have been
+// updated to the latest specification. If you are doing the upgrade,
+// please remove the following line (and this comment).
```
vjeux: Instead of deleting right now, I added a comment for the person that does the upgrade later.

vjeux:
Arg sorry, I'll send an update later this week.

Issue #5277, opened 2018-10-16
If list-different detects differences, it should print command for fixing


melissachang:
It would be nice if `list-different` printed the command for fixing (`prettier --write <files>`) if it detected differences. This way, someone doesn't have to hunt down the syntax.

daubaris:
hi! I'd love to work on this!

kachkaev:
@melissachang I agree with you that printing a command for fixing is a good idea! See https://github.com/prettier/prettier/issues/5520 with some thoughts on this.

kx-chen:
Any chance I could pick this up?

lydell:
Sure!

I think we should leave `--list-different` as it is and add this to `--check`.

### Alternative 1

```
prettier --check "**/*.{css,html,js,md,yml}"

Checking formatting...
html/all-types/index.html
html/experiments/viewport/script.js
Code style issues found in the above file(s). Forgot to run Prettier?
Running something like this should fix the issues:
prettier --write "**/*.{css,html,js,md,yml}"
```

### Alternative 2

```
prettier --check "**/*.{css,html,js,md,yml}"

Checking formatting...
html/all-types/index.html
html/experiments/viewport/script.js
Code style issues found in the above file(s). Forgot to run Prettier?
Running something like this should fix the issues:
prettier --write html/all-types/index.html html/experiments/viewport/script.js
```

### Alternative 3

```
prettier --check "**/*.{css,html,js,md,yml}"

Checking formatting...
html/all-types/index.html
html/experiments/viewport/script.js
Code style issues found in the above file(s). Forgot to run Prettier?
Check how your project runs Prettier! Here are some common ways:
prettier --write path/to/file.js
npm run prettier
????
```

### Alternative 4

Anyone has other ideas?

### Things to think about

- Should we prefix the suggested command with `npx` to increase the chance of the command actually working?
- We need to handle file names with spaces in them / shell quoting if we print globs or file names.

/cc @kachkaev since you came up with the nice `--check` flag.

kx-chen:
Awesome! Thank you so much for the fast reply. 

I personally prefer Alternative 1; it seems simpler. 
Additionally, I think prefixing with `npx` for a high chance of it working is a good idea as well.

I can submit a PR with the general changes, and I'm sure we can discuss more then.

kx-chen:
Actually, on second thought, it seems like using approach 2 works better. It feels more explicit, and might have a lower chance of changing files that were not intended on being changed.

[...]

PR #6406, opened 2019-08-21
Print command for fixing files when using --check

kx-chen:
Prints suggested command for fixing files
when using --check

Closes https://github.com/prettier/prettier/issues/5277

Heya, first time contributor to this repo here. Would love to hear feedback! 

Not sure if I need to add to changelog, it doesn't seem like this is a user-facing change in the way prettier is being used. Please let me know :)

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -433,12 +433,17 @@ function eachFilename(context, patterns, callback) {
   }
 }
 
+function getRunPrettierCommand(files) {
+  return `npx prettier --write ${files.join(" ")}`;
+}
+
 function formatFiles(context) {
   // The ignorer will be used to filter file paths after the glob is checked,
   // before any files are actually written
   const ignorer = createIgnorerFromContextOrDie(context);
 
   let numberOfUnformattedFilesFound = 0;
+  const unformattedFileNames = [];
 
   if (context.argv["check"]) {
     context.logger.log("Checking formatting...");
@@ -583,6 +588,7 @@ function formatFiles(context) {
     ) {
       context.logger.log(filename);
       numberOfUnformattedFilesFound += 1;
+      unformattedFileNames.push(filename);
     }
   });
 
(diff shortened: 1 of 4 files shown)


src/cli/util.js
```diff
@@ -433,12 +433,17 @@ function eachFilename(context, patterns, callback) {
   }
 }
 
+function getRunPrettierCommand(files) {
```
kx-chen: Not too sure about the naming of this -- any thoughts?

kachkaev:
IMO this may cause more confusion that benefit. `npx prettier` can resolve to a different prettier version and config, which may 'break' unfixed files even further. Besides, if the project has a large number of `unformattedFileNames`, the fix command will get huge.

kx-chen:
@kachkaev Ah, makes sense. Should I close this altogether or go for the first approach in the issue?

kachkaev:
@kx-chen I'm not sure how to implement this feature robustly, even using the first approach you refer to. Some thoughts here: https://github.com/prettier/prettier/issues/5277#issuecomment-523388327

lydell:
After reading @kachkaev's comment (https://github.com/prettier/prettier/issues/5277#issuecomment-523388327) I'm not as excited about this feature anymore. Since we have to guess at a suitable command, I think we can cause more harm than good.

The issue (#5277) mentions:

> This way, someone doesn't have to hunt down the syntax.

But thinking about it, I think that's the wrong response to a failed `prettier --check`. One's reaction shouldn't be "Let's google how to fix prettier" or "Let's read through `prettier --help`" but "I wonder how Prettier is set up to run in this project" and then looking that up in the project's README.md or package.json, or asking somebody else working on the project.

@kx-chen What do you think? Sorry for having wasted your time, but I wonder if you shouldn't close this PR and #5277?

kx-chen:
@lydell Yep agreed! Doesn't seem like there's a simple way to get the command correct. I'm open to all thoughts. 

A simple approach would be to leave a small note suggesting that the user run prettier with --write and not worry about the exact command. Thoughts?

lydell:
Hmm, yeah, maybe.

Did you want to make this PR because you were looking for an easy way to get into contributing to Prettier, or because you want the feature a lot yourself? If the latter, what does a typical scenario look like that you use to go through and would like to improve?

kx-chen:
Mostly to start contributing and get my feet wet. Now thinking about it, there definitely isn't too much of a need for this. I'm totally okay with closing it.

lydell:
Alright, we'll close then. Feel free to try some other issue and ask for help when needed!

PR #6429, opened 2019-08-28
Document behavior of line endins for ranges

sheerun:
It's not intuitive so I think it should be documented.

Side node: selecting range of lines to format would be much more useful as e.g. git diff output only contains number of lines, not ranges of characters


thorn0:
Thanks, but this edit cannot be accepted as CRLF line endings count as two characters everywhere, not only on Windows.

Issue #6443, opened 2019-09-01
TypeScript: type assertions (as) aren't wrapped


thorn0:
How has this gone unnoticed for so long?

**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAjAlgMwgCcoBhI3KAcziIFkI0ALOAEwAVaAHAGzgA8AygEMA7q1qYAvJgA6UTJmFoA1gBUiEAG4JhAITi6AInC4JWaAEoRRwqGiUOqPXAFs0cAOI8AnlyZoABLCAF7CRKwAchAQXGhquHAJcADc8orywjpERGx6wjAsRGh6EPyaEK56AK5UVD6R1UQ8AHR2EKw+UHCk0KyMdMI8BXAA8tkjlnAwlBA80nIKSlBQ1cNEvRCGrIK8uGBwaADqQyrHwj4nPo6YKhBgTMLYm4mewq5jUKPYc3AAqlBRHAsql5JlsrlWPlCrQSmUKlVavVGs0FsB0ssOl0en0BkMRuNaJNprMeEgMYo7Kt1pttrsXAdzjwzlcrjc7g8ni8vO9Pt9fgCgSD5ABfNJQcG0SHQopw8oxRF1BpNHgACgx7U63V6UH6aEGw3ghKIxJmUDmbRWa3CtLY9P2hxOzPOlwuNwpt3uj2eMVevNGXx+fEFwJ08gAlCkQAAaECxGboZCgcKaUTscIINDIEBZCC4VgxkDYE1gFTTXbCMCUKjIGBEapwWNMGCuHhHJi4eBoLiVuCCTOd3BaTs+bNgNBZ2OUDxEGDsE1UVzCZD4IYeWMAKzQ-D0JbLMBEHwAMpQ4Cu142QFuhNW+ABFaoQeDnnjrkA94q0bMwPyHMAULgYELLgKFgI580KZAAA4AAZYxAxg4COE0uGzEDDloHRCwAR0feB51iLMUGUABabo2DYQtclw3BcnnYRF2XJBV1fS8PFcXBa3rNjbzgB8nzPZiL1jGAnnA1hIKQAAmESTVwFxqF6Vwl2zc1ukLaoPDUJ4iJYjwRRFIA)
```sh
--parser typescript
```

**Input:**
```tsx
const bifornCringerMoshedPerplexSawder = 
[...]

PR #6450, opened 2019-09-04
TypeScript: format TSAsExpression with same logic as BinaryExpression

kaicataldo:
fixes https://github.com/prettier/prettier/issues/6443.

I took a stab at working on wrapping `TSAsExpression` nodes! ~~In taking a look at this, it looks like the behavior of moving the initial declaration value up to the same line as the `=` operator happens with other declarations and I think the change might have to happen in that logic (here's an example using the [Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBDTBeTAHSk1LLQGsATAKwDMALbNAG2vqcqsZuerpqd+ggPw40VCgM5U205i0nyJc3gxV0KvNAyoS6+zEkz79AbmLFIUDJgBG+IiTKkuHBe0Z9uDbcKHc2kqCzLJSIazBMqrY6rKa2roS+qF0Bnrc5iAANCAQAA4wAJboyKDYAE4VEADuAAqVCGjIINgAbhBFVDkgdhXYYBRwMADK+QNFUADmyDAVAK5wuQwwALYsAOoMRfBo42BwI007RW07AJ4tYGjNuZNocBUwdf1Tq9jIdNgsD7k0aAAPABC-UGwxG2FWcAAMpM4J9vr8QP8ASNJlMWHAAIrzCDwBE-JYgcYVB4VFp2bB2c4saA9fIVSYwDZdGAMZAADgADLkGRAHht+vkWgy4GS2vDcgBHXHwF4FZooZgAWigcDgVA1PQqcBlRR1L2wbw+SC+hNyD1WRVmCyJaHRmJxePhpsRRJgVJZVDZyAATLk5tgiix0QBhCCrd4tKDQSUgeYPAAqVMVZoeAF900A)).~~

~~This PR instead focuses on wrapping the `TSAsExpression` node correctly (ensuring that the line break occurs after the `as` operator).~~

This now takes a different approach and converts the `TSAsExpression` node to match the `BinaryExpression` node interface so we can use the same formatting logic for both.

I'm pretty new to the Prettier codebase, and any and all guidance is very welcome!

### DIFF
--- a/src/language-js/postprocess.js
+++ b/src/language-js/postprocess.js
@@ -2,15 +2,19 @@
 
 const { getLast } = require("../common/util");
 
-// fix unexpected locEnd caused by --no-semi style
 function postprocess(ast, options) {
   visitNode(ast, node => {
     switch (node.type) {
+      // Fix unexpected locEnd caused by --no-semi style
       case "VariableDeclaration": {
         const lastDeclaration = getLast(node.declarations);
         if (lastDeclaration && lastDeclaration.init) {
           overrideLocEnd(node, lastDeclaration);
         }
+        break;
+      }
+      case "TSAsExpression": {
+        convertTSAsExpression(node);
         break;
       }
     }
@@ -42,6 +46,17 @@ function postprocess(ast, options) {
   function locEnd(node) {
     return options.parser === "flow" ? node.range[1] : node.end;
   }
+}
+
+/* Convert TSAsExpression to BinaryExpression interface
+ * https://github.com/estree/estree/blob/master/es5.md#binaryexpression
+ */
+function convertTSAsExpression(node) {
+  node.left = node.expression;
+  node.operator = "as";
+  node.right = node.typeAnnotation;
+  delete node.expression;
+  delete node.typeAnnotation;
 }
 
 function visitNode(node, fn) {
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -3099,12 +3099,29 @@ function printPathNoParens(path, options, print, args) {
       return "unknown";
     case "TSVoidKeyword":
       return "void";
-    case "TSAsExpression":
-      return concat([
-        path.call(print, "expression"),
-        " as ",
-        path.call(print, "typeAnnotation")
+    case "TSAsExpression": {
+      const typeText = options.originalText.slice(
+        options.locStart(n.typeAnnotation),
+        options.locEnd(n.typeAnnotation)
+      );
+      // Defer to expansion logic for objects, arrays, and annotations wrapped in parentheses.
+      const mightBeMultiline = /^[({[]/.test(typeText.trim());
+
+      return conditionalGroup([
+        concat([
+          path.call(print, "expression"),
+          " as ",
+          path.call(print, "typeAnnotation")
+        ]),
+        concat([
+          path.call(print, "expression"),
+          " as ",
+          mightBeMultiline
+            ? path.call(print, "typeAnnotation")
+            : indent(concat([softline, path.call(print, "typeAnnotation")]))
+        ])
       ]);
+    }
     case "TSArrayType":
       return concat([path.call(print, "elementType"), "[]"]);
     case "TSPropertySignature": {
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -537,7 +537,8 @@ function printPathNoParens(path, options, print, args) {
       );
     case "BinaryExpression":
     case "LogicalExpression":
-    case "NGPipeExpression": {
+    case "NGPipeExpression":
+    case "TSAsExpression": {
       const parent = path.getParentNode();
       const parentParent = path.getParentNode(1);
       const isInsideParenthesis =
@@ -3099,29 +3100,6 @@ function printPathNoParens(path, options, print, args) {
       return "unknown";
     case "TSVoidKeyword":
       return "void";
-    case "TSAsExpression": {
-      const typeText = options.originalText.slice(
-        options.locStart(n.typeAnnotation),
-        options.locEnd(n.typeAnnotation)
-      );
-      // Defer to expansion logic for objects, arrays, and annotations wrapped in parentheses.
-      const mightBeMultiline = /^[({[]/.test(typeText.trim());
-
-      return conditionalGroup([
-        concat([
-          path.call(print, "expression"),
-          " as ",
-          path.call(print, "typeAnnotation")
-        ]),
-        concat([
-          path.call(print, "expression"),
-          " as ",
-          mightBeMultiline
-            ? path.call(print, "typeAnnotation")
-            : indent(concat([softline, path.call(print, "typeAnnotation")]))
-        ])
-      ]);
-    }
     case "TSArrayType":
       return concat([path.call(print, "elementType"), "[]"]);
     case "TSPropertySignature": {
@@ -5828,7 +5806,8 @@ function isBinaryish(node) {
   return (
     node.type === "BinaryExpression" ||
     node.type === "LogicalExpression" ||
-    node.type === "NGPipeExpression"
+    node.type === "NGPipeExpression" ||
+    node.type === "TSAsExpression"
   );
 }
 
--- a/tests/typescript_as/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/typescript_as/__snapshots__/jsfmt.spec.js.snap
@@ -33,6 +33,7 @@ const state = JSON.stringify({
 (foo.bar as any)++;
 
 const value1 = thisIsAReallyReallyReallyReallyReallyLongIdentifier as SomeInterface;
+const value1 = thisIsAnIdentifier as thisIsAReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyLongInterface;
 const value2 = thisIsAReallyLongIdentifier as (SomeInterface | SomeOtherInterface);
 const value3 = thisIsAReallyLongIdentifier as { prop1: string, prop2: number, prop3: number }[];
 const value4 = thisIsAReallyReallyReallyReallyReallyReallyReallyReallyReallyLongIdentifier as [string, number];
@@ -76,6 +77,8 @@ const state = JSON.stringify({
 
 const value1 = thisIsAReallyReallyReallyReallyReallyLongIdentifier as
   SomeInterface;
+const value1 = thisIsAnIdentifier as
+  thisIsAReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyLongInterface;
 const value2 = thisIsAReallyLongIdentifier as (
   | SomeInterface
   | SomeOtherInterface
--- a/tests/typescript_as/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/typescript_as/__snapshots__/jsfmt.spec.js.snap
@@ -9,9 +9,12 @@ printWidth: 80
 const name = (description as DescriptionObject).name || (description as string);
 this.isTabActionBar((e.target || e.srcElement) as HTMLElement);
 (originalError ? wrappedError(errMsg, originalError) : Error(errMsg)) as InjectionError;
-'current' in (props.pagination as Object)
-start + (yearSelectTotal as number)
-scrollTop > (visibilityHeight as number)
+'current' in (props.pagination as Object);
+('current' in props.pagination) as Object;
+start + (yearSelectTotal as number);
+(start + yearSelectTotal) as number;
+scrollTop > (visibilityHeight as number);
+(scrollTop > visibilityHeight) as number;
 export default class Column<T> extends (RcTable.Column as React.ComponentClass<ColumnProps<T>,ColumnProps<T>,ColumnProps<T>,ColumnProps<T>>) {}
 export const MobxTypedForm = class extends (Form as { new (): any }) {}
 export abstract class MobxTypedForm extends (Form as { new (): any }) {}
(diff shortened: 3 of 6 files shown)


src/language-js/printer-estree.js
```diff
@@ -3094,12 +3094,29 @@ function printPathNoParens(path, options, print, args) {
...
-        path.call(print, "expression"),
-        " as ",
-        path.call(print, "typeAnnotation")
+    case "TSAsExpression": {
+      const typeText = options.originalText.slice(
+        options.locStart(n.typeAnnotation),
+        options.locEnd(n.typeAnnotation)
+      );
+      // Defer to expansion logic for objects, arrays, and annotations wrapped in parentheses.
+      const mightBeMultiline = /^[({[]/.test(typeText.trim());
```
kaicataldo: Thoughts on a better way to check for this? If we always insert a newline here, the formatting isn't as nice. i.e.

```ts
const value2 = thisIsAReallyLongIdentifier as (SomeInterface | SomeOtherInterface);
```
becomes
```ts
const value2 = thisIsAReallyLongIdentifier as 
  (SomeInterface | SomeOtherInterface);
```
rather than
```ts
const value2 = thisIsAReallyLongIdentifier as (
  | SomeInterface
  | SomeOtherInterface
);
```

src/language-js/printer-estree.js
```diff
@@ -3094,12 +3094,29 @@ function printPathNoParens(path, options, print, args) {
...
+
+      return conditionalGroup([
```
kaicataldo: Is this the most efficient way to do this? I was trying to get it to work with one group, but I started getting line breaks on lines that were shorter than the `printWidth`.

tests/typescript_as/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -32,25 +32,29 @@ const state = JSON.stringify({
...
 <boolean>bValue ? 0 : -1;
 
-
 =====================================output=====================================
 const name = (description as DescriptionObject).name || (description as string);
 this.isTabActionBar((e.target || e.srcElement) as HTMLElement);
-(originalError
-  ? wrappedError(errMsg, originalError)
-  : Error(errMsg)) as InjectionError;
+(originalError ? wrappedError(errMsg, originalError) : Error(errMsg)) as
```
kaicataldo: Is this change desirable? It makes sense, and I actually think it's easier to see what's going on, but others might not like it.

src/language-js/printer-estree.js
evilebottnawi: No need regexp here, you need check ast nodes inside

thorn0:
```
const value2 = thisIsAReallyLongIdentifier as 
  (SomeInterface | SomeOtherInterface);
```
looks better to me

thorn0: And especially this
**Prettier pr-6450**
[Playground link](https://deploy-preview-6450--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuc0DOMAEA3AhgGwFc4AmTAXkxgAsBLNASTQEEAlOA-ATwBloBzBgBMEMWgDNacAE6ZcaTAGUIAWzgNYM8bjBwAPO068BAFS4AHOAAVc03GvjSAfAG4QAGhARzY9MlC20hAA7jbSCGjIILjYELRCHiAARnZgANZwMIrmOrRQ-Mgw0sSe1DAq+ADqdPBoObqKEbRi2M1cUWBokZ55aDIwVnb8KrjI2vh9ngBWaAAeAEKpGVn2cDx5cGMEkyAzs4p5-PhwAIqEEPBbE3CeOdJ90lEwFnBoYNK0PonmH7CV8TRkAAOAAMtyCfUqdnMUR+rxk2E2ngAjud4INvJEUPIALRQOBwEQJTzhVG0cKDXDDUZIcY7PoqWiFYo3EBoQ7HM4XTa07asmC4JL-ISApAkTxFXC0fCHADCqhGUSg0CRIEIfRMgqxdNZ2GIGhEsEU70+MGYUCEimexyufQAvnagA)
```sh
--parser typescript
```

**Input:**
```tsx
const value2 = thisIsAReallyLongIdentifier as SomeInterface<ReallyLongTypeParameter>;
```

**Output:**
```tsx
const value2 = thisIsAReallyLongIdentifier as
  SomeInterface<ReallyLongTypeParameter>;

```

is way better than this

**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEA3AhgGwFc4AmTAXkxgAsBLNASTQEEAlOA-ATwBloBzBgBMEMWgDNacAE6ZcaTAGUIAWzgNYM8bjBwAPMrUB5GjL5RBm6dt0A+ANwgANCAgAHMemShc06RAB3AAVfBDRkEFxsCFohZxAAI2kdAGs4GEU3HVoLZBhpYhdqGBV8AHU6eDQs3UUw2jFsBq4IsDRwlxy0GRgg5P4VXGRtfG6XACs0AA8AIWSwNIzcNR4cuGGCMZBJqcUc-nw4AEVCCHgN0bgXLOlu6QiYLjc4NDBpWg94t3fYMtiaZAADgADNd-N0ysk3BFvi8ZNh1i4AI6neB9dzhFDyAC0UDgcBEcRc0jgKNoJL6uAGQyQIy23RUtDyBSuIDQ+0OJzO61pm1ZMFwCT+QgBSBILnyuFo+H2AGFVIMIlBoIiQIRugAVQWYulwAC+eqAA)
```sh
--parser typescript
```

**Input:**
```tsx
const value2 = thisIsAReallyLongIdentifier as SomeInterface<SomeOtherLongInterface>;
```

**Output:**
```tsx
const value2 = thisIsAReallyLongIdentifier as SomeInterface<
  SomeOtherLongInterface
>;

```

kaicataldo: Is there an easier way to access to the raw string value of the node to check if it’s wrapped in parentheses?

evilebottnawi: You don't need raw, you need look in ast for nodes

kaicataldo: Thanks for the input!

Re the parentheses, I think Prettier should still format it correctly, whether they're required or not.

Re the second comment, I think formatting generics should be treated separately from this PR (and none of the changes here affect that, since it only checks the first token after the `as` operator), to make sure it's as small a patch as possible.

kaicataldo: I don't think the parentheses aren't attached to nodes, though, so we'd have to have access to either the raw text or to iterate over a token list. For the objects and arrays, we definitely can check the AST nodes. Alternatively, we could change the behavior as suggested [above](https://github.com/prettier/prettier/pull/6450#discussion_r320703051) and always add the newline after the `as` operator, and not check this at all.

thorn0: > always add the newline after the as operator

yes, please

thorn0: > I think formatting generics should be treated separately from this PR

It's non-wrapping of `as` that causes #5554. So this PR is going to fix it.

tests/typescript_as/__snapshots__/jsfmt.spec.js.snap
thorn0: I think it's okay.

src/language-js/printer-estree.js
thorn0: As for the parentheses, I'm just saying it's probably a bug that Prettier doesn't remove them. But it's out of scope for this PR, sure.

kaicataldo: The downside of always adding the newline after the `as` operator is that instead of

```ts
const value3 = thisIsAReallyLongIdentifier as {
  prop1: string;
  prop2: number;
  prop3: number;
}[];
```
we would have
```ts
const value3 = thisIsAReallyLongIdentifier as 
  {
    prop1: string;
    prop2: number;
    prop3: number;
  }[];
```

Which isn't a style I think is very common.

thorn0: Yep, this doesn't look good.

thorn0: Overall, I expected `as` to behave the same way as other binary operators. But turnes out other operators add even more line breaks:

**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEA3AhgGwFc4BmTAXkxgAsBLNASTQEEAlOA-ATwBloBzBgBMEMWgDNacAE6YA1AB0omTMCUqVAB2kRNARiSY9AGnUbtugEyHLp5eZ2aShkmYC+IYyF1j0yULjSOgDuAAqBCGjIILjYELRCniAARtK4YADWcDAAyprptFD8yDDSxF7UMAC2+ADqdPBo+WBwOZG0YtgdXNFgaFFehWgyMKFp-FW4yOIEw14AVmgAHgBCaZnZObhVcDyFcNOzcAvLOYX8+HAAioQQ8If4cyD50sPS0cm4yVz40EnahRgtQSNGQAA4AAxeCzDWppTTRbRwN7YA5eACOt3gY10URQuDQAFooHA4CJEl5pHBMbQqWNcBMpkgZo9jiBhlVaCUymy0OdLjc7gdmUcvDAvsChKCkLYQKVcLR8OcAMIQKqTaJQaBokCEYYAFS+eJZwzcbiAA)
```jsx
const value3 =
  thisIsAReallyLongIdentifier +
  {
    prop1: 1,
    prop2: 2,
    prop3: 3
  };

```

kaicataldo: I think it makes sense to treat it the same as other binary operators. I'll explore that and we can see what the team thinks.

kaicataldo:
After thinking about this some more, I took a completely different approach. It feels like we should treat `as` operators similarly to any other binary operator. By postprocessing the `TSAsExpression` to match the interface of a `BinaryExpression`, we can use the exact same logic. Thoughts?

kaicataldo:
I'm having trouble recreating these test failures locally. Any idea what might be going on? I'm running them on Node v10.16.3 in macOS v10.14.6.

src/language-js/postprocess.js
```diff
@@ -2,17 +2,21 @@
...
+      // Fix unexpected locEnd caused by --no-semi style
       case "VariableDeclaration": {
         const lastDeclaration = getLast(node.declarations);
         if (lastDeclaration && lastDeclaration.init) {
           overrideLocEnd(node, lastDeclaration);
         }
         break;
       }
+      case "TSAsExpression": {
+        convertTSAsExpression(node);
```
lydell: Feels pretty hacky to me, but I'm not sure how to do it better. There's quite a bit of code for printing binary expressions. On the other hand – do `as` expressions really need to go through _all_ of the binary expression stuff, like the JSX related parts? Maybe there’s a way to extract or copy the parts needed for `as`?

@vjeux You have had useful insight before on PRs with hacky code – what does your experience tell you here (gut feeling)?

kaicataldo: > Maybe there’s a way to extract or copy the parts needed for as?

My thought here was that keeping the two in sync would be a bigger maintenance burden, but if that's the preferred way of doing things, I can definitely look into it!

I'm also coming from working on babel-eslint, where the parser has to do a lot of massaging of the AST to get it to work with ESLint.

thorn0: Does it modify the passed AST itself or a copy of it? What if someone else wanted to use that AST after Prettier?

thorn0:
> I'm having trouble recreating these test failures locally. Any idea what might be going on? I'm running them on Node v10.16.3 in macOS v10.14.6.

Do you run the tests with `AST_COMPARE=1`?

src/language-js/postprocess.js
thorn0: > do as expressions really need to go through all of the binary expression stuff, like the JSX related parts

Why not? `as` can be used anywhere, just like `+` or say `&&`,

kaicataldo:
@thorn0 Ah, thanks! That's what I was missing!

src/language-js/postprocess.js
kaicataldo: Happy to work on this again once the team decides what the preferred approach is!

kaicataldo:
Friendly ping for the Prettier team - any insight on how we'd like to proceed here?

evilebottnawi:
@kaicataldo need fix CI problems

thorn0:
I don't represent the Prettier team, but I expect them to support the following points:
1. formatting `as`-expressions like binary expressions makes total sense, but
2. modifying the AST is not okay (e.g. see [this case](https://github.com/prettier/prettier/pull/6496#discussion_r325762190)).

kaicataldo:
@thorn0 @lydell Thanks, that should unblock me. I'll get to this ASAP!

kaicataldo:
I'm revisiting this tonight, and it looks like our options are either to:

1. Clone the node and modify it to match the `BinaryExpression` interface before printing.
1. Copy over the current printing `BinaryExpression` logic and modify it to work with the current `TSAsExpression` node interface.

The challenge of the first strategy is that we have to convert the `TSAsExpression` node every time it might exist (which is lots of places! e.g. [here](https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L619)), which adds a lot of complexity throughout the printer.

The challenge of the latter is that we'll have a fork in that logic and will have to make future changes in both code paths. This feels more feasible, but still doesn't feel great to me.

Thoughts?

thorn0:
Might it make sense to replace all occurrences of `.operator`, `.left` and `.right` with function calls (`getOperator`, etc)?

duailibe:
@kaicataldo I think we should start with option 2 and then work from that to minimize the problems with forking later

thorn0:
On the playground for this PR, I see that `as` is not formatted as other binaries (e.g. `+` or `**`), at least in this specific case (from #6623):

**Prettier pr-6450**
[Playground link](https://deploy-preview-6450--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuc0DOMAEBLeAnARkwF5Mw84BDeASX2ojwAoYALbNAOkljwgBt+cPABoy7fgBMxbDpxiUA5pkppMAZQCeUMADEArjpjZoAYQgBbAA7QEMAJQBuADo70WXMIBMJMhWpwdMIMzGhw-ABm3NAwfILCYmAS0phhkfJKKmpaOgZGJlDm1rawTq6uPBg4+ESk5FS09DCMLOxcPLECQqLi2FIybRnKAFTDGtp6hmDGZpY2UHZlblBVnng+df6Nwc2h4VEdcd2JyWJpUQojYzmT+bPFC6UuUCAiIBBWMyvIoJR4fAB3AAKfwQaGQIEoADcINhJK8QAAjPCUMAAazgMHUVlR2CgimQsX0cDerBgFn4AHV2PA0DiwHB1GDcNgobhNBCwGhwW88WE8DAgSjFBZKMgIpR+GE3gArNAADwAQij0Zj1JQLHAADJ4uDiyXSkBy+XqPGKIQARX0EHg+qlJJAOLw-IhME0VjgaHI2E+CKseDxMEpcLYyAAHAAGN7+iBhSkoqwQ-2e4RQvVvACO1vgQo+4JQqgAtAs4JJSwiKFnsBQhUpRXbDWELNhCXhiW80GbLdm9UgJfa3gpEcHJKGkF5Byi+maivWUFBbAj9GEACqURH5-uGqHEmhQMuwdTez4AQX36jdQgbcAAvjegA)
```sh
--parser typescript
```

**Input:**
```tsx
const iter1 = createIterator(this.controller, child, this.tag as SyncFunctionComponent);
const iter2 = createIterator(self.controller, child, self.tag as SyncFunctionComponent);

const iter1 = createIterator(this.controller, child, this.tag ** SyncFunctionComponent);
const iter2 = createIterator(self.controller, child, self.tag ** SyncFunctionComponent);

```

**Output:**
```tsx
const iter1 = createIterator(this.controller, child, this.tag as
  SyncFunctionComponent);
const iter2 = createIterator(self.controller, child, self.tag as
  SyncFunctionComponent);

const iter1 = createIterator(
  this.controller,
  child,
  this.tag ** SyncFunctionComponent
);
const iter2 = createIterator(
  self.controller,
  child,
  self.tag ** SyncFunctionComponent
);

```

kaicataldo:
Sorry, haven’t been able to get back to this. Will close this for now so that someone else can work on it.

Issue #6344, opened 2019-07-31
[BUG] No matching files on filenames with brackets


zomars:
Moved [from `lint-staged` issues](https://github.com/okonet/lint-staged/issues/676).

### Description

I'm using `lint-staged` with `husky` and the `prettier` CLI.

This is related to [Next.js new dynamic routes feature.](https://github.com/zeit/next.js/issues/7607)

### Steps to reproduce

1. Use this configuration in `package.json`:

```json
  …
  "husky": {
    "hooks": {
      "pre-commit": "lint-staged --debug"
    }
  },
  "lint-staged": {
    "*.{js,css,json,md}": [
      "prettier --write",
      "git add"
    ],
    "*.js": [
      "npm run lint:eslint",
      "git add"
    ]
  …
```

2. Try to commit a filename with brackets `[]`.

### Debug Logs

<details>

<summary>expand to view</summary>

```

husky > pre-commit (node v12.6.0)
2019-07-30T23:23:07.320Z lint-staged:bin Running `lint-staged@8.1.7`
2019-07-30T23:23:07.697Z lint-staged:find-bin Loaded package.json using `process.cwd()`
2019-07-30T23:23:07.847Z lint-staged Loading config using `cosmiconfig`
2019-07-30T23:23:07.852Z lint-staged Successfully loaded config from `OMMITED/package.json`:
{
  '*.{js,css,json,md}': [ 'prettier --write', 'git add' ],
  '*.js': [ 'npm run lint:eslint', 'git add' ]
}
2019-07-30T23:23:07.853Z lint-staged:cfg Normalizing config
2019-07-30T23:23:07.855Z lint-staged:cfg Validating config
Running lint-staged with the following config:
{
  linters: {
    '*.{js,css,json,md}': [
      'prettier --write',
      'git add'
    ],
    '*.js': [
      'npm run lint:eslint',
      'git add'
    ]
  },
  concurrent: true,
  chunkSize: 9007199254740991,
  globOptions: {
    matchBase: true,
    dot: true
  },
  ignore: [],
  subTaskConcurrency: 1,
  renderer: 'verbose',
  relative: false
}
2019-07-30T23:23:07.863Z lint-staged:run Running all linter scripts
2019-07-30T23:23:07.863Z lint-staged:run Resolved git directory to be `OMMITED`
2019-07-30T23:23:07.884Z lint-staged:run Loaded list of staged files in git:
[...]

PR #6492, opened 2019-09-17
Add --noglob cli option for processing exact the passed files

AzazKamaz:
I faced problem with configuring git hooks with [lint-staged](https://github.com/okonet/lint-staged) on sapper framework. lint-staged starts prettier and passes changed files with arguments, but in [sapper](https://sapper.svelte.dev) there are routes with name `[name].svelte`, and prettier thinks that is is a regexp. So I have added parameter (--noglob) that disables any logic on filenames and just keeps it as you write, so it will process exactly what you have passed


AzazKamaz:
With this option, prettier will work like [eslint](https://eslint.org). There are no such logic with glob)

kachkaev:
This CLI option should be also helpful to those using dynamic routing in Next.js https://github.com/zeit/next.js#dynamic-routing. The PR resolves https://github.com/prettier/prettier/issues/6344.

Instead of tweaking Prettier, the solution might actually come from https://github.com/okonet/lint-staged/pull/698 (see discussion in https://github.com/okonet/lint-staged/issues/676). The benefit in that approach is that we don't need any extra CLI flags in Prettier, which is great.

If this PR gets accepted, it might be worth updating https://prettier.io/docs/en/precommit.html as well.

AzazKamaz:
> Instead of tweaking Prettier, the solution might actually come from [okonet/lint-staged#698](https://github.com/okonet/lint-staged/pull/698) (see discussion in [okonet/lint-staged#676](https://github.com/okonet/lint-staged/issues/676)). The benefit in that approach is that we don't need any extra CLI flags in Prettier, which is great.

If [okonet/lint-staged](https://github.com/okonet/lint-staged) will provide escaped filenames, than [eslint/eslint](https://github.com/eslint/eslint) will not find them, because there are no globbing on [] and it will try to find file like `\[test\].js`

alexerhardt:
I'm using Next.js with a Prettier file watcher and this feature would be a lifesaver for me. As it stands dynamic routes throw errors on save, and I can't see an easy solution to this problem. Thanks for the effort @AzazKamaz - hope it gets accepted.

lydell:
Random thought: Maybe it would even be nice if Prettier didn’t support `[abc]` syntax at all. I’ve never used it. My globs are all `{a,b,c}`, `*.js` and `src/**/*.css`. ¯\\\_(ツ)\_/¯

evilebottnawi:
Same thread in stylelint https://github.com/stylelint/stylelint/pull/4286, here eslint solution https://github.com/eslint/eslint/blob/d08683e3c807f92daf266894093c70f8d5ac6afa/lib/cli-engine/file-enumerator.js#L292, i think we should use it

lydell:
@evilebottnawi Ah – first treat as literal file name, and if no such file exists then treat as glob? Sounds good to me.

(We could potentially end up in the reverse situation – a file on disk prevents someone’s glob to be treated as a glob, but that should be workaround-able by adjusting the glob pattern.)

lydell:
This is now tracked in #6854.

PR #6519, opened 2019-09-25
Support typescript 3.6

fisker:


evilebottnawi:
What is status?

fisker:
There is a  problem I can't fix

```typescript
type A = & string;
```
this code will format into

```typescript
type A = string;
```
the previous snapshot is also the same  but new version parser parse different ast cause failure test

I tried to update the snap to keep the ast match, but I can't figure a  way to detect

lydell:
What does `type A = & string;` mean in TypeScript?

fisker:
I'm not be familiar with Typescripts types, after a few search, I believe

```typescript
type A = & string;
```

is not equal to

```typescript
type A = string;
```

with `&` leading operator, it's Intersection Type, otherwise it is not, this is the reason AST compare fails, if this is currect, I might fix a possible bug.

We might want some one familiar with this to review

fisker:
why CI not running?

lydell:
Wouldn’t it be nicer if we cleaned `& foo` (and `| foo`) to just `foo`? I get that since the TypeScript update it technically changes the AST, but we already have small deviations when it makes sense (such as for JSX). You should be able to make the tests pass by updating `src/language-js/clean.js`. It already has a couple of TypeScript cases.

fisker:
@lydell 

I really don't know those things, but according to the spec `& foo` `| foo` is different type with `foo`, so we might also add `| foo` test, and keep the leading `|`, maybe after this line https://github.com/prettier/prettier/blob/fced57c1d68971407481247f35fd2c6b329fce4e/tests/typescript/conformance/types/union/unionTypeEquivalence.ts#L4

fisker:
I mean `foo` `& foo` `| foo` has different AST, let me try to clean

fisker:
Good to go

src/language-js/clean.js
```diff
@@ -65,6 +65,14 @@ function clean(ast, newObj, parent) {
     return newObj.typeAnnotation;
   }
 
+  // (TypeScript) `& foo` and `| foo` into `foo`
```
lydell:
```suggestion
  // (TypeScript) Bypass `& foo` and `| foo` into `foo`
  // https://github.com/microsoft/TypeScript/issues/30995
```

scripts/build/config.js
```diff
@@ -42,6 +42,12 @@ const parsers = [
     replace: {
       // node v4 compatibility for @typescript-eslint/typescript-estree
       "(!unique.includes(raw))": "(unique.indexOf(raw) === -1)"
+    },
+    commonjs: {
+      ignore: [
+        // optional typescript etw logger
```
lydell:
```suggestion
        // Optional package for TypeScript that logs ETW events (a Windows-only technology).
```

package.json
```diff
@@ -129,7 +129,7 @@
     "lint": "cross-env EFF_NO_LINK_RULES=true eslint . --format friendly",
     "lint-docs": "prettylint {.,docs,website,website/blog}/*.md",
     "lint-dist": "eslint --no-eslintrc --no-ignore --env=browser \"dist/!(bin-prettier|index|third-party).js\"",
-    "build": "node --max-old-space-size=2048 ./scripts/build/build.js",
+    "build": "node --max-old-space-size=3072 ./scripts/build/build.js",
```
lydell: Does the build fail without increasing this?

fisker: yes, not enough memory

fisker: see https://dev.azure.com/prettier/prettier/_build/results?buildId=1672

lydell: Ouch. I wonder what’s taking so much memory…

PR #6545, opened 2019-09-27
Update `unified` to v8

fisker:


fisker:
rebased

fisker:
Actually this module do not has a node version requirement, but one of the dependency `is-plain-obj` requires node 8, https://github.com/sindresorhus/is-plain-obj/blob/master/package.json#L13

This is the problem I said in https://github.com/prettier/prettier/pull/6224#issuecomment-535930243 , it's not that simple to check than you image 

@evilebottnawi

fisker:
seems nobody cares about #6373 , but that the real solution, even after we drop node 6 in 2.0.0, we can still test against node 8 node 10

PR #6549, opened 2019-09-27
Update `jest-junit` to v8

fisker:


evilebottnawi:
@fisker please rebase

fisker:
On bed already... booting my laptop

PR #6550, opened 2019-09-27
Update `jest-watch-typeahead` to v0.4.0

fisker:


fisker:
rebased

evilebottnawi:
Can't be merged https://github.com/jest-community/jest-watch-typeahead/releases/tag/v0.4.0

fisker:
@evilebottnawi add a tag ? we can merge when we start v2.0

evilebottnawi:
@fisker I'm sure it will require rebase and you can unavailable so the PR will be stale, also before release 2.0.0 we update all deps too so i think better close this

PR #6561, opened 2019-09-27
Update `esutils` to v2.0.3

fisker:


lipis:
we should add dependabot for this project after all these updates.

fisker:
I can do this all day :)

PR #6548, opened 2019-09-27
Update `execa` to v2

fisker:

PR #6563, opened 2019-09-28
Update `execa` to v1

fisker:
see https://github.com/prettier/prettier/pull/6548#issuecomment-536205661


fisker:
@evilebottnawi  Please close #6548 when merge this one

PR #6547, opened 2019-09-27
Update `cross-env` to v6

fisker:

PR #6564, opened 2019-09-28
Update `cross-env` to latest v5

fisker:
Replace #6547


fisker:
please close #6547 when merge this one

PR #6540, opened 2019-09-27
Update `mem` to v5

fisker:

PR #6565, opened 2019-09-28
Update `mem` to latest v4

fisker:
replace #6540


fisker:
please close #6540 when merge this one

PR #6570, opened 2019-09-30
[WIP] Update `@glimmer/syntax` to v0.43.0

fisker:
https://github.com/prettier/prettier/pull/6224#issuecomment-536538898


evilebottnawi:
looks like regression on parser side

package.json
```diff
@@ -13,6 +13,9 @@
   "engines": {
     "node": ">=8"
   },
+  "resolutions": {
+    "simple-html-tokenizer": "0.5.7"
+  },
```
evilebottnawi: It is bad idea

fisker: wait, explaining

evilebottnawi: @fisker we need find a problem and report about this and wait a fix, no need change resolutions

fisker:
Tell you guys a funny story, when I look into this, I found something interesting

When reinstall `@glimmer/syntax`, one of it's dependency `simple-html-tokenizer`, will resolve to `v0.5.8` because this version requirement is not pinned https://github.com/glimmerjs/glimmer-vm/blob/0fd66816d336bbb3ab1b9ac34365dce8721a83c7/packages/%40glimmer/syntax/package.json#L9 . 

and `simple-html-tokenizer@v0.5.8` changed how it treat `<title>`, `<script>`, and `<style>` [ref](https://github.com/tildeio/simple-html-tokenizer/releases/tag/v0.5.8)

two of our handlebars test 
- https://github.com/prettier/prettier/blob/master/tests/handlebars-whitespace/display-none.hbs
- https://github.com/prettier/prettier/blob/master/tests/handlebars-basics/basic-handlebars.hbs

fails.

But our test is fine with `handlebars`.

And this is the reason why #6373 fails, in fact, direct-install-from-github also broken.

you can run npm install prettier/prettier, and format those two file, they will not get formatted.

for `yarn`, we can use "resolutions", I'm not sure how to fix this for `npm`, maybe add to `dependencies`?

fisker:
here is a discussion on `yarn.lock` I found tody
 https://github.com/davidtheclark/cosmiconfig/issues/206

out test pass, but actually the production(I mean direct-install-from-github) is already broken

package.json
fisker: see comment bellow

brodybits:
Interesting. Does the issue need to be fixed on `simple-html-tokenizer`?

package.json
```diff
@@ -13,11 +13,14 @@
...
   "dependencies": {
     "@angular/compiler": "8.2.13",
     "@babel/code-frame": "7.5.5",
     "@babel/parser": "7.7.0",
-    "@glimmer/syntax": "0.41.0",
+    "@glimmer/syntax": "0.43.0",
```
brodybits: Would you mind updating the title to match this change?

rwjblue:
👋 - Taking a look at this PR after @brodybits commented in https://github.com/tildeio/simple-html-tokenizer/pull/69 (we should still create a separate issue to track the path forward).

I'm assuming that the linked input files (from https://github.com/prettier/prettier/pull/6570#issuecomment-550266720) emit a different output after the update, but the CI output is fairly large so I'm not sure what specifically the change in `simple-html-tokenizer` is doing to the output here. Any pointers for quickly reproducing the failures (is it as simple as `npm i && npm test` here)?

---

From the perspective of that library, the change is mostly correct (though I agree that it probably should have been a major version bump in retrospect, sorry about that) for parsing HTML since `<title>`/`<script>`/`<style>` ([per the HTML spec](https://html.spec.whatwg.org/multipage/scripting.html#restrictions-for-contents-of-script-elements)). I think we might be able to add a configurable option for that change so that when in "codemod mode" (which is what codemods like prettier use with `@glimmer/syntax`) we could continue to attempt to parse, but it isn't 100% obvious that we should make assumptions about the contents of these elements even there...

package.json
fisker: was trying other versions, its broken anyway,

fisker:
@rwjblue you could run `npm install  prettier/prettier`  then `npx prettier <file>` on  those two file and compare with the snap in there dir

fisker:
@brodybits thanks for help me with this

fisker:
Can't find a solution without pinning `simple-html-tokenizer` to `v0.5.7`, closing this.

PR #6575, opened 2019-09-30
Update `unicode-regex` to v3

fisker:


ikatyang:
I marked it as breaking change since some Unicode categories are removed. Since we did not using those removed categories and [the CommonMark spec is defined using the general Unicode categories](https://spec.commonmark.org/0.29/#punctuation-character), updating to the newest Unicode standard should be safe.

Issue #5936, opened 2019-03-02
[Styled-components] keyframes are ignored


rodrigojcmello:
**Prettier 1.16.4**
[...]

Issue #4424, opened 2018-05-05
Support library-specific plugin


ikatyang:
from https://github.com/prettier/prettier/pull/4423#pullrequestreview-117811698:

> I think that we should start to figure out a way to do those kind of library-specific kind of transforms into plugins.
> 
> There are a lot of downsides with the current approach of bundling it inside of prettier:
> - Now we need to maintain all those ourselves
> - If jest changes their format, then you need to wait a new prettier release to be able to use it. This dramatically slow down iteration speed
> - Ideally jest would be in control of how their format their templates
> - We have a good relationship with jest but what if some library coming from someone else wanted to format their thing, we now have to figure out what specific library printing is worth being inside of prettier.

ikatyang:
I think we can have a `printerPlugin` entry in a plugin:

```ts
// some-plugin.js
module.exports = {
  printerPlugin: {
    // astFormat
    estree: (path, print, textToDoc, options) => Doc | null;
  }
}
```

which should work something like [`embed()`](https://prettier.io/docs/en/plugins#printers), i.e. potentially switch to a different printerPlugin in every node:

https://github.com/prettier/prettier/blob/1a6cf3d0717c5828d9bf474d13d115e3150ee254/src/main/ast-to-doc.js#L83-L99

suchipi:
I'm super in favor of this; it would also solve https://github.com/prettier/prettier/issues/4260 and let users create plugins on top of prettier for the places we don't want to add options

j-f1:
For example, someone could make a `JSXAttribute` plugin that prints string values with single quotes.

j-f1:
One thing we’ll have to figure out is what to do if two plugins try to customize the printing of a node.

suchipi:
Hyperscript and mithril would be good candidates for llibrary plugins; see https://github.com/prettier/prettier/issues/4424

j-f1:
@suchipi looks like that’s this issue 😕

bakkot:
Drive-by comment - while I think something like this would be a good idea, especially for things like formatting tagged template literals, I'm worried that a really general approach to plugins will mean that people will just start implementing configuration as plugins, which I think would eliminate a lot of [value to end users](https://github.com/prettier/prettier/issues/40) of prettier's conservative approach to options: i.e., people would start fighting about configuration again, just now in the form of "add this third-party plugin to format this really specific thing the way I want it to be formatted" instead of "add this config option to [etc]".

[...]

PR #6626, opened 2019-10-08
Add the 'jsCustomTags' option

MeLlamoPablo:

==========================

Hello!

First of all I want to say that I realize that this PR will be controversial. I have read the [opinion philosophy][1] and #40.

This PR adds a new option `jsCustomTags` with the following form:

```json
{
  "jsCustomTags": {
    "styled-components": ["media"]
  }
}
```

I believe that this option is necesary and must make it into prettier (or any similar option that cover this use case). I understand that we want to keep prettier as opinionated as possible, but in this case this is not a matter of opinion, this is a matter of supporting an use case.

In our code, we use `styled-components`, and we have a custom helper to do media queries:

```ts
const MyComponent = styled.div`
    max-width: 700px;

    ${media.mobile`
        max-width: 90%;
    `};
`;
```

However, in another one of our projects, we do:

```ts
const MyComponent = styled.div`
    max-width: 90%;

    ${from.mobile`
        max-width: 700px;
    `};
`;
```

We use `from` instead of `media` because our queries are mobile-first, and they are more descriptive that way. We are not the only ones to use this pattern: take a look at [this article][2] or [this package][3].

The point here is that we can't just let prettier decide that the only valid tag names for `styled-components` are `styled` and `css`. There's an open issue about supporting `keyframes` too (#5936), but I believe that opening a PR just for `keyframes` is missing the point. What we need to move towards is a customizable list of allowed tags, in which `styled` and `css` are the default, but can also be extended.

There is no standard way to define media queries in styled-components (neither in `styled-components` itself nor there is a well established standard in the community - like [fsa][4] for redux), and while there isn't, we can't have prettier decide which names are acceptable and which aren't.

This not only applies to `styled-components`, but to any potential library that relies on tagged template strings like `graphql`; hence why I named the option `jsCustomTags` and not `styledComponentCustomTags`.

Some final notes:

* I opened this PR knowing that there's a good chance that it won't be merged. I did it instead of opening an issue because I wanted to present a working prototype, and because we can continue to use our own workflow without giving up auto formatting inside media queries.
* I am aware that some tests broke because of this PR, mostly snapshot of configuration help. I didn't want to bother updating them without knowing if this was going to be merged, and if it is, there's a good chance that the API won't be exaclty what I propose, so the tests can be updated later.

Thanks a lot for your time.

[1]: https://prettier.io/docs/en/option-philosophy.html
[2]: https://medium.com/@samuelresua/easy-media-queries-in-styled-components-690b78f50053
[3]: https://github.com/morajabi/styled-media-query
[4]: https://github.com/redux-utilities/flux-standard-action

### DIFF
--- a/scripts/generate-schema.js
+++ b/scripts/generate-schema.js
@@ -106,6 +106,8 @@ function optionTypeToSchemaType(optionType) {
       );
     case "path":
       return "string";
+    case "any":
+      return "object";
     default:
       throw new Error(`Unexpected optionType '${optionType}'`);
   }
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -28,7 +28,7 @@ function embed(path, print, textToDoc, options) {
         isStyledComponents,
         isCssProp,
         isAngularComponentStyles
-      ].some(isIt => isIt(path));
+      ].some(isIt => isIt(path, options));
 
       if (isCss) {
         // Get full template literal with expressions replaced by placeholders
@@ -431,7 +431,7 @@ function getAngularComponentObjectExpressionPredicates() {
 /**
  * styled-components template literals
  */
-function isStyledComponents(path) {
+function isStyledComponents(path, options) {
   const parent = path.getParentNode();
 
   if (!parent || parent.type !== "TaggedTemplateExpression") {
@@ -440,13 +440,20 @@ function isStyledComponents(path) {
 
   const tag = parent.tag;
 
+  const customTags = options.jsCustomTags["styled-components"] || [];
+
+  function isCustomIdentifier(node) {
+    return node.type === "Identifier" && customTags.includes(node.name);
+  }
+
   switch (tag.type) {
     case "MemberExpression":
       return (
         // styled.foo``
         isStyledIdentifier(tag.object) ||
         // Component.extend``
-        isStyledExtend(tag)
+        isStyledExtend(tag) ||
+        isCustomIdentifier(tag.object)
       );
 
     case "CallExpression":
@@ -466,7 +473,7 @@ function isStyledComponents(path) {
 
     case "Identifier":
       // css``
-      return tag.name === "css";
+      return tag.name === "css" || customTags.includes(tag.name);
 
     default:
       return false;
--- a/src/language-js/options.js
+++ b/src/language-js/options.js
@@ -24,6 +24,14 @@ module.exports = {
     ]
   },
   bracketSpacing: commonOptions.bracketSpacing,
+  jsCustomTags: {
+    since: "1.18.2",
+    category: CATEGORY_JAVASCRIPT,
+    type: "any",
+    default: {},
+    description:
+      "A map of languages and custom template literal tag names. Example: { 'styled-components': ['media'] }"
+  },
   jsxBracketSameLine: {
     since: "0.17.0",
     category: CATEGORY_JAVASCRIPT,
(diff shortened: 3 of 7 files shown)


evilebottnawi:
Thanks.

Just notes:
- we should allow to define `foo.bar.baz.other.name` and any deep

What about?
```js
{
  "jsCustomTags": ["styled.div", "styled.keyframes", "from.mobile"]
}
```

- default value should handle all currently supported tagged template expressions to avoid breaking change 

Bonus: Also it can solve problems when we break code in tagged template expressions, after this PR it should be easy to fix, just setup `jsCustomTags` to `[]`.

In fact, we should consider whether we should support formatting template literal or not. 

Here two solutions:
- drop support formatting template literal and implement https://github.com/prettier/prettier/issues/4424 (for `2.0` version)
- support some of them out of box and merge this PR

/cc @prettier/core will be great to get some feedback

thorn0:
Alternatively, why not try to format **all** tagged template literals nested to `styled.div` as CSS?

MeLlamoPablo:
> What about?
>
> ```
> {
>    "jsCustomTags": ["styled.div", "styled.keyframes", "from.mobile"]
> }
> ```

@evilebottnawi, but how would you distinguish between `graphql`, `styled-components` and my own custom DSL that I just wrote a plugin for? I think that it sucks having to deal with a nested structure like that but I don't see any other alternative.

> Alternatively, why not try to format all tagged template literals nested to styled.div as CSS?

@thorn0, that's an interesting idea indeed, but it would prevent me from doing:

```typescript
const responsiveStyles = media.mobile`
    max-width: 90%;
`;

const MyComponent = styled.div`
    max-width: 700px;
    ${responsiveStyles};
`;
```

Personally I think that's confusing so I don't do it, but it's possible so I'm guessing it should be supported too.

evilebottnawi:
@MeLlamoPablo

> but how would you distinguish between graphql, styled-components and my own custom DSL that I just wrote a plugin for? I think that it sucks having to deal with a nested structure like that but I don't see any other alternative.

Maybe support also this:
```
  "jsCustomTags": ["...", "from.mobile"]
```

`...` means use all default tags + own

@thorn0 yep, but here more fundamental problem, in future style-components can implement more helpers and for us it is not easy maintenance all supported tagged template literals

MeLlamoPablo:
@evilebottnawi sorry I didn't explain myself very well. My point is that in this example, we want to parse the content `from`-tagged strings with `styled-components` syntax:

```typescript
const MyComponent = styled.div`
    max-width: 90%;

    ${from.mobile`
        max-width: 700px;
    `};
`;
```

But let's assume that I create a DSL that allows to specify a TCP origin (weird example I know, but let's assume that we need a DSL for that and I write a prettier plugin so it can be properly parsed). We'd have:

```typescript
import { from } from 'my-custom-dsl';

const remote = from`
    protocol ssh origin example.com port 22
`;
```

And we'd want:

```typescript
import { from } from 'my-custom-dsl';

const remote = from`
    protocol ssh
    origin example.com
    port 22
`;
```

So, how will prettier know if it needs to format `from`-tagged strings as `styled-components` or as `my-custom-dsl`? I'd need to specify either:

```json
{
  "jsCustomTags": {
    "styled-components": ["from"]
  }
}
```

or

```json
{
  "jsCustomTags": {
    "my-custom-dsl": ["from"]
  }
}
```

lydell:
I think it might be worth thinking about #5588 when designing this.

JounQin:
We should merge some options to be a nested object IMO.

brandonkal:
I would like to see yaml tagged template support. Prettier already supports both languages standalone, so ideally this PR would be made more generic.

nickmccurdy:
Would this functionality be covered by #5588?

JounQin:
This should already be covered by https://github.com/Sec-ant/prettier-plugin-embed?

cc @fisker

PR #6639, opened 2019-10-12
Update `globby` to v7.1.1

fisker:
~with `globby` v7.1.1, we support `dot pattern` and `expandDirectories` now.~

~fixes: #6085~


evilebottnawi:
> Added expandDirectories option which is on by default. The nodir option is now also true by default. This means that by default it will now only return file paths, not directory paths. cf55e5c

Breaking change

lydell:
Sounds like we need to add some tests, then.

fisker:
> Looks like breaking changem try to use `.` pattern

I don't think this `expandDirectories` and `noDir` effect our code, we only use it to find file

Look for file

https://github.com/prettier/prettier/blob/f8a4a61a6265ff398d817eece67ded0c661b473d/src/cli/util.js#L412-L414


Look for `package.json`

https://github.com/prettier/prettier/blob/3805034bdfe45722e31a1abf019d4368dfb8ae11/src/common/load-plugins.js#L97-L104

Look for test file

https://github.com/prettier/prettier/blob/6c482e2e6c4d771407302fcc999bb3e2b491cbd3/scripts/sync-flow-tests.js#L32-L35

Look for test file

https://github.com/prettier/prettier/blob/4cc99242a886e885853887ee670b21a4698cac1f/scripts/run-external-tests.js#L24

src/cli/util.js
```diff
@@ -410,7 +410,7 @@ function eachFilename(context, patterns, callback) {
 
   try {
     const filePaths = globby
-      .sync(patterns, { dot: true, nodir: true })
+      .sync(patterns, { dot: true, expandDirectories: false })
```
lydell: Is it possible to add tests for this?

scripts/run-external-tests.js
```diff
@@ -21,7 +21,9 @@ function tryFormat(file) {
 }
 
 function runExternalTests(patterns) {
-  const testFiles = globby.sync(patterns);
+  const testFiles = globby.sync(patterns, {
+    expandDirectories: false
+  });
```
brodybits: Do we really need `expandDirectories: false` here or in the other script?

src/cli/util.js
brodybits: I would also say this needs a test.

src/common/load-plugins.js
```diff
@@ -100,7 +100,7 @@ function findPluginsInNodeModules(nodeModulesDir) {
       "@*/prettier-plugin-*/package.json",
       "@prettier/plugin-*/package.json"
     ],
-    { cwd: nodeModulesDir }
+    { cwd: nodeModulesDir, expandDirectories: false }
```
brodybits: Is `expandDirectories: false` really needed here?

src/cli/util.js
brodybits:
```suggestion
      .sync(patterns, { dot: true })
```

works for me when I try with globby@7.1.1 and globby@10.0.1

<https://github.com/prettier/prettier/pull/6639#issuecomment-542054241> indicates to me that there should be no need for `nodir: true` or `expandDirectories: false` here

brodybits:
I have a few more questions:

- Any explanation for the small loss of code coverage?
- Could this resolve #6085?
- Should we consider using globby@10 if we do not support install from GitHub on Node.js 6?

fisker:
I was trying to make tests the another day, then I got distracted by https://github.com/prettier/prettier/pull/6665/files .

Those options are really confusing me, as I understand, 

`nodir` means donot list dirname , since we  are looking for file  we should use default `true` 

`expandDirectories` also means that?

brodybits:
I think part of the confusion is how much globby has changed over quite a few major versions. I think we should try to research and understand this carefully before making any more adjustments. I do think we should also avoid using any options not really needed.

I am now really thinking we should consider using the latest rather than trying to deal with the old, unsupported package versions.

Unfortunately I don't have much time to look into this right now.

brodybits:
I tried raising RFC PR #6748 to try globby@10 update for discussion, seems to be mostly working with a few TODO items. I think that nodir setting is not needed in some newer versions of globby, removing it does not cause any failures. I am also thinking that expandDirectories should not be an issue since Prettier only works with files. But I may prove to be mistaken here.

fisker:
closing in favor of #6748

brodybits:
> closing in favor of #6748

Any ideas how we can resolve the TODO items in PR #6748?

src/cli/util.js
fisker: tests added

src/common/load-plugins.js
fisker: yes

fisker:
@lydell @evilebottnawi @brodybits 

I reopened this, and add `dot pattern` and `expandDirectories` tests.

This will not breaking anything, but support `prettier .` and `prettier src`, we can safely add this to 1.19

tests_integration/__tests__/patterns.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+const runPrettier = require("../runPrettier");
+
+expect.addSnapshotSerializer(require("../path-serializer"));
+
+// TODO: move `multiple-patterns` tests into this one
```
brodybits: Can you explain why we would want to do this in the future?

I would only favor combining tests if they have the exact same output and can help reduce the number of snapshots we have to maintain.

Assuming there would be a good reason to do this, I would really favor adding this comment to `tests_integration/__tests__/multiple-patterns.js` as well to avoid losing track of this TODO item.

brodybits:
Thanks @fisker for taking this work over and for all of your work on these updates in general. I can see that you have worked really hard based on our feedback on this and other proposals you have raised this year. I wish I could understand this better.

tests_integration/__tests__/patterns.js
fisker: not all tests are `multiple-patterns` , name to `patterns` should be more reasonable

lydell:
Supporting `dot pattern` would be such a good feature to get out in 1.19! :clap:

Can you add tests where there are multiple types of files in the directories passed to `prettier`? Such as .md, .css, .ts. Also some that Prettier does not support, like .png, .jpg.

fisker:
```bash
node ../../../bin/prettier.js . -l
```

result: 

```text
directory\file.js
directory\nested-directory\nested-directory-file.js
file.will-never-support[error] No parser could be inferred for file: file.will-never-support
other-directory\file.js
other-directory\nested-directory\nested-directory-file.js
other-regular-modules.js
other-supported-file.css
regular-module.js
```

@lydell is this good ?

lydell:
I pulled your branch and played a little with it.

```
❯ bin/prettier.js .stuff -l
.stuff/css/main.css
.stuff/css/values.txt[error] No parser could be inferred for file: .stuff/css/values.txt
.stuff/foo.png[error] No parser could be inferred for file: .stuff/foo.png
.stuff/readme.md
.stuff/test.js
.stuff/test.ts
```

I think that when you say `prettier some-dir/`, Prettier should only try to run on files that it supports inside that directory.

This is how I would personally be able to set up Prettier:

1. Create a `.prettierignore`. For example, if I don’t want to run Prettier on HTML files, I can add `*.html` in there.
2. Run `prettier --write .` when I want to format all the things.
3. Run `prettier --check .` in CI.

This would make it super easy to set things up, and my editor would automatically know which files to format on save (_all_ files, except those mentioned in `.prettierignore`).

brodybits:
I will also assume that "support `dot pattern` and `expandDirectories`" would continue to work if we would upgrade to globby@10 someday in the future, right?

fisker:
@brodybits yes, both `glob` and `fast-glob` support them

fisker:
@lydell 

currently, `bin/prettier.js .stuff -l` will behave same as `bin/prettier.js .stuff/** -l`, so there will be `[error] No parser could be inferred for file: .stuff/css/values.txt` printed, if we don't want this message, we probably want filter those files without parser, this will not print error for both patterns. but if `bin/prettier.js .stuff/**/*.png -l` will not get a error.

fisker:
@lydell 

I toke a look at the logic, might no easy to change, I suggest we merge this PR first. 

Consider this is a shortcut for  `prettier **/*` -> `prettier .` , `prettier src/**/*` -> `prettier src`

lydell:
I don’t think we should merge this. We don’t want to ship a broken `prettier .` implementation to users. It _feels_ broken to me, even if we justify the behavior. And if we _do_ ship it, changing its behavior later would be a breaking change?

fisker:
@lydell 

If I understand currectly, what do you want is  
`prettier . --write`
`prettier dir/ --write` 
`prettier . --list`
`prettier dir/ --list` 
not print `UndefinedParserError`, is that currect?

lydell:
Yes, that’s correct!

fisker:
1. `prettier dir1 dir2`
2. `prettier dir1 dir2/**/*`
3. `prettier non-exists-dir dir2/**/*`
4. `prettier . dir2/**/*`

Can you explain to me, what  those 4 should be?

lydell:
1. `prettier dir1 dir2` – prettify all files with supported extensions inside `dir1` and `dir2`.
2. `prettier dir1 dir2/**/*` – prettify all files with supported extensions inside `dir1` as well as all files matched by the `dir2/**/*` glob. If any of the latter files have unknown extensions – log an error for them.
3. `prettier non-exists-dir dir2/**/*` – log an error that `non-exists-dir` resulted in 0 files and prettify all files matched by the `dir2/**/*` glob. If any of the latter files have unknown extensions – log an error for them.
4. `prettier . dir2/**/*` – prettify all files in `.` and all files matched by the `dir2/**/*` glob. If any of the latter files have unknown extensions – log an error for them.

I basically want the same behavior as `eslint`.

fisker:
just to clearify not supported extensions, because parser can be override by config.

lydell:
For every file we go through, try to resolve the final config to use for that file like we already do. If `parser` is missing, throw an error if the file came from a full path or glob, and ignore the file if it came from a folder pattern. If a pattern (item in `process.argv`) results in 0 files (non-existent folder, empty folder, folder with no supported files, glob that does not match anything, non-existent file), warn about it resulting in 0 files.

brodybits:
> For every file we go through, try to resolve the final config to use for that file like we already do.
> [...]

Shouldn't this already be covered by the test suite?

If not, I think we should add the missing tests.

fisker:
I've disabled all the behavior changes, it's only package upgrade now.

evilebottnawi:
Also don't forget what plugins can add own extensions and it should be tested also (when we implement `.` support)

src/cli/util.js
```diff
@@ -407,11 +407,13 @@ function eachFilename(context, patterns, callback) {
     patterns = patterns.concat(["!**/node_modules/**", "!./node_modules/**"]);
   }
   patterns = patterns.concat(["!**/.{git,svn,hg}/**", "!./.{git,svn,hg}/**"]);
+  patterns = patterns.filter(pattern => pattern !== ".");
```
evilebottnawi: Let's add comment here why we do this

fisker: done

evilebottnawi:
@fisker please fix conflicts and let's do it for `next`

@prettier/core please review

fisker:
Rebased & CI passed.

fisker:
What are we going to do with this?

evilebottnawi:
@fisker If we can't update globby to latest we need merge this branch

fisker:
Closing in favor of #6776

PR #6400, opened 2019-08-18
 feat: add new parser `babel-ts` to parse TypeScript syntaxes via Babel

JounQin:
- [x] I’ve added tests to confirm my change works.
- [x] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [x] (If the change is user-facing) I’ve added my changes to the `CHANGELOG.unreleased.md` file following the template.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

Motivation: `Babel` itself is able to parse `TypeScript`, but we haven't enabled it, which makes something like [babel-preset-proposal-typescript](https://github.com/JounQin/babel-preset-proposal-typescript) can not integrate with `Prettier` perfectly, but it should be easy to support it.

PR #6660, opened 2019-10-15
enable the allowUndeclaredExports Babel option

thorn0:
Babel 7.4.0 introduced a [check for undefined exported bindings](https://github.com/babel/babel/pull/9589). It could break Prettier's range formatting, but luckily, they also added a [way to disable it](https://github.com/babel/babel/pull/9864) in 7.5.0.

No changes in the API or CLI. Not a user-facing change.

### DIFF
--- a/src/language-js/parser-babylon.js
+++ b/src/language-js/parser-babylon.js
@@ -16,6 +16,7 @@ function babelOptions(extraOptions, extraPlugins) {
       allowImportExportEverywhere: true,
       allowReturnOutsideFunction: true,
       allowSuperOutsideMethod: true,
+      allowUndeclaredExports: true,
       plugins: [
         "jsx",
         "doExpressions",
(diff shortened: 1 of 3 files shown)


thorn0:
I wish there also was a way to disable the [check for duplicate bindings](https://github.com/babel/babel/pull/9493). Because of it, we need to change too many tests in #6400. And overall, it feels like a breaking change for Prettier.

evilebottnawi:
@thorn0 sounds resonable

thorn0:
Babel 7.7.0 is going to get [error recovery](https://github.com/babel/babel/pull/10363), so that we will be able to ignore unneeded errors.

fisker:
@thorn0 Did't I replace them in https://github.com/prettier/prettier/pull/6386

thorn0:
@fisker I'm [trying](https://github.com/prettier/prettier/pull/6400/commits/c89b01dbe237b8cbb5af5c88c8f54541e2724a90#diff-70f3836f361e3398a527b37114b863a4R93) to enable `babel-ts` in the tests for `typescript`.

fisker:
I know what you are doing , when I update babel parser I renamed them all

thorn0:
You did it only in the tests `babel` was enabled for.

fisker:
Ok, haha. I didn't know that

PR #6665, opened 2019-10-16
Fix an expired todo test

fisker:

### DIFF
--- a/src/common/get-file-info.js
+++ b/src/common/get-file-info.js
@@ -19,6 +19,14 @@ const path = require("path");
  * internally by the method wrapper. See withPlugins() in index.js.
  */
 function getFileInfo(filePath, opts) {
+  if (typeof filePath !== "string") {
+    return Promise.reject(
+      new TypeError(
+        `expect \`filePath\` to be a string, got \`${typeof filePath}\``
+      )
+    );
+  }
+
   return createIgnorer(opts.ignorePath, opts.withNodeModules).then(ignorer =>
     _getFileInfo(
       ignorer,
@@ -34,6 +42,12 @@ function getFileInfo(filePath, opts) {
  * @returns {FileInfoResult}
  */
 getFileInfo.sync = function(filePath, opts) {
+  if (typeof filePath !== "string") {
+    throw new TypeError(
+      `expect \`filePath\` to be a string, got \`${typeof filePath}\``
+    );
+  }
+
   const ignorer = createIgnorer.sync(opts.ignorePath, opts.withNodeModules);
   return _getFileInfo(
     ignorer,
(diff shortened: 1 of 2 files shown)


fisker:
I put a specific error there, if someone changes getFileInfo, they can update the Error info

tests_integration/__tests__/file-info.js
```diff
@@ -106,16 +106,17 @@ describe("extracts file-info with inferredParser=foo when a plugin is hand-picke
...
 test("API getFileInfo with no args", () => {
-  // TODO: change this to `rejects.toThrow()` when we upgrade to Jest >= 22
-  // https://github.com/facebook/jest/issues/3601
-  expect.assertions(1);
-  return prettier.getFileInfo().catch(err => {
-    expect(err).toBeDefined();
-  });
+  expect(prettier.getFileInfo()).rejects.toThrow(
+    // this error info might change depends on how getFileInfo works
+    new TypeError(`Cannot read property 'replace' of undefined`)
```
lydell: Where does this error message come from?

fisker: see comment above.

for now its come from undefined.replace

too strict?

fisker: or maybe we should throw a error with specific info?

tests_integration/__tests__/file-info.js
```diff
@@ -106,16 +106,15 @@ describe("extracts file-info with inferredParser=foo when a plugin is hand-picke
 });
 
 test("API getFileInfo with no args", () => {
-  // TODO: change this to `rejects.toThrow()` when we upgrade to Jest >= 22
-  // https://github.com/facebook/jest/issues/3601
-  expect.assertions(1);
-  return prettier.getFileInfo().catch(err => {
-    expect(err).toBeDefined();
-  });
+  expect(prettier.getFileInfo()).rejects.toThrow(
```
SimenB: you need to `return` or `await` it - it's async

fisker: you are right, I forgot, fixed

fisker:
update to use return , it's more readable https://github.com/prettier/prettier/issues/6668

fisker:
update to use return , it's more readable https://github.com/prettier/prettier/issues/6668

src/common/get-file-info.js
```diff
@@ -19,6 +19,14 @@ const path = require("path");
  * internally by the method wrapper. See withPlugins() in index.js.
  */
 function getFileInfo(filePath, opts) {
+  if (typeof filePath !== "string") {
+    return Promise.reject(
```
evilebottnawi: Just use `throw new TypeError`, no need use `Promise.reject`

fisker: no, that would  be a breaking  change since thi function I not async

evilebottnawi: my mistake (i think it is async :smile: )

PR #6666, opened 2019-10-16
API: add resolveConfig option to getFileInfo()

### LINKED ISSUE #5566: `prettier.getFileInfo` doesn’t take `overrides` into account when determining `inferredParser`
I posted this issue in prettier-atom: https://github.com/prettier/prettier-atom/issues/460

”Format on save” isn’t working for files added via overrides in prettierrc like this:

```yml
overrides:
  - files: ['*.module', '*.theme']
    options:
      parser: php
```

According to @robwise this is an issue with Prettier itself (https://github.com/prettier/prettier-atom/issues/460#issuecomment-442170789) in that `prettier.getFileInfo` doesn’t take `overrides` into account when determining `inferredParser`.

**Environments:**
- Prettier Version: 1.15.2
- Running Prettier via: prettier-atom plugin (0.56.0) for Atom (1.32.2)
- Operating System: macOS

kaicataldo:
fixes #5566 

Took a stab at fixing this, since it seems like a useful feature for integrations. I added an option since it does potentially add FS operations, and it doesn't seem like a great thing to force everyone into if they're happy with the current behavio. That would obviously also be a breaking change.

Does this API make sense? I also considered passing the config object directly to `getFileInfo()`, since the consumer should be able to calculate that using the public API to essentially do the same thing done here, but it felt like a nicer UX to have the method take care of it for you.


kachkaev:
`useConfig` sounds like a name of some React hook to me. How about `resolveConfig` or `configResolution`?

fisker:
eslint has a `useEslintrc`, how about we use the same `usePrettierrc`, and `--no-prettierrc` for cli

kaicataldo:
Happy go with consensus on the option name. To be clear, though, I don’t think this will be exposed on the CLI (unless I’m misunderstanding something). This is an enhancement to the Node API.

evilebottnawi:
I think better use `usePrettierrc`, maybe in future we implement this option on CLI too

/cc @prettier/core ideas about name?

kaicataldo:
Friendly ping! Would love to move forward with this :)

kaicataldo:
Sounds good to me!

PR #6675, opened 2019-10-17
Update integrating-with-linters.md

marcospgp:

### DIFF
--- a/docs/integrating-with-linters.md
+++ b/docs/integrating-with-linters.md
@@ -64,6 +64,12 @@ Then in `.eslintrc.json`:
 }
 ```
 
+Please note that `eslint-plugin-prettier` does not install Prettier or ESLint for you. You must install these yourself:
+
+```bash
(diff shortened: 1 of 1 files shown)


lydell:
Hi!

We already mention that you need to install `prettier` yourself near the top of the file:

> All these instructions assume you have already installed `prettier` in your `devDependencies`.

It doesn’t mention installing `eslint`, though. But if we do, we should mention installing `tslint` and `stylelint` as well in their blocks.

But I'm not sure what the best way to document this is yet. It would help if you could provide some more context on this docs change! What did your scenario look like? Are Prettier and ESLint new to you?

One idea I have is to add `xxlint prettier` to the end of all `yarn add` commands on the page, such as:

```
yarn add --dev stylelint-prettier stylelint prettier
```

Even if the user already has `stylelint` and `prettier` installed it doesn’t hurt to install them again. And users can copy just the first part of the command if they know that the rest is already installed.

PR #6688, opened 2019-10-20
Fix tooltip being hidden behind browser scrollbars

mrseanbaines:
Fixes https://github.com/prettier/prettier/issues/6297

Adds a `z-index` to the button tooltips in the footer in the playground to bring the tooltip above the browser scrollbar.

image

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -1014,6 +1014,10 @@ class A {
 }
 ```
 
+#### Website: Fix tooltip being hidden behind browser scrollbars ([#6688] by [@mrseanbaines])
+
+Fixes an issue where the tooltip above the buttons in the footer of the playground would be hidden behind the browser's scrollbars.
+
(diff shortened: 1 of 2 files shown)


lydell:
Thanks, this appears to be working!

We usually don’t mention changes to the website in the changelog, because it’s not interesting to know about when you’re reading about what’s new and thinking about upgrading. (It’s still very much appreaciated, though!)

mrseanbaines:
> Thanks, this appears to be working!
> 
> We usually don’t mention changes to the website in the changelog, because it’s not interesting to know about when you’re reading about what’s new and thinking about upgrading. (It’s still very much appreaciated, though!)

Ha, I wasn't sure if I was supposed to or not for this. Shall I remove the `CHANGELOG` change then?

lydell:
Yes, please!

mrseanbaines:
@lydell All done.. also realised a previous fix I did for this had somehow gotten through, so removed that too. Should just need that one `z-index` change.

lydell:
Thanks!

PR #6692, opened 2019-10-21
Avoid switch and chain indents when tab >2 spaces

grinstead:
I work in a codebase that uses 4 spaces for its indent level and 80 spaces for its width, which means every switch statement case-block will indent 10% (8 spaces up) of the visual space. This is visually distracting by itself, but it also causes wrapping nearly everywhere in the statements, which is even more distracting. With a redux codebase like my team uses, it becomes nearly impossible to read anything quickly.

This diff removes the indent in front of the "case" (but not the block), thus changing:
```
// example from the redux example site
function todoApp(state = initialState, action) {
    switch (action.type) {
        case SET_VISIBILITY_FILTER:
            return Object.assign({}, state, {
                visibilityFilter: action.filter
            })
        case ADD_TODO:
            return Object.assign({}, state, {
                todos: [
                    ...state.todos,
                    {
                        text: action.text,
                        completed: false
                    }
                ]
            })
        case TOGGLE_TODO:
            return Object.assign({}, state, {
                todos: state.todos.map((todo, index) => {
                    if (index === action.index) {
                        return Object.assign({}, todo, {
                            completed: !todo.completed
                        })
                    }
                    return todo
                })
            })
        default:
            return state
    }
}
```

to a more readable:
```
function todoApp(state = initialState, action) {
    switch (action.type) {
    case SET_VISIBILITY_FILTER:
        return Object.assign({}, state, {
            visibilityFilter: action.filter
        })
    case ADD_TODO:
        return Object.assign({}, state, {
            todos: [
                ...state.todos,
                {
                    text: action.text,
                    completed: false
                }
            ]
        })
    case TOGGLE_TODO:
        return Object.assign({}, state, {
            todos: state.todos.map((todo, index) => {
                if (index === action.index) {
                    return Object.assign({}, todo, {
                        completed: !todo.completed
                    })
                }
                return todo
            })
        })
    default:
        return state
    }
}
```

In the same vein, this diff removes the tab in front of method chains, which are already visually unambiguous and (with the indent) often make promise chains indent like mad.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -119,6 +119,13 @@ const {
 
 let uid = 0;
 
+// If clients have a large tab setting (usually 4 spaces), then certain
+// indentation points can detract from the reading experience. Use this
+// function to use an indent only if it is small.
+function indentIfSmallTab(options, content) {
+  return options.tabWidth <= 2 ? indent(content) : content;
+}
+
 function shouldPrintComma(options, level) {
   level = level || "es5";
 
@@ -1999,7 +2006,8 @@ function printPathNoParens(path, options, print, args) {
         ),
         " {",
         n.cases.length > 0
-          ? indent(
+          ? indentIfSmallTab(
+              options,
               concat([
                 hardline,
                 join(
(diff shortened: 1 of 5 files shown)


icopp:
> removes the tab in front of method chains

This is a pretty dramatic change, and one that a lot of people (including me) are likely to dislike for readability reasons.

PR #6699, opened 2019-10-23
docs on whitespace sensitivity: better wording

thorn0:


lydell:
Is it called “whitespace” or “whitespaces”? If you feel like it, you could fix that too while at it (if it needs fixing).

PR #6704, opened 2019-10-24
typescript@3.6.4 update

brodybits:
### What

see title

### Why

Most recent release of TypeScript is *highly desired* to support JavaScript type linting (type checking using JSDoc comments) as we have been working on in multiple PRs such as  #6313 & #6702.

A timely review would be much appreciated.

Unfortunately I don't think we should include TypeScript 3.7 in the RC stage.

/cc @Shinigami92 @lydell

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "resolve": "1.12.0",
     "semver": "6.3.0",
     "string-width": "3.1.0",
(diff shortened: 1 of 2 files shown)


brodybits:
Thanks for the quick merge!

evilebottnawi:
It is just a patch version :smile:

PR #6706, opened 2019-10-24
Add option to keep json numeric literal

NullMDR:
With the new option jsonKeepNumericLiteral set to true,

```json
{ "a": 1.00, "b": 1.200 }
```
will be formatted to
```json
{ 
  "a": 1.00, 
  "b": 1.200
}
``` 
instead of
```json
{ 
  "a": 1, 
  "b": 1.2
}
```


kachkaev:
Hi @NullMDR, just curious: what practical problem does the new option solve? In general, new options are not welcomed as they complicate things: https://prettier.io/docs/en/option-philosophy.html

NullMDR:
@kachkaev I'm the maintainer of https://github.com/Azure/azure-rest-api-specs and currently I need to add a new CI task to check the format for all these json swagger specs. Prettier meets my requirement but except one thing: it tries to transform all these numbers which is quite annoying. Some service team has json with these numbers end by 0 like 60.90 by purpose, e.g. https://github.com/Azure/azure-rest-api-specs/blob/master/specification/consumption/resource-manager/Microsoft.Consumption/preview/2019-05-01-preview/examples/ChargesForBillingPeriodByEnrollmentAccount.json#L19 .  I have search for the option to disable this behavior in prettier but could not find out one so I make this PR. If this PR could not be merged I would consider maintaining my own fork.

NullMDR:
@evilebottnawi I'm wondering what's your advise on my requirement?

thorn0:
> Some service team has json with these numbers end by 0 like 60.90 by purpose

So what is the purpose? Still not clear. Why don't they use strings if those zeroes are important?

kachkaev:
@NullMDR your use case seems to be too narrow to justify a new option. Perhaps, you could try creating a custom [plugin](https://prettier.io/docs/en/plugins.html) for yourself, which would replace the printer. An example of a plugin that works with JSONs I can recall is https://www.npmjs.com/package/prettier-plugin-packagejson

NullMDR:
@kachkaev Thank you for your information! I'll try it then.

lydell:
Sounds like it’s been decided not to merge this, so I'll close! Let me know if I missed anything.

Issue #6703, opened 2019-10-24
Many Check JS errors


brodybits:
as can be seen by one of the following:

- `npx tsc --allowJs --checkJs --noEmit --resolveJsonModule --target es5 src/index.js`
- enable Check JS in VSCode then use VSCode to open a file such as `src/language-js/printer-estree.js`
- try doing `npx tsc` with the tsconfig.json file from PR #6313

see <https://medium.com/@trukrs/javascript-type-linting-5903e9e3625f> for more information

The general idea is to use the TypeScript compiler for static type checking on the JavaScript code, using JSDoc comments to specify the typing. This would be an alternative to porting to TypeScript, as was discussed in #6286. This idea was originally pioneered by the Google Closure compiler.

I resolved a few of these errors in PR #6702, there are many, many more checkJs errors for us to look at.

thorn0:
There is a big problem with your multi-PR approach. Namely, for each specific PR, how can we verify that it improves things? Can we measure this improvement somehow?

Is it possible to setup such a CI check that would show that the number of type errors found by `tsc` went down and we got closer to `Found 0 errors.`?

brodybits:
> Is it possible to setup such a CI check that would show that the number of type errors found by `tsc` went down and we got closer to `Found 0 errors.`?

That should be possible, but I am not personally convinced it would be worth the time and attention that it would divert from other things that I and other people can give to improvements and fixes on this and other projects.

Another thing is that I think resolving the type errors should be done in 2 parts: first without strict mode then with strict mode enabled.

Unfortunately I don't have much more extra time to work on the JSDoc type improvements right now. I would be happy to provide feedback, guidance, and even suggestions to anyone else who wants to take this on.

One idea I have would be to start a big PR that adds tsconfig.json and starts resolving the errors one module at a time until we can resolve most or all errors with no strict mode enabled. Then we could discuss how to integrate all of the changes. I would be happy to propose a starting point if people are interested.

Shinigami92:
> [...] resolving the type errors should be done in 2 parts: first without strict mode then with strict mode enabled [...] 

THIS! 👍

y912765390:
Hello, I don't know why this library is introduced into the Vue project. With this API, the project will not run. Can't find the third party, can you help me? Thank you very much.

PR #6707, opened 2019-10-24
[WIP] [DRAFT] JSDoc fixes for printer-estree.js

brodybits:
with some TODO items

ref: #6703

the following command shows no errors in `src/language-js/printer-estree.js`:

```
 npx tsc --allowJs --checkJs --noEmit --resolveJsonModule --target es5 src/index.js
```

/cc @Shinigami92

TODO items:

- [ ] updates to doc-builders.js should be done elsewhere, such as PR #6710, and moved out of this PR
- [ ] automate JSDoc type checking, as already suggested by @thorn0 elsewhere, before moving forward with this proposal
- [ ] resolve TODO items in the code, if possible

### DIFF
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -1,5 +1,9 @@
 "use strict";
 
+/**
+ * @typedef {string | Object} Doc
+ */
+
 function assertDoc(val) {
   /* istanbul ignore if */
   if (
@@ -11,6 +15,10 @@ function assertDoc(val) {
   }
 }
 
+/**
+ * @param {Doc[]} parts
+ * @returns Doc
+ */
 function concat(parts) {
   if (process.env.NODE_ENV !== "production") {
     parts.forEach(assertDoc);
@@ -25,6 +33,10 @@ function concat(parts) {
   return { type: "concat", parts };
 }
 
+/**
+ * @param {Doc} contents
+ * @returns Doc
+ */
 function indent(contents) {
   if (process.env.NODE_ENV !== "production") {
     assertDoc(contents);
@@ -41,6 +53,11 @@ function align(n, contents) {
   return { type: "align", contents, n };
 }
 
+/**
+ * @param {Doc} contents
+ * @param {Object} [opts]
+ * @returns Doc
+ */
 function group(contents, opts) {
   opts = opts || {};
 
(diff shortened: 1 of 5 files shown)


src/doc/doc-builders.js
```diff
@@ -1,5 +1,9 @@
 "use strict";
 
+/**
+ * @typedef {string | Object} Doc
```
thorn0: My testing shows that TS treats both `object` and `Object` as `any` is JSDoc. So `string | Object` is `any` too. Any idea how this can be improved without introducing an extra `.d.ts` file?

Shinigami92: Yea... instead if `object` you should define it
like `@typedef {string | {type: string, ...}} Doc`

brodybits: > like `@typedef {string | {type: string, ...}} Doc`

Did not quite work in JSDoc, I fixed it a different way.

I confirmed that if I try the following:

```diff
diff --git a/src/doc/doc-builders.js b/src/doc/doc-builders.js
index a35d1597..aa56ffd0 100644
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -2,7 +2,7 @@

 /**
  * @typedef {Object} DocObject
- * @property {string} type
+ * @property {string} tpe
  *
  * @typedef {string | DocObject} Doc
  */
```

then I would get a bunch of type errors from `tsc`.

brodybits:
I am setting this proposal to WIP for now. I am thinking to raise separate PRs to resolve the JSDoc type errors in some other modules that printer-estree.js depends on before continuing with printer-estree.js itself.

brodybits:
Closing for now. I think this should be done in multiple parts.

brodybits:
Reopening as a DRAFT WIP for now.

src/language-js/printer-estree.js
```diff
@@ -4030,7 +4040,9 @@ function printArgumentsList(path, options, print) {
           shouldGroupFirst
             ? concat([
                 "(",
+                // @ts-ignore [TODO: ensure printedExpanded is defined]
```
brodybits: I think we should consider resolving this JSDoc type error (and the one below), rather than simply ignoring it.

Looking through the code, I think this should not be a problem but would rather be 100% sure.

fisker:
Close due to age.

PR #6709, opened 2019-10-25
fix JSDoc type error in `src/doc/doc-printer.js`

### LINKED ISSUE #6703: Many Check JS errors

brodybits:
Quick fix to resolve the JSDoc type error, to help with resolving #6703.

With this update, the following command will show no JSDoc type errors:

```
npx tsc --allowJs --checkJs --noEmit --resolveJsonModule --target es5 src/doc/doc-printer.js
```

I suspect that `PropertyKey` and `MODE` were used in the JSDoc comment for some historical reasons, as part of PR #4563 (dcf44ffbdc2f403de02f12516b0c6d5d5813b16f). Looking through the commits in PR #4563, I see these were introduced in 0892d1780afc7cebd51565040c2b8bbf2ff3bdf7.

In the future, we may want to use a proper JSDoc enum for the mode. From some quick research, I think this would involve some minor source code updates.

### DIFF
--- a/src/doc/doc-printer.js
+++ b/src/doc/doc-printer.js
@@ -4,7 +4,7 @@ const { getStringWidth } = require("../common/util");
 const { convertEndOfLineToChars } = require("../common/end-of-line");
 const { concat, fill, cursor } = require("./doc-builders");
 
-/** @type {{[groupId: PropertyKey]: MODE}} */
+/** @type {{[groupId: string]: number}} */
 let groupModeMap;
 
 const MODE_BREAK = 1;


thorn0:
> With this update, the following command will show no JSDoc type errors

So this command is a kind of test. We should consider adding it to CI. Wouldn't it be a reasonable alternative to what I proposed in https://github.com/prettier/prettier/issues/6703#issuecomment-546330770?

brodybits:
> We should consider adding it to CI. Wouldn't it be a reasonable alternative to what I proposed in [#6703 (comment)](https://github.com/prettier/prettier/issues/6703#issuecomment-546330770)?

I would rather discuss this in #6703, but how about the following idea:

- we start with simple check-js script, perhaps with tsconfig.json, that would only check a limited number of artifacts in src for now (only check those artifacts that would already pass tsc --checkJs)
- for every source file that we fix, we add it to the list in tsconfig.json
- at some point, we should be able to update tsconfig.json to check all JavaScript sources with exceptions such as node_modules, tests, tests_integration, etc.

thorn0:
Sounds like a plan. Would you put together a proof a concept as part of this PR?

brodybits:
> Would you put together a proof of concept as part of this PR?

I may need a few days to work on this.

thorn0:
superseded by #6770

PR #6710, opened 2019-10-25
add & apply doc-builders.js type improvements

brodybits:
- add JSDoc type comments to `src/doc/doc-builders.js` & `doc/index.js`
- use the new Doc type in `src/language-js/printer-estree.js`

based on some of the changes in WIP PR #6707

resolves a few of the checkJs errors in `src/language-js/printer-estree.js`, as part of #6703

`src/doc/doc-builders.js` passes checkJs type checking in strict mode with these changes:

```
npx tsc --allowJs --checkJs --noEmit --resolveJsonModule --target es5 --strict src/doc/doc-builders.js
```

There are a couple of TBD comments that I would appreciate some feedback and guidance on.

### DIFF
--- a/src/doc/doc-builders.js
+++ b/src/doc/doc-builders.js
@@ -1,5 +1,19 @@
 "use strict";
 
+/**
+ * @typedef {Object} DocObject
+ * @property {string} type
+ * @property {boolean} [hard]
+ * @property {boolean} [literal]
+ *
+ * @typedef {string | DocObject} Doc
+ *
+ * @typedef {Object} DocOptions - TBD more specific?
+ */
+
+/**
+ * @param {Doc} val
+ */
 function assertDoc(val) {
   /* istanbul ignore if */
   if (
@@ -11,6 +25,10 @@ function assertDoc(val) {
   }
 }
 
+/**
+ * @param {Doc[]} parts
+ * @returns Doc
+ */
 function concat(parts) {
   if (process.env.NODE_ENV !== "production") {
     parts.forEach(assertDoc);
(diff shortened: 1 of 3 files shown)


src/doc/doc-builders.js
```diff
@@ -41,6 +65,11 @@ function align(n, contents) {
   return { type: "align", contents, n };
 }
 
+/**
+ * @param {Doc} contents
+ * @param {Object} [opts] - TBD ???
```
brodybits: I think we should define a new doc options type, ideally in a separate PR. Any guidance would be much appreciated.

src/doc/doc-builders.js
```diff
@@ -33,6 +53,10 @@ function indent(contents) {
   return { type: "indent", contents };
 }
 
+/**
+ * @param {number} n
```
brodybits: This change would resolve the type error ignored in `markAsRoot` below:

```suggestion
 * @param {number | Doc} n
```

but I am not sure if this is something we really want to do here.

I wonder if we should consider adding a special align function to get what we need for `markAsRoot` instead?

evilebottnawi:
@brodybits Can we do all typescript improvements in one PR? This is not very convenient when spread over several PRs

Shinigami92:
> @brodybits Can we do all typescript improvements in one PR? This is not very convenient when spread over several PRs

We need a starting base, so more then one contributer can participate  to all the JSDocs in parallel.

evilebottnawi:
@Shinigami92 why do not create base in one PR? Just for infromation?

Shinigami92:
> @Shinigami92 why do not create base in one PR? Just for information

Tried here 😉 https://github.com/prettier/prettier/pull/6313
... I think we are starting to move in circles

brodybits:
> > @brodybits Can we do all typescript improvements in one PR?
> > [...]
>
> why do not create base in one PR?

I think these questions are now answered in #6703.

I would also appreciate some guidance or any other input on how to deal with my TBD comments on:

- options object
- use align function from `markAsRoot`

Thanks!

fisker:
Close due to age.

PR #6721, opened 2019-10-28
Revert "Link to `fast-glob` instead of `glob` in docs (#6333)"

### LINKED ISSUE #6719: this PR need revert
this PR 

https://github.com/prettier/prettier/pull/6333

we are  using `globby` , but current version we  are using is not 

> the syntax from the fast-glob module

fisker:
Revert #6333, fixes #6719

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -17,7 +17,7 @@ In practice, this may look something like:
 prettier --single-quote --trailing-comma es5 --write "{app,__{tests,mocks}__}/**/*.js"
 ```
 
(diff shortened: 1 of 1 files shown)


brodybits:
After another look at PR #6333, I am not 100% convinced reverting it is 100% what we want.

The point of PR #6333 is that the document referred to the glob syntax from node-glob, while we are now using whatever glob syntax is supported by globby.

I think we also do not want to point to the README on the master branch of the globby project, since its readme now refers to fast-glob.

I wonder if we should point to something like <https://github.com/sindresorhus/globby/blob/v6.1.0/readme.md#globbing-patterns> until we are ready to use globby@10.

Or maybe it is time to stop supporting old installations (Node.js 6 from GitHub) and use globby@latest?

Food for thought.

PR #6730, opened 2019-10-29
Update `escape-string-regexp` to v2

fisker:


fisker:
@evilebottnawi you are right,

this should not effect 

https://github.com/prettier/prettier/blob/0cbacd31561f981b7e30972afc09051531c7a7c3/src/utils/front-matter.js#L3 

but for this

https://github.com/prettier/prettier/blob/27ef80d56afecca2a02dfb7349be934209896b9d/src/common/util.js#L4 this is a breaking change, but we don't have any actual test on those utils

https://github.com/prettier/prettier/blob/023a8b78df18a3afd8f82f9c2fc51bae02273d24/tests_integration/__tests__/util-shared.js

evilebottnawi:
@fisker maybe we can add tests?

fisker:
This only public api `sharedUtil.getMaxContinuousCount` using this module is tested on #7262

evilebottnawi:
I think will be better do it in next branch

PR #6738, opened 2019-10-29
Run CI on node 12

fisker:
Node 12 is already LTS, let's run CI scripts on Node 12

https://nodejs.org/en/blog/release/v12.13.0/


thorn0:
We should run something on Node 8 too in order to detect situations like https://github.com/prettier/prettier/pull/6544#pullrequestreview-294927535.

fisker:
run tests on node 8 can't guarantee that , this PR will https://github.com/prettier/prettier/pull/6373

PR #6765, opened 2019-10-31
Add `private` to `package.json`

fisker:
use [private](https://docs.npmjs.com/files/package.json#private) to prevent publishing from root dir
 

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "prettier",
+  "private": true,
   "version": "1.18.2",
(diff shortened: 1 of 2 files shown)


thorn0:
> We could even add this to scripts I guess

That's what this commit removes. :)

lydell:
Hehe, didn’t see that :) The reason I suggested `prepublishOnly` is that it seems to run earlier and give a nicer error message.

fisker:
close?

PR #6770, opened 2019-10-31
[...] Start JSDoc type linting

brodybits:
As requested by @thorn0 in PR #6709, here are some updates to start JSDoc type linting using the TypeScript compiler (`tsc`) as described by @rbiggs in: https://medium.com/@trukrs/javascript-type-linting-5903e9e3625f

The idea is that we can enable JSDoc type linting on more and more sources as we are able to resolve the JSDoc type errors. This would be a partial solution to #6703, as an alternative to #6286 (convert to TypeScript).

I am using the tsconfig.json file from PR #6313 by @Shinigami92, with the following updates:

- add more entries to excludes to exclude the src files that do not yet pass JSDoc type linting
- minor updates to settings
- add `checkjs` script and integrate it with the Dev_Lint task on Azure Pipelines
- ~~introduce JSDoc type linting with `strict` mode enabled in tsconfig-strict.json, on a limited number of src files for now, and add it to `checkjs` script~~
- resolve JSDoc type error in `src/doc/doc-printer.js`, as I proposed in PR #6709
- add JSDoc type improvements to `src/doc/doc-builders.js`, as I proposed in PR #6710
- other fixes & updates, in response to comments by @thorn0

In case we do not want strict JSDoc type checking, we can just remove use of tsconfig-strict.json. I do think it would be ideal to use strict type checking as much as possible.

There is now a limited number of JSDoc improvements and fixes within this PR. I think further updates and fixes should be done in other PRs, assuming this proposal is accepted and merged.

The changes in this proposal includes work by multiple authors. At this point the last commit as Co-Authored-By comments which should hopefully preserve the proper co-authorship credits. I would like to ask that the Co-Authored-By comments be preserved if this proposal is accepted and merged.

/cc @Shinigami92 @thorn0 @evilebottnawi @rbiggs

TODO items are now tracked in TBD comments.


package.json
```diff
@@ -122,6 +122,7 @@
     "perf-repeat": "yarn && yarn build && cross-env NODE_ENV=production node ./dist/bin-prettier.js --debug-repeat ${PERF_REPEAT:-1000} --loglevel debug ${PERF_FILE:-./index.js} > /dev/null",
     "perf-repeat-inspect": "yarn && yarn build && cross-env NODE_ENV=production node --inspect-brk ./dist/bin-prettier.js --debug-repeat ${PERF_REPEAT:-1000} --loglevel debug ${PERF_FILE:-./index.js} > /dev/null",
     "perf-benchmark": "yarn && yarn build && cross-env NODE_ENV=production node ./dist/bin-prettier.js --debug-benchmark --loglevel debug ${PERF_FILE:-./index.js} > /dev/null",
+    "checkjs": "tsc && tsc -p ./tsconfig-strict.json",
```
thorn0: "typecheck" would be more self-descriptive than "checkjs"

tsconfig-strict.json
```diff
@@ -0,0 +1,27 @@
...
+    "noEmit": true,
+    "target": "es5",
+    "module": "commonjs",
+    "resolveJsonModule": true,
+    "strict": true,
+    /* Additional Checks */
+    "noUnusedLocals": true,                /* Report errors on unused locals. */
+    "noUnusedParameters": true,            /* Report errors on unused parameters. */
+    "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
+    "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
```
thorn0: These additional checks have nothing to do with type checking. We have ESLint for this. Please remove them.

package.json
brodybits: Thanks for the quick feedback.

I would favor calling it "lint-types", would also maybe favor including it in the "lint" package script.

I would be interested if anyone else has feedback as well.

tsconfig-strict.json
brodybits: Gone

package.json
thorn0: We have `lint-docs` and `lint-dist`. `lint-type` wouldn't be consistent, it would look as if there were a folder or something called 'types'. Also typechecking is an established notion, separate from linting. I would understand 'linting types' as a check for issues in the JSDoc comments themselves, like what [`eslint-plugin-jsdoc`](https://www.npmjs.com/package/eslint-plugin-jsdoc) does.

thorn0: > including it in the "lint" package script

Better keep them separate. It'll look more clear on CI. And `yarn lint` is a little slow already, so I'd prefer to be able to run these checks independently. On the other hand, a command to run them all would be useful too.

src/doc/doc-builders.js
```diff
@@ -1,5 +1,17 @@
 "use strict";
 
+/**
+ * @typedef {Object} DocObject
```
thorn0: This type should eventually become a tagged union.

src/doc/doc-builders.js
```diff
@@ -41,6 +65,11 @@ function align(n, contents) {
   return { type: "align", contents, n };
 }
 
+/**
+ * @param {Doc} contents
+ * @param {Object} [opts] - TBD ???
```
thorn0: https://github.com/prettier/prettier/blob/master/commands.md#group

jest.config.js
```diff
@@ -9,7 +9,8 @@ const requiresPrettierInternals = [
 ];
 
 const semver = require("semver");
-const isOldNode = semver.parse(process.version).major <= 4;
+const version = semver.parse(process.version);
+const isOldNode = version && version.major <= 4;
```
thorn0: Let's focus on src/**/*.js. Build scripts and the like can be taken care of later.

package.json
brodybits: How about "check-types"?

src/doc/doc-builders.js
brodybits: The DocObject is just one of the 2 options in the tagged union which I defined below. Can you think of a better way to do this?

src/doc/doc-builders.js
brodybits: What about `opts.id` below?

Any pointers to what type we should expect for `opts.id`?

Should we consider adding it to GroupOpts?

jest.config.js
brodybits: Reverted this update & reverted other updates related to jest.config.js

package.json
thorn0: Better than "checkjs".

src/doc/doc-builders.js
thorn0: The union you defined is not tagged. We should define a separate type for each kind of Doc, and each subtype's field `type` will be of a string literal type. But it's totally okay as is for now. Let's do it later.

package.json
brodybits: I have now split it into check-types and check-types-strict. (I hope I didn't break the build, we will see.)

I have also updated check-types-strict (in a separate commit) to call check-types first, since check-types-strict would cover a very limited number of source files at this point.

I am now wondering if the check-types-strict should be moved into its own PR?

src/doc/doc-builders.js
brodybits: > https://github.com/prettier/prettier/blob/master/commands.md#group

Done.

My questions about opts.id still stand.

package.json
thorn0: Yep, I wouldn't touch `strict` until we have zero errors without it.

src/doc/doc-builders.js
brodybits: > We should define a separate type for each kind of Doc, and each subtype's field `type` will be of a string literal type.

I would appreciate any pointers.

> Let's do it later.

Any ideas show we should avoid forgetting this?

src/doc/doc-builders.js
```diff
@@ -1,5 +1,21 @@
...
 
+/**
+ * @typedef {Object} DocObject
+ * @property {string} type
+ * @property {boolean} [hard]
+ * @property {boolean} [literal]
+ *
+ * @typedef {string | DocObject} Doc
+ *
+ * @typedef {Object} GroupOpts
```
thorn0: I don't think we need a typedef for this type as we are not going to refer it by name anywhere else. That's why better to move its property definitions to `@param` of the `group` function.

thorn0: As for `id`, define it as `any` and optional.

thorn0: See the syntax for for one-off type specifications in `@param` [here](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#typedef-callback-and-param).

package.json
brodybits: The one thing is that strict mode is needed to ensure that I have added JSDoc comments for all functions in `src/doc/doc-builders.js`, like I proposed in this PR. If we drop strict mode, any ideas how we should verify proper proper documentation of `src/doc/doc-builders.js`?

thorn0: The typechecker will tell us when we work on other files and it finds a call to those functions with unexpected arguments.

src/doc/doc-builders.js
thorn0: https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html

A TBD comment will do.

src/doc/doc-builders.js
brodybits: > As for `id`, define it as `any` and optional.

+1, thanks.

> See the syntax for for one-off type specifications in `@param` [here](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#typedef-callback-and-param).

Thanks for the pointer. But I would actually downvote the one-off syntax for a couple of reasons:

- seems to be more ugly than the typedef that I proposed
- Considering that GroupOpts is defined in the documentation, I would really favor "formally" documenting it in the source code in order to show the level of consistency

thorn0: Seriously, the name `GroupOpts` won't be used anywhere else. It's a one-off type. It doesn't need an alias.

thorn0: It's a standard practice for options objects, which are a basically how named parameters look in JS. Defining it outside the function's jsdoc block makes little sense to me.

src/doc/doc-builders.js
brodybits: > My questions about opts.id still stand.

Answered in <https://github.com/prettier/prettier/pull/6770#discussion_r341402982>, resolving this now.

src/doc/doc-builders.js
brodybits: Would you mind if we just revert this change now, and do the one-off type in a separate PR?

thorn0: Okay, but I don't really understand this decision.

package.json
thorn0: "how we should verify proper proper documentation of src/doc/doc-builders.js"

BTW that's not what typechecking does. It does not verify such things. It verifies that we use functions (classes, variables etc) the way we specified in their type annotations. Many type annotations are optional at that because the typechecker is able to infer many things, e.g. return types.

brodybits:
I think the changes should be almost ready to go. All "questionable" JSDoc comment annotations now have "TBD" comments to hopefully ensure they are not forgotten in the future.

I hope we can get this approved and merged soon to avoid blocking further JSDoc type fixes.

My one question is whether we should expect a "squash merge", a merge commit, or something else?

I would like to rebase and cleanup the commit messages before we go through the final review and merge.

Any further questions or concerns would be appreciated as well. Thanks!

package.json
thorn0: Please disregard my previous comment. I misunderstood what you meant by 'documentation'. Obviously, you just meant 'adding types'.

thorn0:
We always squash merge in this project.

src/doc/doc-printer.js
```diff
@@ -4,7 +4,7 @@ const { getStringWidth } = require("../common/util");
 const { convertEndOfLineToChars } = require("../common/end-of-line");
 const { concat, fill, cursor } = require("./doc-builders");
 
-/** @type {{[groupId: PropertyKey]: MODE}} */
+/** @type {{[groupId: string]: number}} */
```
thorn0:
```suggestion
/** @type {Record<symbol, GroupMode>} */
```

thorn0: Turns out TS doesn't really support symbol indexers: https://github.com/microsoft/TypeScript/issues/1863
It'll be an error if we set `noImpliciyAny` to `true`. But that's how this variable is used. Its keys are symbols, values 1 or 2. Defining its as `{[groupId: string]: number}` is a mistake.

brodybits: Thanks @thorn0, I committed the suggestion. I must admit that this is something that I don't understand too well. But I am really happy that we can get this fixed.

thorn0: What exactly do you not understand?

thorn0:
In general, you don't need to add `@returns`. TS infers it.

src/doc/doc-printer.js
brodybits: How you used Record, symbol, typeof MODE_BREAK, and typeof MODE_FLAT.

Looks nice and clean to me, though.

I guess I gotta to study TypeScript a bit more.

(I wouldn't mind a quick link, if you have one)

brodybits:
> In general, you don't need to add `@returns`. TS infers it.

Thanks for the info, really interesting.

When I tried using `eslint-plugin-jsdoc` in my local workarea, it suggested I was missing the `@returns`.

I would personally favor making it more explicit if it does not make things too messy. I tend to prefer to keep this kind of thing more explicit in JavaDoc & JSDoc, even if it is just documentation internal to the source module itself. Please let me know if you can think of a really compelling reason to take it back out.

package.json
brodybits: I am marking this conversation about the package scripts as resolved now.

I pushed a TBD comment to tsconfig.json to avoid forgetting to try type checking in strict mode in the fuutre.

lipis:
If we switch to (actually when we switch) TS these JSDocs won't be needed.. right?

brodybits:
> If we switch to (actually when we switch) TS these JSDocs won't be needed.. right?

My understanding of the discussion in #6286 is that it is not desired to convert this project to TypeScript. I personally agree with this position very strongly.

Even if someone would convert this project to TypeScript, I would think that the JSDoc comments would still be nice internal documentation, in a similar fashion to how people can use JavaDoc comments to document Java code both internally and externally.

Am I missing something here?

rbiggs:
Well, good TS always has TSDoc comments, which is just a superset of JSDoc. This provides richer intellisense. Just look at the `d.ts` files that the TypeScript install provides: https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2015.core.d.ts

It would be trivial to convert good JSDoc comments to TSDoc comments.

lipis:
> It would be trivial to convert good JSDoc comments to TSDoc comments.

Fair enough! As for converting this project to TypeScript, I think sooner or later every JS project should be converted to TS, and Prettier is not an exception.

rbiggs:
To be honest, I don't think so. 10 years ago everyone had to know how to use JQuery or you weren't considered a good developer. Then came CoffeeScript, again everyone had to write CoffeeScript because it gave you error-free JavaScript. Now the new thing is TypeScript. 10-15 years from now no one will care about TypeScript because JavaScript will have its own type system that works at runtime. FYI JavaScript 4.0 was supposed to have that. At the time Microsoft apposed implementing typed JavaScript because they didn't want to have to re-write IE6. Instead everyone settled on JavaScript 3.5, which as not much compared to what the proposals were for JS4.0.

thorn0:
@brodybits re `@returns`

> Please let me know if you can think of a really compelling reason to take it back out.

I don't feel strongly about it. They are useful mainly for long functions with multiple return statements to control that they don't return something unintended.

src/doc/doc-printer.js
thorn0: * `Record` is a [predefined utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt).
* [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) aren't a TS thing, it's a primitive value type in JS. Their purpose is to represent unique keys that never clash. Unfortunately, TS's type system doesn't fully support them (see the link to the issue in my other message).
* The [`typeof` operator](https://basarat.gitbooks.io/typescript/docs/types/moving-types.html#capturing-the-type-of-a-variable) applied to a constant returns a literal type.

Shinigami92:
:tada:

orta:
This is super cool!

thorn0:
@orta It probably is, but TS's type system seems to be still not flexible enough for our use cases. See https://github.com/microsoft/TypeScript/issues/35045#issuecomment-554272882

PR #6748, opened 2019-10-29
[RFC DRAFT WIP] globby@10 update (with fast-glob implementation)

brodybits:
- use globby@10 (globby@10.0.1) which has fast-glob implementation
- some minor test snapshot updates in tests_integration since fast-glob does not always keep the same file order as node-glob

I am raising this PR for the sake of testing and discussion related to PR #6639 (globby@7 update).

This seems to work from git on macOS and Linux, *mostly* working from git on Windows.

P.S. If we do this update, there should be no need to revert PR #6333 as discussed in #6719 / #6721

TODO items:

- [ ] resolve performance issue with finding plugins
- [ ] consider updating `CHANGELOG.unreleased.md` with information about switch to fast-glob implementation
- [ ] consider updating doc that fast-glob is only used in the new release ref: #6333 /#6719 / #6721

/cc @fisker @lydell @evilebottnawi

- ~~I’ve added tests to confirm my change works.~~
- ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~~
- [ ] (If the change is user-facing) I’ve added my changes to the `CHANGELOG.unreleased.md` file following the template.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #6776, opened 2019-11-01
Update `globby` to v11

fisker:


brodybits:
Comment from @mrmlnc in <https://github.com/prettier/prettier/pull/6748#issuecomment-551043561>:

> JFYI
> 
> > I'm a `fast-glob` maintainer.
> 
> The `fast-glob` package in [most cases](https://github.com/mrmlnc/fast-glob#benchmarks), faster than competitors. But he have a some performance problems for complex patterns ([mrmlnc/fast-glob#156](https://github.com/mrmlnc/fast-glob/issues/156)) like `{*,@*/*}` or [mrmlnc/fast-glob#156 (comment)](https://github.com/mrmlnc/fast-glob/issues/156#issuecomment-485327918). And, yeap, we will work on it.

JounQin:
Can we try [tiny-glob](https://github.com/terkelg/tiny-glob/blob/master/readme.md#benchmarks) according to its benchmarks?

evilebottnawi:
@JounQin `tiny-glob` don't have enough abilities (and has bugs), also the latest version was released a year ago, looks like abandoned

evilebottnawi:
@fisker What is status? Will be great to update globby before fix https://github.com/prettier/prettier/issues/6854

fisker:
I guess If we are going to update `globby` above v7, I need merge some code from #6639

fisker:
Here is the situation:

`globby@7` is the last version use `node-glob` and ready to merge #6639 

`globby@10`(or maybe 11) is using `fast-glob` this is faster than `node-glob` on searching file to format, see https://github.com/prettier/prettier/pull/6748#issuecomment-548640881 . But this module has a problem on our `findPluginsInNodeModules`, it's very slow see https://github.com/prettier/prettier/pull/6748#issuecomment-551043561 , in this PR, I use `node-glob` for `findPluginsInNodeModules` and `globby` for file searchin.

mrmlnc:
JFYI: https://github.com/mrmlnc/fast-glob/issues/156 will be fixed in `fast-glob@3.2.0` (14.02.2010).

fisker:
@mrmlnc great, I'll try later

thorn0:
@evilebottnawi 

> Will be great to update globby before fix #6854

Why? I don't see how this blocks #6854.

evilebottnawi:
@thorn0 node-glob and fast-glob working in difference ways for some characters sequences, also fast-glob have API for which cases and we can reduce code on our side

fisker:
@evilebottnawi 

To resolve #6854 , I thought we are going to try test if pattern is a file first, then pass to glob?

I was going to do this way, but I saw you comment `WIP`

evilebottnawi:
@fisker I looked at code of fast-glob fastly, looks like fast-glob already handle which patterns on own side, so I think the issue was resolved after update update globby, anyway we should add tests

thorn0:
Testing if the pattern is a file before treating it is a glob seems to have nothing to do with changing or updating a library to process globs. Pseudocode:
```js
if (fileExists(pattern)) {
  processFile(pattern);
} else {
  processGlobUsingWhateverLibraryWeLike(pattern);
}
```
Am I missing something?

evilebottnawi:
@thorn0 

Pseudo:

https://github.com/mrmlnc/fast-glob#isdynamicpatternpattern-options

```js
patterns.map((pattern) => {
  if (fg.isDynamicPattern(pattern)) {
    return resolveGlobPattern();
  } else {
    const stats = fs.statSync(pattern);

   if (stat.isDirectory()) {
     return resolveAndRecursiveReadDirectory(pattern);
   } else if (stat.isFile()) {
     return resolveFilePath();
   } else if (stat.isSymbolicLink()) {
     return resolveSymbolicLink();
   } else {
     throw new Error(`Can't handle "${pattern}" pattern`);
   }
  }
});
```

thorn0:
That would miss the point of https://github.com/prettier/prettier/issues/6854:

> [What is a static or dynamic pattern?](https://github.com/mrmlnc/fast-glob#what-is-a-static-or-dynamic-pattern)
> A pattern is considered dynamic if it contains the following characters:
> * `[...]`

But filenames can contain `[` and `]` too. That's why #6854 has been opened.

evilebottnawi:
@thorn0 oh, yes you are right, so pseudo should be:
```
patterns.map((pattern) => {
   const stats = fs.statSync(pattern);

   if (stat.isDirectory()) {
     return resolveAndRecursiveReadDirectory(pattern);
   } else if (stat.isFile()) {
     return resolveFilePath(pattern);
   } else if (stat.isSymbolicLink()) {
     return resolveSymbolicLink(pattern);
   } else {
     return resolveGlobPattern(pattern)
   }
  }
});
```

thorn0:
Which means #6854 is not blocked by any glob library issues, right?

evilebottnawi:
@thorn0 In theory - yes, but `resolveAndRecursiveReadDirectory` can be done using `globby`, `node-glob` has some bugs on windows with directories (spaces in nested directory names and some characters in filename throw an error).

There will also likely be conflicts, to avoid this I asked @fisker focus on that PR instead refactor some parts of our code.

fisker:
@evilebottnawi We might not going use `fast-glob` in v2.0.0, see https://github.com/prettier/prettier/pull/6776#issuecomment-579770505

#6854 should be a simple one to fix, just add `fileExists` check before globbing files

evilebottnawi:
@fisker no, `fileExists` is not enough, we should handle directories with special characters too

fisker:
I don't think we need that, `[id].ts` should only match

if  has `[id].ts`

```text
[id].ts
a/i.ts
b/d.ts
[xxxx]/i.ts
```

if no `[id].ts`

```text
a/i.ts
b/d.ts
[xxxx]/i.ts
```

Did you mean we should also match `a/[id].ts` ?

thorn0:
#6854 is only about files. The major use case is that tools like lint-staged pass full file paths, but Prettier sometimes mistakenly treats them as globs. It'd be good to support directories (including `.`) too, but it's not in the scope of #6854.

evilebottnawi:
Not only `[`/`]` character have problems, filename on linux can have `*` character, for example `example**`, but we resolve that as glob and it is invalid

evilebottnawi:
@thorn0 We can't solve that issue without handling directories, we don’t know what the developer gave us, it can be either a file, a directory, or a glob.

fisker:
> Not only `[`/`]` character have problems, filename on linux can have `*` character, for example `example**`, but we resolve that as glob and it is invalid

In this case I think we check `example**` is a file and give result

```text
./example**
./a/example1
./a/example2
```

should good enought, or don't do glob if find file, just

```text
./example**
```

evilebottnawi:
`example**` can be directory too

fisker:
Make sense, how about if `example**` is file. give that file, if `example**` is dir give all files under it?

thorn0:
@evilebottnawi Strictly speaking, you're right, but from the practical point of view supporting files is much more important because it will fix the broken lint-staged use case.

evilebottnawi:
@fisker https://github.com/prettier/prettier/pull/6776#issuecomment-581363613 it was here that I proposed it :smile:

fisker:
> @fisker [#6776 (comment)](https://github.com/prettier/prettier/pull/6776#issuecomment-581363613) it was here that I proposed it 😄

We don't need care about  special characters  any more, we can `glob('**/*', {cwd: 'example**'})` right?

evilebottnawi:
@thorn0 In theory we can solve that in two difference PRs, but for me it was better to do in one, besides implementation does not look difficult

evilebottnawi:
> https://github.com/prettier/prettier/pull/6776#issuecomment-581383774

I think yes, but need add couple tests, node-glob really buggy in some cases, especial with glob characters in names

thorn0:
I feel that supporting directories might be not as simple as it seems. I'd prefer separate PRs.

evilebottnawi:
@thorn0 solution from `https://github.com/prettier/prettier/pull/6776#issuecomment-581383774` should work, anyway I am glad to see two difference PRs too

fisker:
Actually, might some other problem, we relay on `isFile()`, what if this file is program generated file. 

```
git clone some/repo
prettier repo/example**.js
```

it should format `repo/example**.js`, but that repo delete this file some day, we will format all his `exampleXXXX.js` file

thorn0:
@fisker So we have to make a trade-off between the problem you described and broken lint-staged.  I think the choice is clear.

evilebottnawi:
@fisker Can we try https://github.com/mrmlnc/fast-glob/issues/156#issuecomment-581838152 ?

fisker:
Notice that the package we are really using is `globby`

2 choices:

1. use `resolution` to pin `fast-glob` version
2. need investigate difference of those two packages, use `fast-glob` directly

evilebottnawi:
@fisker We can use `resolution` for testing performance

fisker:
I 'll try.

fisker:
@mrmlnc 

The beta version seems much faster then before, on our `{prettier-plugin-*,@*/prettier-plugin-*,@prettier/plugin-*}` case, this pattern used very slow on old `fast-glob` versio, almost need 1h to complete all tests.

No hard feelings, but I had to say seems still a little bit slower then `node-glob`.

[Azure pipeline job on `next`(we are working on this branch)](https://dev.azure.com/prettier/prettier/_build/results?buildId=5059&view=results)

[Azure pipeline job on this PR](https://dev.azure.com/prettier/prettier/_build/results?buildId=5069&view=results)

time seems take longer on this PR.


Can you help me check, if there any way to improve? Like is there any difference between single pattern and patterns array, or should try other option.

fisker:
@evilebottnawi I think I've done pretty much all I can do. You can try implement what you want to do if you want.

evilebottnawi:
@fisker let's wait @mrmlnc answer

mrmlnc:
I looked at the reports. Can I run something locally? Maybe some unit tests or performance tests? Right now, it is difficult for me to say anything, because I do not know this project well and do not see any benchmarks in CI report.

Unfortunately, right now I don't have any ideas for performance improvements for your case. Yeah, the fast-glob package may be slower than its predecessors in some cases. I have a few ideas for future improvements in General, but now they are very difficult to implement.

fisker:
I guess you can clone this repo and `yarn install` , then glob `{prettier-plugin-*,@*/prettier-plugin-*,@prettier/plugin-*}/package.json` in `node_modules` , this is the real tests running case

fisker:
https://github.com/fisker/prettier/blob/7242e4f551818fce37bcb622195e331e4f847c87/src/common/load-plugins.js#L95-L105

here is the actually code, I'm using

mrmlnc:
I found a bug in the matcher if `cwd` is specified. I will fix it. Perhaps this will give some kind of boost.

Looks like in your case the `fast-glob` package slower by 5ms (10 vs 15). These are single launches, so it's hard for me to say how significant these numbers.

mrmlnc:
Confirm. The bug leads to reading directories that we should not read. I'll trying to fix it in a week.

fisker:
@mrmlnc Great, thanks.

mrmlnc:
I published a new beta (`fast-glob@beta.2`). Actually, `fast-glob` is 2 times slower than `node-glob` in this scenario. Sorry I don't have more to offer right now. In the future, of course, I plan to improve this indicator.

fisker:
Result:

[Azure pipeline job on `next` branch](https://dev.azure.com/prettier/prettier/_build/results?buildId=5096&view=results)

[Azure pipeline job on this PR (with fast-glob@3.2.0-beta)](https://dev.azure.com/prettier/prettier/_build/results?buildId=5069&view=results)

[Azure pipeline job on this PR (with fast-glob@3.2.0-beta.2)](https://dev.azure.com/prettier/prettier/_build/results?buildId=5125&view=results)


Seems `fast-glob@3.2.0-beta.2` almost same as `node-glob`, thank you @mrmlnc .

As we can see in most cases `fast-glob` is faster than `node-glob`, see https://github.com/prettier/prettier/pull/6748#issuecomment-548640881 and https://github.com/prettier/prettier/pull/6748#issuecomment-548702419 .

I guess we can improve `findPluginsInNodeModules` function later, actually we don't even need glob there.

evilebottnawi:
@fisker what are our next actions?

fisker:
@evilebottnawi I think we can merge this when we got a stable release after this beta version.

evilebottnawi:
@fisker @mrmlnc already released `3.2.0` :smile:

fisker:
@evilebottnawi No, he didn't.

image

evilebottnawi:
Oh, yes, I was wrong https://github.com/mrmlnc/fast-glob/issues/251

fisker:
@evilebottnawi @thorn0 

We got the `fast-glob` version we want, should good to merge.


Thanks to @mrmlnc @brodybits

src/cli/util.js
```diff
@@ -398,6 +398,10 @@ function createIgnorerFromContextOrDie(context) {
 }
 
 function eachFilename(context, patterns, callback) {
+  // workaround for fast-glob on Windows ref:
+  // https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows
+  patterns = patterns.map(path => path.replace(/\\/g, "/"));
```
thorn0: Please wrap this into `if (path.sep === "\\")`.

fisker: Done.

src/cli/util.js
```diff
@@ -398,6 +398,12 @@ function createIgnorerFromContextOrDie(context) {
 }
 
 function eachFilename(context, patterns, callback) {
+  // workaround for fast-glob on Windows ref:
+  // https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows
+  if (path.sep === "\\") {
+    patterns = patterns.map(path => path.replace(/\\/g, "/"));
+  }
```
evilebottnawi: It is bad idea, on linux I can use `\` as part of directory or file name

thorn0: That's exactly why I asked to wrap this replace into `if (path.sep === "\\")`.

evilebottnawi: What about WSL? It is allow to use linux filenames and run commands on windows and verse vice. Yes we can try to keep backward compatibility but I think we should throw a warning on using `\` as part of glob

thorn0: No problems with WSL as far as I can see. It behaves just like 'normal' Linux. You can create a file named `\` there, but if you look at that file from Windows it won't be named `\`. The names are somehow mapped.

Please describe the scenario that you worry about. I don't understand. What exactly can go wrong? Are you simply uncomfortable about the fact that commands with `\` aren't cross-platform compatible?

evilebottnawi: Please describe the scenario that you worry about. I don't understand. What exactly can go wrong?

Glob **always**should use `/`, please read https://github.com/micromatch/micromatch#backslashes

> To solve this, you might be inspired to do something like 'foo\\*'.replace(/\\/g, '/'), but this causes another, potentially much more serious, problem.

I can spend time investigation in detail why is dangerous, but I also prefer to trust the documentation and not make some mistakes that we are warned about.

evilebottnawi: I think we must ask developers to use `/` for glob patterns, it is part of breaking changes

thorn0: I know exactly what they mean. See [this comment](https://github.com/prettier/prettier/pull/7587#discussion_r379938060). But I think it's okay. Not accepting `\` as a path separator on Windows is not okay. That's a trade-off.

evilebottnawi: Hm, maybe you are right, eslint has same logic https://github.com/eslint/eslint/blob/cf38d0d939b62f3670cdd59f0143fd896fccd771/lib/cli-engine/cli-engine.js#L711

Also look at that https://github.com/eslint/eslint/blob/cf38d0d939b62f3670cdd59f0143fd896fccd771/lib/cli-engine/cli-engine.js#L708

evilebottnawi: Maybe we should use `path.normalize` to allow using `../../foo`

thorn0: It's a horrible mess, but everyone seems to be okay with it. So I guess we should be okay too.

evilebottnawi: Let's move tests for `normalize` to https://github.com/prettier/prettier/pull/7587

evilebottnawi: Also we will add more tests when implement `expand directories`, so :+1: on that

mrmlnc:
JFYI: This problem may affect your package: https://github.com/mrmlnc/fast-glob/issues/253

thorn0:
I think we're okay. Our yarn.lock contains only picomatch 2.2.1. The "direct install from GitHub" use case can be broken though.

PR #6781, opened 2019-11-02
[WIP] Update `postcss-selector-parser` to v6

fisker:
finnally 


tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -676,7 +676,7 @@ $KeepTopLevelVar: val;
   @KeepLessVar: val;
   @extend %KeepPlaceholderSelector;
 
-  #{$Keep}[#{$Keep}][#{Keep}^="#{Keep-1A}"]:#{$Keep} {
+  #{$Keep}[#{Keep}][#{Keep}^="#{Keep-1A}"]:#{$Keep} {
```
fisker: this might not right ,  https://github.com/postcss/postcss-selector-parser/issues/211

tests/css_attribute/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -218,7 +218,7 @@ a[id='test'] {
 }
 a[id='test'] {
 }
-a[id=func('foo')] {
+a[id=func("foo")] {
```
lydell: These should have been single quotes, right?

fisker: not sure  will take a look  later

fisker: `postcss-selector-parser` throws when parsering this selector `a[id=func("foo")]` 

```text
Thrown:
Error: Unexpected "(" found.
```
so prettier will not touch it

And I've nerver seen a selector like this.

lydell: Yeah, it’s a weird test case. I wonder where it comes from. Neither SCSS nor Less seem to accept it. Maybe we can just remove it.

src/language-css/printer-postcss.js
```diff
@@ -944,6 +961,22 @@ function printCssNumber(rawNumber) {
   );
 }
 
+function preferRowsValue(node) {
+  return (node.raws && node.raws.value) || node.value;
+}
+
+function splitPlaceholder(value) {
```
fisker: planning removing this , instead split every where  I'm going to refactor placeholder logic

thorn0:
@fisker Did you see this issue https://github.com/postcss/postcss-selector-parser/issues/142? The problems mentioned there should be added to the tests.

fisker:
@thorn0 thanks for this information, will check  later

tests/css_case/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: It is serious problem, a lot of projects uses interpolate for attribute selectors

tests/template_literals/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -592,7 +592,7 @@ printWidth: 80
    div.\${expr} {
     color: red;
    }
-  all\${expr} {
+  \${expr} {
```
evilebottnawi: Where is `all`?

tests/css_case/__snapshots__/jsfmt.spec.js.snap
fisker: See https://github.com/postcss/postcss-selector-parser/issues/211

fisker:
This PR should not merge, many problems not solved yet.

fisker:
Close since it's hard to upgrade, maybe take another shoot later.

PR #6788, opened 2019-11-03
JSDoc type added in `src/common/util.js`

Shinigami92:
Added types based on https://prettier.io/docs/en/plugins.html#utility-functions

image

Referencing #6703 

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -229,26 +229,28 @@ defaultOptions: {
 A `util` module from Prettier core is considered a private API and is not meant to be consumed by plugins. Instead, the `util-shared` module provides the following limited set of utility functions for plugins:
 
 ```ts
+type Quote = "\"" | "'";
+type SkipOptions = { backwards?: boolean };
 getMaxContinuousCount(str: string, target: string): number;
 getStringWidth(text: string): number;
-getAlignmentSize(value: string, tabWidth: number, startIndex: number): number;
+getAlignmentSize(value: string, tabWidth: number, startIndex?: number): number;
 getIndentSize(value: string, tabWidth: number): number;
-skip(chars: string|RegExp): number;
-skipWhitespace(text: string, index: number, options: object): number;
-skipSpaces(text: string, index: number, options: object): number;
-skipToLineEnd(text: string, index: number, options: object): number;
-skipEverythingButNewLine(text: string, index: number, options: object): number;
-skipInlineComment(text: string, index: number): number;
-skipTrailingComment(text: string, index: number): number;
-skipNewline(text: string, index: number, options: object): number;
-hasNewline(text: string, index: number, options: object): boolean;
-hasNewlineInRange(text: string, start: number, start: number): boolean;
-hasSpaces(text: string, index: number, options: object): number;
-makeString(rawContent: string, enclosingQuote: string, unescapeUnnecessaryEscapes: boolean): string;
-getNextNonSpaceNonCommentCharacterIndex(text: string, node: object, options: object): number;
+skip(chars: string | RegExp): (text: string, index: number | false, opts?: SkipOptions) => number | false;
+skipWhitespace(text: string, index: number | false, opts?: SkipOptions): number | false;
+skipSpaces(text: string, index: number | false, opts?: SkipOptions): number | false;
+skipToLineEnd(text: string, index: number | false, opts?: SkipOptions): number | false;
+skipEverythingButNewLine(text: string, index: number | false, opts?: SkipOptions): number | false;
+skipInlineComment(text: string, index: number | false): number | false;
+skipTrailingComment(text: string, index: number | false): number | false;
+skipNewline(text: string, index: number | false, opts?: SkipOptions): number | false;
+hasNewline(text: string, index: number | false, opts?: SkipOptions): boolean;
+hasNewlineInRange(text: string, start: number, end: number): boolean;
+hasSpaces(text: string, index: number, opts?: SkipOptions): boolean;
+makeString(rawContent: string, enclosingQuote: Quote, unescapeUnnecessaryEscapes?: boolean): string;
+getNextNonSpaceNonCommentCharacterIndex<N>(text: string, node: N, locEnd: (node: N) => number): number | false;
 isNextLineEmptyAfterIndex(text: string, index: number): boolean;
-isNextLineEmpty(text: string, node: object, options: object): boolean;
-isPreviousLineEmpty(text: string, node: object, options: object): boolean;
+isNextLineEmpty<N>(text: string, node: N, locEnd: (node: N) => number): boolean;
+isPreviousLineEmpty<N>(text: string, node: N, locStart: (node: N) => number): boolean;
 mapDoc(doc: object, callback: function): void;
 ```
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -230,7 +230,7 @@ A `util` module from Prettier core is considered a private API and is not meant 
 
 <!-- prettier-ignore -->
 ```ts
-type Quote = "\"" | "'";
+type Quote = '"' | "'";
 type SkipOptions = { backwards?: boolean };
 function getMaxContinuousCount(str: string, target: string): number;
 function getStringWidth(text: string): number;
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -231,27 +231,27 @@ A `util` module from Prettier core is considered a private API and is not meant 
 ```ts
 type Quote = "\"" | "'";
 type SkipOptions = { backwards?: boolean };
-getMaxContinuousCount(str: string, target: string): number;
-getStringWidth(text: string): number;
-getAlignmentSize(value: string, tabWidth: number, startIndex?: number): number;
-getIndentSize(value: string, tabWidth: number): number;
-skip(chars: string | RegExp): (text: string, index: number | false, opts?: SkipOptions) => number | false;
-skipWhitespace(text: string, index: number | false, opts?: SkipOptions): number | false;
-skipSpaces(text: string, index: number | false, opts?: SkipOptions): number | false;
-skipToLineEnd(text: string, index: number | false, opts?: SkipOptions): number | false;
-skipEverythingButNewLine(text: string, index: number | false, opts?: SkipOptions): number | false;
-skipInlineComment(text: string, index: number | false): number | false;
-skipTrailingComment(text: string, index: number | false): number | false;
-skipNewline(text: string, index: number | false, opts?: SkipOptions): number | false;
-hasNewline(text: string, index: number, opts?: SkipOptions): boolean;
-hasNewlineInRange(text: string, start: number, end: number): boolean;
-hasSpaces(text: string, index: number, opts?: SkipOptions): boolean;
-makeString(rawContent: string, enclosingQuote: Quote, unescapeUnnecessaryEscapes?: boolean): string;
-getNextNonSpaceNonCommentCharacterIndex<N>(text: string, node: N, locEnd: (node: N) => number): number | false;
-isNextLineEmptyAfterIndex(text: string, index: number): boolean;
-isNextLineEmpty<N>(text: string, node: N, locEnd: (node: N) => number): boolean;
-isPreviousLineEmpty<N>(text: string, node: N, locStart: (node: N) => number): boolean;
-mapDoc(doc: object, callback: function): void;
+function getMaxContinuousCount(str: string, target: string): number;
+function getStringWidth(text: string): number;
+function getAlignmentSize(value: string, tabWidth: number, startIndex?: number): number;
+function getIndentSize(value: string, tabWidth: number): number;
+function skip(chars: string | RegExp): (text: string, index: number | false, opts?: SkipOptions) => number | false;
+function skipWhitespace(text: string, index: number | false, opts?: SkipOptions): number | false;
+function skipSpaces(text: string, index: number | false, opts?: SkipOptions): number | false;
+function skipToLineEnd(text: string, index: number | false, opts?: SkipOptions): number | false;
+function skipEverythingButNewLine(text: string, index: number | false, opts?: SkipOptions): number | false;
+function skipInlineComment(text: string, index: number | false): number | false;
+function skipTrailingComment(text: string, index: number | false): number | false;
+function skipNewline(text: string, index: number | false, opts?: SkipOptions): number | false;
+function hasNewline(text: string, index: number, opts?: SkipOptions): boolean;
+function hasNewlineInRange(text: string, start: number, end: number): boolean;
+function hasSpaces(text: string, index: number, opts?: SkipOptions): boolean;
+function makeString(rawContent: string, enclosingQuote: Quote, unescapeUnnecessaryEscapes?: boolean): string;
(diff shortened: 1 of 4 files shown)


src/common/util.js
```diff
@@ -38,6 +38,10 @@ function getPenultimate(arr) {
   return null;
 }
 
+/**
+ * @param {string | RegExp} chars
+ * @returns {(text: string, index: number, opts?: object) => any}
```
lydell: From a quick glance at the function body, doesn’t this return `number | false`? If so, the `skip*` functions below need to be updated as well.

Shinigami92:
The errors in `Check JSDoc types` are

```shell
src/common/util.js(51,9): error TS2367: This condition will always return 'false' since the types 'number' and 'boolean' have no overlap.
src/common/util.js(104,7): error TS2367: This condition will always return 'false' since the types 'number' and 'boolean' have no overlap.
src/common/util.js(105,5): error TS2322: Type 'false' is not assignable to type 'number'.
src/common/util.js(124,7): error TS2367: This condition will always return 'false' since the types 'number' and 'boolean' have no overlap.
src/common/util.js(125,5): error TS2322: Type 'false' is not assignable to type 'number'.
src/common/util.js(145,7): error TS2367: This condition will always return 'false' since the types 'number' and 'boolean' have no overlap.
src/common/util.js(146,5): error TS2322: Type 'false' is not assignable to type 'number'.
```

They can be ignored at this moment and in this PR, because I don't want to touch the code in this PR

src/common/util.js
lydell: Would it make sense to try to type `opts` more exactly than `object`?

lydell: This is probably the reason for the errors you mention below.

Shinigami92: Yes, definitive 
Also `node` could be typed more exactly

But I would/want to do this 1. later and 2. in another PR

Shinigami92: But what is correct? The website or the code?

lydell: What website?

lydell: Why in a different PR?

Shinigami92: https://prettier.io/docs/en/plugins.html#utility-functions

lydell: The code.

Shinigami92: 1. later => It is `01:47 AM` at my location 😅
2. want to improve the types step by step
  and so others (like @brodybits) can work on this and help :)

Shinigami92: Oh okay 👍 

I will go to bed now 💤 
I can resolve this tomorrow

brodybits: I would really favor a TBD comment in a case like this.

Unfortunately I cannot promise to work on anything specific within the near future due to some other priorities.

I would be happy to answer any questions you may have, though.

brodybits:
Build error is because of some JSDoc type errors. You can do `yarn check-types` if you want to check it yourself (I hope I didn't get the spelling wrong).

And I think it would be ideal if we could automatically synchronize the types in the documentation and the JSDoc types somehow.

Shinigami92:
Found a nice / helpful site: https://github.com/google/closure-compiler/wiki/Generic-Types#declaring-a-bounded-generic-type

src/common/util.js
```diff
@@ -185,10 +251,22 @@ function isNextLineEmptyAfterIndex(text, index) {
...
+ * @returns {boolean}
+ */
 function isNextLineEmpty(text, node, locEnd) {
   return isNextLineEmptyAfterIndex(text, locEnd(node));
 }
 
+/**
+ * @param {string} text
+ * @param {number} idx
+ * @returns {number}
```
Shinigami92: Should this be `@returns {number}` or `@returns {number | false}`?

Shinigami92:
Current type check errors

```shell
src/common/util.js(274,5): error TS2322: Type 'number | false' is not assignable to type 'number'.
  Type 'false' is not assignable to type 'number'.
src/common/util.js(275,5): error TS2322: Type 'number | false' is not assignable to type 'number'.
  Type 'false' is not assignable to type 'number'.
src/common/util.js(276,5): error TS2322: Type 'number | false' is not assignable to type 'number'.
  Type 'false' is not assignable to type 'number'.
src/common/util.js(277,5): error TS2322: Type 'number | false' is not assignable to type 'number'.
  Type 'false' is not assignable to type 'number'.
```

src/common/util.js
lydell: This function returns `idx`. `idx` contains the result of `skip*` functions. Since they return `number | false`, so must this function, unless I’m missing something. So `@returns {number | false}`

lydell:
I think you can solve those errors by changing the name of the parameter to `passedIdx` or something and do `let idx = passedIdx` before the loop. You might need to annotate `idx` there to `number | false`.

Shinigami92:
> I think you can solve those errors by changing the name of the parameter to `passedIdx` or something and do `let idx = passedIdx` before the loop. You might need to annotate `idx` there to `number | false`.

Have made the code change
I wanted to let the name of the param be `idx`, so I used a local variable named `nextIdx`

src/common/util.js
```diff
@@ -185,41 +251,82 @@ function isNextLineEmptyAfterIndex(text, index) {
...
+/**
+ * @template N
+ * @param {string} text
+ * @param {N} node
+ * @param {(node: N) => number} locEnd
+ * @returns {string}
+ */
 function getNextNonSpaceNonCommentCharacter(text, node, locEnd) {
   return text.charAt(
+    // @ts-ignore => TBD: can return false, should we define a fallback?
```
Shinigami92: What should I do here?
I don't know if there is something like `as` / `trust-me` operator syntax for JSDoc

src/common/util.js
```diff
@@ -144,13 +180,26 @@ function skipNewline(text, index, opts) {
   return index;
 }
 
+/**
+ * @param {string} text
+ * @param {number | false} index
```
lydell: Is this function really supposed to take `number | false` as input? Is anyone calling it with `false`?

src/common/util.js
lydell: I like how the type checking finds bugs like these. `"abc".charAt(false) === "a"`. So it seems like this function returns `text[0]` if there is no “next non-space-non-comment character.”

lydell:
If you have the energy it would be great if you could update the website docs as well to mention `number | false` and `SkipOptions`.

src/common/util.js
Shinigami92: As a human I would say: no
As a compiler I would say: yes, because L251 (`isNextLineEmptyAfterIndex`) is calling it with `idx: number | false`
And it uses many of the `skip*` functions

src/common/util.js
Shinigami92: Tried it in Chrome ^^
image

To make it safe: should I add `|| 0` as fallback? Will not change runtime, but makes the compiler 😊

lydell: I think we’ll just leave this as a TBD comment like you’ve done for now.

Shinigami92:
> If you have the energy it would be great if you could update the website docs as well to mention `number | false` and `SkipOptions`.

We currently don't emit `d.ts` files out of our JSDoc
So currently there is no such thing as `SkipOptions`, it's just virtual and is not `export`ed and usable by anyone out there

I would suggest to let us define and improve the code step by step with JSDocs
and after that we could try to emit `d.ts` files (with TS 3.7 https://github.com/microsoft/TypeScript/wiki/Roadmap#37-november-2019, https://github.com/Microsoft/TypeScript/issues/29490, https://github.com/Microsoft/TypeScript/issues/7546) and update the docs

lydell:
I meant that we could mention what `SkipOptions` is in the docs. Feels easy enough to update them manually for now?

src/common/util.js
Shinigami92: OK

lydell: For others reading, see https://github.com/prettier/prettier/pull/6788#discussion_r341847047

src/common/util.js
lydell: Then it sounds to me like L251 should be:

```js
return idx === false ? false : hasNewline(text, idx)
```

And that `hasNewline` should only accept `number` for `idx`.

Shinigami92:
What should I do with
```shell
src/main/comments.js(513,28): error TS2345: Argument of type 'number | false' is not assignable to parameter of type 'number'.
  Type 'false' is not assignable to type 'number'.
```

lydell:
Maybe:

```js
const index = skipNewline(text, options.locEnd(comment);
if (index !== false && hasNewline(text, index))
```

Shinigami92:
What is failing there?
How can I fix this?

lydell:
We have add more conditions / code paths, but have not added tests for them. From a type perspective the conditions are needed, but in reality we might never hit those edge cases in the current implementation. I think it should be fine merging this even if it means -0.02% coverage.

Shinigami92:
Ok, then this PR is ready to merge 🚀

docs/plugins.md
```diff
@@ -229,26 +229,28 @@ defaultOptions: {
 A `util` module from Prettier core is considered a private API and is not meant to be consumed by plugins. Instead, the `util-shared` module provides the following limited set of utility functions for plugins:
 
 ```ts
+type Quote = "\"" | "'";
+type SkipOptions = { backwards?: boolean };
 getMaxContinuousCount(str: string, target: string): number;
 getStringWidth(text: string): number;
```
lydell:
```suggestion
function getStringWidth(text: string): number;
```

Can you do this on all the functions below? Then this code block becomes valid TypeScript, which means that Prettier will format it!

Shinigami92: Yes I had the same idea ^^

lydell: Nice. You forgot to run Prettier on the file though ;)

Shinigami92: Oh... common
I used `Save without format` so the lines don't wrap 😂
Should I add `<!-- prettier-ignore -->` or should I format it?

lydell: Yeah, maybe it makes sense to add prettier-ignore.

src/common/util.js
```diff
@@ -161,7 +210,16 @@ function hasNewlineInRange(text, start, end) {
...
+ * @template N
+ * @param {string} text
+ * @param {N} node
+ * @param {(node: N) => number} locStart
+ */
 function isPreviousLineEmpty(text, node, locStart) {
+  /**
+   * @type {number | false}
+   */
   let idx = locStart(node) - 1;
```
thorn0: Please don't add types to local variables. They are inferred by the typechecker and are not part of API.

src/common/util.js
```diff
@@ -182,44 +248,85 @@ function isNextLineEmptyAfterIndex(text, index) {
...
-    idx = skipTrailingComment(text, idx);
-    idx = skipNewline(text, idx);
-  }
-  return idx;
-}
-
+  /**
+   * @type {number | false}
+   */
+  let nextIdx = idx;
```
thorn0: However, if you need to override the inferred type like in this case, it's a valid reason to specify a type for a variable.

thorn0: Lets's use a more compact syntax:
```js
let nextIdx = /** @type {number | false} */ (idx);
```

brodybits: I would not favor adding extra parens unless absolutely necessary, which I think is not the case here. The following solution works for me (passes `yarn check-types`):

```js
  /** @type {number | false} */
  let nextIdx = idx;
```

src/common/util.js
```diff
@@ -182,44 +248,85 @@ function isNextLineEmptyAfterIndex(text, index) {
   }
   idx = skipTrailingComment(text, idx);
   idx = skipNewline(text, idx);
-  return hasNewline(text, idx);
+  return idx === false ? false : hasNewline(text, idx);
```
brodybits:
```suggestion
  return idx !== false && hasNewline(text, idx);
```

src/common/util.js
```diff
@@ -144,13 +180,26 @@ function skipNewline(text, index, opts) {
...
 
+/**
+ * @param {string} text
+ * @param {number} index
+ * @param {SkipOptions=} opts
+ * @returns {boolean}
+ */
 function hasNewline(text, index, opts) {
   opts = opts || {};
+  // @ts-ignore => TBD: false - 1 is not allowed
```
brodybits: I think this @ts-ignore comment is not needed; `yarn check-types` seems OK with this comment removed.

src/common/util.js
brodybits: I would also favor adding something similar for the `let oldIdx` declaration on the line above, for the sake of someone reading the code for the first time, and some extra spacing. So I think the first few lines of the function could look something like this:

```js
function getNextNonSpaceNonCommentCharacterIndexWithStartIndex(text, idx) {
  /** @type {number | false} */
  let oldIdx = null;

  /** @type {number | false} */
  let nextIdx = idx;

  while (nextIdx !== oldIdx) {
    oldIdx = nextIdx;
```

src/common/util.js
lydell: This is needed to tell TypeScript that `idx` won’t just be a number, but can also be `false`?

thorn0: Right, sorry.

src/common/util.js
lydell: That’s correct! We forgot to remove it when this problem was fixed.

docs/plugins.md
```diff
@@ -228,28 +228,31 @@ defaultOptions: {
...
-hasNewline(text: string, index: number, options: object): boolean;
-hasNewlineInRange(text: string, start: number, start: number): boolean;
-hasSpaces(text: string, index: number, options: object): number;
-makeString(rawContent: string, enclosingQuote: string, unescapeUnnecessaryEscapes: boolean): string;
-getNextNonSpaceNonCommentCharacterIndex(text: string, node: object, options: object): number;
-isNextLineEmptyAfterIndex(text: string, index: number): boolean;
-isNextLineEmpty(text: string, node: object, options: object): boolean;
-isPreviousLineEmpty(text: string, node: object, options: object): boolean;
-mapDoc(doc: object, callback: function): void;
+type Quote = "\"" | "'";
```
lydell:
```suggestion
type Quote = '"' | "'";
```

docs/plugins.md
thorn0: Why avoid the wrapping? It'd look better than scrolling: https://deploy-preview-6788--prettier.netlify.com//docs/en/plugins.html#utility-functions

Shinigami92: Very opinionated
I (as a consumer) prefer to be able to do Side-Scroll in the preview window

src/common/util.js
```diff
@@ -108,6 +138,12 @@ function skipTrailingComment(text, index) {
 // This one doesn't use the above helper function because it wants to
 // test \r\n in order and `skip` doesn't support ordering and we only
 // want to skip one newline. It's simple to implement.
+/**
```
lydell: Should this comment be part of the JSDoc? Same question for some comments further down.

Shinigami92: This comment -> I think no
Because it is really an internal comment
Or in other words: When I want to call this function, I did not expect the content of this comment

Shinigami92:
🎉

Issue #6469, opened 2019-09-11
Release 1.19 checklist


lipis:
- prepare
  - [x] Complete [milestone items](https://github.com/prettier/prettier/milestone/32)
  - [x] Run on other projects to check for regressions
  - [x] Write release notes
    - #6787 add a Markdown file named [<code>2019-11-03-1.19.0.md</code>](https://github.com/prettier/prettier/blob/master/website/blog/2019-09-16-1.19.0.md) in [`website/blog`](https://github.com/prettier/prettier/tree/master/website/blog)
    - [prs-merged-since](https://npm.im/prs-merged-since) may be useful
- publish
  - [x] Run [release script](https://github.com/prettier/prettier/tree/master/scripts/release)
- postpublish
  - [x] Adjust release notes date
  - [x] Merge release notes
  - [x] Wait for release notes to deploy
  - [x] Tweet from [`@PrettierCode`](https://twitter.com/PrettierCode) account via TweetDeck
  - [x] Close milestone
  - [x] Send schema PR generated by `scripts/generate-schema.js` to [SchemaStore](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/prettierrc.json) if we added a new option.
  - [x] Update the [release checklist](https://github.com/prettier/prettier/wiki/Release-Checklist) with possible changes – opened #6887
  - [x] Close #3888

Then monitor [Twitter](https://twitter.com/search?q=%40PrettierCode&src=typed_query&f=live)/[GitHub](https://github.com/prettier/prettier/issues) for any immediate issues 😄

evilebottnawi:
I think will be great update all possible dependencies before do release

squidfunk:
Any updates on 1.19? When will it be released?

lydell:
@squidfunk  We’ve finally ticked off all blocking items in the [1.19 milestone](https://github.com/prettier/prettier/milestone/32)!

mrmckeb:
Is there a way to test this out? I'd be happy to run against a few projects to help test.

lipis:
> Is there a way to test this out? I'd be happy to run against a few projects to help test.

```
npm install prettier/prettier
# or
yarn add prettier/prettier
```

mrmckeb:
This is working well, thanks @lipis. I'll let you know if we have any issues over the coming days.

lydell:
@prettier/core If anyone could help with “Run on other projects to check for regressions” that would be amazing!

thorn0:
[...]

PR #6229, opened 2019-06-16
Printing arguments: short circuiting some cases

duailibe:
So when investigating #4784, I've managed to verify the issue was in `printArgumentsList()`, more specifically the part where we check if it should hug the first of the last argument, and later, verified the issue was in the `conditionalGroup`, and more specifically, the first possible group, with a combination of `ifBreak()` and `softline`.

The previous code had the following meaning: if the group breaks, print all arguments broken down; else, just concat the arguments with the wrapping parenthesis (note that there are some cases where we print a new line but the group doesn't break!). 

So the first thing I changed was: if none of the arguments break beforehand (this is before printing, so this means the argument "doc" already includes something that makes it break. It can break later when printing if it exceeds the print width), we can try a simple concat. If it fails, it will continue to the next cases in the conditional group. And only if one of the arguments break beforehand, we use the `ifBreak()`.. another change is instead of using `softline` for the ifBreak case, just print a hardline instead.

These changes didn't break any existing tests and fix the performance degradation.

```js
var prettier = require("./");
var code = `
[...]

PR #6791, opened 2019-11-03
disable optimization from #6229 for calls with type arguments

thorn0:
fix for https://github.com/prettier/prettier/issues/6469#issuecomment-549093687

**✨[Try the playground for this PR](https://deploy-preview-6791--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuc0DOMAEAnOaAO6cmAvJmhALZwwAWAllAOYB0AJLTDPi0zQDwAVBswCGlUQCMIkgMpxsAN3pg4LACI1R9ADZp1omKIB8ACgA6UTJgAGo-PQD0AMwgQWotLQAeLtwFoAEy1dNEc2YEoATwBJQIBfGwAaS2tgTDBRMFo4JEwARwBXQu8WTm4AYSycpMwYemoIQqx4ywBKAG5LECSQCHx69GRQUWxsCAB3AAVRhDRkEFFFCHpAnpBJbCyAaxpZfCzGJmQYbEK4Xs5KHQB1BngCLLh5KDR6euUYKIWwNHnexhoBQwKZbJgSZDOUR6C4gABWaG8ACEtmBdjBZOI4AAZRhwSHQoG9BHeWRHHRwACKhQg8AJMN6B2wQOwCy++DwYGw9AG63w3NgN1WdGQAA4AAyM8ZAm5bfALfl4BSKfG9Iq0uCg-rzFCefxQOBwYJrXq4Ir0XCg0Tg0T0okgIGUegnM6wt7MCnUjV22FGSRCwIipAAJl6p20OiOFSoEJQUGgqpAhSBgikOqhDJAinOMSgwVgsi5PJgAEE87IvhSffF4kA)✨**

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -4024,7 +4024,9 @@ function printArgumentsList(path, options, print) {
       somePrintedArgumentsWillBreak ? breakParent : "",
       conditionalGroup(
         [
-          !somePrintedArgumentsWillBreak
+          !somePrintedArgumentsWillBreak &&
+          !node.typeArguments &&
+          !node.typeParameters
             ? simpleConcat
             : ifBreak(allArgsBrokenOut(), simpleConcat),
           shouldGroupFirst
--- a/tests/break-calls/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/break-calls/__snapshots__/jsfmt.spec.js.snap
@@ -231,3 +231,23 @@ function MyComponent(props) {
 
 ================================================================================
 `;
+
+exports[`type_args.js 1`] = `
+====================================options=====================================
+parsers: ["flow", "typescript"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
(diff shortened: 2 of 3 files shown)


src/language-js/printer-estree.js
```diff
@@ -4024,7 +4024,9 @@ function printArgumentsList(path, options, print) {
       somePrintedArgumentsWillBreak ? breakParent : "",
       conditionalGroup(
         [
-          !somePrintedArgumentsWillBreak
+          !somePrintedArgumentsWillBreak &&
+          !node.typeArguments &&
+          !node.typeParameters
```
lydell: You added two things here, but only one test. Does the test cover both? I can’t tell by looking at the test because I’m unsure what the difference between `typeArguments` and `typeParameters` is.

thorn0: It's a discrepancy between the AST formats. The same thing is called `typeArguments` by Flow and `typeParameters` by `typescript`. The test runs for both parsers.

Issue #6790, opened 2019-11-03
Proposal: refactor css placeholder for TemplateLiteral


fisker:
`CSS in template literals` is first added in #2102 `@prettier-placeholder` is choosed to be a placeholder for template  quasis. 

Now problem is TemplateLiteral could be anywhere of the css code

as tag
```
${as-tag}{}
@prettier-placeholder{}
```
tag selector became a `at-rule`, but we handled in `at-rule`

as part of tag
```
foo${as-part-of-tag}{}
foo@prettier-placeholder{}

foo[${as-part-of-selector}]{}
foo[@prettier-placeholder]{}
```

`foo@prettier-placeholder` is not a legal tag,`postcss-selector-parser` can parse currectly, but might cause problem in future,

[...]

Issue #6392, opened 2019-08-13
styled component


soojeen:
If we have multiple styled components interpolated and separated by comma, the expected newline after the comma is missing.

Styled Components allow interpolation of nested styled components, see [here for the syntax](https://www.styled-components.com/docs/advanced#referring-to-other-components).

Prettier will put css selectors separated by `,` on new lines, introduced here: https://github.com/prettier/prettier/pull/2047. This works in styled components, see second block in the playground link.

**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBBTBeTDATwBs4ATACgH1sBKAAwB0pNMASYAIQF8AaDsADCPTMBZs2xMkkwA3AIYkArnADcEzDxbMoLSFAw58hGKQo16utgDowJBWjTYBdh065jNUszPlLVDVYtHRYQPhAIAAcYAEt0ZFAFACdkiAB3AAUUhDRkEAU5CFjycJAAI2SFMABrOBgAZSjq2KgAc2QYZNUIgAsYAFsSAHVe2Pg0ZrA4Btzx2LlxonywJzLWtDhkmEyqtoGFZAAzJU2IgCs0AA8uKtr6hoUBuAAZVrhj07gL64bWtrIAEVlBB4J8SGcQM1kptkvlygpynASGUoslWjBhiUYL1kAAOAAMETREE2wyqUXyaLgsLkHwiAEcQfBdtE8ihHABaKBwCgUMrJOBM2KC3YKfaHJAnCHfECbAaxTrdWVof5A5kfKVfCIwRFY8g45AAJh1VViJH+QggAwO+Sg0HpIGUmwAKoj2dLNjweEA)
```sh
--parser babel
```

**Input:**
```jsx
const A = styled(_A)`
  ${B}, ${C} {
    style: value;
  }
`

const A = styled(_A)`
  .classA, .classB {
    style: value;
  }
`

```

**Output:**
```jsx
const A = styled(_A)`
  ${B}, ${C} {
    style: value;
  }
`;

const A = styled(_A)`
  .classA,
  .classB {
    style: value;
  }
`;

```

**Expected behavior:**
```jsx
const A = styled(_A)`
  ${B},
  ${C} {
    style: value;
  }
`;

const A = styled(_A)`
  .classA,
  .classB {
    style: value;
  }
`;

```

fisker:
minimum reproduction

**Prettier 1.18.2**
[...]

Issue #6259, opened 2019-06-30
breaks CSS in JS ${Cls}:not() -> ${Cls}:not ()


brandonkal:
**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACSKAZ3wHFsIBXDPAXjxIE8AbOAEwDo2BLANwAMAOlDx4AtgENsAc25QkeAAwBuYcNEc5GSvmDrR4qbPlLVIvAF815gCTAAyjBbsAklG0wLSKBBgAKJAAzbmwSAFowAAtuZjYASjw9c1FJGTkFFX0rKH0AMjwAPjxvXz8OOy0dAHVsCQwLOJL-OwAhHRhoBsT9UR4iDGYJRgVA1jQzUWyhKBAAGhAIDBhuaCJkUCkKAHcABSkENZQJXghuNjmQACNasABrOBh7DAkwOWlkGGxKOHnImDFmFVovB+i84PYDtxlrwoYxkOAiGt5nIiHBcDtatJJMhAhJmKj5gArIhoFo3e6PCRiOAAGTkcBxeIJIGJaHsb1YAEVKL4GUEmT8QM9Qmj4U4MHAiGBsNwlhcMDLYFUzjBIsgAByKeYKiComp1eEKyVo3gM+YARx58Axi0OIAkRDCUDg7HYF2wcEtITgGIkWIkjPxgtRYm4Hy+wY5cG5vMDzJgEkuyrYquQACZ5p8JDE3gBhCBibEoHzOi6UVEAFUTh1xQYsFiAA)
```sh
--parser typescript
```
I am using Typescript but this does not matter here.

**Input:**
```tsx
export const Group = styled.div`
  margin: 0;

  .input {
    margin: 0;
  }

  ${StyledInput}:not(:first-child) {
    margin: 0;
  }

  & > :not(.${inputWrap}):not(${Button}) {
    display: flex;
  }
`

```

**Output:**
```tsx
export const Group = styled.div`
  margin: 0;

  .input {
    margin: 0;
  }

  ${StyledInput}:not (:first-child) {
    margin: 0;
  }

  & > :not(.${inputWrap}):not(${Button}) {
    display: flex;
  }
`;

```
**Expected behavior:**
No space should be inserted after not as this breaks the CSS.

Updated with minimal reproduction. It appears that having a child selector is required. Oddly, the space is not always inserted as you can see in the last declaration.

[...]

Issue #5961, opened 2019-03-12
Formatting styled component for nth-child() is wrong 


wooxudong:
**Input:**
```jsx
const Steps = styled.div`
  @media (min-width: ${breakpoints.lg}) {
    ${Step} {
      margin-bottom: 90px;
    }

    ${Step}:nth-child(odd) {
      ${StepItemDescription} {
        grid-row: 1;
        grid-column: 3 / span 3;
      }
      ${Image} {
        grid-row: 1;
        grid-column: 7 / span 6;
      }
    }

    ${Step}:nth-child(even) {
      ${Image} {
        grid-row: 1;
        grid-column: 3 / span 6;
      }
      ${StepItemDescription} {
        grid-row: 1;
        grid-column: 10 / span 3;
      }
    }
  }
`;
```

**Output:**
```jsx
const Steps = styled.div`
  @media (min-width: ${breakpoints.lg}) {
    ${Step} {
      margin-bottom: 90px;
    }

    ${Step}:nth-child (odd) {
      ${StepItemDescription} {
        grid-row: 1;
        grid-column: 3 / span 3;
      }
      ${Image} {
        grid-row: 1;
        grid-column: 7 / span 6;
      }
    }

    ${Step}:nth-child (even) {
      ${Image} {
        grid-row: 1;
        grid-column: 3 / span 6;
      }
      ${StepItemDescription} {
        grid-row: 1;
        grid-column: 10 / span 3;
      }
    }
  }
`;
```

**Expected behavior:**
Notice the space added between the `nth-child (odd)`, this would not make the nth-child selector work. I am not sure how I can use prettier but still make this work.

fisker:
minimum reproduction

**Prettier 1.18.2**
[...]

PR #6794, opened 2019-11-04
Refactor css placeholder for TemplateLiteral 

fisker:
fixes: #6790 , with this changes, `src/language-css/printer-postcss.js` can remove all placeholder logic, ~will clean later~ update: already did

fixes: 
- #6259 #4520
- #5465 #5614
- #5219 
- #6392 
- #5961
- #5886
- #4355
- #2636
- #2350
- #2883


fisker:
need help here, does any one know: where is this function been try/catch , can't find the SyntaxError

https://github.com/prettier/prettier/blob/12ba99720714ef965c2c9580a6d922519bc282f8/src/language-js/embed.js#L47

tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -264,17 +265,18 @@ styled.div\`
 \`;
 
 styled.span\`
-  \${foo}
-  \${bar}
+  \${foo};
```
fisker: Anyone know is this going tobe a problem?

fisker:
ready for review

tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
lydell: https://github.com/prettier/prettier/issues/2350
https://github.com/prettier/prettier/issues/2636

fisker: if you're sure about this I'll try to remove this extra semi, it can be done

lydell: Yes, that would be great!

tests/template_literals/styled-jsx-with-expressions.js
```diff
@@ -23,7 +221,7 @@
     color: red;
   }
   }
-  @font-face {
+@font-face {
```
fisker: also this is not right, seems this at-rule with 
`a:placeholder;` (we are faking single placeholder to this) cant parse as expected, is there other thing we can fake? tried  fake into at-rule and comment , some other problem come up

fisker: misstaken, not snap

tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
fisker: done.

fisker:
ready for review

fisker:
got a error from [playground](https://deploy-preview-6794--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBnGBPANnAJgOlwEsA3AAwBJgBDAXyoCN7gxayBuEAGhAgAcYRaKmShqAJ3EQA7gAUJCESmokIRXNxANx1MAGs4MAMp9dRKAHNkMcQFc4PABYwAtlgDqjovFSmwcI0VvUm8MZHBUER5zVDhxGFkdCxdqZAAzaixYngArVAAPACEdfUMjahc4ABlzOHTM7JA8-KNzCxwARVsIeHqshxBTcVjxcIZqBjgsTT5xcxh3dRhHZAAOAAYeWYhY9x0+cNm4EZI6ngBHbvhE-iUQalQAWig4PDxNcThLok-E6mTUkgMv0eLEXERrHYBqg2p0rnUgQ0BjAJotcMtkAAmHg2ahELBtADCEBcKXCUGgZxAtliABUJkpgY0SPYAJJQXAIYxgOYCACCHKMmBwfVitFoQA), maybe there is a length limitation

fisker:
@evilebottnawi  #6790

fisker:
`src/language-js/printer-estree.js` seems lost a lot coverage, there might some dead code inside, but I'm not able to check on [codecov](https://codecov.io/gh/prettier/prettier/compare/12ba99720714ef965c2c9580a6d922519bc282f8...5bd8d4424be46ecc6b2527e13110b578f2383ad1/changes) maybe because this file is too big

tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -287,11 +289,11 @@ styled.span\`
 \`;
 
 styled.span\`
-  \${foo}: \${bar}
+  \${foo}: \${bar};
 \`;
 
 styled.span\`
-  \${foo}: \${bar}
+  \${foo}: \${bar};
```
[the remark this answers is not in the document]
fisker: @evilebottnawi if we consider this case, previously we remove the semi also not right, exta semi is better, but

```js
const color= 'red'
const width = '1px'
const Button = styled.a`
  color: ${color}
  width: ${width}
`
```
remove this might cause problem, am I right?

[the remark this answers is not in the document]
lydell: Why would anyone ever put a semicolon in `test` if they intend to use it as a property value? I could just as well see people reporting _not_ inserting the semicolon as a bug: “Why is `color: red` turned into `color: red;`, but not `color: ${test}` into `color: ${test};`?”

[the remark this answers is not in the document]
evilebottnawi: @lydell so https://github.com/prettier/prettier/issues/2636 not solved, but it is in issues list (above), as I said earlier, I'm pretty sure there exists edge case when we break code, but it’s not so easy to solve, I understand, so maybe the solution is better than now

evilebottnawi:
Thanks for great work! I will look and play with this tomorrow

tests/template_literals/styled-jsx-with-expressions.js
```diff
@@ -23,9 +329,9 @@
     color: red;
   }
   }
-  @font-face {
-    ${expr}
-  }
+         @font-face {
+${expr}
+}
```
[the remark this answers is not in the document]
fisker: no that's the source, I put some extra space to confirm this works, original one can't see if it formated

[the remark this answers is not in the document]
fisker: snap on Line 1390

tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
fisker: @evilebottnawi [playground](https://deploy-preview-6794--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACSKAZ3wCEBXGGaAdWwEMMM5s8BePEgTwBs4ATAHQAjStSgADADpQ8eACTBh9InAC+MuYoAWEAG4sNshcAgZ6AR3LrNJ4TwhgA1ka3B6YGAEsDrk-y8iensBP0UISh4vKBtjRX4VbVDbRSJzMGiAcyMJAG4QABoQM29oImRQemxsCAB3AAUqhHKUej0IL35CkGEGZzgYAGV0rOQYbGsi7RgAWx4abS94NI84QealnyWuZHAicqLo1Vx6hkyZ+mQAM3oeVSKAKyI0Uj6nAcH6GbgAGWi4a63e4gJ5oQZZPgARXIEHggLucCK5mwx12ymEcB43Qw2GiMBonRg2mQAA4AAxImqqOiMXY4uDHAzdKywuCnMwtEAqAC0MQEAm62DgVi8QtO9HOlyQNwRRVUMy8YwmiJARAhcGhrPhwJgwQJ-CJyAATEVxvQvFEoJkAMIQGYXXZQaAAorkVQAFWCLRlwL01gAklB+AghmBcRgYABBIODGC8AHSoHqNRAA) it's solved, haha, magic~~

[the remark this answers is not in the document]
fisker: different case, if you read #6790  carefule , you can understand why they are different,

add a prop before the last one, you'll see [link](https://deploy-preview-6794--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACSKAZ3wCEBXGGaAdWwEMMM5s8BePEgTwBs4ATAHQAjStSgADADpQ8eACTBh9InAC+MuYoAWEAG4sNshcAgZ6AR3LrNJ4TwhgA1ka3B6YGAEsDrk-y8iensBP0UISh4vKBtjRX4VbVDbADMICCQTInMwaIBzIwkAbhAAGhAzb2giZFB6bGwIAHcABXqEGpR6PQgvfjKQYQZnOBgAZRz85Bhsa3LtGABbHhptL3hsjzgxjvWfda5kcCIa8ujVXBaGPMX6ZBT6HlVygCsiNFJhp1Gx+kW4AAy0Tg90ezxAbzQY3yfAAiuQIPBQU84OVzNgLkdlMI4DwBhhsNEYDQ+jBtMgABwABjRjVUdEYRwJcAuBgGVkRcCuZk6IBUAFoYgIBANsHArF4xVd6Dc7kgHijyqpFl5prNUSAiDC4PDOcjwTBgiT+GTkAAmcozeheKJQPIAYQgi1uRyg0BB5XIqgAKsFOgrwXprABJKD8BDjMCEjAwACCYbGMF4IPlYPUaiAA) and [link2](https://deploy-preview-6794--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACSKAZ3wCEBXGGaAdWwEMMM5s8BePEgTwBs4ATAHQAjStSgADADpQ8eACTBh9InAC+igBYQAbizUy5iiBnoBHcuoVKeEMAGs1AbkPX6YGAEs9B2df6eRPTCfPy+RsAQlDyeUFYA9PF4UBCccAC2nq6K-CqaAs54iWmZrgBmEBBI1kSmYLEA5mpFSTr0PJYyEk4gADQgJl7QRMig9NjYEADuAArjCCMo9DoQnvx9IMIMDnAwAMp1jcgw2Jb9mjDpPDSanvC17nB7C3fed1zI4EQj-bGquDMGA10vRkGV2qp+gArIhoUjbey7Pb0dJwAAysTgYIhcGhsL2jT4AEVyBB4NieJCQKZsP9PsphHAeBsMNhYjAaGsYJpkAAOAAM-VZEFUdEYn1ZcH+eg2FjJcEBJkWIBUAFo4gIBBtsHALJ4dYD6MDQUhwZTcSBVJljqcLURCXASfKKVSYMFOfxucgAEz9E70TwxKANADCEHSIM+KTiG3IqgAKsFFmaqTpLABJKD8BD7MBsjAwACCWb2MF4WNNOLUaiAA)

fisker:
FYI:

```
div {
foo#{expr}bar;
}
```

this weird case is not handled yet, I ready add test cases and FIXME there

lydell:
@fisker I wonder what that case even means, and where it comes from.

fisker:
@lydell update  the example, i made that up.

fisker:
@lydell @evilebottnawi 

Since you guys review, I did some changes

1. some code rewrite
2. `div{foo${expr}bar;}` handled
3. split the tests

welcome  for more test cases

fisker:
@lydell @evilebottnawi

Did you guys find any problem? Should I start writing changelog and get this merged.

evilebottnawi:
@fisker sorry didn’t have time to look, but I’m doing it in the near future, no need rush, i think we can merge this for `1.20`, because `1.19` release should be done in near future and aimed at typescript 3.7

src/language-js/embed.js
```diff
@@ -16,6 +17,35 @@ const {
...
+const placeholderPiecesToStringArray = pieces =>
+  pieces.map(({ isPlaceholder, string, placeholder }) =>
+    isPlaceholder ? placeholder : string
+  );
+const removeCSSComments = string =>
+  string
+    .replace(/\/\*\s*prettier-ignore\s*\*\//g, CSS_PRETTIER_IGNORE_PLACEHOLDER)
+    .replace(/\/\/\s*prettier-ignore/g, CSS_PRETTIER_IGNORE_PLACEHOLDER)
+    .replace(/\/\*[\s\S]*?\*\//g, "")
+    .replace(/\/\/.*/g, "")
```
lydell: Can this accidentally remove things inside strings? Such as `a[href="https://foo"]`, `a { content: "/* prettier-ignore */" }`?

fisker: could be, I'll check later.

FYI: this function mean to keep `prettier-ignore`, and only used to check some before after string. logic on L114-L142

fisker: you are right, I  break [this case](https://deploy-preview-6794--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAucBnVADAOlA2gCwCc4AzAXixAHoqSIJKBdAAmGYBJg4APAB0IC+AbgE4MIADQgIvGAEtoqZKACGhQhADuABTUIlKFQDcIcgCaSQAI0IqwAazgwAyrztyoAc2QxCAVzgpfBgAWwAbAHV8OXhUNzA4Z30YuSMYgE9kNCUpD1Q4QhhtW08QlWQSFTD8qQArVG4AIVsHJ2cVELgAGQ84CqqakHruZw9PMLgART8IeH7qwJA3QnzCLKsVKzgwy34PGAjzGHxkAA4ABil+CHyI214s-jhVoz6pAEcZ+GKZAxAVVAAWigcDgZjBlmInzkxGKKlK5SQlQWUnyITkPn8i1QYwm01mfSRA0WME2hzMx2QACYpL4VHIwmMAMIQEJlLJQaBvEB+fIAFU2BmRgyMAQAklBwbBnGBCHJZABBSXOGDpCbzfICARAA)

fisker: handled

src/language-js/embed.js
```diff
@@ -25,12 +25,18 @@ const placeholderPiecesToStringArray = pieces =>
   pieces.map(({ isPlaceholder, string, placeholder }) =>
     isPlaceholder ? placeholder : string
   );
-const removeCSSComments = string =>
+
+const removeCSSCommentsAndQuotedString = string =>
   string
+    // remove quoted strings
+    .replace(/(['"]).*?\1/, "$1$1")
```
lydell: We also need to handle escapes in strings. This regex should do it:

```
(['"])(?:(?!\2)[^\\\r\n\f]|\\(?:\r\n|[\s\S]))*(\2)
```

Copied from https://github.com/lydell/css-tokens

We also need to handle unquoted URLs, such as:

```
background-image: url(https://example.com/foo.png);
```

Regex:

```
url\((?!\s*["'])\s*(?:[^"'()\\\s]|\\(?:[\da-fA-F]{1,6}\s?|[^\r\n\f]))*\s*\)
```

fisker: I understand to not break css code, escapes need to be handled, but this function only need make sure strings after `//`(not a comment) is not lost, I even strip all strings inside quotes. 

and I already add test for escaped quote

`url()` is hot handled

fisker: **Prettier pr-6794**
[Playground link](https://deploy-preview-6794--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAucBnVADAOlAhgAmACNcwBrAcwCcIBXKAEwFoBLAW1wriX1qoBsAFAAsYMAA6okAemlwAHrjbj+cAHSQ20gGYQIa8VAoBKANwASYLgC+1nBhAAaEBHEwW0VMlC4qNAO4ACr4IXii4AG4QLAxOIERUpGRwMADK4qQsRsgwVLRwzqJs-ADqwizwqBlgcKmhFSwRFQCeyGhezlmocFQwgYkUHMjauPzdzgBWqPIAQonkKalKcAAyWXDDo+MgU-KpWRSqAIq0EPCbYwUgGVTdVG0kRHD8ceJUWTAlMTDCyAAcAAZnG8IN0SolxG03nA7hENs4AI6neD9VxhEC4VBMKBwOAMPFxKhwJEsIn9ThDJAjS7ObpsFg5PJXVAHY7IjZUrZXGC4IhfBg-ZAAJmcuVwLH4BwAwhA2JSQFBoPCQLRugAVXlhanbCL5ACSjAQaTA7zcAEFGKkYM1VBdurYgA)
```sh
--parser babel
```

**Input:**
```jsx
css`
a {background-image: url(https://example.com/foo.png);${a}}
`
```

**Output:**
```jsx
SyntaxError: (postcss) CssSyntaxError Unknown word (2:55)
  1 | 
> 2 | a {background-image: url(https://example.com/foo.png);pprreettttiieerrfgfyruzjaaaaaaaaaaaaaaaarreeiitttteerrpp}
    |                                                       ^
  3 | 
```

lydell: Hmm, I guess I need to dive deeper into how this PR actually works…

fisker: this part is clean before/after to check if `templateLiteral` is a `orphan placeholder`, beause `postcss` can't parse

```css
a {
  placeholder
}
```

I need fake it into
```css
a {
  prop: placeholder;
}
```

I clean before/after comments, and check is this `templateLiteral`

```js
           (!after ||
              endsWithLineBreak ||
              after[0] === ";" ||
              after[0] === "}") &&
            (!before ||
              before.slice(-1) === ";" ||
              before.slice(-1) === "{" ||
              before.slice(-1) === "}")
```

fisker: seem I don't even need handle `url(//example.com)` because you can't wrote this in jsx,

you need escape `url(\/\/example.com)`

src/language-js/embed.js
```diff
@@ -16,6 +17,38 @@ const {
...
+    .replace(/\/\*\s*prettier-ignore\s*\*\//g, "_")
+    .replace(/\/\/\s*prettier-ignore/g, "_")
+    // remove block comments
+    .replace(/\/\*[\s\S]*?\*\//g, "")
+    // remove quoted strings
+    .replace(/(['"]).*?(\1)/g, "_")
+    // remove strings in parentheses
+    .replace(/\(.*?\)/g, "_")
+    // remove inline comments
+    .replace(/\/\/.*/g, "");
```
fisker: @lydell I renamed this function and replace all of then to `_`, It really doesn't matter :)

evilebottnawi: Looks like RegExp hell, they will ever kill us :smile: 

Don't forget some of css-in-js support Sass/Less/etc (for example https://github.com/4Catalyzer/astroturf) and using regexp potentially can break code, especially quoted strings/strings in parentheses, I'm sure there are problems

src/language-js/embed.js
```diff
@@ -16,6 +17,38 @@ const {
...
+const cssIdentityRegExp = /[^$a-zA-Z\d_-]/;
```
evilebottnawi: Lie for Sass/Less and non standard CSS-in-JS

src/language-js/embed.js
```diff
@@ -32,18 +65,95 @@ function embed(path, print, textToDoc, options) {
...
+          if (
+            (!after ||
+              endsWithLineBreak ||
+              after[0] === ";" ||
+              after[0] === "}") &&
+            (!before ||
+              before.slice(-1) === ";" ||
+              before.slice(-1) === "{" ||
+              before.slice(-1) === "}")
+          ) {
```
evilebottnawi: Move this stuff under function with good names like `isEndOfDeclaration`/`isEndOfRule`/etc, because it is hard to understand what you mean

src/language-js/embed.js
```diff
@@ -16,6 +17,38 @@ const {
   utils: { mapDoc, stripTrailingHardline }
 } = require("../doc");
 
+const cssPlaceholder = new Placeholder("prettier");
+const CSS_PROPERTY_PLACEHOLDER = new Placeholder("prettier").get(0);
+const CSS_EXTRA_SEMICOLON_MARK = new Placeholder("prettier").get(0);
```
evilebottnawi: Why do you use UPPERCASE notation?

fisker: they are const string

src/language-js/embed.js
```diff
@@ -1,5 +1,6 @@
 "use strict";
 
+const Placeholder = require("id-placeholder");
```
evilebottnawi: maybe we can avoid using extra dep?

fisker: you want those logic here? I made this module for this PR

src/language-js/embed.js
```diff
@@ -255,62 +365,112 @@ function transformCssDoc(quasisDoc, path, print) {
...
+
+        // find following semicolon
+        let semicolonIndex = -1;
+        for (let i = index + 1; i < original.length; i++) {
+          const value = original[i];
+          if (typeof value !== "string" || !value.trim()) {
+            continue;
+          }
+          if (value !== ";") {
+            throw new Error(
```
evilebottnawi: Very bad idea, we should not throw the error on broken code, just do not format it

src/language-js/embed.js
```diff
@@ -255,62 +365,112 @@ function transformCssDoc(quasisDoc, path, print) {
...
+          break;
+        }
+        original[semicolonIndex] = "";
+      }
+      part = part.replace(new RegExp(CSS_EXTRA_SEMICOLON_MARK + ";", "g"), "");
+
+      // clean css prop placeholder
+      if (part === CSS_PROPERTY_PLACEHOLDER) {
+        if (original[index + 1] !== ":") {
+          throw new Error(
```
evilebottnawi: https://github.com/prettier/prettier/pull/6794/files#r347824455

src/language-js/embed.js
fisker: this whole function is `try/catch` ed

src/language-js/embed.js
evilebottnawi: This module looks very overloaded for such simple actions, do not forget about performance.

fisker: free feel to improve it

evilebottnawi: Good/fast/stable solution https://github.com/ai/nanoid

src/language-js/embed.js
evilebottnawi: hm, can't find try/catch here

src/language-js/embed.js
evilebottnawi: Good solution above

src/language-js/embed.js
fisker: https://github.com/prettier/prettier/blob/64ab703d4122c2c835741dcf888b3df663c33c2a/src/main/ast-to-doc.js#L109

fisker: also, this error is put there for safety, actually I never find a way to trigger it, unless you modify ast manually

src/language-js/embed.js
fisker: I need indexed placeholder, I want it alphabets only, I already explained in #6790 , you prefer this module over a module designed for this feature? how can you say it's not good/fast  have you tested it.

src/language-js/embed.js
evilebottnawi: :+1:

src/language-js/embed.js
lydell: `cssIdentifierRegExp` would also be a better name

lydell: Here’s "ident-token" in the CSS spec: https://www.w3.org/TR/css-syntax-3/#ident-token-diagram

fisker: maybe I named it not good, I can explain what is this for, 

```css
a {
  before${js}after;
}
```

it will be

```css
a {
  beforePLACEHOLDERafter;
}
```

as I explained, postcss can't parse this,

I need do

```css
a {
  beforePLACEHOLDERafter;
}
```
but, I split them into ['....before', PLACEHOLDER, 'after...'], I wan't move before and after back, this RegExp is used to check what character can be moved.

I'm open to a good solution

src/language-js/embed.js
evilebottnawi: @fisker Unfortunately, I can not approve of the use of this module here due to performance problems, and its congestion and unfortunately I don’t have time to improve it, so I suggest using a popular and stable solution. 
Some problems:
- Low entropy - https://github.com/ai/nanoid/blob/master/random.js vs https://github.com/fisker/id-placeholder/blob/master/src/string.js#L15 
- Performance (unnecessary options overloading and extra checks) https://github.com/fisker/id-placeholder/blob/master/src/placeholder.js#L27 vs https://github.com/ai/nanoid/blob/master/index.js#L21
- Size of package

I can do more analysis of the problems, but I really don't have much time.

src/language-js/embed.js
fisker: test cases are on top of this file 

https://github.com/prettier/prettier/blob/3bddcd07d2ca72dfa06f77105e2162a141774ec1/tests/template_literals/css-orphan.js

evilebottnawi: Let's rename variable to avoid misleading

src/language-js/embed.js
fisker: `namespace` and `identity` is designed to identity what this placeholder is used for, check is necessary, because `prefix` `suffix` `identity` all can be passed by user, but it may not alphabets only, may cause parse problem. 

and size? kidding???

fisker: also, for templateLiteral restore we need index, what do you suggest to find index, using nanoid, cache them all?

maybe some methods are not useful for a placeholder generate, but useful to split and test

lydell: @evilebottnawi I think you misunderstand what id-placeholder is doing. It’s not just generating an ID, like ai/nanoid. id-placeholder is specifically made for Prettier’s use case as far as I understand, which is to put stuff in place of `${...}` so that our parsers manage to parse the code, and the be able to substitute the `${...}`s back.

I haven’t looked into the code of id-placeholder yet, but I think it’s worth trying it out. This PR solves a _ton_ of css-in-js issues. That’s great! If we release that and get reports of Prettier being way too slow I’m sure we’ll be able to fix performance then.

Also, it’s just [1.7 kB minified](https://bundlephobia.com/result?p=id-placeholder@1.0.2).

evilebottnawi: @lydell @fisker 
> Low entropy - https://github.com/ai/nanoid/blob/master/random.js vs https://github.com/fisker/id-placeholder/blob/master/src/string.js#L15

This is a pretty serious problem that can give same value multiple times and lead to strange problems that will be extremely difficult to identify and demonstrate

evilebottnawi: We really simplify this and avoid unnecessary overloading

fisker: > This is a pretty serious problem that can give same value multiple times 

we don't even need it random , if you like  I can set `identity` to `css`.

evilebottnawi: Why we can't use just a string?

fisker: fyi: 1.7k including extra an esm build and not minified, it could be even smaller

fisker: > Why we can't use just a string?

because there is small chance people actually wrote `pprreettttiieerrCSSaaaaaaaaaaaaaaaarreeiitttteerrpp` random string lower this chance to `1 / (26 ** 8)`

lydell: Maybe it’s better to have a static string so that Prettier is more deterministic?

fisker: that's why this module has `option`

evilebottnawi: > Maybe it’s better to have a static string so that Prettier is more deterministic?

I think the same

evilebottnawi: @fisker we can use `-prettier-randomstringwithbiglength` instead of generation, it is unnecessary

fisker: could you take one minute to check the readme , https://github.com/fisker/id-placeholder/blob/master/readme.md

evilebottnawi: README is very misleading, it lacks examples and a description of why certain methods/options are needed and what use is made of them.

Improving packages is beyond the scope of this PR; we do not need such a package for such simple actions, sorry, I will not give approval for usage this, simple string is enough in this case

evilebottnawi: > so you are saying , if I copy/paste those codes here , it will not "beyond" the scope.

No, i saying what we can use simplify here instead package, we need string + couple simple method.

It is easy maintenance, it is easy to fix, it is reduce unnecessary overloading, we don't need all these stuff what provide package.

evilebottnawi: Honestly, I don’t understand why you started to implement the package without discussing it with other developers. This is not entirely logical and may lead to such situations. I do not see any advantage in using the package, since all this can be simplify. I also want to say that additional dependencies sometimes make project management difficult, so you should avoid them whenever possible, and now this is exactly the situation. 

In the future, I would recommend discussing ideas or creating PoC. This would allow finding the best solution instead of wasting time.

fisker: ask you one question, have you ever check some of our deps like `camelcase` `html-tag-names` `parse-srcset`, how many lines in there package? why are we  using them?

fisker: > I don’t understand why you started to implement the package

because I like this idea I came up with, not only for this PR , I thought I could use it for my own stuff. I made a lot of stuff, some I dont even use, I dont need anyone to approve

fisker: another thing, I don't even use `css-in-js`, and I haven't use `Less` for about 3 years.

so maybe I should not even code here

evilebottnawi: @fisker do not be toxic and be calm, emotions do not help solve problems. I will try to convey my point of view. 

All packages that you have listed have their own logic that cannot be simplified (yes, maybe they can certainly be optimized, but they offer unique logic). The package that you offer does not make sense since all the code can be simplified to a few lines, there is no unique logic, it's just a string generator with a couple of simple methods.

> like this package https://github.com/fisker/set-input-files/blob/e3d773614ec4b98ab2e850ea80fe651b59934461/src/index.js#L5 its one line module, and I never used it, I just want to build it

If you like to do this, it doesn’t mean that it’s good practice. You can create 100 such packages every day. You yourself say that you never used it, then why generally create such things, maybe I missed something, it makes no sense to me.

> another thing, I don't even use css-in-js, and I haven't use Less for about 3 years.
> so maybe I should not even code here

You, me and other contributors also do not use part of the opportunities/languages/etc and this does not mean that we should help other developers, as well as each other. I just said that it’s worth discussing some points before starting, maybe it’s even better to make PoC and then improve it as you get feedback especially with such PRs, which rewrite a rather important part of the logic.

I do not quite understand you, you created the PR and expect feedback, I made recommendations for improvement as a developer/contributor. This is a normal development process when you get feedback and it turns out that some things can be improved, simplified, rewritten. No one ever writes perfect code.  

I even wrote you that most of the work is done good - https://github.com/prettier/prettier/pull/6794#pullrequestreview-318915994.

lydell: My first git feeling when I saw the `id-placeholder` package was – is this going to be useful to anyone besides Prettier? It don’t think so. Since the library is so small, I think it could help moving the code into Prettier itself. Then if we need to fix something in it in the future and @fisker isn’t around (maybe on vacation or whatever), we don’t need to fork the package.

So I’d be in favor of copying in the code of id-placeholder into Prettier. Since that will be private code in Prettier, we can probably simplify it a little since there’s no nice public API to worry about. And @evilebottnawi could push to this PR with improvements he might think of.

@fisker You have done some amazing work with this PR and we would be super happy if you’d like to take it across the finishing line! I think the above discussion took a bad turn because of language barriers (none of us have English as their mother language, right?) and too little time (all of us are really busy with lots of work).

evilebottnawi:
@fisker Why close?

fisker:
sorry for wasting you time

evilebottnawi:
Stop closing issue

evilebottnawi:
I will continue development

fisker:
### 

-

evilebottnawi:
@fisker man what are you doing? I want to continue development based on your code, why you remove branch

/cc @prettier/core guys it is a violation code of conduct

fisker:
@evilebottnawi 

I restored this branch. You can reopen this.

Maybe you can continue work on it. 

I saw too many multi-parser problems this PR can resolve, and we can't even use pure `postcss` parser (as you said, we should not use `postcss-less` for css), because old placeholder uses `@` .

evilebottnawi:
Let's reopen, I think we can fix it in near future

fisker:
@evilebottnawi You can't reopen this old one? If not I can send another PR

thorn0:
> we should not use postcss-less for css

A bit off topic, but what's wrong with using a Less parser for CSS?

evilebottnawi:
@thorn0 because we should throw an error on invalid syntax, Less allow to use many strange things and many things in Less is not compatibility with pure CSS, at least it would be more logical to use Sass/SCSS parser, because SCSS superset of CSS like typescript superset of javascript

lydell:
@thorn0 @evilebottnawi Sounds like #1784?

evilebottnawi:
Yep

evilebottnawi:
Let's solve it after `2.0` released, because here no breaking changes, just fixes and improvements

evilebottnawi:
@fisker let's postpone it for v2.2, what do you think? I will help you with it

fisker:
I think I made myself clear, I don't want to involve this!
You do whatever you want.

evilebottnawi:
> I think I made myself clear, I don't want to involve this!

Why? Something hard that you can't solve?

sosukesuzuki:
I'm so not related to this PR but I agree with to postpone.

PR #6796, opened 2019-11-04
fix: numeric separators were removed from BigInt literals

thorn0:


src/language-js/printer-estree.js
```diff
@@ -1610,15 +1610,8 @@ function printPathNoParens(path, options, print, args) {
...
-        printNumber(
-          n.extra
-            ? n.extra.rawValue
-            : // TypeScript
-              n.value
-        ),
-        "n"
-      ]);
+      // babel: n.extra.raw, typescript: n.raw
+      return (n.extra ? n.extra.raw : n.raw).toLowerCase();
```
thorn0: Lowercasing is the only part of `printNumber` that applies to bigints. So I removed the `printNumber` call.

PR #6801, opened 2019-11-05
ci: move `Dev_Test` to the top

fisker:
We don't care lint result that much, move `Dev_Test` to the top, so we can see the test result immediately when we click link to azure, no need click the tab on left

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,22 +8,12 @@ variables:
   ENABLE_CODE_COVERAGE: "" # only enable coverage on the fastest job
 
 jobs:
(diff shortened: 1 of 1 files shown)


fisker:
@evilebottnawi 

yes we do, here is why I want this, when I write code, I don't always run full tests on my machine, because it cost a lost of time to finish. when I commit something , I wait for the CI result, I think not only me, other developer do this.

When we check the CI log, we navigate to azure, and click `dev_test` tab then see the log.

And there is PR not changing much, we can still check the lint result.

evilebottnawi:
I am not against this change, let's wait other review

PR #6815, opened 2019-11-05
test with duplicated class name

brodybits:
@thorn0 pointed out in <https://github.com/prettier/prettier/issues/6469#issuecomment-549821981> that the behavior in case of `class A {} class A {}` had changed since 1.18.2, due to an update of `@babel/parser`.

Here is a test that should confirm the current behavior and hopefully ensure that it does break in the future. The test description and snapshot are based on the new behavior, but it should be straightforward (trivial) to change it for the old behavior if needed.

I did also confirm that rollback to @babel/parser@7.2.0 would give us the old behavior.


brodybits:
closing in favor of #6816

PR #6830, opened 2019-11-06
Upgrade flow-parser from 0.89 to 0.111

gkz:
In flow-parser version 0.90, Flow changed how we report offsets in the location information to use bytes. This was useful internally, but doesn't work well with JS consumers. This caused Prettier to be unable to upgrade past 0.89. Recently, we added an option to report offsets in a JS friendly way, and made this the default for JS consumers of flow-parser.

The snapshot changes are due to the fact that the offsets returned in 0.89 had some bugs when dealing with multibyte characters. With the rewrite of this functionality, these bugs have been fixed, and Prettier is now correctly preserving newlines in these situations.


thorn0:
Great news. Thanks for the PR. Was the fix for https://github.com/facebook/flow/issues/8159 included to this release of flow-parser? I wanted to remove a [workaround](https://github.com/prettier/prettier/blob/c26f08758f5d24e552c3418784396dd51350a216/src/language-js/needs-parens.js#L673) for it, but the tests fail as if the bug is still there.

gkz:
Hi @thorn0, that fix is not released yet, but now that we've fixed the issues with updating the flow-parser dependency, you will be able to regularly update it every two weeks without issue.

We expect that fix to be in the next version, to be released a couple of weeks from now.

fisker:
I tryed to update this, tests fails since `0.90.0`, how this version became compatible

fisker:
why not `0.111.3`?

gkz:
There weren't any parser changes in the point releases.

PR #6831, opened 2019-11-06
Include Flow & TypeScript in objects test

brodybits:
- move `tests/objects/expression.js` into `tests/objects/expressions` subdirectory (and update test snapshots)
- include Flow & TypeScript in `tests/objects/jsfmt.spec.js`

rationale: all tested modules in `tests/objects` except for `tests/objects/expression.js` can be handled by Flow & TypeScript parsers


thorn0:
The folder needs a more descriptive name than `expressions`. How about `babel-only`?

brodybits:
> The folder needs a more descriptive name than `expressions`. How about `babel-only`?

I renamed the folder to: `tests/objects/expressions-babel-only`.

Any objections?

FYI I have a few more like this coming, may be a little slow since I am kinda multitasking right now. I do not expect any more renames at this point.

My apologies if these changes are a distraction from the upcoming release:)

thorn0:
My thinking was that more Babel-only tests might be created in this folder in the future.

brodybits:
> My thinking was that more Babel-only tests might be created in this folder in the future.

Got it. I renamed the folder as requested and did a clean rebase.

Just curious what kind of likelyhood that Babel would be used with TypeScript here?

thorn0:
> what kind of likelyhood that Babel would be used with TypeScript here?

You mean in this folder or in Prettier in general? #6400 is WIP. If we add tests for `babel-ts` or `babel-flow` to the same folder and they use TS/Flow-specific syntax, then `babel` won't be able to format them, so it won't make sense to do so. Is this what you wanted to clarify?

PR #6843, opened 2019-11-06
[WIP] Include Babel in break-calls tests

brodybits:
- move tests/break-calls/type_args.js into own subdirectory
- cover Babel in tests/break-calls/type_args, which has somewhat different formatting from Flow & TypeScript in this case
- include Babel in other break-calls tests


tests/break-calls/type_args/jsfmt.spec.js
```diff
@@ -0,0 +1,2 @@
+run_spec(__dirname, ["flow", "typescript"]);
+run_spec(__dirname, ["babel"]);
```
thorn0: Please remove this line. `babel` does not and will not support this syntax. So what's the point?

brodybits: I think it would be good for it to be known and remain known how the Babel parser would handle the input, even if it is not designed to do anything special.

thorn0: Good for what? How does it help anything? Why not add also `markdown` and `html` to this test then?

brodybits: Show what should Prettier do then if a user puts this input when using the Babel parser?

thorn0: Are we going to test every possible input and parser combination?

fisker:
Close since `babel` doesn't support flow syntax anymore.

PR #6844, opened 2019-11-06
[WIP] Include Babel in many other tests

brodybits:
I did a quick `%s` operation in vim to include Babel in many other tests where this was possible, with the help from `npm jest -u` to update the snapshots.

In case @prettier/core wants me to split the changes into individual commits, I should be able to do this sometime in the next 1-2 weeks.

There are still a lot of tests with Babel not included. I hope to finish this in the next few days.


thorn0:
> In case @prettier/core wants me to split the changes into individual commits

The opposite would be better I think. I mean if you took the changes from your other "Include Babel..." PRs (except the ones that include more sophisticated changes than just adding `babel` to the parser list) to this one and closed them.

thorn0:
Please add `innerComments` to the beginning of `src/language-js/clean.js` where `leadingComments`, `trailingComments`, etc., are deleted

brodybits:
> Please add `innerComments` to the beginning of `src/language-js/clean.js` where `leadingComments`, `trailingComments`, etc., are deleted

I do not understand, cannot promise when I would get a chance to look in to this kind of update. I just wanted to update tests according to the existing behavior.

brodybits:
Closing in favor of PR #6917

Issue #6769, opened 2019-10-31
Formatting .mongo files


aymericbouzy:
I'm trying to format .mongo files in VS Code using prettier/prettier-vscode extension as explained in my PR https://github.com/prettier/prettier-vscode/pull/1009

Here is an example of a .mongo file (it is valid javascript) : 

```js
db.users.find({ someField: { $exists: true } });
```

These files are created by Azure Cosmos DB, an extension for VS Code. Renaming them to .js files doesn't work for me, because then I cannot run the query to the DB, it only works for files with the extension .mongo.

VS Code seems to treat mongo as a different language from javascript or typescript. I would love to open a PR to add support for .mongo files in Prettier : can someone give me a general direction? I'm new to the Prettier codebase.

thorn0:
I guess you need to change [this line](https://github.com/prettier/prettier/blob/56a2cd8a1b809e5f1b9f0f08b0d5bad4bbfd61d1/src/language-js/index.js#L13).

thorn0:
@aymericbouzy Do you have further questions?

aymericbouzy:
Perfect! I'm opening a PR right away. Thank you so much!

PR #6848, opened 2019-11-06
Add mongo as a VS Code supported language

aymericbouzy:
Closes https://github.com/prettier/prettier/issues/6769

This PR adds support for the VS Code mongo language (alias of javascript)

- ~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -1367,7 +1367,7 @@ const bigints = [200_000n, 0x0000_000an, 0b0111_1111n];
 
 #### VS Code: add support for .mongo files ([#6848] by [@aymericbouzy])
 
-When using the Azur Cosmos DB extension for VS Code, you can create .mongo files to write MongoDB queries, which use Javascript syntax. This change allows VS Code to format your file using Prettier.
+When using the Azure Cosmos DB extension for VS Code, you can create .mongo files to write MongoDB queries, which use Javascript syntax. This change allows VS Code to format your file using Prettier.
 
 ```js
 db.users.find({ someField: { $exists: true } });
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -692,7 +692,7 @@ Previously, Prettier only supported the most common HTML entities, such as `&nbs
 [error] stdin: SyntaxError: Unknown entity "excl" - use the "&#<decimal>;" or  "&#x<hex>;" syntax (1:6)
 [error] > 1 | <p>Hi&excl;</p>
 [error]     |      ^
-[error]   2 | 
+[error]   2 |
 -->
 
 <!-- Output (Prettier master) -->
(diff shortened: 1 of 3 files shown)


aymericbouzy:
I have not added any test for this change, can someone point me to where I should add them if they are required in this situation?

thorn0:
Just run `yarn test -u` to update the snapshots.

aymericbouzy:
@thorn0 ready for review 😄

CHANGELOG.unreleased.md
```diff
@@ -1365,6 +1365,14 @@ const bigints = [200000n, 0x0000000an, 0b01111111n];
 const bigints = [200_000n, 0x0000_000an, 0b0111_1111n];
 ```
 
+#### VS Code: add support for .mongo files ([#6848] by [@aymericbouzy])
+
+When using the Azure Cosmos DB extension for VS Code, you can create .mongo files to write MongoDB queries, which use Javascript syntax. This change allows VS Code to format your file using Prettier.
+
+```js
+db.users.find({ someField: { $exists: true } });
+```
```
evilebottnawi: Can you format changelog item as we do in others?

thorn0: I don't think this makes sense in this case.

aymericbouzy: Is that what you mean?

``````md
<!-- prettier-ignore -->
```js
// Input
db.users.find({someField:{$exists: true}});

// Output
db.users.find({ someField: { $exists: true } });
```
``````

Comment (evilebottnawi) [reply CHANGELOG.unreleased.md]: yep, but maybe we really can simplify it here as you do

Comment (aymericbouzy) [reply CHANGELOG.unreleased.md]: I'm not making any change unless you tell me to 😄 I believe too it's not really necessary in this case.

**Comment by aymericbouzy** [comment]
What's the next step for this? Anything else I should do?

**Comment by thorn0** [comment]
Did you try to patch Prettier locally to check that this change does what you need?

**Comment by aymericbouzy** [comment]
I did try this other PR I had opened previously https://github.com/prettier/prettier-vscode/pull/1009 and it did work locally. I have no tried to patch prettier locally, and to be honest, I'm not exactly sure how I should proceed ... I don't think VS Code can pickup that Prettier can be used to format documents unless my PR to prettier-vscode is merged? Please tell me if I'm wrong of course.

**Comment by aymericbouzy** [comment]
To make it clear, I only opened this PR because my PR to the prettier-vscode codebase was closed with this recommandation. Since I'm new to the codebase, I tried to do as recommended, but I still have not understood why my first PR was not enough to solve my use-case since it was working fine locally.

**Comment by thorn0** [comment]
If the VS Code extension finds that Prettier is locally installed in the project, then it uses this local version. So you can do `npm install aymericbouzy/prettier#vscode-mongo-support` and check if it has any effect.

**Comment by aymericbouzy** [comment]
wow thanks for showing me the `npm install aymericbouzy/prettier#vscode-mongo-support` command! I'm amazed it works so well! npm knows where to look for this kind of package, how to build it, which package version it should override in my package.json ... very impressive! Does it work by default for any package published to npm, or is there some configuration to make it work with prettier?

worth documenting in the CONTRIBUTING.md section I believe.

**Comment by aymericbouzy** [comment]
Well it doesn't work for me : I get 

```
There is no formatter for 'mongo'-files installed.
```

Should I reopen https://github.com/prettier/prettier-vscode/pull/1009 ?

thorn0:
I just saw it work. Did you try restarting VS Code?

As for you question, it's a feature of npm. It can install packages from different sources, including git repositories. See https://docs.npmjs.com/cli/install. It doesn't work for any npm package because the repository structure doesn't necessarily reflect what is published. E.g. `package.json` might be not in the root of the repository.

aymericbouzy:
Thanks for the explanation regarding regarding the npm feature 👍 

I did restart VS Code, and it wasn't working at first, but after a few seconds, it worked 🎉 boot time I supposed. Thank you so much!

aymericbouzy:
When do you expect there will be a new prettier version including this change?

evilebottnawi:
@aymericbouzy in near future, we are working on new version

aymericbouzy:
It's now released

PR #6850, opened 2019-11-06
Update `cosmiconfig` to v6

fisker:
fixes #6810

### DIFF
--- a/src/config/resolve-config.js
+++ b/src/config/resolve-config.js
@@ -48,13 +48,10 @@ const getExplorerMemoized = mem(opts => {
     }
   });
 
-  const load = explorer.load;
-  const search = explorer.search;
-
   return {
     // cosmiconfig v4 interface
     load: (searchPath, configPath) =>
-      configPath ? load(configPath) : search(searchPath)
+      configPath ? explorer.load(configPath) : explorer.search(searchPath)
   };
 });
 
--- a/src/config/resolve-config.js
+++ b/src/config/resolve-config.js
@@ -10,7 +10,8 @@ const resolveEditorConfig = require("./resolve-config-editorconfig");
 const loadToml = require("../utils/load-toml");
 
 const getExplorerMemoized = mem(opts => {
-  const explorer = thirdParty.cosmiconfig("prettier", {
+  const cosmiconfig = thirdParty["cosmiconfig" + (opts.sync ? "Sync" : "")];
+  const explorer = cosmiconfig("prettier", {
     cache: opts.cache,
     transform: result => {
       if (result && result.config) {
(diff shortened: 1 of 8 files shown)


src/config/resolve-config.js
```diff
@@ -47,8 +48,8 @@ const getExplorerMemoized = mem(opts => {
     }
   });
 
-  const load = opts.sync ? explorer.loadSync : explorer.load;
-  const search = opts.sync ? explorer.searchSync : explorer.search;
+  const load = explorer.load;
+  const search = explorer.search;
```
evilebottnawi: We can simplify this - `const { load, search } = explorer;`

fisker: sure

fisker: we don't even need this const, use `explorer.load` directly

fisker:
must be async function

fisker:
includes....

fisker:
should do something with the build script

fisker:
another funny story, bug is from `import-fresh` , it clean require cache from parent module, make sense right? because calling require is from parent module no need clean `import-fresh` self, but when we bundle this module, there will be no parent, haha. too tired to confirm this.

https://github.com/fisker/prettier/pull/151

last two commits, going to sleep now, 2am...

fisker:
@brodybits

evilebottnawi:
@fisker I can take care of updating if you have problems

brodybits:
I just tried `yarn prepare-release` and got `ReferenceError: regeneratorRuntime is not defined` on `tests_integration/__tests__/config-resolution.js`.

I think we should be able to resolve it by using `babel-plugin-transform-async-to-promises`, just like you did for the globby@10 update in PR #6776. I think we would need to keep this for browser support, even if we drop Node.js pre-8.0 support.

I would not personally favor pushing this update into the upcoming Prettier 1.19 release. Definitely a good job, though.

fisker:
@brodybits i know that, success build bundle also not work

fisker:
@evilebottnawi go for it

fisker:
changed base to `next` branch

evilebottnawi:
@fisker why next? Here breaking change?

fisker:
@evilebottnawi 

Can't make bundle work on node@4 

- async functions need `babel-plugin-transform-async-to-promises` or `regenerate-runtime`,
- `util.promisify` requires another shim, like `pify`
- also need transform `includes` to `indexOf`
- most import, `require-fresh` not play well on node 4, hard to find a solution

not worse to dig

evilebottnawi:
Okey, let's do it for next

PR #6852, opened 2019-11-06
fix: issue #6813 (Zero-based lists are broken)

### LINKED ISSUE #6813: [Markdown] Zero-based lists are broken
As provided by @alexandrtovmach in https://github.com/gatsbyjs/gatsby/pull/18942#issuecomment-549749393, parsing zero-based lists is broken.

**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAGAdAAgDIEsDOMAOlAIxZ6EkBMFBMIANCBAA4y7T7KgCGATvwgB3AAoCE3FLwBuEXABMmIAEb9eYANZwYAZVYbcUAObIY-AK5xmACxgBbADYB1G7nj4DYOLsnvcMu4Ansjg+NzMRvhw-DCi6sb2vMgAZryO0cwAVvgAHgBC6lo6urz2cHhQcKnpmSA5ubpGxo5wAIoWEPA1GdYgBvzR-KFJ-JoKIlDKrPxGMM6KMDbIAByozDMQ0c7qrKEzcEMy1cwAjp3w8WxSILz4ALRVcArPyvxw57jv8byJyUhpXrMaL2XBmSx9fDNVodLrVAG1PowXgqBYKJbIajMcy8XCOZoAYQg9iSoSg0BOIAs0QAKiipIDogBfJlAA)
```sh
--parser markdown
```

**Input:**
```md
0. List
1. List
2. List
```

**Output:**
```md
0. List
1. List
1. List

```

**Expected behavior:**
```md
0. List
1. List
2. List

```

evilebottnawi:

fixes #6813 + many tests


src/language-markdown/utils.js
```diff
@@ -162,6 +162,34 @@ function getOrderedListItemInfo(orderListItem, originalText) {
...
+function hasGitDiffFriendlyOrderedList(node, options) {
+  if (!node.ordered) {
+    return false;
+  }
+
+  if (node.children.length < 2) {
+    return false;
+  }
+
+  const startNumber = Number(
```
j-f1:
```suggestion
  const firstNumber = Number(
```

src/language-markdown/utils.js
```diff
@@ -162,6 +162,34 @@ function getOrderedListItemInfo(orderListItem, originalText) {
...
+    getOrderedListItemInfo(node.children[0], options.originalText).numberText
+  );
+
+  const number = Number(
```
j-f1:
```suggestion
  const secondNumber = Number(
```

src/language-markdown/utils.js
```diff
@@ -162,6 +162,34 @@ function getOrderedListItemInfo(orderListItem, originalText) {
...
+    getOrderedListItemInfo(node.children[1], options.originalText).numberText
+  );
+
+  if (startNumber === 0 && node.children.length > 2) {
+    const nextNumber = Number(
```
j-f1:
```suggestion
    const thirdNumber = Number(
```

tests/markdown_list/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2458,11 +2458,200 @@ proseWrap: "always"
...
+
+0. abc
+1. def
+2. ghi
+
+---
+
+1. abc
+1. def
+1. ghi
```
j-f1: Shouldn’t `0`-`1`-`1` be printed as `0`-`1`-`2` just like `1`-`2`-`2` is printed as `1`-`2`-`3`?

evilebottnawi: No, we (not only me, it is common practice for many other developers) use `1` for git friendly markers (for example you want to start with `45` and have git friendly markers):
```md
45. list
1. list
1. list
1. list
1. list
```

Same for `0`, you may want start with `0`:
```md
0. list
1. list
1. list
1. list
1. list
```

evilebottnawi: `1`-`2`-`2` is git friendly also but it is out of this PR because in out code we use only `1` as git friendly market, you can open an issue about this and i will fix it (it is easy)

j-f1: Why not do this then?

```markdown
0. list
0. list
0. list
0. list
0. list
```

evilebottnawi: Because @ikatyang use `1.` for git friendly markers (and it is common practice) :smile: I think it is really other issue, developers use `1.` for git friendly markers usually (for example https://docs.microsoft.com/en-us/contribute/how-to-write-use-markdown, i can find more links if it is necessary)

j-f1: That makes sense. Thanks! 👍

j-f1: That makes sense, thanks! 👍

fisker:
@evilebottnawi

how about

```js
const [firstNumber, secondNumber, thirdNumber] = node.children
  .slice(0, 3)
  .filter(Boolean)
  .map(child =>
    Number(getOrderedListItemInfo(child, options.originalText).numberText)
  );
```

evilebottnawi:
thirdNumber may not exist, also it is unnecessary get third node when list contains only 2 items

fisker:
I used `.filter(Boolean)`

evilebottnawi:
@lydell why 2?

evilebottnawi:
There are a lot of changes for the second version, we should still have an intermediate version 1.20

lydell:
The plan is to release 2.0 early January 2020. I don’t think we’ll be able to jam another release in between. It’s just two months away and there’s Christmas and all.

lydell:
Yes, sounds good.

src/language-markdown/utils.js
```diff
@@ -162,6 +162,34 @@ function getOrderedListItemInfo(orderListItem, originalText) {
   return { numberText, marker, leadingSpaces };
 }
 
+function hasGitDiffFriendlyOrderedList(node, options) {
```
fisker: why not name `hasDiffFriendlyOrderedList`

evilebottnawi: Because it is git diff, not diff, it is original name, so it is out of this PR

changelog_unreleased/markdown/pr-6852.md
```diff
@@ -0,0 +1,19 @@
+#### Markdown: handle zero-based lists correctly ([#6852](https://github.com/prettier/prettier/pull/6852) by [@evilebottnawi](https://github.com/evilebottnawi))
```
lydell:
```suggestion
#### Handle zero-based lists correctly ([#6852](https://github.com/prettier/prettier/pull/6852) by [@evilebottnawi](https://github.com/evilebottnawi))
```

PR #6853, opened 2019-11-06
audit(critical): handlebars@4.4.5 in package resolutions

brodybits:
resolves critical `yarn audit` issue due to: https://www.npmjs.com/advisories/755

without causing any regressions (failures) in the existing test suite

I hope this update can be part of the release 1.19 checklist (#6469).

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,9 +13,6 @@
   "engines": {
     "node": ">=8"
   },
-  "resolutions": {
-    "handlebars": "4.4.5"
-  },
   "dependencies": {
     "@angular/compiler": "8.2.13",
     "@babel/code-frame": "7.5.5",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
   "engines": {
     "node": ">=8"
   },
+  "resolutions": {
+    "handlebars": "4.4.5"
+  },
   "dependencies": {
     "@angular/compiler": "8.2.13",
     "@babel/code-frame": "7.5.5",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3434,20 +3434,10 @@ growly@^1.3.0:
   version "1.3.0"
   resolved "https://registry.yarnpkg.com/growly/-/growly-1.3.0.tgz#f10748cbe76af964b7c96c93c6bcc28af120c081"
 
-handlebars@^4.0.13:
-  version "4.1.1"
-  resolved "https://registry.yarnpkg.com/handlebars/-/handlebars-4.1.1.tgz#6e4e41c18ebe7719ae4d38e5aca3d32fa3dd23d3"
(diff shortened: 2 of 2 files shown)


fisker:
bad idea , npm don't support this, i use this just for testing

brodybits:
> npm don't support this

but whatever version of handlebars is installed by Yarn would be bundled when the dist is build

I tried building the dist, there seem to be no dependencies in the generated `dist/package.json`.

So I think this would be for the benefit of most users, who install from npm, would not affect installation from GitHub.

Am I missing something here?

evilebottnawi:
Why not just update the lock file?

package.json
```diff
@@ -13,6 +13,9 @@
   "engines": {
     "node": ">=8"
   },
+  "resolutions": {
+    "handlebars": "4.4.5"
+  },
```
evilebottnawi: Using `resolutions` is always bad idea, it helps in applications to broken/vuln dependency releases, but not in libraries or tools

brodybits:
> Why not just update the lock file?

Done. I just removed the resolutions field, evidently not needed now that yarn.lock is updated.

Is there anything else I can do to help get this into Release 1.19 (#6469)?

brodybits:
Thanks for the quick attention and merge!

After the fact, I really should have updated the title to reflect the end result, which is just an update to yarn.lock.

fisker:
@brodybits @lydell  

This need revert, every time I run yarn, `yarn.lock` changes.
I do this a lot, because I need switch from different branches.

fisker:
> This need

lydell:
@brodybits How did you update the lockfile?

@fisker You could use `yarn --frozen-lockfile` for now?

brodybits:
Just look through the commits in this PR. Yarn updates the dependencies in
lock file if we add resolutions, which I think is good if we resolve audit
issue. Removing resolutions did not cause Yarn to do anything else in this
case.

I can take a better look in the next few hours if you have any more
questions. I really do not see what should be a problem here.

> --
Sent from my mobile

brodybits:
Or maybe it would be less drastic to fix the lock file to use handlebars
4.1.2 rather than 4.4.5. This should  still resolve the audit issue without
affecting anything else.

I would be happy take a look some to later today, if needed.
-- 
Sent from my mobile

fisker:
@brodybits @lydell no need revert, sending a PR fix it

brodybits:
6868, right?

It looks ok to me. Unfortunately I really do not understand what problem it
solves now and how it solves the problem.
-- 
Sent from my mobile

fisker:
#6867 I don't know either, I ran yarn and open it check, it's still new version. so I send it

PR #6864, opened 2019-11-07
Don't require parens for same-operator logical expressions

jridgewell:
Multiple same-operator logical expressions do not require parentheses to disambiguate.

- [ ] ~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~


lydell:
Hi!

I _think_ this is safe and wanted.

This technically changes the AST. That’s why the CI tests are failing – they run the tests with `AST_COMPARE=1` which is more thorough but slower. To make the tests pass, you need to update src/language-js/clean.js.

src/language-js/clean.js
```diff
@@ -37,6 +37,15 @@ function clean(ast, newObj, parent) {
     return null;
   }
 
+  // We remove unneeded parens around same-operator LogicalExpressions
+  if (
+    ast.type === "LogicalExpression" &&
+    newObj.type === "LogicalExpression" &&
+    ast.operator === newObj.operator
+  ) {
+    return null;
```
jridgewell: Should I be returning null here? That feels wrong.

lydell: Yes, that’s wrong. Returning `null` means “remove this node from the AST entirely.” Which is not what Prettier is doing :) Compare the AST for `a && b && c` and `a && (b && c)`. The code needs to normalize the ASTs so that the input AST and the output AST can be deeply compared.

CHANGELOG.unreleased.md
```diff
@@ -1403,6 +1403,26 @@ async function f() {
 }
 ```
 
+#### JavaScript: Don't require parens for same-operator logical expressions ([#6864] by [@jridgewell])
```
thorn0: Could you please merge the changelog entries for this PR and for #6404: https://github.com/prettier/prettier/blob/master/CHANGELOG.unreleased.md#javascript-update--precedence-to-match-stage-3-proposal-6404-by-vjeux

Otherwise it might look like there are 2 fixes for the same thing in changelog. The existing entry is written in such a way that I was sure that PR fixed this issue too.

src/language-js/clean.js
jridgewell: Done.

src/language-js/clean.js
```diff
@@ -39,11 +39,23 @@ function clean(ast, newObj, parent) {
...
+      left: {
+        type: "LogicalExpression",
+        operator: newObj.operator,
+        left: newObj.left,
+        right: newObj.right.left,
+        loc: {}
+      },
+      right: newObj.right.right,
+      loc: {}
+    };
```
thorn0: This doesn't work for `foo && (bar && (baz && qux))`.

jridgewell: You're right. I updated to recursively rebalance the trees.

tests/handlebars-basics/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -20,12 +20,10 @@ printWidth: 80
 =====================================output=====================================
 <script id="entry-template" type="text/x-handlebars-template">
   <div class="entry">
-    <h1>
-      {{title}}
-    </h1>
+    <h1>{{title}}</h1>
```
lydell: This shouldn’t have changed, should it?

fisker: he must ran `yarn` , changed lock https://github.com/prettier/prettier/pull/6570#issuecomment-550266720

lydell: I thought we fixed that problem?

fisker: no, not fixable for now

fisker: sorry, I mean he resintalled, not `yarn`,  maybe `npm install`?

jridgewell: Reverted. Sorry, I didn't review the changed files before committing.

PR #6865, opened 2019-11-07
Show invalid config filename in error message

fisker:
Show invalid config filename in error message, makes it easier to find problem

### DIFF
--- a/tests_integration/path-serializer.js
+++ b/tests_integration/path-serializer.js
@@ -1,12 +1,12 @@
 "use strict";
 
-const replaceCWD = path => {
+const replaceCWD = text => {
   const cwd = process.cwd();
-  while (path.indexOf(cwd) !== -1) {
-    path = path.replace(cwd, "<cwd>");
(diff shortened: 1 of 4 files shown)


tests_integration/path-serializer.js
```diff
@@ -1,9 +1,18 @@
 "use strict";
 
+const replaceCWD = text => {
+  const cwd = process.cwd();
+  while (text.indexOf(cwd) !== -1) {
+    text = text.replace(cwd, "<cwd>");
+  }
+
+  return text;
+};
```
evilebottnawi: Bad idea, better output absolute path, because you can run prettier using Node API and change cwd so developer do not know cwd

fisker: Original implemention is replace to `<cwd>`, I only change this, because there might be two path in message.

this is for test, absolute path can't match snap

evilebottnawi: My mistake, sorry

PR #6866, opened 2019-11-07
Add `core-js` support for build script

fisker:
Since we've already [drop Node 6 support for direct install from github](https://github.com/prettier/prettier/pull/6657), we could start using some new Syntax like `async`, `Array.includes` and [more](https://node.green/), at least for `core` bundles. 

Also, some new version of dependencies requires `async` and `Array#include`, but our build script can't handle, like `globby@10`(https://github.com/prettier/prettier/pull/6776), `cosmiconfig@6`(https://github.com/prettier/prettier/pull/6850).

I add this as an option, and not enable for any bundle yet, also only allow to use for core bundles.


lydell:
As far as I understand, we won’t need this in [2.0](https://github.com/prettier/prettier/issues/6888) so I’m closing. Let me know if I’ve misunderstood anything.

lydell:
@fisker Feel free to continue discussion here. We’ll reopen if needed.

fisker:
@lydell You are right, we can reopen it when we  need it

Issue #5586, opened 2018-12-03
angular vs angularjs: operator priority problem with ?: (ternary) and | (pipe)


thorn0:
**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIaYwA2cAvADojDC4AkYMATkQHSRFECGADgM5zpcAflwByAKIAPbpyjpRuJGIDCEDj34KAPrkZxeEAK6MwcXAF9zlAHyoA9FmzXyUDDlwBtAsTgBdCiD0TKzsXHwCwmJSMnIKSqKq6uHauvpGJnA29o7WIAA0IBDcBNC8yKCcjIwQAO4ACpUIZSic2BCY6PkgAEaMnGAA1nAwAMoyYJhQAObITIZwBQAWMAC2RADqi4T643AjTYQ4hACeyOC8ZQWT-IwwdX1TK5zIAGacRPwFAFa8kgBCfUGwxGnBWcAAMpM4K93p8QD9JCNJlMSABFQwQeAwj4LEAyRg3M6yKaGLiMLrcRiTGDrDowRbIAAcAAYCpSIPx1n1uGdKfo4IxsNCCnoAI6GTB6e6cR7PJBvHEFfgrTCzRjzJXItEYrHy2G4mCcbq09D05AAJgKTE4mCIyNUKyeZyg0GFIEM-AAKkbmgr+JYgA)
```sh
--parser angular
```

**Input:**
```html
<div title="{{ $ctrl.collapsed ? 'Expand' : 'Collapse' | resource }}"></div>
<div [title]="$ctrl.collapsed ? 'Expand' : 'Collapse' | resource"></div>
```

**Output:**
```html
<div title="{{ $ctrl.collapsed ? 'Expand' : ('Collapse' | resource) }}"></div>
<div [title]="$ctrl.collapsed ? 'Expand' : ('Collapse' | resource)"></div>

```

**Expected behavior:**

Same as input.

**UPD**

Turns out it is a grammar difference between AngularJS and Angular.

In AngularJS, `a ? b : c | x` means `(a ? b : c) | x`.

In Angular, `a ? b : c | x` means `a ? b : (c | x)`.

It would make sense if Prettier kept such expressions as is to stay compatible with both. Now, it corrupts AngularJS templates.

ikatyang:
[...]

Issue #5466, opened 2018-11-13
AngularJS: one-time binding syntax


thorn0:
**Prettier 1.15.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeADgPgDpQAR7DB5JIBuAtgHQCuAznAE54C+LuqA9FiADQgR0MAJbQ6yUAENGjCAHcACtITiUkshGEATPiABGjSWADWcGAGV0R4VADmyGIxpx+ACxgUANgHVXw+HRWYHDmKv7CZP4AnsjgdOL8NgyMMAqGthSSyABmkp4M-ABWdAAeAEKGJmbmkhRwADI2cDl5BSDFJeY2tp5wAIo0EPAt+S4gVozJsZJ2NJ7SuuiMNjDe2jCuyAAcAAz8SxAM3obosUtwyWTN-IxwAI40wrdpkhlZSLmj-AwUwg5OYzo3V6AyGzQ+rTGMEkejWWg2yAATPxHJJhJ5ugBhCAUTKxKDQa4gehwAAqMNUnwYbCAA)
```sh
--parser angular
```

**Input:**
```html
<p>
  {{ ::vm.user }}
</p>
```

**Output:**
```html
SyntaxError: Unexpected token :
    at o (https://prettier.io/lib/parser-angular.js:1:43175)
    at t.parseNgInterpolation (https://prettier.io/lib/parser-angular.js:1:44299)
    at n (https://prettier.io/lib/parser-angular.js:1:54796)
    at Object.t.parseInterpolation (https://prettier.io/lib/parser-angular.js:1:55134)
    at https://prettier.io/lib/parser-angular.js:1:55771
    at Object.parse (https://prettier.io/lib/parser-angular.js:1:55427)
    at Object.parse$2 [as parse] (https://prettier.io/lib/standalone.js:12931:19)
    at textToDoc (https://prettier.io/lib/standalone.js:15724:23)
    at https://prettier.io/lib/standalone.js:15712:14
    at Object.embed$2 [as embed] (https://prettier.io/lib/standalone.js:20444:55)
```

**Expected behavior:**

No error.

Notice that AngularJS and Angular aren't the same framework. The one with JS is an older incompatible version. However, their expression syntaxes seem to be mostly compatible except for this one-time binding thing. Docs: https://docs.angularjs.org/guide/expression#one-time-binding

ikatyang:
[...]

PR #6869, opened 2019-11-07
Rudimentary support for the most used directives of AngularJS 1.x

thorn0:
While there are two syntax incompatibilities ([one-time bindings](https://github.com/prettier/prettier/issues/5466) and [the precedence of `|`](https://github.com/prettier/prettier/issues/5586)) between the expression languages of the old AngularJS and the new Angular, overall the two languages are compatible enough for legacy and hybrid AngularJS-based projects to be able to benefit from using Prettier, which I know from my own experience. An ideal solution would be to add a separate parser for AngularJS, but given the [LTS status](https://docs.angularjs.org/misc/version-support-status) of the framework, such an effort would hardly make sense. On the other hand, unofficially impoving the partial support like I'm doing with this PR is very cheap in terms of effort and code complexity, but would significantly improve the experience for those of our fellow programmers who still have to deal with AngularJS.


thorn0:
Thanks. I maintain one project where this would be really helpful. :)

PR #6872, opened 2019-11-08
Fix unpkg links in docs

fisker:

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -3,7 +3,7 @@ id: browser
 title: Browser
 ---
 
-Run Prettier in the browser with the `standalone.js` UMD bundle shipped in the NPM package (starting in version 1.13). The new UMD bundle only formats the code and has no support for config files, ignore files, CLI usage, or automatic loading of plugins.
+Run Prettier in the browser with the `standalone.js` UMD bundle shipped in the NPM package (starting in version 1.13). The UMD bundle only formats the code and has no support for config files, ignore files, CLI usage, or automatic loading of plugins.
 
 ### `prettier.format(code, options)`
 
@@ -15,12 +15,14 @@ See [Usage](#usage) below for examples.
 
 ### Global
 
-<!-- prettier-ignore -->
 ```html
-<script src="https://unpkg.com/prettier@1.13.0/standalone.js"></script>
-<script src="https://unpkg.com/prettier@1.13.0/parser-graphql.js"></script>
-<script type="text/javascript">
-prettier.format("query { }", { parser: "graphql", plugins: prettierPlugins });
+<script src="https://unpkg.com/prettier@1.18.2/standalone.js"></script>
+<script src="https://unpkg.com/prettier@1.18.2/parser-graphql.js"></script>
+<script>
+  prettier.format("query { }", {
+    parser: "graphql",
+    plugins: prettierPlugins
+  });
 </script>
 ```
 
@@ -40,8 +42,8 @@ prettier.format("query { }", {
 
 ```js
 define([
-  "https://unpkg.com/prettier@1.13.0/standalone.js",
-  "https://unpkg.com/prettier@1.13.0/parser-graphql.js"
+  "https://unpkg.com/prettier@1.18.2/standalone.js",
+  "https://unpkg.com/prettier@1.18.2/parser-graphql.js"
 ], (prettier, ...plugins) => {
   prettier.format("query { }", { parser: "graphql", plugins });
 });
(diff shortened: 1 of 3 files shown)


docs/browser.md
```diff
@@ -60,7 +62,7 @@ This syntax doesn't necessarily work in the browser, but it can be used when bun
 ### Worker
 
 ```js
-importScripts("https://unpkg.com/prettier@1.13.0/standalone.js");
-importScripts("https://unpkg.com/prettier@1.13.0/parser-graphql.js");
+importScripts("https://unpkg.com/prettier@1.18.2/standalone.js");
+importScripts("https://unpkg.com/prettier@1.18.2/parser-graphql.js");
```
evilebottnawi: Maybe `unpkg` has alias like `stable`/`latest` to avoid using versions here, it is not good update them on each new release

thorn0: Simply omit the version number, if you want latest.

evilebottnawi: Maybe we can create issue in `unpkg` and support alias `latest` as latest stable version?

thorn0: Like a wrote, they already have such an URL, but as it is dynamic, it can't be cached and is implemented as a redirect.

fisker: yes, `https://unpkg.com/prettier/standalone.js` is the latest. but we should suggest people to pin version

evilebottnawi: @thorn0 i was misleading, i mean support latest without redirect :smile:

fisker: jsdelivr has this feature? forgot

thorn0: That would miss the point of CDN, which is to provide URLs that never change and can be cached forever.

thorn0: @fisker Doesn't Docusaurus have a feature for substitutions like that?

thorn0: > If you want users to be able to use the latest version when you cut a new release, the best policy is to put the version number in the URL directly in your installation instructions. This will also load more quickly because we won't have to resolve the latest version and redirect them.

-- https://unpkg.com/#cache-behavior

fisker: I was going to put a placeholder there on first thought, then I think may confuse people read this `.md`  directly

thorn0: How about processing all the links to `unpkg.com/prettier`, without relying on `previousVersionOnMaster`?

fisker: > Anyway, using `previousVersionOnMaster` looks fragile.

do you prefer 

```js
/https:\/\/unpkg\.com\/prettier@(.*?)\//g
```
?

fisker: I was using this at first, then I thought `previousVersionOnMaster` should be safer

thorn0: I'd prefer this, yes.

fisker: I can change it, one sec

fisker: remove the protocol, we may won't strip it some day.

... or, should I strip it now?

scripts/release/steps/update-version.js
```diff
@@ -17,15 +17,12 @@ async function bump({ version, previousVersionOnMaster }) {
...
-    }
-
-    return content;
-  });
+  processFile("docs/browser.md", content =>
+    content.replace(
+      /\/\/unpkg\.com\/prettier@(.*?)\//g,
+      `//unpkg.com/prettier@${version}/`
+    )
+  );
```
evilebottnawi: Good idea

docs/browser.md
```diff
@@ -15,12 +15,14 @@ See [Usage](#usage) below for examples.
 
 ### Global
 
-<!-- prettier-ignore -->
```
fisker: I can understand why here is a ignore now, must be https://github.com/prettier/prettier/pull/6423

docs/browser.md
thorn0: Let's remove it while you are at it.

fisker: @evilebottnawi  `https://unpkg.com/prettier@1.13.0/standalone.js` or `//unpkg.com/prettier@1.13.0/standalone.js` ?

fisker:
One more thing, I just found

```md
This syntax doesn't necessarily work in the browser, but it can be used when bundling the code with browserify, Rollup, webpack, or another bundler.
```

Should I remove `browserify`?

thorn0:
> Should I remove browserify?

It's out of scope for this PR.

docs/browser.md
lydell: `https`. There’s never a reason to use anything else, is it?

fisker: ok , ready to merge

PR #6873, opened 2019-11-08
Docs: Update `resolveConfigFile` usage example

fisker:
update codes from #6568 , should use chained promise


fisker:
this docs still seem not right...

fisker:
After reading the source code, I don't think `prettier.resolveConfig` accept `configFile`

fisker:
I can't find a real world example for this api usage, because no function accepts `configFile`, only cli accept it.

so, use like this?

```js
prettier.resolveConfigFile(filePath).then(configFile => {
  execa('prettier', ['**/*', '--config', configFile])
});
```

why this exists?

fisker:
blame https://github.com/prettier/prettier/blame/289b98c1250be8b7f668f91458f5c61a43889d5f/index.js#L52

fisker:
first introduced in https://github.com/prettier/prettier/pull/4364

PR #6880, opened 2019-11-08
Run CI on latest OS

fisker:
Run CI on latest OS, and move prod_test to linux, its' faster than macos


azure-pipelines.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   - job: Dev_Test_Windows
     displayName: Dev Test on Windows Node v12
     pool:
-      vmImage: vs2017-win2016
+      vmImage: "windows-latest"
```
thorn0: Wasn't it you who wrote recently that pinned versions should be preferred?

evilebottnawi:
@thorn0 i think better use latest os, developers other use latest versions

thorn0:
The test environment should be stable. If it changes constantly, how can we be sure that CI fails are caused by our code, not by something in the OS?

evilebottnawi:
@thorn0 os problems is very rare and we can't fix them on our side, also testing on latest os can help us to catch problems in modules on the latest operating systems

thorn0:
Then we should have tests for both latest and non-latest OS versions to understand whether problems are of that kind.

evilebottnawi:
@thorn0 sounds good, i think we can add latest and non latest jobs to improve testing

fisker:
Setting up github actions https://github.com/prettier/prettier/pull/7431 , so we don't need this anymore.

PR #6881, opened 2019-11-08
Fix lost adjacent JSX when using Babel

thorn0:
Otherwise Prettier formats `<a/><b/>` to `<a/ >;`.

### DIFF
--- a/src/language-js/parser-babylon.js
+++ b/src/language-js/parser-babylon.js
@@ -101,7 +101,7 @@ function tryCombinations(fn, combinations) {
   let error;
   for (let i = 0; i < combinations.length; i++) {
     try {
-      return rethrowSomeRecoveredErrors(fn(combinations[i]));
+      return fn(combinations[i]);
     } catch (_error) {
       if (!error) {
         error = _error;
@@ -109,20 +109,6 @@ function tryCombinations(fn, combinations) {
     }
   }
   throw error;
-}
-
-function rethrowSomeRecoveredErrors(ast) {
-  for (const error of ast.errors) {
-    if (
-      typeof error.message === "string" &&
-      error.message.startsWith(
-        "Adjacent JSX elements must be wrapped in an enclosing tag."
-      )
-    ) {
-      throw error;
-    }
-  }
-  return ast;
 }
 
 function parseJson(text, parsers, opts) {
--- a/src/language-js/parser-babylon.js
+++ b/src/language-js/parser-babylon.js
@@ -101,7 +101,7 @@ function tryCombinations(fn, combinations) {
   let error;
   for (let i = 0; i < combinations.length; i++) {
     try {
-      return fn(combinations[i]);
+      return rethrowSomeRecoveredErrors(fn(combinations[i]));
     } catch (_error) {
       if (!error) {
         error = _error;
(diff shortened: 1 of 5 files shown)


src/language-js/parser-babylon.js
```diff
@@ -111,6 +111,20 @@ function tryCombinations(fn, combinations) {
   throw error;
 }
 
+function rethrowSomeRecoveredErrors(ast) {
+  for (const error of ast.errors) {
+    if (
+      typeof error.message === "string" &&
+      error.message.startsWith(
+        "Adjacent JSX elements must be wrapped in an enclosing tag."
```
thorn0: @nicolo-ribaudo Would be more convenient if errors had some kind of short codes in order for such checks to not rely on the wording of the message.

lydell:
Ouch! Good find!

This makes me a little bit worried. What other things could Prettier potentially format in a very unexpected way now? Why does Babel even recover from this error?

nicolo-ribaudo:
I'd categorize this as a bug in Babel

lydell:
@nicolo-ribaudo Thanks!

@thorn0 Should we do this the other way around – _whitelist_ what we know is safe, and rethrow everything else? We wouldn’t want Prettier to delete parts of people’s code.

nicolo-ribaudo:
I'm preparing a PR for Babel right now, to not recover from this error. Apart from this everything should be safe (I manually checked a lot of tests, but they were so many that I missed this one).

lydell:
@nicolo-ribaudo Cool, thank you so much! I’ll hold off making the Prettier 1.19 release for a bit then.

thorn0:
Why hold off the release? Let's use my workaround for now.

lydell:
@thorn0 Yeah, if you feel confident about your workaround we could merge this and clean up later after the release.

nicolo-ribaudo:
This workaround looks safe to me.

When do you plan to release 1.19? I can try to get a new Babel version published at most on monday if you prefer.

lydell:
The plan is to release it in the next couple of hours. So let’s merge this PR as-is. Then there’s no rush fixing it on your side.

nicolo-ribaudo:
@lydell Hold off the release, I'm releasing Babel!

lydell:
Alright! :+1:

nicolo-ribaudo:
https://github.com/babel/babel/releases/tag/v7.7.3

lydell:
Yay, thanks for the super quick fix! 🏆 

@thorn0 Would you like to update this PR? Otherwise I can do it.

thorn0:
Yep, I'll do it.

yarn.lock
```diff
@@ -378,7 +378,12 @@
...
     js-tokens "^4.0.0"
 
-"@babel/parser@7.7.2", "@babel/parser@^7.1.2", "@babel/parser@^7.1.5", "@babel/parser@^7.4.5", "@babel/parser@^7.5.5", "@babel/parser@^7.7.0", "@babel/parser@^7.7.2":
+"@babel/parser@7.7.3":
+  version "7.7.3"
+  resolved "https://registry.yarnpkg.com/@babel/parser/-/parser-7.7.3.tgz#5fad457c2529de476a248f75b0f090b3060af043"
+  integrity sha512-bqv+iCo9i+uLVbI0ILzKkvMorqxouI+GbV13ivcARXn9NNEabi2IEz912IgNpT/60BNXac5dgcfjb94NjsF33A==
+
+"@babel/parser@^7.1.2", "@babel/parser@^7.1.5", "@babel/parser@^7.4.5", "@babel/parser@^7.5.5", "@babel/parser@^7.7.0", "@babel/parser@^7.7.2":
   version "7.7.2"
```
nicolo-ribaudo: Is it possible to deduplicate this?

thorn0: done

PR #6882, opened 2019-11-08
deduplicate entries in yarn.lock

thorn0:
I used this tool: https://github.com/atlassian/yarn-deduplicate

- [ ] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [ ] (If the change is user-facing) I’ve added my changes to the `CHANGELOG.unreleased.md` file following the template.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #6884, opened 2019-11-08
deduplicate entries in yarn.lock - part 2

thorn0:
Unfortunately, yarn.lock merged in #6882 is unstable. Yarn rewrites it on `yarn install`. This PR fixes this issue.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -227,6 +227,15 @@
     "@babel/traverse" "^7.7.0"
     "@babel/types" "^7.7.0"
 
+"@babel/highlight@7.0.0-beta.46":
+  version "7.0.0-beta.46"
+  resolved "https://registry.yarnpkg.com/@babel/highlight/-/highlight-7.0.0-beta.46.tgz#c553c51e65f572bdedd6eff66fc0bb563016645e"
+  integrity sha512-r4snW6Q8ICL3Y8hGzYJRvyG/+sc+kvkewXNedG9tQjoHmUFMwMSv/o45GWQUQswevGnWghiGkpRPivFfOuMsOA==
+  dependencies:
+    chalk "^2.0.0"
+    esutils "^2.0.2"
+    js-tokens "^3.0.0"
+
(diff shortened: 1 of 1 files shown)


lydell:
Nice, I was just going to ask if it happened to you too.

kachkaev:
I've added a small alias to my `.bash_profile` to avoid bumping into this problem and it works well so far:

```
alias yyy='yarn; yarn-deduplicate; yarn'
```

We also run this small script in CI to ensure yarn.lock is always deduplicated:

```bash
if ! npx yarn-deduplicate --list --fail; then
  echo
  echo "Please run 'npx yarn-deduplicate; yarn install' locally to make sure there are no duplicates in yarn.lock"
  echo
  exit 1
else 
  echo
  echo "No duplicates found in yarn.lock!"
  echo
fi
```

Ultimately, `yarn-deduplicate` can be added to the pre-commit hook for 100% automation.

thorn0:
@kachkaev Thanks for sharing. Unfortunately, it's not an option for Prettier for now. In #6882, I had to revert some changes in yarn.lock manually for the tests to work on Node 4. Do you happen to know if there is a way to check that the lock file is stable _without_ calling `yarn-deduplicate`?

kachkaev:
@thorn0 I see.

>  Do you happen to know if there is a way to check that the lock file is stable _without_ calling `yarn-deduplicate`? 

We simply do this in CI:

```bash
yarn install

if ! git diff --exit-code yarn.lock; then
  echo "Changes were detected in yarn.lock file after running 'yarn install', which is not expected. Please run 'yarn install' locally and commit the changes.";
  exit 1;
fi
```

Perhaps not the most elegant solutions, but it works. In theory, you should be able to just call `yarn --frozen-lockfile`, but this command has bugs, e.g. it does not work for monorepos (https://github.com/yarnpkg/yarn/issues/5840). I can't imagine anything more robust than a good old `git diff` 🙂

PR #6893, opened 2019-11-09
Enable `corejs` option for core bundle `bin-prettier.js` and `third-party.js`

fisker:
this PR requires #6866 , fixes #6891 , I tried to make a test for it, but failed. local test

before

```bash
$ cat index.js | node node_modules/prettier/bin-prettier --stdin --parser babel
[error] regeneratorRuntime is not defined
```

after

```bash
$ cat index.js | node dist/bin-prettier --stdin --parser babel
"use strict";

module.exports = require("./src/index");
```

diff

```diff
npm notice
-npm notice package: prettier@1.19.0
+npm notice package: prettier@1.20.0-dev
npm notice === Tarball Contents ===
npm notice 468B    package.json
-npm notice 1.4MB   bin-prettier.js
+npm notice 1.6MB   bin-prettier.js
npm notice 70.6kB  doc.js
npm notice 1.3MB   index.js
npm notice 1.1kB   LICENSE
npm notice 57.4kB  parser-angular.js
npm notice 266.2kB parser-babylon.js
npm notice 1.5MB   parser-flow.js
npm notice 139.6kB parser-glimmer.js
npm notice 51.4kB  parser-graphql.js
npm notice 110.7kB parser-html.js
npm notice 241.1kB parser-markdown.js
npm notice 351.8kB parser-postcss.js
npm notice 2.6MB   parser-typescript.js
npm notice 167.7kB parser-yaml.js
npm notice 3.7kB   README.md
npm notice 1.1MB   standalone.js
-npm notice 146.7kB third-party.js
+npm notice 296.0kB third-party.js
npm notice === Tarball Details ===
npm notice name:          prettier
-npm notice version:       1.19.0
+npm notice version:       1.20.0-dev
-npm notice filename:      prettier-1.19.0.tgz
+npm notice filename:      prettier-1.20.0-dev.tgz
-npm notice package size:  2.2 MB
+npm notice package size:  2.3 MB
-npm notice unpacked size: 9.6 MB
+npm notice unpacked size: 9.9 MB
-npm notice shasum:        0a355225fa968b1a716739508f2689d18a0a513a
+npm notice shasum:        5dc617bfb5759264b838f9d34d238578b220a850
-npm notice integrity:     sha512-p32VAB78i7JQT[...]txg82TWSxTi1A==
+npm notice integrity:     sha512-DLkijM5TPo6uU[...]PoR2Dnm5WbNGw==
npm notice total files:   18
npm notice
-prettier-1.19.0.tgz
[...]

Issue #6888, opened 2019-11-09
Prettier 2.0 – 2.0 2020


lydell:
We already have a Prettier 2.0 issue – #3503 – but tell you what: Time’s up for that one. As of this writing, Node.js 8 is scheduled to expire early on December 31, 2019. And we still support Node.js 4. In two months we will be 6 versions behind!

Supporting Node.js 4 and Node.js 6 has been a real hassle lately as our dependencies (rightfully!) has stopped supporting them. We have spent a lot of unnecessary time getting tests to pass in Node.js 4.

So here’s my suggestion for a 2.0 of a Prettier 2.0 issue:

**The main focus should be dropping support for unsupported Node.js versions.** Let’s aim for releasing Prettier 2.0 early in January 2020 with support for only Node.js 10+.

Apart from that we could do some easy breaking changes:

- Change the default for `trailingComma` to `es5`, as discussed in the original Prettier 2.0 thread.
- Remove deprecated stuff.

When that is done – if we have time, we could do more things. We’ll see. For example, it would be really nice addressing #5588 (making all unsafe template literal formatting **opt-in**). I also have a suggestion for lots of improvements for Prettier’s CLI and API, but I’ll open a separate issue for that.

/cc @prettier/core @brodybits @fisker

azz:
Sounds good to me! Let's get this shipped.

fisker:
@lydell you fogot Change the default for `singleQuote` to `true`

lydell:
@fisker No, I didn’t. I don’t think we should do that. As far as I can tell, there’s been no consensus on that change. Let’s not discuss `singleQuote` for this release so that we can stay focused on the important parts.

brodybits:
I hope we can drop use of a bundled `typescript` dependency in favor of babel-ts as proposed in PR #6400. I think this would be much less bulky, also reduced variety of bundled dependency licenses (I think TypeScript is Apache 2.0.).

lydell:
That would be nice. Maybe we can change to `babel-ts` as the default TypeScript parser in 2.0 and remove the `typescript` parser in 3.0 if it all works out.

thorn0:
On the other hand, we might want to go the other way round and use TypeScript to parse JS. If Prettier is going to format JSDoc, TypeScript already parses it for us.

![image](https://user-images.githubusercontent.com/94334/68574765-4e527b80-0473-11ea-8137-6cae8912e5c9.png)

manuelbieh:
[...]

PR #6894, opened 2019-11-09
Quick-fix for stdin being broken in 1.19.0

### LINKED ISSUE #6891: 1.19 Regression with --stdin
**Environments:**
- Prettier Version: 1.19.0
- Running Prettier via: CLI
- Runtime: Node v10.14
- Operating System: MacOs

**Steps to reproduce:**
I did a clean install in a new directory of prettier:
```sh
$ yarn install prettier
$ cat test.js | prettier --stdin --stdin-filepath test.js
[error] regeneratorRuntime is not defined
```

Seems like the built "binaries" of `node_modules/prettier/bin-prettier.js` `node_modules/prettier/index.js` and `node_modules/prettier/third-party.js` all have this undefined name `regeneratorRuntime` presumably from a babel generator transformation, probably only used during stream/stdin reading. I double checked that 1.18.2 does not have this issue.

lydell:
Closes #6891.

This downgrades the get-stream package from 5.x to 4.x, which does not
use `async` functions which works around the problem of
`regeneratorRuntime` not being defined (it shouldn’t be needed).

This is an alternative solution to #6893 because I don’t know yet if we
want to enable `corejs` given that we will most likely drop support for
Node.js 8 and older in the next version.

Just like #6893 this PR comes without tests, but I verified locally that
`--stdin` works after this change.

/cc @fisker @evilebottnawi

### DIFF
--- a/changelog_unreleased/cli/pr-6894.md
+++ b/changelog_unreleased/cli/pr-6894.md
@@ -1,0 +1,12 @@
+#### Fix `--stdin` regression in 1.19.0 ([#6894](https://github.com/prettier/prettier/pull/6894) by [@lydell](https://github.com/lydell))
+
+<!-- prettier-ignore -->
+```jsx
+// Prettier stable
+$ echo "test" | prettier --stdin --parser babel
+[error] regeneratorRuntime is not defined
+
+// Prettier master
+$ echo "test" | prettier --stdin --parser babel
+test;
+```
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "find-parent-dir": "0.3.0",
     "find-project-root": "1.1.1",
     "flow-parser": "0.111.3",
-    "get-stream": "5.1.0",
+    "get-stream": "4.1.0",
     "globby": "6.1.0",
     "graphql": "14.5.8",
     "html-element-attributes": "2.2.0",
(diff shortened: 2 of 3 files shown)


fisker:
Do you want me send a pr transform async to promies? I can do that

lydell:
@fisker So my thinking is – once we start working on [2.0](https://github.com/prettier/prettier/issues/6888) we want to remove all of these things from the build script, right? So wouldn’t it be easier to just downgrade get-stream? Then we don’t have to clean up as much later.

fisker:
@lydell 

I'm thinking  differently, instead of removing things, we should let babel(mostly corejs) to take care of any syntax problem.

Yes, we drop old version of node support, but node 10 will be old someday, we don't want start  adding thing again.

lydell:
Let’s take that decision when Node.js 10 is old (in a year or so).

fisker:
Are we going to remove babel, when we start 2.0?

lydell:
I don’t know yet.

fisker:
I really dont think we can , for this issue I'm okay with this solution. But  https://github.com/prettier/prettier/pull/6866 is only a option, I add it to prepare for some deps like this, we can't always go back, some deps might  requires lot of changes, we can use it 
just in case, please keep #6866 open

lydell:
Sorry, I don’t understand what you mean.

fisker:
please reopen #6866, we can discuss there

PR #6908, opened 2019-11-10
Update build scripts to target Node.js 10

fisker:
follows #6907


fisker:
here, another problem, netlify not running on 10, I didn't saw a config file, is it on there site?

fisker:
netlify not running

j-f1:
https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript

.nvmrc
```diff
@@ -0,0 +1 @@
+10
```
lydell: Shouldn’t it be up to each developer to choose which version they want to develop with?

j-f1: I think it would be good to standardize (maybe using something like [asdf](https://asdf-vm.com/#/)) so that everyone has the exact same versions of Node/Python/whatever other languages we use.

lydell: I removed it, at least for now, to unblock the PR. We can re-add later if it makes sense.

fisker:
@lydell  why did you remove nvmrc?

fisker:
@lydell removing this will break netlify deploy

lydell:
Can’t we fix netlify deploy some other way?

fisker:
I'm not sure, is there any settings in netlify site?

fisker:
I checked my netlify account, on netlify site  settings page

`Environment`  section , maybe we can set `NODE_VERSION` to `10` or `12`

but in `Deploy contexts` section, it seems only one branch is allowed

fisker:
Someone taking care of this? 

/cc @prettier/core

j-f1:
@azz is the owner of the Prettier site on Netlify.

---

If you have the time, could you email Netlify support and ask them to create an OSS team for us so all the core maintainers can access the site settings @azz?

PR #6912, opened 2019-11-10
Disable trailingComma for Angular internal parser

fisker:
extra comma, cause parse error, fond in #6910

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -121,7 +121,12 @@ let uid = 0;
 
 function shouldPrintComma(options, level) {
   // angular does not allow trailing comma
-  if (options.parser === "__ng_interpolation") {
+  if (
+    options.parser === "__ng_interpolation" ||
+    options.parser === "__ng_action" ||
+    options.parser === "__ng_binding" ||
+    options.parser === "__ng_directive"
+  ) {
     return false;
   }
 
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -120,6 +120,11 @@ const {
 let uid = 0;
 
 function shouldPrintComma(options, level) {
+  // angular does not allow trailing comma
+  if (options.parser === "__ng_interpolation") {
+    return false;
+  }
+
   level = level || "es5";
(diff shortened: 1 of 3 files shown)


src/language-js/printer-estree.js
```diff
@@ -120,6 +120,11 @@ const {
 let uid = 0;
 
 function shouldPrintComma(options, level) {
+  // angular does not allow trailing comma
+  if (options.parser === "__ng_interpolation") {
```
fisker: I'm not sure about `__ng_action`, `__ng_binding` , `__ng_directive`, not using angular

thorn0: Trailing comma should be disabled for all of them.

fisker: you are right, done

fisker:
we need this for #6910

src/language-js/printer-estree.js
```diff
@@ -120,6 +120,16 @@ const {
 let uid = 0;
 
 function shouldPrintComma(options, level) {
+  // angular does not allow trailing comma
+  if (
+    options.parser === "__ng_interpolation" ||
+    options.parser === "__ng_action" ||
+    options.parser === "__ng_binding" ||
+    options.parser === "__ng_directive"
+  ) {
```
evilebottnawi: Can we simplify this using regexp `^__ng_`?

fisker: really nessary? only 4, but I was thinking override here, https://github.com/prettier/prettier/blob/4ed525350904a3c19f4a7504e53a27a98b12531e/src/language-js/parser-angular.js#L5 what do you suggest?

fisker: Just realize this options is not passing to print function

```js
options.trailingComma = "none";
```

seems not safe. maybe I should keep check in shouldPrintComma

fisker: @evilebottnawi still prefer regexp?

evilebottnawi: It is just question, i am not familiar with angular very good,all fine

PR #6917, opened 2019-11-11
[RFC ...] more complete js parser testing

brodybits:
- update tests_config/run_spec.js to support js-all & babel-all, in a similar fashion to using wildcards
- update many (many) tests to run on all supported JavaScript parsers babel, babel-flow, flow, and typescript
- update some other tests to run on babel, babel-flow, and flow parsers

Supersedes PR #6844

I think these changes could help us test the proposed "babel-ts" parser in PR #6400 more completely.

This seems to be a pretty massive set of changes. I would be happy to split it into multiple PRs if needed.

NOT INCLUDED: There seem to be many tests that can run on all parsers but with different snapshot results depending on which parser is used. I think this part should be done in a separate PR.


fisker:
Close since no active in years, and we alreay auto test `babel-flow` when testing `flow`.

PR #6935, opened 2019-11-12
Style: use async functions

fisker:


src/cli/util.js
```diff
@@ -360,33 +360,31 @@ function applyConfigPrecedence(context, options) {
...
-      writeOutput(context, format(context, input, options), options);
-    })
-    .catch(error => {
-      handleError(context, relativeFilepath || "stdin", error);
-    });
+    writeOutput(context, format(context, input, options), options);
+  } catch (error) {
+    handleError(context, relativeFilepath || "stdin", error);
+  }
 }
```
fisker: strange, am I doing this wrong?

fisker: this has to be sync to pass tests

fisker:
waiting for #6908

PR #6700, opened 2019-10-24
[ImgBot] Optimize images

MichaelDeBoey:
## Beep boop. Your images are optimized!

Your image file size has been reduced!

<details>
<summary>
Details
</summary>

| File | Before | After | Percent reduction |
|:--|:--|:--|:--|
| /website/static/images/get_started_rects.svg | 2.73kb | 2.49kb | 8.70% |
| /website/static/images/editors/editor_atom.svg | 0.88kb | 0.83kb | 4.91% |
| /website/static/images/languages/markdown_small_grey.svg | 0.38kb | 0.37kb | 4.57% |
| /website/static/images/users/used_by_zeit.svg | 0.13kb | 0.12kb | 3.79% |
| /website/static/images/users/used_by_jest.svg | 1.42kb | 1.37kb | 3.64% |
| /website/static/images/users/used_by_react.svg | 1.23kb | 1.19kb | 3.49% |
| /website/static/images/editors/editor_vscode.svg | 0.67kb | 0.64kb | 3.38% |
| /website/static/images/languages/tools_gql.svg | 1.40kb | 1.35kb | 3.35% |
| /website/static/images/languages/graphql_small_grey.svg | 1.41kb | 1.36kb | 3.33% |
| /website/static/images/npm.svg | 0.25kb | 0.24kb | 3.16% |
| /website/static/images/users/heroku.svg | 1.88kb | 1.83kb | 3.16% |
| /website/static/images/npm_grey.svg | 0.27kb | 0.27kb | 2.86% |
| /website/static/images/users/metric-ai-white.svg | 1.72kb | 1.68kb | 2.27% |
| /website/static/images/users/estalee.svg | 1.83kb | 1.80kb | 2.08% |
| /website/static/images/languages/tools_html.svg | 0.25kb | 0.25kb | 1.94% |
| /website/static/images/editors/editor_webstorm.svg | 1.00kb | 0.98kb | 1.77% |
| /website/static/images/languages/python_small_grey.svg | 1.02kb | 1.00kb | 1.73% |
| /website/static/images/editors/editor_vs.svg | 0.31kb | 0.31kb | 1.57% |
| /website/static/images/users/buildcom.svg | 2.10kb | 2.06kb | 1.54% |
| /website/static/images/languages/tools_wip.svg | 1.15kb | 1.13kb | 1.53% |
| /website/static/images/editors/editor_emacs.svg | 1.25kb | 1.23kb | 1.41% |
[...]

PR #6939, opened 2019-11-12
Optimize images with SVGO and added script

lipis:


package.json
```diff
@@ -130,6 +131,8 @@
     "build": "node --max-old-space-size=3072 ./scripts/build/build.js",
     "build-docs": "node ./scripts/build-docs.js",
     "check-deps": "node ./scripts/check-deps.js",
-    "spellcheck": "npx -p cspell@4.0.31 cspell {bin,scripts,src}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md"
+    "spellcheck": "npx -p cspell@4.0.31 cspell {bin,scripts,src}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md",
+    "svgo": "svgo --pretty --indent=2 --precision=1",
+    "svgo:all": "find website | grep '\\.svg$' | xargs -Iz -n 1 yarn svgo z"
```
evilebottnawi: Can't work for non linux users, so contributing from windows will be not easy :smile:

kachkaev: Could this command be added to CI? This will helps us ensure

1. it works on Windows
2. There are no produced git diffs following its execution (i.e. svgs are optimised by the committer)

lipis: `svgo:all` will run once.. and who ever else wants to add new images could run `yarn svgo path/to/image.svg`

lipis: I removed it..

lydell:
svgo can optimize all svgs in a folder using `svgo --folder=icons` I think. Could that help?

lipis:
But it doesn't do it recursively :(

JounQin:
You can also try [@pkgr/imagemin](https://github.com/rx-ts/pkgr/tree/master/packages/imagemin) which supports almost all images and works perfectly with `lint-staged`.

Usage: `i file1 file2` or `i **/*.{jpg,svg,png}`

Or with lintstaged:

```json
{
  "*.{jpg,svg,png}": ["i", "git add"]
}
```

Then no custom npm script is required.

How do you think?

@lipis @evilebottnawi @lydell

package.json
```diff
@@ -130,6 +131,7 @@
     "build": "node --max-old-space-size=3072 ./scripts/build/build.js",
     "build-docs": "node ./scripts/build-docs.js",
     "check-deps": "node ./scripts/check-deps.js",
-    "spellcheck": "npx -p cspell@4.0.31 cspell {bin,scripts,src}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md"
+    "spellcheck": "npx -p cspell@4.0.31 cspell {bin,scripts,src}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md",
+    "svgo": "svgo --pretty --indent=2 --precision=1"
```
JounQin: Why should we output pretty content?

lipis: Because we are Prettier :P

JounQin: I doesn’t make sense to me. Is the author of these svg files wrote it manually? I supposed not. They are image files like `jpg/png`, we should not care about if they are `prettier` IMO.

And if we do care about it, we should support prettier `svg` by ourselves then.

lipis:
Shall we proceed with this?

JounQin:
> Shall we proceed with this?

I think my question has not been resolved.
https://github.com/prettier/prettier/pull/6939#discussion_r347111220

package.json
lipis: We have Prettier for XML now :)

but SVGO does more than what Prettier would do.. it also cleans it up.

JounQin: You still didn't answer why should we prettier these files?

> I doesn’t make sense to me. Is the author of these svg files wrote it manually? I supposed not. They are image files like jpg/png, we should not care about if they are prettier IMO.

I think we should minify them instead.

JounQin:
@lipis Should we care about other types of images also? I recommend to use `@pkgr/imagemin` which works for most images and perfectly with `lint-staged` so that we don't need this npm script at all.

lipis:
@JounQin Feel free to open a new PR with it.. I'm closing this one.

lipis:
As for the lint staged, I think we didn't wanted hooks back in the day.. not sure if that changed.

lipis:
Maybe it's better to use the ImgBot after all.. No dependencies :) 
 https://github.com/prettier/prettier/pull/6700

JounQin:
I read #4019 but still can't understand why should we do like https://github.com/prettier/prettier/pull/4019#issuecomment-367499786

Personally I love `lint-staged` for all files, not only for images.
Should I open another issue for discussion?

PR #6941, opened 2019-11-12
Support string-to-package config in JSON schema:

lydell:
Prettier supports having just a string in the config file to use a
shared config from a package. Example:

    "@company/prettier-config"

But the JSON schema generated by running `node
scripts/generate-schema.js` did not include that possibility.

See https://github.com/SchemaStore/schemastore/pull/855 for more
information.


evilebottnawi:
Can we add test?

thorn0:
There already exists a test for this in `tests_integration/cli/config/external-config`.

PR #6942, opened 2019-11-12
Sort the package.json alphabetically

lipis:
https://sortjson.com/

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,17 +1,7 @@
 {
-  "name": "prettier",
-  "version": "1.20.0-dev",
-  "description": "Prettier is an opinionated code formatter",
+  "author": "James Long",
   "bin": {
     "prettier": "./bin/prettier.js"
-  },
-  "repository": "prettier/prettier",
-  "homepage": "https://prettier.io",
-  "author": "James Long",
-  "license": "MIT",
-  "main": "./index.js",
-  "engines": {
-    "node": ">=8"
   },
   "dependencies": {
     "@angular/compiler": "8.2.13",
@@ -73,6 +63,7 @@
     "vnopts": "1.0.2",
     "yaml-unist-parser": "1.1.1"
   },
+  "description": "Prettier is an opinionated code formatter",
   "devDependencies": {
     "@babel/core": "7.7.2",
     "@babel/preset-env": "7.7.1",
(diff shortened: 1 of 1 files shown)


merlinnot:
I think it's worth having much more readable `package.json` than be able to git blame this particular file. It makes it much easier for contributors to discover all settings and scripts.

lipis:
Couldn't agree more!

lydell:
Then I’d argue the current package.json is more readable. It feels like 90% of the time I open a package.json file I want to look at `scripts` so I appreciate it being near the top.

lipis:
But I like it in the end because of the `cat package.json`.. it's all preferences, that's why sorted is the right way.. easier to navigate! all the dependencies are sorted, why not the rest?

lydell:
Let’s not make this steal time from working on 2.0.

lipisdev:
How is this going to steal time?

lipis:
@lydell can you reopen that issue and at least wait a few more opinions before taking a decision?

qw-in:
Having `"version"` near the bottom would just feel wrong

lipis:
> Having `"version"` near the bottom would just feel wrong

Subjective...

kachkaev:
Our teams use https://github.com/matzkoh/prettier-plugin-packagejson. It is a dependency of `@company/prettier-config` and thanks to plugin autodiscovery, all our package.json files look consistent and clean :)

merlinnot:
> Our teams use https://github.com/matzkoh/prettier-plugin-packagejson. It is a dependency of `@company/prettier-config`, and thanks to Prettier’s plugin autodiscovery, all our package.json files look consistent and clean :)

I looked into this in the past, it uses https://github.com/keithamus/sort-package-json, which unfortunately doesn't sort alphabetically.

lydell:
This is 100% bike shedding. Why should we spend time on it?

bennyn:
Being an open source developer I work with many different projects and from my experience I can tell that it is really annoying when properties are not sorted in a `package.json` file. 

The format of a `package.json` file is quite loose. There are only a few mandatory properties and a lot of optional properties and non-standard properties. When viewing the `package.json` file of a new project I am working with, I want to be able to easily navigate within that file. When properties are ordered it's very simple to find entries like "[scripts](https://docs.npmjs.com/files/package.json#scripts)" and "[version](https://docs.npmjs.com/files/package.json#version)" because by the nature of the alphabet these should be found at the bottom of the file, so I would naturally scroll towards the end of a `package.json`. When properties are unordered, it makes it hard to know if something like "scripts" is at the beginning or at the end. 

Another example is the insertion of new properties. Where would you place "[bin](https://docs.npmjs.com/files/package.json#bin)" or "[bugs](https://docs.npmjs.com/files/package.json#bugs)"? Would you put the "bin" field at the beginning of the file because it's important to know the executable and "bugs" at the end of the file because it's rarely viewed? How to determine which properties are more relevant than others? With an alphabetical order, such a question does not even arise.

The larger a config files becomes, the more helpful is an alphabetical sorting. There are some good examples, such as the `tslint.json` file from Microsoft's TypeScript: https://github.com/microsoft/TypeScript/pull/16252/files

If you check the above PR, you will see a lot of entries that have a common prefix (`no-in-operator`, `no-inferrable-types`, `no-internal-module`, ...). Having them all alphabetically ordered makes it so much easier finding a Linter property when you need to change it. I don't see anything wrong in applying such good manners to a `package.json` file.

I'd also like to point out that the Prettier project has always been committed to deliver a consistent style. Having an alphabetical order in config files (such as `package.json`, `tslint.json`, `eslint.json`, etc.) is consistent. Sorting properties by personal relevance is not.

luke-john:
Under scripts, I'll often group related scripts (sometimes with a fake comment script to provide more info). 

Prettier doing alphabetical sorting here would be frustrating, and require scrips to be renamed in order to keep grouping (and I'd no longer be able to do comment scripts (ie `"// used on ci": "",`)

As a much smaller concern, this is changing runtime behaviour as well, any scripts importing this from node would have different key ordering.

ffflorian:
@luke-john you know that this is only about sorting a single file and not your personal package.json file, right? 😉

Andarist:
IMHO this could make sense - but a custom order (priority-based) should be applied if any (alphabetical doesn't make much sense to me). I doubt everyone could agree on the priority though - given that there are many "custom" fields (including browser, react-native aliasing fields, tool configs etc)

evilebottnawi:
Why :-1: :smile: It is our internal change, we **do not change order your package.json**

merlinnot:
Funny to see how everyone (including myself) spends so much time and effort on this issue, where we have so many bigger fishes to fry! 2.0 anyone? 😇

I personally sort every single json file lexicographically using this plugin: https://marketplace.visualstudio.com/items?itemName=richie5um2.vscode-sort-json

It makes me much more productive. But if there is such a strong opposition, is it really worth it?

Let's maybe just make a simple poll and whatever the result is, I guess the minority can get used to it? Do you think it's a good idea? If so, vote with:
- ❤️ for a current setup
- 🎉 for [package.json plugin](https://github.com/matzkoh/prettier-plugin-packagejson)
- 🚀 to sort lexicographically

lipis:
I think people. just got confused that prettier will start sorting package.json file in general 😂

merlinnot:
@lydell What do you think about my proposal above?

evilebottnawi:
@merlinnot out of scope this PR, please open new issue for this changes, but i think better use https://github.com/tclindner/npm-package-json-lint for that

lipis:
What are we talking about? I don't think that we want to add more linters here. Just sort the package.json once and keep it sorted. That's my proposal.

merlinnot:
@evilebottnawi Yes, I perfectly understand that. The small poll I proposed is just to reformat this one single file, `package.json` in this repository. Sorry if I phrased it in confusing way.

lydell:
@merlinnot I unsubscribed from this PR days ago.

JounQin:
How about prettier it by a prettier plugin [prettier-plugin-pkg](https://github.com/rx-ts/prettier/tree/master/packages/pkg#prettier-plugin-pkg--)?

PR #6944, opened 2019-11-12
Add the XML plugin to the docs

kddeisz:


lipis:
It looks like it's not there: https://prettier.io/docs/en/plugins.html#official-plugins

lydell:
@lipis Fixed in commit 992e410aa742e7ee0156fcaa8ea6fe30e3f37b3f

PR #6949, opened 2019-11-13
Fix MDX html parsing errors

### LINKED ISSUE #6943: MDX with "html" tag in JSX fails
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAkMOCMIDEIIAaQgIQEMd8BffAMxwnVPwHIA6APRlUADxgBXHHF4AdKPIDE+ABJwANuuqkA7rnUATAITz5AHgoR8AIwDmAXl7SDvAHzyiZg6gBu+AM4wAJ7qcPbAJITe-pjq9EFIRLzWWmAA1ryshMzuUDz4ZgxMRK4AFqikZoJFufkFgt4+uVWWuSBUIBCYMKjQ-sigjJw6AAqMCP0o9D4QqAbtINY49OlwMADKmCuoULbIMDgScB2lMOjqAOrl8DErcOsTqD0+T0HI4P79HTv+cHgjy1s6HoyDY9HUvw6ACt-KJaMtVht6Og4AAZHZwUHgyEgGGidY7WyhACKEgg8CxEOOIC2OF+OHe6AMogWmBwOxgFzmMFKyAAHAAGDpsiC-C7LTDvNlwek+TEdACOZPgAK6kxA9H8AFooHA4AZ9QtpErUNIAfQgSCkGCqR1fuhUPtDtT-ISScrMdbsdSYPRrFyDDzkAAmDoHeiodSEgDCXGB7yg0HlIAkvwAKn7JjacT4jgBJKAG2DrMDs7oAQUL62CoUpv2YzCAA)
```sh
--parser mdx
```

**Input:**
```mdx
import     {     Foo,  Bar } from     './Fixture'

# Hello,    world!

<Foo bg='red'>
   <div style={{   display:   'block'}   }>
      <Bar    >hi    </Bar>
       </div>
</Foo>

```

**Output:**
```mdx
SyntaxError: Unexpected closing tag "div". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (4:8)
  2 | 
  3 | # Hello,    world!
> 4 | 
    | ^
  5 | <Foo bg='red'>
  6 |    <div style={{   display:   'block'}   }>
  7 |       <Bar    >hi    </Bar>
```

**Expected behavior:**

In prettier 1.18.2 this worked fine, but now it seems that the div tag is parsed as html (judging from the stack trace) and fails. The example above also renders correctly in the MDX playground at <https://mdxjs.com/playground>. Result JSX is:

```jsx
    <h1>{`Hello,    world!`}</h1>
    <Foo bg='red' mdxType="Foo">
   <div style={{
        display: 'block'
      }}>
      <Bar mdxType="Bar">hi    </Bar>
       </div>
    </Foo>
```

Tigge:
For MDX wrap JSX parsing in a fragment before sending it of to babel for parsing and formatting. In MDX adjacent JSX tags are permitted in the root.
    
An option sent to the babel parser omits printing the root fragment and instead just prints its children.
    
Fixes #6943

**✨[Try the playground for this PR](https://deploy-preview-6949--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeAKgQwOYAIBumANgK5wC8wwumSuAOiAEaO4C+bAfPAM4yoB6LNk64AjACYAzPSioAYhAi4m2SowBOcACaMus3LlTaAlvlx8AnkQpVcpngAcimS3UZMiEMAGtWHTgNDIwAhTA1OAAsTIwEwiKDDYEi4Ii82RNxgAQAqGigIGBSNXEgAWzKEGFwcgQyoQ0FTfEC5AUUIUQAWAFYANhAAGhAIRxgTaB5kUHCNCAB3AAVwhCmUTHwIE20h5g1MXzgYAGVHA5MobGQYDTJhyJgyogB1aN4zsDhj1ZNx-F-LMhwDwpsMLjw4BoYIt9tgyphkAAzYgQ4YAKx4AA8QvtDidMJUADIXOBIlFwdFY44XbA2ACKJEKpKQyKIqJAZw0EI0QLK2kxu0cGguMGe2yKyAAHAAGYZCiAQ577RxAoVwbn4UnDACOjPgMNGaxAmB4AFooHAdDpdlpdSYtDCcPCyWyKSAIWUTNdbm6eDT6Xrmaz2TBMEwxdoJUgJMMbpgTEQaQBhCAVBEoAoW3YkCFYJhrYNu-BkACSUG0VWOYGFYwAguXjjBrEHyRwgA)✨**

### DIFF
--- a/src/language-markdown/embed.js
+++ b/src/language-markdown/embed.js
@@ -59,7 +59,7 @@ function embed(path, print, textToDoc, options) {
     case "jsx":
       return textToDoc(`<$>${node.value}</$>`, {
         parser: "__js_expression",
-        __maybe_mdx_adjacent: true
+        rootMarker: "mdx"
       });
   }
 
--- a/src/language-markdown/embed.js
+++ b/src/language-markdown/embed.js
@@ -57,7 +57,10 @@ function embed(path, print, textToDoc, options) {
     case "importExport":
       return textToDoc(node.value, { parser: "babel" });
     case "jsx":
-      return textToDoc(node.value, { parser: "__js_expression" });
+      return textToDoc(`<>${node.value}</>`, {
+        parser: "__js_expression",
+        __maybe_mdx_adjacent: true
+      });
   }
 
   return null;
--- a/src/language-markdown/embed.js
+++ b/src/language-markdown/embed.js
@@ -57,7 +57,7 @@ function embed(path, print, textToDoc, options) {
     case "importExport":
       return textToDoc(node.value, { parser: "babel" });
     case "jsx":
-      return textToDoc(`<>${node.value}</>`, {
+      return textToDoc(`<$>${node.value}</$>`, {
         parser: "__js_expression",
         __maybe_mdx_adjacent: true
       });
--- a/tests/mdx/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/mdx/__snapshots__/jsfmt.spec.js.snap
@@ -276,7 +276,8 @@ printWidth: 80
 
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
+123
 
 ---
 
@@ -285,16 +286,19 @@ printWidth: 80
 </Hello>
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
+123
 
 ---
 
 <Hello>
   test <World /> test
-</Hello> 123
+</Hello>{" "}
+123
 <Hello>
   test <World /> test
-</Hello> 234
+</Hello>{" "}
+234
 
 ---
 
@@ -352,7 +356,8 @@ semi: false
 
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
+123
 
 ---
 
@@ -361,16 +366,19 @@ semi: false
 </Hello>
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
+123
 
 ---
 
 <Hello>
   test <World /> test
-</Hello> 123
+</Hello>{" "}
+123
 <Hello>
   test <World /> test
-</Hello> 234
+</Hello>{" "}
+234
 
 ---
 
@@ -434,6 +442,117 @@ const a = 1;
 const a = 1
 ;<div>foo</div>
 \`\`\`
+
+================================================================================
+`;
+
+exports[`levels.mdx 1`] = `
+====================================options=====================================
+parsers: ["mdx"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+import     {     Foo,  Bar } from     './Fixture'
+
+# Hello,    world!
+
+<Foo bg='red'>
+   <div style={{   display:   'block'}   }>
+      <Bar    >hi    </Bar>
+       {  hello       }
+       {     /* another comment */}
+       </div>
+</Foo>
+
+asdfsdf <strong style={{fontWeight: 'bolder'}}>asdfasdf</strong>
+
+<Foo/>
+test
+
+<ul>
+      <li   >item    {' '} </li>
+      <li/>
+        </ul >
+
+
+=====================================output=====================================
+import { Foo, Bar } from "./Fixture";
+
+# Hello, world!
+
+<Foo bg="red">
+  <div style={{ display: "block" }}>
+    <Bar>hi </Bar>
+    {hello}
+    {/* another comment */}
+  </div>
+</Foo>
+
+asdfsdf <strong style={{fontWeight: 'bolder'}}>asdfasdf</strong>
+
+<Foo />
+test
+
+<ul>
+  <li>item </li>
+  <li />
+</ul>
+
+================================================================================
+`;
+
+exports[`levels.mdx 2`] = `
+====================================options=====================================
+parsers: ["mdx"]
+printWidth: 80
+semi: false
+                                                                                | printWidth
+=====================================input======================================
+import     {     Foo,  Bar } from     './Fixture'
+
+# Hello,    world!
+
+<Foo bg='red'>
+   <div style={{   display:   'block'}   }>
+      <Bar    >hi    </Bar>
+       {  hello       }
+       {     /* another comment */}
+       </div>
+</Foo>
+
+asdfsdf <strong style={{fontWeight: 'bolder'}}>asdfasdf</strong>
+
+<Foo/>
+test
+
+<ul>
+      <li   >item    {' '} </li>
+      <li/>
+        </ul >
+
+
+=====================================output=====================================
+import { Foo, Bar } from "./Fixture"
+
+# Hello, world!
+
+<Foo bg="red">
+  <div style={{ display: "block" }}>
+    <Bar>hi </Bar>
+    {hello}
+    {/* another comment */}
+  </div>
+</Foo>
+
+asdfsdf <strong style={{fontWeight: 'bolder'}}>asdfasdf</strong>
+
+<Foo />
+test
+
+<ul>
+  <li>item </li>
+  <li />
+</ul>
 
 ================================================================================
 `;
--- a/tests/mdx/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/mdx/__snapshots__/jsfmt.spec.js.snap
@@ -286,19 +286,16 @@ printWidth: 80
 </Hello>
 <Hello>
   test <World /> test
-</Hello>
-123
+</Hello>123
 
 ---
 
 <Hello>
   test <World /> test
-</Hello>{" "}
-123
+</Hello> 123
 <Hello>
   test <World /> test
-</Hello>{" "}
-234
+</Hello> 234
 
 ---
 
@@ -366,19 +363,16 @@ semi: false
 </Hello>
 <Hello>
   test <World /> test
-</Hello>
-123
+</Hello>123
 
 ---
 
 <Hello>
   test <World /> test
-</Hello>{" "}
-123
+</Hello> 123
 <Hello>
   test <World /> test
-</Hello>{" "}
-234
+</Hello> 234
 
 ---
 
--- a/changelog_unreleased/mdx/pr-6949.md
+++ b/changelog_unreleased/mdx/pr-6949.md
@@ -1,0 +1,43 @@
+#### Fix MDX html parsing errors ([#6949](https://github.com/prettier/prettier/pull/6949) by [@Tigge](https://github.com/Tigge))
+
+MDX parsing for JSX failed when encountering JSX elements that where not
+parsable as HTML. Such as `<Tag value={{a: 'b'}}>test</Tag>`
+
+<!-- prettier-ignore -->
+````md
+<!-- Input -->
+
+<Tag  value={ {a :  'b'   } }>test</ Tag>
+
+<Foo bg={ 'red' } >
+   <div  style={{   display:   'block'}   }>
+      <Bar    >hi    </Bar>
+       {  hello       }
+         {     /* another comment */}
+            </div>
+</Foo>
+
+<!-- Prettier stable -->
+
+```
+SyntaxError: Unexpected closing tag "Tag". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (1:35)
+> 1 | <Tag  value={ {a :  'b'   } }>test</ Tag>
+    |                                   ^
+  2 | 
+  3 | <Foo bg={ 'red' } >
+  4 |    <div  style={{   display:   'block'}   }>
+```
+
+<!-- Prettier master -->
+
+<Tag value={{ a: "b" }}>test</Tag>
+
+<Foo bg={"red"}>
+  <div style={{ display: "block" }}>
+    <Bar>hi </Bar>
+    {hello}
+    {/* another comment */}
+  </div>
+</Foo>
+
+````
--- a/changelog_unreleased/mdx/pr-6949.md
+++ b/changelog_unreleased/mdx/pr-6949.md
@@ -1,4 +1,4 @@
-#### Fix MDX html parsing errors ([#6949](https://github.com/prettier/prettier/pull/6949) by [@Tigge](https://github.com/Tigge))
+#### Fix MDX html parsing errors ([#6949](https://github.com/prettier/prettier/pull/6949) by [@Tigge](https://github.com/Tigge) & [@thorn0](https://github.com/thorn0))
 
 MDX parsing for JSX failed when encountering JSX elements that where not
 parsable as HTML. Such as `<Tag value={{a: 'b'}}>test</Tag>`
--- a/src/language-js/preprocess.js
+++ b/src/language-js/preprocess.js
@@ -10,7 +10,8 @@ function preprocess(ast, options) {
       return Object.assign({}, ast, {
         type: options.parser.startsWith("__") ? "JsExpressionRoot" : "JsonRoot",
         node: ast,
-        comments: []
+        comments: [],
+        rootMarker: options.rootMarker
       });
     default:
       return ast;
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5388,8 +5388,14 @@ function printJSXElement(path, options, print) {
     containsMultipleAttributes ||
     containsMultipleExpressions;
 
+  const isMdxBlock =
+    options.__maybe_mdx_adjacent &&
+    path.getParentNode().type === "JsExpressionRoot";
+
   const rawJsxWhitespace = options.singleQuote ? "{' '}" : '{" "}';
-  const jsxWhitespace = ifBreak(concat([rawJsxWhitespace, softline]), " ");
+  const jsxWhitespace = isMdxBlock
+    ? concat([" "])
+    : ifBreak(concat([rawJsxWhitespace, softline]), " ");
 
   const isFacebookTranslationTag =
     n.openingElement &&
@@ -5509,6 +5515,10 @@ function printJSXElement(path, options, print) {
     ? fill(multilineChildren)
     : group(concat(multilineChildren), { shouldBreak: true });
 
+  if (isMdxBlock) {
+    return content;
+  }
+
   const multiLineElem = group(
     concat([
       openingLines,
@@ -5517,13 +5527,6 @@ function printJSXElement(path, options, print) {
       closingLines
     ])
   );
-
-  if (
-    options.__maybe_mdx_adjacent &&
-    path.getParentNode().type === "JsExpressionRoot"
-  ) {
-    return concat(children);
-  }
 
   if (forcedBreak) {
     return multiLineElem;
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5518,6 +5518,13 @@ function printJSXElement(path, options, print) {
     ])
   );
 
+  if (
+    options.__maybe_mdx_adjacent &&
+    path.getParentNode().type === "JsExpressionRoot"
+  ) {
+    return concat(children);
+  }
+
   if (forcedBreak) {
     return multiLineElem;
   }
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5388,9 +5388,7 @@ function printJSXElement(path, options, print) {
     containsMultipleAttributes ||
     containsMultipleExpressions;
 
-  const isMdxBlock =
-    options.__maybe_mdx_adjacent &&
-    path.getParentNode().type === "JsExpressionRoot";
+  const isMdxBlock = path.getParentNode().rootMarker === "mdx";
 
   const rawJsxWhitespace = options.singleQuote ? "{' '}" : '{" "}';
   const jsxWhitespace = isMdxBlock
--- a/src/language-markdown/parser-markdown.js
+++ b/src/language-markdown/parser-markdown.js
@@ -7,7 +7,6 @@ const parseFrontMatter = require("../utils/front-matter");
 const { mapAst, INLINE_NODE_WRAPPER_TYPES } = require("./utils");
 const mdx = require("./mdx");
 const remarkMath = require("remark-math");
-const htmlParser = require("../language-html/parser-html").parsers.html;
 
 /**
  * based on [MDAST](https://github.com/syntax-tree/mdast) with following modifications:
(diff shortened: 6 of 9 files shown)


tests/mdx/jsx.mdx
```diff
@@ -18,15 +18,6 @@
 
 ---
 
-<Hello>
-    test   <World />   test
-</Hello>       123
-<Hello>
-    test   <World />   test
-</Hello>       234
```
thorn0: Please don't remove it. Better add a `FIXME` comment if you don't find a way to fix this.

thorn0:
Thanks. This issue with text doesn't look critical. It doesn't break code. That's the usual way Prettier formats JSX. This can be changed in a separate PR if needed.

tests/mdx/jsx.mdx
Tigge: Restored. Forgot to revert this change before submitting. Thanks!

tests/mdx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -361,16 +366,19 @@ semi: false
...
 
 <Hello>
   test <World /> test
-</Hello> 123
+</Hello>{" "}
+123
 <Hello>
   test <World /> test
-</Hello> 234
+</Hello>{" "}
```
JounQin: Not good

thorn0: Why?

JounQin: Not `prettier` to me.

`mdx` is both `html` and `jsx`.

JounQin: For example, the following is valid mdx:

```html
Inline <Test><!-- it's a comment --></Test>
```

thorn0: This specific text is JSX, not HTML. And that's how Prettier formats JSX.

JounQin: > This specific text is JSX, not HTML. And that's how Prettier formats JSX.

I have to say that's not true, see my comment above.

thorn0: Please clarify what exactly is not true. I don't understand.

thorn0: > For example, the following is valid mdx:
> And the following is not jsx:

Both your examples aren't JSX and seem to have nothing to do with what we discuss. 🤔

JounQin: I understood the text should be parsed as `jsx` now, but still think `html` style is `prettier` in this case and they are literally same, we don't need to consider it as plain `jsx`, just like our default `json` output is actually `js_expression`, right?.

thorn0: Do you have a real-life use case for text like this?

thorn0: > we don't need to consider it as plain jsx

We should consider it JSX. At least because it can contain JSX expression.

Anyway, I think it's a totally separate thing and the current PR should not take care of it. Can you open an issue?

Tigge: Yeah, this looks a bit weird, but it is how prettier formats the same code in a JSX file. It seems to make sense to format the code the same - but perhaps there are reasons to have different formatting in MDX?

thorn0:
@Tigge Please add a changelog entry.

tests/mdx/__snapshots__/jsfmt.spec.js.snap
JounQin: Well, although it's a bug fix PR, but it will change current behavior (whatever whether it was included as intended), so I think we should finish it up here.

JounQin:
```jsx

<div/> {' '}123




<div/>

{' '}123
```

Current behavior is much better for refactoring.

thorn0: @JounQin I don't understand your last example. Please show some real-life actual code that this PR would make look significantly worse.

JounQin: @thorn0 Would the following examples make sense to you?

```jsx
// original code, current behavior
<span>1</span> 2

---

// this PR's output
<span>1</span>{' '}
2

---

// If we would to refactor it by adding two lines
<span>1</span>

{' '}
2

---

// We'll only add two lines simply on current behavior
<span>1</span>

2

```

JounQin: > Yeah, this looks a bit weird, but it is how prettier formats the same code in a JSX file. It seems to make sense to format the code the same - but perhaps there are reasons to have different formatting in MDX?

As I said, `mdx` is both `html` and `jsx`, and current behavior is much easier for refactoring as above.

thorn0: Now I see your point. But it doesn't look convincing to me. I still it's more important that people be aware that this text is JSX, which Prettier would make more clear. When those two lines are added, the text becomes markdown. To avoid mistakes, the content author must understand this clearly anyway.

JounQin: And also, `{' '}` in following examples are different:

```jsx
{' '}<span>my text</span>{' '}

prefix <span>my text</span>{' '}

<span>my text</span>{' '}
```

image


In general, inline `{' '}` is a bad idea even if it is actually `jsx`.

thorn0: You can say the same about anything in curly braces (`{ foo }`). Or in other separators, e.g. `**foo**`.

thorn0: @ikatyang What are your thoughts on this?

JounQin: I'm not hiding the fact that it's `jsx`, but `mdx` extends `markdown` and `markdown` extends `html`.

The fact is that they are just been implemented as `jsx` but they should be considered as `html` in these confusing cases. That's why the inline comments are `html` style rather than `jsx` style.

thorn0: A paragraph (thing separated by double line breaks) is either markdown (with embedded HTML) or JSX. That's it. HTML inline comments don't work inside JSX paragraphs because they make JSX invalid.

> just been implemented as jsx but they should be considered as html

That's a very wrong conclusion, I would say.

Tigge:
> @Tigge Please add a changelog entry.

I've added a changelog entry

tests/mdx/__snapshots__/jsfmt.spec.js.snap
JounQin: @thorn0

```html
Inline <Test><!-- it's a comment --></Test>
```

I mean the above example, it's `jsx` + `html` style comment.

thorn0: No, it's not JSX. Come on.

JounQin: <Test> is a valid jsx element, why it's not jsx?

thorn0: Still it's not JSX. It's something else.

JounQin: So whatever `mdx` is a likely kind of *special* `jsx` like `jsx` is likely a kind of *special* `html`, we should not simply consider all of them as plain `jsx`.

I'm strongly recommending that we should not change current behavior.

thorn0: Example:
```
Inline <h3 title={ 'foo bar' }>{ 'baz' }<!-- it's a comment -->_qux_</h3> **quux**
```
It understands expressions in attributes (`{ 'foo bar' }`), but not in the inner text (`{ 'baz' }`). Also it understands markdown and HTML comments in the inner text. So this is indeed some kind of extended HTML.

But this works only in markdown paragraphs. Pure JSX paragraphs contain pure normal JSX.

thorn0: The change we discuss doesn't affect that special inline quasi-JSX. Pure JSX paragraphs where it happens are plain JSX.

**Prettier pr-6949**
[Playground link](https://deploy-preview-6949--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEATABAHlQSwG4D0AfOqiADQgQAOMO0AzsqAIYBObEA7gArsJMULPBBxlKAIzYswAazgwAytRk4oAc2Qw2AVziUAFjAC2AGwDqBnPAYqwcRQOv5rAT2TgGTSmoZw2MDzS6sYsyABmLKZ+lABWDAAeAELScgqKLMZwADJqcBFRMSDxCYpq6qZwAIo6EPAF0fogKmx+bB7GqAkUzWxqMOZiMAbIABwADJTUnH7m0tQe03BtePmUAI618EE0giAsDAC0UHBwqGc9bHCbOFdBLCFhSJGNlH7GOFq6TQzllTV1fLPQpNGAsCSDVDDZAAJko2hYOFM5QAwhBjKEPFBoGsQDo-AAVcGCF5FPB6ACSUHOsEUYD6tAAgtTFDBXJUGn4AL5coA)
```sh
--parser mdx
```

**Input:**
```mdx
d <div/> d
```

**Output:**
```mdx
d <div/> d

```

JounQin: Then the different outputs are very confusing and make refactoring bad experience as I mentioned. Current behavior doesn't change the finally output content and easier for document author.

JounQin: And I still cann't get your point that why we must change current behavior which works just fine and friendly to document authors.

thorn0: @JounQin  Uh, okay, you convinced me.

@Tigge Seems like this PR is blocked by this problem.

src/language-markdown/embed.js
```diff
@@ -57,7 +57,10 @@ function embed(path, print, textToDoc, options) {
     case "importExport":
       return textToDoc(node.value, { parser: "babel" });
     case "jsx":
-      return textToDoc(node.value, { parser: "__js_expression" });
+      return textToDoc(`<>${node.value}</>`, {
```
JounQin:
```suggestion
      return textToDoc(`<$>${node.value}</$>`, {
```

`<$></$>` is not considered as `jsx` in `mdx`, so it will not break `</>` error location reports literally.

tests/mdx/__snapshots__/jsfmt.spec.js.snap
JounQin: @thorn0 Glad to hear that. 🍺

Then I'm going to finish my original PR based on @Tigge's great work, maybe tonight.

Tigge: I'm a little bit unsure from the discussion what is the problem. Is it the `{' '}` blocks or the lack of inline formatting? Or something else?

JounQin: @Tigge Inline text format whatever prefix or suffix, whitespace or non-whitespace.

thorn0: @Tigge The requirement seems to be as follows: in block-level JSX (paragraphs) that end with text or a JSX expression (not an element or a fragment), the line break should not be inserted between the last tag of the last element (or fragment) and the JSX node that follows it.

So implementation-wise, we need to tweak the way we print `JSXText` and `JSXExpressionContainer` nodes if they are descendants of a node marked with `__maybe_mdx_adjacent`. Might be not easy because the JSX printing code in Prettier is quite sophisticated.

Also it's not clear if we should respect `--prose-wrap`. It's ignored in JSX paragraphs now:

**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAZCAnOBbABOrPAnfTE4osyw0mi28qh6xup+j9rtn1vlgcyGdeggDpQJEgDwUAfBJH9h3ZUsHqVorWtUa9O-ds2bcuGQHp5IADQgIABxgBLaAGdkoAIYYMEAO4ACj4IHiheAG4QzgAmtiAARhheYADWcDAAyg4pzlAA5sgwGACucHYAFjDYADYA6hXO8G45YHCZoU3OEU0Ansjgbh52eW5wGDCByfnYXsgAZl41Y3YAVm4AHgBCyWkZmV7YcKh5cAtLKyDrG5l5+TVwAIolEPDny+UgORhjGAPYMQ28QcGDyMDqsRgFWQAA4AAx2EEQMZ1ZIOAZQOARcbxACOL3gU0cYRAXjcAFpMXAYtT4lh8c4sFMvDM5khFh87GNsM4iqVPm47g9nq8zuyLp8YF4EhCYlDkAAmOzFLzOGp3ADCEGwswx0DOdhKYwAKtKwhzLhEygBJKA02CZMCgpwAQTtmRgvQe7zGAF9fUA)
```sh
--parser mdx
--prose-wrap never
```

**Input:**
```mdx
Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem 


<Lorem>
Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem   
</Lorem>
```

**Output:**
```mdx
Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem

<Lorem>
  Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem
  Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem Lorem
  Lorem Lorem Lorem Lorem Lorem Lorem
</Lorem>

```

src/language-markdown/embed.js
thorn0: @JounQin Can you expand on this? What's wrong with using a fragment here?

JounQin: @thorn0 Imagine original code `123</>456`, after wrapped it's `<>123</>456</>`, it's definitely invalid jsx, but babel's error location would be wrong even if we catch the exception and try to restore the original error location.

With `<$>123</>456</$>`, it's definitely clear where is the error location and we restore it by catching parsing error.

See my PR for detail.

Actually, if we've merged it three months ago, it will not be an issue like current one here.

thorn0: Why would we wrap `123</>456` in a fragment?

thorn0: So you mean, let's say we have this code:
```
<Test>123</>456</Test>
```
The MDX parser decides that it's block JSX and we pass it to Babel. Babel tries to parse it and says that this code is invalid.

Now let's suppose we replaced `<>...</>` with `<$>...</$>`. What has changed? If we parse this:
```
<Test>123</$>456</Test>
```
we'll find ourselves in the same situation.

Please tell me what I'm missing.

thorn0:
@JounQin I added a commit to this PR where I tried to address your comments. Please [play with it](https://deploy-preview-6949--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAkMOCMICEBDAL3wF98AzHCdU-AcgDoB6AMVQAeMAK4447ADpQM2PPhJlhNOkSYsOPfkNHipUuANkF8kKAGdjjCBFIBeeVMIALVEjYADXsIAmAczj4ACTAlFSc3qhmmAA2FACeAHIU6AEAPqkcoey07gA0jvgUwjBOuGZuANoFROxMcN7s+VBsLexmFNHRcZLNRAC6UrR6UADE+AAScJ0QuaQA7rjR3gCEw44AkuysFM0Uc3BmLAGEmBQ4FL7nmE6cwwA8ywC0j-ib2yYsKbD4zwB897xrPgAEa+WzAdhiBq0f69Uh3Sg4Ui-Fzw7iI2EteTOKbRGwtIZw0gkUjcABUhSgEBKcCRkHQXwIZO4hLufGsv3wAEYAEwAZgALABWYbuMVSeAWD7eODAvFgADWUjF7lFYoAVmYpOgIN5hNE4JwDEYzPh7OxJTAeiq1e4zE4pFBMKxUD9fFYIMrxVApBkACoHYz+ihy474VK+0hIZ4vUgZaMxwgRqAZQgAYQg0S5SfwGeiPLjw2NuEsMoYRWilkIAApgGAXEsxHDaABKM2cu74CIAN1+JMI9dQjYQtDZPdhIFyIAgmBgqGgZmQoDOzDmAAUzghFygKN2IKhvJOQMDzoq4DAAMqnMCoKC+ZAwHDCOBTpwwdDRADqLkl17gF63VA527IC4mQcAzEXKdbzMWkYDXc5fHQChkHLaJYKnTUBDIU8FXPC9kjgAAZW84FQjoMJALCL1vXwDQARWEakyKQNDKNOHBYJwcD0G8AQj0wHBbxgT8DxKZAAA4AAYp0EiBYM-K5wMEg5aW7MipwARyY+AEJnbcQAoMxHigOB6nqI8xG01AxAQi5kPI9CXxAWD0FQB8n2csxaIYnSWLY5yYBDUTvHEpAeSnR8KCHWiMwZFCUCpUyj2EWC-RDbcAqnbtn3WKAZVgC8wCE2cAEF8ovGA4gNRzYNoWggA) and tell me what you think now.

src/language-markdown/embed.js
JounQin: `<Test>123</$>456</Test>` you are right here, but it is very uncommon to use `<$>` in `mdx` or `jsx`, it's a balance choose to me.

tests/mdx/__snapshots__/jsfmt.spec.js.snap
JounQin: > Also it's not clear if we should respect `--prose-wrap`.

If it does not affect the final output and user experience, we should not address it here IMO.

src/language-markdown/embed.js
thorn0: I still think there isn't any issue here. If Prettier can't parse an embedded language block (e.g. JSX in MDX), it just skips this block and formats the rest of the file. The error you might see on the playground is playground-only. The user doesn't see it normally. So if an invalid JSX block is encountered, it simply stays unformatted. What is the problem?

tests/mdx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -276,7 +276,8 @@ printWidth: 80
 
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
```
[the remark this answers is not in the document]
thorn0: Oh. So far it's the only incompatibility between MDX's block-level JSX and React. Before trying to fix this, I'd first make sure it's not a bug.

src/language-markdown/embed.js
JounQin: A correct error message is better for user to find output why they're not formatted as expected.

Besides, I'm the author of [eslint-mdx](https://github.com/rx-ts/eslint-mdx) which works very well with `eslint-plugin-prettier`, so the error location does make sense.

thorn0: The user doesn't see any error message about embedded languages in Prettier. What you see on the playground is debug-only information.

JounQin: I mean the error message from `eslint`.

thorn0: Does is get that error from Prettier?

tests/mdx/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
JounQin: Yep, another evidence proving `mdx` is not simply `jsx`, right?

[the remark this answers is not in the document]
thorn0: The spec has clear distinction between block-level and inline JSX. And so far it looked like the block-level one is supposed to be 100% React-compatible so that it was possible to copy-paste React code from JS to MDX. Could the creators of MDX deliberately break this use case by introducing this incompatibility? I'm not sure. We need to ask them.

src/language-markdown/embed.js
JounQin: I'm not for sure, it emits from rule `prettier/prettier`.

But whatever, my point is that we can do better with correct error location, then why should we not?

tests/mdx/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
JounQin: cc @johno @wooorm Any advices here?

src/language-markdown/embed.js
thorn0: Because currently there is no error at all, so I don't understand what we're even talking about.

tests/mdx/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
thorn0: Also asked here https://spectrum.chat/mdx/general/whitespace-in-block-level-jsx~38817f54-db82-4971-be9e-584e1c4d20ff

src/language-markdown/embed.js
thorn0: But okay, this change is too simple to argue about it, I'll add it

tests/mdx/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
thorn0: @JounQin What would you say to closing this to merging this PR as is and opening an issue about that whitespace?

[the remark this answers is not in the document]
JounQin: Cool to me then, but we should post an issue with reference first before merging.

evilebottnawi:
feel free to merge :+1:

tests/mdx/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
thorn0: https://github.com/prettier/prettier/issues/6961

Tigge:
A few hours at work and a lot of things can happen. Thank you all for all the input, help and updates to this merge request!

PR #6954, opened 2019-11-14
Update `jest` to v24

fisker:

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "eslint-plugin-prettier": "3.1.1",
     "eslint-plugin-react": "7.16.0",
     "execa": "3.2.0",
-    "jest": "23.3.0",
+    "jest": "24.9.0",
     "jest-junit": "9.0.0",
     "jest-snapshot-serializer-ansi": "1.0.0",
     "jest-snapshot-serializer-raw": "1.1.0",
--- a/tests_integration/__tests__/__snapshots__/load-toml.js.snap
+++ b/tests_integration/__tests__/__snapshots__/load-toml.js.snap
@@ -11,18 +11,12 @@ Object {
       8002,
     ],
     "server": "192.168.1.1",
-    Symbol(type): Symbol(table),
-    Symbol(declared): true,
   },
   "owner": Object {
     "dob": 1979-05-27T15:32:00.000Z,
     "name": "Tom Preston-Werner",
-    Symbol(type): Symbol(table),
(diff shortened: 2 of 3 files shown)


evilebottnawi:
good job, need look what is problem with CI

fisker:
what is the problem could be

fisker:
strange thing, `windows` and `linux` passed, only fails on `macOS`

fisker:
Should be a `toMatch` assert from `jest` or `jasmine` bug

tests_integration/__tests__/loglevel.js
```diff
@@ -1,21 +1,22 @@
...
 const runPrettier = require("../runPrettier");
+const stripAnsi = require("strip-ansi");
 
 test("do not show logs with --loglevel silent", () => {
   runPrettierWithLogLevel("silent", null);
 });
 
 test("do not show warnings with --loglevel error", () => {
-  runPrettierWithLogLevel("error", ["[error]"]);
+  runPrettierWithLogLevel("error", [/\[error]/]);
```
fisker: @thorn0 do you think I should restore this to string?

thorn0: I have no idea why those tests failed. It's really difficult to read them because their output is cluttered with uninterpreted color codes. But if you don't mind, let's try to restore it and see what happens.

fisker: The failure reason already found, on macOS `[error]` is colored, so I use stripAnsi remove color first.

this change doesn't mater anymore, `string` also passing

thorn0: Great. Let's revert this change then.

fisker: sure

fisker:
merge?

tests_integration/__tests__/loglevel.js
```diff
@@ -40,11 +41,13 @@ function runPrettierWithLogLevel(logLevel, patterns) {
 
   expect(result).not.toEqual(0);
```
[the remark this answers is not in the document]
SimenB: drive-by comment: should probably not be negated at all to assert on the actual value instead of what it's _not_ to avoid false negatives such as this.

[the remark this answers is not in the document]
fisker: I'm going to change it to

```js
expect(result.status).toEqual(2);
```

Issue #6936, opened 2019-11-12
Add `quoteProps: es3`


ExE-Boss:
When contributing new&nbsp;polyfills to&nbsp;[**ES‑Shims**](https://github.com/es-shims/es-shim-api), which&nbsp;have to&nbsp;use&nbsp;**ES3**, I&nbsp;still&nbsp;want to&nbsp;use&nbsp;Prettier for&nbsp;**ES‑Shim**&nbsp;packages that&nbsp;I&nbsp;own.

Since&nbsp;**Prettier**&nbsp;strips the&nbsp;quotes required by&nbsp;**ES3** for&nbsp;object&nbsp;properties named&nbsp;with&nbsp;keywords, I&nbsp;have&nbsp;to&nbsp;run&nbsp;`eslint ‑‑fix` after&nbsp;running&nbsp;`prettier --write`, which&nbsp;is&nbsp;annoying.

**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#.....)

**Input:**
```js
function __generator(thisArg, body) {
    var g;
    return g = { next: verb(0), "throw": verb(1), return: verb(2) }, typeof Symbol === "function" && (g[Symbol.iterator] = function() { return this; }), g;
    function verb(n) { return function (v) { return step([n, v]); }; }
    function step(op) {
        // …
    }
}
```

**Output:**
```js
function __generator(thisArg, body) {
	var g;
	return (
		(g = { next: verb(0), throw: verb(1), return: verb(2) }),
		typeof Symbol === "function" &&
			(g[Symbol.iterator] = function() {
				return this;
			}),
		g
	);
	function verb(n) {
		return function(v) {
			return step([n, v]);
		};
	}
}
```

**Expected behavior:**

```js
function __generator(thisArg, body) {
	var g;
	return (
		// `throw` and `return` are quoted,
		// as they're invalid property names in ES3.
		(g = { next: verb(0), "throw": verb(1), "return": verb(2) }),
		typeof Symbol === "function" &&
			(g[Symbol.iterator] = function() {
				return this;
			}),
		g
	);
	function verb(n) {
		return function(v) {
			return step([n, v]);
		};
	}
}
```

j-f1:
@ExE-Boss I‘d recommend using [`prettier-eslint`](https://github.com/prettier/prettier-eslint-cli) to apply these fixes after Prettier runs.

[...]

PR #6963, opened 2019-11-15
Set `trailingComma` default value to `es5` 

fisker:
#6910


fisker:
@evilebottnawi come on it's 2019 still some `engines.node` requires `none`

evilebottnawi:
@fisker no, but some people can write es3 code in theory can inserting trailing commas can break code, but if we don't support es3 we can set `es5` by default, but for better 0CJS we can respect node version

fisker:
Self checked, didn't find any wrong result.

This is hard to review, but I think it's ready

src/language-js/options.js
```diff
@@ -76,16 +76,17 @@ module.exports = {
...
     ],
     description: "Print trailing commas wherever possible when multi-line.",
     choices: [
-      { value: "none", description: "No trailing commas." },
       {
         value: "es5",
         description:
           "Trailing commas where valid in ES5 (objects, arrays, etc.)"
       },
+      { value: "none", description: "No trailing commas." },
```
ExE-Boss: https://github.com/prettier/prettier/pull/6910#pullrequestreview-314631630

fisker: https://github.com/prettier/prettier/pull/6910#discussion_r346168674

> @ExE-Boss why? after changing default to es5, shouldn't we put it on the top?

tests/angular_component_examples/jsfmt.spec.js
```diff
@@ -1,2 +1,2 @@
+run_spec(__dirname, ["typescript"], { trailingComma: "none" });
```
fisker: https://github.com/prettier/prettier/pull/6910#discussion_r344485493

> I change the test orders to help review, snap changes should only be `options` part

lydell:
@evilebottnawi We already turned down #6936 so adding automatic support for ES3 here won’t make sense. I think we should keep things simple. `es5` is always the default.

lipis:
So merginnnnn?

evilebottnawi:
Let's do it

fisker:
👍

kachkaev:
@fisker I'm working on #6929 and am about to create `changelog_unreleased/api/pr-7430.md` as asked in https://github.com/prettier/prettier/pull/7417#issuecomment-577651439. By looking at your PR, I cannot find a corresponding entry to use it as an example. If I simply can't find it, could you please share a link? Otherwise, could you please add a corresponding blogpost section for your change? I'll write mine is a similar style.

fisker:
@kachkaev I didn't wrote one, if it's required, hope someone can help me out. I'm really bad at docs and English :(

PR #6973, opened 2019-11-16
Fix loglevel test

fisker:
found this problem when upgrading `jest` https://github.com/prettier/prettier/pull/6954

according to [jasmine toMatch source](https://github.com/jasmine/jasmine/blob/15f969bee7066526ba44e562407c728d4a1740df/src/core/matchers/toMatch.js#L22), string are treat as RegExp, original `"[error]"` are not testing against `[error]` string, `[` should be escaped. I changed patterns to RegExp Literals, easier to write.

### DIFF
--- a/tests_integration/__tests__/loglevel.js
+++ b/tests_integration/__tests__/loglevel.js
@@ -7,15 +7,15 @@ test("do not show logs with --loglevel silent", () => {
 });
 
 test("do not show warnings with --loglevel error", () => {
-  runPrettierWithLogLevel("error", ["[error]"]);
+  runPrettierWithLogLevel("error", [/\[error]/]);
 });
 
 test("show errors and warnings with --loglevel warn", () => {
-  runPrettierWithLogLevel("warn", ["[error]", "[warn]"]);
+  runPrettierWithLogLevel("warn", [/\[error]/, /\[warn]/]);
 });
 
 test("show all logs with --loglevel debug", () => {
-  runPrettierWithLogLevel("debug", ["[error]", "[warn]", "[debug]"]);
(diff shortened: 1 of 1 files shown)


fisker:
wait

lydell:
We wait for green CI before merging :)

fisker:
no, it should fail

lydell:
I understand that. But after it has failed in we’ve verified stuff you are going to make it pass, right? I meant that we won’t press Merge until then.

fisker:
yes

fisker:
strange, CI passed. It should fail, because `[error]` is colored, this is why #6954 fails, and only on macOS

lydell:
Maybe you can add this commit to #6954 instead if it makes things easier?

fisker:
I think master also requires this fix, let me check #6954 for sure

fisker:
I've confirmed, only jest@24 on macOS has color, this is good enough for `master`

thorn0:
@fisker 

> according to jasmine toMatch source

Unless I am missing something, we don't use Jasmine at all. Jest's `toMatch` [doesn't convert strings to regexps](https://github.com/facebook/jest/blob/116303b/packages/expect/src/matchers.ts#L837).

fisker:
@thorn0  

thank you.

you are right, I tested

```
expect('f').toMatch('[foo]')
```

it fails as expected.

I thought `jest` is using `jasmine` asserts. looks like I made a misstake.

thorn0:
`expect` used to be an independent package, but it was [donated to Jest](https://github.com/facebook/jest/issues/1679).

fisker:
I was told `jest` is based on `jasmine`, long ago. never looked into it

PR #6974, opened 2019-11-16
set registry to https://registry.yarnpkg.com/

fisker:
I have access to `yarnpkg`, but my default `registry` is not `yarnpkg` on all my computers, so every time when I want change `yarn.lock` on this repo, I have to use

`yarn --registry https://registry.yarnpkg.com/`
`yarn --registry https://registry.yarnpkg.com/ add some-dep --dev`

adding this `.npmrc` can lock registry to yarnpkg when work on this repo

This might not useful for many developers, but really can help other developers like me.

approve?

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.yarnpkg.com/


fisker:
@lydell @evilebottnawi

evilebottnawi:
I am :-1: on this, why do not use default register?

fisker:
@evilebottnawi network issue, in China we normally use `taobao` mirror for work, almost every one

fisker:
You might not notice, others change registry too,

https://github.com/prettier/prettier/blob/master/yarn.lock#L2411-L2415

`registry.npmjs.org` is used, commited by @ikatyang

JounQin:
I don't think it is a good practice.

> @evilebottnawi network issue, in China we normally use `taobao` mirror for work, almost every one

I've never used it, you can always use proxy instead.

fisker:
@JounQin 

> I have access to `yarnpkg`

JounQin:
> I have access to yarnpkg

So there is no reason to use `taobao` registry then.

fisker:
Did this PR effect anyone installing dependencies? 
no!

What changes? 
an extra file.

Benefits? 
avoid yarn.lock changes from another registry, 
help people using different settings easier to contribute

fisker:
But never mind.

PR #6975, opened 2019-11-16
Update `ignore` to v5

fisker:


src/cli/util.js
```diff
@@ -443,7 +443,11 @@ function formatFiles(context) {
   }
 
   eachFilename(context, context.filePatterns, filename => {
-    const fileIgnored = ignorer.filter([filename]).length === 0;
+    let fileIgnored = false;
+    try {
+      fileIgnored = ignorer.ignores(path.relative(process.cwd(), filename));
+    } catch (_) {}
```
fisker: empty catch not allowed? `eslint-disable` or `change eslintrc` ?

src/cli/util.js
```diff
@@ -443,7 +443,13 @@ function formatFiles(context) {
   }
 
   eachFilename(context, context.filePatterns, filename => {
-    const fileIgnored = ignorer.filter([filename]).length === 0;
+    let fileIgnored = false;
+    try {
+      fileIgnored = ignorer.ignores(path.relative(process.cwd(), filename));
+    } catch (_) {
+      // eslint-disable-line no-empty
```
lydell: You can change this to `// Do nothing`. A comment inside a block counts as non-empty in ESLint. This way it’s explicit that you actually intended to ignore the error.

lydell: Btw, why _are_ we ignoring errors here?

fisker: https://github.com/kaelzhang/node-ignore#upgrade-4x---5x

`ignorer.ignores('../file-from-parent')`

throws

fisker:
```suggestion
      // Invalid path
```

src/common/get-file-info.js
```diff
@@ -68,8 +68,15 @@ function _getFileInfo({
   resolveConfig = false,
   sync = false
 }) {
+  let ignored = true;
+  try {
+    ignored = ignorer.ignores(path.relative(process.cwd(), filePath));
+  } catch (_) {
+    // eslint-disable-line no-empty
```
fisker:
```suggestion
    // Invalid path
```

src/cli/util.js
fisker: I did tried `isValidPath`, but they are the same.

fisker: > You can change this to `// Do nothing`. A comment inside a block counts as non-empty in ESLint. This way it’s explicit that you actually intended to ignore the error.

thanks

src/cli/util.js
```diff
@@ -443,7 +443,13 @@ function formatFiles(context) {
   }
 
   eachFilename(context, context.filePatterns, filename => {
-    const fileIgnored = ignorer.filter([filename]).length === 0;
+    let fileIgnored = false;
+    try {
+      fileIgnored = ignorer.ignores(path.relative(process.cwd(), filename));
+    } catch (_) {
+      // Invalid path
```
lydell: Would it be better to do this update in `next` instead and show this error to the user? I’m not sure what an “invalid path” is, but it sounds useful to get to know that some of my patterns are wrong?

fisker: I don't know. I'm not changing current behave

lydell:
```suggestion
      // Invalid path. TODO: Should we show this error to the user, rather than swallowing it?
```

lydell: @evilebottnawi Do you have experience from ESLint or stylelint?

src/cli/util.js
```diff
@@ -443,7 +443,13 @@ function formatFiles(context) {
   }
 
   eachFilename(context, context.filePatterns, filename => {
-    const fileIgnored = ignorer.filter([filename]).length === 0;
+    let fileIgnored = false;
+    try {
+      fileIgnored = ignorer.ignores(path.relative(process.cwd(), filename));
+    } catch (_) {
+      // Invalid path. TODO: Should we show this error to the user, rather than swallowing it?
```
evilebottnawi: Maybe we should add note for next major release about that?

fisker:
Not sure what else I need do

fisker:
After reading  https://github.com/sindresorhus/globby/pull/120 , https://github.com/sindresorhus/globby/pull/112 and  https://github.com/sindresorhus/globby/pull/104 I think  I did something wrong

Issue #6969, opened 2019-11-15
Redundant parens inserted around JSX inside NewExpressions


rtsao:
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEU4HcAEAxCEAUAPAMIQC2ADtAjJgPQB8AlANwgA0IEFMAltAGdkoAIYAnMRHQAFcQiEoRANwi8AJhxAAjMSLABrODADKFPbygBzZDDEBXOJwAWMMgBsA6k97wBZsHDG8j68Sj4AnsjgAkKcFgJwYjDSupZkIsgAZiJuCZwAVgIAHgBCugZGxiJkcAAyFnBZOXkghUXGFpZucACKdhDwTbmOIGZiCWJRWiJacG6aFGIWMB7qME7IABwADJyLEAkeuhRRi3ATSo2cAI798CncCiAiAgC0aHBqn5picLe8vxSIjSGSQ2WGnASZF4NnsIwEnW6fQGjTBzRGMBmqzU62QACZOLYRLw3J1SGR0lEoNRNHYEgAVGYKcEtJQOACSUC+sGMYCWPAAglzjDBwt0hgkAL6SoA)
```sh
--parser babel
```

**Input:**
```jsx
new Foo(<Component />);
```

**Output:**
```jsx
new Foo((<Component />));

```

**Expected behavior:**

```jsx
new Foo(<Component />);
```

No parens should be inserted (like is the case when the `new` keyword is omitted)

thorn0:
Thanks for the issue. Would you like to submit a PR with a fix for it? It looks simple enough to be a good first contribution.

PR #6980, opened 2019-11-16
Fix redundant parens around JSX inside NewExpressions

rtsao:
Fixes https://github.com/prettier/prettier/issues/6969


tests/jsx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4097,6 +4097,8 @@ f?.(<div/>);
 (<div/>)();
 (<div/>)?.();
 
+new Foo(<Component/>);
```
lydell: Could you add a test checking that `new Foo((<Component/>))` is turned into `new Foo(<Component/>)` as well?

Issue #5420, opened 2018-11-09
Less Parse Error: Variable Body that starts with `&`


fabb:
**Prettier 1.15.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAVgVwM4wLTwAe+MAngA5x4CWU5m+2cANnGDBAE7ZIAEAZEigQYACgDaZSgF4A5ADNqrWQF0AlEJESpcOWAAWbANYAjCIVUbhYyRV2zOAQwAm1CJc02dcp1ADmcB7W2nZy+tTOzgiWANwAOlAgADQgEOQwblDYyKCOnJwQAO4ACnkI2SiOAG4QEckgJk5gRnAwAMrkjmC0fsgwnJhwKfowALbMAOrh8NidYHBt5dQZVcukyODY2Sm0TJwwxU5+o47I8o7MTCno2IQAQk0t7Y6jcAAytHBnF1cgN4RtHqsACKmBEXyQ50uQxAnW4cE4G1YW3q5E4tBgEwiMH0yAAHAAGFJoiBMCZOcgbNFwPZVL4pThwACOmGojMOjmOp0hPxhTFG1D6Az5QLgoPB32hKRgjhMWOcOOQACZpU4lD0AMIQUYnDbCKD0kA4OAAFVlFShTAAvlagA)
```sh
--parser less
```

**Input:**
```less
@just-text-type-input-selectors: &:not([type='file']):not([type='checkbox']):not([type='radio']):not([type='range']):not([type='hidden']);

```

**Output:**
```less
SyntaxError: (postcss) CssSyntaxError Missed semicolon (1:34)
> 1 | @just-text-type-input-selectors: &:not([type='file']):not([type='checkbox']):not([type='radio']):not([type='range']):not([type='hidden']);
    |                                  ^
  2 | 
```

**Expected behavior:**
The line should parse fine, since it is valid less syntax, and the less compiler parses this fine.

evilebottnawi:
@fabb Looks as invalid Less syntax can you provide example where it is work?

fabb:
here is an example:
```
@just-text-type-input-selectors: &:not([type='file']):not([type='checkbox']):not([type='radio']):not([type='range']):not([type='hidden']);

input {
	@{just-text-type-input-selectors} {
		color: blue;
	}
}
```

E.g. these online less compilers translate it correctly:
* http://lesscss.org/less-preview/
* http://ecomfe.github.io/est/fiddle/


[...]

Issue #4524, opened 2018-05-22
[less] mixin parameters not prettified


benmerckx:
**Prettier 1.12.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEACVAdKA6AtgSwA98oAKYLdABwgGd8Z9olUBDAI1ogBsBXeANyVUAE3y0q3VgE8WAM25xCQqOgDu+ETAAWLAIwAGAwFIV6bXHwBzbTH1HTw7Bp0lUFVelQQqrMA1lUAzMvRTk7IJD0GB8WYOEAXywEgEoBEAAaEB9GaFpkUFYAJyKINQAFYoR8lFYANwhNTJB2Ir8AazgYAGVffygrZBgi3jgsklo4IphytqtcVmQ5Vm5JrIArWkIAITawTp7WXDgAGRI4JZW1kF8iyaLkEEVafKyqIpIYAHVNHWQADgMb1Kky+bSoj3ecHudQuWSKcAAjrx8AjZqx5oskMtVmMQJMCEMRnj6ANFABFXgQeCXXFZGAcH5abTIABM9La+G4JCsAGEILgFo8oNA4SBeJMACocGo4yYJBJAA)
```sh
--parser less
```

**Input:**
```less
  
.mixin({
  position: absolute;
  display: flex;
  width: 100%;
  height: 100%;
  .within {
    opacity: 0;
    left: 0;
    top: 0;
  }
});
```

**Output:**
```less
.mixin(
  {position: absolute; display: flex; width: 100%; height: 100%; .within {opacity:
    0; left: 0; top: 0;}}
);

```

**Expected behavior:**

evilebottnawi:
@benmerckx known issue, will be fixed in `1.13`, thanks for issue

evilebottnawi:
FYI: need use `postcss-values-parser` for less `mixin`

evilebottnawi:
Problem in parser `css-decl` should contain all properties with values into `nodes` field. Right now we have this:
```
Rule {
      raws: 
       { before: '\n  ',
         between: '',
         semicolon: true,
         after: '',
         selector: '.mixin({\n    position: absolute;\n    display: flex;\n    width: 100%;\n    height: 100%;\n    .within {\n      opacity: 0;\n      left: 0;\n      top: 0;\n    }\n  })',
[...]

PR #6981, opened 2019-11-17
Update `postcss-less` to v3

fisker:

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -406,7 +406,7 @@ function parseNestedCSS(node, options) {
       }
 
       // Whitespace between variable and colon
-      if (node.name.includes(":") && !node.params) {
+      if (node.name.indexOf(":") !== -1 && !node.params) {
         node.variable = true;
         const parts = node.name.split(":");
         node.name = parts[0];
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -385,6 +385,12 @@ function parseNestedCSS(node, options) {
       node.value = parseValue(value);
     }
 
+    // Less variable
+    if (node.type === "css-atrule" && node.variable) {
+      node.params = parseValue(node.params);
+      return node;
+    }
+
     if (node.type === "css-atrule" && params.length > 0) {
       const { name } = node;
       const lowercasedName = node.name.toLowerCase();
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -199,10 +199,15 @@ function parseSelector(selector) {
   // the content of the comment as selectors which turns it into complete
   // garbage. Better to print the whole selector as-is and not try to parse
   // and reformat it.
-  if (selector.match(/\/\/|\/\*/)) {
+  if (/\/\/|\/\*/.test(selector)) {
+    let value = selector.replace(/^ +/, "").replace(/ +$/, "");
+    // last line has inline comment
+    if (/\/\//.test(selector.split("\n").pop())) {
+      value += "\n";
+    }
     return {
       type: "selector-unknown",
-      value: selector.replace(/^ +/, "").replace(/ +$/, "")
+      value
     };
   }
 
@@ -385,17 +390,17 @@ function parseNestedCSS(node, options) {
       node.value = parseValue(value);
     }
 
-    // Less whitespace between variable and colon
-    if (
-      node.type === "css-atrule" &&
-      node.params.startsWith(":") &&
-      node.name !== "custom-selector"
-    ) {
-      // if (!(options.parser === "css" && node.name === "custom-selector")) {
-      node.variable = true;
-      node.params = node.params.slice(1);
-      // }
-    }
+    // // Less whitespace between variable and colon
+    // if (
+    //   node.type === "css-atrule" &&
+    //   node.params.startsWith(":") &&
+    //   node.name !== "custom-selector"
+    // ) {
+    //   // if (!(options.parser === "css" && node.name === "custom-selector")) {
+    //   node.variable = true;
+    //   node.params = node.params.slice(1);
+    //   // }
+    // }
 
     // Less whitespace between variable and colon
     if (node.type === "css-atrule" && node.name.includes(":") && !node.params) {
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -413,7 +413,7 @@ function parseNestedCSS(node, options) {
         node.value = parseValue(parts.slice(1).join(":"));
       }
 
-      // Whitespace between variable and colon
+      // Missing whitespace between variable and colon
       if (
         ["page", "nest"].indexOf(node.name) === -1 &&
         node.params &&
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -376,8 +376,11 @@ function parseNestedCSS(node, options) {
     }
 
     if (node.type === "css-atrule") {
+      const isLessParser =
+        options.parser === "css" || options.parser === "less";
+
       // mixin
-      if (node.mixin) {
+      if (isLessParser && node.mixin) {
         const source =
           node.raws.identifier +
           node.name +
@@ -389,12 +392,12 @@ function parseNestedCSS(node, options) {
       }
 
       // function
-      if (node.function) {
+      if (isLessParser && node.function) {
         return node;
       }
 
       // only css support custom-selector
-      if (node.name === "custom-selector" && options.parser === "css") {
+      if (options.parser === "css" && node.name === "custom-selector") {
         const customSelector = node.params.match(/:--\S+?\s+/)[0].trim();
         node.customSelector = customSelector;
         node.selector = parseSelector(
@@ -405,7 +408,7 @@ function parseNestedCSS(node, options) {
       }
 
       // Whitespace between variable and colon
-      if (node.name.indexOf(":") !== -1 && !node.params) {
+      if (isLessParser && node.name.indexOf(":") !== -1 && !node.params) {
         node.variable = true;
         const parts = node.name.split(":");
         node.name = parts[0];
@@ -414,6 +417,7 @@ function parseNestedCSS(node, options) {
 
       // Missing whitespace between variable and colon
       if (
+        isLessParser &&
         ["page", "nest"].indexOf(node.name) === -1 &&
         node.params &&
         node.params[0] === ":"
@@ -423,7 +427,7 @@ function parseNestedCSS(node, options) {
       }
 
       // Less variable
-      if (node.variable) {
+      if (isLessParser && node.variable) {
         delete node.params;
         return node;
       }
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -434,6 +434,12 @@ function parseNestedCSS(node, options) {
       const { name } = node;
       const lowercasedName = node.name.toLowerCase();
 
+      if (lowercasedName === "import") {
+        node.import = true;
+        node.filename = node.params;
+        return node;
+      }
+
       if (name === "warn" || name === "error") {
         node.params = {
           type: "media-unknown",
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -393,6 +393,7 @@ function parseNestedCSS(node, options) {
         return node;
       }
 
+      // only css support custom-selector
       if (node.name === "custom-selector" && options.parser === "css") {
         const customSelector = node.params.match(/:--\S+?\s+/)[0].trim();
         node.customSelector = customSelector;
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -200,14 +200,9 @@ function parseSelector(selector) {
   // garbage. Better to print the whole selector as-is and not try to parse
   // and reformat it.
   if (/\/\/|\/\*/.test(selector)) {
-    let value = selector.replace(/^ +/, "").replace(/ +$/, "");
-    // last line has inline comment
-    if (/\/\//.test(selector.split("\n").pop())) {
-      value += "\n";
-    }
     return {
       type: "selector-unknown",
-      value
+      value: selector.trim()
     };
   }
 
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -288,16 +288,6 @@ function parseNestedCSS(node, options) {
       node.raws.selector = selector;
     }
 
-    // postcss-less@2.0.0 parse `custom-selector` as `css-decl`
-    if (
-      options.parser === "css" &&
-      node.type === "css-decl" &&
-      node.prop === "@custom-selector"
-    ) {
-      selector = node.value;
-      node.raws.value = selector;
-    }
-
     let value = "";
 
     if (typeof node.value === "string") {
@@ -390,35 +380,7 @@ function parseNestedCSS(node, options) {
       node.value = parseValue(value);
     }
 
-    // // Less whitespace between variable and colon
-    // if (
-    //   node.type === "css-atrule" &&
-    //   node.params.startsWith(":") &&
-    //   node.name !== "custom-selector"
-    // ) {
-    //   // if (!(options.parser === "css" && node.name === "custom-selector")) {
-    //   node.variable = true;
-    //   node.params = node.params.slice(1);
-    //   // }
-    // }
-
-    // Less whitespace between variable and colon
-    if (node.type === "css-atrule" && node.name.includes(":") && !node.params) {
-      node.variable = true;
-      const a = node.name.split(":");
-      node.name = a[0];
-      node.params = a[1];
-    }
-
-    // Less variable
-    if (node.type === "css-atrule" && node.variable) {
-      if (node.params) {
-        node.params = parseValue(node.params);
-      }
-      return node;
-    }
-
-    // Less mixin
+    // mixin
     if (node.type === "css-atrule" && node.mixin) {
       const source =
         node.raws.identifier +
@@ -431,14 +393,41 @@ function parseNestedCSS(node, options) {
     }
 
     if (node.type === "css-atrule" && params.length > 0) {
-      const { name } = node;
-      const lowercasedName = node.name.toLowerCase();
+      if (node.name === "custom-selector" && options.parser === "css") {
+        const customSelector = node.params.match(/:--\S+?\s+/)[0].trim();
+        node.customSelector = customSelector;
+        node.selector = parseSelector(
+          node.params.slice(customSelector.length).trim()
+        );
+        delete node.params;
 
-      if (lowercasedName === "import") {
-        node.import = true;
-        node.filename = node.params;
         return node;
       }
+
+      // Whitespace between variable and colon
+      if (node.name.includes(":") && !node.params) {
+        node.variable = true;
+        const parts = node.name.split(":");
+        node.name = parts[0];
+        node.params = parts.slice(1).join(":");
+      }
+
+      // Whitespace between variable and colon
+      if (node.params.startsWith(":")) {
+        node.variable = true;
+        node.params = node.params.slice(1);
+      }
+
+      // Less variable
+      if (node.variable) {
+        if (node.params) {
+          node.params = parseValue(node.params);
+        }
+        return node;
+      }
+
+      const { name } = node;
+      const lowercasedName = node.name.toLowerCase();
 
       if (name === "warn" || name === "error") {
         node.params = {
@@ -468,6 +457,8 @@ function parseNestedCSS(node, options) {
       }
 
       if (lowercasedName === "import") {
+        node.import = true;
+        node.filename = node.params;
         node.params = parseValue(params);
         return node;
       }
@@ -496,16 +487,6 @@ function parseNestedCSS(node, options) {
         params = params.replace(/^(?!if)(\S+)\s+\(/, "$1(");
 
         node.value = parseValue(params);
-        delete node.params;
-
-        return node;
-      }
-
-      if (name === "custom-selector") {
-        const customSelector = params.match(/:--\S+?\s+/)[0].trim();
-
-        node.customSelector = customSelector;
-        node.selector = parseSelector(params.slice(customSelector.length));
         delete node.params;
 
         return node;
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -413,7 +413,7 @@ function parseNestedCSS(node, options) {
       }
 
       // Whitespace between variable and colon
-      if (node.params.startsWith(":")) {
+      if (["page"].indexOf(node.name) === -1 && node.params.startsWith(":")) {
         node.variable = true;
         node.params = node.params.slice(1);
       }
@@ -459,7 +459,7 @@ function parseNestedCSS(node, options) {
 
       if (lowercasedName === "import") {
         node.import = true;
-        node.filename = node.params;
+        delete node.filename;
         node.params = parseValue(params);
         return node;
       }
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -4,7 +4,7 @@ const createError = require("../common/parser-create-error");
 const parseFrontMatter = require("../utils/front-matter");
 const lineColumnToIndex = require("../utils/line-column-to-index");
 const { hasPragma } = require("./pragma");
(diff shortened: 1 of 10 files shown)


tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -364,13 +364,13 @@ $KeepScssVar: val;
 
   @KeepDetachedRuleset();
 
-  &: EXTEND(.Keep ALL);
+  &:EXTEND(.Keep ALL);
```
fisker: Anyone know what is this? space matters?

lydell: I think I added this test case when I added support for lowercasing of stuff back in the day. Testing on http://lesscss.org/less-preview/, uppercase `EXTEND` does not appear to be valid Less. So it’s a weird test case. `&:extend(.Keep ALL)` seems to be valid syntax, while `&: extend(.Keep ALL)` is a syntax error.

lydell:
Also feel free to switch to the `next` branch if Node.js 4 is giving you trouble – we should not waste more time on that old version.

tests/css_case/__snapshots__/jsfmt.spec.js.snap
fisker: what output is expect?

lydell: My guess is that the input should be changed to `&:extend(.Keep ALL)` and the output should be the same (`&:extend(.Keep ALL)`).

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -217,7 +217,7 @@ $KeepScssVar: val;
 
     @KeepDetachedRuleset();
 
-    &:EXTEND(.Keep ALL);
+    &:extend(.Keep ALL);
```
fisker: https://github.com/prettier/prettier/pull/6981#discussion_r347130066

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -364,13 +364,13 @@ $KeepScssVar: val;
 
   @KeepDetachedRuleset();
 
-  &: EXTEND(.Keep ALL);
+  &:extend(.Keep ALL);
```
fisker: https://github.com/prettier/prettier/pull/6981#discussion_r347130066

scripts/build/babel-plugins/replace-array-includes-with-indexof.js
```diff
@@ -2,29 +2,27 @@
 
 //
 // BEFORE:
-//   [__something__].includes(__something__)
-//   CONSTANT_CASE.includes(__something__)
+//   foo.includes(bar)
```
fisker: changes for this file is from seperate PR #6985

fisker:
ready for review

thorn0:
[`each`](http://lesscss.org/functions/#list-functions-each) is still not handled correctly, but at least it's not a parser error as it was before.

-  `@` should not be added in front of `each`
- whitespace should not be added in front of `@{value}`
- anonymous ruleset should not be formatted inline

**Prettier pr-6981**
[Playground link](https://deploy-preview-6981--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEABAznANnMMIBO6SABAEZYCucANCQOYFwJ1MAmA3ADpQ9wCGYABYAKDNlz4idYDxIkAdJiwBaVMABu-KnAC+JWVHnz+pMtyMldPXQEoOIGiAgAHGAEto6ZKH4ECEADuAAp+CN4o-BoQ7myOIGQEggDWcDAAyi6C7lD0yDAE1E5CMAC2WADqQu7w6FlgcOnhNe4aNQCeyODo3k45mAQwwUn0pfzIAGbamE4AVugAHgBCSWCpGfylcAAyOXCT07Qg8wvpOfQ4AIqUEPAHWDMgWURwBF04PfEuBDkwFbEwITIAAcAAYnN8IJgKkkXF1vnABhp9k4AI43eDDVwRED8dAqKDMNhwOJOJjo9xMYb8UbjJBTB5HTCldz5QpM85XDH7emHJwwfhkf5sQHIABM-KS7iw5wAwhBSmMulBoCiQJRMAAVQURBmPDTUACSUGJsHSYB+bgAgib0jB2jh7phdLogA)
```sh
--parser less
```

**Input:**
```less
@selectors: blue, green, red;

each(@selectors, {
  .sel-@{value} {
    a: b;
  }
});
```

**Output:**
```less
@selectors: blue, green, red;

@each (@selectors, {.sel- @{value} {a: b;}}) ;

```

fisker:
looking into this, different `at-rule` with `function: true`

fisker:
@thorn0 add a test for it, print as it is, maybe we can make it pretty later

fisker:
**Prettier pr-6981**
[Playground link](https://deploy-preview-6981--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEABAznANnMMIBO6SABAEZYCucANCQOYFwJ1MAmA3ADpQ9wCGYABYAKDNlz4idYDxIkAdJiwBaVMABu-KnAC+JWVHnye-UmW5GSunroCUHEDRAQADjACW0dMlD8CBBAA7gAK-gg+KPwaEB5sTiBkBIIA1nAwAMqugh5Q9MgwBNTOQjAAtlgA6kIe8OjZYHAZEbUeGrUAnsjg6D7OuZgEMCHJ9GX8yABm2pjOAFboAB4AQslgaZn8ZXAAMrlwUzO0IAuLGbn0OACKlBDwh1izINlEcATdOL0JrgS5MJVxGBCZAADgADM4fhBMJVkq5uj84IMNAdnABHW7wEZuSIgfjoFRQZhsODxZxMDEeJgjfhjCZIaaPY6YMoeApFZkXa6Yg4Mo7OGD8MgAthA5AAJgFyQ8WAuAGEIGVxt0oNBUSBKJgACpCyKMp4aagASSgJNgGTAv3cAEFTRkYB0cA9MLpdEA)
```sh
--parser less
```

**Input:**
```less
@selectors: blue, green, red;

each(@selectors, {
  .sel-@{value} {
    
a: b;
  }
});
```

**Output:**
```less
@selectors: blue, green, red;

each(@selectors, {
  .sel-@{value} {
    
a: b;
  }
});

```

thorn0:
@fisker :+1: I'm okay with whatever you decide as long as the output stays valid.

thorn0:
#5420 is not a parser error now, but selectors assigned to a variable ([docs](http://lesscss.org/features/#variables-feature-selectors)) get corrupted by whitespace inserted after colons.

**Prettier pr-6981**
[Playground link](https://deploy-preview-6981--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEABAznANnMMIBOSABAGRJQQwAUA2jAJ4AOcAvAOQBmAljuwLoBKClTqMWHMAAtcAawBGEAB4DhlGvWZt2BAIYATbhFUiN47XqgBzOCfVitHKd336EqgNwAdKCAA0IBBMMEZQ6MigugQEEADuAApRCOEougBuEC7+IPJ6YLJwMADKTLpg3NbIMAQArnABUjAAtlgA6s7w6KVgcEXJ3CFpAwzI4OjhARWYBDDxelZNusiculiYAQBW6EoAQnkFxbpNcAAyFXDLq+sgW0pFFVY4AIo1VBdIK2v1IKUE06M4cbZJgECowVouGBSZAADgADAEQRBMK09ExRiC4NM0hcAgBHV7wOZBFIgXToAC0UDgcDc+myBDgBO4jLmugWSw+V2+mCa3CqtR5D2ehPen2uMF08gh+ihyAATAFqrpeA8AMIQJqLUaUanZGqYAAqUpS4u+aTqAEkoG5YEUwKDggBBG1FRg4S5fAC+XqAA)
```sh
--parser less
```

**Input:**
```less
@selector: &:not([type='file']):not([type='checkbox']):not([type='radio']):not([type='range']):not([type='hidden']);

```

**Output:**
```less
@selector: &: not([type= "file" ]): not([type= "checkbox" ]):
  not([type= "radio" ]): not([type= "range" ]): not([type= "hidden" ]);

```

fisker:
variable support selector, maybe we need do a `isSelector` check, but should do in a separate PR

src/language-css/printer-postcss.js
```diff
@@ -189,6 +202,50 @@ function genericPrint(path, options, print) {
...
+                    printNodeSequence(path, options, print)
+                  ])
+                ),
+                softline,
+                "}"
+              ])
+            : "",
+          isTemplatePlaceholderNodeWithoutSemiColon ? "" : ";"
+        ]);
+      }
```
evilebottnawi: Each `node.mixin`/`node.function`/`node.variable` should be:
```
if (isLessParser) { 
  if (node.mixin) { /* ... */ }
  if (node.function) { /* ... */ }
}
```

It is better for performance and other parsers can potentially have same properties

src/language-css/printer-postcss.js
```diff
@@ -132,7 +140,12 @@ function genericPrint(path, options, print) {
         node.important ? " !important" : "",
         node.nodes
           ? concat([
-              " {",
+              node.selector &&
+              node.selector.type === "selector-unknown" &&
+              node.selector.value.endsWith("\n")
```
evilebottnawi: Why? It is bad idea use ` node.selector.value.endsWith("\n")`

src/language-css/printer-postcss.js
```diff
@@ -115,12 +115,20 @@ function genericPrint(path, options, print) {
...
-          return concat(["// ", rawText]);
+        if (node.raws.inline || node.inline) {
+          const needBreakAfter = !(
+            node.source.input.css.split("\n")[node.source.end.line] || ""
+          ).trim();
+          return concat([
+            "//",
+            node.raws.left + rawText,
+            needBreakAfter ? "\n" : ""
+          ]);
```
evilebottnawi: "\n"? We should use prettier builders
`node.source.input.css` should be never used

this place is very bad, need rewrite this

src/language-css/parser-postcss.js
```diff
@@ -422,6 +471,8 @@ function parseNestedCSS(node, options) {
       }
 
       if (lowercasedName === "import") {
+        node.import = true;
+        delete node.filename;
```
evilebottnawi: Why?

src/language-css/parser-postcss.js
```diff
@@ -204,10 +204,15 @@ function parseSelector(selector) {
   // the content of the comment as selectors which turns it into complete
   // garbage. Better to print the whole selector as-is and not try to parse
   // and reformat it.
-  if (selector.match(/\/\/|\/\*/)) {
+  if (/\/\/|\/\*/.test(selector)) {
+    let value = selector.replace(/^ +/, "").replace(/ +$/, "");
+    // last line has inline comment
+    if (/\/\//.test(selector.split("\n").pop())) {
+      value += "\n";
+    }
```
evilebottnawi: Why? Using `\n` break code for developers who using CRLF

src/language-css/parser-postcss.js
```diff
@@ -390,6 +385,60 @@ function parseNestedCSS(node, options) {
...
+        node.params &&
+        node.params[0] === ":"
+      ) {
+        node.variable = true;
+        node.value = parseValue(node.params.slice(1));
+      }
+
+      // Less variable
+      if (node.variable) {
+        delete node.params;
```
evilebottnawi: ?

src/language-css/parser-postcss.js
```diff
@@ -390,6 +385,60 @@ function parseNestedCSS(node, options) {
...
+
+      // Missing whitespace between variable and colon
+      if (
+        ["page", "nest"].indexOf(node.name) === -1 &&
+        node.params &&
+        node.params[0] === ":"
+      ) {
+        node.variable = true;
+        node.value = parseValue(node.params.slice(1));
+      }
```
evilebottnawi: Why?

src/language-css/parser-postcss.js
```diff
@@ -390,6 +385,60 @@ function parseNestedCSS(node, options) {
       node.value = parseValue(value);
     }
 
+    if (node.type === "css-atrule") {
+      // mixin
+      if (node.mixin) {
```
evilebottnawi: `isLessParser && node.mixin`

src/language-css/parser-postcss.js
```diff
@@ -390,6 +385,60 @@ function parseNestedCSS(node, options) {
...
+      }
+
+      // function
+      if (node.function) {
+        return node;
+      }
+
+      // only css support custom-selector
+      if (node.name === "custom-selector" && options.parser === "css") {
+        const customSelector = node.params.match(/:--\S+?\s+/)[0].trim();
```
evilebottnawi: Don't use regexp here, it is very bad

src/language-css/parser-postcss.js
```diff
@@ -390,6 +385,60 @@ function parseNestedCSS(node, options) {
...
+        node.customSelector = customSelector;
+        node.selector = parseSelector(
+          node.params.slice(customSelector.length).trim()
+        );
+        delete node.params;
+        return node;
+      }
+
+      // Whitespace between variable and colon
+      if (node.name.indexOf(":") !== -1 && !node.params) {
```
evilebottnawi: We should improve this check, Scss allow to use `:` in interpolations and can't have params, so this invalid

evilebottnawi:
Please reduce using regexps, it is always bad, if you don't have enough information on node, please open issue in postcss-less instead using hacks

src/language-css/printer-postcss.js
fisker: will do

src/language-css/parser-postcss.js
fisker: filename is a mess after format, suggestion?

src/language-css/parser-postcss.js
fisker: variable, use `value` property

src/language-css/parser-postcss.js
fisker: this is original impelementation

fisker:
@evilebottnawi 

https://github.com/shellscape/postcss-less#variables

> Note: LESS variables are strictly parsed - a colon (:) must immediately follow a variable name.

`postcss-less` stop parsing 

```less
@foo :bar;
```

what do we do?

src/language-css/parser-postcss.js
evilebottnawi: using delete reduce performance, we should avoid doing this if we really do not need it

src/language-css/parser-postcss.js
evilebottnawi: https://github.com/prettier/prettier/pull/6981#discussion_r347321768

src/language-css/parser-postcss.js
evilebottnawi: hm, okey, let's keep this as is

evilebottnawi:
@fisker if less support these, we should open issue and fix problem in parser

src/language-css/parser-postcss.js
thorn0: Also, in general, we're trying not to modify the AST during the printing phase.

fisker:
@evilebottnawi  https://github.com/shellscape/postcss-less/pull/101#issuecomment-392495714

evilebottnawi:
@fisker we should investigate how Less interpreted this, if parser don't parse that propertly we can't update parser until it will be fixed

lydell:
I'm no Less expert, but it seems to be valid on http://lesscss.org/less-preview/

```less
@foo : bar;

a {
  color: @foo;
}
```

becomes:

```css
a {
  color: bar;
}
```

evilebottnawi:
@fisker it is mean we can't update parser while it is not fix bug

thorn0:
But it works!

**Prettier pr-6981**
[Playground link](https://deploy-preview-6981--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEABAZhCACJ2BGAhgE4DcIANCBAA4wCW0AzsqCcRAO4AKJCLKQgDcI9ACaUQ+YoTABrODADKNWfSgBzZDGIBXOFQAWMALYAbAOqH68JqrBwl-G-SE2AnsnBMWVdUzhiGG4ZDRNCZHRCMwCqACsmAA8AIRl5RSVCEzgAGXU4SOjYkATEpXUNMzgARV0IeEKYgxBVYgDiLyqfSRpidRgLcRhDZAAOAAYqXogAixkaL164dqECqgBHOvgQ2gEQQiYAWig4ODEzyWI4TforkMIwiKQopqoAk3ptPWamCqra+oFZ5FZowQj4QZiYbIABMVB0hHoZgqAGEICZwl4oNA1iBdAEACrggQvYpCfQASSg51gSjAfToAEFqUoYO4qo0AgBfLlAA)
```sh
--parser less
```

**Input:**
```less
@foo : bar;
```

**Output:**
```less
@foo: bar;

```

src/language-css/parser-postcss.js
evilebottnawi: @thorn0 it is okay, but we should reduce number of `delete` where is unnecessary

tests/css_case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -364,13 +364,13 @@ $KeepScssVar: val;
...
 
   .Keep;
   .Keep();
   .Keep(4px) !important;
   .Keep() when (@Keep=Keep);
-  .Keep() when (@Keep=12PX);
+  .Keep() when (@Keep=12px);
```
evilebottnawi: We should keep `12PX`

evilebottnawi:
@fisker let's add more tests for https://github.com/prettier/prettier/pull/6981#issuecomment-555431766 cases

src/language-css/parser-postcss.js
thorn0: @evilebottnawi How many imports do Less files usually include? I don't believe this specific line can have any real performance implications.

evilebottnawi: @thorn0 Depends on the size of the application, this is my recommendation, nothing more

fisker:
@evilebottnawi this is already covered. https://github.com/prettier/prettier/blob/master/tests/css_less/less.less#L1672-L1673

src/language-css/parser-postcss.js
fisker: Can't do that, `--parser=css` also need this, since we use less parser to parse css

src/language-css/printer-postcss.js
evilebottnawi: Please add `isLessParser` check here

src/language-css/printer-postcss.js
```diff
@@ -115,12 +115,20 @@ function genericPrint(path, options, print) {
...
-          return concat(["// ", rawText]);
+        if (node.raws.inline || node.inline) {
+          const needBreakAfter = !(
+            node.source.input.css.split(/[\r\n]/)[node.source.end.line] || ""
+          ).trim();
+          return concat([
+            "//",
+            node.raws.left + rawText,
+            needBreakAfter ? line : ""
+          ]);
```
evilebottnawi: I think we can simplify that, it was workaround for less parsers bugs, so maybe they were fixed

src/language-css/parser-postcss.js
```diff
@@ -382,7 +382,7 @@ function parseNestedCSS(node, options) {
 
     if (node.type === "css-atrule") {
       // mixin
-      if (node.mixin) {
+      if (options.parser !== "scss" && node.mixin) {
```
evilebottnawi: The `css` parser should not support `minix` and other stuff because mixins doesn't exists in `css`, yes we use `postcss-less` for parsing pure CSS, but it is not good and in future we should switch from `postcss-less` on built-in css parser inside `postcss`

src/language-css/printer-postcss.js
fisker: In fact we can't use `options.parser === "less"`, since we has `parser=css` also use `less` parser, so I use `options.parser !== "scss"` in this commit

https://github.com/prettier/prettier/pull/6981/commits/07010de356026f1568864bfceb14c271a39e8088

What do you think? or maybe `options.parser === "less" || options.parser === "csss"`?

evilebottnawi: To be honestly we should throw error on less syntax when developer set `css` parser. What about change the `options.parser` value change when we switch on other syntax?

evilebottnawi:
Good job, some small fixes and we can merge it, i will look at all deeply tomorrow

src/language-css/printer-postcss.js
fisker: Maybe we can do this later in a seperate PR, shall we?

src/language-css/printer-postcss.js
fisker: Will check this

src/language-css/printer-postcss.js
evilebottnawi: Yes, don't forget open new issue after merge, anyway let's encapsulate logic in one util function `isLessParser`, it is allow to fix problem in one places

src/language-css/printer-postcss.js
fisker: Seems hard to fix, maybe fix this later

tests/css_case/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Can we add test for `.Keep() when (@Keep="12px");`?

tests/css_less/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,52 @@
...
+  .sel-@{value} {
+    a: b;
+  }
+});
+
+// TODO: fix this
+each(@selectors, 
+{
+  .sel-@{value} {
+a: b;
```
evilebottnawi: Please open issue about that after merge

src/language-css/parser-postcss.js
```diff
@@ -390,6 +380,64 @@ function parseNestedCSS(node, options) {
       node.value = parseValue(value);
     }
 
+    if (node.type === "css-atrule") {
+      const isLessParser =
+        options.parser === "css" || options.parser === "less";
```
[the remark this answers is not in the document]
fisker: https://github.com/prettier/prettier/pull/6981/commits/305b20cc1169c57a8d4ceedf46129ebecf5c90e7

src/language-css/printer-postcss.js
```diff
@@ -189,6 +202,52 @@ function genericPrint(path, options, print) {
     }
     case "css-atrule": {
       const parentNode = path.getParentNode();
+      const isTemplatePlaceholderNodeWithoutSemiColon =
+        isTemplatePlaceholderNode(node) &&
+        !parentNode.raws.semicolon &&
+        options.originalText[options.locEnd(node) - 1] !== ";";
+      const isLessParser =
+        options.parser === "css" || options.parser === "less";
```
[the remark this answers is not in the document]
fisker: https://github.com/prettier/prettier/pull/6981/commits/305b20cc1169c57a8d4ceedf46129ebecf5c90e7

src/language-css/printer-postcss.js
```diff
@@ -925,6 +982,10 @@ function adjustStrings(value, options) {
   return value.replace(STRING_REGEX, match => printString(match, options));
 }
 
+function lastLineHasInlineComment(value) {
+  return /\/\//.test(value.split(/[\r\n]/).pop());
+}
```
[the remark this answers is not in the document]
fisker: https://github.com/prettier/prettier/pull/6981/commits/2639c6cfa8f44f4c014f6951fe7fb148c212c98d

fisker:
> Can we add test for .Keep() when (@Keep="12px");?

https://github.com/prettier/prettier/pull/6981/commits/dba16d0bff0954c5decca549ef0b5377870b4e31 , this is expected, right?

fisker:
> we should keep EXTEND as is

add test noth `&:extend(.Keep ALL);` and `&:EXTEND(.Keep ALL);`

fisker:
Forgot this?

fisker:
I didn't notice there is conflicts, I'm not near my laptop, can you do that for me?

thorn0:
I'm going to look into this PR and resolve the conflicts.

tests/css_less/__snapshots__/jsfmt.spec.js.snap
thorn0: I think it's https://github.com/prettier/prettier/issues/4524, not a new issue.

PR #6982, opened 2019-11-17
Enable `replace-array-includes-with-indexof` for all bundles

fisker:
It's really annoying, and hard to locate code, let's enable it for all bundles.

maybe we can also merge #6967 to `master` instead of `next`


lydell:
Is this PR fixing any issue? If not, this will just cause merge conflicts with the `next` branch which doesn’t need this transform at all, right?

fisker:
https://github.com/prettier/prettier/pull/6981 need transform too, maybe other incoming dependencies, no harm to enable it. conflict will be only one line, I think

PR #6993, opened 2019-11-18
Remove deprecated parser & update `sharedUtil`

fisker:


src/common/internal-plugins.js
```diff
@@ -131,11 +124,6 @@ module.exports = [
         return eval("require")("../language-markdown/parser-markdown").parsers
           .remark;
       },
-      // TODO: Delete this in 2.0
-      get markdown() {
```
fisker: can't find history, why this removed?

thorn0: I guess the original idea was to call the parser `remark` instead of `markdown`, like we have `babel`, not `javascript`. Not sure we still want to go that way.

fisker: I guess probably we remove this after we support `language` option

src/common/internal-plugins.js
```diff
@@ -131,7 +124,6 @@ module.exports = [
         return eval("require")("../language-markdown/parser-markdown").parsers
           .remark;
       },
-      // TODO: Delete this in 2.0
```
fisker: https://github.com/prettier/prettier/issues/4097#issuecomment-579482013

tests_integration/__tests__/util-shared.js
```diff
@@ -24,8 +24,9 @@ test("shared util has correct structure", () => {
   expect(typeof sharedUtil.getNextNonSpaceNonCommentCharacterIndex).toEqual(
     "function"
   );
-  expect(typeof sharedUtil.mapDoc).toEqual("function");
   expect(typeof sharedUtil.makeString).toEqual("function");
+  // Removed in v2.0.0
+  expect(typeof sharedUtil.mapDoc).toEqual("undefined");
```
evilebottnawi: I think we don't need that test

fisker: removed.

src/language-markdown/parser-markdown.js
```diff
@@ -109,8 +109,6 @@ const mdxParser = { ...baseParser, parse: createParse({ isMDX: true }) };
 module.exports = {
   parsers: {
     remark: markdownParser,
-    // TODO: Delete this in 2.0
-    markdown: markdownParser,
```
thorn0: https://github.com/prettier/prettier/issues/4097#issuecomment-579482013

PR #7005, opened 2019-11-19
test: css-in-js

### LINKED ISSUE #2291: Removes linebreaks from styled-components tagged template literals in an odd way

### LINKED ISSUE #2601: CSS-in-JS ampersand unexpected behavior
Hey,

we ran into this slightly awkward behaviour / edge case with ampersands in `CSS-in-JS` scenarios. Prettier transforms code that looks like this:

``` js
const Foo = styled.p`
  max-width: 980px;
  ${mediaBreakpointOnlyXs`
    && {
      font-size: 0.8rem;
    }
  `}

  &.bottom {
    margin-top: 3rem;
  }
`;
```

to this:

``` js
const Foo = styled.p`
  max-width: 980px;
  ${mediaBreakpointOnlyXs`
    && {
      font-size: 0.8rem;
    }
  `} &.bottom {
    margin-top: 3rem;
  }
`;
```

Expected would be something along the lines of:

``` js
const Foo = styled.p`
  max-width: 980px;
  ${mediaBreakpointOnlyXs`
    && {
      font-size: 0.8rem;
    }
  `};

  &.bottom {
    margin-top: 3rem;
  }
`;
```

See the positioning of `&.bottom` after the prettifying. Like said it might be an edge case that only happens with interpolated things that don't end with ; and are followed with &, but will check if this occurs with other similar cases (and in LESS/SASS situations) if needed.

Thanks for prettier, it saves me hours every week ❤️ ❤️

### LINKED ISSUE #2623: [styled-components] Automatically add semicolon for template string interpolation
Maybe a duplicate of https://github.com/prettier/prettier/issues/2402, but can we add a semi colon if the interpolation is in the middle of rule list? so the next rule not dangling at the end of the interpolation

[Demo](https://prettier.io/playground/#%7B%22content%22%3A%22const%20A%20%3D%20styled.a%60%5Cn%20%20display%3A%20inline-block%3B%5Cn%20%20color%3A%20%23fff%3B%20%20%5Cn%20%20%24%7Bprops%20%3D%3E%20props.a%20%26%26css%60%5Cn%20%20%20%20display%3A%20none%3B%5Cn%20%20%60%7D%5Cn%20%20%20height%3A%2030px%3B%5Cn%60%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Afalse%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D)

evilebottnawi:

Just tests

fixes #2601 fixes #2291 fixes #2623

### DIFF
--- a/tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/multiparser_js_css/__snapshots__/jsfmt.spec.js.snap
@@ -165,6 +165,96 @@ const StyledComponent2 = styled.div\`
   }
 \`;
 
+const Direction = styled.span\`
+  \${({ up }) => up && \`color: \${color.positive};\`}
+  \${({ down }) => down && \`color: \${color.negative};\`}
+\`;
+
+const Direction2 = styled.span\`
+  \${({ up }) => up && \`color: \${color.positive}\`};
+  \${({ down }) => down && \`color: \${color.negative}\`};
+\`;
+
+const mixin = css\`
+  color: \${props => props.color};
+  \${props => props.otherProperty}: \${props => props.otherValue};
+\`;
+
+const foo = styled.div\`
+  display: flex;
+  \${props => props.useMixin && mixin}
+\`;
+
+const Single1 = styled.div\`
+  color: red
+\`;
+
+const Single2 = styled.div\`
+  color: red;
+\`;
+
+const Dropdown = styled.div\`
+  /* A comment to avoid the prettier issue: https://github.com/prettier/prettier/issues/2291 */
+  position: relative;
+\`;
+
+const bar = styled.div\`
+  border-radius: 50%;
+  border: 5px solid rgba(var(--green-rgb), 0);
+  display: inline-block;
+  height: 40px;
+  width: 40px;
+
+  \${props =>
+    (props.complete || props.inProgress) &&
+    css\`
+      border-color: rgba(var(--green-rgb), 0.15);
+    \`}
+  
+  div {
+    background-color: var(--purpleTT);
+    border-radius: 50%;
+    border: 4px solid rgba(var(--purple-rgb), 0.2);
+    color: var(--purpleTT);
+    display: inline-flex;
+
+    \${props =>
+    props.complete &&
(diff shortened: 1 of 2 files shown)


thorn0:
Snapshots will be easier to read and understand if you split the tests into multiple files with descriptive names. Not a requirement, just a recommendation.

evilebottnawi:
@thorn0 i think we can avoid this, because these are common usage practices

thorn0:
E.g. to review this PR, the reviewer has to scroll constantly back and forth to compare input and output. That's not how it is supposed to be.

evilebottnawi:
@thorn0 okay, what are the ideas for naming? As I said earlier, the usual standard use cases, nothing special

thorn0:
Each test illustrates something: a syntax combination, a use case, a problem, etc. Any name will do. Even something like `issue-2601.js` is way better than everything in one big file. But I'm fine if you leave this specific PR as is, it's just something to keep in mind in the future. (Sorry za zanudstvo.)

evilebottnawi:
@thorn0 
> Even something like issue-2601.js is way better

yes, let's do it in future :+1: 

> But I'm fine if you leave this specific PR as is

Great, i think we can merge this :smile: 

> Sorry za zanudstvo

nichego strahnogo, vse horosho :smile:

Issue #3905, opened 2018-02-06
[css] comments around declaration


evilebottnawi:
**Prettier 1.10.2**
[Playground link](https://deploy-preview-3897--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBDABMAOldf0D0AVOpALZkIzoCM6RBO+hJ5ls6ATPQegA6oAJoICWUAOYtSEClXQBmHuiRS2cgCxKaABj4APVTPbUArFt0Hi02RwBsPANyGb1AOw8m+K2o4AOD7j4MABOqFAAzgBmEMFkKiFh4QA2qPAAFN5GcgCc5vrOxrTaPACUDp54kEkxKpkutHQM6MFwggVyNNxNAIQiZHwxMGHUdYU0ik1Oox2aDBXtHDRmc1AAviAANCAQfDAi0OHIoKjBwRAA7gAKJwiHKKgAbhAigpto4TDIkahJ4XBbACNQmAANZwGAAZQEYDE4mQIQArv8QIIIGAvj8-lsxH9gjBLqFxGRUBjfsiAFbhPQAIWBYMhqEoABkxHBSVjtgiYHwuZx2ciBMFcchwOFDls+MExDAAOovGAAC2Qvm0ErOfxloT4IslcFxDzZWxaAEcESIWgTUESSUhvmStn8yCJ4cEkQ7YUk4ABFBEQeD8rZDAFywSK5CcQOhERJWEAYSMNpAUGghpACL+ABVUAC7na-qtVkA)

**Input:**
```css
a {
    /* comment 1 */
    /* comment 2 */ padding /* comment 3 */ : /* comment 4 */ 10px /* comment 5 */ 10px /* comment 6 */; /* comment 7 */
    /* comment 8 */
    transform: translate(/* comment 9 */ 10px /* comment 10 */);
    color: /* comment 11 */ red /* comment 12 */ !important /* comment 13 */ ; /* comment 14 */
    /* comment 15 */
}
```

**Output:**
```css
a {
  /* comment 1 */
  /* comment 2 */
  padding/* comment 3 */ : /* comment 4 */ 10px /* comment 5 */ 10px
    /* comment 6 */; /* comment 7 */
  /* comment 8 */
  transform: translate(/* comment 9 */ 10px /* comment 10 */);
  color: /* comment 11 */ red /* comment 12 */ !important /* comment 13 */ ; /* comment 14 */
  /* comment 15 */
}
```

[...]

PR #7009, opened 2019-11-19
fix: comment at end of at-rule with selector

### LINKED ISSUE #5941: Sass directives without blocks can produce invalid code
**Prettier 1.16.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAhjAtAJwhGAAgDoAzfAHSgHpqSAjdHK4KwwqgXxABoQIADjACW0AM7JQTPAHcACkwQSU6AG4RhAE14h6OdGADWcGAGUBB4VADmyGDgCucPgAsYAWwA2AdRfD4YhZgcKZK-sKq-gCeyOBiEnxWYnA4MHL61u7oyKTonsl8AFZiAB4AQvpGJqbo7nAAMlZwOXkFIMUlplbWnnAAig4EzUi5+c4gFjjJOLFiYPE6AjhWMN5aMC7IABwADHxLEMne+gKxS3DTqs18OHAAjg7Ct+nomdkjrePJ7sJ2jl-dXoDIYtMZ8GDoehrTQbZAAJnB+mEnm6AGEIO4srEoNBriAHMkACqQ5SjZKcThAA)
```sh
--parser scss
```

**Input:**
```scss
@at-root .foo
// .bar
{
  
}
```

**Output:**
```scss
@at-root .foo
// .bar {
}

```

**Expected behavior:**
Prettier should generate valid Sass code. The output should match the correct formatting that is produced without the `@at-root` directive ([playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA6AZhCAdKB6PAAlQCMBDAJ12FwF8QAaECABxgEtoBnZUSiiAHcACpQQ8UZAG4R2AE0YgSFMmADWcGAGUWq9lADmyGBQCucJgAsYAWwA2AdUvt4XXWDhbxL9lJcBPZHAuHiZ9LjgKGGEVAxsyZHQyOwimACsuAA8AIRV1TS0yGzgAGX04ROTUkAzMrX0DOzgARVMIeEqUixBdCgiKIK4wEMUWCn0YB3kYS2QADgAGJjGICIcVFiCxuH6pCqYKOABHU3ZDmLI4hKQkrqYIm3ZjM26uBqbW9oqbqu6YMhIUzkM2QACYmCYyOw7A0AMIQGzxIJQaD7ECmCIAFQBEluEVotCAA)):
```scss
.foo
// .bar
 {
}
```

The behavior seems to generally be caused by `@` directives with no associated `{}`-blocks (like `@media` or `@supports` are required to have).

In the example above, any arbitrary `@rule` in place of `@at-root` can cause the incorrect output, for example:

```scss
@qux .foo
// .bar {
}

```

evilebottnawi:
Fixes #5941

~partial #5941, still ugly, but don't break code~


thorn0:
Do you think this isn't worth mentioning in the changelog?

evilebottnawi:
@thorn0 forgot to commit :smile:

evilebottnawi:
@thorn0 Done

fisker:
Need rebase, and there is new `lastLineHasInlineComment`, we can use.

evilebottnawi:
@fisker feel free to send a new PR, don't have time on this right now :disappointed:

fisker:
I fixed the conflicts, let's see what will it be.

fisker:
I changed a little bit, 

```css
@at-root .foo
// .bar
   {
}
```

to

```css
@at-root .foo
// .bar
{
}
```

I think this is better.

evilebottnawi:
@fisker agree

fisker:
Failed cases in #5941,  all fixed.

evilebottnawi:
@fisker :+1: I can't review, because it is my PR :smile:

thorn0:
@evilebottnawi I'm not a member myself :)

evilebottnawi:
@thorn0 I can send a invite :smile:

thorn0:
@evilebottnawi  Okay, now that I'm a member, I'm good with adding Fisker to members. :D

thorn0:
BTW, I'm working on something like https://github.com/prettier/prettier/issues/3905#issuecomment-393621184 that will fix this too. But have I little idea when I finish that.

evilebottnawi:
@thorn0 It is why we should use pure CSS parser for CSS, because pure CSS and SCSS/Less have difference grammar, in an ideal world we should get `SyntaxError` in your case

fisker:
That one should be a parser error.

Issue #7019, opened 2019-11-20
format `@use` and `@forward` scss at-rules


evilebottnawi:
**Prettier 1.19.1**
[...]

PR #7057, opened 2019-11-26
Add support for @use and @forward rules in Scss

jbravar:
Add detection for new `@use` and `@forward` rules included in Sass 1.23. Addresses #7019.

### DIFF
--- a/tests/css_scss/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css_scss/__snapshots__/jsfmt.spec.js.snap
@@ -1085,6 +1085,42 @@ label {
   @include margin-bottom-1\\/3;
 }
 
+@forward "src/list";
+@forward "src/list" as list-*;
+@forward "src/list"    as    list-*   ;
+@forward "src/list"    hide     list-reset   ,     $horizontal-list-gap   ;
+@forward "src/list"    hide     list-reset   ,     $horizontal-list-gap   , $horizontal-list-gap-foo, $horizontal-list-gap-bar;
+
+
+@use 'src/code';
+@use "src/code" as c;
+@use "src/code"    as    c;
+@use "src/code"    as    *;
+@use 'library'    with (
+$black: #222,
+  $border-radius: 0.1rem
+
+);
+@use 'library'    with (
+$black: #222,
+$border-radius: 0.1rem,
+$left: 10px,
+$right: 400px
+);
+@use 'library'    with (
+$black: #222,
+$border-radius: 0.1rem,
+$left: 10px,
+$right: 400px,
+$box-shadow: 0 0.5rem 1rem rgba($black, 0.15)
+);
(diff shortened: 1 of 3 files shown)


tests/css_scss/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2048,6 +2084,41 @@ label {
   @include margin-bottom-1\\/3;
 }
 
+@forward "src/list";
+@forward "src/list" as list- *;
+@forward "src/list" as list- *;
```
evilebottnawi: Bug, should be `list-*`

tests/css_scss/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2048,6 +2084,41 @@ label {
...
+  (
+    $black: #222,
+    $border-radius: 0.1rem,
+    $left: 10px,
+    $right: 400px,
+    $box-shadow: 0 0.5rem 1rem rgba($black, 0.15),
+    0 0.5rem 1rem rgba($black, 0.15),
+    0 0.5rem 1rem rgba($black, 0.15),
+    0 0.5rem 1rem rgba($black, 0.15)
+  );
```
evilebottnawi: Ugly, we should format that:
```scss
@use 'veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVery' with (
  $black: #222,
  $border-radius: 0.1rem
);
```

alexander-akait (changes requested):
Good job, need some improvements

tests/css_scss/__snapshots__/jsfmt.spec.js.snap
AndrewLeedham: @evilebottnawi any ideas on what type the AST node should be. I tried converting the `with` to a function with a nested `paren_group` but that formats it as
```scss
@use 'veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVery' with(
  $black: #222,
  $border-radius: 0.1rem
);
```

evilebottnawi: No ideas :disappointed: just point on bad output

michaelfaith:
Did this PR get abandoned?

AndrewLeedham:
@michaelbeaumont I believe so. I took a shot at addressing the PR feedback in https://github.com/prettier/prettier/pull/9069. But hit a roadblock due to a lack of understanding: https://github.com/prettier/prettier/pull/9069#discussion_r477314002

michaelfaith:
@AndrewLeedham dang. Looks like you were close too.  At least patching it til a new parser can be implemented.  
@alexander-akait is that something anyone's working on?  With sass moving away from `@import` entirely, more and more people will be moving to their module syntax and need this capability. We're having to completely ignore some of our important sass files, while this issue remains.

PR #7061, opened 2019-11-27
Update FUNDING.yml

azz:
Do any other maintainers have GitHub Sponsors accounts? 

@duailibe @evilebottnawi noticed you removed yourselves in https://github.com/prettier/prettier/commit/5a1aeb3aa27ae3a4ddaa57ab42215a6506b9844d. Have you applied for Sponsors?

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,3 @@
 tidelift: npm/prettier
 open_collective: prettier
+github: [azz]


evilebottnawi:
@azz i think better collect funds in one collective instead to each developer and then distribute them, otherwise it may not be completely honest

azz:
That's a good point! Could I get added to the OpenCollective?

PR #7074, opened 2019-12-01
Add docs on using prettier with git-format-staged in a pre-commit hook

hallettj:
I am the author of [git-format-staged](https://github.com/hallettj/git-format-staged) which provides another option for running prettier in a pre-commit hook. I believe that git-format-staged is valuable because it operates fundamentally differently compared to other tools that format partially-staged files, and therefore provides a different set of guarantees. This PR documents usage of git-format-staged with prettier in a pre-commit hook, and provides a brief description of its merits.

Should I add an entry to `changelog_unreleased/` for a documentation change?

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -367,6 +367,8 @@
         "unpause",
         "unpkg",
         "unrestrict",
+        "unstage",
+        "unstaged",
         "untracked",
         "valourous",
         "Vanderwerff",
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -84,7 +84,40 @@ and add this config to your `package.json`:
 
 Read more about this tool [here](https://github.com/JamesHenry/precise-commits#2-precommit-hook).
 
-## Option 5. bash script
+## Option 5. [git-format-staged](https://github.com/hallettj/git-format-staged)
+
+**Use Case:** Great for when you want to format partially-staged files, and you have run into issues with other options.
+
+Git-format-staged is used to run any formatter that can accept file content via stdin. It operates differently than other tools that format partially-staged files: it applies the formatter directly to objects in the git object database, and merges changes back to the working tree. This procedure provides several guarantees:
+
+1. Changes in commits are always formatted.
+2. Unstaged changes are never, under any circumstances staged during the formatting process.
+3. If there are conflicts between formatted, staged changes and unstaged changes then your working tree files are left untouched - your work won't be overwritten, and there are no stashes to clean up.
+4. Unstaged changes are not formatted.
+
+Use it with [husky](https://github.com/typicode/husky):
+
+```bash
+yarn add --dev husky prettier git-format-staged
+```
+
+and add this config to your `package.json`:
+
+```json
+{
+  "husky": {
+    "hooks": {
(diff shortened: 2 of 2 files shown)


lydell:
Hi!

Have you seen the [lint-staged](https://github.com/okonet/lint-staged#-help-test-lint-stagedbeta) beta? Is your solution different?

hallettj:
> Have you seen the [lint-staged](https://github.com/okonet/lint-staged#-help-test-lint-stagedbeta) beta? Is your solution different?

Hello! Yes, git-format-staged is different. I looked at the [beta branch](https://github.com/okonet/lint-staged/tree/beta) of lint-staged. AFAICT it behaves similarly to my understanding of how the stable version works:

1. stash unstaged changes (actually it stashes everything, and computes a patch to bring back staged changes only)
1. apply prettier (or other linters or formatters) to files with staged changes
1. stage changes from formatting
1. restore unstaged changes - effectively this is done by computing a patch of differences between unformatted staged changes plus unstaged changes, and that patch is applied to files with staged changes after they have been formatted

Git-format-staged operates on the git object database. When a file is staged a blob is created in the object database that is a copy of the file with staged changes, and without unstaged changes. Git-format-staged operates like this:

1. apply prettier (or other linters or formatters) to blobs representing staged versions of files - this produces a new blob in the object database for each file with formatting changes
1. update the index to point to the new blobs
1. compute a patch for each staged file with the changes made by the formatter
1. apply patches to working tree files

The result is similar, but there are two key differences:

1. Git-format-staged leaves working tree files alone until the last step. If formatting changes cannot be cleanly merged with unstaged changes the fallback behavior leaves working tree files untouched. The commit is not blocked, and committed files are formatted properly. Lint-staged puts working tree modifications in the stash up front. If I understand what I'm reading from the source correctly if lint-staged cannot merge formatting changes with unstaged changes it resets everything, staged and unstaged changes, back to their initial states.
1. Both tools merge formatting changes with unstaged changes. Git-format-staged computes patches of formatting changes, and applies them to working tree files. Lint-staged computes a patch of unstaged changes, and applies it to formatted files with staged changes. Neither approach is obviously better; but there might be cases where one works and the other doesn't.

My understanding of lint-staged is not especially deep, so my apologies if there is something I have misunderstood.

lydell:
Thanks for the detailed answer! Though I have to admit, even though I find git very interesting, I had a hard time understanding all the details :)

And that’s what I’m worried about. lint-staged is a fairly popular project. How are we going to explain the differences to users?

I’m also a little bit worried about the tool being written in Python. I like Python myself, but unlike Linux and mac it does not come pre-installed on Windows, right?

hallettj:
I think I can boil down what distinguishes git-format-staged from lint-staged to,
- failure to merge formatting changes with unstaged changes does not block the commit
- changes are never stashed
- it works differently, and might provide different results

Please correct me if lint-staged also does not block the commit in the same scenario.

I realize that the way I wrote up git-format-staged's guarantees might imply that other tools do not share several of the same guarantees, which is obviously not true. I can adjust the wording to make it clear that those guarantees overlap with lint-stage's, or to focus on what makes git-format-staged different instead.

You make a good point about Python - I forgot to call that out. I've updated the PR with a note that Python is required. I chose Python because I want to be able to use the tool with projects in languages other than JavaScript, such as Rust, and Python is the most widely-deployed interpreter that I could think of that I was willing to work with.

PR #7075, opened 2019-12-01
Don't print trailing commas after rest elements in tuples

thorn0:
* A rest element is always the last element of a tuple type. Nothing can be added after it.
* While TS is okay with this comma, Babel doesn't parse it. https://github.com/babel/babel/issues/10769
* In function parameters and array destructuring, such a comma is a syntax error. Keeping it in tuples is inconsistent.


evilebottnawi:
Good job, thanks!

PR #7129, opened 2019-12-11
fix: configure pre-commit hook to only output changed files

jonathanmorley:
Currently, the pre-commit hook will output all files that prettier analyzes.

The changed lines are highlighted in the terminal, but this highlighting is not always successful (e.g. we have seen the highlighting disappear when running inside Jenkins)

This sets prettier (when running within pre-commit) to only output changed files

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,6 @@
 - id: prettier
   name: prettier
-  entry: prettier --write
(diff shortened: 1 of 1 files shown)


jonathanmorley:
To test these changes, I added `args: ['--list-different']` to the prettier hook in one of my `.pre-commit-config.yaml`s.

This successfully outputted only the changed files

PR #7204, opened 2019-12-29
javascript: fix Template strings aren't moved to new line, where '' or "" strings are

### LINKED ISSUE #4719: Template strings aren't moved to new line, where '' or "" strings are
**Prettier 1.13.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBDTBeAHSk0wAMA3OAJwE8AbaAcwyoEspHLaGOX3Pq9Jnw5chvGG1GCezSf1JEikKBkwAjfJgIgxskQO7D50oxKmHxci3uMWdIADQgIABxit0yUNipUIAO4ACr4IaMgg2OQQrAAmTiDqVNhgANZwMADKrin8yJIArnDO7GjUMEHJjAC22MgAZth0Zc4AVmgAHgBCyWkZmdjVcAAy7HANTS0gOVRlVBHq2OriCa5SMADqcTAAFsgAHAAMzmsQZRvJrhFrcHOUCVRwAI4FrI+V2DV1SI3NxSBlaqsfJUIrOND8OhwACKBQg8Amf2cMCWW1iu2QACZkclWHR+ABhCDVWoRKDQcbOAplAAqS3CP0mcAAvsygA)

**Input:**
```jsx
const a =
  `verylongstringverylongstringverylongstringverylongstringverylongstring`

const b = "verylongstringverylongstringverylongstringverylongstringverylongstring"
```

**Output:**
```jsx
const a = `verylongstringverylongstringverylongstringverylongstringverylongstring`;

const b =
  "verylongstringverylongstringverylongstringverylongstringverylongstring";
```

**Expected behavior:** The template string is moved to the next line after the `const a =`.

I'd assume that this happens because it assumes about indentation, as that is part of the
string in multiline template strings.

sasurau4:
fix #4719

### DIFF
--- a/tests/angular_component_examples/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/angular_component_examples/__snapshots__/jsfmt.spec.js.snap
@@ -27,11 +27,12 @@ class     TestComponent {}
 =====================================output=====================================
 @Component({
   selector: "app-test",
-  template: \`
-    <ul>
-      <li>test</li>
-    </ul>
-  \`,
+  template:
+    \`
+      <ul>
+        <li>test</li>
+      </ul>
+    \`,
   styles: [
     \`
       :host {
@@ -76,11 +77,12 @@ class     TestComponent {}
 =====================================output=====================================
 @Component({
   selector: "app-test",
-  template: \`
-    <ul>
-      <li>test</li>
-    </ul>
-  \`,
+  template:
+    \`
+      <ul>
+        <li>test</li>
+      </ul>
+    \`,
   styles: [
     \`
       :host {
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -5740,7 +5740,8 @@ function printAssignmentRight(leftNode, rightNode, printedRight, options) {
       // do not put values on a separate line from the key in json
       options.parser !== "json" &&
       options.parser !== "json5") ||
-    rightNode.type === "SequenceExpression";
+    rightNode.type === "SequenceExpression" ||
+    (leftNode.type === "Identifier" && rightNode.type === "TemplateLiteral");
 
   if (canBreak) {
     return group(indent(concat([line, printedRight])));
--- a/tests/decorators-ts/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/decorators-ts/__snapshots__/jsfmt.spec.js.snap
@@ -66,9 +66,10 @@ export class HeroButtonComponent {
 =====================================output=====================================
 @Component({
   selector: "toh-hero-button",
-  template: \`
-    <button>{{ label }}</button>
-  \`
+  template:
+    \`
+      <button>{{ label }}</button>
+    \`
 })
 export class HeroButtonComponent {
   @Output() change = new EventEmitter<any>();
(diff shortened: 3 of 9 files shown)


thorn0:
I think this has been discussed somewhere else, not only in #4719, but so far I can't find that discussion. The current behavior might be intended.

bakkot:
This seems relevant to #2482.

tests/angular_component_examples/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -76,11 +77,12 @@ class     TestComponent {}
 =====================================output=====================================
 @Component({
   selector: "app-test",
-  template: \`
-    <ul>
-      <li>test</li>
-    </ul>
-  \`,
+  template:
+    \`
```
thorn0: Changes like this definitely don't look good. I'd say the current behavior should be kept for multiline template literals.

lipis:
I doubt that it's intended, looks more like a bug to me.. I don't see why it shouldn't change for `

[Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBDTBeTAHRG1LPIsquptrvOIG5CpIoMd9MByev-gYO7NW6LLgIADQTNlzJjEABoQEAA4wAlumShsAJ30QA7gAUDCNMhIA3CJoAmykACN92MAGs4MAMpqPTSgAc2QYfQBXOBUACxgAWwAbAHUYzXg0ALA4X0t0zRt0gE9rMDQrFSC0OH0YU3dg+OxkADNsROqVACs0AA8AIXcvH19seLgAGSC4VvbOkB7e3yDgxLgARQiIeFmO6JAA-Wr9axdsFzhE5zV9IJhkxxgY5AAOAAYVG4hq5Pc1axucGONhmKgAjlt4PV1FYUNg0ABaKBwOAOVHOfRwCGaTH1bCNZpINp7FTVeKaMKRfZoFZrTbbGZEub7GDnB4OJ7IABMKnC2E0iRWAGEIPEmtYoNBQSAItUACrnWHE+Y2KIASSgaNgvjAtw0AEFNb4YEU1rtqgBfC1AA)

thorn0:
@lipis In single-line cases, yes, probably, but look at the changed snapshots. Do they look good to you?

G-Rath:
@thorn0 This might already have been cleared up in one of the linked issues, but just to throw my hat into the ring: 

> I think this has been discussed somewhere else
> The current behavior might be intended.

From my understanding the current behaviour is indeed completely intended; from what I understand (and understood at the time when I read the discussion I too can't find) it's there as an escape hatch for formatting strings, that tied in to how prettier would behave when trying to pretty template literals that had expressions.

The behaviour was effectively that prettier bailed out when it encountered a template string literal, as it was line-breaking as soon as it hit an expression. The behaviour was changed to bail out of template literals, and only format template expressions if they had expressions in them (i.e `${target}` will never be broken on, but `${target ? target.toUpperCase() : 'NO TARGET'}` would).

Personally, while I am a fan of prettier & consistency, I've always liked having this as an escape hatch, and so for what it's worth would vote against changing this.

Issue #7225, opened 2020-01-05
typescript: TSIndexSignature without type parameters


armano2:
This is similar issue as #7174 but for TSIndexSignature

**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#.....)
```sh
--parser typescript
```

**Input:**
```ts
type foo = { [a: string] }
```

**Output:**
```ts
type foo = { [a: string]:  };
```

2nd run

```
SyntaxError: Type expected. (1:28)
> 1 | type foo = { [a: string]:  };
    |                            ^
  2 | 
```

**Expected behavior:**

```ts
type foo = { [a: string]; }
```

#### Additional info

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/typescript-estree/src/ts-estree/ts-estree.ts#L1336-L1344
```ts
export interface TSIndexSignature extends BaseNode {
  type: AST_NODE_TYPES.TSIndexSignature;
  parameters: Parameter[];
  typeAnnotation?: TSTypeAnnotation;
  readonly?: boolean;
  accessibility?: Accessibility;
  export?: boolean;
  static?: boolean;
}
```

Issue #7224, opened 2020-01-05
typescript: TSIndexSignature with multiple parameters 


armano2:
`TSIndexSignature` allows to output multiple parameters, this syntax is no longer valid, but typescript produces ast for it

**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAZhCbAXm2AB0ptsBtAQyWwGcYAnASygHMAabAIwbN2XALqDWHTgG4KAXxDcQEDDDbRGyULRYsIAdwAK2hBpS0AbhDYATBSD4taYANZwYAZQxPJyVgFc4RQALGABbABsAdSC2eEYvMDh3E1i2c1i0ZHBGDUUORjgWGANHTlDaZFxacILFACtGAA8AIUcXN3daULgAGQ44SurakAbG90lwuABFPwh4QZrAkC8WApYs9CxGMHYVOwxhGEibGCDkAA4ABkUDiALIxwwsg7g18wHFAEdZ+BLlUxAtEYAFooHA4NYIXYWHBvmwYSVaGUKkgqotFAVQmxfCwAhiJtMfgNUUMljBaHxjtZTsgAEyKVi0NjhSQAYQgoXKWSg0A+ID8BQAKhTTGjhuYAgBJKCQ2DuHZsFQAQRl7nQkwWBVksiAA)

```sh
--parser typescript
```

**Input:**
[source](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/shared-fixtures/fixtures/typescript/errorRecovery/index-signature-parameters.src.ts)
```ts
type foo = {
  [a: string, b: string]: string;
}
```

**Output:**
```jsx
type foo = {
  [a: stringb: string]: string;
};
```

**Expected behavior:**
```ts
type foo = {
  [a: string, b: string]: string;
}
```

#### Additional info

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/typescript-estree/src/ts-estree/ts-estree.ts#L1336-L1344
```ts
export interface TSIndexSignature extends BaseNode {
  type: AST_NODE_TYPES.TSIndexSignature;
  parameters: Parameter[];
  typeAnnotation?: TSTypeAnnotation;
  readonly?: boolean;
  accessibility?: Accessibility;
  export?: boolean;
  static?: boolean;
}
```

armano2:
fixed by #7228

PR #7228, opened 2020-01-06
Fix printing both index signatures without type annotations and multi-parameter index signatures

cola119:
fixes https://github.com/prettier/prettier/issues/7225


cola119:
I guess I can fix https://github.com/prettier/prettier/issues/7224 quickly, but should I commit and push to this branch to avoid conflicts? Or is it better to create a new PR for that?

changelog_unreleased/typescript/pr-7228.md
```diff
@@ -0,0 +1,11 @@
+#### Typescript: Remove unnecessary colons for TSIndexSignature without type parameters ([#7228](https://github.com/prettier/prettier/pull/7228) by [@cola119](https://github.com/cola119))
```
thorn0:
```suggestion
#### Fix printing index signatures without type annotations ([#7228](https://github.com/prettier/prettier/pull/7228) by [@cola119](https://github.com/cola119))
```

tests/typescript_index_signature/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -11,11 +11,14 @@ class LocalStorage implements Storage {
   [key: string]: any;
 }
 
+type A = { [key: string] };
```
thorn0: As an index signature without a type annotation is not valid TS, please move the test for this case to a separate file in `tests/typescript_error_recovery`.

thorn0:
@cola119 I'd say it's okay to fix #7224 in this PR.

cola119:
@thorn0 thanks your reviews. I modified the pointed out items and fixed https://github.com/prettier/prettier/issues/7224. 
Sorry in my poor English but please give me a review :bow:

cola119:
@sosukesuzuki san,
I fixed to support long expressions and trailing comma. Please let me know if there are any problems.

PR #7235, opened 2020-01-09
use lodash/* instead of lodash.*

thorn0:
The `lodash.*` packages are deprecated and abandoned. The recommended thing is to use `lodash/*` instead. See https://github.com/lodash/lodash/issues/3838


thorn0:
@evilebottnawi Maybe, but I just was going to use more functions from there. :) E.g. [sortedIndexOf](https://lodash.com/docs#sortedIndexOf) to work with sorted arrays of tokens in #7161.

PR #7251, opened 2020-01-11
Enable `unicorn/prefer-string-slice`

fisker:
Remove usage of `String#substr` and `String#substring`

deps on https://github.com/prettier/prettier/pull/7250 , `eslint-plugin-unicorn@15` can fix more cases on this rule


src/language-graphql/printer-graphql.js
```diff
@@ -669,7 +669,7 @@ function determineInterfaceSeparator(originalSource) {
   if (end === -1) {
     end = originalSource.length;
   }
-  return originalSource.substr(start, end).includes("&") ? " & " : ", ";
+  return originalSource.slice(start, end).includes("&") ? " & " : ", ";
```
thorn0: And this. Please double check other changes.

src/language-graphql/printer-graphql.js
```diff
@@ -288,7 +288,7 @@ function genericPrint(path, options, print) {
               " implements ",
               join(
                 determineInterfaceSeparator(
-                  options.originalText.substr(
+                  options.originalText.slice(
                     options.locStart(n),
                     options.locEnd(n)
```
thorn0: This doesn't look right.

fisker: Good catch, not equal. But I think old one is wrong, will think a case to test and confirm

fisker:
Don't merge until bugs in #7305 is fixed.

src/language-graphql/printer-graphql.js
fisker: fixed in #7305

src/language-graphql/printer-graphql.js
fisker: fixed in #7305

src/language-js/utils.js
```diff
@@ -645,7 +645,9 @@ function isLastStatement(path) {
 function isFlowAnnotationComment(text, typeAnnotation, options) {
   const start = options.locStart(typeAnnotation);
   const end = skipWhitespace(text, options.locEnd(typeAnnotation));
-  return text.substr(start, 2) === "/*" && text.substr(end, 2) === "*/";
+  return (
+    text.slice(start, start + 2) === "/*" && text.slice(end, end + 2) === "*/"
+  );
```
fisker: I feel this is not right, but tests passed.

fisker: No, it's right

scripts/build/babel-plugins/transform-custom-require.js
```diff
@@ -22,7 +22,7 @@ module.exports = function(babel) {
         if (isEvalRequire(node.callee) && node.arguments.length === 1) {
           let arg = node.arguments[0];
           if (t.isLiteral(arg) && arg.value.startsWith(".")) {
-            const value = "." + arg.value.substring(arg.value.lastIndexOf("/"));
+            const value = "." + arg.value.slice(arg.value.lastIndexOf("/"));
```
fisker: This one is not equal, since 'foo'.slice(-1) === 'o', 'foo'.substring(-1) === 'foo'

But `require` startsWith `.` always has `/`(except `require('.')` or `require('..')`), I think.

PR #7260, opened 2020-01-11
Update `mem` to v6

fisker:
We use this module in two files:

- [src/config/resolve-config-editorconfig.js](https://github.com/prettier/prettier/blob/master/src/config/resolve-config-editorconfig.js)

As I understand `editorconfigAsyncNoCache` and `editorconfigSyncNoCache` accept two params, they are all valid JSON data, they can all safely serialized by `JSON.stringify`.

So `{ cacheKey: JSON.stringify }` should good enough.

- [src/config/resolve-config.js](https://github.com/prettier/prettier/blob/master/src/config/resolve-config.js)

`getExplorerMemoized` only accept one argument, so no need use `cacheKey`

fixes: #7079


fisker:
You are right, I read the source again, we need calculate cacheKey, I didn't think through.

PR #7265, opened 2020-01-11
Use `lodash/partition`

fisker:
We have `lodash` now from #7235 , but if 

> Maybe we can remove lodash in future? Should be not hard

https://github.com/prettier/prettier/pull/7235#pullrequestreview-340606697

Feel free to close this PR.


evilebottnawi:
To be honestly I prefer avoid using lodash, but it is a development tool, so size is not matter

thorn0:
I think using popular libraries makes the project more contributor-friendly.

evilebottnawi:
@thorn0 yes, but we don't use a lot of functions/helpers/etc from lodash so we can reduce package size

thorn0:
I guess it's included anyway in Prettier's dependencies.

PR #7269, opened 2020-01-11
Simplify `getSupportInfo`

fisker:

### DIFF
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -28,10 +28,10 @@ function getSupportInfo(version, opts) {
   )
     .filter(option => filterSince(option) && filterDeprecated(option))
     .sort((a, b) => (a.name === b.name ? 0 : a.name < b.name ? -1 : 1))
+    .map(mapDeprecated)
+    .map(mapInternal)
     .map(option => {
-      option = {
-        ...mapInternal(mapDeprecated(option))
-      };
+      option = { ...option };
 
       if (Array.isArray(option.default)) {
         option.default =
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -6,15 +6,13 @@ const currentVersion = require("../../package.json").version;
 const coreOptions = require("./core-options").options;
 
 function getSupportInfo(version, opts) {
-  opts = Object.assign(
-    {
-      plugins: [],
-      showUnreleased: false,
-      showDeprecated: false,
-      showInternal: false
-    },
-    opts
-  );
+  opts = {
+    plugins: [],
+    showUnreleased: false,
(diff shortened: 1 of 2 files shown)


src/main/support.js
```diff
@@ -133,19 +107,16 @@ function getSupportInfo(version, opts) {
     if (!object.deprecated || opts.showDeprecated) {
       return object;
     }
-    const newObject = Object.assign({}, object);
-    delete newObject.deprecated;
-    delete newObject.redirect;
+    // eslint-disable-next-line no-unused-vars
+    const { deprecated, redirect, ...newObject } = Object.assign({}, object);
```
thorn0: Let's enable https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings , shall we?

fisker: I thought about it, but I thought this case 

```js
const {a, ...b} = c
```

if a and b both not used, it  will not report an error.

But I'm not sure now.. not near my laptop.

thorn0: BTW, `Object.assign` is not needed here.

fisker: Oops, missed this one.

fisker: Fixed.

Also enabled ignoreRestSiblings,

```js
const {a, ...b} = c
```

If a&b not used, report b is not used.
If a used b not used, report b is not used.
If b used a not used, no error.

src/main/support.js
```diff
@@ -25,43 +23,33 @@ function getSupportInfo(version, opts) {
...
-      ),
-      coreOptions
-    ),
+    Object.assign({}, ...plugins.map(({ options }) => options), coreOptions),
     "name"
   )
-    .sort((a, b) => (a.name === b.name ? 0 : a.name < b.name ? -1 : 1))
     .filter(option => filterSince(option) && filterDeprecated(option))
-    .map(mapDeprecated)
-    .map(mapInternal)
```
thorn0: This looked good. Why change?

fisker: Old one `map` 3 times, generate two extra new array, should merge into one `map`, for better performance.

thorn0: This place is not performance-critical, so please revert.

src/main/support.js
```diff
@@ -25,43 +23,33 @@ function getSupportInfo(version, opts) {
...
+    .sort((a, b) => (a.name === b.name ? 0 : a.name < b.name ? -1 : 1))
```
fisker: I changed `filter` and `sort` order, so array for `sort` could be smaller.

Issue #7236, opened 2020-01-09
HTML tag with colon breaks formatting


louis-bompart:
**Prettier 1.19.1**
[...]

PR #7273, opened 2020-01-11
Fix HTML elements with colon & lower case HTML tags everywhere

fisker:
fixes: #7236 

- [stable playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeA7gSxgCyZAG2gD4AdKVAE0wDdiACIiZl1t9jzr7rgc3oBGRMAGt6qAPTU65KrQZCIo8VPly6g4WMnSG6hVpW79mpdtV0Ti5TvmyADgCc4DJ3HpPycSW7IV4AB4wAIbOwa7OALTojsH2Ho5e0bHxkoEhYX6oAM72wVAMTDzFJaUl-JhQBJXukrn5WfUF9JXVULUSTcQ5ec2tNSpd4kP97YO93SNVA3UTshJYuPgQRAUgADQgEPYwmNDZyKChjhDoAAqhCAcowTQQmJQbIAKxonAwAMp5YJW8yDCOACucE2OBgAFsCAB1HDYOD1MBwD5XbC0bAAT2Q4GyB02lWycEcMDOsV44OCyAAZsECATNgArbIBABCrxE7w+wXBcAAMjUqTS6SBGQEPr8CHAAIqAiDwAW0kEgPKOAmOLFgyFPJyVGBQh64ZAADgADJsnBACVCUli3KqaHAngBHGXwEnba4gYLZSLtOCUP1PZzOzDOEnBMkUpDUhWbAngzD-IGK7LiqUuh1RwWKkICPWUA1IABMmwBwUwbV4AGEIODyVioNAHZtAQSACrBATXaNCmjAgCSUH9sA+YEcmB2AEFBx8YOiJfKCQBfRdAA)

- [pr playground](https://deploy-preview-7273--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeA7gSxgCyZAG2gD4AdKVAE0wDdiACIiZl1t9jzr7rgc3oBGRMAGt6qAPTU65KrQZCIo8VPly6g4WMnSG6hVpW79mpdtV0Ti5TvmyADgCc4DJ3HpPycSW7IV4AB4wAIbOwa7OALTojsH2Ho5e0bHxkoEhYX6oAM72wVAMTDzFJaUl-JhQBJXukrn5WfUF9JXVULUSTcQ5ec2tNSpd4kP97YO93SNVA3UTshJYuPgQRAUgADQgEPYwmNDZyKChjhDoAAqhCAcowTQQmJQbIAKxonAwAMp5YJW8yDCOACucE2OBgAFsCAB1HDYOD1MBwD5XbC0bAAT2Q4GyB02lWycEcMDOsV44OCyAAZsECATNgArbIBABCrxE7w+wXBcAAMjUqTS6SBGQEPr8CHAAIqAiDwAW0kEgPKOAmOLFgyFPJyVGBQh64ZAADgADJsnBACVCUli3KqaHAngBHGXwEnba4gYLZSLtOCUP1PZzOzDOEnBMkUpDUhWbAngzD-IGK7LiqUuh1RwWKkICPWUA1IABMmwBwUwbV4AGEIODyVioNAHZtAQSACrBATXaNCmjAgCSUH9sA+YEcmB2AEFBx8YOiJfKCQBfRdAA)


changelog_unreleased/html/pr-7273.md
```diff
@@ -0,0 +1,29 @@
+#### Fix HTML elements inside unknown namespace element ([#7273](https://github.com/prettier/prettier/pull/7273) by [@fisker](https://github.com/fisker))
```
thorn0: In HTML5, colons don't have special meaning in tag names, so mentioning "namespaces" is the changelog is incorrect. If not a bug, they're just an internal thing of the Angular parser.

src/language-html/utils.js
```diff
@@ -488,12 +488,28 @@ function getNodeCssStyleDisplay(node, options) {
...
       return "block";
     default:
+      if (node.type === "element" && isUnknownNamespaceElement) {
+        const display = CSS_DISPLAY_TAGS[node.name];
+        if (display) {
+          return display;
+        }
+
+        if (!HTML_TAGS[node.name]) {
+          return "block";
```
thorn0: This should be `inline` (`CSS_DISPLAY_DEFAULT`). The formatting of unknown (custom) tags should not depend on whether those tags are nested in a tag whose name contains a colon.

fisker: If set to inline.

Will format to 
```html
<foo:bar>
  <div>block block block</div>
  <div>block block block</div>
  <div>block block block</div>
  <span> inline inline </span><span> inline inline </span
  ><span> inline inline </span><span> inline inline </span
  ><html:button> known inline</html:button></foo:bar
>
```

because `<foo:bar>` is inline, seems wired.

thorn0: `<foo-bar>` is treated the same way, the safest way to deal with unknown elements. There should be no difference between `foo-bar` and `foo:bar`.

src/language-html/utils.js
```diff
@@ -494,6 +494,18 @@ function getNodeCssStyleDisplay(node, options) {
...
+      ) {
+        const display = CSS_DISPLAY_TAGS[node.name];
+        if (display) {
+          return display;
+        }
+      }
+
       return (
         (node.type === "element" &&
           (!node.namespace || isInSvgForeignObject) &&
```
thorn0: Can't we just add `|| !["html", "svg"].includes(node.namespace)` to this `if` instead?

src/language-html/utils.js
```diff
@@ -496,7 +496,10 @@ function getNodeCssStyleDisplay(node, options) {
     default:
       return (
         (node.type === "element" &&
-          (!node.namespace || isInSvgForeignObject) &&
+          (!node.namespace ||
+            isInSvgForeignObject ||
+            (!node.hasExplicitNamespace &&
+              !["html", "svg"].includes(node.namespace))) &&
```
thorn0: The `getNodeCssStyleWhiteSpace` function below contains suspiciously similar logic. Could you investigate whether it needs fixing too and whether they can be unified?

fisker: Right, `<pre>` lost whitespace

fisker: Working on 3 repos, just now. no time to reply, sorry

fisker:
Should be good now, thanks @thorn0 . going to sleep now

thorn0:
Looks good, but unfortunately there still is a difference between formatting `<foo-bar>` and `<foo:bar>`:

**Prettier pr-7273**
[Playground link](https://deploy-preview-7273--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeAZhCBaARgQwCcA+AHSgAILUAHQuYiuuK51w8qlz6gembJRUvTDgIly5DFiTjBVWvUZsm9dj26U+AqSJlyQAGhAQaMAJbQAzslBFCEAO4AFIghsp8ANwjmAJkYguIT4YADWcDAAyjSh5lAA5sgwhACucMYAFjAAtgA2AOqZ5vBWsWBwUe4l5l4lAJ7I4FY2xvFWcIQwziEJOfjI6Ph5HcYAVlYAHgBCIeGRUfg5cAAy8XCDw6MgE5NR8Ql5cACKqRDwmyMZILGEHYRN2fmBdPEwBf4wmcgAHAAMxjoEA6BRCNCazHuXg2xgAjmd4D1TB4QPgrNgoHA4H5sYF6PDzPQevg+gMkEMrsYOjlzMk0tcrAcjqdzhtyVtrjB8LgPn4vsgAEzGFL4cx5A4AYQgOX6TSg0BhIFSHQAKtyPBTtl50gBJKA42BRMCEcxmACC+qiMHqR0uHQAvvagA)
```sh
--parser html
```

**Input:**
```html
<foo-bar>
  <pre> pre  pre  pr
  e 
  </pre>
</foo-bar>

<foo:bar>
  <pre> pre  pre  pr
  e 
  </pre>
</foo:bar>
```

**Output:**
```html
<foo-bar>
  <pre>
 pre  pre  pr
  e 
  </pre>
</foo-bar>

<foo:bar>
  <pre
> pre  pre  pr
  e 
  </pre>
</foo:bar>

```

fisker:
will check latet

fisker:
Found another problem, uppercase tag not normalized

fisker:
Tests failed, should this `span` lowercased?

```html
<x:root><SPAN>foreign tag name should not be lower cased</SPAN></x:root>
```

thorn0:
> should this span lowercased?

I'd say it shouldn't.

HTML5 [supports two syntaxes](https://www.w3.org/TR/html5-diff/#syntax): "HTML" and "XML".

- In the HTML syntax:
  - colons have no special meaning in tag names
  - tag names are case-insensitive (except for embedded SVG and MathML)
- In the XML syntax:
  - colons are used for namespace prefixes in tag names
  - _all_ tag names are case-sensitive, no matter what parents the element has

As we can see, what Prettier does is already incompatible with the XML syntax, so it's totally okay to keep ignoring that syntax. If somebody wants XML, there exists https://github.com/prettier/plugin-xml for them.

However, this all applies only to 'vanilla' HTML5. Angular and Vue might have their own peculiarities in this regard.

thorn0:
Sorry, I edited my previous message, but you might have not noticed that:

> I'd say it should be lowercased, just like any other span.

Or do you have some other reason for this [commit](https://github.com/prettier/prettier/pull/7273/commits/4cb5272bb4b8c021c73be42540c85625d08659b2)?

fisker:
@thorn0 I thought you want them keep upercased.

evilebottnawi:
@thorn0 

> As we can see, what Prettier does is already incompatible with the XML syntax

In which places we are incompatibility?

thorn0:
@evilebottnawi All names are case-sensitive in XML. Prettier lowercases tag names.

evilebottnawi:
@thorn0 oh, yes, i think we should documented that (preferable, because xhtml is old stuff) or we can look at doctype and do not lowercase if it is xml

thorn0:
@evilebottnawi Let's merge this PR and open an issue about XHTML support, shall we?

evilebottnawi:
@thorn0 Sounds good!

PR #7296, opened 2020-01-13
Don't swallow multi-parser-error on test

fisker:
Don't swallow multi-parser-error on test

### DIFF
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -119,7 +119,7 @@ function callPluginPrintFunction(path, options, printPath, args) {
       }
     } catch (error) {
       /* istanbul ignore if */
-      if (process.env.PRETTIER_DEBUG) {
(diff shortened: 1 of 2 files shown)


src/main/ast-to-doc.js
```diff
@@ -119,7 +119,10 @@ function callPluginPrintFunction(path, options, printPath, args) {
       }
     } catch (error) {
       /* istanbul ignore if */
-      if (process.env.PRETTIER_DEBUG) {
+      if (process.env.PRETTIER_DEBUG || process.env.NODE_ENV === "test") {
```
thorn0: I think this should be controlled by something more dynamic than environment variables. In a perfect world, the playground would have a checkbox to control showing these debug-only errors (with "don't show" as the default value), and the tests would be able to pass an option to unsuppress these error on demand, only in those tests where this is needed. But again, by default, both the tests and the playground should behave exactly like Prettier behaves in production, namely suppress these errors.

PR #7302, opened 2020-01-13
Upgrade Node.js version and drop mkdirp

shisama:
Node.js v8 moved to End-of-Life on Dec 31, 2019. ( Please see https://github.com/nodejs/Release)
Node.js v10.13.0 is minimum LTS version now.
Moreover, new API's options in Node.js Core are added in Node.js v10.13.0. 
- [withFileTypes](https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options) option for `fs.readdirSync` can make code faster than using `fs.stat`.
- [recursive](https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_options) option for `fs.mkdirSync` can recursively make directories so that `mkdirp` package is no longer need.

### DIFF
--- a/scripts/sync-flow-tests.js
+++ b/scripts/sync-flow-tests.js
@@ -3,7 +3,6 @@
 const fs = require("fs");
 const flowParser = require("flow-parser");
 const globby = require("globby");
-const mkdirp = require("mkdirp");
 const path = require("path");
 const rimraf = require("rimraf");
 
@@ -65,7 +64,7 @@ function syncTests(syncDir) {
     const specFile = path.join(dirname, SPEC_FILE_NAME);
     const specContent = specContents[specFile] || DEFAULT_SPEC_CONTENT;
 
-    mkdirp.sync(dirname);
+    fs.mkdirSync(dirname, { recursive: true });
     fs.writeFileSync(newFile, content);
     fs.writeFileSync(specFile, specContent);
   });
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -20,12 +20,13 @@ global.run_spec = (dirname, parsers, options) => {
     throw new Error(`No parsers were specified for ${dirname}`);
   }
 
-  fs.readdirSync(dirname).forEach(basename => {
+  fs.readdirSync(dirname, { withFileTypes: true }).forEach(dirent => {
+    const basename = dirent.name;
     const filename = path.join(dirname, basename);
 
     if (
       path.extname(basename) === ".snap" ||
-      !fs.lstatSync(filename).isFile() ||
+      !dirent.isFile() ||
       basename[0] === "." ||
       basename === "jsfmt.spec.js"
     ) {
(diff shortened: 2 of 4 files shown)


tests_config/run_spec.js
```diff
@@ -22,12 +22,13 @@ global.run_spec = (dirname, parsers, options) => {
     throw new Error(`No parsers were specified for ${dirname}`);
   }
 
-  fs.readdirSync(dirname).forEach(basename => {
+  fs.readdirSync(dirname, { withFileTypes: true }).forEach(dirent => {
```
fisker: This part can commit to `next` branch, already requires node 10. maybe also change to `for...of`

scripts/sync-flow-tests.js
```diff
@@ -66,7 +65,7 @@ function syncTests(syncDir) {
     const specFile = path.join(dirname, SPEC_FILE_NAME);
     const specContent = specContents[specFile] || DEFAULT_SPEC_CONTENT;
 
-    mkdirp.sync(dirname);
+    fs.mkdirSync(dirname, { recursive: true });
```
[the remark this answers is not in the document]
shisama: @fisker Thanks. Since Node.js v10 provides `fs.mkdirSync` with `recursive` option, I think `mkdirp` package is unnecessary. This part has to be changed from `mkdirp` to `fs.mkdirSync` to remove `mkdirp` package from dependencies.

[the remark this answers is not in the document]
fisker: I mean just use `fs.mkdirSync` don't care `engine`.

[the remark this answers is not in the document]
shisama: you mean `mkdirp` doesn't have to be removed?

tests_config/run_spec.js
shisama: fixed it. Thank you.

tests_config/run_spec.js
```diff
@@ -20,16 +20,18 @@ global.run_spec = (dirname, parsers, options) => {
     throw new Error(`No parsers were specified for ${dirname}`);
   }
 
-  fs.readdirSync(dirname).forEach(basename => {
+  const dir = fs.readdirSync(dirname, { withFileTypes: true });
+  for (const dirent of dir) {
```
fisker:
```suggestion
  for (const file of files) {
```

yarn.lock
```diff
@@ -5150,7 +5150,7 @@ mixin-deep@^1.2.0:
     for-in "^1.0.2"
     is-extendable "^1.0.1"
 
-mkdirp@0.5.1, mkdirp@^0.5.1:
+mkdirp@^0.5.0, mkdirp@^0.5.1:
```
fisker: Is this right? Please run `yarn` check if `yarn.lock` changes.

shisama: No. I made mistake. I reverted this.

sosukesuzuki: Has this fixed already ? CI seems to still fail due to `yarn.lock`.

fisker: @shisama Just run `yarn` and commit it, should be good.

shisama: @sosukesuzuki @fisker Thank you for your help. CI passed.

evilebottnawi:
Thanks!

PR #7305, opened 2020-01-13
Improve separator between interfaces detection

fisker:
- [stable playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAKpuARmwEsBbDAGzjIRgGdsBBAGmwCEAdKAYm2GwAybNgA8AWnHYASjQgA3HDAAWJRpDK1Y2MMoCGUAOZxG9OBj0AnPTAiXSUbPRh6ARtW7A9SbHoC+3HwAyuZWNnbYANZw5r6MJDCk8Y4qajp6ZtzoWHgEAEykFNRaDMxs7ELYAMJePv4gLCAQGDAk0PTIoFaWEADuAApWCB0oevIQJAAmDSCu1mDRMEEWYCRGyDCWAK5wjcowZJQA6qrw9CtwIVD0CSTyCWjI4PQdjWtmljD91oZkesgAMz0lDMjQAVvQAB7seaLIJ6WgAGTWcEBwNBIAhkKCa0M1AAilsIPA0SDdiALJYPk9DNYMMoAI6UGYYSxrGBHKYqZAADgADI1WRAzEc6U9WSY4JZFDMGUT4N9miMQBlxFAYpM4NNGpY4HKSLrvnpfv8kECyY0zGQSBttuSbkYCfLUWb0eSXK5OZNuUg8o1NnoSJRcVUIJpTSAoNBUY0tmZcG4RuaMfIdgBJKCa2BBMBslpMTNBdDUUlmPx+IA)

### DIFF
--- a/changelog_unreleased/graphql/pr-7305.md
+++ b/changelog_unreleased/graphql/pr-7305.md
@@ -9,7 +9,6 @@ type Type1 implements A, B
 # { &  <-- Removing this comment changes the separator in stable
 {a: a}
 
-# Separator keep as it is in this case
 type Type2 implements A, B & C{a: a}
 
 # Prettier stable
@@ -18,7 +17,6 @@ type Type1 implements A & B {
   a: a
 }
 
-# Separator keep as it is in this case
 type Type2 implements A & B & C {
   a: a
 }
@@ -29,7 +27,6 @@ type Type1 implements A, B {
   a: a
 }
 
-# Separator keep as it is in this case
 type Type2 implements A, B & C {
   a: a
 }
--- a/changelog_unreleased/graphql/pr-7305.md
+++ b/changelog_unreleased/graphql/pr-7305.md
@@ -1,6 +1,6 @@
-#### Improve separator between interfaces detection ([#7305](https://github.com/prettier/prettier/pull/7305) by [@fisker](https://github.com/fisker))
+#### Improve detection of separator between interfaces ([#7305](https://github.com/prettier/prettier/pull/7305) by [@fisker](https://github.com/fisker))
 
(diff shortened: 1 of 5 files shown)


fisker:
@thorn0 I make a test case

```graphql
# SOME COMMENT BEFORE, this effects string slice
type AA implements C, D
# { &
# Should separat by \`,\` not \`&\`
```

So this part https://github.com/prettier/prettier/blob/master/src/language-graphql/printer-graphql.js#L291-L294

will substr string to `` type AA implements C, D\n# { &\n# Should separat by `,` not `&` ``

then pass to `determineInterfaceSeparator` function https://github.com/prettier/prettier/blob/617d3d6b1e1504b8c44262e2ebab28f0d367a4b2/src/language-graphql/printer-graphql.js#L672

this line substr will be `implements C, D\n# { &\n# Sh` then check `&`, but `&` in a comment.

I'm not familiar with `graphql`, but this definitely not right.

fisker:
Even use `.slice()` seems still not right,

```graphql
type AA implements C, D
# { &     <-------------------------------------will targeting the wrong `{`
# Should separat by \`,\` not \`&\`
{
...
}
```

Maybe should be `C, D` range, but also maybe comment between them


```graphql
type AA implements C, # &&&&&&&&&&&&&&& <- actually the Separator is `,`
D
&
{
...
}
```

thorn0:
Sure, it's a bug. Include these tests in #7251.

fisker:
I was thinking maybe we can improve detection.

fisker:
@thorn0 I make a improvement, please review.

fisker:
Can we merge this?

PR #7245, opened 2020-01-10
Upgrade @glimmer/syntax to latest

dcyriller:
## Description

Commit messages will explain minor cleanups done on the way.

Two things are worth noticing:
#### a test removal (EDIT: fixed in https://github.com/prettier/prettier/pull/7306)
```
<script id="entry-template" type="text/x-handlebars-template">
  <div class="entry">
    <h1>{{title}}</h1>
    <div class="body">
      {{body}}
    </div>
  </div>
</script>
```
is legacy hbs syntax not supposed to be parsed by `glimmer`. Here is the AST output of this code: https://astexplorer.net/#/gist/188be5ea74c0d409791417f60a8ea40f/08ee2a309d8e9e0961902f9df82a7fa7ade12767

It was failing, so I've removed the file testing it.

#### a workaround (EDIT: fixed)

glimmer throws on parsing `<TITLE></TITLE`. I am unsure why. So I have changed the tag to `<title></title` while https://github.com/glimmerjs/glimmer-vm/pull/1007 is being answered.

## Checklist

- [ ] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [ ] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [ ] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

Issue #5340, opened 2018-11-05
Support Handlebars


suchipi:
I am opening this issue to track Handlebars support. It was previously tracked in https://github.com/prettier/prettier/issues/1882 but that was closed when we added HTML support, so this issue is for Handlebars only.

buschtoens:
Am I correct in assuming, that work has already started in [`language-handlebars`](https://github.com/prettier/prettier/tree/master/src/language-handlebars), but is not usable yet?

suchipi:
@buschtoens yes

peterjmag:
For anyone who might be watching this, you can already try this out like so:

```sh
yarn prettier '**/*.hbs' --parser glimmer
```

Stolen from @ikatyang's comment here: https://github.com/prettier/prettier/pull/3534#issuecomment-355588250.

ikatyang:
Please note that the glimmer printer is currently unrelated to the HTML printer, which means their formatting rules are not the same. To _support_ Handlebars, we need to somehow combine them together, see https://github.com/prettier/prettier/issues/5511#issuecomment-440511001 for more info.

mydea:
I've tried this on our codebase (with `--parser glimmer`), and in general it seemed to be working quite well, except for one issue: It removes all blank lines, which makes the resulting code rather hard to understand, sometimes. Was this a conscious decision? It is at odds with this: https://prettier.io/docs/en/rationale.html#empty-lines and how this works in JS/CSS.

lydell:
@mydea That does not sound intended. Please share an example on the [playground](https://prettier.io/playground/).

mydea:
Hmm, I can't select the `glimmer` parser there, as far as I can see?

But what it does is format this:

```hbs
{{test}}

{{other-test}}


{{final-test}}

```

to 

```hbs
{{test}}
{{other-test}}
{{final-test}}
```

lydell:
Ah, sorry, I didn't know glimmer wasn't available there.

Alonski:
I think adding Glimmer support to the playground should be done anyways so that we can at least experiment with it there.

I would really love to make this official.

mfeckie:
When processing conditionals the parser / formatter takes liberties with multi level nesting.  On occasion it combines the conditions in ways that end up with broken logic.

[...]

Issue #4908, opened 2018-07-30
[QUEST] Glimmer


tchak:
## Objective

We want to get Prettier’s support for Ember templates up to the point where it is complete and maintainable enough to be officially released.

This quest issue was created to track the outstanding work required to meet this objective and to help volunteers coordinate to get the work done.

## Contributing

You can follow and participate in the work on the [collaborative issue]( https://github.com/jgwhite/prettier/issues/1)

tchak:
Maybe I missed some discussion/context. In which case you can point me to some reading :)

ikatyang:
We can officially release it once it's able to prettier all the supported syntaxes (i.e., every cases mentioned in its spec or its official test cases), I'm not familiar with handlebars so I'm not sure if it's ready.

tchak:
It's not :)
I will continue to fix obvious missing syntax and will try to find what would be the official reference.

tchak:
@rwjblue what is the most up to date document with all the Glimmer syntaxes?

Also, do you have any idea on why `{{collection.[2]}}` is parsed as `{{collection.2}}` and I have no hint in the AST about the original form?

Alonski:
@tchak Can anything be done to help get all the issues for this fixed?

efx:
On a related note, I brought up the above syntax in discord and it apparently shouldn't work in Ember templates: https://github.com/ember-template-lint/ember-cli-template-lint/issues/437

tchak:
@Alonski I think the best thing you can do is to try it and report any issues/suggestions in here. I will remove the`Array paths` issue problem from the list as it appears to not be officially supported and it is a pain to implement.

Alonski:
@tchak So it seems that maybe this is more ready than we though? I am in the process of getting my dev team to start using it and to autofix all templates.

Maybe all that is missing is some tests around various templates before/after?

tchak:
I think some new stuff (like modifiers) was added to glimmer since last time I looked and would need to be checked.

duailibe:
@GavinJoyce is working through some changes, if you want to collaborate.

jgwhite:
@tchak @GavinJoyce et al. we (Heroku) are also motivated to improve Glimmer support in Prettier. Shall we try to coordinate our efforts somehow? Quest issue? ⚔️ 😅

GavinJoyce:
@jgwhite nice one!

[...]

PR #7306, opened 2020-01-13
Support printing inline handlebars in html (through embed function)

dcyriller:
### Context

An ancient way of using handlebars is through a specific script tag in html. For instance, the code _inside_ the `script` tag is actually also handlebars:
```html
<span>in an html file:</span>
<script type="text/x-handlebars-template">
  <h1>{{title}}</h1>
</script>
```

### Description

In handlebars test suite, there was an old test using this syntax. It was hidden in a handlebars file. This PR moves it to a html file to process it as a html file. This PR proposes to support printing this html code with glimmer printer.

### Why?

The goal is to move forward with `@glimmer/syntax` upgrade. See related discussion here: https://github.com/prettier/prettier/pull/7245 and https://github.com/prettier/prettier/pull/6570#issuecomment-550266720

@thorn0 this might be a better solution to pinning `simple-html-tokenizer` on next branch.

Thank you @rwjblue for the initial idea!

### An Unresolved Question

If accepted as is, this PR will would _always_ run glimmer parser for these inline handlebars. A problem is that handlebars is still an experiment in Prettier. For this reason, glimmer parser shoud be manually activated with the `--parser=glimmer` flag. Any pointer to how to resolve this issue is welcome.

### Checklist


thorn0:
> An Unresolved Question

Do I understand you correctly that you're asking whether it's okay to use an experimental feature (Glimmer formatter) inside a stable feature (HTML formatter)?

dcyriller:
Yes, you do. It is what I mean

tests/handlebars-basics/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,72 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`basic-handlebars.hbs 1`] = `
```
fisker: why this deleted.

dcyriller: Because it was basically empty.

This file contained two examples:
´´´
<script id="entry-template" type="text/x-handlebars-template">		
   <div class="entry">		
     <h1>{{title}}</h1>		
     <div class="body">		
       {{body}}		
     </div>		
   </div>		
 </script>
´´´	
and
´´´
  <div class="{{hello}} {{world}}"></div>
´´´

The first one is extracted to its own folder. I felt the second one is two basic to be relevant. Class attributes are tested in depth in test/handlebars-mustache-statement folder. Do you want me to revert the file deletion?

fisker: Last time I try to update `glimmer` package, this file did report an error, not sure what error, better keep it.

thorn0: @fisker This PR is a solution exactly for that problem. It's okay.

thorn0:
@dcyriller It seems to be okay, but to give a good answer, I need to understand why Glimmer support is still experimental. Does it lack some important functionality? Or is it time to move it out of experimental? (Sorry, I am really in the dark on this case. Trying to focus mostly on JS/TS issues.)

dcyriller:
No worries, I can understand that.

Handlebars support doesn't lack any important functionality. But I think it's not official yet because glimmer printer is still a bit eager about formatting whitespaces and newlines. Please [see this issue](https://github.com/prettier/prettier/blob/1.19.1/tests/handlebars-whitespace/__snapshots__/jsfmt.spec.js.snap#L406) or [this issue](https://github.com/prettier/prettier/blob/1.19.1/tests/handlebars-whitespace/__snapshots__/jsfmt.spec.js.snap#L38). This is a problem that could be solved by implementing [this comment](https://github.com/prettier/prettier/issues/5340#issuecomment-442674510).

I have been using Prettier for handlebars in two production apps for a moment now. I need to play with the `.prettierignore` file from time to time though. 😄 But with the next release, I'll be able to delete this file at all.

A few useful links: two links tracking progress on handlebars support [link 1](https://github.com/prettier/prettier/issues/5340) and [link 2](https://github.com/prettier/prettier/issues/4908). The second link might be closed for this [colloborative issue](https://github.com/jgwhite/prettier/issues/1). Also, I've forked a website using handlebars and formatted its hbs files with prettier#master. You can have a look [here](https://github.com/dcyriller/crates.io/commit/d72829c4f1df8eb962b3626842bc62c9e93241fd).

thorn0:
So it seems like the biggest blocker is the lack of support for [`--html-whitespace-sensitivity`](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting). Once it's implemented, I think we'll be able to make Glimmer support official.

As for this PR, given the legacy status of this `script` syntax, I think it's okay to proceed.

thorn0:
@dcyriller Please add a changelog entry.

changelog_unreleased/handlebars/pr-7306.md
```diff
@@ -0,0 +1,17 @@
...
+<script type="text/x-handlebars-template">{{component 'hey'}}</script>
+
+<!-- Output (Prettier stable) -->
+<script type="text/x-handlebars-template">
+  {{component 'hey'}}
+</script>
+
+<!-- Output (Prettier master) -->
+<script type="text/x-handlebars-template">
+  {{component "hey"}}
```
thorn0: Doesn't really look like an improvement over Prettier stable. Please make the input uglier. :)

dcyriller: Did you notice how it fixes a bug regarding quotes? ^^

I can make the change more obvious though.

dcyriller: Will do (but I need a bit more time 😬)

dcyriller: Here we go, it was worth it :)

PR #7320, opened 2020-01-14
Add changelog lint script

fisker:
So many errors in changelogs


fisker:
Do you know how to get PR number and commit files on CI, I think it's possible.

evilebottnawi:
Maybe it is in env variables? Need look at documentation, but for `commit files` you need run git commands i think

fisker:
If we can get the PR number, test can be stricter.

thorn0:
> If we can get the PR number, test can be stricter

Let's not overcomplicate things. Sometimes a PR is a fix for another PR in the same release, so they share one changelog entry.

fisker:
Then good to go

changelog_unreleased/handlebars/pr-7051.md
```diff
@@ -1,4 +1,4 @@
-#### handlebars: Fix superfluous line breaks in ConcatStatement ([7051](https://github.com/prettier/prettier/pull/7051) by [@dcyriller](https://github.com/dcyriller))
+#### handlebars: Fix superfluous line breaks in ConcatStatement ([#7051](https://github.com/prettier/prettier/pull/7051) by [@dcyriller](https://github.com/dcyriller))
```
thorn0: Can you please also validate that headers 1) have four `#`, 2) don't start with redundant prefixes like "handlebars:", "HTML:", etc.?

cspell.json
```diff
@@ -122,6 +122,7 @@
         "gitattributes",
         "githook",
         "gitignore",
+		"gitkeep",
```
thorn0: tabs instead of spaces?

scripts/lint-changelog.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+const authorRegex = /by \[@(.*?)\]\(https:\/\/github\.com\/\1\)/;
+
+const template = fs.readFileSync(
+  path.join(CHANGELOG_ROOT, TEMPLATE_FILE),
+  "utf8"
+);
+const templateComment = template.match(/<!--[\s\S]*?-->/);
```
thorn0: It still works because include converts arrays to string, but you forgot `[0]`.

fisker: fixed https://github.com/prettier/prettier/pull/7320/commits/a6fa99696b76e840e8e8fe40093457b4e40b8e6b

cspell.json
fisker: fixed https://github.com/prettier/prettier/pull/7320/commits/e8dbd2369d616bdf75a56d47803c7867cd0621a8

changelog_unreleased/handlebars/pr-7051.md
fisker: Done.

thorn0: yes

fisker: Done https://github.com/prettier/prettier/pull/7320/commits/baa548761a13638de335d1741a17fe601cf96626

PR #7330, opened 2020-01-15
Use array destructuring for some simple case

fisker:

### DIFF
--- a/src/document/doc-printer.js
+++ b/src/document/doc-printer.js
@@ -154,9 +154,7 @@ function fits(next, restCommands, width, options, mustBeFlat) {
     }
 
     const x = cmds.pop();
-    const ind = x[0];
-    const mode = x[1];
-    const doc = x[2];
+    const [ind, mode, doc] = x;
 
     if (typeof doc === "string") {
       out.push(doc);
@@ -254,9 +252,7 @@ function printDocToString(doc, options) {
 
   while (cmds.length !== 0) {
     const x = cmds.pop();
-    const ind = x[0];
-    const mode = x[1];
-    const doc = x[2];
+    const [ind, mode, doc] = x;
 
     if (typeof doc === "string") {
       out.push(doc);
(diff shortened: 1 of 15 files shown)


src/document/doc-printer.js
```diff
@@ -154,9 +154,7 @@ function fits(next, restCommands, width, options, mustBeFlat) {
     }
 
     const x = cmds.pop();
-    const ind = x[0];
-    const mode = x[1];
-    const doc = x[2];
+    const [ind, mode, doc] = x;
```
jamesgeorge007:
```suggestion
    const [ind, mode, doc] = cmds.pop();
```

jamesgeorge007: I can't see the variable `x` serving it's purpose anywhere else.

src/document/doc-printer.js
```diff
@@ -254,9 +252,7 @@ function printDocToString(doc, options) {
 
   while (cmds.length !== 0) {
     const x = cmds.pop();
-    const ind = x[0];
-    const mode = x[1];
-    const doc = x[2];
+    const [ind, mode, doc] = x;
```
jamesgeorge007:
```suggestion
    const [ind, mode, doc] = cmds.pop();
```

src/document/doc-printer.js
```diff
@@ -254,9 +252,7 @@ function printDocToString(doc, options) {
 
   while (cmds.length !== 0) {
     const x = cmds.pop();
```
jamesgeorge007:
```suggestion
```

src/document/doc-printer.js
```diff
@@ -154,9 +154,7 @@ function fits(next, restCommands, width, options, mustBeFlat) {
     }
 
     const x = cmds.pop();
```
jamesgeorge007:
```suggestion
```

thorn0:
I'm really not sure the readability improves in simple cases like this (with only one variable declared):
```diff
-  const child = node.children[0];
+  const [child] = node.children;
```
Let's keep them as is.

src/document/doc-printer.js
fisker: Good idea, updated. Thanks!

PR #7334, opened 2020-01-15
Tiny refactor

fisker:

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -3927,25 +3927,24 @@ function printJestEachTemplateLiteral(node, expressions, options) {
       }
     }
 
-    const maxColumnCount = tableBody.reduce(
-      (maxColumnCount, row) => Math.max(maxColumnCount, row.cells.length),
-      headerNames.length
+    const maxColumnCount = Math.max(
+      headerNames.length,
+      ...tableBody.map(row => row.cells.length)
     );
 
-    const maxColumnWidths = Array.from(new Array(maxColumnCount), () => 0);
-    const table = [{ cells: headerNames }].concat(
-      tableBody.filter(row => row.cells.length !== 0)
-    );
-    table
-      .filter(row => !row.hasLineBreak)
-      .forEach(row => {
-        row.cells.forEach((cell, index) => {
-          maxColumnWidths[index] = Math.max(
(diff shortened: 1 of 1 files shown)


src/language-js/printer-estree.js
```diff
@@ -3927,25 +3927,24 @@ function printJestEachTemplateLiteral(node, expressions, options) {
...
-    table
-      .filter(row => !row.hasLineBreak)
-      .forEach(row => {
-        row.cells.forEach((cell, index) => {
-          maxColumnWidths[index] = Math.max(
-            maxColumnWidths[index],
-            getStringWidth(cell)
-          );
-        });
+    const maxColumnWidths = Array.from({ length: maxColumnCount }).fill(0);
```
armano2:
```suggestion
    const maxColumnWidths = new Array(maxColumnCount).fill(0);
```

fisker: No , thanks, I  think this is exactly why es has `Array.from` `Array.of`.

armano2: https://2ality.com/2018/12/creating-arrays.html#cheat-sheet%3A-creating-arrays
https://2ality.com/2018/12/creating-arrays.html#recommended-patterns

fisker: Personally I always use `Array.from({length})` to generate array of this length, I think it's best practice.

Another PR for eslint also use it https://github.com/eslint/eslint/pull/12608/files

armano2: they are using it in different way, they are using callback to prefill values

```suggestion
    const maxColumnWidths = Array.from({ length: maxColumnCount }, () => 0);
```

armano2: they are using it in different way, they are using callback to prefill values

```suggestion
    const maxColumnWidths = Array.from({ length: maxColumnCount }, () => 0);
```

armano2: they are using it in different way, they are using callback to prefill values

```js
    const maxColumnWidths = Array.from({ length: maxColumnCount }, () => 0);
```

fisker: I wrote them differently on purpose. There is no difference in this case

fisker: There is no they, still me in that PR.

PR #7336, opened 2020-01-15
Fix `prefer-object-spread` in `/scripts`

fisker:


scripts/generate-schema.js
```diff
@@ -80,17 +78,15 @@ function generateSchema(options) {
...
+  return {
+    description: option.description,
+    default: option.default,
+    ...(option.array ? wrapWithArraySchema : identity)(
       option.type === "choice"
         ? { oneOf: option.choices.map(choiceToSchema) }
         : { type: optionTypeToSchemaType(option.type) }
     )
-  );
+  };
```
fisker: I think no tests cover this part, need review.

scripts/generate-schema.js
```diff
@@ -21,10 +21,8 @@ function generateSchema(options) {
     definitions: {
       optionsDefinition: {
         type: "object",
-        properties: options.reduce(
-          (props, option) =>
-            Object.assign(props, { [option.name]: optionToSchema(option) }),
-          {}
+        properties: Object.assign(
+          ...options.map(option => ({ [option.name]: optionToSchema(option) }))
         )
```
fisker: I think no tests cover this part, need review.

scripts/generate-schema.js
fisker: And this

scripts/generate-schema.js
thorn0: `tests_integration\__tests__\schema.js` seems to cover it.

PR #7338, opened 2020-01-15
Fix `prefer-object-spread` in `/website`

fisker:

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -14,7 +14,7 @@ plugins:
 rules:
   curly: error
   dot-notation: error
-  eqeqeq: 
+  eqeqeq:
     - error
     - always
     - null: ignore
@@ -25,6 +25,9 @@ rules:
   no-inner-declarations: error
   no-unneeded-ternary: error
   no-useless-return: error
+  no-unused-vars:
+    - error
+    - ignoreRestSiblings: true
   no-var: error
   object-shorthand: error
   one-var:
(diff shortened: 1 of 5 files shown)


website/playground/buttons.js
```diff
@@ -39,7 +39,7 @@ export class ClipboardButton extends React.Component {
   render() {
     const { children } = this.props;
     const { showTooltip, tooltipText } = this.state;
-    const rest = Object.assign({}, this.props);
+    const rest = { ...this.props };
     delete rest.children;
     delete rest.copy;
```
thorn0:
```js
const { children, copy, ...rest } = this.props;
```

.eslintrc.yml
```diff
@@ -25,6 +25,9 @@ rules:
   no-inner-declarations: error
   no-unneeded-ternary: error
   no-useless-return: error
+  no-unused-vars:
+    - error
+    - ignoreRestSiblings: true
```
fisker: See https://github.com/prettier/prettier/pull/7269#discussion_r365513743

website/playground/buttons.js
fisker: Fixed

fisker:
You guys prefer

```js
const { children, copy, ...rest } = this.props;
```

or

```js
const { children, copy: _, ...rest } = this.props;
```

?

In this case `copy` is not using

@sosukesuzuki @thorn0 @evilebottnawi

evilebottnawi:
Looks better:
```js
const { children, copy, ...rest } = this.props;
```

fisker:
better, but second can tell `copy` is not using

PR #7348, opened 2020-01-15
Update overrides.css

vnoitkumar:
Css changes for logos in the home page.

Before:
Before

After:
After

### DIFF
--- a/website/static/overrides.css
+++ b/website/static/overrides.css
@@ -372,14 +372,15 @@ em {
   color: #6f7b81 !important;
 }
 .usersSection a {
-  display: flex;
-  align-items: center;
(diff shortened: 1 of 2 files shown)


website/static/overrides.css
```diff
@@ -389,8 +389,7 @@ em {
 }
 
 .usersSection .user {
-  height: 100px;
-  width: 200px;
+  height: 80px;
```
armano2: what do you think about keeping width and changing height to: 
```suggestion
  max-height: 100px;
```
and align them vertically 
```css
  display: flex;
  align-items: center;
```

image

lipis:
Thank you!

PR #7355, opened 2020-01-16
Fix `prefer-object-spread` in `src/common`

fisker:
Part of #7339


fisker:
@evilebottnawi I do made #7339 , but that one seems too big, I thought easier to review in this way.

lipis:
I think we can do it all at once.. too much noice with separate PRs.

PR #7354, opened 2020-01-16
Fix `prefer-object-spread` in `src/config`

fisker:
Part of #7339

PR #7358, opened 2020-01-16
Fix `prefer-object-spread` in `src/language-js` (part 1)

fisker:
Part of #7339

### DIFF
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -222,7 +222,7 @@ function escapeTemplateCharacters(doc, raw) {
       }
     });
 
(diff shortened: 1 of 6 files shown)


armano2:
you guys should really do this in semi-auto way with eslint and `prefer-object-spread`
https://eslint.org/docs/rules/prefer-object-spread

fisker:
@armano2 I'm doing these PRs with `prefer-object-spread` fixes, but some code require adjustments, like https://github.com/prettier/prettier/pull/7354 , planning enable that rule after all codes fixed first.

Issue #6502, opened 2019-09-21
Print <!doctype html> lowercase rather than uppercase


k-yle:
**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAhAEwmGBPABzgAIALGAWwBsA+AHSlXOvqmOPgA8YHUB6ZrRAAaEBAIwAltADOyUAEMATkogB3AArKEclAoBuESZhEgARkoVgA1nBgBlAlclQA5shhKArnFGCA6qSS8DJOYHD2OsGS+sF4yOAycqIuMnBKMBqWrhQKyABmClRpogBWMpwAQpY2dvYKFHAAMi5wBUUlIOWc9i6uVHAAil4Q8O3FviBOSmlKCYKmBEouMP7GMKTIABwADKJLEGn+lgQJS3Cz+m2iAI4j8FniuiAKMgC0UHBwmN+mSnB3ST-LIKHJ5JCFCaiNIUSQebyTGR9AbDUZtCEdSYwBRmNaYDbIABMok8CkkVD6AGEIBRcgkoNBriAvGkACo43SQtIAX25QA)
```sh
--parser html
```

**Input:**
```html
<!doctype html>
<html>
  text
</html>
```

**Output:**
```html
<!DOCTYPE html>
<html>
  text
</html>

```

**Expected behavior:**
It'd be great if there were an option to disable transforming the html doctype tag into uppercase. Per [the spec](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype), the doctype is case insensitive - even [google.com](https://google.com) has its doctype declaration in lowercase. For normal HTML tags, prettier accepts all uppercase, e.g. `<DIV>` or all lowercase, e.g. `<div>`, however the doctype tag must be uppercase.

I would expect prettier not to treat the doctype differently from other HTML tags.

lydell:
Hi!

I’m not sure what you mean. Prettier normalizes the case of case insensitive stuff in HTML:

**Prettier 1.18.2**
[...]

PR #7391, opened 2020-01-18
Print HTML5 `doctype` in lowercase

fisker:
Please DO NOT MERGE until we have consensus that this should be accepted. 

Fixes: #6502


lipis:
I'm down!

thorn0:
It would be another incompatibility with XML. Why introduce it without a reason?

fisker:
@thorn0 https://github.com/prettier/prettier/issues/6502#issuecomment-533840708

lipis:
xml requires uppercase?

thorn0:
@lipis yep https://www.w3.org/TR/xml/#NT-doctypedecl

evilebottnawi:
I think we should discuss about supporting xhtml (xml)

thorn0:
How about lowercasing it only if it's the HTML5 doctype (`<!doctype html>`)?

evilebottnawi:
@thorn0 Problem what some developers can use not fully html template (with doctype) and we can potential break code

thorn0:
I mean lowercasing only in this one case, when doctype is `<!doctype html>`. If it contains something else (e.g. after html`), it should be uppercase.

thorn0:
As far as I understand, `<!DOCTYPE html>` is meaningless from the XML point of view, so lowercasing it should affect anything. Probably... :)

kydesnic:
I understand the concerns about merging this as is. This is the explicit change of the current behaviour.
But couldn't this be done with a new configuration option? If the default configuration value will produce the current result, the existing users will not be affected. But those, who want to change the behaviour, will be able to tune it.

fisker:
@kydesnic  We don't add options.

fisker:
Can we include this in V3?

thorn0 (approved):

JounQin:
AFAIK, prettier does not change the inputs but only prints, why it is changed here?

JounQin:
AFAIK, prettier does not change the inputs but only prints, why it is changed here?

JounQin (changes requested):
AFAIK, prettier does not change the inputs but only prints, why it is changed here?

fisker:
> prettier does not change the inputs but only print

We change html tags to lowercase.

> why it is changed here

Check original issue https://github.com/prettier/prettier/issues/6502

JounQin:
> We change html tags to lowercase.

So the correct change to me should be keeping the input as-is instead.

JounQin:
@fisker It seems disagree with my opinion here which follows https://prettier.io/docs/en/rationale.html#what-prettier-is-_not_-concerned-about and it is mentioned a lot of times by @thorn0 elsewhere.

Maybe I ask why we are not consistent here?

fisker:
I don't talk.

thorn0:
Prettier usually normalizes case-insensitive things: CSS property names, HTML tag names, JS hex numbers, etc. Often it's done by the parser. When something is syntactically case-insensitive, it means case is styling/formatting, which is what Prettier deals with. Of course, there aren't strict definitions of what "formatting", "linting", "transform", etc. are, so the debates about this stuff can be endless.

JounQin:
> Of course, there aren't strict definitions of what "formatting", "linting", "transform", etc. are, so the debates about this stuff can be endless.

@thorn0 

It makes sense to me. And then can we define what are considering as formatting into the policy on prettier's side in details? Otherwise it could still be confusing in the future.

For example, when the content is case-insensitive we usually print them in lowercase.

thorn0:
> can we define what are considering as formatting into the policy on prettier's side in details

I don't believe it's possible to come up with clear and satisfying definitions. It's all too complicated and full of tradeoffs. E.g., what are numeric separators in JS? Syntax or formatting? Are the two ways to make a code block in Markdown (fenced and indented) just two different styles for the same syntactic thing, so they should be normalized to one style? The general idea is kind of simple: "preserve the syntax and regenerate formatting from scratch (keeping some elements of the original formatting when practical)", but when it comes to defining the line between syntax and formatting, things become tricky. I don't think this boundary is formalizable. Decisions should be made on the case-by-case basis. 🤷‍♂️

> Otherwise it could still be confusing in the future

I'd say it gets confusing mostly in bikesheddy cases like this doctype issue, where the practical value of doing it one way or another is really questionable.

yoyo837:
https://github.com/prettier/prettier/issues/15096 Personally I want it to stay uppercase regardless of the standard, how do I control it from being converted to lowercase by prettier?

eligrey:
DOCTYPE is case sensitive in XML contexts. With this change, prettier no longer supports my XHTML.

eligrey:
Re: @thorn0 https://github.com/prettier/prettier/pull/7391#issuecomment-577099818

> As far as I understand, <!DOCTYPE html> is meaningless from the XML point of view anyway because it doesn't include a link to DTD. It's only [needed for the browser not to switch to a legacy mode](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype), and browsers don't care about its case. So lowercasing it should not affect anything. Probably... :)
> 
> But also there is (or was, it's "no longer in active maintenance") a thing called [Polyglot markup](https://www.w3.org/TR/html-polyglot/#dfn-polyglot-markup), a way to write markup compatible with both syntaxes at the same time, and it [requires](https://www.w3.org/TR/html-polyglot/#doctype) the doctype to be uppercase.

The ['polyglot' markup](https://blog.whatwg.org/xhtml5-in-a-nutshell) is and has been supported by all major browsers (even IE & EdgeHTML) for over a decade. It has rare use in some sites and frameworks (including many of my own).

yoyo837:
[Polyglot markup](https://www.w3.org/TR/html-polyglot/#dfn-polyglot-markup), doctype doc https://www.w3.org/TR/html-polyglot/#doctype Uppercase needed.

Rock-n-Roll-CRC:
Are W3C standards a joke to you? _If something works, why would you ever want to ruin it?_ Revert the changes concerning the `<!DOCTYPE html>` casing.

atesgoral:
DOCTYPE is not an HTML tag. It's a preamble. "We lowercase all HTML tags" shouldn't have applied here.

ProximaB:
How to change that behavior?

ptu14:
From: https://html.spec.whatwg.org/multipage/syntax.html#the-doctype

> A DOCTYPE must consist of the following components, in this order:
> 
> - A string that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "<!DOCTYPE".
> - One or more [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
> - A string that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "html".
> - Optionally, a [DOCTYPE legacy string](https://html.spec.whatwg.org/multipage/syntax.html#doctype-legacy-string).
> - Zero or more [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
> - A U+003E GREATER-THAN SIGN character (>).

> In other words, `<!DOCTYPE html>`, case-insensitively.

PR #7400, opened 2020-01-20
Enable ESLint `prefer-rest-params` rule

fisker:
Notice: Similar changes inside `fast-path.js` was merged and reverted.  ( #6225 and #6226 )

But I see no reason not use rest params (node 10 supports this), we already used it a lot.

### DIFF
--- a/src/common/fast-path.js
+++ b/src/common/fast-path.js
@@ -56,10 +56,7 @@ FastPath.prototype.getParentNode = function getParentNode(count) {
 // reference to this (modified) FastPath object. Note that the stack will
 // be restored to its original state after the callback is finished, so it
 // is probably a mistake to retain a reference to the path.
-FastPath.prototype.call = function call(
-  callback,
-  /* name1, name2, ... */ ...names
-) {
+FastPath.prototype.call = function call(callback, ...names) {
   const s = this.stack;
   const origLen = s.length;
   let value = s[origLen - 1];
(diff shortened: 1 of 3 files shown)


thorn0:
Do we really still need to transpile rest parameters? What for?

fisker:
`bin-prettier.js` keeps `rest params syntax`, `standalone.js` still transpile it.

https://dev.azure.com/prettier/prettier/_build/results?buildId=4387&view=artifacts&type=publishedArtifacts

thorn0:
But what for? For [obscure browsers](https://browserl.ist/?q=%3E0.25%25%2C+not+ie+11%2C+not+op_mini+all) like Safari 5 for Windows? Let's stop doing this, shall we? Something tells me Prettier doesn't work there anyway. We don't do any test in those browsers and don't really care.

fisker:
How about https://github.com/eslint/eslint/blob/1b8a3ce15237b9085f2761dcf73655207e6169a6/webpack.config.js#L37

thorn0:
`>0.5%, not ie 11, not safari 5.1, not op_mini all`

src/common/fast-path.js
```diff
@@ -83,14 +86,17 @@ FastPath.prototype.callParent = function callParent(callback, count) {
 // accessing this.getValue()[name1][name2]... should be array-like. The
 // callback will be called with a reference to this path object for each
 // element of the array.
-FastPath.prototype.each = function each(callback /*, name1, name2, ... */) {
+FastPath.prototype.each = function each(
+  callback,
+  /* name1, name2, ... */ ...names
```
thorn0: `/* name1, name2, ... */` - please remove such comments

thorn0:
There are also `Array.prototype.push/unshift.apply` and `[].push.apply` calls in the code base. Can you rewrite them too while you're at it?

src/common/fast-path.js
fisker: Done

PR #7402, opened 2020-01-20
Clarify API docs

jacksteamdev:
It is not immediately obvious how to set the language (`options.parser`), hopefully this will save others time.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -9,7 +9,7 @@ const prettier = require("prettier");
 
 ## `prettier.format(source [, options])`
 
-`format` is used to format text using Prettier. Set `options.parser` to the language you are formatting. [Options](options.md) may be provided to override the defaults. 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -9,7 +9,7 @@ const prettier = require("prettier");
 
 ## `prettier.format(source [, options])`
 
-`format` is used to format text using Prettier. [Options](options.md) may be provided to override the defaults.
+`format` is used to format text using Prettier. Set `options.parser` to the language you are formatting. [Options](options.md) may be provided to override the defaults.
```
thorn0:
```suggestion
`format` is used to format text using Prettier. [Options](options.md) may be provided to override the defaults. Set `options.parser` according to the language you are formatting (see the [list of available parsers](options.md#parser)).
```

PR #7432, opened 2020-01-25
Use `ci-info` instead of `is-ci`

fisker:
Use `ci-info` directly.

https://github.com/watson/is-ci/blob/master/index.js#L3


evilebottnawi:
Looks coverage was not sent

fisker:
Yes, coverage sent https://codecov.io/gh/prettier/prettier/compare/93f71d499af7038bbb6f80bb132dcdc17b6fb885...d55836f0aa9b7d6fd41338efefeafad8c771b9f3/commits

just no diff

PR #7431, opened 2020-01-25
Followup work on setup github actions

fisker:
Todo:

- [x] blocker https://github.com/prettier/prettier/pull/7432 & https://github.com/prettier/prettier/pull/7433
- [x] Coverage report
- [x] Standalone tests

#7321

PR #7433, opened 2020-01-25
Always enable `ENABLE_TEST_RESULTS` and `AST_COMPARE` on CI

fisker:
We don't have to setup those variables on CI. And makes https://github.com/prettier/prettier/pull/7431 simplier.


lipis:
Less is more :)

PR #7435, opened 2020-01-26
Change `endOfLine` to `lf` by default for 2.0

### LINKED ISSUE #6932: Change `endOfLine` to `lf` by default for 2.0
Refs: #6888 (about Prettier 2.0), #6929 (about changing `arrowParens` to `always`)

[`End of Line option`](https://prettier.io/docs/en/options.html#end-of-line) was added in https://github.com/prettier/prettier/pull/5327 and is available since Prettier 1.15. Given that using LF (Linux line endings) is a default choice in the software dev world, I decided to open this issue to bring the topic to a broader discussion.

I understand that changing defaults can cause large git diffs in some projects. However for the benefit of newly created projects, I suggest that the problem can be addressed by changing `prettier.config.js` in existing codebases. V2.0 is a great opportunity for breaking changes like this.

WDYT folks?

kachkaev:
Closes #6932


changelog_unreleased/api/pr-7435.md
```diff
@@ -0,0 +1,3 @@
+#### Change default value for `endOfLine` to `lf` ([#7435](https://github.com/prettier/prettier/pull/7435) by [@kachkaev](https://github.com/kachkaev))
+
+<!-- Work in progress -->
```
kachkaev: TODO

kachkaev: ✅

PR #7439, opened 2020-01-26
Simplify `createUsage`

fisker:

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -586,12 +586,16 @@ function createUsage(context) {
   );
 
   const groupedOptions = groupBy(options, option => option.category);
+
+  const firstCategories = constant.categoryOrder.slice(0, -1);
+  const lastCategories = constant.categoryOrder.slice(-1);
+  const restCategories = Object.keys(groupedOptions).filter(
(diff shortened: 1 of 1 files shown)


src/cli/util.js
```diff
@@ -586,14 +586,13 @@ function createUsage(context) {
...
-      !firstCategories.includes(category) && !lastCategories.includes(category)
-  );
-  const allCategories = firstCategories.concat(restCategories, lastCategories);
+  const allCategories = [
+    ...constant.categoryOrder.slice(0, -1),
+    ...Object.keys(groupedOptions).filter(
+      category => !constant.categoryOrder.includes(category)
+    ),
+    ...constant.categoryOrder.slice(-1)
+  ];
```
evilebottnawi: Previous version will be more readable, just my opinion

fisker: Main change in this PR is change `!firstCategories.includes(category) && !lastCategories.includes(category)` to `!constant.categoryOrder.includes(category)` they are equal. no need test twice.

fisker: @evilebottnawi How about now?

PR #7465, opened 2020-01-29
Remove `normalize-path` package

fisker:


thorn0:
This breaks in the browser:
```console
> prettier.format("a < b > ( c )", { plugins: prettierPlugins, filepath: 'foo.ts' });
19:46:09.835 standalone.js:13126 Uncaught TypeError: path.basename is not a function
    at inferParser (standalone.js:13126)
    at normalize$1 (standalone.js:13028)
    at formatWithCursor (standalone.js:16704)
    at standalone.js:31935
    at Object.format (standalone.js:31944)
    at <anonymous>:1:10
inferParser @ standalone.js:13126
normalize$1 @ standalone.js:13028
formatWithCursor @ standalone.js:16704
(anonymous) @ standalone.js:31935
format @ standalone.js:31944
(anonymous) @ VM36:1
```

fisker:
@thorn0 how to do this test ?

fisker:
Oh. script include standalone.js ?

thorn0:
https://jsbin.com/giqiroy/edit?html,js,console

fisker:
I think `filepath.split(/[/\\]/).pop()` should work.

thorn0:
That should work. I wonder why the tests didn't catch this. There is an Azure job configured to run the tests with `TEST_STANDALONE=true`, but it didn't catch this.

fisker:
I'll figure

thorn0:
I found why: `parser` is passed explicitly in all those tests.

fisker:
@thorn0 `STANDALONE` only test `tests/`
 https://github.com/prettier/prettier/blob/6a497b691440dd6ce8b713fec1326a2b6a23e3e6/scripts/test-dist.js#L24

thorn0:
Right. And `parser` is explicitly specified in all those tests.

fisker:
This function is `inferParser`, and all tests in `tests/` has options `parser`. so we don't have to look for parser, right?

fisker:
I'm trying to add a `no-parser` test

fisker:
Let's see if https://github.com/prettier/prettier/pull/7465/commits/7a9fb9fc46a03b451c6ffb10e6e219e881551f89 can catch this error

fisker:
Error caught https://dev.azure.com/prettier/prettier/_build/results?buildId=4915&view=logs&j=cc0821cc-6ec9-5a14-8bb9-4961b231143e&t=0577ecdd-8efd-50b6-992b-f5e447724f2c&l=1959

tests_config/run_spec.js
```diff
@@ -15,8 +15,11 @@ const prettier = !TEST_STANDALONE
...
 
 global.run_spec = (dirname, parsers, options) => {
-  // istanbul ignore next
-  if (!parsers || !parsers.length) {
+  // `isNoParserTests` mean to test `inferParser` on `standalone`
+  const isNoParserTests = dirname.includes("no-parser-tests");
+  if (isNoParserTests) {
+    parsers = [];
+  } else if (!parsers || !parsers.length) {
     throw new Error(`No parsers were specified for ${dirname}`);
```
thorn0: Is this check really needed? I think we can remove it.

fisker: Maybe new tests forgot add parsers, I think we should keep it

fisker: Added in https://github.com/prettier/prettier/pull/3356 , let's keep it. but `// istanbul ignore next` can be removed...

scripts/build/shims/path.js
```diff
@@ -1,9 +1,16 @@
...
 export function extname(path) {
+  path = path.split(sep).pop();
   const dotIndex = path.lastIndexOf(".");
   if (dotIndex === -1) return "";
   return path.slice(dotIndex);
 }
 
+export function basename(path) {
+  return path.split(sep).pop();
+}
```
fisker: I'm going to fix with this shim.

scripts/build/shims/path.js
```diff
@@ -1,9 +1,16 @@
+const sep = /[\\/]/;
+
 export function extname(path) {
+  path = path.split(sep).pop();
```
fisker: this fixes `/a.b/c`

tests_config/run_spec.js
thorn0: What's wrong with allowing Prettier to infer the parser if it's what the author of the test wants? I guess this check was added when the parser inference wasn't as good as it is today.

fisker: I guess, if someone want test 'foo.js' on `flow` but forgot add parser, the default parser `babel` will work, and a little hard to find the mistake?

tests_config/run_spec.js
```diff
@@ -15,8 +15,11 @@ const prettier = !TEST_STANDALONE
   : require("prettier/standalone");
 
 global.run_spec = (dirname, parsers, options) => {
-  // istanbul ignore next
-  if (!parsers || !parsers.length) {
+  // `isNoParserTests` mean to test `inferParser` on `standalone`
+  const isNoParserTests = dirname.includes("no-parser-tests");
```
thorn0: Let's rename the folder to `parser-inference` to better describe what is actually tested.

fisker: sure

fisker:
> I see `eval("require").resolve` in multiple places, could we factor this out somehow?

Suggestion?

brodybits:
> > I see `eval("require").resolve` in multiple places, could we factor this out somehow?
> 
> Suggestion?

<https://github.com/fisker/prettier/pull/300>

A side point is that we could do something similar for eval("require")(...) in `src/common/internal-plugins.js` and `src/cli/util.js`.

fisker:
@brodybits if we want do this, maybe better `const xx = eval('require')` some where

See https://github.com/prettier/prettier/blob/master/src/common/internal-plugins.js

brodybits:
> `const xx = eval('require')` some where
> 
> See https://github.com/prettier/prettier/blob/master/src/common/internal-plugins.js

Yes. I am starting to wonder if we should do this in a separate (followup) PR and factor more eval calls out of that module at the same time?

fisker:
I'm not suggesting to do this way, I think `eval('require')` is good, can't find a better solution, as we may don't need it someday. Also `eval('require').resolve` should be good.

fisker:
> > `const xx = eval('require')` some where
> > See https://github.com/prettier/prettier/blob/master/src/common/internal-plugins.js [](/prettier/prettier/blob/HEAD@{2020-02-02T12:16:17Z}/src/common/internal-plugins.js)
> 
> Yes. I am starting to wonder if we should do this in a separate (followup) PR and factor more eval calls out of that module at the same time?

I don't think we should do anything, it's a hack. It works just like `require/*please-keep-me-as-it-is*/(foo)`

brodybits:
Please disregard my comments on eval, they were intended for another PR. I will raise a new PR to propose what I would like to do.

PR #7466, opened 2020-01-29
Change default of singleQuote to true

### LINKED ISSUE #4102: Change the default value for singleQuote to true
From https://github.com/prettier/prettier/issues/3503#issuecomment-359863688

karlhorky:
Change default of singleQuote to `true`.

Closes #4102

## Why?

Single quotes are more common. This will make Prettier more zero-config for the majority.

Additionally:

* make it easier for beginners to compare their code with examples
* votes in #4102 (👍 124, 👎 28)
* a reference to style guides and popular open-source repos
* many other sources from [my review of the ecosystem](https://github.com/prettier/prettier/issues/4102#issuecomment-579912268)
* the original decision was made based on a single person's opinion
* [191 likes in 18 hours on Twitter](https://twitter.com/karlhorky/status/1222540916608700416)
* supporting Prettier members below

### Votes from Prettier Collaborators / Members:

👍 @j-f1 
👍 @lipis 
👍 @evilebottnawi 
👍 @sosukesuzuki

👎 @lydell
👎 @Shinigami92


karlhorky:
Uh, this is my first PR that changes the defaults... I'm not sure that I did everything correctly. It seems like there aren't a lot of changes to the snapshots, which I was expecting when I ran:

```sh
yarn test -u
yarn lint --fix
```

karlhorky:
Ok I think I figured it out - I changed the wrong options file.

lipis:
You should target the `next` branch.. and it's better to checkout to that branch first..

lipis:
try

```
git pull https://github.com/prettier/prettier.git next
```

Merge it and push back to your branch..

lipis:
Otherwise branch out from next do your changes and create a new PR :)

karlhorky:
Yep, just cherry-picked my changes to a new branch, branched from `next`.

Just updating the snapshots now...

docs/rationale.md
```diff
@@ -13,7 +13,7 @@ The first requirement of Prettier is to output valid code that has the exact sam
 
 ### Strings
 
-Double or single quotes? Prettier chooses the one which results in the fewest number of escapes. `"It's gettin' better!"`, not `'It\'s gettin\' better!'`. In case of a tie, Prettier defaults to double quotes (but that can be changed via the [`--single-quote`](options.html#quotes) option).
+Double or single quotes? Prettier follows the JavaScript ecosystem, which has largely standardized on single quotes, with it being the most popular style seen in the wild.
```
thorn0: The 'fewest number of escapes' part shouldn't get removed. Please keep it somehow.

karlhorky: Ah, sorry! I misunderstood that this was describing a separate feature. I've added this back.

thorn0:
500+ files changed. This PR is going to be really difficult to review.

karlhorky:
I think a lot of it can be quickly checked (maybe even partially automated), since the vast majority is in the snapshots.

thorn0:
The problem is it's easy to overlook something important in this batch of trivial changes.

karlhorky:
Apart from that, I would like to update this PR to avoid changing the CSS rules (CSS is more commonly written with double quotes).

I tried doing the following (in `src/language-css/options.js`), but it doesn't seem to make a change when I update the snapshots again with `yarn jest -u --testPathPattern=css`:

```diff
 "use strict";
 
-const commonOptions = require("../common/common-options");
+const CATEGORY_COMMON = "Common";
 
 // format based on https://github.com/prettier/prettier/blob/master/src/main/core-options.js
 module.exports = {
-  singleQuote: commonOptions.singleQuote
+  singleQuote: {
+    since: "0.0.0",
+    category: CATEGORY_COMMON,
+    type: "boolean",
+    default: false,
+    description: "Use single quotes instead of double quotes."
+  }
};
```

kachkaev:
> Apart from that, I would like to update this PR to avoid changing the CSS rules (CSS is more commonly written with double quotes).

I believe that many teams using Prettier today with `{ "singleQuote": true }` also like their CSS files using single quotes as well. If the option no longer applies to CSS, this will be a breaking change with no mitigation.

fisker:
I have a idea, we can make a temp snapshot serializer , swap `'` and `"` so we can make review esaier. after review we restore it back and update snapshot to real one.

karlhorky:
> I believe that many teams using Prettier today with `{ "singleQuote": true }` also like their CSS files using single quotes as well.

What is this assumption based on? I think Prettier should format to the common style for the language.

Those teams that would want `singleQuote: true` in their CSS config would still have an option - to use [an override](https://prettier.io/docs/en/configuration.html#configuration-overrides):

```json
{
  "overrides": [
    {
      "files": "*.css",
      "options": {
        "singleQuote": true
      }
    }
  ]
}
```

karlhorky:
> a temp snapshot serializer , swap ' and " so we can make review esaier.

Oh wow, that sounds much better than the automation that I was thinking of (I was thinking of modifying the git diff to remove all snapshots that now have `'` instead of `"`).

thorn0:
@karlhorky Would `overrides` be required? Wouldn't simple top-level `singleQuote: true` suffice?

karlhorky:
> Wouldn't simple top-level `singleQuote: true` suffice to override the defaults for all languages?

Oh right, yeah I guess so! Will edit that in my comment above.

lydell:
> Apart from that, I would like to update this PR to avoid changing the CSS rules (CSS is more commonly written with double quotes).

Isn’t that an argument against changing the default? Sounds easier to have double quotes as the default for everything. Why complicate things?

karlhorky:
> Isn’t that an argument against changing the default?
> Why complicate things?
> How would you motivate not preferring single quotes in some languages? 

I think these three questions all have the same answer, which is also the explanation in the pull request:

> The JavaScript ecosystem has largely standardized on single quotes, with it being the most popular style seen in the wild.

https://github.com/prettier/prettier/pull/7466/files#diff-6afbd7a88bf7df26bba2e36de87663ffR15-R16

It's the most common, expected, survey-winning style seen in the JavaScript ecosystem.

---

Then again, playing the devil's advocate for CSS, I don't have as much experience with whether this is a big deal. Quotes are not as common in CSS, so maybe that's one reason that it doesn't matter as much as keeping things consistent.

ExE-Boss:
Because of&nbsp;<https://github.com/prettier/prettier/issues/4102#issuecomment-579842398>, I&nbsp;vote to&nbsp;defer to&nbsp;3.0 or&nbsp;later.

changelog_unreleased/api/pr-7466.md
```diff
@@ -0,0 +1,18 @@
...
+```js
+// Input
+const foo = 'bar';
+// Prettier stable
+const foo = "bar";
+// Prettier master
+const foo = 'bar';
+```
+
+The JavaScript ecosystem has largely standardized on single quotes, with it being the most popular style seen in the wild.
```
lydell: People are going to ask for sources on this.

karlhorky: Hm, interesting perspective - I think they won't. I think that most JavaScript developers have had this experience by trolling through documentation and examples from all over the place.

But it's not about your opinion or my opinion, so maybe we can get some statistics together on this!

Some sources:

- GitHub usage
- public documentation of large JavaScript projects
- public blog posts
- developer surveys (@mvolkmann here: https://github.com/prettier/prettier/issues/4102#issuecomment-370476135)
- style guides

karlhorky: But it's not about your opinion or my opinion, so maybe we can get some statistics together on this!

Some sources:

GitHub usage
public documentation of large JavaScript projects
public blog posts
developer surveys (@mvolkmann here: https://github.com/prettier/prettier/issues/4102#issuecomment-370476135)
style guides

karlhorky: I will ask for some help in #4102.

lydell: > I think that most JavaScript developers have already had this experience of single quotes being the norm by trolling through documentation

I write JS every day, and I honestly have no clue what’s the norm.

nickmccurdy:
JSON itself, JSON.stringify, and JSX all use double quotes. I don't think single quotes would promote consistency, as not enough JS developers use them.

changelog_unreleased/api/pr-7466.md
karlhorky: Interesting take, maybe you just don't notice which quotes there are? Because I do, and I read about JavaScript from a ton of sources, from beginner level all the way up to spec level. I notice when I see double quotes, because it's so rare.

karlhorky: @lydell Just finished a tour of many JavaScript library documentation pages, so go take a look over there: https://github.com/prettier/prettier/issues/4102#issuecomment-579912268

karlhorky: Will continue my research (and get some other people to help too). Single quotes are almost everywhere, so collecting them isn't tough to do.

karlhorky:
> JSX

@nickmccurdy JSX is based on the visual appearance of HTML, which almost never has single quotes. So it's another language.

> JSON itself, JSON.stringify

This is just one thing - JSON. Which, albeit related to JavaScript in many ways, is a different format (not even a language), meant to interop with many other languages. JavaScript is different enough to warrant a different decision.

> not enough JS developers use them

Super interesting to hear that from someone - never heard this perspective before 1v1! Can you shed some light on where you get this from?

I have just finished a survey of many documentation pages of popular JavaScript libraries, maybe take a look. Single quotes almost everywhere: https://github.com/prettier/prettier/issues/4102#issuecomment-579912268

nickmccurdy:
> > JSX
> 
> @nickmccurdy JSX is based on the visual appearance of HTML, which almost never has single quotes. So it's another language.

Still, Prettier should stay consistent with how it formats across file. Prettier uses double quotes in HTML and JSON files because they're required/recommended, so they should also be used in JS files. Prettier wouldn't use different indent levels in different languages when given the same config.

> > JSON itself, JSON.stringify
> 
> This is just one thing - JSON. Which, albeit related to JavaScript in many ways, is a different format (not even a language), meant to interop with many other languages. JavaScript is different enough to warrant a different decision.

The history of JSON used to be that it was very similar to JS, but now it's becoming a literal subset of modern ES specs. As a result, we can think of ECMAScript as being an extension of the JSON spec that adds scripting on the data.

> > not enough JS developers use them
> 
> Super interesting to hear that from someone - never heard this perspective before 1v1! Can you shed some light on where you get this from?
> 
> I have just finished a survey of many documentation pages of popular JavaScript libraries, maybe take a look. Single quotes almost everywhere: [#4102 (comment)](https://github.com/prettier/prettier/issues/4102#issuecomment-579912268)

Most of those packages are older packages for Node, which traditionally has more single quote usage relatively. On the other hand, most JavaScript usage is on the web with double quotes.

karlhorky:
> Still, Prettier should stay consistent with how it formats across file

This is exactly what I'm arguing against.

I believe a formatter for any language should do the most common, expected thing for the language. The language in question is JavaScript, no other language.

> Most of those packages are older packages for Node

As I mentioned in my methodology over there in that comment, I included both older and newer popular libraries.

> On the other hand, most JavaScript usage is on the web with double quotes.

Can you provide examples of these libraries with their public documentation with double quotes?

changelog_unreleased/api/pr-7466.md
styfle: Microsoft is really the only counter-example I can think of. They use doubled quotes.

The typescript source code: https://github.com/microsoft/TypeScript/blob/master/src/tsserver/server.ts

The typescript documentation: https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html

docs/rationale.md
```diff
@@ -13,7 +13,7 @@ The first requirement of Prettier is to output valid code that has the exact sam
 
 ### Strings
 
-Double or single quotes? Prettier chooses the one which results in the fewest number of escapes. `"It's gettin' better!"`, not `'It\'s gettin\' better!'`. In case of a tie, Prettier defaults to double quotes (but that can be changed via the [`--single-quote`](options.html#quotes) option).
+Double or single quotes? Prettier formats to the one which results in a fewest number of escapes. `'We call it "Prettier".'`, not `"We call it \"Prettier\"."`. In case of a tie, Prettier defaults to single quotes (but that can be changed via the [`--single-quote`](options.html#quotes) option).
```
Shinigami92: `results in a fewest number` sound wrong to me 🤔 
IMO it should stay as `results in the fewest number`

changelog_unreleased/api/pr-7466.md
karlhorky: @styfle Thanks! I've added this to the list.

docs/rationale.md
karlhorky: Thanks, fixed! This was an artifact from a previous wording that I changed back (but not completely!).

pluma:
Wait, how does this affect JSX? There was a huge kerfuffle when StandardJS started switching JSX literals to single quotes so I expect this PR doesn't do that, but will JS code in JSX files use single quotes too?

And what about Typescript? Most TS code I've seen uses double quotes. But having inconsistent formatting between TS and JS would be a little weird.

karlhorky:
> I expect this PR doesn't [switch to single for JSX]

Yes, I expect that it doesn't either. And it seems that it doesn't, in [the playground](https://deploy-preview-7466--prettier.netlify.com/playground/)

> Typescript

TypeScript will also be formatted with singles. It's a superset of JS, so same rules as JS.

The Microsoft docs use double quotes - I've documented that in [my ecosystem review](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html). Is there anywhere else that uses doubles? I seem to recall seeing singles a lot with TypeScript code.

> I also expect this PR won't affect CSS in template strings in JS/JSX/TS/TSX files because double quotes are the norm for CSS?

I would like to retain double quotes default for CSS, if possible: https://github.com/prettier/prettier/pull/7466#issuecomment-579783887

phpnode:
imo this PR is not worth the disruption, people who like single quotes already have that option, why introduce this churn? I see the argument about the popularity of single quotes in JS, and maybe if prettier was brand new and didn't have the level of adoption it already has then this change would make a lot of sense, but this feels like "principle over pragmatism" to me. All this change will do is upset users.

kachkaev:
At the moment, a need for the change is justified by:
* votes in #4102
* a reference to style guides and popular open-source repos (some of which were created many years ago)
* a list of supporting Prettier maintainers in the first comment to this PR

To sum up, the only persuasion methods currently used are ‘social proof’ and ‘authority’. Such methodology is quite shallow, which means that there is a chance of making a harmful decision in the end. I [suggest doing a proper research](https://github.com/prettier/prettier/issues/4102#issuecomment-580138779) before changing the default value.

karlhorky:
@kachkaev you're missing out a lot of reasons. Really feels like you're operating in bad faith here, just based on your personal opinion:

- make it more zero-config for the majority
- make it easier for beginners to compare their code with examples
- the original decision was made based on a single person's opinion
- many other sources from [my review of the ecosystem](https://github.com/prettier/prettier/issues/4102#issuecomment-579912268)

karlhorky:
> this PR is not worth the disruption
> All this change will do is upset users

@phpnode This PR is meant for a major version bump, which in semver means breaking changes.

Luckily, for those who oppose the formatting changes to their codebase, there is still the option of writing a single line of configuration.

Not very much disruption at all, especially for a major version bump.

phpnode:
The point is that people who want to maintain the status quo (which is likely to be most people) will now need a configuration file when they didn't before. Most people are likely using prettier with default config, including in IDEs etc. It's making busy-work for thousands of devs for extremely questionable benefit. 

It's not about whether single quotes are more prevalent in the eco-system, it's about whether churning something like this is worth it when the prettier user base is so large.

karlhorky:
> Most people are likely using prettier with default config

Is this just your opinion? Or do you have any data on that?

phpnode:
Of course I don't have any data on that, and I don't see an acceptable way for anyone to collect that data, but it makes sense, no? It's an opinionated code formatter which works out of the box, I think it's very very likely that most people won't have bothered to configure it

pluma:
I for one welcome our new single quote overlords even if single quotes are normally annoying to type on a QWERTZ keyboard layout (both require chords but single quotes are on the right hand side).

Let's not kid ourselves about the public vote here though. Twitter likes are irreverent and ~200 GitHub reactions aren't representative for a project with this many installs.

The real reason isn't the votes but that this has been a long-standing wish within the prettier dev team based on the widespread preference within the JS open source ecosystem. That is absolutely legit but I'd prefer we stick to that rationale rather than democracy-wash what is clearly not intended (nor required) to be an opinion poll or popularity contest.

lydell:
> I for one welcome our new single quote overlords even if single quotes are normally annoying to type on a QWERTZ keyboard layout (both require chords but single quotes are on the right hand side).

That’s irrelevant want using Prettier. You can type whichever quote is easier and Prettier will fix them for you.

peterjuras:
How will this behave in IDEs like VSCode? Right now I can copy and paste a JavaScript object /array from a JS file into a JSON file and the prettier extension will format it to make it JSON compatible.

Will this still work 100% of the times? Or will it change everything to single quotes when the file format is not 100% clear, thus breaking JSON?

I can understand breaking changes that reformat long lines at different breakpoints for higher (supposed) readability. But this change doesn't make sense to me since it's just an opinionated change which could be reverted in a few years because double quotes might get cool again.

I don't want to use or introduce a configuration file with prettier. Keeping it in sync across multiple repositories in a team is a pain. Therefore the argument of "you can easily configure this" is not working for me, it would rather completely invalidate this PR because if this would be the case then you would simply apply a single quite configuration to your repos and would have no need to change the default style.

I'm indifferent about whether this makes into the core of prettier our not, but in case that I do feel any downgrade in editor experience when dealing with JSON, it would leave a really bad impression about prettier for me.

pgsandstrom:
@peterjuras Prettier can format a json-file using single quote to using double quote. So there should be no problem copy-pasting javascript-objects to json-files.

glen-84:
Please consider @Shinigami92's [suggestion](https://github.com/prettier/prettier/issues/4102#issuecomment-580127727) before making such a drastic change.

I won't argue about single quotes being more common in JavaScript, but Prettier is used with many other languages as well. The default should not be based on what is most common in JavaScript, but by what is most common across _all_ languages.

A probably-common case of JavaScript (single), HTML (double), and CSS (double), would now require an override with 2 file extensions instead of 1. I can't imagine someone using single quotes for HTML, so even if you only needed that one override, you'd just be swapping `*.js` for `*.html` – what does that achieve?

changelog_unreleased/api/pr-7466.md
styfle: @karlhorky Maybe you could move the content of that comment into the original post because github auto-collapses comments in the middle so its unlikely someone will see that who is just finding this PR (like myself).

karlhorky: Good idea. I moved it yesterday already actually, under the link text "my review of the ecosystem"

thorn0:
I unintentionally closed this and several other PRs because I deleted the `next` branch, but let's keep this one closed for now.

karlhorky:
@thorn0 @lipis @j-f1 @evilebottnawi Should this be rebased on the 2.0 branch now and a new PR opened?

I think it probably shouldn't just stay closed just because it was accidentally closed...

wouterds:
> I think it probably shouldn't just stay closed just because it was accidentally closed...
I agree.

thorn0:
> Should this be rebased on the 2.0 branch now and a new PR opened?

Only if there is a final decision about changing this default, I guess. Otherwise, what for?

karlhorky:
> Only if there is a final decision about changing this default, I guess. Otherwise, what for?

To have a PR that can be easily merged if and when the decision is made to merge (for example, for 3.0).

Even though more than a few people and their friends / colleagues supporting double quotes suddenly appeared in this issue to make this into a heated discussion, it seems like given the numbers that they are the vocal minority - which seems to me should not be the reason that Prettier decides to do or not do something.

I still feel that given the research that has been already done, this change to a default single quote would better fit JavaScript and benefit the majority of developers who write it.

Every time a poll has been done, the majority supports this choice.

If the research is required for 3.0 that a larger, more wide reaching and impartial poll or survey of existing code be done (which I believe is overblown for a default on a configurable option), then I'm sure that the numbers would also tell the same story as they have up until now.

karlhorky:
Even though more than a few people and their friends / colleagues supporting double quotes suddenly appeared in this issue to make this into a heated discussion, it seems like given the numbers that they are the vocal minority - which seems to me should not be the reason that Prettier decides to do or not do something.

I still feel that given the research that has been already done, this change to a default single quote would better fit JavaScript and benefit the majority of developers who write it.

Every time a poll has been done, the majority supports this choice.

If the research is required for 3.0 that a larger, more wide reaching and impartial poll or survey of existing code be done (which I believe is overblown for a default on a configurable option), then I'm sure that the numbers would also tell the same story as they have up until now.

Issue #5810, opened 2019-01-28
Empty lines unexpectedly added after table tags


thorn0:
**Prettier 1.16.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeGBDARgGzgPgB0oACE9AJ31QHoYrjbNcDGATASwDdqbOeQAGhAQADjA7QAzslAYKFCAHcACvIQyUGLhA5shILBQxgA1nBgBlUSY5QA5snoBXOMIAWMALY4A6u454KRswOEsNQO5AgE9kcCkZYTspOAoYFWN7LwxkADMMHBThACspAA8AIWMzC0sMLzgAGTs4PIKikFKyyzt7PABFZwh4NsK3EBsKFIo4zx8DUQo7GF89GHdkAA4ABmFFiBTfY1E4xbhprlbhCjgAR2cOG4yMLJykfLHhFK8OJwpXL69AZDEbvdrjZirNjrZAAJmE9AwHBwvQAwhAvNk4lBoFcQM4UgAVbCaD4pAC+5KAA)
```sh
--parser html
```

**Input:**
```html
<table>
  <tr></tr>
</table>
<div></div>
```

**Output:**
```html
<table>
  <tr></tr>
</table>

<div></div>

```

**Expected behavior:**
```html
<table>
  <tr></tr>
</table>
<div></div>

```

Happens only if `tbody` tags are omitted.

ikatyang:
The issue here is that the position of `</table>` was eaten by the pseudo node `<tbody>`.

ikatyang:
Pseudo `<tbody>` for `<tr>` is removed in [angular-html-parser@1.4.0](https://github.com/ikatyang/angular-html-parser/blob/master/packages/angular-html-parser/CHANGELOG.md#140-2020-01-28), updating the parser should fix the issue.

thorn0:
closed by #7461

thorn0:
Seems like some cleanup is needed here, see https://codecov.io/gh/prettier/prettier/compare/07b9465dcfeb7228eef9dff5a959dea645854697...f40fcba710f7a0d4e82e5458334dca97d994c94a/src/src/language-html/parser-html.js

thorn0:
closed by #7468

PR #7468, opened 2020-01-29
Cleanup dead code from pr7461

fisker:
Close: #5810


thorn0:
I've already added a test: `html_basics/tr-without-tbody-5810.html`, but `html_tags` is seems to be a better location for it.

fisker:
@thorn0 I didn't notice that.

thorn0:
Feel free to (re)move that test.

What about the line 76 in `src/language-html/parser-html.js`? Its coverage became partial, which means `node.nameSpan` is always truthy now. Please clean it up too.

PR #7469, opened 2020-01-29
glimmer: unpin simple-html-tokenizer

thorn0:
🚧 Don't merge. 🚧

For now, this PR exists just to have the playground where the consequences of this change can be explored.


dcyriller:
Might be superseeded by https://github.com/prettier/prettier/pull/7245

PR #7483, opened 2020-01-30
refactoring: Extract isGlimmerComponent to an utils file

dcyriller:
#### Description

This is a refactoring PR for glimmer printer. I want to use this util in a coming PR. But on the way, I'll have quite a few other functions to extract this utils file.

My plan is to submit one PR for this small change, another one to extract some other utils and last one using `isGlimmerComponent` util.

I don't know if you prefer to review small chunks one by one or if you prefer only one PR extracting all the things. Feel free to tell me.

#### Checklist


src/language-handlebars/utils.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+function isGlimmerComponent(node) {
+  if (node.type !== "ElementNode") {
+    return false;
+  }
+
+  const tagFirstChar = node.tag && node.tag[0];
+  const isLocal = node.tag.indexOf(".") !== -1;
```
fisker: This will be a eslint error, when merge `master` into `next`,  how about commit to `next`

dcyriller: Out of curiosity: why https://github.com/prettier/prettier/pull/6967 did not target master?

fisker: https://github.com/prettier/prettier/pull/6968#issuecomment-554629174

src/language-handlebars/utils.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+function isGlimmerComponent(node) {
+  if (node.type !== "ElementNode") {
+    return false;
+  }
+
+  const tagFirstChar = node.tag && node.tag[0];
```
fisker: This part `node.tag && ` seems useless, you use `node.tag` bellow

Issue #7229, opened 2020-01-06
API: prettier.format reloads plugins on every call


velut:
**Description**

Calling the `prettier.format` Node.js API causes plugins to be reloaded every time, even if the formatting options remain the same.
This appears to slow down formatting, since plugin directories are (seemingly) accessed synchronously each time.

**Environments:**
- Prettier Version: `1.19.1`
- Running Prettier via: `Node.js API`
- Runtime: `Node.js v12.14.0`
- Operating System: `Windows`

**Steps to reproduce:**

Call `prettier.format` API multiple times as seen in the following example.

```js
// main.js
//
// Flamegraph command: clinic flame -- node main.js

const prettier = require('prettier');

// Options are fixed
const options = {
    semi: true,
    singleQuote: true,
    trailingComma: 'es5',
    tabWidth: 4,
    printWidth: 60,
    endOfLine: 'lf',
    arrowParens: 'always',
    parser: 'typescript',
};

// Three differently named functions to show different call stacks
function foo() {
    prettier.format('', options);
}

function bar() {
    prettier.format('', options);
}

function baz() {
    prettier.format('', options);
}

// Repeatedly calling `prettier.format` with the same options
// causes plugins to be constantly reloaded from directories
// as seen in the flamegraph below.
const max = 100;
for (let i = 0; i < max; i++) {
    foo();
    bar();
    baz();
}
```

**Expected behavior:**

It should be possible to load plugins (if necessary) only once for a given set of options and then use the returned formatter to format different input texts.

**Actual behavior:**

Plugins are reloaded from directories with each call to `prettier.format` as seen in [this flamegraph](https://upload.clinicjs.org/public/66c03b913bf6053ddb6dbffc4181266671f0e98adf6a45655731583d159067b7/16704.clinic-flame.html)

velut:
I'm sharing a temporary workaround to improve performance for API users.

Using [patch-package](https://www.npmjs.com/package/patch-package) and applying this patch

```diff
diff --git a/node_modules/prettier/index.js b/node_modules/prettier/index.js
index 96cfec8..d455497 100644
--- a/node_modules/prettier/index.js
+++ b/node_modules/prettier/index.js
@@ -44202,7 +44202,8 @@ function _withPlugins(fn) {
     var args = Array.from(arguments);
     var opts = args[1] || {};
     args[1] = Object.assign({}, opts, {
-      plugins: loadPlugins_1(opts.plugins, opts.pluginSearchDirs)
[...]

PR #7485, opened 2020-01-30
Memoize `loadPlugins`

fisker:
Fixes: #7229


thorn0:
It's "[memoize](https://en.wiktionary.org/wiki/memoize)", without R.

thorn0:
Not sure we want to make the API more complex by adding a separate function for clearing that cache. Let's clear it in `prettier.clearConfigCache`. If someone needs more granularity, we can add it later.

fisker:
@thorn0 need rename?

thorn0:
Not rename. There already exist the function [`clearConfigCache`](https://prettier.io/docs/en/api.html#prettierclearconfigcache). I suggest that it clear both caches:
```js
clearConfigCache: () => {
  config.clearCache();
  plugins.clear(); // BTW, let's rename this one to `clearCache`
},
```

docs/api.md
```diff
@@ -81,7 +81,7 @@ Use `prettier.resolveConfigFile.sync([filePath])` if you'd like to use sync vers
 
 ## `prettier.clearConfigCache()`
 
-As you repeatedly call `resolveConfig`, the file system structure will be cached for performance. This function will clear the cache. Generally this is only needed for editor integrations that know that the file system has changed since the last format took place.
+When searching Prettier configuration file and Prettier plugins, the file system structure will be cached for performance. This function will clear the cache. Generally this is only needed for editor integrations that know that the file system has changed since the last format took place.
```
thorn0:
```suggestion
When Prettier loads configuration files and plugins, the file system structure is cached for performance. This function will clear the cache. Generally this is only needed for editor integrations that know that the file system has changed since the last format took place.
```

fisker:
@thorn0 Last time you ask me put  `clearPluginsCache` in `clearConfigCache`, I asked 

> need rename?

I mean do we need rename `clearConfigCache` , plugins cache seems not part of config?

thorn0:
Plugins are kind of configuration too, so I don't think that function needs renaming.

fisker:
I was thinking `Prettier.clearCache()`.

fisker:
Like  this  reverted `--only-changed` feature, https://github.com/prettier/prettier/pull/5910 we  can  also  clean that cache  by `prettier.clearCache()`

thorn0:
I really think this PR is fine the way it is now.

@evilebottnawi Could you review, please?

PR #7501, opened 2020-01-31
Fix a failed test on `flow` parser

fisker:
I found a failed test

```js
const foo: ?boolean = true;
const bar = true || (foo || false);
```

This will format to 

```js
const foo: ?boolean = true;
const bar = true || foo || false;
```

AST_COMPARE shows diff

```diff
@@ -67,13 +67,11 @@
                    "source": null,
                  },
                  "type": "Literal",
                  "value": true,
                },
-               "loc": Object {
-                 "source": null,
-               },
+               "loc": Object {},
                "operator": "||",
                "right": Object {
                  "loc": Object {
                    "source": null,
                  },
@@ -81,14 +79,12 @@
                  "optional": false,
                  "type": "Identifier",
                  "typeAnnotation": null,
                },
                "type": "LogicalExpression",
-             },
-             "loc": Object {
-               "source": null,
              },
+             "loc": Object {},
              "operator": "||",
              "right": Object {
                "loc": Object {
                  "source": null,
                },
```

I'm not sure I'm doing this right.

### DIFF
--- a/src/common/parser-include-shebang.js
+++ b/src/common/parser-include-shebang.js
@@ -12,7 +12,6 @@ function includeShebang(text, ast) {
     value: shebang,
     range: [0, index],
     loc: {
-      source: null,
       start: {
         line: 1,
         column: 0
--- a/src/language-js/clean.js
+++ b/src/language-js/clean.js
@@ -17,6 +17,10 @@ function clean(ast, newObj, parent) {
     delete newObj[name];
   });
 
+  if (ast.loc && ast.loc.source === null) {
+    delete newObj.loc.source;
+  }
+
   if (ast.type === "BigIntLiteral") {
     newObj.value = newObj.value.toLowerCase();
   }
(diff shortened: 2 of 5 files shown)


fisker:
@thorn0 Can you help me with this? Or maybe you want investigate this case?

fisker:
Thanks, if it's not important.  Which one is better?

1.

```js
  if (ast.loc && ast.loc.source === null) {
    delete newObj.loc.source;
  }
```

2.

```js
  if (ast.source === null) {
    delete newObj.source;
  }
```

3.

```js
delete newObj.loc.source;
```

4.

```js
delete newObj.source;
```

4.

```js
delete newObj.loc;
```

thorn0:
The first (current) option looks safest. We don't want to accidentally delete something unrelated.

fisker:
Sure,  if we decide use first one, maybe I should change https://github.com/fisker/prettier/blob/28735c8e7e4bd583966d12aaace887dba3cca683/src/language-js/clean.js#L12-L13

to this style.


Also, merge maybe I should delete this line. https://github.com/prettier/prettier/blob/6c482e2e6c4d771407302fcc999bb3e2b491cbd3/src/common/parser-include-shebang.js#L16

thorn0:
`start` and `end` can stay as is. Babel creates properties with these names directly on AST nodes, so it's safe to presume that these names won't get used for something else (e.g. for a new type of AST node). But we can't be that sure about `source`.

thorn0:
> maybe I should delete this line.

You meant the line 15, not 16, right?

fisker:
Yes, 15, typo

PR #7504, opened 2020-01-31
Don't add parens to `undefined` in `ConditionalExpression` with `jsx`

### LINKED ISSUE #3552: JSX ternaries: `null` is collapsed, but `undefined` is not
**Prettier 1.9.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAggB0wIQIYBOAOlAASkCWkUAShQOYAWMAvMAM4QC2cMjFUeqQD8pVO0x4oAPgA20IfAAeMUnzyqK7UvMGl6EONsZwCcUpzWMNVraQBGZvAGt2qAPQSp00klJQAV1lZAF8ScipoABk4ADNWDm5efj1RcUkZXUU4FSsbOyz9Q2NTc0t1VT47RzgXN08Mnz8AqAATOIE4VrCod2kQABoQCEwYCmh2ZFBCAggAdwAFQgRJlDwANwgKVsGQPHYYZFi8WXY4Icc8MGdeAGVJMAF6ZBgCAPOQVogwI5OzoYEZwIMAWBDw9C4eF+pw+ACt2EocGDrnc8Dwop1of9hgEYJhcQAmLEfSQEIHIED2PD2ACeul2mAIAhgAHVtnxkAAOAAMQ0ZEDOLLBmApjKMpnWcF2ZgAjgEKGZQeDIcShmcuBQXm8PuwnrI4ABFAIQeCqkAwals1ocpAEoavPAUWRPADC3BVKCg0ClQwCZwAKtTVscYSEQkA)

**Input:**
```jsx
<AppBar
  iconRight={something ? <span>long text that is long goes here so that this breaks</span> : null}
  iconLeft={something ? <span>long text that is long goes here so that this breaks</span> : undefined}
/>
```

**Output:**
```jsx
<AppBar
  iconRight={
    something ? (
      <span>long text that is long goes here so that this breaks</span>
    ) : null
  }
  iconLeft={
    something ? (
      <span>long text that is long goes here so that this breaks</span>
    ) : (
      undefined
    )
  }
/>;

```

**Expected behavior:**
```jsx
<AppBar
  iconRight={
    something ? (
      <span>long text that is long goes here so that this breaks</span>
    ) : null
  }
  iconLeft={
    something ? (
      <span>long text that is long goes here so that this breaks</span>
    ) : undefined
  }
/>;

```

When I wrote the JSX ternary handling that special-cased `null`, I didn't also do so for `undefined` because it was an invalid node type at the time. However, since it is a valid node type in React 16, it should probably be treated the same as `null`. Maybe `false` should collapse, too (though it's kinda weird to put in a ternary).

fisker:
Fixes #3552


tests/jsx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -756,6 +756,10 @@ cable ? (
   ) : null}
 </div>;
 
+// #3552
+foo ? <span>loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong jsx</span> :
+undefined
+
```
evilebottnawi: Do we have test for `null`?

fisker: L756

fisker: Two lines above this line

PR #7505, opened 2020-01-31
Refactor `FastPath#get{Name,Value}` to `getter`

fisker:

### DIFF
--- a/src/common/fast-path.js
+++ b/src/common/fast-path.js
@@ -23,7 +23,7 @@ class FastPath {
 
   // The name of the current property is always the penultimate element of
   // this.stack, and always a String.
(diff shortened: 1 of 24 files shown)


evilebottnawi:
I do not see any benefits from the approach except that we will break the backward compatibility of all

fisker:
@evilebottnawi Don't you think it's easier to use? If you consider `backward compatibility` , how about deprecate them first.

evilebottnawi:
Why? I really see no benefits

evilebottnawi:
I think it would be better to focus on problems for 2.0 than break backward compatibility without benefits

Issue #7407, opened 2020-01-21
Document how to integrate in pre-commit with plugins


Yajo:
**Environments:**
- Prettier Version: 1.19.1
- Running Prettier via: pre-commit 
- Runtime: node v12.13.1 
- Operating System: Linux 

**Steps to reproduce:**

All I want to do is to integrate the prettier-xml plugin into pre-commit.

I tried with this, but it doesn't work:

```yaml
  - repo: https://github.com/prettier/prettier
    rev: "1.19.1"
    hooks:
      - id: prettier
        name: prettier xml plugin
        additional_dependencies:
          - "@prettier/plugin-xml@0.5.0"
        args:
          - --with-node-modules
        files: \.xml$
```

**Expected behavior:** Reformat xml files

**Actual behavior:**

Running `pre-commit run -a`:

```
prettier xml plugin......................................................Failed
hookid: prettier

[error] No parser could be inferred for file: web_responsive/views/web.xml
[error] No parser could be inferred for file: web_responsive/static/src/xml/apps.xml
[error] No parser could be inferred for file: web_notify/views/res_users_demo.xml
[error] No parser could be inferred for file: web_responsive/static/src/xml/form_view.xml
[error] No parser could be inferred for file: web_notify/views/web_notify.xml
[error] No parser could be inferred for file: web_tree_many2one_clickable/views/assets.xml
[error] No parser could be inferred for file: web_dialog_size/static/src/xml/web_dialog_size.xml
[error] No parser could be inferred for file: web_environment_ribbon/view/base_view.xml
[error] No parser could be inferred for file: web_dialog_size/templates/assets.xml
[error] No parser could be inferred for file: web_ir_actions_act_view_reload/views/web_ir_actions_act_view_reload.xml
[error] No parser could be inferred for file: web_responsive/views/assets.xml
[error] No parser could be inferred for file: web_responsive/views/res_users.xml
[error] No parser could be inferred for file: web_responsive/static/src/xml/navbar.xml
[error] No parser could be inferred for file: web_environment_ribbon/data/ribbon_data.xml
[error] No parser could be inferred for file: web_decimal_numpad_dot/views/web_decimal_numpad_dot.xml
[error] No parser could be inferred for file: web_widget_bokeh_chart/views/web_widget_bokeh_chart.xml
```

**Workaround**:

Configuration from https://github.com/prettier/plugin-xml/issues/17#issuecomment-555445966 works, but I guess that's not what you want to recommend to use.

[...]

PR #7508, opened 2020-02-01
Remove `resolve` package

fisker:
I think `require.resolve` can do the job.

Tests from separate PR https://github.com/prettier/prettier/pull/7517

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -31,7 +31,7 @@ function loadPlugins(plugins, pluginSearchDirs) {
 
   const externalManualLoadPluginInfos = externalPluginNames.map(pluginName => ({
     name: pluginName,
-    requirePath: require.resolve(pluginName, { paths: [process.cwd()] })
+    requirePath: eval("require").resolve(pluginName, { paths: [process.cwd()] })
   }));
 
   const externalAutoLoadPluginInfos = pluginSearchDirs
@@ -60,7 +60,7 @@ function loadPlugins(plugins, pluginSearchDirs) {
 
       return findPluginsInNodeModules(nodeModulesDir).map(pluginName => ({
         name: pluginName,
-        requirePath: require.resolve(pluginName, {
+        requirePath: eval("require").resolve(pluginName, {
           paths: [resolvedPluginSearchDir]
         })
       }));
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -29,10 +29,25 @@ function loadPlugins(plugins, pluginSearchDirs) {
     plugin => typeof plugin === "string"
   );
 
-  const externalManualLoadPluginInfos = externalPluginNames.map(pluginName => ({
-    name: pluginName,
-    requirePath: eval("require").resolve(pluginName, { paths: [process.cwd()] })
-  }));
+  const externalManualLoadPluginInfos = externalPluginNames.map(pluginName => {
+    let requirePath;
+    try {
+      // try local files
+      requirePath = eval("require").resolve(
+        path.resolve(process.cwd(), pluginName)
+      );
+    } catch (_) {
+      // try node modules
+      requirePath = eval("require").resolve(pluginName, {
+        paths: [process.cwd()]
+      });
+    }
+
+    return {
+      name: pluginName,
+      requirePath
+    };
+  });
 
   const externalAutoLoadPluginInfos = pluginSearchDirs
     .map(pluginSearchDir => {
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -5,7 +5,6 @@ const partition = require("lodash/partition");
 const fs = require("fs");
 const globby = require("globby");
 const path = require("path");
-const resolve = require("resolve");
 const thirdParty = require("./third-party");
 const internalPlugins = require("./internal-plugins");
 
@@ -30,20 +29,10 @@ function loadPlugins(plugins, pluginSearchDirs) {
     plugin => typeof plugin === "string"
   );
 
-  const externalManualLoadPluginInfos = externalPluginNames.map(pluginName => {
-    let requirePath;
-    try {
-      // try local files
-      requirePath = resolve.sync(path.resolve(process.cwd(), pluginName));
-    } catch (e) {
-      // try node modules
-      requirePath = resolve.sync(pluginName, { basedir: process.cwd() });
-    }
-    return {
-      name: pluginName,
-      requirePath
-    };
-  });
+  const externalManualLoadPluginInfos = externalPluginNames.map(pluginName => ({
+    name: pluginName,
+    requirePath: require.resolve(pluginName, { paths: [process.cwd()] })
+  }));
 
   const externalAutoLoadPluginInfos = pluginSearchDirs
     .map(pluginSearchDir => {
@@ -71,8 +60,8 @@ function loadPlugins(plugins, pluginSearchDirs) {
 
       return findPluginsInNodeModules(nodeModulesDir).map(pluginName => ({
         name: pluginName,
-        requirePath: resolve.sync(pluginName, {
-          basedir: resolvedPluginSearchDir
+        requirePath: require.resolve(pluginName, {
+          paths: [resolvedPluginSearchDir]
         })
       }));
     })
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "regexp-util": "1.2.2",
     "remark-math": "1.0.6",
     "remark-parse": "5.0.0",
-    "resolve": "1.15.1",
     "semver": "7.1.3",
     "srcset": "2.0.1",
     "string-width": "4.2.0",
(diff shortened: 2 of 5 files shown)


src/common/load-plugins.js
```diff
@@ -30,20 +29,10 @@ function loadPlugins(plugins, pluginSearchDirs) {
...
-    }
-    return {
-      name: pluginName,
-      requirePath
-    };
-  });
+  const externalManualLoadPluginInfos = externalPluginNames.map(pluginName => ({
+    name: pluginName,
+    requirePath: eval("require").resolve(pluginName, { paths: [process.cwd()] })
+  }));
```
fisker: This part might different with old one.

If  `pluginName = 'my-plugin'` not `pluginName = './my-plugin'`, result might different, will look for `my-plugin` package, insteadof `./my-plugin`, but no tests cover it.

thorn0: We should keep the existing behavior. I'd like to understand though what was the original motivation for using this module instead of `require.resolve`. Any ideas?

fisker: https://github.com/prettier/prettier/pull/3536/files .

`require.resolve` supports `options` until `v8.9.0` https://nodejs.org/api/modules.html#modules_require_resolve_paths_request

I think that's the reason

fisker: Logic of this part has restored.

thorn0: Can you please add a test for that case?

brodybits: How important is it to support the old behavior in the new release (#6888)?

fisker: This should take a while, I don't know where pluginName come from yet.

fisker: This line seems mean to test this, but did't catch it.

https://github.com/prettier/prettier/blob/master/tests_integration/__tests__/plugin-resolution.js#L148

fisker: I add a print

```js
    console.log(
      JSON.stringify(
        {
          pluginName,
          "with-path.resolve": eval("require").resolve(
            path.resolve(process.cwd(), pluginName)
          ),
          "with-require.resolve(..., options)": eval("require").resolve(
            pluginName,
            {
              paths: [process.cwd()]
            }
          )
        },
        null,
        2
      )
    );
```

Got same result on `./automatic/node_modules/prettier-plugin-bar` and `automatic/node_modules/prettier-plugin-bar`, not sure where `.js` lost

```json
     {
       "pluginName": "./automatic/node_modules/prettier-plugin-bar",
       "with-path.resolve": "<repo>\\prettier\\tests_integration\\plugins\\automatic\\node_modules\\prettier-plugin-bar\\index.js",
       "with-require.resolve(..., options)": "<repo>\\prettier\\tests_integration\\plugins\\automatic\\node_modules\\prettier-plugin-bar\\index.js"
     }
```


```json
    {
       "pluginName": "automatic/node_modules/prettier-plugin-bar",
       "with-path.resolve": "<repo>\\prettier\\tests_integration\\plugins\\automatic\\node_modules\\prettier-plugin-bar\\index.js",
       "with-require.resolve(..., options)": "<repo>\\prettier\\tests_integration\\plugins\\automatic\\node_modules\\prettier-plugin-bar\\index.js"
     }
```

But I ran on cmd, this failed as expected

```text
<repo>\prettier\tests_integration\plugins>node
Welcome to Node.js v13.5.0.
Type ".help" for more information.
> require.resolve("automatic/node_modules/prettier-plugin-bar")
Uncaught Error: Cannot find module 'automatic/node_modules/prettier-plugin-bar'
Require stack:
- <repl>
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:17)
    at Function.resolve (internal/modules/cjs/helpers.js:78:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '<repl>' ]
}
```

fisker: @thorn0 have a look

fisker:
I found something interesting, I print the `require.resolve.toString()` on test,

```js
"(moduleName, options) =>\n      this._requireResolve(from.filename, moduleName, options)"
```

But, on console

```console
$ node -p "require.resolve.toString()"
function resolve(request, options) {
    validateString(request, 'request');
    return Module._resolveFilename(request, mod, false, options);
  }
```

They are different

So Jest not using the node.js require ???? Didn't find any docs about that.

fisker:
Found it in code

https://github.com/facebook/jest/blob/c9127bfec38882f02abed4bf41514f1843708ecb/packages/jest-runtime/src/index.ts#L1059

fisker:
Test that can catch this error added #7517

SimenB:
Jest implements `require` on its own to support transpilation and mocks

brodybits:
> Jest implements `require` on its own to support transpilation and mocks

Is this documented somewhere?

fisker:
@SimenB yes, I figured, and there is a bug `require.resolve('a-path/to/file')` should try resolve `a-path` module not `a-path/to/file.js` file

SimenB:
I don't think so

src/common/load-plugins.js
```diff
@@ -34,11 +33,16 @@ function loadPlugins(plugins, pluginSearchDirs) {
...
+      requirePath = eval("require").resolve(
+        path.resolve(process.cwd(), pluginName)
+      );
+    } catch (_) {
       // try node modules
-      requirePath = resolve.sync(pluginName, { basedir: process.cwd() });
+      requirePath = eval("require").resolve(pluginName, {
+        paths: [process.cwd()]
+      });
     }
```
fisker: I've merged #7517 into this one, to catch that error, remove this `try/catch` part and run 

```bash
yarn test tests_integration/__tests__/plugin-resolution.js
```

fisker:
@SimenB You can check my comment https://github.com/prettier/prettier/pull/7508#discussion_r373860098 , 

If you want I can show you how to find the difference from jest `require.resolve` and node.js `require.resolve` .

brodybits:
> If you want I can show you how to find the difference from jest `require.resolve` and node.js `require.resolve` .

I think it would be ideal if someone could do this in Q&A style on Stack Overflow. It would also help `require.resolve` newbies such as myself if someone can describe how this PR was able to overcome this limitation.

SimenB:
> @SimenB You can check my comment [#7508 (comment)](https://github.com/prettier/prettier/pull/7508#discussion_r373860098) ,
> 
> If you want I can show you how to find the difference from jest `require.resolve` and node.js `require.resolve` .

If you could open up an issue at fb/jest with a small reproduction that would be wonderful yeah

fisker:
@SimenB Let me try first on my machine

fisker:
@SimenB 

checkout this repo

https://github.com/fisker/jest-require-resolve-bug

If you confirm that or something I did wrong, let me know here first.

brodybits:
> https://github.com/fisker/jest-require-resolve-bug

It took me a little while to understand the difference between `require.resolve` on Node.js vs Jest. The difference I found is that `require.resolve('foo/node_modules/bar', {paths: [process.cwd()]})` throws an error in Node.js but succeeds in Jest. Looks like a subtle bug in Jest.

I think it would also really help users in the future if someone could document what Jest does to the require statements and why. I am also wondering if Jest would do something similar to ES6 import statements.

SimenB:
Yeah, that's a bug. Seems like we're missing something in the implementation from https://github.com/facebook/jest/pull/6471. Something about making things relative while they're not seems off. Could you open up an issue with it?

---

@brodybits there should be no observable differences between Jest's `require` and node's implementations (by invoking the functions, not `toString` etc) beyond options passed to Jest that change it, like `moduleNameMapper` and code transforms. So nothing really to document. It'll be the same with ESM when that's implemented

SimenB:
Seeing as `require.resolve` with `paths` is broken in Node 10 (fixed in 12: https://github.com/nodejs/node/pull/23683) I'd suggest to not land this until you also drop node 10

fisker:
@SimenB Doesn't matter anymore, new test code won't relay `require.resolve` to throw

SimenB:
I'm not talking about test code, I'm talking about loading user plugins, which is where you use the `resolve` package today. Using `require.resolve('asdasd', {paths: []})` when you _know_ it's broken on Node 10 seems irresponsible to me, especially if it's just about avoiding a single dependency and not some feature `resolve` doesn't support or `resolve` messing up browser builds or some such.

From the OP:

>I think `require.resolve` can do the job.

It cannot, it's broken on node 10 for the use case you want to use it.

---

I'll unsubscribe since it doesn't really impact me (I don't use custom plugins), and I don't really have any interest in discussing this. Thanks for finding the inconsistency in Jest's implementation though! 😀 We'll replicate the implementation in the version of node we're running on (at least that's the current plan)

fisker:
Actually the node 10 behavior is more we want, when resolve plugin `foo`, we try `<CWD>/foo` first. So node 10 broken `require.resolve` doesn't effect us, we are not breaking anything.

We found that problem is because I tried to remove `<CWD>/foo` first, but didn't catch a error. So we improved the test, make sure we are not changing the logic.

thorn0:
Might fix https://github.com/prettier/prettier/issues/7407 if `require.resolve` uses `NODE_PATH`.

PR #7514, opened 2020-02-02
Print hex value in uppercase

fisker:
Currently, we print all numeric(number, bigint) value in lower case, proposal print hex value in uppercase , both `number` and `bigint`
`0xdeedbeef` -> `0xDEEDBEEF`
`0xdeedbeefn` -> `0xDEEDBEEFn`
it's more readable.

`eslint-plugin-unicorn` have a rule do the same thing, [`number-literal-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/number-literal-case.md)


kachkaev:
Science suggests that upper case is actually _less_ readable. Lower case letter have a higher distinction between themselves because of ascenders and descenders, which eases their cognition:

image

Upper case is suitable only in small captions when such styling is justified by a professional designer, who understands the context and the trade-offs. When using all caps, it is common to increase spacing between letters to improve readability.

Not sure this PR is useful for Prettier community.

fisker:
`0x@deed_beefn` vs `0xDEED_BEEEFn`

thorn0:
I unintentionally closed this and several other PRs because I deleted the next branch, but I think the consensus here is against merging, so let's keep this PR closed.

PR #7516, opened 2020-02-02
Put space after generator star, not before

bakkot:
This is a partial revert of https://github.com/prettier/prettier/pull/3903. That PR also included an unrelated change to the spacing of the `*` used for generator declarations and `yield*`, after very brief discussion in https://github.com/prettier/prettier/issues/7028, which this PR reverts (so that they are spelled `function*` and `yield*`, rather than `function *` and `yield *`).

This is, of course, just a style preference. But generally speaking I think prettier shouldn't change things which there is broad agreement on without extremely good reason, and here it looks to me like there was broad agreement on the previous style: see [earlier discussion in the thread preceding #3903](https://github.com/prettier/prettier/issues/3847#issuecomment-363449250). I assume this discussion was missed when #7028 was being considered.


bakkot:
> Please do not send PR before it was resolved

Oh, sorry. When did that policy change? It used to be that we'd make PRs whenever something was being considered, and just not merge it until there was agreement.

evilebottnawi:
@bakkot You can still do it, but it does not make sense, it is not a difficult change, so I would prefer not to waste time until there is a final solution, just my recommendation

changelog_unreleased/javascript/pr-3903.md
```diff
@@ -1,36 +1,18 @@
...
+#### Always add a space after the `function` keyword ([#3903](https://github.com/prettier/prettier/pull/3903) by [@j-f1](https://github.com/j-f1), [@josephfrazier](https://github.com/josephfrazier), [@sosukesuzuki](https://github.com/sosukesuzuki), [@thorn0](https://github.com/thorn0); [#7516](https://github.com/prettier/prettier/pull/7516) by [@bakkot](https://github.com/bakkot))
 
-Previously, a space would be added after the `function` keyword in function declarations, but not in function expressions. Now, for consistency, a space is always added after the `function` keyword, including in generators. Also, a space is now added between `yield` and `*`, which can help with [catching bugs](https://github.com/prettier/prettier/issues/7028#user-content-bugs).
+Previously, a space would be added after the `function` keyword in function declarations, but not in function expressions. Now, for consistency, a space is always added after the `function` keyword. The only exception is generator declarations where `function*` is treated as a whole word.
 
 <!-- prettier-ignore -->
 ```js
```
glen-84: This should be `ts` if there are generics?

PR #7517, opened 2020-02-02
Improve `plugin-resolution` test

fisker:
Add a test with same packageName and fileName, fix a missing case from #7508

### DIFF
--- a/tests_integration/__tests__/plugin-resolution.js
+++ b/tests_integration/__tests__/plugin-resolution.js
@@ -5,7 +5,7 @@ const EOL = "\n";
 
 describe("automatically loads 'prettier-plugin-*'", () => {
   runPrettier("plugins/automatic", ["file.txt", "--parser=bar"]).test({
-    stdout: "bar+contents" + EOL,
+    stdout: "content from `prettier-plugin-bar` package + contents" + EOL,
     stderr: "",
     status: 0,
     write: []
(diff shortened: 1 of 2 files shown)


tests_integration/__tests__/plugin-resolution.js
```diff
@@ -173,7 +173,20 @@ describe("loads --plugin by relative path to its directory without leading ./ (a
...
+    status: 0,
+    write: []
+  });
+});
+
+describe("loads --plugin by filename without leading ./ and ext, should resolve to file, not package", () => {
+  runPrettier("plugins/automatic", [
+    "file.txt",
+    "--parser=bar",
+    "--plugin=prettier-plugin-bar"
```
fisker: This part do the trick, if we don't try local file first, will resolve to the package `prettier-plugin-bar` , instead of `prettier-plugin-bar.js` file

brodybits: I would favor documenting this in the test itself.

brodybits:
Seems to be motivated by review of PR #7508.

@fisker would you mind describing some of the context in this PR, for the sake of plugin resolution newbies such as myself?

tests_integration/__tests__/plugin-resolution.js
```diff
@@ -173,7 +173,20 @@ describe("loads --plugin by relative path to its directory without leading ./ (a
...
+  ]).test({
+    stdout: "content from `prettier-plugin-bar.js` file + contents" + EOL,
```
fisker: @brodybits I put different content here

fisker:
@brodybits If plugin is passed 'foo', prettier try '<cwd>/foo' first, then try `foo` package. In #7508, I remove the first part, `require.resolve` should throw , but it didn't on jest.

thorn0:
This changes only a test, so I'm going ahead and merging this.

PR #7528, opened 2020-02-03
refactoring(glimmer): Extract utils from printer

dcyriller:
#### Description

Extract a few glimmer printer utils into a file, similar to what is done for others languages.

#### Checklist

### DIFF
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -65,9 +65,29 @@ function getNextNode(path) {
   }
 }
 
+function isPrettierIgnoreNode(node) {
+  if (!isNodeOfSomeType(node, ["MustacheCommentStatement"])) {
+    return false;
+  }
+
+  return (
+    typeof node.value === "string" && node.value.trim() === "prettier-ignore"
+  );
+}
+
+function hasPrettierIgnore(path) {
+  const n = path.getValue();
+  const previousPreviousNode = getPreviousNode(path, 2);
+  const isIgnoreNode = isPrettierIgnoreNode(n);
+  const isCoveredByIgnoreNode = isPrettierIgnoreNode(previousPreviousNode);
+
+  return isIgnoreNode || isCoveredByIgnoreNode;
+}
+
 module.exports = {
   getNextNode,
   getPreviousNode,
+  hasPrettierIgnore,
   isGlimmerComponent,
   isNextNodeOfSomeType,
   isNodeOfSomeType,
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -16,10 +16,7 @@ function isWhitespaceNode(node) {
 }
 
 function isNodeOfSomeType(node, types) {
-  if (node) {
-    return types.some(type => node.type === type);
-  }
-  return false;
+  return node && types.some(type => node.type === type);
 }
 
 function isParentOfSomeType(path, types) {
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -66,12 +66,12 @@ function getNextNode(path) {
 }
 
 module.exports = {
-  getPreviousNode,
   getNextNode,
+  getPreviousNode,
+  isGlimmerComponent,
+  isNextNodeOfSomeType,
   isNodeOfSomeType,
   isParentOfSomeType,
   isPreviousNodeOfSomeType,
-  isNextNodeOfSomeType,
-  isWhitespaceNode,
-  isGlimmerComponent
+  isWhitespaceNode
 };
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -35,20 +35,20 @@ function isNextNodeOfSomeType(path, types) {
   return isNodeOfSomeType(nextNode, types);
 }
 
-function getPreviousNode(path, lookBack = 1) {
+function getSiblingNode(path, offset) {
   const node = path.getValue();
   const parentNode = path.getParentNode(0) || {};
   const children = parentNode.children || parentNode.body || [];
   const index = children.indexOf(node);
-  return index !== -1 && children[index - lookBack];
+  return index !== -1 && children[index + offset];
+}
+
+function getPreviousNode(path, lookBack = 1) {
+  return getSiblingNode(path, -lookBack);
 }
 
 function getNextNode(path) {
-  const node = path.getValue();
-  const parentNode = path.getParentNode(0);
-  const children = parentNode.children || parentNode.body || [];
-  const index = children.indexOf(node);
-  return index !== -1 && children[index + 1];
+  return getSiblingNode(path, 1);
 }
 
 function isPrettierIgnoreNode(node) {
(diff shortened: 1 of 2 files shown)


src/language-handlebars/utils.js
```diff
@@ -11,6 +11,87 @@ function isGlimmerComponent(node) {
...
+function isWhitespaceNode(node) {
+  return node.type === "TextNode" && !/\S/.test(node.chars);
+}
+
+function isNodeOfSomeType(node, types) {
+  if (node) {
+    return types.some(type => node.type === type);
+  }
+  return false;
+}
```
fisker:
```suggestion
function isNodeOfSomeType(node, types) {
  return node && types.some(type => node.type === type);
}
```

dcyriller: I applied your suggestion

src/language-handlebars/utils.js
```diff
@@ -11,6 +11,84 @@ function isGlimmerComponent(node) {
...
+
+function isPrettierIgnoreNode(node) {
+  if (!isNodeOfSomeType(node, ["MustacheCommentStatement"])) {
+    return false;
+  }
+
+  return (
+    typeof node.value === "string" && node.value.trim() === "prettier-ignore"
+  );
+}
```
fisker:
```suggestion
function isPrettierIgnoreNode(node) {
  return (
    isNodeOfSomeType(node, ["MustacheCommentStatement"]) &&
    typeof node.value === "string" &&
    node.value.trim() === "prettier-ignore"
  );
}
```

src/language-handlebars/utils.js
```diff
@@ -11,6 +11,84 @@ function isGlimmerComponent(node) {
...
+
+function getNextNode(path) {
+  const node = path.getValue();
+  const parentNode = path.getParentNode(0);
+
+  const children = parentNode.children || parentNode.body;
+  if (children) {
+    const nodeIndex = children.indexOf(node);
+    if (nodeIndex < children.length) {
```
fisker: This seems can't be false?

fisker: Is this mean to check `< .length - 1`?

fisker: but still unnessary `children[nodeIndex + 1]` also get `undefined`

src/language-handlebars/utils.js
```diff
@@ -11,6 +11,84 @@ function isGlimmerComponent(node) {
...
+      const nextNode = children[nodeIndex + 1];
+      return nextNode;
+    }
+  }
+}
```
fisker:
```suggestion
function getNextNode(path) {
  const node = path.getValue();
  const parentNode = path.getParentNode(0);
  const children = parentNode.children || parentNode.body;
  return children[children.indexOf(node) + 1];
}
```

fisker: We are not check `index === -1`, I don't think need check `children` either

dcyriller: My initial goal was more to extract these utils than to refactor them. But there is definitely work to do in this area. 😄

dcyriller: Nice one

fisker: Should be safe

fisker: `getPreviousNode` checked `nodeIndex > 0` , I think we should also check it in `getPreviousNode`

fisker: I updated the suggestion

src/language-handlebars/utils.js
```diff
@@ -11,6 +11,84 @@ function isGlimmerComponent(node) {
...
+
+  const children = parentNode && (parentNode.children || parentNode.body);
+  if (children) {
+    const nodeIndex = children.indexOf(node);
+    if (nodeIndex > 0) {
+      const previousNode = children[nodeIndex - lookBack];
+      return previousNode;
+    }
+  }
+}
```
fisker:
```suggestion
function getPreviousNode(path, lookBack = 1) {
  const node = path.getValue();
  const parentNode = path.getParentNode(0) || {};
  const children = parentNode.children || parentNode.body || [];
  const index = children.indexOf(node)
  return index !== -1 && children[index - lookBack];
}
```

src/language-handlebars/utils.js
fisker: How about ...

```js
const getNextNode = node => getPreviousNode(node, -1);
```

dcyriller: > getPreviousNode checked nodeIndex > 0 , I think we should also check it in getNextNode, that one also checked parentNode

agreed

src/language-handlebars/utils.js
dcyriller: I have applied your suggestions. It unlocks a few good improvements. Do you mind having a look at my last commit?

src/language-handlebars/utils.js
```diff
@@ -1,16 +1,80 @@
...
+
+function getNextNode(path) {
+  const node = path.getValue();
+  const parentNode = path.getParentNode(0);
+  const children = parentNode.children || parentNode.body || [];
+  const index = children.indexOf(node);
+  return index !== -1 && children[index + 1];
+}
+
```
fisker: Looks like my review get lost, some how,

Make a `getSiblingNode(path, offset)`, and `getNextNode`, `getPreviousNode` as shortcuts

dcyriller: CI was red because of a typo

Good advice, I'll add it.

dcyriller: Done, neat refactoring in the end. Thanks for your help :)

src/language-handlebars/utils.js
```diff
@@ -1,16 +1,80 @@
...
-  }
+  return (
+    isNodeOfSomeType(node, ["ElementNode"]) &&
+    typeof node.tag === "string" &&
+    (isUppercase(node.tag[0]) || node.tag.includes("."))
+  );
+}
+
+function isWhitespaceNode(node) {
+  return isNodeOfSomeType(node, ["TextNode"]) && !/\S/.test(node.chars);
```
fisker:
```suggestion
  return isNodeOfSomeType(node, ["TextNode"]) && /\s/.test(node.chars);
```

fisker: Seems don't need negative, right?

fisker: Wait, not equal, if all `\s`

fisker: Sorry, I think my suggestion is not right.

fisker:
```js
!/\S/.test('')
true
/\s/.test('')
false
```

fisker:
@dcyriller need undo last one, my fault.

src/language-handlebars/utils.js
fisker: /^\s*$/ should be equal, but old one better

dcyriller: Removed the last commit 😉

PR #7529, opened 2020-02-03
refactoring(glimmer): Create a clean file

dcyriller:
#### Description

This PR proposes to extract clean function from glimmer-printer to its file, similarly to what is done with language-js / language-html.

#### Checklist

### DIFF
--- a/src/language-handlebars/clean.js
+++ b/src/language-handlebars/clean.js
@@ -6,9 +6,9 @@ module.exports = function(ast, newNode) {
 
   // (Glimmer/HTML) ignore TextNode whitespace
   if (ast.type === "TextNode") {
-    if (ast.chars.replace(/\s+/, "") === "") {
+    if (ast.chars.trim() === "") {
       return null;
     }
-    newNode.chars = ast.chars.replace(/^\s+/, "").replace(/\s+$/, "");
+    newNode.chars = ast.chars.trim();
   }
 };
--- a/src/language-handlebars/clean.js
+++ b/src/language-handlebars/clean.js
@@ -1,0 +1,14 @@
+"use strict";
+
+module.exports = function(ast, newNode) {
+  delete newNode.loc;
+  delete newNode.selfClosing;
+
+  // (Glimmer/HTML) ignore TextNode whitespace
+  if (ast.type === "TextNode") {
+    if (ast.chars.replace(/\s+/, "") === "") {
+      return null;
+    }
+    newNode.chars = ast.chars.replace(/^\s+/, "").replace(/\s+$/, "");
+  }
+};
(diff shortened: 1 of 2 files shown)


src/language-handlebars/clean.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+module.exports = function(ast, newNode) {
+  delete newNode.loc;
+  delete newNode.selfClosing;
+
+  // (Glimmer/HTML) ignore TextNode whitespace
+  if (ast.type === "TextNode") {
+    if (ast.chars.replace(/\s+/, "") === "") {
```
fisker: how about trim()?

src/language-handlebars/clean.js
```diff
@@ -0,0 +1,14 @@
...
+      return null;
+    }
+    newNode.chars = ast.chars.replace(/^\s+/, "").replace(/\s+$/, "");
```
fisker: trim

src/language-handlebars/printer-glimmer.js
```diff
@@ -622,5 +609,5 @@ function isPrettierIgnoreNode(node) {
 
 module.exports = {
   print,
-  massageAstNode: clean
+  massageAstNode: require("./clean")
```
fisker: move to top?

src/language-handlebars/clean.js
evilebottnawi: Just for you information - trim !== .replace(/\s+/, ""), trim === `/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, but here we can use it

dcyriller: good to know :)
I'll give it a try.

src/language-handlebars/printer-glimmer.js
dcyriller: I'm not sure

src/language-handlebars/clean.js
dcyriller: my middle-term goal it to remove that to make it whitespace sensible (like html)

src/language-handlebars/printer-glimmer.js
fisker: I mean `const clean = require("./clean")` at top of this file

src/language-handlebars/clean.js
```diff
@@ -0,0 +1,14 @@
...
+module.exports = function(ast, newNode) {
+  delete newNode.loc;
+  delete newNode.selfClosing;
+
+  // (Glimmer/HTML) ignore TextNode whitespace
+  if (ast.type === "TextNode") {
+    if (ast.chars.trim() === "") {
+      return null;
+    }
+    newNode.chars = ast.chars.trim();
```
fisker:
```suggestion
	const trimmed = ast.chars.trim();
    if (!trimmed) {
      return null;
    }
    newNode.chars = trimmed;
```

dcyriller:
Thank you @thorn0

PR #7530, opened 2020-02-03
Prettier needs an absolute path instead of relative

nixolas1:
Changed the relative path variable to an absolute one, to get prettier to work.


fisker:
Can someone confirm this? If it's correct, [this image](docs/assets/webstorm/file-watcher-prettier.png] need update too.

thorn0:
@prigara Could you probably comment on this? Does this change make sense? Is `$FilePathRelativeToProjectRoot$` always relative to `$ProjectFileDir$` or can `$ProjectFileDir$` be different from the project root?

prigara:
@thorn0 as far as I can see, they can be different in case of multi-module projects in IntelliJ IDEA. 
I'm fine with the changes - using full paths when prettier is installed globally seems to be a good idea.

prigara:
@fisker probably it's ok to remove this screenshot completely since this setup is for older IDE versions only.

PR #7533, opened 2020-02-04
Remove `babylon` parser

fisker:
ref: #4097

### DIFF
--- a/changelog_unreleased/api/pr-7533.md
+++ b/changelog_unreleased/api/pr-7533.md
@@ -1,3 +1,3 @@
 #### Remove deprecated parser `babylon` ([#7533](https://github.com/prettier/prettier/pull/7533) by [@fisker](https://github.com/fisker))
 
-Parser `babylon` renamed to `babel` in _v1.16.0_, but aliased to keep backwards compatibility, now it's removed.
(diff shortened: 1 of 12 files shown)


src/main/support.js
```diff
@@ -77,11 +76,6 @@ function getSupportInfo(version, opts) {
       // Prevent breaking changes
       if (language.name === "Markdown") {
         parsers = ["markdown"];
-        // "babylon" was renamed to "babel" in 1.16.0
-      } else if (useBabylonParser && language.parsers.includes("babel")) {
-        parsers = language.parsers.map(parser =>
-          parser === "babel" ? "babylon" : parser
-        );
```
fisker: Guys, do we need keep this? otherwise, the support info seems not right for old versions.

thorn0: That's a good question. What for exactly do we need this `version` parameter in [`prettier.supportInfo`](https://prettier.io/docs/en/api.html#prettiergetsupportinfoversion)? What are its use cases? Can we probably simply remove it in 2.0?

@azz It was [you](https://prettier.io/blog/2017/11/07/1.8.0.html#implement-getsupportinfo-and-use-it-for-inference-3033httpsgithubcomprettierprettierpull3033-by-azzhttpsgithubcomazz) who implemented it. What was the original use case? #3082?

Issue #5482, opened 2018-11-14
[HTML] Format values of attributes with defined semantics


nunocastromartins:
As briefly discussed in #5479, this issue targets the formatting of HTML attribute values that have defined semantics: for example, in regard to white spaces or commas.

An example of such an attribute is `class`, where `class=" a   b      c "` should be semantically equivalent to `class="a b c"`. However, note that if the user is parsing `element.className` directly, this change is **not safe**. This could be controller via `--html-whitespace-sensitivity` or a new flag.

This issue would first require some research effort into finding out which HTML attributes have values with semantics that allow Prettier to manipulate them. A list of all (most?) HTML attributes can be found at: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

Implementation note: some attribute semantics are bound to the HTML element they are part of, _e.g._ the semantics of `name` are different between a `button` element and a `meta` element; `class` is a global attribute, so its semantics should be valid for all elements.

**Prettier 1.15.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIwA2AhgM4kC8AOiEbnbgEaVT367r3UB8qA9Fti7NUAWzgxakWAhhUQuAO6Z0MABbl0cbJjBwAtEpWq6AGmZ1MUTDExECekmDtxyARgB0ABjMtWfvwCuJHAATg5OxAwELlAQ1P68QlColgAOATC4MACeqS7UAGaY0fFEYLqpsvGYIkQA5nC8dZgFdOb+HZ1dJrg19Y0AVnl1uNwgJiAQlZjQJMigRCEhEAoACosIcyhE2BDK4yAMIWUA1uIAyqlllnXIMCEBcBOqMCIEAOqq1nAkV7rnm2sOGs2WQ4DIB0swRCMFWxzqtWQBTswQmAxIAA8AELHMBnGDnIhiAAyljgSJRTxA6Ix5xu0QAigEIPAKQRUSAriFoWCXm8DqkQpYYO9lGpkAAObyc5bBd7HVJgwU-ULYckTEJwACOAUwmrh9URSGR7KpwREmDuDzN9LgTJZ5ONlImEgYoqMyAATC7jsUbgBhCAiI0gWJQdUgIJwAAqRAYWxNwQAvkmgA)
```sh
--parser html
```

**Input (`class`, `content`, and `accept` examples):**
```html
<div class="a    b
    c d   "></div>
[...]

Issue #5479, opened 2018-11-14
[HTML] Format the content of the style attribute


nunocastromartins:
I'm not absolutely sure that this is a safe change, but still, I thought it would be worth opening an issue for. This likely applies to Angular and Vue as well.

At least in Angular there might be an issue if there is an `Input` named `style` which doesn't expect a "CSS-like" (for the lack of a better name) string; which would cause Prettier to throw an error. 

Perhaps, at least for Angular, Prettier could "try" to make the change, but if it somehow "throws", leave the content as-is. Of course, Angular users could always use `[style]="'whatever'"` instead of `style="whatever"` to work around the Prettier formatting.

**Prettier 1.15.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIDOMAngDZwC8AOiLrZCRAE65K6NzoDclUtf-fLPgAOJAIZF+STFBIy4uagD4etVAHos2JSAA0ICMJiZo+ZKDGNGEAO4AFSwjMox2CJnR6QAI0ZiwANZwMADKwv4yAObIMIwArnD6ABYwALYkAOpJmPAi-nAhTjk4OUTI4Phm+jL4cIwwdn6RqWLIAGZiJLX6AFb4AB4AQn6BwSFiqXAAMvLtnd0gff0hUWQAinEQ8HNdiSDhjLWM5SnpXsKMMjAZHjBJyAAcAAz6FxC1GX7C5RdwR9hwLzsACOcUw7EaYmarSQHV2+lqqUwMXie3wqzgGy2gNh8z2MDE3hu6DuyAATPpYmJMHIoJEAMIQVItcpQaCA-RxWoAFUJzjhtQAvoKgA)
```sh
--parser html
```

**Input:**
```html
<div style="   color : red;
            display    :inline ">
  </div>
```

**Output:**
```html
<div
  style="   color : red;
            display    :inline "
></div>

```

**Expected behavior:**
```html
<div style="color: red; display: inline"></div>
```

thorn0:
It's a feature request (formatting the `style` attribute), not a bug.

nunocastromartins:
Maybe also worth discussing (or even worth opening another issue): what about other attributes?
For example, `class`:
```html
<div class=" a bbb    cc  
ddddd  "></div>
```
Could be formatted to:
```html
<div class="a bbb cc ddddd"></div>
```
[...]

PR #7555, opened 2020-02-06
Format value of HTML `class` attribute

fisker:
Refs: #5479 #5482


changelog_unreleased/html/pr-7555.md
```diff
@@ -0,0 +1,17 @@
...
+
+<!-- Prettier stable -->
+<div
+  class="  foo
+bar baz"
+></div>;
+
+<!-- Prettier master -->
+<div class="foo bar baz"></div>
+```
```
evilebottnawi: Can you add example with many clases when we don't fit in line?

fisker: done

evilebottnawi: /cc @prettier/core What do you think, maybe we should print each class on newline if we don't fix in line for readability?

lipis: hmm.. that's quite an edge case.. but maybe ignore the filelength in that case?

```html
<div
  class="another element with so many classes even can not fix one line really a lot and lot of classes"
></div>
```

WDYT?

fisker: @evilebottnawi I tried print each on newline , seems not good, it's common to have many classes, and we even don't print css on newline (#7556) , css normally longer then class names.

fisker: @lipis I don't think this is a edge case, and not a good idea to ignore the line length

[This case](https://deploy-preview-7555--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeAFjAtgGwHwA6UABMagEYQAmAnoScUaalQJYBuTxYOAhgM78AvARC8AtOThZJOCGADWxCVJnk5i8eIDurGOnFZqrAGZwAToxBE8zNpwY8Bw0Sumz5CgPpe4OaQgwypLu6p7ifgGwWrr6hsZmlrw4-BDivFAQ+hYeiqI2LBxcToIiYiFqGt6+-liBwaq5ChG1gTF6BkZsicopaRlZ6Dlhij6RdbD5UHioAPT29KTM8xyLZCvseCAANCAQAA4wrND8yKC85uYQ2gAKFwinKMnavDSnu+TmvIpwMADK+2+rCgAHNkDBzABXOC7TC4ADq6D0cH4gLAcD+Dz0HD0NGQ4EEOxAwP4FhgNy+IKwvGQJmSpN2ACt+AAPABCXx+-14dQAMsC4LT6TCQMyWX9gSD-ABFSFZQVIOkpEWA8yk8z4uE4In7czAmDw1hUfTIAAcAAZdrqIKT4V99vjdSiLOxBbsAI5y+AUg6PMT8cRQOBwKghonmOCe1gRim8Kk0xXC3akrCscFQkX8SUyr0KpUMkAwXjkQ3G9DIABMuwhvFYOElAGEIFhqfiUQBWImQ0kAFWLj3zIvY0IAklBQ7A-mA9YcAILjv4wGj+IXKgC+a6AA) , I think could be a real world case.

evilebottnawi: I think respect line width in that cases is bad idea, maybe print it in one line always is better

evilebottnawi: It is really hard to read and will create big git diff when you add a new class

fisker: Actually I'm ok with both, so just `.split(/\s/).join(' ')` ?  So simple, :)

evilebottnawi: Sounds good for me to inline that, will be great to hear other opinion

PR #7565, opened 2020-02-08
Test second format

### LINKED ISSUE #7055: Idempotence testing: prettier(code) == prettier(prettier(code))
The tests don't contain such a check as of now. Let's add it.

Previous discussions: https://github.com/prettier/prettier/issues/2803#issuecomment-329671900, https://github.com/prettier/prettier/issues/3512#issuecomment-352622275

fisker:
Fixes #7055  (Can't link issue, maybe need close it manually)


j-f1:
What’s the status of this?

fisker:
@j-f1 need fix bugs

j-f1:
Do you want to try to fix all the idempotency issues here or figure out a way to ignore them and make an issue about it?

fisker:
I don't think we can fix all issues here. I've opened https://github.com/prettier/prettier/issues/7745 and 
https://github.com/prettier/prettier/issues/7746 ,  we can 

1. fix the failed tests first then enable this test
2. ignore them and fix later

j-f1:
How about gating the reformat test behind an env variable, then only enabling it for one run? (i.e. create a new job that only runs on Node 13 in dev mode and tests the second format) You could set its `continue-on-error` attribute to `true` so CI will still be green.

thorn0:
I think better to add an option to `run_spec` to disable this check for specific tests, like I did with the `disableBabelTS` option.

fisker:
We have filename, we can just simply list those failed files, and fix them later.

I'm AFK now, feel free to take over.

package.json
```diff
@@ -22,7 +22,7 @@
     "angular-html-parser": "1.4.0",
     "camelcase": "5.3.1",
     "chalk": "3.0.0",
-    "ci-info": "2.0.0",
+    "ci-info": "watson/ci-info#f43f6a1cefff47fb361c88cf4b943fdbcaafe540",
```
fisker: see https://github.com/watson/ci-info/pull/42#issuecomment-585014879

tests_config/run_spec.js
```diff
@@ -144,17 +142,12 @@ global.run_spec = (dirname, parsers, options) => {
...
-      !TEST_CRLF
+      !TEST_CRLF &&
+      !unstableTests.has(filename)
     ) {
       test(`${filename} second format`, () => {
         const secondOutput = format(output, filename, mainOptions);
-        if (unstableTests.has(filename)) {
-          // To keep eye on failed tests, this assert never supposed to pass,
-          // if it fails, just remove the file from `unstableTests`
-          expect(secondOutput).not.toEqual(output);
```
thorn0: Seemed to be a good idea. Why have you decided to remove this?

fisker: I forgot which file, but it formats twice with different options in `jsfmt.spec.js` run this twice, and one fail one pass, can't assert.

fisker: I'll add it back

fisker: Done.

fisker:
@j-f1 I can't request review from this PR, so friendly ping.

tests_config/run_spec.js
fisker: @thorn0 I think this is a better way to manage these todos, we should refactor `disableBabelTS` option to this way.

fisker: #7786

thorn0: I'm not sure this is a better way. Why do you think so?

thorn0: I mean we can keep these assertions and still use an option to mark unstable/disabled/etc tests? Why clutter `run_spec.js` with these lists?

fisker: > I'm not sure this is a better way. Why do you think so?

I thought they are todos

lydell:
Oh, wow. Having tests for second format is going to be so good! 👏

PR #7587, opened 2020-02-14
Check pattern existence before glob

### LINKED ISSUE #6854: Issue finding/writing files with brackets ([]).
**Environments:**
- Prettier Version: 1.18.2
- Running Prettier via: CLI, CLI via Editor Tool (WebStorm File Watcher)
- Runtime: Node v12.13.0
- Operating System: macOS

**Steps to reproduce:**
```fish
prettier --write src/pages/posts/[id].jsx
```

**Expected behavior:**
It's expected to find, parse, update & write the file, as a prettier does.

**Actual behavior:**
```fish
[error] No matching files. Patterns tried: src/pages/posts/[id].jsx !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
```

There is another issue previously opened about this (#6344), I am opening this one because the other is closed & locked so can't be commented on.

It was said in the above referenced issue that this was probably an issue for `lint-staged` instead of `prettier` but I feel a bit differently about that given that this—at least to my imagination—affects tooling of many kinds. I could escape the file path like this: `"src/pages/posts/\[id\].jsx"` to get prettier to work, but that isn't really possible for any tools that insert an arbitrary file path into the CLI command. In my case I'm using a File Watcher in WebStorm… this is really just using the CLI with whatever the current file path is.

fisker:
Fixes #6854


src/cli/util.js
```diff
@@ -397,33 +399,113 @@ function createIgnorerFromContextOrDie(context) {
...
+    const absolutePath = path.resolve(cwd, pattern);
+    const stat = statSafeSync(absolutePath);
+
+    if (
+      stat &&
+      stat.isDirectory() &&
+      // `dot pattern` and `expand directories` support need handle differently
+      // for backward compatibility reason only expand `directories` like a glob pattern
+      // see https://github.com/prettier/prettier/pull/6639#issuecomment-548949954
+      isGlobPattern(pattern)
```
evilebottnawi: Just infromation: `test*` can be directory and can be glob :smile: But maybe it is good idea to check on directory firstly, also we can use `https://github.com/mrmlnc/fast-glob#isdynamicpatternpattern-options` to check on glob, I think that's why I decided that we need to do an update first

fisker: @evilebottnawi `is-glob` is original from `eslint`, we don't have `fast-glob` yet, we don't want both `fast-glob` and `globby`, we can switch to `globby` function later https://github.com/sindresorhus/globby#globbyhasmagicpatterns-options

evilebottnawi: Yes, agree :+1:

fisker:
@evilebottnawi ready for review.

src/cli/util.js
```diff
@@ -397,33 +399,129 @@ function createIgnorerFromContextOrDie(context) {
...
+  const patterns = [];
+
+  for (const pattern of maybePatterns) {
+    const absolutePath = path.resolve(cwd, pattern);
+    const stat = statSafeSync(absolutePath);
+    if (
+      stat &&
+      path
+        .relative(process.cwd(), absolutePath)
+        .split(/[\\/]/)
```
thorn0:
```suggestion
        .split(path.sep)
```
Linux file names can contain backslashes.

src/cli/util.js
```diff
@@ -397,33 +399,129 @@ function createIgnorerFromContextOrDie(context) {
...
+    if (
+      stat &&
+      stat.isDirectory() &&
+      // `dot pattern` and `expand directories` support need handle differently
+      // for backward compatibility reason only expand `directories` like a glob pattern
+      // see https://github.com/prettier/prettier/pull/6639#issuecomment-548949954
+      isGlobPattern(pattern)
+    ) {
+      files = [
+        ...files,
```
thorn0: What's wrong with `files.push(...`? Why reassign the variable?

fisker: I considered use that, but is there arguments length limit in this? I'm  not sure.

thorn0: That's an interesting concern. Indeed, the spread syntax seems to be able to handle more elements. On my machine:
```console
> [].push(...Array(1e5))
100000
> [].push(...Array(2e5))
RangeError: Maximum call stack size exceeded
> [...Array(2e5)]
[ undefined,
  undefined,
  undefined,
...
```
Okay, let's keep it this way.

fisker: This is also why people don't use `String.fromCharCode.apply(String, codePoints)` to build string.

Alternate way, loop and `.push(...x.slice(1e5))` one time.

src/cli/util.js
thorn0: I don't understand this `if`. Why include `<dir>/**/*` if `<dir>` looks like a pattern, but is a directory? I read https://github.com/prettier/prettier/pull/6639#issuecomment-548949954, but it doesn't make it any clearer to me.

fisker: Run 
```console
$ npx prettier src

[error] No matching files. Patterns tried: src !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
```

We don't expand dirname, unless it's pattern like `!dir` and it's a dir, we search all files in this `!dir`, because this is we want to do in this pr, otherwise, we still say we don't glob anything in `src`

thorn0: Let's say we have two directories (with formattable files inside): `foo` and `[foo]`.
Now, both `npx prettier foo` and `npx prettier [foo]` give the same result: "No matching files".

Why does this PR change the behavior for `prettier [foo]`? Sorry if it's something obvious, but somehow I can't see it.

fisker: #6854 ask for support to find `[foo]` file, and @evilebottnawi want also support `[foo]` dir https://github.com/prettier/prettier/pull/6776#issuecomment-581363613

fisker: If we don't expand this `[foo]` dir , people will still not able to format all files inside `[foo]` , because `prettier [foo]/**/*` will not glob any file, he has to `prettier [foo]/1.js [foo]/2.js [foo]/3.js`, but we do this, he can `prettier [foo]`. And we are going to support `prettier foo` #6128 , just not in this one.

thorn0: That's why there exists escaping syntax in globs. fast-glob uses backslashes for that, which, unfortunately, makes the Windows story messier... I'll check how ESLint handles this.

thorn0: What a mess. The same command `eslint "x\[foo\]/*.js"` matches different files on Linux (`x[foo] → foo.js`) and Windows (`x → [foo → ] → foo.js`).

thorn0: An [alternative escaping syntax](https://github.com/isaacs/minimatch/issues/64#issuecomment-258709901) works though: `eslint "x[[]foo]/*.js"`.

src/cli/util.js
```diff
@@ -397,33 +399,130 @@ function createIgnorerFromContextOrDie(context) {
...
+        ...extraPatterns
+      ].join(" ")}`
     );
-    // Don't exit the process if one pattern failed
     process.exitCode = 2;
+    return;
+  }
+  files = uniq(files.map(file => path.relative(process.cwd(), file)))
+    // keeping file orders for backward compatibility
+    .sort((a, b) => a.localeCompare(b));
```
thorn0: Is this compatibility thing important? This sorting prevents us from refactoring this function into a generator, which would be the best solution to the problems of big arrays.

fisker: @thorn0 It's not important, but makes it stable, it seem required for tests. And this order is not even good, I tried to find a better/stable sort package, but seems `sort-paths` `path-sort` `path-sort2` all not that popular. So I decide to ignore this now.

src/cli/util.js
thorn0: > If we don't expand this [foo] dir , people will still not able to format all files inside [foo] ,

So, as we have the escaping syntax, this concern about 

> And we are going to support prettier foo #6128 , just not in this PR.

Let's move all the work with directories to that PR. For now, in this PR, let's just skip directories. To format files inside `[foo]`, people have the escaping syntax.

src/cli/util.js
thorn0: Valid points.

> And this order is not even good

What's wrong with it?

fisker:
```text
a.js
b/a/a.js
b/b.js
c.js
```

I prefer by depth and file first

```text
a.js
c.js
b/b.js
b/a/a.js
```

Like

```js
.sort(
  (a, b) => (a.split('/').length - b.split('/').length) || a.localeCompare(b)
)
```

thorn0: I see, that's not what 1.19.1 does anyway.

docs/cli.md
```diff
@@ -17,9 +17,13 @@ In practice, this may look something like:
 prettier --single-quote --trailing-comma es5 --write "{app,__{tests,mocks}__}/**/*.js"
 ```
 
-Don't forget the quotes around the globs! The quotes make sure that Prettier expands the globs rather than your shell, for cross-platform usage. The [glob syntax from the glob module](https://github.com/isaacs/node-glob/blob/master/README.md#glob-primer) is used.
+> Don't forget the **quotes** around the globs! The quotes make sure that Prettier CLI expands the globs rather than your shell, which is important for cross-platform usage.
 
-Prettier CLI will ignore files located in `node_modules` directory. To opt-out from this behavior use `--with-node-modules` flag.
+Prettier CLI uses the [glob syntax from the `fast-glob` module](https://github.com/mrmlnc/fast-glob#pattern-syntax). To escape special glob characters, one of the two escaping syntaxes can be used: `prettier "\[my-dir]/*.js"` or `prettier "[[]my-dir]/*.js"`. Both match all JS files in a directory named `[my-dir]`, however the latter syntax is preferable as the former doesn't work on Windows, where backslashes are treated as path separators.
```
fisker: This `fast-glob` module is not true ... https://github.com/prettier/prettier/pull/6721

But we are going to merge https://github.com/prettier/prettier/pull/6776 , so it's not important.

How about link to `globby` ?

thorn0: Yes, I wrote that with #6776 in mind. We need a link that describes the syntax of the patterns. Globby doesn't have such docs.

fisker: Sure.

fisker:
> Let's add tests for `./directory/directory/directory/../../file.js`, `./directory/./file.js`

https://github.com/prettier/prettier/pull/7587/commits/74936476d8f8f363c7d2d02e37ebe6c236ed774d

evilebottnawi:
Let's fix conflicts

fisker:
WIP

src/cli/util.js
```diff
@@ -401,51 +402,106 @@ function createIgnorerFromContextOrDie(context) {
...
+    }
   }
+}
 
-  // The '!./' globs are due to https://github.com/prettier/prettier/issues/2110
-  const ignoreNodeModules = context.argv["with-node-modules"] !== true;
-  if (ignoreNodeModules) {
-    patterns = patterns.concat(["!**/node_modules/**", "!./node_modules/**"]);
+// `dot pattern` and `expand directories` support need handle differently
+// for backward compatibility reason temporary remove `.` and set `expandDirectories=false`
```
thorn0: I think it's forward compatibility, not backward, isn't it? :)

src/cli/util.js
```diff
@@ -401,51 +402,106 @@ function createIgnorerFromContextOrDie(context) {
...
+// see https://github.com/prettier/prettier/pull/6639#issuecomment-548949954
```
thorn0: This comment is still confusing to me. Yes, we're going to expand directories sometime in the future. But what does this have to do with backward compatibility?

fisker: Try read little more from https://github.com/prettier/prettier/pull/6639#issuecomment-548634927 , if this still  
 confusing you, I'll try to explain.

thorn0: I think I understand it, it's just that I wouldn't call it backward compatibility. Never mind.

Issue #7586, opened 2020-02-14
`--ignore-path` not respected when prettier executes in a subdirectory


heylookltsme:
**Environments:**
- Prettier Version: 1.19.1
- Running Prettier via: CLI
- Runtime: Node.js v13.8.0
- Operating System: macOS, etc.

**Steps to reproduce:**
Repo for reproducing: https://github.com/heylookltsme/prettier-bug

1. `.prettierignore` file at the root of a project
2. `.prettierrc` in subdirectories of the project
3. Run prettier in a subdirectory, passing the `--ignore-path`
4. The `.prettierignore` file is not respected

Correct behavior when run from the root:
<img width="794" alt="corinne_hoener_vimeo____projects_prettier-bug-2" src="https://user-images.githubusercontent.com/1299305/74543467-6ecb1f80-4f13-11ea-969f-0a056abbc1c5.png">

Bad behavior when run from a subdirectory:
<img width="782" alt="corinne_hoener_vimeo____projects_prettier-bug_web1" src="https://user-images.githubusercontent.com/1299305/74543521-81ddef80-4f13-11ea-8427-3b3b887c99d3.png">

**Expected behavior:**
The `.prettierignore` file is respected when prettier is run in a subdirectory. 

**Actual behavior:**
Prettier fails to ignore files when it's run in a subdirectory. 

It seems like the issue is that `.prettierignore` must be at the root of the repo, but if prettier is executing in a subdirectory,  then the ignored patterns aren’t matching because it’s comparing the file path relative to the _current working directory_, not to the root where the .prettierignore file lives. 

I _think_ I have a fix. [This fixes](https://github.com/prettier/prettier/compare/master...heylookltsme:fix-prettierignore-bug) the issue for me locally. I'll submit a PR for feedback. 

Thanks!

thorn0:
`--ignore-path` seems to have been inspired by ESLint. How does ESLint behave in this situation?

heylookltsme:
Good question! 

ESLint behaves as I would expect. `.eslintignore` is not respected when eslint is run from a subdirectory, _unless_ `--ignore-path` is passed, in which case it _is_ respected. 

<img width="957" alt="corinne_hoener_vimeo____projects_prettier-bug_web1" src="https://user-images.githubusercontent.com/1299305/74657169-a9c09380-515d-11ea-8222-6845c3611a65.png">

thorn0:
Fixed by #7588

heylookltsme:
[...]

PR #7588, opened 2020-02-14
Respect prettierignore when executed in subdirectories

heylookltsme:
This fixes https://github.com/prettier/prettier/issues/7586. It changes the filename we use in `ignore.filter` to be relative to the location of the `ignore-path` and not the current working directory. 

This is my first exposure to the prettier source code, so please school me! This fixes the issue I have locally, but I have no idea if other changes might be needed or if there are other scenarios I'm not considering here. Thanks!

- [ ] **N/A** (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)

### DIFF
--- a/changelog_unreleased/cli/pr-7588.md
+++ b/changelog_unreleased/cli/pr-7588.md
@@ -1,0 +1,4 @@
+#### Respect `--ignore-path` when prettier executes from a subdirectory ([#7588](https://github.com/prettier/prettier/pull/7588) by [@heylookltsme](https://github.com/heylookltsme))
+
+Changes the filename used when filtering ignored files to be relative to the
+`--ignore-path`, if present, rather than the current working directory.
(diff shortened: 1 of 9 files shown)


heylookltsme:
@evilebottnawi Will do 👍

heylookltsme:
@evilebottnawi Tests added 👍 Please let me know if you have any suggestions about anything, particularly naming conventions. I'm not in love with `ignore-in-subdirectories`, but it's probably clear enough. 

Passes in branch:
corinne_hoener_vimeo____projects_prettier

Fails in master:
corinne_hoener_vimeo____projects_prettier-2

heylookltsme:
I realized I also needed to cover the case where `--stdin-filepath` is being used so I've updated to include that. 

I'm not entirely sure what I need to do in order to get the codecov checks to pass. I'm guessing it's failing because I'm adding additional conditionals and not providing coverages. But I'm not what what tests I need to edit or add. Any guidance on that would be much appreciated!

heylookltsme:
Thanks @thorn0! Added 👍

Issue #6107, opened 2019-05-07
Improve formatting of chain of arrow functions


mgtitimoli:
**Prettier 1.17.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBzOMCSkUAolNgIa4C2CMAKgJ4AOcAYgE4RWYC8mAFAgrVajFmiSYAgu3bkGAHlLC4NWGLgA+AJS9NAHSiZMASyLLKq0czh7MQy2vo20AOhNQwAGwCuAEzg0fjNoCxF1G20AbkMQABoQCCYYE3RkUHJZCAB3AAVMhDRkEHIANwgTP3iQACM5MABrPABlJnIwD2xkGHYfOASACxgqLwB1AZN4NDawOGbCyZNSyYZisDQihI80OHYYXLlsKnJkADNyLx2EgCs0AA8AIXqmmGbyGgAZDzgzi6uQW53ZqdLxwACKPgg8F+l36IDa7B27GKNXINQYXmg1SY7A8MFGlRgA2QAA4AAwJHEQHajORMYo4wK7Uo-BIAR0h8AOSSKKHIaAAtFA4HAAlUEuw4ByTJKDpRjjD-jsqCZur04WgQeDOT8kOdYQkYGiCX4icgAEyGuQmLydADCXAVKCg0FZIB8OzoaN5+p2AF8-UA)

**Input:**
```jsx
const getIconEngagementTypeFrom = (engagementTypes: Array<EngagementType>) => 
  iconEngagementType => engagementTypes.includes(iconEngagementType);
```

**Output:**
```jsx
const getIconEngagementTypeFrom = (
  engagementTypes: Array<EngagementType>
) => iconEngagementType => engagementTypes.includes(iconEngagementType);
```

**Expected behavior:**
```jsx
const getIconEngagementTypeFrom = (engagementTypes: Array<EngagementType>) => 
  iconEngagementType => engagementTypes.includes(iconEngagementType);
```
---
I'm opening this one, after having asked @vjeux about his thoughts and [validated](https://twitter.com/Vjeux/status/1125627702516387840) he also thinks chain of arrow functions should be formatted this way.

thorn0:
I'd probably prefer for such arrow chains to be formatted this way:
```ts
const getIconEngagementTypeFrom =
  (engagementTypes: Array<EngagementType>) => 
  iconEngagementType => 
  engagementTypes.includes(iconEngagementType);
```

[...]

Issue #4166, opened 2018-03-16
Better handling or better options for curried function definitions


mike-robertson:
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBbAngYQFcAnYgSzgBMAxKTAXkwENiBzARgYD5MAdKXjBasATNz4CYg4QGZxwzgGpmbMctkBuEABoQEAA4wy6ZKBbEIAdwAKLBGmQgmANwhlKOpxmQAzJgBs0OF0AI2ImMABrOBgAZX0IsihWZBhiQmCQSggwXwCg3SSg4hhrcNZsJjzAzIArNAAPACFwqJjYpmw4ABkkuGqCvUIYfWGRAcyE4mLHEKYQ3H9oT31yWAB1dxgAC2QADgAGXVWIIPXw-UdVuGLnft1iOABHQjJHsqYKqqQ-Gt0g7BkVLpTJoJKsfxwACKhAg8AmuiEIU2lB2yBEiPCZH84PwEGwlUcUGg9xAhCCABV5g4fvk4ABfelAA)

Example 1:
**Input:**
```jsx
const myCurriedFn = arg1 => 
  arg2 => 
    arg3 => arg1 + arg2 + arg3;
```

**Output:**
```jsx
const myCurriedFn = arg1 => arg2 => arg3 => arg1 + arg2 + arg3;

```

**Expected behavior:**

I would want this to be preserved because it's much easier to read a curried function like this rather than as one long line. You can easily see how many invocations it requires.

```jsx
const myCurriedFn = arg1 => 
  arg2 => 
    arg3 => arg1 + arg2 + arg3;
```

Example 2:
**Prettier 1.11.1**
[...]

PR #7589, opened 2020-02-14
Declared arrow fn

qlonik:
Initial work on issues #6107 and #4166. Format is not finalized in issues, so the pull request is not final either. 

The last commit of this pr, shows changes between how prettier prints functions with these changes, instead of one possible desired printing style.

Things that are not done, which would be nice to do:
 - [ ] each level of curried function should be aligned, instead of indented at each level
 - [ ] return type of the very last function should go to the new line too
 - [ ] function implementation does not break on `=>` right now


qlonik:
Seems like most of this work is done with #9992

Issue #2835, opened 2017-09-15
CSS: `url(unquoted)` is broken: Can be lowercased, normalized and broken across lines


sampi:
**Prettier 1.7.0**
[Playground link](https://prettier.io/playground/#%7B%22content%22%3A%22%40font-face%20%7B%5Cn%5Ctfont-family%3A%20'Some%20Font'%3B%5Cn%5Ctfont-style%3A%20normal%3B%5Cn%5Ctfont-weight%3A%20300%3B%5Cn%5Ctsrc%3A%20local('Some%20Font')%2C%20url(data%3Aapplication%2Ffont-woff%3Bcharset%3Dutf-8%3Bbase64%2CThisIsNormalBut%2F%2B0ThisIsLowerCased)%20format('woff')%3B%5Cn%7D%22%2C%22options%22%3A%7B%22ast%22%3Afalse%2C%22bracketSpacing%22%3Atrue%2C%22doc%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22output2%22%3Afalse%2C%22parser%22%3A%22postcss%22%2C%22printWidth%22%3A80%2C%22semi%22%3Atrue%2C%22singleQuote%22%3Afalse%2C%22tabWidth%22%3A2%2C%22trailingComma%22%3A%22none%22%2C%22useTabs%22%3Atrue%7D%7D)
```sh
--parser postcss
--use-tabs
```

**Input:**
```css
@font-face {
	font-family: 'Some Font';
	font-style: normal;
	font-weight: 300;
	src: local('Some Font'), url(data:application/font-woff;charset=utf-8;base64,ThisIsNormalBut/+0ThisIsLowerCased) format('woff');
}
```
[Full example on Gist](https://gist.github.com/sampi/33248fa89ebbe01ff0bbc5063052424f)

**Output:**
```css
@font-face {
	font-family: "Some Font";
	font-style: normal;
	font-weight: 300;
	src: local("Some Font"),
		url(data:application/font-woff;charset=utf-8;base64,ThisIsNormalBut/+0thisislowercased)
			format("woff");
}

```
[Full example on Gist](https://gist.github.com/sampi/d2826fe92402f2c2b2dbf81ffa1d053e)

**Expected behavior:**
Some characters are lowercased inside the base64-encoded `@font-face` `url()` rendering them useless.
Anything inside `url()` should be left untouched.

lydell:
[...]

PR #7592, opened 2020-02-16
CSS: Fix url() unquoted content manipulation (#2835)

mattiacci:
Improve the handling of unquoted URL content in CSS/SCSS/Less. This doesn't
address the underlying parsing issues, but at least ensures Prettier doesn't
modify URLs, by stringifying more cases of unquoted content in `url()`.

(Not changing the API or CLI)


src/language-css/parser-postcss.js
```diff
@@ -92,29 +97,70 @@ function parseValueNodes(nodes) {
...
+}
 
-  return before + value + unit + after;
+function groupListStartsWithSassVariable(groups) {
+  return !!(
+    groups &&
+    groups[0] &&
+    groups[0].type === "word" &&
+    groups[0].value.startsWith("$")
+  );
```
evilebottnawi: Please move that to utils and rename to `isScssVariable`

src/language-css/parser-postcss.js
```diff
@@ -92,29 +97,70 @@ function parseValueNodes(nodes) {
...
+        groups[i - 1].value.endsWith("#")
+      ) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+function groupListContainsStringOrFunctionNode(groups) {
```
evilebottnawi: Please move that to utils and rename to `isFunction` (maybe best name)

src/language-css/parser-postcss.js
```diff
@@ -92,29 +97,70 @@ function parseValueNodes(nodes) {
...
+        groups[i].value === "{" &&
+        groups[i - 1].type === "word" &&
+        groups[i - 1].value.endsWith("#")
+      ) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
```
evilebottnawi: Please move that to utils and rename to `hasScssInterpolation`

src/language-css/parser-postcss.js
```diff
@@ -92,29 +97,70 @@ function parseValueNodes(nodes) {
...
+  const before = (node.raws && node.raws.before) || "";
+  const quote = (node.raws && node.raws.quote) || "";
+  const atword = node.type === "atword" ? "@" : "";
+  const value = node.value || "";
+  const unit = node.unit || "";
+  const group = node.group ? stringifyNode(node.group) : "";
+  const after = (node.raws && node.raws.after) || "";
+
+  return before + quote + atword + value + quote + unit + group + after;
+}
```
evilebottnawi: Move that to printer/util

src/language-css/parser-postcss.js
mattiacci: I'll see what I can do, but note that these new utility functions, along with some existing ones, aren't very portable.
For example, I wanted to use isURLFunctionNode on line 27, instead of manually checking the type and value, but isURLFunctionNode only works in the printer, where the parser has already processed the nodes and converted "func" to "value-func".
Should I be specifying somewhere whether these will work in the parser, printer, or both (or should there be separate utils for these cases)?

evilebottnawi: You can use name `isGroupListLikeStringOrFunction` (maybe less name)

src/language-css/parser-postcss.js
mattiacci: Moved to utils and renamed to `isSCSSVariable` (to match local convention).

src/language-css/parser-postcss.js
mattiacci: Moved to utils and renamed to `hasStringOrFunction`.

src/language-css/parser-postcss.js
mattiacci: Moved to utils and renamed to `hasSCSSInterpolation` (to match local convention).

src/language-css/parser-postcss.js
mattiacci: Moved to util.

thorn0:
@mattiacci Looks good. Please add code from https://github.com/prettier/prettier/issues/5654 as a test.

src/language-css/parser-postcss.js
```diff
@@ -23,23 +31,32 @@ function parseValueNodes(nodes) {
...
+      if (
+        hasSCSSInterpolation(groupList) ||
+        (!hasStringOrFunction(groupList) && !isSCSSVariable(groupList[0]))
+      ) {
+        const stringifiedContent = stringifyNode({
+          groups: node.group.groups,
+        });
+        node.group.groups = [stringifiedContent.trim()];
+      }
+    }
```
evilebottnawi: Not sure it is a good solution, we really should fix it on parser side, it is still output broken url with interpolation and variables

thorn0: Broken in what way? Not formatted? Or really broken? Show an example please.

thorn0:
I found a regression in CSS-in-JS.

**Prettier pr-7592**
[Playground link](https://deploy-preview-7592--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEBnGBPANnAJgOlwEsA3AA0iwgCclq8BuAIwEMwBrAc2ogFcpcSAAS9qWABQALGDAAOSAPQK4ADxYBbWTnyR1AfgCOAXgAkwAGYQIAXwCUZEABoQEWTCLRUyUC2o8A7gAKvgheKCxY-iwYXs5M1GzscDAAyrJsRFCcyDDUvHDO0upYAOqSRPCo6WBwKaEVpBUYyOCosSCZqHDUMIEJnOosyOYRXc4AVqgqAEIJHMkpGnAAMplww6MFIJMqKZmcOACKvBDwG1hjIOnUXdQtrExwWE5X1JkwJUS4MJLIABwABmcsh4XRKCVkLRBcFuJHWzgMJ3gfVcYRALFQAFooHA8HgXvREUR6H0WAMhkgRhctl11EQcnkafsjkj1pTNs4YCwmJ9vr8kAAmTkJIhYfYAYQg6kGLRhAFYXrwugAVblhKmXEj5ACSAgQqTAbzcAEEBClMDhzl1rNYgA)
```sh
--parser babel
```

**Input:**
```jsx
styled.div`color:red;background: url(http://example.com?q=${foo})`
```

**Output:**
```jsx
Error: Couldn't insert all the expressions
    at transformCssDoc (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:24094:13)
    at Object.embed$3 [as embed] (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:23938:20)
    at Object.printSubtree (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:16794:30)
    at callPluginPrintFunction (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:16915:33)
    at https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:16879:49
    at Object.printComments (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:16552:21)
    at printGenerically (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:16879:24)
    at FastPath.call (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:16672:22)
    at printPathNoParens (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:27753:92)
    at Object.genericPrint$3 [as print] (https://deploy-preview-7592--prettier.netlify.com/lib/standalone.js:26247:32)
```

mattiacci:
Thanks for bringing up the [regression](https://github.com/prettier/prettier/pull/7592#issuecomment-606688062), but I'm not sure the best approach to address it - it's my first time contributing to Prettier.

It seems Prettier uses atwords to represent the interpolated JS variables, and that case you brought up is broken by this PR. That said, it seems [remarkably easy](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBnGBPANnAJgOlwEsA3AAwCMBDMAawHMAnCAVylyQAIXGsAKADogAFjBgAHJAHopcAB5UAtuJz5IigPwBHALwASYADMIEAL5CAlGRAAaEBHEwi0VMlBVGzAO4AFDwlcUKiwvKgxXOwpGGlo4GABlcRoiKHpkGEYWODtRRSwAdWEieFQksDh4gOLSYoxkcFQIkBTUOEYYH2j6RSpkQ2DWuwArVDkAIWi6OPilOAAZFLg+geyQEbl4lPocAEUWCHhlrEGQJMZWxnrqCjgsW1PGFJh8olwYYWQADgAGO3FmVr5aLier-OAXEhLOxafbwToOQIgKioAC0UDgeDw90YcBhRBxnSo3V6SH6x1WrUURHSmQpW12sKWpJWdhgVAoLzeHyQACZWdEiFgtgBhCCKHr1cEAVnuLFaABV2YEyScSFkAJLsBAJMCPRwAQXY8UwOCOrVMpiAA) to break Prettier in the css-in-js interpolated string use today:
```
styled.div`background: url("http://example.com?q=${foo}")
```

This presumably happens because the JS variable is replaced with the Prettier placeholder, resulting in the CSS parser seeing:
```
styled.div`background: url("http://example.com?q=@prettier-placeholder-1-id")
```

But Prettier seems to assume that the placeholders will be parsed as at `atnode`, so it doesn't find the placeholder and fails. My change triggers the same behavior, since it stringifies the unquoted URL.

Perhaps Prettier should look at the raw string output for the placeholders, rather than using the node tree? Something else?

thorn0:
The CSS-in-JS regression doesn't happen anymore because an underlying issue has been fixed by #8097.

tests/css/inline-url/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -24,26 +24,88 @@ parsers: ["css"]
...
+.validUnquotedUrlsThatAreParsedBySass {
+  background: url($foo);
+  background: url($foo + $bar);
+  background: url($foo*3);
+  background: url($foo/$bar);
+}
+
+.invalidUnquotedUrlsButWeParseThemAnyway {
+  background: url(--var(foo-bar, #dadce0));
+  -fb-sprite: url(fbglyph:cross-outline, fig-white);
```
evilebottnawi, line 108: Let's add tests with nested function
```scss
$var: "http://example.com/img.png";

@function concat($numbers...) {
  $sum: '';
  @each $number in $numbers {
    $sum: $sum + $number;
  }
  @return $sum;
}

a {
 backround: url(concat('http://', 'example.com', '/image.png'));
}
```

thorn0:
@mattiacci Thanks for squashing this ancient bug!

Issue #7615, opened 2020-02-17
Remove the version parameter from prettier.getSupportInfo()


thorn0:
This [parameter](https://prettier.io/docs/en/api.html#prettiergetsupportinfoversion) adds a significant maintenance burden (see #4097, #6887), but its value is dubious. What's the use case? I found only that it was used by prettier-vscode (see #5822), but then they understood that it didn't make much sense and stopped using it. I have also checked all the other editor integrations listed on the main page of Prettier's site. They don't use it. So let's get rid of this parameter, shall we?

`getSupportInfo` was added in https://github.com/prettier/prettier/pull/3033.

evilebottnawi:
Previously many plugins recommend to install prettier globally, so developers don't know what is support what is not, and this function allow to them know about supported features based on versions, I don't have strong opinion on that, so if most of core developers vote :+1: I voting :+1: too

thorn0:
I don't really understand. If there is a global instance of Prettier and a local one, you just call `getSupportInfo()` (without arguments) on each of them to see what they support. Why use the parameter?

evilebottnawi:
@thorn0 I tried to find an example of use, but could not, let's remove it

lydell:
This was done in #7620.

PR #7620, opened 2020-02-18
api: remove the version parameter of getSupportInfo

thorn0:
closes https://github.com/prettier/prettier/issues/7615

### DIFF
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -14,7 +14,7 @@ function getSupportInfo({
   showUnreleased = false,
   showDeprecated = false,
   showInternal = false
-}) {
+} = {}) {
   // pre-release version is smaller than the normal version in semver,
   // we need to treat it as the normal one so as to test new features.
   const version = currentVersion.split("-", 1)[0];
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -9,16 +9,12 @@ const arrayify = require("../utils/arrayify");
 const currentVersion = require("../../package.json").version;
 const coreOptions = require("./core-options").options;
 
-function getSupportInfo(
-  _,
-  // This has to be the second parameter. See _withPlugins in src/index.js.
-  {
-    plugins = [],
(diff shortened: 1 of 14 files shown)


src/main/support.js
```diff
@@ -9,30 +9,26 @@ const arrayify = require("../utils/arrayify");
...
+    showUnreleased = false,
+    showDeprecated = false,
+    showInternal = false
   }
-
-  const { plugins } = opts;
+) {
+  // pre-release version is smaller than the normal version in semver,
+  // we need to treat it as the normal one so as to test new features.
+  const version = currentVersion.split("-", 1)[0];
```
fisker: I never know split takes a second argument.

thorn0: It's not super useful anyway.

fisker:
We may want switch those support infos to JSON scheme some day.

thorn0:
@fisker Would we need that `version` parameter for that?

fisker:
I agree that we don't need it. I mean after removing this, support info is just some property validate info, we can use ajv to validate.

src/main/support.js
```diff
@@ -9,30 +9,26 @@ const arrayify = require("../utils/arrayify");
...
-    showInternal: false,
-    ...opts
-  };
-
-  if (!version) {
-    // pre-release version is smaller than the normal version in semver,
-    // we need to treat it as the normal one so as to test new features.
-    version = currentVersion.split("-", 1)[0];
+function getSupportInfo(
+  _,
```
fisker: Can we remove this parameter, and wrap in the export?

thorn0: That would break type inference and overall I think the code is cleaner (less confusing) the way it is now.

thorn0: Nevertheless, I did something with this.

thorn0:
Merging this to move forward with other changes.

Issue #4559, opened 2018-05-27
Remove --stdin CLI Flag


azz:
@azz
> [...] can't we remove `--stdin` entirely? What's the use case?
>
> For Prettier 2.0, I would imagine we could do:
>
> ```shellsession
> $ echo ".selector {color:red}" | prettier --stdin-filepath style.css
> $ # OR:
> $ echo ".selector {color:red}" | prettier --language css
> .selector {
>  color: red;
> }
> ```

@lydell 
> The `--stdin` flag seems to have been added in commit 9ffbd32723b7b9a5ad7bf454c113fe16436a862d (2017-01-11) as the only way to read from `--stdin`. commit 239521e97df824e477187c361c57cf9911fc6421 (2017-01-26) added stdin auto detection, but it looks like we've kept around the flag anyway since then.
>
> I can't think of any use case for the flag either.

We could just change this line:
https://github.com/prettier/prettier/blob/938d2f06be95012c0e512828958f2ee54a645958/src/cli/index.js#L54-L55

to:
```js
const useStdin = !hasFilePatterns;
```

lydell:
I think we still need to check `!process.stdin.isTTY` because we want just running `prettier` to print help, not wait for stdin to be typed be the user. Thinking about it – if you want to "type stdin", I think using the `--stdin` is the only way. Not sure if it is ever useful to do that, though. I always go with `echo 'something' | prettier` or make a temporary file. Less error prone and more shell history friendly.

azz:
Another option is using `-` as a file pattern instead of a CLI flag, like `tar` (and probably many others).

But if the issue is typing `prettier` and expecting the help, then we can just do:

```js
const useStdin = !hasFilePatterns && (argv["stdin-filepath"] || argv["language"]);
```

j-f1:
@lydell

> Thinking about it – if you want to "type stdin", I think using the `--stdin` is the only way.

You can always use `cat | prettier`.

j-f1:
Related: #4097

duailibe:
Let's add a deprecation warning this time..

azz:
This is for Prettier 2.0.

duailibe:
I know that.. doesn't mean we shouldn't warn users already. Kind of like react does it

azz:
Something like this?

```
[warn] Deprecated: --stdin will be removed in prettier@2, see https://github.com/...
```

thorn0:
closed by #7668

PR #7668, opened 2020-02-24
remove --stdin

thorn0:
Closes https://github.com/prettier/prettier/issues/4559
See https://github.com/prettier/prettier/issues/4097


thorn0:
`echo 123 | prettier --parser markdown`

thorn0:
> why eslint doesn't use same logic

Probably because ESLint isn't supposed to work as a stream transformer (`stdin → transformation → stdout`).

PR #7726, opened 2020-03-07
Swap minimatch for micromatch

stramel:
Swaps `minimatch` out for `micromatch`. [Why use micromatch?](https://github.com/micromatch/micromatch#why-use-micromatch)

`micromatch` is [optimized better](https://github.com/micromatch/micromatch#benchmarks) and has [better cross-platform support](https://github.com/micromatch/micromatch#why-use-micromatch).

This should likely improve the performance of `prettier` as a byproduct of better cross-platform support.

Switching as documented _should just be_ drop-in and call the correct function. [Switching to micromatch](https://github.com/micromatch/micromatch#switching-to-micromatch)

Support for directories with `[]` characters is something that `minimatch` has had issues with since 2014 and came up when I was working on a PR for `lint-staged`. Next.js started using `[param]` to denote dynamic routes and thus my long journey to this spot. [Dynamic Routes w/ Next.js](https://nextjs.org/docs/routing/dynamic-routes)

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "linguist-languages": "7.7.0",
     "lodash": "4.17.15",
     "mem": "5.1.1",
-    "minimatch": "3.0.4",
+    "micromatch": "4.0.2",
     "minimist": "1.2.0",
     "n-readlines": "1.0.0",
     "normalize-path": "3.0.0",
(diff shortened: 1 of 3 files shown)


fisker:
Does this still  improve the performance?

Issue #767, opened 2017-02-22
Import sorting


philipp-spiess:
In order to avoid unnecessary git changes, we sort our import statements using the default sort feature of Atom or VSCode. We do this by selecting all import lines and running the sort action.

This works okayish but breaks complelty when we break multiple imports into multiple lines, e.g:

```js
import {
  Foo,
  Bar,
  Baz
} from 'foobarbaz';
```

Since this will (obviously) be sorted to:

```js
  Bar,
  Baz
  Foo,
} from 'foobarbaz';
import {
```

Are there any plans to support sorting of import blocks using prettier?

The idea would be to detect blocks that consists of import lines like:

```js
import Bar from 'foo';
import Foo from 'foo';
import foo from 'foo';
import bar from 'foo';
import { 
  Foo,
  Bar
} from 'foo';
import * from 'foo';
import { Bar } from 'foo';
import {
  foo,
  bar
} from 'foo';
import { Foo } from 'foo';
import type Foo from 'foo';
import type Bar from 'foo';
import { bar } from 'foo';
import { foo } from 'foo';

```

and turn it to something like:

```js
import { bar } from 'foo';
import { Bar } from 'foo';
import { foo } from 'foo';
import { Foo } from 'foo';
import { 
  foo,
  bar
} from 'foo';
import {
  Foo,
  Bar
} from 'foo';
import * from 'foo';
import Bar from 'foo';
import bar from 'foo';
import foo from 'foo';
import Foo from 'foo';
import type Bar from 'foo';
import type Foo from 'foo';
```

demoneaux:
Prettier is just a code formatter and avoids actual changes to the code which modifies the AST and hence might affect execution. You should try to create a plugin to do that.

lydell:
Unfortunately, importing modules can have side effects so changing the order of imports is not safe. So I guess this one has to be closed for the same reasons as #323 and #766. You could use the [sort-imports](http://eslint.org/docs/rules/sort-imports) ESLint rule, though.

philipp-spiess:
Right, side effects, I kinda forgot about that. Thank you for your quick reply :) Consider this closed.

dolanmiu:
But could this be a feature, but then have an option to turn it off in `.prettierignore` or inline in the file?

I have noticed that I have to sort imports more often than not.

There is only a few files in projects with side effects from ordering. But 99% of the time, I have to sort...

j-f1:
[...]

PR #7758, opened 2020-03-12
docs: add prettier-plugin-organize-imports to list of community plugins

simonhaenisch:
- [ ] ~I’ve added tests to confirm my change works.~
- [ ] ~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~
- [ ] ~(If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.~

Just adding a plugin I wrote to organize tyepscript imports to the list of community plugins cause I find it quite handy and want other people to be able to discover it and benefit from it as well. Not sure whether this list is open to additions but I hope so!

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -316,6 +316,7 @@
         "Shigeaki",
         "Shinigami",
         "Simen",
+        "simonhaenisch",
         "singleline",
         "skratchdot",
         "smirea",
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -55,6 +55,7 @@ Providing at least one path to `--plugin-search-dir`/`pluginSearchDirs` turns of
 - [`prettier-plugin-solidity`](https://github.com/prettier-solidity/prettier-plugin-solidity) by [**@mattiaerre**](https://github.com/mattiaerre)
 - [`prettier-plugin-svelte`](https://github.com/UnwrittenFun/prettier-plugin-svelte) by [**@UnwrittenFun**](https://github.com/UnwrittenFun)
 - [`prettier-plugin-toml`](https://github.com/bd82/toml-tools/tree/master/packages/prettier-plugin-toml) by [**@bd82**](https://github.com/bd82)
+- [`prettier-plugin-organize-imports`](https://github.com/simonhaenisch/prettier-plugin-organize-imports) by [**@simonhaenisch**](https://github.com/simonhaenisch)
 
 ## Developing Plugins
 


kachkaev:
I'm not sure this plugin matches Prettier’s philosophy, so it probably should not be recommended to Prettier users. See context here: https://github.com/prettier/prettier/issues/767

simonhaenisch:
@kachkaev thanks for the quick reply. I get that the philosophy prevents it from going into the core of Prettier, but I didn't think that philosophy would extend to plugins... in fact, that's why I implemented this as a plugin, because plugins are the only way to extend Prettier beyond its philosophy... no offense, philosophy has its place, but it can sometimes be limiting in a practical sense and the plugins API allows to overcome that and make Prettier even more useful depending on the user's need (I like to think this is one reason the plugins API came along in the first place).

In your mentioned issue https://github.com/prettier/prettier/issues/767#issuecomment-281643014 there's this advice:

> You should try to create a plugin to do that.

And BTW, in the community plugins list there's already `prettier-plugin-package` and `prettier-plugin-packagejson` which sort json by keys, i. e. AST modifications, and `prettier-plugin-svelte` adds a few config options which is also against the philosophy.

I don't see a reason not to recommend any useful plugin, because after all, the user can decide which plugins they want to use. I was thinking of this list as an official "awesome-prettier-plugins". But I respect that the decision is up to the maintainers.

fisker:
There is a spellcheck issue

```text
npx -p cspell@4.0.31 cspell {bin,scripts,src}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md
npx: installed 83 in 13.727s
/home/vsts/work/1/s/docs/plugins.md:58:114 - Unknown word (simonhaenisch)
CSpell: Files checked: 228, Issues found: 1 in 1 files
```

lipis:
This plugin is changing the AST (abstract syntax tree) and it should be an ESLint plugin instead.

kachkaev:
@simonhaenisch given the points on Prettier's philosophy about ASTs, could you update [your plugin's READMe](https://github.com/simonhaenisch/prettier-plugin-organize-imports) by mentioning those points? I think it's important that people that see your plugin are aware of other options.

To me, it still feels more correct to sort imports with ESLint, not Prettier. Sorting imports in a js / ts file is not the same as using `prettier-plugin-packagejson`. The former is more likely to lead to bugs caused by ‘formatting’.

simonhaenisch:
@lipis does reopening mean that you're reconsidering? I know there's ESLint plugins but not all of my projects use ESLint, and I like to have this as part of the "prettification" toolchain in many projects (and judging by the feature requests for this, many others as well).

Here's the rationale: I wouldn't really consider the import statements part of the module's code, they are more like directives for the module system, kind of like the `#include` preprocessing directive in C or the `using` directive in C#. This plugin only affects Typescript files and doesn't affect dynamic imports or uses of `require` in the code. It just makes sure that the import statements stay consistent no matter who writes them and what their preferences are, and that's why me and my team have found this plugin very helpful. I think it's for the same reason that import sorting has been an often requested feature and the plugin API now makes it possible to add this functionality directly into Prettier if desired by the user.

There's nothing in the docs that mentions that plugins should not do AST changes (as I mentioned, some plugins doing that are already on the list). Don't want to repeat myself but I don't see a point in forcing plugins to follow the same philosophy as Prettier because the plugins API enables us to go even beyond that in user land (if the user wants it, because the user decides what plugins they want). It would be difficult to enforce anyway, because plugins might start affecting the AST after they have been added to the list 🤷🏻‍♂️ Or there might be an `awesome-unofficial-prettier-plugins` repo soon 😉

---

Another idea (might need a bit more maintenance): change the plugins list to a table with checkboxes, to indicate how "well" it follows the Prettier philosophy/scope/rationale, e. g.:

| Plugin | Author | No AST modifications | No options | ... |
|-|-|-|-|-|
| `prettier-plugin-organize-imports` | [@simonhaenisch](https://github.com/simonhaenisch) | ❌ | ✅ | |

(or just add some tags like "changes ast" to plugins who go beyond the scope)

---

Looking forward to a decision from the maintainers about this.

thorn0:
There is another thing about this plugin that bothers me. It overrides the `preprocess` method of the `typescript` parser, which is undefined now, but can be added in future versions of Prettier... That's why it seems to be more of a hack than using the Plugin API the way it's supposed to be used.

I guess we still can add a link to it to the docs, but under a separate subheader, e.g. "Crazy hacks you can do with plugins", not under "Community plugins". It shouldn't look like such hacks are officially supported.

simonhaenisch:
Ok thanks for the hint, it respects an existing typescript preprocessor now:

https://github.com/simonhaenisch/prettier-plugin-organize-imports/commit/5ccd0c95099250091c74f69b32196d4086df7f81

(released as `1.0.2`). Let me know if there's anything else that makes you consider this a hack.

lipis:
I like the idea of changing the plugin list to a table!

lipis:
But that can come as another PR..

Issue #6122, opened 2019-05-14
JSDoc typedef parentheses are removed if member access is done on a typedef'd object variable


rsimha:
**Prettier 1.17.1**

**Before:**
```
foo = /** @type {!Foo} */ (/** @type {!Baz} */ (baz).bar );
```

**After:**
```
foo = /** @type {!Foo} */ /** @type {!Baz} */ (baz.bar);
```

Playground link [here](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VACA8gIwFZxgwYCGYYcAzpRtBjAJ4AOcAJnAGYA6U6GA6gCdoAcx4cIEDAF4MqAFTyMAAUYsMwAIQAxSQF8M8zAAoFS1czgbNAIRIAvA0YzGcDgJQA6N4IzuA3Dx8AMIQgoKEMOKSMnKKKmpWWroQTiam8RbqWnaOhiZu9u5ePn6BUDw8IAA0IBBMMACW0JTIoCThEADuAAodCK0oJABuEI2sNSA4gmQA1nAwAMpMZI1QIsgwggCucLUAFjAAtgA2-PuN8JQrFIsDl43DlwzI4NSTa5RwgjA9MyJHEjIDgkE5fWp4SgADxsMzA8yWJCOcAAMms4MDQeCQJCoYs1iITnAAIrbCDwTFgvYgFaCL6CV5uHAME7QSZMQRrGD8cYwfbIAAcAAZahyIF8hCQmK8OVRvsMMbUAI5k+B-eqDEAkSgAWigcDYbEmERVjQifxIAKBSBBVNqXyOjU2O2plAJRNJ5IxNqx1JgJBwPNYfOQACZalsSI0TgTQkdAa8oNBFSBtl8ACoBwa2r56PRAA)

Related to #4799

thorn0:
A more isolated bug report:

**Prettier 1.19.1**
[...]

PR #7782, opened 2020-03-17
Fix printing of nested closure typecasts

jridgewell:
Fixes https://github.com/prettier/prettier/issues/6122.

This goes through the rather complicated steps to determine when parens surround a typecast comment, including when both the node and the ancestor have typecasts.

I had to add _really_ hacky comment nodes in order to insert the parens in the correct locations. This is because using the normal `needsParens` would insert multiple typecasts comments before the first `(`. It order to get it right, we need to print the first typecast, then the first `(`, then the next typecast, so on. I was only able to get it right by splicing the comments in.

### DIFF
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -17,11 +17,10 @@ function hasClosureCompilerTypeCastComment(text, path) {
 
   const n = path.getValue();
   const parenStart = getParenStart(n);
-  // console.log(n);
 
   return (
-    parenStart <= n.start &&
-    (hasTypeCastComment(n, parenStart, 0) ||
+    parenStart !== -1 &&
+    (hasTypeCastComment(n, parenStart, -1) ||
       hasAncestorTypeCastComment(0, parenStart))
   );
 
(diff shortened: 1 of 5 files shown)


src/language-js/needs-parens.js
```diff
@@ -16,37 +16,95 @@ function hasClosureCompilerTypeCastComment(text, path) {
...
+    if (typeCastIndex === -1) {
+      return (
+        ancestorParenStart === -1 &&
+        hasAncestorTypeCastComment(index + 1, parenStart)
+      );
+    }
+
+    // There's an annoying issue if both our node contains a typecast, and this
+    // ancestor contains a typecast. Unfortunately, both comments are applied
+    // to ancestor. So we need to detect this situation, and insert hack paren
```
thorn0: Unfortunately, we can't rely on what node the comment will be attached to. It's not always the ancestor. E.g. see https://github.com/prettier/prettier/issues/7775.

thorn0:
Please target the `next` branch, not `master`.

src/language-js/needs-parens.js
jridgewell: Ahh, I see. This is attaching two casts onto the same expression. I can fix this in a follow up PR using the same paren comment hack, if you'd like?

jridgewell:
> Please target the next branch, not master.

Done.

src/language-js/needs-parens.js
```diff
@@ -16,37 +16,95 @@ function hasClosureCompilerTypeCastComment(text, path) {
...
+        comments.isBlockComment(comment) &&
+        isTypeCastComment(comment.value)
     );
   }
 
-  function isParenthesized(node) {
+  function getParenStart(node) {
     // Closure typecast comments only really make sense when _not_ using
     // typescript or flow parsers, so we take advantage of the babel parser's
     // parenthesized expressions.
```
thorn0: I forgot this feature was Babel-only. We should use Babel's `createParenthesizedExpressions` option then. This should give us nested parens in the AST. Could you look into that?

jridgewell:
Closing for #7791.

PR #7783, opened 2020-03-17
Fix broken HTML anchor in the Rationale documentation

Calinou:
The italic Markdown formatting is included in the [automatically generated ID](https://prettier.io/docs/en/rationale.html#what-prettier-is-_not_-concerned-about), so it must be present in the references as well.


thorn0:
Thanks!

PR #7786, opened 2020-03-18
Run verify test on  disabled `babel-ts` tests

fisker:
~Refactor disabled `babel-ts` tests to a better way to manage.~

~Seems there is already a fixture fixed see https://github.com/prettier/prettier/commit/95edf2ee7b7b7d8193070a9fce8d845d20858a0d~

Run verify test on disabled `babel-ts` tests, make sure they throws `SyntaxError`

### DIFF
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -101,24 +101,28 @@ global.run_spec = (dirname, parsers, options) => {
     });
 
     const parsersToVerify = parsers.slice(1);
-    if (
-      parsers.includes("typescript") &&
-      !parsers.includes("babel-ts") &&
-      !(
-        options &&
-        (options.disableBabelTS === true ||
-          (Array.isArray(options.disableBabelTS) &&
-            options.disableBabelTS.includes(basename)))
-      )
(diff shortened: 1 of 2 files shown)


tests_config/run_spec.js
```diff
@@ -16,6 +16,36 @@ const prettier = !TEST_STANDALONE
   ? require("prettier/local")
   : require("prettier/standalone");
 
+// TODO: make these test files support `babel-ts` parser
```
thorn0: These tests aren't TODO. They all contain invalid code. TS supports them because its parser is super permissive by design (as it powers the language service, which should work with invalid/incomplete code in IDEs), but it's unlikely Babel will ever parse them.

fisker: Good to know.

thorn0: No, we shouldn't. But the `toThrow` assertion still is a good thing to have.

fisker: Okay, I'll keep the `disableBabelTS` and add `toThrow`

PR #7788, opened 2020-03-18
Remove `index.js` and `standalone.js ` links

fisker:
These two files are created in #5054

> `/index.js` and `/standalone.js` are preserved as links to not break builds that's installed from git.

### `/index.js`

With updating `main` field in `package.json`, removing `/index.js` won't break `require('prettier')`, unless someone use `require('prettier/index')`.

### `/standalone.js`

People use `require('prettier/standalone')` ? If this is considered breaking change, it's good chance to do this in v2.0.0


j-f1:
Using `prettier/standalone` is recommended in [our browser docs](https://prettier.io/docs/en/browser.html).

fisker:
But that's accessing from bundles right? So remove this file won't break it.

j-f1:
Ah, yes. Good point!

thorn0:
What if someone wants to install Prettier from GitHub AND use it for a browser application via a bundler? That's what `standalone.js` in the root is for, I guess. But we can probably just use the `browser` field in `package.json` instead. And overall, we should check if this scenario works and decide if we want to support it.

fisker:
Closing since we support install from github.

PR #7820, opened 2020-03-21
Tweak build script to not corrupt import-fresh module

### LINKED ISSUE #7818: [2.0] Invalid configuration file `...`: parent.eval is not a function
**Environments:**

- Prettier Version: 2.0.0
- Running Prettier via: CLI
- Runtime: Node 10.10 AND 12.16
- Operating System: mac osx

**Steps to reproduce:**

Once updating to 2.0, all my prettier commands are failing, across all my projects. It's always the same error:

```
[error] Invalid configuration file `packages/config/src/configs/babel.ts`: parent.eval is not a function
```

I tried the following command on Node 10 AND 12, and it's the same error.

```
npx prettier packages/config/src/configs/babel.ts --write
```

Also fails with or without `--write`.

My `prettier.config.js`:

```js
module.exports = {
  arrowParens: 'always',
  bracketSpacing: true,
  endOfLine: 'lf',
  jsxBracketSameLine: false,
  printWidth: 100,
  proseWrap: 'always',
  semi: true,
  singleQuote: true,
  tabWidth: 2,
  trailingComma: 'all',
  useTabs: false,
};
```

**Expected behavior:**

It works.

**Actual behavior:**

It doesn't.

thorn0:
fixes #7818

### DIFF
--- a/scripts/build/config.js
+++ b/scripts/build/config.js
@@ -169,6 +169,7 @@ const coreBundles = [
       // cosmiconfig@5 -> import-fresh uses `require` to resolve js config, which caused Error:
       // Dynamic requires are not currently supported by rollup-plugin-commonjs.
       "require(filePath)": "eval('require')(filePath)",
+      "parent.eval('require')(filePath)": "parent.require(filePath)",
       "require.cache": "eval('require').cache",
       // cosmiconfig@6 -> import-fresh can't find parentModule, since module is bundled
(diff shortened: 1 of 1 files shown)


thorn0:
`test:dist` didn't catch it because `runPrettier` overrides cosmiconfig with the non-bundled module. Let's investigate this problem with tests later in a separate issue.

PR #7824, opened 2020-03-22
Remove azure pipelines

fisker:
Include codes from #7431

- Remove `.azure-pipelines/*`
- Remove `azure-pipelines.yml`
- Remove `html` `cobertura` reporter from jest config
- Remove `jest-junit`

More?


thorn0:
Any idea why commits to `master` aren't deployed to prettier.io? E.g. https://github.com/prettier/prettier/commit/fac9ce07fea57f1d852988fdbc217db9c8cf858a

fisker:
I have no idea. 

Did we change something on netlify website? When we are on `next`?

lipis:
Was it Azure?

thorn0:
Before these changes, it deployed fine. E.g. https://github.com/prettier/prettier/commit/0d813b7d4ab8dc14dd551fdd9cf78bd3d403759c was deployed.

thorn0:
Sorry, false alarm. That docs update was published to `/docs/en/next`, here: https://prettier.io/docs/en/next/install.html

thorn0:
However the coverage doesn't work. We consistently get this error:
```
ERROR: The coverage upload was rejected because the workflow run is either not in-progress or finished a while ago. Please make sure the workflow run is in-progress and is not stale for uploads to process correctly.
##[error]Codecov failed with the following error: The process 'bash' failed with exit code 1
```
https://github.com/prettier/prettier/runs/528790846#step:6:36

fisker:
I got this error first on https://github.com/prettier/prettier/pull/7431 too, and turns green in  https://github.com/prettier/prettier/commit/7e61875862e762d3190591985b0db4e3a50d1195 , I guess the file format is not the real reason. 

Maybe because our test cost too much time?

I searched this error message, didn't get too much usefully information.

fisker:
Maybe we can try upload coverage with token?

thorn0:
GitHub Actions has been having problems lately (even though githubstatus.com doesn't admit that). Let's see how the coverage will work when GH Actions are back to normal.

fisker:
I  read the  azure config file  again , seems  we  are  missing one  test, https://github.com/prettier/prettier/pull/7824/files#diff-fec826feae04e51c0d94076385408bdcL25

thorn0:
There was no force pushes in https://github.com/prettier/prettier/pull/7855, but still: "**65** successful and 2 neutral checks". Is it expected?

fisker:
@thorn0 No, it's not. I'll fix it.

PR #7837, opened 2020-03-22
Add please-upgrade-node

thorn0:
https://github.com/typicode/please-upgrade-node#readme

- (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.

### DIFF
--- a/tests_integration/__tests__/early-exit.js
+++ b/tests_integration/__tests__/early-exit.js
@@ -85,3 +85,19 @@ describe("throw error and show usage with something unexpected", () => {
     status: "non-zero",
   });
 });
+
+describe("node version error", () => {
+  const originalProcessVersion = process.version;
+  try {
+    Object.defineProperty(process, "version", {
+      value: "v8.0.0",
+      writable: false,
+    });
+    runPrettier("cli", ["--help"]).test({ status: 1 });
+  } finally {
+    Object.defineProperty(process, "version", {
+      value: originalProcessVersion,
+      writable: false,
+    });
+  }
+});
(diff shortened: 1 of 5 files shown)


fisker:
Is this really a good idea? There should be some `SyntaxError` on old Node.js version, then `pleaseUpgradeNode()` can't run.

thorn0:
You're right, but at least it catches Node 8.

tests_integration/__tests__/early-exit.js
```diff
@@ -85,3 +85,19 @@ describe("throw error and show usage with something unexpected", () => {
...
+      writable: false,
+    });
+    runPrettier("cli", ["--help"]).test({ status: 1 });
+  } finally {
+    Object.defineProperty(process, "version", {
+      value: originalProcessVersion,
+      writable: false,
+    });
+  }
+});
```
evilebottnawi: I think we need test exit code here too

thorn0: That's what `test({ status: 1 });` does.

fisker:
@thorn0 Add a entry file, make it ES5 or add post-install script?

tests_integration/__tests__/early-exit.js
evilebottnawi: I should be a little more attentively sometimes, sorry

thorn0:
@fisker Ideally, yes, but I think it's good enough as is. At least for now.

fisker:
Let's see if people complains.

GrahamCampbell:
This is causing false positives when prettier is packed with https://github.com/zeit/pkg. I'm unable to use prettier with node 12.16.1 after this change.

fisker:
@GrahamCampbell I don't know if your problem still exists, but you can use `cli.js` as an entry now.

Issue #7826, opened 2020-03-22
Formatting incorrect using styled-component component selector pattern


marclemagne:
**Prettier 2.0.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBJSVMC8mGAngDZwAmAdJQJYBuABgDr6YBmFAHkplNDgBuNpkwwATgEMoaOjDrQ+HOmTKYADNQBMAVjQj2AdzqUYACz4AWABwAHbobHm4dAObmY1+47ajMACTAADJ0UADWAL5I5hAMcBKYwP5iKmp8EnAARnBgYFJ2AK4SdhROmJFsTEIgADQgEHYK6MigUhISEEYACu0IaMggUmRGUiQD9VnSYOFwMADKdlJgYW7IkoVw9Z4AtmQA6ubycGhLYHDz-fKM8iSDYGgTIGFoCTDd0m47Usgcw6-1ABWaG4ACFprMFlIdnBQlA4L9-lsQMDuPNVhQAIqFCDwRFkAEgJYSV4SQZZKQ5Mh1IkSMIwfamCzIGwaep2TqvfbSOyDDknBLxGkARxx8A+jQGKCkaAAtPCqFQaZlRXRMh8pF8fkg-gTka8dnR1hJNvU5FA3FixQidUj6jBKYyzOZkNp7dJVKsAMIQHbfQYnXQ0wqvAAqlKlusJDE22CglAQCzAdKaAEF4-MYOQbVG4JFIkA)
```sh
--parser babel
```

**Input:**
```jsx
const Icon = styled.div`
  flex: none;
  transition: fill 0.25s;
  width: 48px;
  height: 48px;

  ${Link}:hover {
    fill: rebeccapurple;
  }
`;
```

**Output:**
```jsx
const Icon = styled.div`
  flex: none;
  transition: fill 0.25s;
  width: 48px;
  height: 48px;

  ${Link}: hover{
    fill: rebeccapurple;
  }
`;

```

**Expected behavior:**

```jsx
const Icon = styled.div`
  flex: none;
  transition: fill 0.25s;
  width: 48px;
  height: 48px;

  ${Link}:hover {
    fill: rebeccapurple;
  }
`;
```

[...]

PR #7842, opened 2020-03-23
fix formatting of pseudo-elements and pseudo-classes in styled-component

thorn0:
fixes https://github.com/prettier/prettier/issues/7826

The solution is pretty dirty, but hopefully it will do for now.

### DIFF
--- a/changelog_unreleased/javascript/pr-7842.md
+++ b/changelog_unreleased/javascript/pr-7842.md
@@ -1,0 +1,21 @@
+#### Fix formatting of pseudo-elements and pseudo-classes in styled-components template literals ([#7842](https://github.com/prettier/prettier/pull/7842) by [@thorn0](https://github.com/thorn0))
+
+<!-- prettier-ignore -->
+```jsx
+// Input
+const Foo = styled.div`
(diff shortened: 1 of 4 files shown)


thorn0:
@fisker This is the first PR without Azure Pipelines checks. Could you please check that everything works as expected here?

fisker:
#6794 can also fix it, and many styled-component issues.

thorn0:
#7826 is a regression that I'd like to fix quickly and release 2.0.2.

fisker:
Sure, I'll let you know if something missing.

fisker:
This is taking forever.... Still `Queued`, this is why I prefer keep both for a while.

thorn0:
We might have exceeded some limitations. After I merged #7431, existing PRs went crazy like this:

image

fisker:
This is because github actions don't clear previous build status on force push.

thorn0:
Is this fixable?

thorn0:
The CI fail is related to coverage (`The coverage upload was rejected because the workflow run is either not in-progress or finished a while ago.`). All the tests pass.

Issue #4878, opened 2018-07-23
Extra indentation of lines following comments within SCSS maps


morewry:
**Prettier 1.13.7**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASAZgQwDaQE4C0k2EeAzkgAQAUAOlJZQPROV5xkCu2MlE6lACaY8AawTUAxAGY40gJzSAHABpKAJgAMASnqMARtkxhRWzVUmb5mpTZV7mrALacyMEYMqu4n9KUrwAB4wDi5uHmYWYOhKACwAjJr2DJRh7niCWhZwOer60smM7IKJ5pSS6JXS6OqFlGRwmH6YTlnlggCsNXAAbHUsbBzcvPyU2ACWAOYAFvBQUtIKmJjxah06DvA4bZKCcOjVHf2s7Fw8fAITM3MLSytqsRspW9iPFvpKcB1wSvTaANwgFQgCAABxg42gZGQoBEeAgAHcAAoiBDQlCYABuEHGgiBIH0eGM4hgAGVQcZxlBJsgYHhOHBgVSGngYEiiZMnJhkFhsA1gQArMiBABCRJMcDJLTgABkqXAeTh+SAKeQ4HhkCAyGAyNDgaC8FSYAB1XEwabIWz6+ENY1E0Gag0cdWYhXA9gAR0443Y7MwnO5SF5yoaTnGtPpjK1VMm2DgAEVOBB4Iq+VH3PpTYJzchaiA6ZhxhNqQBhCBOLmaqDQN0gbwAFUw+nRwbgAF820A)
```sh
--parser scss
```

**Input:**
```scss
$falcor-colors: (
  // result of darken(#3e3938, 20)
  black200: #090808,
  // mustard used for text
  mustard200: #cf8410,
  mustard20: #eee2b3,
  red100: #fff3f2,
  seafoam20: #d5f2e6,
  // result of lighten(#339aa1, 50)
  teal20: #def3f5,
  // result of lighten(#339aa1, 40)
  teal40: #b8e5e8
);
```

**Output:**
```scss
$falcor-colors: (
  // result of darken(#3e3938, 20)
    black200: #090808,
  // mustard used for text
    mustard200: #cf8410,
  mustard20: #eee2b3,
  red100: #fff3f2,
  seafoam20: #d5f2e6,
  // result of lighten(#339aa1, 50)
    teal20: #def3f5,
  // result of lighten(#339aa1, 40)
[...]

PR #7844, opened 2020-03-23
fix: Update scss comment indentation

boyenn:
Force dedent upon hardline when working with inline scss comments.

Fixes https://github.com/prettier/prettier/issues/4878

Used existing tests.
*Not relevant*

**✨[Try the playground for this PR](https://deploy-preview-7844--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEASAtgTwLToIYAOSABABQA6UxxA5AGYQQ0kCMANMQPSfEBijlajQBGeAE7NiAJg7diAIXGUAlAG5KINiAgEYAS2gBnZKHFiIAdwAK4hMZR4ANhbyZjW4WLxgA1nBgAygTeelAA5sgwYgCucFoAFjDojgDq8XrwhsFgcAF2GXoAbhmYyOCG7iChhnBiMFZeYfjIdE41WgBWhgAe8l6+-gF46HAAMqFwLW1xIF3dAaFhjnAAitEQ8FOO7SDBYjViZWAVmrtioTApegAmMPHIABwADFoE5jUpXgRlb3AHhZMtABHdbwBo6ewgPCGbBQOBwa4I05iOAgvQohp4Jp4LY7GroPSRGIzQyLZZrDaTJCtbYzGB4YRXW73JAyEBRPB6RyLADCEHQzRQfwArKdojUACoM+w0naFWIASSgiNgATA510AEFlQEYJhlri4ABfI1AA)✨**


tests/css_less/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3276,7 +3276,7 @@ li + li {
 
 // Usage
 @import // Comment
-  "@{themes}/tidal-wave.less";
+"@{themes}/tidal-wave.less";
```
evilebottnawi: broken

boyenn: Is expected behaviour 
```
@import "@{themes}/tidal-wave.less";
```
or the old behaviour ?

boyenn: Fixed it by checking if we're in an atRule and whether there is a prevNode. 

I'm not sure this is the right way to go about it?

tests/css_less/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3276,7 +3276,7 @@ li + li {
 
 // Usage
 @import // Comment
-"@{themes}/tidal-wave.less";
+  "@{themes}/tidal-wave.less";
```
evilebottnawi: Same situation can be for `SCSS` map too

boyenn: I'm not sure how ? I tried a single value map

tests/css_comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -707,10 +707,10 @@ $my-list2:
 =====================================output=====================================
 $my-list: "foo",
   // Foo
-    "bar"; // Bar
+  "bar"; // Bar
 
 $my-list2: a // a
-  b c;
+b c;
```
evilebottnawi: Look here

boyenn:
Noticed it only goes wrong if the parent node is `value-paren_group`, seems like the extra identation is added there. Will workaround that instead.

boyenn:
@evilebottnawi Not sure what to do with the failing check ?

boyenn:
Squashed commits to trigger failing checks again.

boyenn:
Did a rebase to keep the PR up to date.

src/language-css/printer-postcss.js
```diff
@@ -697,6 +697,10 @@ function genericPrint(path, options, print) {
 
         // Add `hardline` after inline comment (i.e. `// comment\n foo: bar;`)
         if (isInlineValueCommentNode(iNode)) {
+          if (parentNode.type === "value-paren_group") {
```
evilebottnawi, line 700: Potential unsafe, `value-paren_group` can be used in functions and nested function, need more tests, I just feel that it breaks something

boyenn:
FYI: 
I'm trying to find some free time to update/add more tests. Did not abandon this yet.

fisker:
@boyenn Friendly ping

src/language-css/printer-postcss.js
boyenn: Having a look at this now, I'm not sure what kind of tests would suffice. I'll add nested functions for now, but I'm not sure if I'm writing the kind of tests you're looking for.

boyenn:
Having a look at this now, I'm not sure what kind of tests would suffice. I'll add nested functions for now, but I'm not sure if I'm writing the kind of tests you're looking for.

fisker:
Thanks for fixing this!

PR #7851, opened 2020-03-23
Collapse single CSS rules onto a single line.

jdelStrother:
This is a POC for collapsing single (or empty) CSS rules onto a single line.  The relative verbosity of prettier's CSS styling has always been a sticking point for integrating it into our project, and it seems like a common request for others: https://github.com/prettier/prettier/issues/5948

To my eyes, most of the changes to the test snapshots are clear improvements.  There's definitely a couple that are a little worse, but thought I'd throw this out first to see if something like this would be accepted as a default setting.


jdelStrother:
Doh, I missed those test failures.  Will try and get them fixed up.

alexander-akait (changes requested):
:-1: On this

> .x .y { }

I have never seen this in any code base

> .reference { composes: selector from "a/long/file/path/exceeding/the/maximum/length/forcing/a/line-wrap/file.css"; }

Ugly and unreadable

> div { grid-template-areas ... }

Unreadable

And a lot of other problems

jdelStrother:
> > .x .y { }
> 
> I have never seen this in any code base

I mean...  it's an empty CSS selector, so is pretty useless on it's own, maybe that's why you've never seen it in any code base?
Why do you find that so objectionable, when javascript like this is perfectly fine?

```js
function foo() {}
const x = { bar: 1 };
```

---

> > .reference { composes: selector from "a/long/file/path/exceeding/the/maximum/length/forcing/a/line-wrap/file.css"; }
>
> Ugly and unreadable

I think that's unfair.  The original is barely any prettier or more readable:

```css
.reference {
  composes: selector from "a/long/file/path/exceeding/the/maximum/length/forcing/a/line-wrap/file.css";
}
```

---

> > div { grid-template-areas ... }
>
> Unreadable

Agreed.  I'd definitely like to fix it so that multiline values like this stay on multiple lines:
```css
div { grid-template-columns:
    [full-start] minmax(1.5em, 1fr)
    [main-start] minmax(0.4ch, 75ch)
    [main-end] minmax(1em, 1fr)
    [full-end]; }
```

but it sounds like even if I fix that you won't accept this PR?

src/language-css/printer-postcss.js
```diff
@@ -120,6 +120,7 @@ function genericPrint(path, options, print) {
       return isInlineComment ? text.trimEnd() : text;
     }
     case "css-rule": {
+      const nodeSeparator = singleLineRule(node) ? " " : hardline;
```
j-f1:
```suggestion
      const nodeSeparator = singleLineRule(node) ? line : hardline;
```

This should (?) make it still break when the rule itself breaks.

evilebottnawi:
@jdelStrother I am afraid that this change causes a lot of disputes. Most developers use:
```
.selector {
  prop: value;
}
```

even for one property. 

I do not see a real problem. The current version is very readable and I would leave it as it is

thorn0:
> but it sounds like even if I fix that you won't accept this PR?

As long as the issue is marked as `needs:discussion`, PRs aren't accepted.

jdelStrother:
> @jdelStrother I am afraid that this change causes a lot of disputes. Most developers use:
> 
> ```
> .selector {
>   prop: value;
> }
> ```
> 
> even for one property.
> 
> I do not see a real problem. The current version is very readable and I would leave it as it is

It makes functional css three times longer than it should be.  Obviously readability is subjective, but to my eyes 90% of the snapshot changes are improvements.  95% if I were to fix the multiline value issue like that grid-template-columns example. 

image

image


image


It sounds like you're strongly opposed to making this the default behaviour.  What about either:

* having it as a user-configurable option?  (I'm well aware of, and in favour of, Prettier making opinionated choices rather than having user-configurable options.)

* allowing CSS to stay on one line if the original source is on one line?  There's precedence for this in javascript objects, where both of these objects stay as they were originally written:

```js
const foo = { x: 1 };
const bar = {
  y: 2
};
```

- though I'm not entirely sure how that's implemented, would appreciate any pointers.

j-f1:
Here’s how it’s done for objects:

https://github.com/prettier/prettier/blob/16c64635effe0b0d84ae8c829973ad38c4c41b63/src/language-js/printer-estree.js#L1331-L1337

I’m definitely in favor of fixing the formatting for `.foo {}`, although I think to follow the style of empty braces in JS it should not have a space (i.e. `{}` not `{ }`.

jdelStrother:
> > but it sounds like even if I fix that you won't accept this PR?
> 
> As long as the issue is marked as `needs:discussion`, PRs aren't accepted.

I'm not trying to force this through without discussion, I just want to know if I'm wasting time even suggesting this.

lydell:
Doesn’t this result in worse diffs when adding a second declaration?

```diff
-input[type="search"] { font-weight: bold; }
+input[type="search"] {
+  font-weight: bold;
+  border-radius: 0;
+}
```

vs

```diff
 input[type="search"] {
   font-weight: bold;
+  border-radius: 0;
 }
```

jdelStrother:
> Doesn’t this result in worse diffs when adding a second declaration?

It can do. I suppose I’m approaching this as a user of functional/atomic css, where most/many rules have a single declaration, and it’s actually relatively rare to change a rule once it’s been written.

jdelStrother:
I've pushed an updated version.  It collapses empty declarations to a single line:

```css
.foo {}
```

and allows declarations to remain on a single line, if they only contain one rule & no newlines.

jdelStrother:
The new version is causing headaches with that failing second-format test.  I started work on another version that removes hardlines from css declarations entirely, instead wrapping them in a group with `shouldBreak`, depending on whether the source declaration contains a newline

```js
function printDeclarationList(path, options, print) {
  const node = path.getValue();
  if (node.nodes.length === 0) {
    return "";
  }
  const shouldBreak = !singleLineNode(node, options);
  return group(
    concat([
      indent(concat([line, printNodeSequence(path, options, print)])),
      line,
    ]),
    { shouldBreak }
  );
}
function singleLineNode(node, options) {
  // Declarations with a single rule that already contains a newline
  // should continue using multiple lines
  if (
    hasNewlineInRange(
      options.originalText,
      options.locStart(node),
      options.locEnd(node)
    )
  ) {
    return false;
  }

  const child = node.nodes[0];
  return (
    child.type === "css-rule" ||
    child.type === "css-decl" ||
    child.type === "css-atrule" ||
    child.type === "css-comment"
  );
}
```


This version allows multiple declarations on one line as long as it's less than the `printWidth` -  eg
```css
/* input */
.foo { color: red; font-weight: bold; }
.some-long-long-long-long-long-long-long-long-long-long-line { color: red; font-weight: bold; }
/* output */
.foo { color: red; font-weight: bold; }
.some-long-long-long-long-long-long-long-long-long-long-line {
  color: red;
  font-weight: bold;
}
```

but it's turning into a bit of a rabbithole, and is then inconsistent with how selectors are formatted.    Currently, selectors are always broken onto newlines - eg
```css
/* input */
.foo, .bar { color: red; }
/* output */
.foo,
.bar {
  color: red;
}
```

Before I try and go much further, I was hoping to get some feedback on what people thought to allowing single-line output if the input is on a single line. Is that acceptable at all, or only for declarations, or for declarations & selectors but not something else?

thorn0:
@jdelStrother It sucks to write this to someone who's trying to contribute, but the whole idea of opinionated code formatter is about leaving as few things up to the developer as possible (to keep the developer's brain completely free from formatting-related concerns). That's why Prettier strategically strictly resists adding options and respecting the original formatting. The latter is basically an option in disguise, i.e. something left up to the developer. As for JS object literals, exception is exception. It's not to be used as a justification for new exceptions, otherwise the exception simply becomes the norm and the whole intent is compromised. The result might be good, even great, bit it'd be a different project with a different value proposition and target audience. All this means it's really unlikely this PR will be accepted (at least that part of it ). I'm sorry. An alternative is to fork Prettier or probably create a plugin (which would consist of a forked CSS printer).

jdelStrother:
> @jdelStrother It sucks to write this to someone who's trying to contribute, but the whole idea of opinionated code formatter is about leaving as few things up to the developer as possible

I think that’s reasonable, it’s one of the main reasons we use prettier for js. It’s just a shame that it means we don’t want to use prettier for our css.  Let’s just rule out all the other options while we’re here...

In the original version of this PR, all CSS declarations with a single rule were collapsed to one line.  To take the simplest example, does `a {color:red}` have to be broken onto multiple lines like this, in prettier’s approach?

```
a {
  color: red;
}
```

Or are there circumstances where that’s optional / heuristically decided / enforced?

thorn0:
> Or are there circumstances where that’s optional / heuristically decided / enforced?

AFAIK, no.

jdelStrother:
Sorry, my question was poorly worded. Do the prettier maintainers think that a selector like that ought to always appear on 3 lines, or is there any discussion to be had around condensing it?

If you’re still opposed to even condensing simple selectors like that, maybe https://github.com/prettier/prettier/issues/5948 should be closed as a wontfix?

thorn0:
Maybe. The huge problem with issues like #5948 is that there is no good way to measure the real demand. Given how popular Prettier has become, neither GitHub reactions nor Twitter polls are representative. That's why I'd like you to consider publishing a fork or a plugin. If someone did that, we could see its stats, issues, etc. and get a better sense of the demand. Overall, encouraging an ecosystem of plugins feels like a much healthier direction for the project than endless discussions about adding options. Prettier's plugin API might be not as granular as it could be (e.g. now it's impossibly to override only how function calls are printed in JS, you need to override the whole JS printer), but this should improve over time.

samuelfarrell:
@jdelStrother Did you ever end up forking/writing a plugin? As per #5948, it looks like there's quite a bit of demand for it and its also apparent the maintainers of Prettier have no interest in re-evaluating how Prettier prints single line CSS declarations. A plugin like this might make me give Prettier another shot for CSS.

jdelStrother:
> @jdelStrother Did you ever end up forking/writing a plugin? As per #5948, it looks like there's quite a bit of demand for it and its also apparent the maintainers of Prettier have no interest in re-evaluating how Prettier prints single line CSS declarations. A plugin like this might make me give Prettier another shot for CSS.

No, I confess I got dispirited after this PR and put it aside for a while.  It sounded like @jakecasey was interested in it but don't know if he got anywhere.  (Or even if that's the right Jake Casey - sorry if not!)

JakeCasey:
> No, I confess I got dispirited after this PR and put it aside for a while. It sounded like @JakeCasey was interested in it but don't know if he got anywhere. (Or even if that's the right Jake Casey - sorry if not!)

Honestly, we were just spending our time arguing in the issue hoping to change their minds. I haven't earnestly tried yet. I want to, but have absolutely 0 experience working w/ prettier's source, so the task seems a little daunting.

evilebottnawi:
Issue was not closed and still discussed, if you want to achieve this using a plugin (you can't do it, because it is not supported yet) feel free to create repo and continue discussion there

fisker:
Close since maintainers are against this change. We can continue the discussion in the linked issue.

PR #7866, opened 2020-03-25
Tweak test and CI configs

fisker:
- Set `jest` verbose to false by default
- Correct `doc-debug.js` path
- Remove outdated ignore patterns
- Exclude Node.js 13 on macos

### DIFF
--- a/.github/workflows/dev-test.yml
+++ b/.github/workflows/dev-test.yml
@@ -24,6 +24,9 @@ jobs:
           - os: "ubuntu-latest"
             node: "12"
             ENABLE_CODE_COVERAGE: true
+        exclude:
+          - os: "macos-latest"
+            node: "13"
     env:
       ENABLE_CODE_COVERAGE: ${{ matrix.ENABLE_CODE_COVERAGE }}
     name: Node.js ${{ matrix.node }} on ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


.github/workflows/dev-test.yml
```diff
@@ -24,6 +24,9 @@ jobs:
           - os: "ubuntu-latest"
             node: "12"
             ENABLE_CODE_COVERAGE: true
+        exclude:
+          - os: "macos-latest"
+            node: "13"
```
thorn0: Do you exclude it simply because macos tests take longer and are flakier or is there something else?

fisker: Github Actions limit Maximum concurrent macOS jobs to 5, https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

thorn0:
Why do we need `verbose: true` at all, even on CI? Won't the tests run faster without it?

fisker:
I don't think this will slow down the tests, and I found more information is very useful for new contributors.

thorn0:
The log of a test run now has almost 20000 lines. Scrolling it in the browser is painful, especially on mobile. Is there a way probably to keep the benefits that `verbose` gives us without printing the names of individual tests? It's those names that make the log so big.

PR #7874, opened 2020-03-25
First commit for smart-padded-blocks ; let the haters hate

kswope:


j-f1:
This is interesting! I’m not sure if we’re going to take this, but if we did:

- it shouldn’t be an option
- like with breaking objects, it should only look for blank lines at the beginning of a block so it’s easy to unpad a block by deleting one character.

Issue #7790, opened 2020-03-18
Prettier assumes untagged template string parameter is graphql


tdjone1:
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEA5gE4CGADgBYCOANgBRQQyF1L4DOMxOUhAlOy48++YAB0o+fMTgx0xKUxZ18AanwByThAC2cyr0KbJAX0mTIULvgAe+ALxEyVOvQAGwHABNT7-hZQAPRB+KgQtLQQAO5G+JDecPixkfGk6BxJ5LIwuHDE+PnEEMTsAMoAnrCk9gCixMWl+LW25HDYcN74AHKk+gA0YRCYXab49ACMSABMAcGhVjYVjs4UNAyefvwg-SAQ5LjQHMigpA0xAApnCMcopABuED47IABGZGAA1nJl5KRgRmQ3HQcF2lBguloAHVDPAOH8wHAyjccLh7qiKshwBxjrteJliDALmRCLpSMhUKRaJldgArDi2ABCH2+MDKfTgABleHAKVSaSB6bYykZaHAAIroZi8pCU6mgkB-YgErEwCptDhgHgHF7ZXgwKE+GCUZAADgADLtshBMlCXFjsnACfdebtqFL4MT9rcQKQOABaKBwTqdF6yd04WTE0ik8my-kKzK6HBA4gg3YcUUSj0yuUCmCkV6G7zG5DTXbcUg4WhGADCejJWKYQZeGTgABVC7c8wr7iCAJJQRKwMpanAHACCQ7KarFfPlplMQA)
```sh
--parser typescript
```

**Input:**
```tsx
function graphql(notgql: string): string {
  return notgql + ' something'
}

const x = graphql(`{id}`)

// following code will cause prettier error: Syntax Error: Unexpected Name "x" (1:1)
// const y = graphql(`x {id}`)
```

**Output:**
```tsx
function graphql(notgql: string): string {
  return notgql + " something";
}

const x = graphql(
  `
    {
      id
    }
  `
);

// following code will cause prettier error: Syntax Error: Unexpected Name "x" (1:1)
// const y = graphql(`x {id}`)

```

**Expected behavior:**

The template strings passed to `graphql()` should not be reformatted because there is no guarantee that they are graphql queries.

evilebottnawi:
Do not use `graphql` name, it is special name and have special formatting, duplicate of #5588

[...]

PR #7875, opened 2020-03-26
Add --embedded-language-formatting={auto,off} option

bakkot:
This adds an option to completely disable language-based formatting of template literals, code blocks in markdown, etc, for cases like https://github.com/prettier/prettier/issues/7790 or https://github.com/prettier/prettier/issues/5588#issuecomment-470418654.

There's been a lot of discussion about how to handle formatting of template literals (https://github.com/prettier/prettier/pull/6626, https://github.com/prettier/prettier/issues/5588, https://github.com/prettier/prettier/issues/4424, many other places). We've talked a bit about making this pluggable, which I think is an interesting idea. But no one has gotten around to that yet, so I want to propose something simpler to start. Specifically, `--embedded-language-formatting={auto,off}`, which would allow toggling between the current behavior and no formatting.

This would be extensible with new values (beyond `auto` and `off`) if we ever get around to implementing something more customizable (like https://github.com/prettier/prettier/issues/4424 or https://github.com/prettier/prettier/pull/6626).

I've added a simple test ~but haven't yet added documentation or updated the changelog. I'll do those if there's support for this~ and added documentation and updated the changelog. cc @thorn0 and @evilebottnawi since you were [discussing this recently](https://github.com/prettier/prettier/issues/7790#issuecomment-601647683).


thorn0:
> This would be extensible with new values (beyond auto and off) if we ever get around to implementing something more customizable (like #4424 or #6626).

We need a more future-proof name for the option. A future customizable solution for managing nested languages might support not only template literals in JS but also code blocks in Markdown (https://github.com/prettier/prettier/issues/6679), script tags in HTML, etc.

bakkot:
I am happy to go with another name. Thoughts on `--nested-language-formatting`?

thorn0:
Disabling it only in JS will do for now as it seems to be a pain point for many users. We just need to reflect that in the option value. E.g. `--nested-languages disable-in-js`.

bakkot:
Sure. I would go with `--nested-languages 'in-js:off'`, maybe?

thorn0:
Or maybe even `--nested-languages js:off`. It can later develop into something like this:
```console
# media.mobile`max-width: 90%;`
prettier --nested-languages js:template:media.*:css

# query(`{ foo { bar } }`)
prettier --nested-languages js:call:query:graphql
```
We need to hear from other collaborators to proceed.

bakkot:
The reason I suggest `in-js` rather than `js` is that just `js` is ambiguous whether it's talking about js as the language in which other things are embedded, or as the thing which is being embedded in something else (e.g. in HTML).

suchipi:
I'm not convinced that `// prettier-ignore` isn't good enough for these situations

bakkot:
@suchipi Without this option it is impossible to run prettier and trust that the resulting program will behave the same as the input program. You have to know that you're not using any of the magic tags in a way prettier doesn't expect. For a formatter, that's very, very bad.

suchipi:
Could we instead have a config mapping template tag names to plugins, and you can pass an empty config to do no template formatting (or you could override one of the defaults)? That would also let users map eg ruby\`...\` to the community ruby plugin.

suchipi:
Also, if one of our builtin plugins is breaking the input program (html not supporting amp, etc), then that's a bug that should be fixed in our plugin.

bakkot:
@suchipi 

> Could we instead have a config mapping template tag names to plugins

That would be better but is also something we have been considering for [years](https://github.com/prettier/prettier/issues/4424), and which no one has implemented yet. I would like to get this simpler thing in first, which we could later extend to support something more configurable.

> Also, if one of our builtin plugins is breaking the input program

They have been [for years](https://github.com/prettier/prettier/issues/5588#issuecomment-470418654). It's not a bug in the plugin; the problem is the decision to format template literals. There is no way to know that it is safe to do this. Hence my ask for an option to disable that entirely.

suchipi:
Alright, I understand and agree, then. But I want to make sure this option doesn't get in the way of a future library or template tag formatting plugin API.

Since there are already workarounds (renaming the template tag or using prettier-ignore), and because it could complicate supporting a more extensible option in the future, I don't think this config option should be super configurable; especially since it's not only tagged template literals that get treated special (there's also styled-jsx and styled-components, for instance). I think just `auto` or `off` is all we should do here, and if it's set to `off`, then we bail out of any tagged template literal formatting, no matter where it appears.

Otherwise, the question of "what if they want css formatting on, but styled-components formatting off?" comes up, and I think that is better answered by a future library plugins API.

bakkot:
Great; that's the behavior this PR currently implements (at least, I think). Do you have thoughts on @thorn0's [comment](https://github.com/prettier/prettier/pull/7875#issuecomment-604439619) above about choosing a name which we could re-use for something more extensible later? Or do you think it's better to call this `--template-formatting`  or something (maybe `--js-template-formatting`?) and later introduce a new name if we do something more extensible?

suchipi:
Maybe `--embedded-language-formatting={auto|off}`? We could make it impact fenced code blocks in markdown, too (anywhere we use the multiparser)

bakkot:
Sounds good. I'll update this PR to add that when I get a moment.

Other than JS templates and markdown code blocks, is there anywhere else this comes up? script and style tags in HTML, I'd guess?

suchipi:
I think so, but I'm not familiar with the html printer, so I'm not 100% sure.

@lydell mentioned a few non-obvious js examples here: https://github.com/prettier/prettier/issues/5588#issuecomment-444221185

It looks like they're all in language-js/embed.js: https://github.com/prettier/prettier/blob/master/src/language-js/embed.js

lydell:
In my ideal-config-world (#7073 ) there would be no CLI flag. Only config per language.

```json
{
  "javascript": {
    "embedded-language-formatting": ["graphql"]
  },
  "html": {
    "embedded-language-formatting": ["javascript", "css"]
  }
}
```

Not sure if it helps the discussion or not.

bakkot:
@lydell The idea with this PR is specifically to get something simple done now (in a way which could later be extended) rather than waiting until the larger discussions about config get resolved, because those discussions have been ongoing for years while this issue has continued tripping people up.

thorn0:
@lydell A config written this way doesn't contain enough information. Say we want to format templates tagged with `foo` as CSS. Where is the place for `foo` in your config? Also Prettier formats some non-tagged template literals (`graphql(...)` calls, inline templates in Angular components), so a way is needed to describe such patterns too.

lydell:
```json
{
  "javascript": {
    "embedded-language-formatting": ["graphql", "angular", "styled-components"]
  }
}
```

thorn0:
@lydell What about configuring additional tags for CSS? Users keep asking for it. E.g. https://github.com/prettier/prettier/issues/5936

lydell:
We just add them all to the `styled-components` option value.

bakkot:
OK, I've renamed this to `--embedded-language-formatting` (still with only valid values `auto` and `off`) and made it apply to all file types (with tests for HTML, Markdown, and JS, which are the only ones I saw). Are we happy with that? If so I'll add docs/changelog (and rename the PR).

thorn0:
LGTM

suchipi:
Looks great to me :+1:

bakkot:
OK, I've added docs and updated the changelog; I think it should be good to go.

Since this is a new option, I think it warrants a longer window for feedback than most PRs, so let's leave it open for at least a few more days? And ideally try to get feedback from more of the maintainers, if possible. I don't know how to go about requesting that though.

changelog_unreleased/javascript/pr-7875.md
```diff
@@ -0,0 +1,27 @@
...
+
+<!-- prettier-ignore -->
+```js
+// Input
+html`
+<p>
+I am expecting this to come out exactly like it went in.
+`;
+
+// using --template-formatting=auto (or omitting this option)
```
evilebottnawi: Typo `--template-formatting=` to `--embedded-language-formatting`

bakkot: Thanks, fixed.

docs/options.md
```diff
@@ -364,3 +364,22 @@ Valid options:
...
+Sometimes this behavior is undesirable, particularly in cases where you might not have intended the string to be interpreted as code. This option allows you to switch between the default behavior (`auto`) and disabling this feature entirely (`off`).
+
+Valid options:
+
+- `"auto"` – Format embedded code if Prettier can automatically identify it.
+- `"off"` - Never automatically format embedded code.
+
+| Default  | CLI Override                                    | API Override                                   |
+| -------- | ----------------------------------------------- | ---------------------------------------------- |
+| `"auto"` | <code>--embedded-language-formatting=off</code> | <code>embeddedLanguageFormatting: "off"</code> |
```
fisker, line 385:
```suggestion
| `"auto"` | `--embedded-language-formatting=off` | `embeddedLanguageFormatting: "off"` |
```

bakkot: Done, though there's a bunch of other usages of `<code>` in this file.

fisker: Fixed in #7937

fisker:
If  I  remember correctly the vue and angular interpolation is  also  using embedded, should them  control by  this  flag? I'm not near  my  laptop, can't confirm this.

I was  trying to  check  on the  playground, seems this flag is  not  added to  playground, we  should add  it.

bakkot:
@fisker It does apply to Vue; I've added a test confirming this. I'm not sure how to check Angular.

I also don't know how to update the website, so I'm not planning to do it in this PR. Someone can do it as a followup or by pushing more commits to this branch.

fisker:
@bakkot I think this controls the options on playground https://github.com/prettier/prettier/blob/04e4ef375844b011ab586072ecc97678afe838b4/website/playground/Playground.js#L26

docs/options.md
```diff
@@ -361,9 +361,9 @@ Valid options:
 - `"auto"` - Maintain existing line endings
   (mixed values within one file are normalised by looking at what's used after the first line)
 
-| Default | CLI Override                      | API Override                     |
-| ------- | --------------------------------- | -------------------------------- |
-| `"lf"`  | `--end-of-line <lf|crlf|cr|auto>` | `endOfLine: "<lf|crlf|cr|auto>"` |
+| Default | CLI Override                                                | API Override                                               |
+| ------- | ----------------------------------------------------------- | ---------------------------------------------------------- |
+| `"lf"`  | <code>--end-of-line <lf&#124;crlf&#124;cr&#124;auto></code> | <code>endOfLine: "<lf&#124;crlf&#124;cr&#124;auto>"</code> |
```
fisker, line 366: Is it a bug in `docusaurus`, if it can't render correctly?

thorn0: Probably.

fisker:
[Angular `*ngFor`](https://deploy-preview-7875--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AJ-4BUUA5gGIQBOAvADoj4A2cMB+AFnJRG7hAGa9O3OAGd6APlQB6LNgkgANCAgAHGJmijkoAIaVuAdwAK+hNpS7Gh3QE9tygEaVdYANYsAyqteZSyGEoAVzhlOABbRzh0dGiAGV1SIN0SOApKcN0YDX8UAX4lEHYYcMYAdXZMeFEfMDhPcyqcKttkcFEHED9RLhhjFxJM5H4rHuUAK1EADwAhF3cvXXC4OL84YdHQkEmpzz8SZgBFIIh4DcYxkB9KHso2xJIgxn1C1Uo-GDLMdBh2ZAAOAAMyjeEB6ZRcqjabzEXGw62UAEcTvB+moLCBdKIALRQODRaKFShwZGYYn9FJDJAjC5bHrhTABYJ0-ZHFHrambZQwXSOL4-P5IABM3JcmEY+wAwhBwlSQGIAKyFII9AAqvIsNMu2BCAEkoLFYJ4wO91ABBA2eGC2ZjnHoAX3tQA) and [vue `script`](https://deploy-preview-7875--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzmATgSwA4wB8AOlADZwwAEAClpTDnFlQLxUBmEWAtgIbVWhKsFJVxVMRKniZkqBPmK5cgJ4QArlgB0AK3QAKAOQwAFpoDmpowEoqAX1KoA9JlwFCIADQgIBHNDoyKB8WFgQAO40oQhBKHxkEXyqQT4ARlh8YADWlADKeFk4UBbIMFgacD5wPGlwACb1DQAyfCUafBZwAGLc-DCMJci+HBzeIKYwPGQA6qY48OiFYHB5sQs4AG4LqsNg6KkgxejMMHSd-MgcCSc++gAeAEKZOfl8PHDNxXBXN1UgDzyxQsFAAihoIPBfmRbiBClgTlhhptKuM8LhYDMcPUzMgABwABh86IgJxmmTww3RcERmx+PgAjhD4HQ-HEQHx0ABaKBwBoNcb0Jk4ejnCyXJDXGH-E48HBlCoy4Fg5k-SV-HwwPhpLE40zIABMmsyODIwIAwhAeBKQDSAKzjDQnAAq2riUthKLgAEkoE1YHlsPgYABBP15GCqCjQk72exAA) and [`<style>` inside html](https://deploy-preview-7875--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzjAngGzgPgENgAdKAAkvLKuoqoFcBzHLJcsddG8gX1QD0mXARAAaEBAAOMAJbR0yUIQBOKiAHcACqoSKUhHBsJZFEgEYrCYANZwYAZSnXZUJshgqGcCXAC25nAAJkHBADKEbgyETHAAYhAqfoQwcm7IkgBmmeIgABYwfjgA6nmy8OjOYHAOeuWyAG7lWBmcZiCu6HAqMFpWTMnImYZdEgBW6AAeAEJWtvYOhH5wYa5wQyM+IBOTDq4scACKDBDwGzijIM4qXSoZDd65UiquMMWyQTB5yAAcAAwSZ4QLrFKxSDLPOC3BrrCQARxO8D60n0IEI6AAtFA4MFgrkVHAEbICX0YoMkMMLlsun5ZB4vNT9nhjqd1hTNhIYIRzO9Pt8kAAmTlWWQ4fYAYQgfnJIChAFZcgwugAVbn6SmXB5wACSUFCsAcYBeMgAgnqHNg8OcurxeEA) won't parse as expected.

But I don't think those should controlled by this option.

Vue file always has `script`, we just don't want parser something like html tag inside script 

```vue
<script>
export default {
	data() {
		return {a: html`<div>
			HTML inside this should not formatted.
</div>`}
	}
}
</script>
```

WDYT?

thorn0:
It's good enough as it is. New values can be added later to this option for more granular control.

fisker:
This option is good, but I really don't like my `script`/`style` in `vue`/`html` disabled formatting by `--embedded-language-formatting=off`.

thorn0:
Don't set this option to `off` for your Vue files then. What's your use case? "we just don't want parser something like html tag inside script"? Do you actually use `` html`...` `` in Vue files?

fisker:
Yes, right. I didn't think through. I don't use that.

fisker:
@bakkot Can you rebase this? I guess there will be a snapshot title change, because test scripts updated in #7934 .

bakkot:
@fisker Done.

fisker:
Thanks!

Merge? @prettier/core

jonathanstiansen:
@thorn0 is it intentional that the latest release in npm is 2.0.5, not 2.1? 
https://www.npmjs.com/package/prettier
Since the docs on the homepage reference 2.1... :\

jquense:
@fisker Is the plan still to expand this option to allow configuring which tags/comments denote which embedded language? Is anyone working on that, if not would ya'll be open to a contribution adding it, or is there more planning/discussion needed?

fisker:
> Is the plan still to expand this option to allow configuring which tags/comments denote which embedded language?

Yes.

> Is anyone working on that

There is not AFAIK.

fisker:
We may want define what `auto` is.

jquense:
from nothing, i'd expect `auto` to handle stuff more like language tags on markdown code fences, but given that prettier already does much more by codifying library APIs it seems hard to make it a real "auto" and not "prettier's defaults". E.g I'd expect `css`, `scss` template tags to format in a true `auto` but not `styled()`. 

Maybe `auto` should expand to cover more basic cases, with the library specific formatting deprecated and removed in v3?

nmatei:
js```
// --embedded-language-formatting=html
const tpl = `<div class="my-div-class"><span class="my-icon-class">icon</span><span class="my-text-class">text</span></div>`;
```
will be grate, as we can use html template in pure js files, 
and now will only formate if I use invalid js syntax, but then I have to remove html before my template.
js```
const tpl = html`<div class="my-div-class">...</div>`;
```

Anyway but is a grate starting point! Thanks!

PR #7876, opened 2020-03-26
Update changelog Template

fisker:
This is much clear in this way,

Before: https://github.com/fisker/prettier/blob/df52e9781b809deb1332ed0d88a58db67f354a70/changelog_unreleased/html/pr-7867.md

After: https://github.com/fisker/prettier/blob/c888f2df3359481a3657668472e1698f95701ae0/changelog_unreleased/html/pr-7867.md

@lydell WDYT?

### DIFF
--- a/changelog_unreleased/TEMPLATE.md
+++ b/changelog_unreleased/TEMPLATE.md
@@ -16,9 +16,7 @@
 
 6. Change ```jsx to your language. For example, ```yaml.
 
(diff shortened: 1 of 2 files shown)


lydell:
I think this makes sense! :+1: I haven’t checked how it looks when rendered, though.

thorn0:
I don't like how it would look in a blog post.

fisker:
Close because rejected.

PR #7890, opened 2020-03-28
add Tradie Training to users of Prettier

ryan-senn:
Thanks for all the work that goes into prettier!


j-f1:
Thank you for using Prettier!

PR #7891, opened 2020-03-28
Uniform line endings for multi line strings/comments

### LINKED ISSUE #6501: CRLF: CR is removed in multiline comments and strings
```
type IAmIncredibleLongParameterType = {};
const IAmAnotherFunctionName = (_0: IAmIncredibleLongParameterType) => {};
export const IAmIncredibleLongFunctionName = IAmAnotherFunctionName(
    (_0: IAmIncredibleLongParameterType) => {
        setTimeout(() => {
            /*
                Multiline comment
                Multiline comment
                Multiline comment
            */
            console.log(
                'Multiline string\
                 Multiline string\
                 Multiline string'
            );
        });
    }
);
```
**Before running `lint --fix`:**

before

Errors:
```
   6:15  error  Delete `␍`  prettier/prettier
   7:34  error  Delete `␍`  prettier/prettier
   8:34  error  Delete `␍`  prettier/prettier
   9:34  error  Delete `␍`  prettier/prettier
  12:35  error  Delete `␍`  prettier/prettier
  13:35  error  Delete `␍`  prettier/prettier
```
**After running `lint --fix`:**

after

**What version of `eslint` are you using?**
6.0.1

**What version of `prettier` are you using?**
1.18.2

**What version of `eslint-plugin-prettier` are you using?**
3.1.0

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
```
{
  "printWidth": 100,
  "tabWidth": 4,
  "singleQuote": true,
  "quoteProps": "consistent"
}
```
```
module.exports = {
    root: true,
    parser: '@typescript-eslint/parser',
    parserOptions: {
        ecmaVersion: 6,
        sourceType: 'module'
    },
    plugins: ['import', 'folder-schema', 'processors-stub', 'react-hooks'],
    extends: ['plugin:import/typescript', 'plugin:prettier/recommended'],
    rules: {
        'react-hooks/rules-of-hooks': 'error',
        'react-hooks/exhaustive-deps': 'warn'
    }
};
```

**What source code are you linting?**
*.tsx

**What did you expect to happen?**
Equal line endings for the whole file.

**What actually happened?**
The plugin is requiring and replacing CRLF with LF in multiline strings and comments

sidharthv96:
Fixes #6501 
* Removed double parsing of the doc
* Moved newline check to printDocToString


tests/end_of_line/multiline.js
```diff
@@ -0,0 +1,16 @@
+const foo = bar(
+  (_0: baz) => {
```
thorn0: I can't reproduce the bug in 2.0.2 with this snippet. Please use the original snippet from #6501 as a test instead.

sidharthv96: Done.

j-f1:
How does this work with e.g. template strings containing LFs? Does it preserve them or does it change their content?

sidharthv96:
```javascript
console.log(
        `Multiline \n string\
         Multiline string\
         Multiline string`
      );
```
Will have the `\n` preserved (As they are escaped in the text representation). This only changes the actual line endings in the file.

sidharthv96:
@thorn0 any pointers on how to handle the codecov failure?

thorn0:
@sidharthv96 Failing because of "+-0.01%" looks like a misconfiguration. Ignore it.

thorn0:
@j-f1 I think your question was about **non-escaped** line breaks in template literals. So I did a little research. Turns out (see the [spec](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-template-literal-lexical-components)) template literals treat any _LineTerminatorSequence_ as LF, no matter how it's represented in the source code: 

image

This means it's totally safe to convert line endings in source files. Prettier's correctness goal isn't compromised.

tests/end_of_line/multiline.js
thorn0: The change log entry still contains the old snippet though.

thorn0:
@sidharthv96 Thanks for fixing this!

PR #7896, opened 2020-03-28
An attempt to fix coverage after updating Jest to 25.2.0

thorn0:
Seems like the coverage wasn't collected because of `transform:{}` in `jest.config.js`.
Any idea why we needed this line there? Apparently, it disables the Babel transform, so we probably should disable it conditionally or use `coverageProvider: "v8"`: https://github.com/prettier/prettier/pull/7897

closes https://github.com/prettier/prettier/issues/7893

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -28,7 +28,6 @@ module.exports = {
     "prettier/standalone": "<rootDir>/tests_config/require_standalone.js",
   },
   testEnvironment: "node",
-  transform: {},
   watchPlugins: [
     "jest-watch-typeahead/filename",
     "jest-watch-typeahead/testname",


SimenB:
Hey! This wasn't supposed to change with 25.2. we need Babel (or v8) to instrument code, but that should happen regardless of a transformer being defined or not

SimenB:
The issue is this line: https://github.com/facebook/jest/blob/2c666c21c6efc1fcadaf820cc9b47bf0ac10505a/packages/jest-transform/src/ScriptTransformer.ts#L230

Removing it, or forcing it to `false`, fixes the issue. It was added in https://github.com/facebook/jest/pull/9460

thorn0:
@SimenB It was late too, I couldn't make sure it's indeed a bug to report. Would you recommend switching to `coverageProvider: "v8"` on Node 13 or should we stick with Babel for now?

SimenB:
The V8 report is faster, while the Babel one is more accurate. I'd take a look at the html report locally and compare them, and decide based on that 🙂

thorn0:
E.g. V8 [reports that the line 212](https://codecov.io/gh/prettier/prettier/compare/412694e1c5ca77daba767c2a19d50f8b567f462b...c8435e81e4f73b69c7674d7ee81a3b62d6bb166e/src/src/main/comments.js#L212) is covered now whereas [according to Babel it's not](https://codecov.io/gh/prettier/prettier/compare/412694e1c5ca77daba767c2a19d50f8b567f462b...82bb912a4505efcad9da293fb48edb5f4a9ee2ba/src/src/main/comments.js#L212). Whom should I trust? :)

SimenB:
It's half red, not sure what that means?
image

thorn0:
These pages compare coverage for two given commits. Half-red means the line was not covered in the older commit but is covered now.

SimenB:
Aha. Probably wrong then 😀 If you could reduce it down and open up a bug report that'd be awesome. Probably something for `v8-to-istanbul`, but might be the implementation in Jest as well, so let's start there

thorn0:
So, sticking with Babel then.

> If you could reduce it down and open up a bug report that'd be awesome.

Added this to my OSS to-do list.

BTW, what should actually mean "covered" in this specific case?
 * "`() => false` was the result of the conditional operator",
 * "`() => false` was called",
 * either of the above happened,
 * both?

SimenB:
I _think_ ideally covered should mean "both" - coverage should be able to differentiate between "`else` case in ternary hit" and "function defined was called".

thorn0:
superseded by https://github.com/prettier/prettier/pull/7901

Thanks @SimenB !

PR #7900, opened 2020-03-29
Revert "Add please-upgrade-node"

GrahamCampbell:
This is causing false positives when prettier is packed with https://github.com/zeit/pkg. I'm unable to use prettier with node 12.16.1 after this change. Reverts prettier/prettier#7837.


thorn0:
How do you call `pkg`? What value is passed to the `--targets` option?

Issue #7910, opened 2020-03-30
Error When Combining Method Chaining with Sparse Array Arguments


han-tyumi:
**Environments:**

- Prettier Version: 2.0.2
- Running Prettier via: CLI via node_modules or VS Code
- Runtime: Node.js v12.14.1
- Operating System: macOS

**Steps to reproduce:**

Try formatting the following TypeScript code...
```typescript
new Test()
  .test()
  .test([, 0])
  .test();
```

**Expected behavior:**

The code is formatted without error.

**Actual behavior:**

```
src/index.ts[error] src/index.ts: TypeError: Cannot read property 'type' of null
[error]     at isSimpleCallArgument (/project/node_modules/prettier/bin-prettier.js:37144:30)
[error]     at isChildSimple (/project/node_modules/prettier/bin-prettier.js:37142:34)
[error]     at Array.every (<anonymous>)
[error]     at isSimpleCallArgument (/project/node_modules/prettier/bin-prettier.js:37163:26)
[error]     at /project/node_modules/prettier/bin-prettier.js:41502:109
[error]     at Array.every (<anonymous>)
[error]     at /project/node_modules/prettier/bin-prettier.js:41502:96
[error]     at Array.some (<anonymous>)
[error]     at printMemberChain (/project/node_modules/prettier/bin-prettier.js:41502:67)
[error]     at printPathNoParens (/project/node_modules/prettier/bin-prettier.js:38826:18)
```

**Observations**

I believe the issue is in the combination of `[, 0]` and method chaining.
It only occurs when methods are indented onto newlines and an array element is omitted.
Changing `[, 0]` to `[undefined, 0]` will fix the issue or by reducing the number of methods chained so that there is no need for newlines.

PR #7911, opened 2020-03-31
Fix bug with holes in array literals

bakkot:
Fixes https://github.com/prettier/prettier/issues/7910, so that running Prettier against

```js
new Test()
  .test()
  .test([, 0])
  .test();
```
no longer throws an error.


src/language-js/utils.js
```diff
@@ -958,7 +958,7 @@ function isSimpleCallArgument(node, depth) {
     );
   }
   if (node.type === "ArrayExpression") {
-    return node.elements.every(isChildSimple);
+    return node.elements.every((x) => x == null || isChildSimple(x));
```
fisker: Shouldn't this be `=== null` ?  holes always `null`

PR #7922, opened 2020-04-01
Update `isSCSS` test regex

fisker:
Fix `css_atrule/include.css` second format.

I don't really understand this regex, but space after `:` should be optional.

It's added in https://github.com/prettier/prettier/pull/4317/files#diff-29bcb0b461c4ae51930e975e53edfdd9L481

### DIFF
--- a/src/language-css/utils.js
+++ b/src/language-css/utils.js
@@ -60,7 +60,7 @@ function getPropOfDeclNode(path) {
 
 function isSCSS(parser, text) {
   const hasExplicitParserChoice = parser === "less" || parser === "scss";
-  const IS_POSSIBLY_SCSS = /(\w\s*: ?[^}:]+|#){|@import[^\n]+(url|,)/;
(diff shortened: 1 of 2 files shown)


src/language-css/utils.js
```diff
@@ -60,7 +60,7 @@ function getPropOfDeclNode(path) {
 
 function isSCSS(parser, text) {
   const hasExplicitParserChoice = parser === "less" || parser === "scss";
-  const IS_POSSIBLY_SCSS = /(\w\s*: [^}:]+|#){|@import[^\n]+(url|,)/;
+  const IS_POSSIBLY_SCSS = /(\w\s*: ?[^}:]+|#){|@import[^\n]+(url|,)/;
```
fisker, line 63: Should it be `*`?

evilebottnawi: I can't understand regexp too :smile: , I think we should rewrite it, on looking extension

fisker: You mean file extension? File extension is not always available, maybe check actual parser?

fisker:
Let's keep it this way to make test pass, we won't need it anymore after https://github.com/prettier/prettier/pull/7933

Issue #7401, opened 2020-01-20
Let's use GitHub actions / workflows for linting tests instead of Azure


lipis:
Suppose to be faster and it's without using 3rd party software. WDYT?

BPScott:
Yo @lipis, getting stuff consistent across the org would be handy - shall we say GH Actions is the preferred CI flow for all stuff? If so I'll switch eslint-plugin-prettier / stylelint-prettier over to GH Actions too.

thorn0:
We've done this, however the action `codecov-action` is [flaky](https://github.com/codecov/codecov-action/issues/68). If it stays like this, we should consider using Azure again for coverage-enabled testing as it was stable there.

lipis:
@fisker can you check what is going on?

fisker:
I really don't know why, people has this error too https://community.codecov.io/t/error-the-coverage-upload-was-rejected-because-the-workflow-run-is-either-not-in-progress-or-finished-a-while-ago/1116 , I suggest we can try upload with token.

thorn0:
@fisker "Keep in mind that secrets are not available to forks of repositories." It means that won't work for pull requests.

fisker:
babel use token https://github.com/babel/babel/blob/548cb3ee89552ffc08ee5625b084bd33f8107530/.github/workflows/coverage.yml#L29

lipis:
Let's do this!

fisker:
@lipis Can you setup `CODECOV_TOKEN` on https://github.com/prettier/prettier/settings/secrets?

lipis:
@fisker  I sent you an invite.. you should get access now.

fisker:
I don't have access to that page.

fisker:
Just click `Add a new secret` 

Name: CODECOV_TOKEN
Value: [token from this page](https://codecov.io/gh/prettier/prettier/settings)

lipis:
Added

PR #7930, opened 2020-04-01
Try upload coverage with token

fisker:
Ref: #7401

Remove `verbose` as requested in https://github.com/prettier/prettier/pull/7866#issuecomment-605892388


fisker:
Let's merge and see if it works on other incoming PRs?

thorn0:
Okay, let's see.

PR #7934, opened 2020-04-02
Wrap tests in `describe`

fisker:
We are running `format()` function outside `test`, if error throws, we won't see the failed filenames, it should called in `test` or `describe`.

- (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.


thorn0:
BTW, is there a good reason for not wrapping the preparation code in `beforeEach` or `beforeAll`?

fisker:
We definitely don't want `beforeEach` because we are  testing  different stuff, I  guess `beforeAll` is useful when we need do some  async job, in that  way  we  don't need  put  test in `.then`, or maybe  jest  need  early call  to do something ?  I guess  sync  code  has  no  different with  or  without `beforeAll`.

fisker:
With `--maxWorkers`, so tests are not running sequentially. So we have to wrap tests in `describe`

<details>
<summary>if SyntaxError can see the file name</summary>
<code><pre>
 FAIL  tests/pipeline_operator/jsfmt.spec.js
  fsharp_style_pipeline_operator.js
    × encountered a declaration exception (2ms)
  minimal_pipeline_operator.js
    √ format (3ms)
  smart_pipeline_operator.js
    √ format

  ● fsharp_style_pipeline_operator.js › encountered a declaration exception

    SyntaxError: Identifier directly after number (2:7)
      1 | promise
    > 2 |   |> 1await
        |       ^
      3 |   |> x => doubleSay(x, ', ')
      4 |   |> capitalize
      5 |   |> x => x + '!'

      3 | function createError(message, loc) {
      4 |   // Construct an error similar to the ones thrown by Babel.
    > 5 |   const error = new SyntaxError(
        |                 ^
      6 |     message + " (" + loc.start.line + ":" + loc.start.column + ")"
      7 |   );
      8 |   error.loc = loc;

      at createError (src/common/parser-create-error.js:5:17)
</pre></code>
</details>

tests_config/run_spec.js
```diff
@@ -69,126 +69,126 @@ global.run_spec = (dirname, parsers, options) => {
...
+          }
         });
       }
-    }
+
+      if (AST_COMPARE) {
+        const formatted = output.replace(CURSOR_PLACEHOLDER, "");
+
+        if (formatted !== input) {
+          test("parse", () => {
```
thorn0, line 183:
```suggestion
          test("compare AST", () => {
```

PR #7967, opened 2020-04-06
Refactor glimmer-printer

dcyriller:
#### Description

I'm doing a POC to introduce a `hbsWhitespaceSensitivity` option.

On the way, I noticed that we could group / simplify / extract / remove some code. It might be easier to review it commit by commit. I could split this PR in smaller chunks if you'd like.

#### Checklist

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -18,6 +18,7 @@ const {
   getPreviousNode,
   hasPrettierIgnore,
   isNextNodeOfSomeType,
+  isNodeOfSomeType,
   isParentOfSomeType,
   isPreviousNodeOfSomeType,
   isVoid,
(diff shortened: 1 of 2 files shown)


src/language-handlebars/printer-glimmer.js
```diff
@@ -355,18 +248,13 @@ function print(path, options, print) {
...
-            trailingLineBreaksCount,
-            maxLineBreaksToPreserve
-          ),
-        ].filter(Boolean)
-      );
+      return concat([
+        ...generateHardlines(leadingLineBreaksCount, maxLineBreaksToPreserve),
+        n.chars
+          .replace(/^[\s ]+/g, leadingSpace)
+          .replace(/[\s ]+$/, trailingSpace),
```
fisker, line 255:
```suggestion
          .replace(/^\s+/g, leadingSpace)
          .replace(/\s+$/, trailingSpace),
```

dcyriller: strange indeed 🤔

thorn0:
As for `hbsWhitespaceSensitivity`, let's use the existing option `htmlWhitespaceSensitivity`. Otherwise, it'd be strange that there is no `vueWhitespaceSensitivity` and `angularWhitespaceSensitivity`.

dcyriller:
Noted 👍

Issue #7949, opened 2020-04-03
Potential Breaking Change in `resolveConfig`


ntotten:
It seems that 2.0 may have introduced an unintended breaking change in the `resolveConfig` function. This function is used in the Prettier VS Code extension here: https://github.com/prettier/prettier-vscode/blob/master/src/ConfigResolver.ts#L99

Since upgrading to 2.0 we are now seeing errors when trying to resolve configs:

```
Error: Cannot find module 'prettier-config-greenelab' from 'c:\Users\Vincent\Desktop\adage-frontend'
	at Function.Module._resolveFilename (internal/modules/cjs/loader.js:717:15)
	at Function.n.resolve (c:\Program Files\Microsoft VS Code\resources\app\out\vs\loader.js:15:584)
	at Object.transform (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\index.js:23283:40)
	at run (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11472:51)
	at async cacheWrapper (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11300:20)
	at async cacheWrapper (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11300:20)
	at async cacheWrapper (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11300:20)
	at async cacheWrapper (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11300:20)
	at async cacheWrapper (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11300:20)
	at async Explorer.search (c:\Users\Vincent\.vscode\extensions\esbenp.prettier-vscode-4.0.0\node_modules\prettier\third-party.js:11457:22)
	at async Promise.all (index 0)
```
 
For repro steps and more details see: https://github.com/prettier/prettier-vscode/issues/1289

One thing to note, is that everything behaves as expected from the CLI which is odd.

ntotten:
@lipis Opened an issue in prettier repo to track.

[...]

PR #7973, opened 2020-04-07
Throw original error when loading external config

fisker:
See https://github.com/prettier/prettier/issues/7949#issuecomment-608628125


fisker:
These tests won't pass, I'm still thinking about solution.

thorn0:
You can probably use `expect.stringMatching` and `expect.stringContaining` instead of snapshots, like I did in `patterns-dirs.js`.

fisker:
Thanks I'll try

PR #7974, opened 2020-04-07
issue template: insist on using the Playground

thorn0:
Now people tend to ignore this "tip".

### DIFF
--- a/.github/ISSUE_TEMPLATE/formatting.md
+++ b/.github/ISSUE_TEMPLATE/formatting.md
@@ -15,16 +15,35 @@ BEFORE SUBMITTING AN ISSUE:
 
 2.  We get a lot of requests for adding options, but Prettier is
     built on the principle of being opinionated about code formatting.
-    This means we have a very high bar for adding new options.
+    This means we add options only in the case of strict technical necessity.
     Find out more: https://prettier.io/docs/en/option-philosophy.html
 
-👉 Still sure you want to submit an issue? 🛑 DON'T USE THIS FORM.
-🤖 Let a program fill the issue template for you!
+Don't fill the form below manually! Let a program create a report for you:
 
 1. Go to https://prettier.io/playground
 2. Paste your code and set options
 3. Press the "Report issue" button in the lower right
-4. Enter a title and describe the expected behavior
-5. Submit
 
 -->
+
+**Prettier 2.2.1**
+[Playground link](https://prettier.io/playground/#.....)
+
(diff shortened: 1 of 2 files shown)


fisker:
How about remove template bellow, so people have to use playground?

fisker:
Why not just 1-7

```markdown

1.  Search for your issue on GitHub: https://github.com/prettier/prettier/issues
    A large number of opened issues are duplicates of existing issues.
    If someone has already opened an issue for what you are experiencing,
    you do not need to open a new issue — please add a 👍 reaction to the
    existing issue instead.

2.  We get a lot of requests for adding options, but Prettier is
    built on the principle of being opinionated about code formatting.
    This means we have a very high bar for adding new options.
    Find out more: https://prettier.io/docs/en/option-philosophy.html

3.  Go to https://prettier.io/playground

4.  Paste your code and set options

5.  Press the "Report issue" button in the lower right

6.  Enter a title and describe the expected behavior

7.  Submit
```

.github/ISSUE_TEMPLATE/formatting.md
```diff
@@ -18,32 +18,13 @@ BEFORE SUBMITTING AN ISSUE:
...
 
-Tip! Don't write this stuff manually.
+👉 Still sure you want to submit an issue? 🛑 DON'T USE THIS FORM.
+🤖 Let a program fill the issue template for you!
 
 1. Go to https://prettier.io/playground
 2. Paste your code and set options
 3. Press the "Report issue" button in the lower right
+4. Enter a title and describe the expected behavior
+5. Submit
```
evilebottnawi, line 28: I am afraid please will be ignore an issue templat - with simple words, they did not see the comment. From my experience, everything in the comments is ignored by many

thorn0:
> Why not just 1-7

Because "Search for your issue on GitHub..." looks like standard text that people skip. We need to draw their attention to the playground somehow.

fisker:
I like [this issue template](https://github.com/vuejs/vue/issues/new), simple.

seaneking:
Hey so as someone who ran into this myself, if you really want to stop people using GH issues directly I'd take a harder line. Eg: something like Emotion's support question issue template:

Screen Shot 2020-05-02 at 11 41 13 AM

fisker:
I  visit the  vue  project again, they are  using  a  direct external link to fill  the  issue now.

https://github.com/vuejs/vue/pull/10777/files

PR #7978, opened 2020-04-07
Update steps for running Prettier on save in WebStorm

prigara:
WebStorm 2020.1 makes it easier to configure Prettier to [run on save](https://www.jetbrains.com/webstorm/whatsnew/#running-prettier-on-save).

I've updated the docs accordingly. 


prigara:
Sorry about having all the previous commits I've made visible in the PR. They should have been all merged back in 2018.

thorn0:
What about the WebStorm link on the main page of prettier.io? Should we update it too?

prigara:
@thorn0 the support for Prettier in WebStorm is about out-of-the-box. No changes here.
Do you suggest adding a link to https://prettier.io/docs/en/webstorm.html?

thorn0:
>  support for Prettier in WebStorm is about out-of-the-box

Whereas the link leads to a plugin. Is that okay?

> Do you suggest adding a link to https://prettier.io/docs/en/webstorm.html?

Probably. Or we can put multiple links there. There are multiple links for other editors. You decide what would work best for WebStorm users.

prigara:
Yeah, let's point the link to https://prettier.io/docs/en/webstorm.html. Will push a change to this branch.

PR #7980, opened 2020-04-07
An attempt to fix unexpected GitHub Actions failures

thorn0:
Seems like GitHub Actions have the `CI` environment variable set now whereas some tests (added in #5804) depended on this variable.


thorn0:
Yes, this is it. Does GitHub announce such changes anywhere?

fisker:
Great!

fisker:
How  about `{mockIsCI: () => true}`,  clear to me .

thorn0:
It's not dynamic. Why define this option as a function?

fisker:
I was thinking `cosmiconfig` mock has argument, in this way we can allow option to mock it.

PR #8011, opened 2020-04-10
add information about how to run Prettier on save

ryababukha:
WebStorm 2020.1 comes with an option for running Prettier on save, so file watchers are no longer required. More info is available here: https://blog.jetbrains.com/webstorm/2020/04/webstorm-2020-1/#tools, it might be worth adding a link to this blog post as well.


thorn0:
Thanks, but that information is already there in this file.

fisker:
#7978

PR #8012, opened 2020-04-10
Update spellcheck config and scripts

fisker:
- Ignore bold names, [plugins authors are wrapped with `**`](https://github.com/prettier/prettier/pull/7994/files#diff-0916b758c148ca9d7bbfb241d34d747eR55
).
- Lint more files.
- Add `cspell.json` clean script.
- Cleanup `cspell.json`

-----


fisker:
Also, there are 80 words not using, should I clean it?

thorn0:
Probably we just need to include more files to `lint:spellcheck`. Those words might have come from those files (e.g. `jest.config.js`).

tests_integration/__tests__/file-info.js
```diff
@@ -77,7 +77,7 @@ describe("extracts file-info with inferredParser=null for file.foo", () => {
   });
 });
 
-describe("extracts file-info with inferredParser=foo when plugins are autoloaded", () => {
+describe("extracts file-info with inferredParser=foo when plugins are auto loaded", () => {
```
thorn0, line 80: https://en.wiktionary.org/wiki/autoload

cspell.json
```diff
@@ -68,376 +63,333 @@
...
         "linebreak",
         "linebreaks",
         "literalline",
         "literallines",
-        "lockfile",
         "loglevel",
         "lowercased",
         "lowercasing",
         "lydell",
-        "macos",
```
thorn0, line 214: Seems like `.github/` is still excluded. "macos" is from the GitHub Actions files.

fisker: Seems hidden files excluded by default

package.json
```diff
@@ -125,7 +125,7 @@
     "lint:changelog": "node ./scripts/lint-changelog.js",
     "lint:prettier": "prettier \"**/*.{md,json,yml,html,css}\" --check",
     "lint:dist": "eslint --no-eslintrc --no-ignore --env=es6,browser --parser-options=ecmaVersion:2016 \"dist/!(bin-prettier|index|third-party).js\"",
-    "lint:spellcheck": "cspell {bin,scripts,src,website}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md",
+    "lint:spellcheck": "cspell **/* .github/**/*",
```
thorn0, line 128: I guess we should put quotes around `**/*`, otherwise it's expanded by the shell.

fisker: Right

thorn0: `.github/workflows/*` must have been excluded for this reason.

scripts/clean-cspell.js
```diff
@@ -0,0 +1,41 @@
...
+
+  console.log("Running spellcheck with empty words ...");
+  try {
+    await execa("yarn lint:spellcheck");
+  } catch (error) {
+    const { stdout } = error;
+    const words = [
+      ...stdout.matchAll(/ - Unknown word \((.*?)\)/g),
+    ].map(([, word]) => word.toLowerCase());
+    config.words = [...new Set(words)].sort();
```
thorn0, line 30: Please use the same order as the VS Code extension: https://github.com/streetsidesoftware/vscode-spell-checker/blob/2fde3bc5c658ee51da5a56580aa1370bf8174070/packages/client/src/settings/CSpellSettings.ts#L78

cspell.json
```diff
@@ -1,18 +1,17 @@
 {
     "version": "0.1",
     "words": [
-        "ACMR",
```
thorn0, line 4: Please don't lowercase words. cSpell **is** case-sensitive: if it has only "ASTs" in the dictionary, it marks "asts" as an error.

package.json
```diff
@@ -125,7 +125,7 @@
     "lint:changelog": "node ./scripts/lint-changelog.js",
     "lint:prettier": "prettier \"**/*.{md,json,yml,html,css}\" --check",
     "lint:dist": "eslint --no-eslintrc --no-ignore --env=es6,browser --parser-options=ecmaVersion:2016 \"dist/!(bin-prettier|index|third-party).js\"",
-    "lint:spellcheck": "cspell {bin,scripts,src,website}/**/*.js {docs,website/blog,changelog_unreleased}/**/*.md",
+    "lint:spellcheck": "cspell \"**/*\" \".github/**/*\"",
```
thorn0, line 128: Is `.github/**/*` still needed after you added the quotes around `**/*`?

fisker: Yep, still not checking

```text
node -p "require('./cspell.json').words.find(x => x.toLowerCase() === 'macos')" && yarn lint:spellcheck
undefined
yarn run v1.22.4
$ cspell "**/*"
CSpell: Files checked: 1450, Issues found: 0 in 0 files
Done in 30.45s.
```

package.json
fisker: I guess because `node-glob` ignore `dot` file by default https://github.com/isaacs/node-glob#options

Can't see any override https://github.com/streetsidesoftware/cspell/blob/master/packages/cspell-glob/src/GlobMatcher.ts

cspell.json
```diff
@@ -22,45 +22,50 @@
         "autofix",
         "autogenerated",
         "autolink",
+        "Autolinks",
         "autolinks",
```
thorn0, line 26: If the word in the dictionary is all-lowercase, then cspell is case-insensitive with regard to that word. E.g. if it has "autolinks", then "Autolinks" is not an error, but not the other way around. Please remove "Autolinks" and "ALGOLIA" if the dictionary already includes "autolinks" and "algolia".

fisker: Okay...

thorn0:
Did I break the CI with https://github.com/prettier/prettier/commit/8557ea69af830cdaa8a887a472aca35a66a44c82 ? 🤔

fisker:
I don't think so

> GitHub Actions has encountered an internal error when running your job.

Try restart

thorn0:
Where did you read that? Note that the checks in this PR didn't run too.

fisker:
I got that from master jobs, maybe broken...

Try add that `patch-release` branch

fisker:
Jobs can't cancel either.

thorn0:
I created the `patch-release` branch. I'm going to close and reopen this PR to see what will happen.

fisker:
I've revert that change, we can try update only one `lint` job later.

I think you can change workflow config file directly in in `patch-release` branch if you want do release now.

thorn0:
The Actions are back! I think I'll just open a PR for `patch-release`.

fisker:
I'm testing on my fork. Seems no problem with that commit https://github.com/fisker/prettier/commits/master 

PR https://github.com/fisker/prettier/pull/342

thorn0:
Probably GitHub's UI just needed time to sync with the new settings.

thorn0:
https://www.githubstatus.com/incidents/2wx06v7dbzhg

fisker:
> This incident has been resolved.

https://www.githubstatus.com/incidents/2wx06v7dbzhg

PR #8027, opened 2020-04-12
Cache CI build result

fisker:
The build cache system seems [designed to speed up CI build](https://github.com/prettier/prettier/pull/4693), but lost in last CI setup.

With cache, ~12s https://github.com/prettier/prettier/pull/8027/checks?check_run_id=580817579#step:7:22

For reference: this caching was implemented in https://github.com/prettier/prettier/pull/4693

### DIFF
--- a/.github/workflows/prod-test.yml
+++ b/.github/workflows/prod-test.yml
@@ -24,6 +24,23 @@ jobs:
       - name: Install Dependencies
         run: yarn install --frozen-lockfile
 
+      - name: Cache Build Results
+        # Prevent error on re-run
+        # https://github.com/actions/cache/issues/107#issuecomment-556037278
+        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        id: build-cache
+        uses: actions/cache@v1
+        with:
+          path: .cache
+          key: v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.ref }}-
+          restore-keys: |
+            v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.ref }}-
+            v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.base_ref }}-
+            v1-build-cache-${{ hashFiles('yarn.lock') }}-master-
+
+      # Unknown reason `scripts/build/build.js` created dir can't work
+      - run: mkdir -p .cache/files
+
       - name: Build Package
         run: yarn build
 
--- a/cspell.json
+++ b/cspell.json
@@ -206,6 +206,7 @@
         "mitermayer",
         "mixins",
         "mjml",
+        "mkdir",
         "Moeller",
         "Monteiro",
         "Morrell",


.github/workflows/prod-test.yml
```diff
@@ -24,6 +24,23 @@ jobs:
...
+        with:
+          path: .cache
+          key: v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.ref }}-
+          restore-keys: |
+            v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.ref }}-
+            v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.base_ref }}-
+            v1-build-cache-${{ hashFiles('yarn.lock') }}-master-
+
+      # Unknown reason `scripts/build/build.js` created dir can't work
+      - run: mkdir -p .cache/files
```
fisker, line 42: Littler hard to debug on CI, havn't found the reason. Maybe js created folder don't have the right permission.

Without this, the `manifest.json` file seems can't write.

.github/workflows/prod-test.yml
```diff
@@ -24,6 +24,23 @@ jobs:
...
 
+      - name: Cache Build Results
+        # Prevent error on re-run
+        # https://github.com/actions/cache/issues/107#issuecomment-556037278
+        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        id: build-cache
+        uses: actions/cache@v1
+        with:
+          path: .cache
+          key: v1-build-cache-${{ hashFiles('yarn.lock') }}-${{ github.ref }}-
```
fisker, line 35: The hash not really necessary, but the cache can't be deleted/updated we need a way to update the cache. https://github.com/actions/cache/issues/2

.github/workflows/prod-test.yml
```diff
@@ -24,6 +24,23 @@ jobs:
...
         run: yarn install --frozen-lockfile
 
+      - name: Cache Build Results
+        # Prevent error on re-run
+        # https://github.com/actions/cache/issues/107#issuecomment-556037278
+        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        id: build-cache
+        uses: actions/cache@v1
+        with:
+          path: .cache
```
evilebottnawi, line 34: Shouldn't it be `yarn cache dir`? https://github.com/actions/cache/blob/master/examples.md#node---yarn

fisker: You misunderstood, we have a dir [`.cache` in project root](https://github.com/prettier/prettier/blob/232661c13652d0649113aa10f0332f9934b86df4/scripts/build/build.js#L69), after building process, the result will cached there, so on later build, if cache hit, we won't need bundle this file.  Nothing related to yarn.

fisker:
I'm merging this, hope nothing goes  wrong :pray:  .

lipis:
> I'm merging this, hope nothing goes wrong 🙏 .

Last famous words!

PR #8031, opened 2020-04-12
GH Actions: Cancel Previous Runs

fisker:
Cancel redundant workflows


fisker:
Can't make it work, will try later

PR #8040, opened 2020-04-13
website: clean up readme, remove old deploy script

thorn0:


fisker:
Want run `draft-blog-post.js` when building website? So we can see how changelog will be 
 rendered.

thorn0:
@fisker It'd good to publish it somehow, but not in the blog. Rather somewhere under https://prettier.io/docs/en/next/

PR #8058, opened 2020-04-14
Add `prettier/standard` to list of eslintrc extends. 

hadron-mhchu:
This will disable additional rules like "standard/computed-property-even-spacing" which may conflict with prettier for some long lines (especially with arrays)

This is a simple documentation change. No tests.

### DIFF
--- a/docs/integrating-with-linters.md
+++ b/docs/integrating-with-linters.md
@@ -25,7 +25,7 @@ Then in `.eslintrc.json`:
 
 ```json
 {
-  "extends": ["prettier"]
+  "extends": ["prettier", "prettier/standard"]
 }
 ```
 


thorn0:
Thanks, but we want to keep that page simple. Please consider contributing to https://github.com/prettier/eslint-config-prettier instead.

hadron-mhchu:
Ah, thanks for the clarification - I did not realize that we needed it only because we were using `standard` ESLint plugin. I guess that makes sense then to leave it out. I have to admit that eslint + prettier integration was not as smooth as I had hoped and we keep finding small incompatibilities that need to be resolved. In this case, including `prettier/standard` and I misunderstood why we needed to include it. Thanks for the explanation!

PR #8061, opened 2020-04-15
Refactor printer-estree.js: extract printing of member chains and call args to separate files

thorn0:
I know this is not a great idea from the git blame point of view, but IMHO this needs to be done. `printMemberChain` is a huge function and changing it in #7889 is painful. After extracting it to a separate module, we'll be able to decompose it further into smaller functions scoped to that module.

### DIFF
--- a/src/language-js/print/call-arguments.js
+++ b/src/language-js/print/call-arguments.js
@@ -1,8 +1,8 @@
 "use strict";
 
-const comments = require("../main/comments");
-const { getLast, getPenultimate } = require("../common/util");
-const { isNextLineEmpty } = require("../common/util-shared");
+const comments = require("../../main/comments");
+const { getLast, getPenultimate } = require("../../common/util");
+const { isNextLineEmpty } = require("../../common/util-shared");
 const {
   hasLeadingComment,
   hasTrailingComment,
@@ -10,7 +10,7 @@ const {
   isJSXNode,
   isLongCurriedCallExpression,
   shouldPrintComma,
-} = require("./utils");
+} = require("../utils");
 
 const {
   builders: {
@@ -25,7 +25,7 @@ const {
     breakParent,
   },
   utils: { willBreak },
-} = require("../document");
+} = require("../../document");
 
 function printCallArguments(path, options, print) {
   const node = path.getValue();
--- a/src/language-js/utils.js
+++ b/src/language-js/utils.js
@@ -1002,6 +1002,26 @@ function isTSXFile(options) {
   return options.filepath && /\.tsx$/i.test(options.filepath);
 }
 
+function shouldPrintComma(options, level) {
+  level = level || "es5";
+
+  switch (options.trailingComma) {
+    case "all":
+      if (level === "all") {
+        return true;
+      }
+    // fallthrough
+    case "es5":
+      if (level === "es5") {
+        return true;
+      }
+    // fallthrough
+    case "none":
+    default:
+      return false;
+  }
+}
+
 module.exports = {
   classChildNeedsASIProtection,
   classPropMayCauseASIProblems,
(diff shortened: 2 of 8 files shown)


fisker:
How about a directory? The code will keep  grow, some day we'll say  too much files in  this dir .

src/language-js/print/call-arguments.js
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  // useEffect(() => { ... }, [foo, bar, baz])
+  if (
+    args.length === 2 &&
+    args[0].type === "ArrowFunctionExpression" &&
+    args[0].params.length === 0 &&
+    args[0].body.type === "BlockStatement" &&
+    args[1].type === "ArrayExpression" &&
+    !args.find((arg) => arg.comments)
```
fisker, line 49:
```suggestion
    !args.some((arg) => arg.comments)
```

src/language-js/print/call-arguments.js
```diff
@@ -0,0 +1,295 @@
...
+      ", ",
+      path.call(print, "arguments", 1),
+      ")",
+    ]);
+  }
+
+  // func(
+  //   ({
+  //     a,
+
```
fisker, line 63:
```suggestion
  //
```

DI think this empty line in code?

src/language-js/print/call-arguments.js
```diff
@@ -0,0 +1,295 @@
...
+}
+
+function couldGroupArg(arg) {
+  return (
+    (arg.type === "ObjectExpression" &&
+      (arg.properties.length > 0 || arg.comments)) ||
+    (arg.type === "ArrayExpression" &&
+      (arg.elements.length > 0 || arg.comments)) ||
+    (arg.type === "TSTypeAssertion" && couldGroupArg(arg.expression)) ||
+    (arg.type === "TSAsExpression" && couldGroupArg(arg.expression)) ||
```
fisker, line 236: This 4 branches can merge to 2.

src/language-js/utils.js
```diff
@@ -1002,6 +1002,26 @@ function isTSXFile(options) {
...
+    // fallthrough
+    case "es5":
+      if (level === "es5") {
+        return true;
+      }
+    // fallthrough
+    case "none":
+    default:
+      return false;
+  }
```
fisker, line 1022:
```js
  return options.trailingComma == level && (level == "all" || level = "es5");
```

thorn0:
@fisker This PR simply extracts the code from `printer-estree` as is. Let's do further refactoring as separate changes.

src/language-js/print/call-arguments.js
fisker: Sorry, wrong.

fisker:
Please take a look at the build log when you merge,  should only js related file (maybe  index.js standalone.js parser-{babel,flow,typescript.js)need rebuild.

thorn0:
@fisker 

```
yarn run v1.22.4
$ node --max-old-space-size=3072 ./scripts/build/build.js
 Building packages 
index.js........................  DONE  
doc.js.......................... CACHED 
standalone.js...................  DONE  
bin-prettier.js.................  DONE  
third-party.js.................. CACHED 
parser-babel.js................. CACHED 
parser-flow.js.................. CACHED 
parser-typescript.js............ CACHED 
parser-angular.js............... CACHED 
parser-postcss.js............... CACHED 
parser-graphql.js............... CACHED 
parser-markdown.js.............. CACHED 
parser-glimmer.js............... CACHED 
parser-html.js.................. CACHED 
parser-yaml.js.................. CACHED 
Done in 49.66s.
```

> While you are at this, why not? We'll foget.

Because changes have their scope. The intent of this change is only to move the code across files to make it more modular.

PR #8062, opened 2020-04-15
Fix master ref

fisker:
Seems the master ref is not right, let's see if this PR hit the cache.

Fix  mistake in  https://github.com/prettier/prettier/pull/8027


fisker:
First commit hit the master cache https://github.com/prettier/prettier/runs/589185294?check_suite_focus=true#step:5:10

Waiting for second hit PR cache.

fisker:
Great, work  as  expected  https://github.com/prettier/prettier/runs/589213198?check_suite_focus=true#step:5:11

PR #8063, opened 2020-04-15
Avoid line-breaking inside a method chain before breaking the chain itself

thorn0:
Extracted from #7889. See https://github.com/prettier/prettier/pull/7889#issuecomment-610725563


rattrayalex:
@thorn0 would you like help getting this over the line?

PR #7951, opened 2020-04-04
workaround for require.resolve in prettier-vscode

thorn0:
fixes #7949

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #8072, opened 2020-04-17
Use `resolve` if `require.resolve` is overridden

### LINKED ISSUE #8047: Uncaught TypeError: createRequire is not a function
**Environments:**

- Prettier Version: 2.0.4
- Running Prettier via: atom-prettier
- Runtime: ?
- Operating System: Linux, macOS, windows

**Steps to reproduce:**

Please see https://github.com/prettier/prettier-atom/issues/725

Thanks!

fisker:
~This reverts #7508 & #7951~

Fixes #8047


thorn0:
No, we shouldn't revert it. Let's just use it if `require.resolve` is broken. Otherwise we'll break https://github.com/prettier/prettier/issues/7407.

fisker:
I missed that.

fisker:
I need go out for couple hours now.

src/common/resolve.js
```diff
@@ -4,9 +4,14 @@ let { resolve } = eval("require");
 
 // In the VS Code extension `require` is overridden and `require.resolve` doesn't support the 2nd argument.
 if (resolve.length === 1) {
-  const Module = eval("require")("module");
-  const createRequire = Module.createRequire || Module.createRequireFromPath;
-  resolve = createRequire(__dirname).resolve;
+  resolve = (id, options) => {
+    let baseDir;
```
thorn0, line 8: It's all-lowercase: `basedir`

src/common/resolve.js
```diff
@@ -4,9 +4,14 @@ let { resolve } = eval("require");
...
+    if (options && options.paths && options.paths.length === 1) {
+      baseDir = options.paths[0];
+    }
+
+    return require("resolve").sync(id, { baseDir });
```
thorn0, line 13: With this conditional `require` call, we might face the problem you mentioned [here](https://github.com/prettier/prettier/pull/8036#issuecomment-614620428). We should check how it will look in the bundle, or better think of a way to test this fallback.

src/common/resolve.js
fisker: Thanks.

src/common/resolve.js
fisker: Don't know, need check dist.

fisker: I was trying to wrote a test couple days ago, but failed.

I tried to override `require.resolve` function before test, and restore after test.

thorn0: Overriding is too complicated. I suggest that we simply use an environment variable to trigger the fallback: `resolve.length === 1 || process.env.PRETTIER_FALLBACK_RESOLVE`.

fisker:
I'm going to disable length check to run a test , so it will always use `resolve`, just make sure we are not breaking things.

src/common/resolve.js
fisker: Yes, let's use env to run the test, but before that. I'm going to refactor production test first.

.github/workflows/prod-test.yml
```diff
@@ -131,3 +131,10 @@ jobs:
       - name: Run Tests (standalone) (Linux and Windows)
         if: matrix.os != 'macos-latest'
         run: yarn test:dist-standalone --maxWorkers=2
+
+      # #8073 test
+      - name: Run Tests (PRETTIER_FALLBACK_RESOLVE)
```
thorn0, line 136: We need to change the order of the steps. The coverage report from this run is uploaded to Codecov.

fisker: I thought they can merge.

thorn0: Seems like [they can](https://docs.codecov.io/docs/merging-reports) if you upload them separately: run test1, upload report1, run test2, upload report2.

fisker: Sorry, I know codecov can, I mean jest should.

thorn0: No, it doesn't do that.

fisker:
@thorn0 I checked the `index.js` file in `dist`,  on L24254

```js
let {
  resolve: resolve$1
} = eval("require");
```

I remember this eval can't run in atom? 

[Link](https://github.com/prettier/prettier/issues?q=is%3Aissue++sort%3Aupdated-desc+atom+eval)

fisker:
`eval('require').cache` used in `third-party.js` because https://github.com/prettier/prettier/blob/ea8330a22fb944bb6a0ccb0ec1546d842db7a97f/scripts/build/config.js#L173

src/common/resolve.js
```diff
@@ -2,11 +2,16 @@
 
 let { resolve } = eval("require");
```
thorn0, line 3:
```suggestion
let resolve = eval("require").resolve;
```

fisker: I've tried this,  won't work , this will replaced by babel https://github.com/prettier/prettier/blob/master/scripts/build/babel-plugins/transform-custom-require.js
to `require.resolve` , then rollup replace it to `commonJsRequire.resolve`

fisker: I'm not sure, but `global.require.resolve` seems a  good solution.

thorn0: Are you sure? I tested it too locally and it stays this way in `dist/index.js`:
```js
let resolve$1 = require.revolve; // In the VS Code and Atom extension `require` is overridden and `require.resolve` doesn't support the 2nd argument.

if (resolve$1.length === 1 || process.env.PRETTIER_FALLBACK_RESOLVE) {
```

thorn0: `global.require` exists only in the REPL.

thorn0: Strangely. locally ESLint doesn't issue a warning for this line for me. 🤔

fisker: Hmm, I guess I tried `let { resolve } = require;` and`let resolve = eval("require").resolve;`. 

Eslint didn't warn on my machine either, we need add comments and disable the rule on it anyway.

thorn0:
@fisker I think everything is good here. Could you please re-check?

src/common/resolve.js
```diff
@@ -1,12 +1,18 @@
 "use strict";
 
-let { resolve } = eval("require");
+// eslint-disable-next-line prefer-destructuring
+let resolve = eval("require").resolve;
```
fisker, line 4: We should add comment explain why this is disabled, because this can't understand `/scripts/build/babel-plugins/transform-custom-require.js` 

I'm not near my laptop, can you do this?

fisker:
Everything should be good, thanks. Will you fix this https://github.com/prettier/prettier/pull/8072#issuecomment-616482737 ?

thorn0:
Is it broken? I thought that replacement happens for a reason.

src/common/resolve.js
thorn0: ok

fisker:
I remember you  add it ? I don't know if it can run in atom , I just saw it in `third-party.js`

thorn0:
It wasn't me. I fixed only this thing: https://github.com/prettier/prettier/pull/7820

fisker:
#5558 add it, if  it's true  that atom  can't run  eval , we  need fix it.

thorn0:
Okay. I'm downloading Atom. :D

fisker:
Can you also check if it support `new Function("return this")()`? 

I believe it's also used.

fisker:
I'm  going to sleep, good luck!

fisker:
@thorn0 I tested in atom, both prettier files and `prettier.config.js` file can run `eval`.

thorn0:
I haven't been able to reproduce problems with `eval` and `new Function` in Atom. However, I confirmed that the fallback from this PR works there.

fisker:
And I logged `console.log({global, require})` in our source, and played with it in console

```js
temp2.global.require
ƒ require(path) {
    try {
      exports.requireDepth += 1;
      return mod.require(path);
    } finally {
      exports.requireDepth -= 1;
    }
  }


temp2.require
ƒ require(e){return n.require(e)}


temp2.global.require.resolve
ƒ resolve(request, options) {
    if (typeof request !== 'string') {
      throw new ERR_INVALID_ARG_TYPE('request', 'string', request);
    }
    return Module._resolveFilename(request, mod, false, op…


temp2.require.resolve
ƒ (e){return get_Module()._resolveFilename(e,n)}
```

`global.require.resolve` looks like the native `require.resolve`

thorn0:
The native `require.resolve` should have 2 parameters.

thorn0:
It's not native. `Module._resolveFilename` that it calls has `length === 2` and ignores `options`. Execute `debugger; global.require.resolve('a', {paths:['b']})` to explore that function in the debugger.

PR #8073, opened 2020-04-17
Refactor production tests

fisker:
- Auto install production packge before running tests in `jest.config.js`(can't figure a better way).
- Refactor `STANDALONE` tests to the same way as `production-test`

In this way, we can add more seperate tests on bundles.

To make [this test](https://github.com/prettier/prettier/pull/8072#discussion_r410216481) runs on bundles:

Without this change, we have to modify `scripts/test-dist.js` add more `if/else`

Now, We only need 
- `cross-env PRETTIER_FALLBACK_RESOLVE=1 jest tests_integration/__tests__/{config-resolution,config-invalid,plugin-resolution}.js` to run tests on source
- `cross-env NODE_ENV=production PRETTIER_FALLBACK_RESOLVE=1 jest tests_integration/__tests__/{config-resolution,config-invalid,plugin-resolution}.js`  to run tests on bundles

Also easier to run production tests with pattern `npx cross-env NODE_ENV=production jest tests/big-int`

---
I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).


.github/workflows/prod-test.yml
```diff
@@ -116,8 +116,13 @@ jobs:
...
-        run: yarn test:dist
+      - name: Run Tests (macOS)
+        if: matrix.os == 'macos-latest'
+        run: yarn test:dist --maxWorkers=4
+
+      - name: Run Tests (Linux and Windows)
+        if: matrix.os != 'macos-latest'
+        run: yarn test:dist --maxWorkers=2
 
   standalone:
```
fisker, line 127: I'll try to merge `production-test` and `standalone-test` later.

jest.config.js
```diff
@@ -1,5 +1,11 @@
 "use strict";
+
+const installPrettier = require("./scripts/install-prettier");
+
 const ENABLE_CODE_COVERAGE = !!process.env.ENABLE_CODE_COVERAGE;
+if (process.env.NODE_ENV === "production") {
```
fisker, line 6: I don't think use `NODE_ENV` environment is good idea, maybe change to another variable  later

thorn0:
> To make this test runs on bundles:
> Without this change, we have to modify scripts/test-dist.js add more if/else

I don't understand this part. Why do we have to modify that file to add that test? If anything, I thought we would modify `tests_integration/runPrettier.js` to mock `process.env` and restore it after the test. Adding `try ... finally` around the `runPrettier` could work too. Aren't you overthinking this a bit?

fisker:
> > To make this test runs on bundles:
> > Without this change, we have to modify scripts/test-dist.js add more if/else
> 
> I don't understand this part. Why do we have to modify that file to add that test? If anything, I thought we would modify `tests_integration/runPrettier.js` to mock `process.env` and restore it after the test. Adding `try ... finally` around the `runPrettier` call could work too. Aren't you overthinking this a bit?

How to run full `config-resolution.js` and `plugin-resolution.js` twice?

fisker:
Anyway, I think better in this way.

> run production tests with pattern `npx cross-env NODE_ENV=production jest tests/big-int`

This is much easier.

thorn0:
> How to run full config-resolution.js and plugin-resolution.js twice?

By wrapping them in a loop. But I think just one test for that fallback would be enough.

thorn0:
After a second thought, I think it's okay to test it in a separate step of the GitHub Actions workflow. It just feels unusual to me.

PR #8080, opened 2020-04-18
Correct `standalone test` environment variable

fisker:
refs:
- [old azure pipeline settings](https://github.com/prettier/prettier/commit/34115cbd3c7101a6be733e8d3288fe46d0eb10f5#diff-fec826feae04e51c0d94076385408bdcL90)
- [mistake](https://github.com/prettier/prettier/pull/7431/files#diff-2d1d1197987e7769ed317c9ef37ea11dR110)

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "prepare-release": "yarn && yarn build && yarn test:dist",
     "test": "jest",
     "test:dist": "cross-env NODE_ENV=production jest",
-    "test:dist-standalone": "cross-env NODE_ENV=production STANDALONE=1 jest tests/",
+    "test:dist-standalone": "cross-env NODE_ENV=production TEST_STANDALONE=1 jest tests/",
     "test:integration": "jest tests_integration",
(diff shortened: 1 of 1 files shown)


fisker:
How do I know it's correct or not? I guess have to check locally.

thorn0:
> How do I know it's correct or not? I guess have to check locally.

One way is to add a test that standalone doesn't have methods like `getFileInfo` and `clearConfigCache`.

fisker:
Problem is  I don't even know if it's running standalone test.

thorn0:
Right. So we can add a test that would check  `prettier.getFileInfo` should be `undefined` if and only if this env variable is set.

fisker:
But if I set the variable name  wrong again, jest will run the default tests, still get pass.

thorn0:
The default test should assert the opposite: `prettier.getFileInfo` should be defined.

thorn0:
Also we could add some kind of report to which each test run would add something (only on CI) and in the end a separate step will check that this report is correct.

PR #8088, opened 2020-04-20
Run tests on snippets instead of files

fisker:
Some codes can't test pieces together, for example, we can't test two json string in one file.

```js
run_spec(
  {
    dirname: __dirname,
    codes: [
      {
        name: "JSON object",
        code: '{"foo": 1}',
      },
      {
        name: "JSON array",
        code: '["foo", "bar"]',
      },
    ],
  },
  ["json"]
);
```

Some codes testing special characters, we can use js to create those files.

```js
run_spec(
  {
    dirname: __dirname,
    codes: [
      {
        name: "empty with BOM",
        code: "\uFEFF",
      },
      {
        name: "BOM",
        code: `\uFEFF
          const foo = bar
        `,
      },
      {
        name: "`CRLF` as linebreak",
        code: "const foo=1;\r\nconst bar=2;",
      },
    ],
  },
  ["babel"]
);
```

We don't need create small files anymore, or worry about editor replacing those special characters.

`name` and `filename` both optional

```js
run_spec(
  {
    dirname: __dirname,
    codes: [
      // just code
      'code snippet 1',
      'code snippet 2',
    ]
  },
  ["babel"]
)
```


tests/js_empty/jsfmt.spec.js
```diff
@@ -1 +1,18 @@
...
+      // empty
+      "",
+      // empty lines
+      "\n",
+      "\n\n\n\n",
+      // semicolons
+      ";",
+      ";;;;",
+      ";\n",
+      ";\n\n;;;\n",
```
thorn0, line 14: This looks fine.

tests/parser-inference/jsfmt.spec.js
```diff
@@ -1 +1,48 @@
...
+  dirname: __dirname,
+  codes: [
+    {
+      name: "html",
+      filename: "prettier.html",
+      code: dedent`
+        <foo>
+                        <bar/>
+        </foo>
+      `,
```
thorn0, line 13: But this looks worse than files IMHO. There is nothing wrong with small files. They should be small.

thorn0:
Also let's use `code` or `snippets` instead of `codes`. The word code is uncountable when it means "source code".

tests/parser-inference/jsfmt.spec.js
fisker: I restored them.

fisker:
Sorry for my poor english, 

```js
{
  code: [
    {
      code: ''
    }
  ]
}
```

seems weird, so `snippets`?


[Eslint called their tests `scenarios`](https://github.com/eslint/eslint/blob/b398e0b9ff455e4e7d70f19d5ccea5819c21eb86/lib/rule-tester/rule-tester.js#L436) , I don't know if it's a good name?

thorn0:
`snippets` is fine.

fisker:
@thorn0 I checked the `index.js` file in `dist`,  on L24254

```js
let {
  resolve: resolve$1
} = eval("require");
```

I remember this can't run in atom?

tests_config/run_spec.js
```diff
@@ -67,32 +70,50 @@ global.run_spec = (dirname, parsers, options) => {
...
-      !file.isFile() ||
-      basename[0] === "." ||
-      basename === "jsfmt.spec.js"
-    ) {
-      continue;
-    }
+  const snippets = (fixtures.snippets || []).map((test, index) => {
+    test = typeof test === "string" ? { code: test } : test;
+    return {
+      name: `code: ${test.name || `#${index}`}`,
```
thorn0, line 76: Would be logical:
```suggestion
      name: `snippet: ${test.name || `#${index}`}`,
```

fisker:
I need add this to https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md

PR #8105, opened 2020-04-21
Treat `.prettierrc` as yaml

### LINKED ISSUE #6040: .prettierrc parsed as js/json by prettier
**Environments:**
- Prettier Version: 1.16.4
- Running Prettier via: cli
- Runtime: node v11.11.0
- Operating System: macOS

**Steps to reproduce:**

- `echo 'printWidth: 100' > .prettierrc`
- `prettier .prettierrc`

**Expected behavior:** either parses as yaml, or ignores as not parseable

**Actual behavior:** error:

```
.prettierrc[error] .prettierrc: SyntaxError: Unexpected token (1:11)
[error] > 1 | printWidth: 100
[error]     |           ^
[error]   2 |
```

(hitting this because I'm doing `ls | xargs prettier --write` basically)

fisker:
Fixes #6040


thorn0:
`.prettierrc` can be either JSON or YAML. https://prettier.io/docs/en/configuration.html

fisker:
I know, but yaml parser can parse json.

thorn0:
Oh, right.

tests/parser-inference/jsfmt.spec.js
```diff
@@ -1 +1,41 @@
...
+      name: ".prettierrc in json(empty)",
+      filename: ".prettierrc",
+      code: "{}",
+    },
+    {
+      name: ".prettierrc in json",
+      filename: ".prettierrc",
+      code: dedent`
+        {printWidth: 100,
+        overrides: [
```
thorn0, line 16: This is not JSON.

fisker: I'll add another one.

thorn0: You're aware of the trailing comma issue, right?

fisker: I try to add `json5` comment `// comment` , seems yaml parser can't parse, so I delete that test.

trailing comma should not be a problem, it can be parsed by yaml.

fisker: You are right, the output not right.

thorn0: Prettier doesn't support comments in `.prettierrc` if it's written in JSON.

fisker: Trailing comma can be recognized by `cosmiconfig`, 

```console
$ cat .prettierrc
{
  "parser": "babel",
  "quoteProps": "as-needed",
}
$ node -p "require('.').resolveConfig.sync()"
{ parser: 'babel', quoteProps: 'as-needed' }
```

fisker:
What do you think the idea 

> Consider `.prettierrc` in json is "embed json in yaml"

:smile:

thorn0 (approved):

thorn0:
Just realized it won't work with `--embedded-language-formatting off`.

fisker:
Yes, I mean to make it possible to disable it as JSON by this option, normal ".yaml" file in "JSON" format like the one before I change it, people might want that.

thorn0:
Kind of makes sense. Okay, let's keep it this way.

fisker:
It's awkward we are doing this https://github.com/prettier/prettier/blob/00a4798b9042bffcf89d341e60aa6cddc3a14102/.prettierrc#L5-L7, just because I don't want write it in JSON.

PR #8119, opened 2020-04-22
Build: exclude `index.js` from `bin-prettier`

fisker:
This `index.js` we already bundled to `dist/index.js`, we don't need include it again.

This also easier to make `please-upgrade-node` work on old nodes, if we want.

`bin-prettier.js`

```diff
- Before: 1536KB
+ After:   416KB
```


src/index.js
```diff
@@ -66,6 +66,12 @@ module.exports = {
 
   util: sharedUtil,
 
+  // Internal shared
+  /* istanbul ignore next */
+  __shared: {
+    errors: require("./common/errors"),
+  },
```
fisker, line 73: Is it okay to share module like this? We can share more modules from `index.js`, `bin-prettier.js` can get even smaller.

fisker:
Any one with me on this?

@thorn0 @evilebottnawi

src/index.js
thorn0: It's necessary to do something like that. Otherwise there will be multiple instances of the same modules, which might lead to difficult-to-catch errors.

thorn0:
Would be good to understand why it wasn't done this way from the beginning. Just technical debt or were there reasons?

fisker:
Actually , the `bin-prettier.js`/`index.js`/`standalone.js` all have duplicated `core`. 

I'm trying to reuse some code in `bin-prettier.js` from `index.js` in PR, seems no problem.

Ideally, we should share `core` from `standalone.js` -> `index.js` -> `bin-prettier.js`, and `cli/*` should not allow to import any file from other dir, they should come from `core`.

If all these get done, our package size will reduce about 2M.

thorn0:
To be safe (to avoid having multiple instances of the same modules at run time), we'll have to do the same thing you did with `common/errors` with the rest of the `require`s from other dirs.

> and cli/* should not allow to import any file from other dir

Exactly. Eventually, those things exported by `index.js` should become public API, so that it'd be possible to create alternative CLIs using them.

I'm skeptical about reusing `standalone` in `index` though. It might make the code needlessly convoluted.

src/cli/util.js
```diff
@@ -12,19 +12,20 @@ const fromPairs = require("lodash/fromPairs");
...
-const createIgnorer = require("../common/create-ignorer");
 const { expandPatterns, fixWindowsSlashes } = require("./expand-patterns");
-const errors = require("../common/errors");
 const constant = require("./constant");
-const coreOptions = require("../main/core-options");
-const optionsModule = require("../main/options");
-const optionsNormalizer = require("../main/options-normalizer");
-const thirdParty = require("../common/third-party");
-const arrayify = require("../utils/arrayify");
-const isTTY = require("../utils/is-tty");
```
thorn0, line 27: Can we use https://eslint.org/docs/rules/no-restricted-modules to enforce these import rules?

fisker: Yes, I was going to add. eslint has a good example https://github.com/eslint/eslint/blob/master/.eslintrc.js

fisker: But we need decide this first,

How do we import `fast-glob` `lodash/flatten` `json-stable-stringify` `lodash/fromPairs` etc.

For example `lodash/fromPairs`, `index.js` might not need it, but it could be used someday in `index.js`, so all come from `index.js`?

thorn0: 3rd party libraries are extremely unlikely to be stateful, so it's safe to keep them as is. I support a stricter separation of the CLI from the rest of the code, but I don't think the size of the package is important.

thorn0: On the other hand, isn't that what `third-party.js` is for?

fisker: `third-party.js` only keep modules need mock during test.

fisker:
Rules added, `import/no-relative-parent-imports` don't support `require`, 
`node/no-restricted-require` need absolute path.

`no-restricted-modules` + `eslint-disable` should be good enough.

scripts/build/build.js
```diff
@@ -21,6 +21,7 @@ process.on("unhandledRejection", (err) => {
   process.exit(1);
 });
 
+const CACHE_VERSION = "v22"; // This need update when updating build scripts
```
fisker, line 24: I always have trouble finding this, I know it's somewhere just can't remember, hope moving to top can help.

thorn0:
`isTTY` is used only in `cli`. Please move it to `cli`.

.eslintrc.yml
```diff
@@ -76,3 +76,10 @@ overrides:
       strict: off
     globals:
       run_spec: false
+  - files:
+      - src/cli/**/*.js
+    rules:
+      no-restricted-modules:
```
lipis, line 82: I've noticed that you added bunch of disables for this rule in the code, so the question is how important it is to have it here?! I would suggest to remove this rule.

thorn0: The idea is to disallow importing anything in `cli` from other directories except for two modules: `../index` and `../common/third-party`. But even though the rule supports negative patterns for specifying such exceptions (e.g. `"!baz-module/good"`), internally it uses the [`ignore`](https://www.npmjs.com/package/ignore) module, which doesn't play well with relative paths that contain `..`. The reason is that that module is an implementation of the `.gitignore` rules where such paths can't be used. So the negative pattern `!../common/third-party` doesn't work. `!../index` seems to work though.

BTW, Prettier uses the same `ignore` module and exhibits similar bugs if the file specified by the `--ignore-path` CLI option isn't in the root of the project. We need to open an issue about that.

PR #8138, opened 2020-04-23
docs: add two plugins `pkg` and `sh`

JounQin:
`prettier-plugin-pkg` for `package.json`
`prettier-plugin-sh` for shell scripts


fisker:
I think  we  should add some  descriptors to  these plugins.

@thorn0 @evilebottnawi

thorn0:
@fisker Agree. This list probably needs to be turned into a table.

kachkaev:
Should the plugins be inserted so that the list remained alphabetical?

lipis:
Yes I agree that we have to turn this thing into a table.. but in a separate PR.. 

Some suggestions for the columns:

- Name
- Description
- AST (if it's changing the AST or not)
- Additional options
- (Anything else?)

PR #8139, opened 2020-04-23
remark-parse: v8.0.2

saramarcondes:
`remark-parse`'s version has drifted 3 major releases from the current major release. There are a lot of bug fixes and new features.

All I've done in this PR is bump the version, replaced `loose` with `spread` (even though [they're not equivalent](https://github.com/remarkjs/remark/pull/364)) and updated the snapshot tests.

Looking for feedback for which changes to the markdown output are acceptable and which aren't.


saramarcondes:
So there are some changes that are clearly not acceptable:
1. footnotes
2. arguments passed to code blocks

Of course there are many changes to the ways lists print due to the loose/spread distinction. I'd love feedback on those specifically.

PR #8142, opened 2020-04-24
test: Allow assert output

fisker:
Some whitespace hard to see in snapshots, easier to assert output directly.

I understand this will make the test hard to update when format logic changes, but still worth.


tests/jsx-whitespace/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,63 +0,0 @@
...

[</div>;
-slash_v = <div>]


[</div>;
-non_breaking_spaces = <div>]   [</div>;
-em_space = <div>]   [</div>;
-hair_space = <div>]  [</div>;
-zero_width_space = <div>]​​​[</div>;
```
fisker, line 53: These spaces really hard to know what's in there.

package.json
```diff
@@ -105,6 +105,7 @@
     "jest-snapshot-serializer-ansi": "1.0.0",
     "jest-snapshot-serializer-raw": "1.1.0",
     "jest-watch-typeahead": "0.5.0",
+    "outdent": "0.7.1",
```
fisker, line 108: See #8143

PR #8164, opened 2020-04-27
Build: remove `doc.js`

fisker:
Why are we building this file? It's not used anywhere.

Should use `const {doc} = require("prettier")` instead of `const doc = require("prettier/doc")`

### DIFF
--- a/scripts/build/config.js
+++ b/scripts/build/config.js
@@ -142,13 +142,6 @@ const coreBundles = [
     },
   },
   {
-    input: "src/document/index.js",
-    name: "doc",
-    type: "core",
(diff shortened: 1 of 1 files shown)


fisker:
After a little digging, it was add on purpose...

https://github.com/prettier/prettier/pull/6009#issuecomment-476282558

fisker:
@rattrayalex Are you using it?

evilebottnawi:
@fisker some plugins also can you it, for me it is breaking change

fisker:
Seems plugins actually using this, close for now.

PR #8165, opened 2020-04-27
Remove use of `assert`

fisker:
Ref: https://github.com/prettier/prettier/pull/4449/files#r190597247

### DIFF
--- a/src/language-html/printer-html.js
+++ b/src/language-html/printer-html.js
@@ -43,7 +43,6 @@ const {
 } = require("./utils");
 const { replaceEndOfLineWith } = require("../common/util");
 const preprocess = require("./preprocess");
-const assert = require("assert");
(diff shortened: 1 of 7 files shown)


src/language-html/printer-html.js
```diff
@@ -884,7 +883,10 @@ function printOpeningTagStartMarker(node) {
 }
 
 function printOpeningTagEndMarker(node) {
-  assert(!node.isSelfClosing);
+  /* istanbul ignore next */
+  if (node.isSelfClosing) {
+    throw new Error("node.isSelfClosing should be false");
+  }
```
thorn0, line 889: :-1: 4 lines instead of 1 and what for? Assertions are an idiomatic way to do the same thing this `if` does. I see no reason to remove them. If they're considered development-only, we should configure the build system to remove them.

PR #8173, opened 2020-04-28
Support HTML-like Comments

fisker:
Support [`HTML-like Comments`](https://tc39.es/ecma262/#sec-html-like-comments)


src/language-js/parser-babel.js
```diff
@@ -7,7 +7,7 @@ const postprocess = require("./postprocess");
 
 function babelOptions(extraPlugins = []) {
   return {
-    sourceType: "module",
+    sourceType: "unambiguous",
```
fisker, line 10: Oldest line of code, since first commit https://github.com/vjeux/prettier/commit/29627ee520307494c45073701b20ef64effd3a8f#diff-86100135effef64b185e301e9346e528R13

src/language-js/parser-babel.js
```diff
@@ -5,9 +5,9 @@ const { hasPragma } = require("./pragma");
 const locFns = require("./loc");
 const postprocess = require("./postprocess");
 
-function babelOptions(extraPlugins = []) {
+function babelOptions({ sourceType, extraPlugins = [] }) {
   return {
-    sourceType: "module",
```
fisker, line 10: Oldest line of code, since first commit https://github.com/vjeux/prettier/commit/29627ee520307494c45073701b20ef64effd3a8f#diff-86100135effef64b185e301e9346e528R13

src/language-js/parser-babel.js
```diff
@@ -65,6 +65,10 @@ function createParse(parseMethod, ...pluginCombinations) {
     // Inline the require to avoid loading all the JS if we don't use it
     const babel = require("@babel/parser");
 
+    // HTML-like Comments https://tc39.es/ecma262/#sec-html-like-comments
+    // requires sourceType to be `script` https://github.com/babel/babel/issues/7802#issuecomment-492522145
+    const sourceType = /<!--|-->/.test(text) ? "unambiguous" : "module";
```
bakkot, line 70: This regular expression doesn't do the thing you want; e.g. it would match

```js
let x=10;while(x-->0)console.log(x)
await(import('mod'))
```
which does not actually have contain HTML-like comment and which has a substantially different AST when interpreted as a module vs as a script.

fisker: This is a  real good case,  HTML-like comment and dynamic import , I'll add to tests.

This also why I set it to `unambiguous` instead of `script`, the output seems same as stable version.

[stable](https://prettier.io/playground/?N4Igxg9gdgLgprEAuEBDJsAWBaANqgZxmzEwEtcATACgCYoACAagYEYBKAHUYYwhmrcGwhgDoCEALZwS+AgSEjxUmf0xwATrMIKew5dOwA3TQE88ES1es2rUAObb5i4aly5FXKC3RY8hYlIKGnpmNi9hPgEXMQlDMDldEViVbDVNJySlOJkTDXNcWyKbB0yYtw8eCN4kACM4ADMIDThgAF8Gbi6oUXQ4AA94KBpRMigABwBXGCRKMgJUWtw4SnZQABoQCHGYMmgCZFBUDQ0IAHcABWOEA5Q3M9RTA83ajVQwAGs4GABlcfexvZkDANJM4JtMDBJLgAOrkeAEf5gOA-G5kXZGdGmZDgeQgTZjAiaGAXN72SSoZANNxEzYAKwI-QAQm9Pt8fqhpAAZMZwKk08EgBn9H6A5YARUm-D5SGpuFpIH+GiJGhxtUWcFw+MVGjGMBhZEoMEwyAAHAAGTbjU5EmFvcY461wFUmbUARyl8FJ21uaAI2CgcBWK21LQ9ZBapNQ5MpsoFmyJkjIwNBgoIYrgkul-PlgpgiwNRpNSFomxBqAogIAwlIKTjnQBWbWTIkAFUWtzlCqMYIAksMEL8wLqdgBBYY-GCmZY5oltNpAA#N4Igxg9gdgLgprEAuEAbOMAEAPAvARgAYBuAdwAsBLdACmwFp6A+QgSkigGcJ0A6VCAHM6rADpQAhqQmUYNSgFsADhABOcgOQKIAEw2tWIADQgISmJWidkoCatURSABTsJrKCamkBPayYBGqhJgANYYAMpKwZRQgsgwqgCucCbkMAqoAOpU8JxRYHDhbrKUAG6y3sjgnH4gMZxw6k5BggoSyABmng0mAFac2ABCQaEREgpwADIxcJ3dKSD92OExgugAiokQ8HOoPSBRqg2qVf4S-nCoxgeqMTCZlDow5MgAHIQmSg4NmUFKVV84MdSrMTABHLbwZpmdwgCScehQOBwHQo66qOAQygY5oSVrtJBdPYLBoKSjxJIk1YbSGzQnzEwwc4PJ4vJAAJkZQWoqwAwhAFG0qkCAKzXRINAAq53cRP2pWSAEkoKjYOEwLdzABBFXhGDedC7BoAX2NQA) vs [pr](https://prettier.io/playground/?N4Igxg9gdgLgprEAuEBDJsAWBaANqgZxmzEwEtcATACgCYoACAagYEYBKAHUYYwhmrcGwhgDoCEALZwS+AgSEjxUmf0xwATrMIKew5dOwA3TQE88ES1es2rUAObb5i4aly5FXKC3RY8hYlIKGnpmNi9hPgEXMQlDMDldEViVbDVNJySlOJkTDXNcWyKbB0yYtw8eCN4kACM4ADMIDThgAF8Gbi6oUXQ4AA94KBpRMigABwBXGCRKMgJUWtw4SnZQABoQCHGYMmgCZFBUDQ0IAHcABWOEA5Q3M9RTA83ajVQwAGs4GABlcfexvZkDANJM4JtMDBJLgAOrkeAEf5gOA-G5kXZGdGmZDgeQgTZjAiaGAXN72SSoZANNxEzYAKwI-QAQm9Pt8fqhpAAZMZwKk08EgBn9H6A5YARUm-D5SGpuFpIH+GiJGhxtUWcFw+MVGjGMBhZEoMEwyAAHAAGTbjU5EmFvcY461wFUmbUARyl8FJ21uaAI2CgcBWK21LQ9ZBapNQ5MpsoFmyJkjIwNBgoIYrgkul-PlgpgiwNRpNSFomxBqAogIAwlIKTjnQBWbWTIkAFUWtzlCqMYIAksMEL8wLqdgBBYY-GCmZY5oltNpAA#N4Igxg9gdgLgprEAuEAbOMAEAPAvARgAYBuAdwAsBLdACmwFp6A+QgSkigGcJ0A6VCAHM6rADpQAhqQmUYNSgFsADhABOcgOQKIAEw2tWIADQgISmJWidkoCatURSABTsJrKCamkBPayYBGqhJgANYYAMpKwZRQgsgwqgCucCbkMAqoAOpU8JxRYHDhbrKUAG6y3sjgnH4gMZxw6k5BggoSyABmng0mAFac2ABCQaEREgpwADIxcJ3dKSD92OExgugAiokQ8HOoPSBRqg2qVf4S-nCoxgeqMTCZlDow5MgAHIQmSg4NmUFKVV84MdSrMTABHLbwZpmdwgCScehQOBwHQo66qOAQygY5oSVrtJBdPYLBoKSjxJIk1YbSGzQnzEwwc4PJ4vJAAJkZQWoqwAwhAFG0qkCAKzXRINAAq53cRP2pWSAEkoKjYOEwLdzABBFXhGDedC7BoAX2NQA)

bakkot: That's surprising, because when parsed as a script this should be a function call, not an actual `await`.

fisker: yes, I  know that , but your case should be a  seperate issue.

fisker: By the way, do you know any other syntax requires `script`?

bakkot: The only cases where there's something which is valid in both a script and a module but with a different interpretation are HTML comments and `await`. There's a few things which are legal in scripts but not modules, like octal literals (`x = 01` is legal as a script but not a module), and of course import/export are legal in modules but not scripts.

The only interesting case with HTML comments when you have `<!--` in a module, which doesn't really come up. e.g.

```js
let x = 2;
if (-1<!--x) console.log(1); if(0
==-1) console.log(2);
```
prints 1 in a script and 2 in a module - but who writes that?

lydell:
Just a question: What is your motivation for adding this? Like, ”it’s in the spec so we should support it, and it’s fun”, or because you or someone else have had a need for this? If so, what was the use case? Is this worth the trouble?

fisker:
1. It's spec
2. Makes prettier easier to run in old  html project (I  do have some old  static html website, one site  contains html generated by  program with  template, but I lost the database and program)

thorn0:
How about enabling it only in `<script>` tags (check `options.parentParser`)?

lydell:
> How about enabling it only in `<script>` tags (check `options.parentParser`)?

Interesting idea! The good thing about that is that we can look for `type="module"` to know if we should parse the code as module or script – then Babel should let us parse the HTML-style comments for us without hacks.

fisker:
> enabling it only in <script> tags

It is a good idea.

> look for type="module" to know if we should parse the code as module or script

You mean if not `type="module", use  `unambiguous` (not `script`), right ? We  can't use `script` because some bundler don't require type to be `module` but  allow use  `import`.

lydell:
I might be wrong, but as far as I know browsers _always_ parse `<script>` tags as _script,_ unless `type="module"` is present (then they are parsed as _module_).

fisker:
@lydell Yes, the browser do, but before your html publish, there might some tool bundle your html, and allow you use `import` without `type="module"`, let me try to fix a example.

fisker:
I thought parcel should support it, but it's not

```text
parcel index.html
Server running at http://localhost:1234
| Building index.html...Browserslist: caniuse-lite is outdated. Please run next command `npm update`
×  <CWD>index.html:1:7: Imports and requires are not supported inside inline <script> tags yet.
> 1 | import "./index.mjs"
```

Maybe we can do this?

src/language-js/parser-babel.js
```diff
@@ -65,6 +65,9 @@ function createParse(parseMethod, ...pluginCombinations) {
     // Inline the require to avoid loading all the JS if we don't use it
     const babel = require("@babel/parser");
 
+    const sourceType =
+      opts && opts.__babelSourceType === "script" ? "script" : "module";
```
thorn0, line 69: No need for `&&`. `opts` is always defined here.

fisker: No... I forgot how. Maybe called by another parser

thorn0: Okay, if you say so.

PR #8185, opened 2020-04-29
Fix unstable format with `no-semi`

### LINKED ISSUE #1510: Need to run prettier twice before it settles
There seem to be cases in which a single prettier run is not sufficient. A second run will change the already formatted file again. After the second run no more formatting will happen.

A small example I found:

```
a;

[].b();
```
The [first formatting](https://prettier.github.io/prettier/#%7B%22content%22%3A%22a%3B%5Cn%5Cn%5B%5D.b()%3B%22%2C%22options%22%3A%7B%22printWidth%22%3A50%2C%22tabWidth%22%3A1%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22all%22%2C%22bracketSpacing%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Afalse%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%7D%7D) result is:
```
a

;[].b()
```
The [second formatting](https://prettier.github.io/prettier/#%7B%22content%22%3A%22a%5Cn%5Cn%3B%5B%5D.b()%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A50%2C%22tabWidth%22%3A1%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22all%22%2C%22bracketSpacing%22%3Afalse%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Afalse%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%7D%7D) result (on the result of the first formatting) is:
```
a
;[].b()
```
Note the **removed** line.

I assumed 🤔  that prettier formatting is idempotent. Am I wrong 🙇  or is this a bug?

The only important option seems to be `semi: false`.

fisker:
Fixes #1510


fisker:
@thorn0 I need some help here, I was trying to fix #1510 (This one is really old!!), 

```js
a

;[].b()
```

I found the reason why the line after `a` is missing. It's because the `;` is count in range of `ExpressionStatement`, so there is no empty line after this `ExpressionStatement`, I tried to change `ExpressionStatement` end to skip the `;` and space, It will format as expected. Though there is some other problem need solve.

Question: 

1. Is this the right way?
2. If we want fix the range, should we 
	1. skip only `;`
	2. skip `space` + `;`
	3. fix to `ExpressionStatement.expression.end`


I think we should ignore `;`, and I also found if we set skip the comments, we will fix https://github.com/prettier/prettier/blob/master/tests_config/run_spec.js#L26
https://github.com/prettier/prettier/blob/master/tests_config/run_spec.js#L27
https://github.com/prettier/prettier/blob/master/tests_config/run_spec.js#L39

too.

thorn0:
You can read how I was going to fix it here: https://github.com/prettier/prettier/pull/7161#issuecomment-571300833

The idea is that if a statement (not necessary an expression statement, it happens to all kinds of statements) ends with a semicolon, we find the token preceding the semicolon (the penultimate token of the statement) and assign `end` of that token to `end` of the statement.

fisker:
Close due to age.

PR #8240, opened 2020-05-07
Switch `dedent` to `outdent`

### LINKED ISSUE #8143: Dependency `dedent` has bug
`dedent` is in `package.json`, I just found it's buggy, I'll check how it effect our code.

```console
$ node -p "`\u0070\u0072\u0065\u0074\u0074\u0069\u0065\u0072`"
prettier

$ node -p "require('dedent')`\u0070\u0072\u0065\u0074\u0074\u0069\u0065\u0072`"
\u0070\u0072\u0065\u0074\u0074\u0069\u0065\u0072

$ node -p "require('outdent').outdent`\u0070\u0072\u0065\u0074\u0074\u0069\u0065\u0072`"
prettier
```

fisker:
`dedent` is working fine in other places, but can't use in `tests/`, it can't work with escaped characters, more details see #8143.

But having two similar package may cause confusion.

Switching to `outdent`

Fixes #8143


scripts/build/bundler.js
```diff
@@ -37,6 +39,33 @@ const EXTERNALS = [
   "graceful-fs",
 ];
 
+const entries = [
```
thorn0, line 42: What is this change doing in this PR?

fisker: The entries of `outdent` need point to commonjs build,  and I tried fix in 4 bundles, still not enough, seems it's
 imported everywhere, so I move them to commom alias.

scripts/build/bundler.js
```diff
@@ -118,6 +147,9 @@ function getRollupConfig(bundle) {
 
   const babelConfig = getBabelConfig(bundle);
 
+  const alias = { ...bundle.alias };
+  alias.entries = [...entries, ...(alias.entries || [])];
```
lydell, line 151: Does this work?

```suggestion
  const alias = { ...bundle.alias, entries: [...entries, ...(bundle.entries || [])] };
```

fisker: Yes, I think so.

fisker: Wait, no... bundle.alias can be undefined.

thorn0:
Would be cool to replace this with a build-time transformation (Babel plugin) one day. :)

PR #8245, opened 2020-05-07
Test: Remove `graceful-fs` hack

fisker:


thorn0:
`graceful-fs` is also mentioned twice in `scripts/build/bundler.js`. Should we do something about that?

fisker:
I didn't see the other one, the comment in `jest.config.js` means the one I removed, I checked the history.  But I'll take a look.

fisker:
I download the artifacts, compared 
- the dist from this PR, 
- the dist from master branch, 
- local build

They are all the same, except the issue I found #8268

PR #8248, opened 2020-05-07
Categorize tests part 2

fisker:


thorn0:
I'm good with these changes as long as it's possible for Git to trace the history of the files. Please don't modify them too much when moving.

fisker:
I won't

PR #8251, opened 2020-05-07
Categorize tests part 4(CSS, flow, misc)

fisker:


fisker:
@thorn0  I've done. You may want make some adjustments?

thorn0:
I thought we had tests for union and intersection types shared between Flow and TS, but turns out we didn't. If we add such tests, what directory should we put them in?

fisker:
Maybe `js`, unlike other  languages,  I moved original `typescript-*` in `typescript`, kind of `typescript-only` in `typescript`, `flow-only` in `flow` , for example js don't has `interface`, I don't know where to put it, I moved to `js`. Do you have good idea?

thorn0:
I wish I had. All possible solutions seem to be non-obvious for contributors.

fisker:
Let's think this way, image we don't have `inline-comment` tests, you want test it in `scss` now, you should put it in `scss/`, and I'll say, please add tests in `less/` too, or please verify with `less` parser too. Someday, `css` support this too, we can add separate tests in `css/`, or add `css` parser to tests in `scss`.

So `union and intersection types`, you can 

- put in `typescript/` and verify with `flow`/`babel-flow`
- put in `flow/` and verify with `typescript`/`babel-ts`
- put in duplicated tests both `typescript/` and `flow/`

I think better in this way.

--- 

To enforce this way, I'll update this check https://github.com/prettier/prettier/blob/master/tests_config/utils/check-parsers.js#L15

To something like:

|category|allowed parser|allowed verify parser|
|:-:|:--|:--|
|typescript|`["typescript", "babel-ts"]`|`["babel", "flow", "babel-flow"]`|
|css|`["css"]`|`["scss", "less"]`|
|scss|`["scss"]`|`["css", "less"]`|
|less|`["less"]`|`["css", "scss"]`|

WDYT?

PR #8263, opened 2020-05-08
[DO NOT MERGE] Add async version babel parser

fisker:
Trying to investigate possibility of making the parser async.


fisker:
I think the effect is quite small, if we add a new version of `format` method, should easy to implement


## API

- core.format
- formatRange
- parser.parse

## CLI

- run
- formatFiles

fisker:
We already support async parser in next branch.

PR #8266, opened 2020-05-08
refactor(css): calculate correct location data for value-* nodes

thorn0:
Once all the nodes have the correct location data (`selector-*` and `media-*` are TBD), it'll be possible to use Prettier's comment attachment logic for CSS like discussed in #3905. Here's the plan:

- Have correct location data in the AST.
- Preparse the input with a simple parser (similar to what we already have in `language-css/loc.js`, see `replaceQuotesInInlineComments` there) and collect comments.
- Remove the comments from the input, so that the `postcss-*` parsers don't have to deal with them.
- Use the comment attachment logic like it's done for JS.

A big problem here is that it's not always possible to remove a comment safely (e.g. [`div/**/span`](https://mobile.twitter.com/thorn0/status/1257251001448398854)), so some of them will have to stay and be handled the way they're handled now. But at least all SCSS and LESS single line comments are safe to remove.

### DIFF
--- a/src/language-css/loc.js
+++ b/src/language-css/loc.js
@@ -5,6 +5,9 @@ const { getLast, skipEverythingButNewLine } = require("../common/util");
 
 function calculateLocStart(node, text) {
   if (node.source) {
+    if (typeof node.source.sourceIndex === "number") {
+      return node.source.sourceIndex; // value-* nodes have this
+    }
     return lineColumnToIndex(node.source.start, text) - 1;
   }
   return null;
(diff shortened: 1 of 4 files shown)


src/language-css/loc.js
```diff
@@ -25,26 +28,68 @@ function calculateLocEnd(node, text) {
...
-  if (node && typeof node === "object") {
-    if (node.source) {
-      node.source.startOffset = calculateLocStart(node, text);
-      node.source.endOffset = calculateLocEnd(node, text);
+  if (node.source) {
+    node.source.startOffset = calculateLocStart(node, text);
+    node.source.endOffset = calculateLocEnd(node, text);
+  }
+
+  for (const key in node) {
```
fisker, line 36:
```suggestion
  for (const [key, child] of Object.entries(node)) {
```

src/language-css/loc.js
```diff
@@ -25,26 +28,68 @@ function calculateLocEnd(node, text) {
...
+    const child = node[key];
```
fisker, line 37:
```suggestion
  for (const [key, child] of Object.entries(node)) {
```

thorn0: I'm not sure about the performance impact of this change.

fisker: `for...in` is faster.

PR #8274, opened 2020-05-10
Add arm-template to vscodeLanguageIds of JSON and JSONC

bergmeister:
Related: https://github.com/microsoft/vscode-azurearmtools/issues/465
Second attempt after https://github.com/prettier/prettier-vscode/pull/1335

cc @ntotten 

### DIFF
--- a/src/language-js/index.js
+++ b/src/language-js/index.js
@@ -47,7 +47,7 @@ const languages = [
   createLanguage(require("linguist-languages/data/JSON"), (data) => ({
     since: "1.5.0",
     parsers: ["json"],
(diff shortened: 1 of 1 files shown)


thorn0:
It should be added only to one language. To `jsonc` I guess. Also you need to run `yarn test -u` to update snapshots.

However, if this language has syntax incompatible with JSON, it should be implemented as a Prettier plugin. If we merge this change now, and someone creates a plugin later, will the VS Code extension use that plugin or will it still associate this language with the `json` parser? I'm not sure.

fisker:
Close due to age.

Issue #8272, opened 2020-05-09
<select> and <optgroup> can safely be formatted nicer with htmlWhitespaceSensitivity=css, just like <datalist>


lydell:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAznANnMMB8qEADjAJbT4BCWArnKgPQnmVGkVT4DiATnAiYtOhFgHNeEWsQAEWAIYAjbAF4AOiAAi83gGs4vDaI6VtemTXpCTXdqy5ndMvgKjX7hZqQlTinzDh4+GpuZFDEtDByZOgwKpBYELzowW4AJvIw8lgxUWRp8RCJycYelgxeHnYiLoKVIkwZWTmxqSAANCDC0OjIoDqSAO4ACjoIvSjZg-IAnr2dirzyYPowAMrEy2FiyDC89J0AFjAAtlgA6odk8OibYHBr49dkAG7XM8jg6PMgYZi8MGGSzEJ3kyAAZtlMJ0AFboAAeVCWKzg63kJzgABkwnAIVC4LCEWttjgAIq0CDwPFYaEgTbJAyfY5nDp03hhGDnfIwQ7IAAcAAZOsRJJhzktiJ8RXB-i9cZ0AI4U+BAkgTEDydAAWigAjScDSrP4SrI-CB8hBYKQkJpBJAmBOZF2+zt6BJcHJlNx1vxnSyii5aR5yAATH6lmQclAxABhCAnUGfGUAVlZtEwABUlBMbbSXvQAJJQfWwNZgdmkACCxbWMBmOGpmAAvk2gA)
```sh
--parser html
```

**Input:**
```html
<select><option>Blue</option><option>Green</option><optgroup label="Darker"><option>Dark Blue</option><option>Dark Green</option></optgroup></select>
<input list=colors>
<datalist id=colors><option>Blue</option><option>Green</option></datalist>

```

**Output:**
```html
<select
  ><option>Blue</option
  ><option>Green</option
  ><optgroup label="Darker"
    ><option>Dark Blue</option><option>Dark Green</option></optgroup
  ></select
>
<input list="colors" />
<datalist id="colors">
  <option>Blue</option>
  <option>Green</option>
</datalist>

```

**Expected behavior:**
```html
<select>
  <option>Blue</option>
  <option>Green</option>
  <optgroup label="Darker">
    <option>Dark Blue</option>
    <option>Dark Green</option>
[...]

PR #8275, opened 2020-05-10
Treat <select> and <optgroup> as inline-block for whitespace sensitivity

thorn0:


fisker:
Is this #8272 ask for? 

I think @lydell mean `htmlWhitespaceSensitivity: "css" / "strict" / "ignore"` the same?

lydell:
In theory I think it’s safe to do the nicer formatting even in `htmlWhitespaceSensitivity: ignore`, but that would be new machinery that we don’t have yet.

thorn0:
I think with `strict`, users are supposed to add all the whitespace manually.

lydell:
Oops, I meant `strict` in my comment.

Issue #6921, opened 2019-11-11
Revisit function literal heuristic in Prettier 1.19


rassie:
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBnGAnOBDAtgRgBIA6ABwEtS4AKAHSgAIGB3cmACwBlt50AxDBFzV0WPACZCASgA09JgBNy6clDAwAqrHIAbAMLtsUAOZwF1WfIYK4AIwgBXNXAAq5XDQCsABin0-UCAyIBCkMOTQqMig2BiCzAAKsQhRKNgAbhDkCkEgthjYYADWcDAAyqSFqsbImA5wwewwuDoA6uxscKiVYHBlKWzk6WwAnsjgqFHBqqhwGDAJBca42MgAZtg6s8EAVqgAHgBCBcWlZXhwnKpw65vbIHv7ZdU6cACKDhDwt1sNIJUYWYYca2bC2OA6XKkQSzVoFUjjaFdObpG7BACOn3gi1CqRA2FQAFooHAzGZclhMeQsItsMtVkgNr9grNcORahh6iyXu8sTckHU-jAwa1shxkOJgphsLpqnohCtxlBoGiQA5Zi4wakmfd0vUAJJQGywMpgDCUGAAQSNZRgI1eP3u0NUMFFCnFSHw3m8AF8fUA)
```sh
--parser babel
--print-width 100
--single-quote
```

**Input:**
```jsx
stream1$.pipe(
  withLatestFrom(stream2$),
  distinctUntilChanged(),
  debounceTime(50)
)

```

**Output:**
```jsx
stream1$.pipe(withLatestFrom(stream2$), distinctUntilChanged(), debounceTime(50));

```

**Expected behavior:**
Function literal heuristic introduced in https://github.com/prettier/prettier/pull/6033 is insufficient for RxJS, since it ignores a lot of operators which at their core are functional compositions, but not in syntax. I don't have a solution proposal, but it's breaking existing (and expected) behavior right now.

thorn0:
The output looks perfectly readable to me.

rassie:
[...]

PR #8300, opened 2020-05-13
Functional composition experiments

thorn0:
Playing with different ideas to address #6921.

Or test it locally: **`npm install thorn0/prettier#functional-composition-experiments`**


fisker:
Run this on our code base, result https://github.com/fisker/prettier/pull/429/files

fisker:
Accept this?

evilebottnawi:
```
`${new Date()
    .toISOString()
    .replace(/T.+/, "")}-${version}.md`
```

Ugly

fisker:
> ```
> `${new Date()
>     .toISOString()
>     .replace(/T.+/, "")}-${version}.md`
> ```
> 
> Ugly

This one seems not related. It's changed by #7889

[#7889 Playground](https://deploy-preview-7889--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEADAJMKcDuACAEQEN4AKASgDoYIBJAZQHl6YAnASygHMLLW4ADgBsiYOKQD0AFUoBqCQBo8AHRCryAXwC0mAG5xWAZ3bQNlALYATVCAUgIAmCaiHkoIq1YQcABQ8JXFCIhHCIAT1c7ACNWUQBrOBh6AVFOLmQ2AFc4OwALGHMhAHVc9nhDFLF6ALL2XTKw5HBDSJBOQwMYH1iucyJkADNgjrsAK0MADwAhWLAEpKJzOAAZTjhB4ZyQcYn6NKE4AEVMiHgNoRGQFKMDJqiiKLghWyuOWCL2Sxhc5AAOAAY7AIvB0irEBE1gXAOqx9C8AI4neDdByBEBEQxabBwSw4l78RHsfjdIi9fpIIYXLYdczsDKsbJ2YzcA7HU7rCmbOwwB4fL4-JAAJm5sXYQjSAGEIOY+k1oQBWF6ZDpSB6BSmXXTZWhQXGwehgDiOACCupYYQO5w6Gg0QA)

brodybits:
Definitely looks like a positive change to me. Wouldn't this looks like a breaking change to some users?

thorn0:
> `${new Date()`
> This one seems not related. It's changed by #7889

Yes. I wonder what to do with it. Should we special-case `Date` to exempt it from the "always break constructor-call headed chains" rule or is it better to disable that rule inside template literals?

thorn0:
As for this PR, what do you all think the reformatting of the `mapDoc` calls? E.g. here:
https://github.com/fisker/prettier/pull/429/files#diff-b943e07d378ae28b03b5ebd46e7e86ac
Does that look okay? I'm not sure.

thorn0:
@brodybits 

> Wouldn't this looks like a breaking change to some users?

Where exactly do you think this change might look unexpected?
Do you have any specific libraries/usage in mind?

brodybits:
> Where exactly do you think this change might look unexpected?

If a massive set of changes would suddenly happen to a code base after a minor update of Prettier, like we would see in Prettier itself (<https://github.com/fisker/prettier/pull/429>), I would be a bit surprised. Or am I too pedantic?

thorn0:
@brodybits I think people are okay with massive changes when these massive changes are perceived as in improvement.

fisker:
> what do you all think about the reformatting of the mapDoc calls?

I think it's good.

karptonite:
> @brodybits I think people are okay with massive changes when these massive changes are perceived as in improvement.

Recall also that, to some extent, this sort of rolls back another "massive" change that appeared in 1.19.

rassie:
Any progress and/or resolution on this one? I've rebased this branch on master and tried running my codebase through it. It's somewhat better than the original, but there are still those weird single-line `pipe`s, which used to be multi-line.

thorn0:
@rassie This PR wasn't meant to change the situation for RxJS pipes as I don't know a solution to it.

src/language-js/print/call-arguments.js
```diff
@@ -253,21 +258,58 @@ function couldGroupArg(arg) {
...
-    !hasComment(lastArg, CommentCheckFlags.Trailing) &&
-    couldGroupArg(lastArg) &&
-    // If the last two arguments are of the same type,
-    // disable last element expansion.
-    (!penultimateArg || penultimateArg.type !== lastArg.type) &&
+
+  if (
+    hasComment(
+      lastArg,
+      CommentCheckFlags.Leading | CommentCheckFlags.Trailing
```
thorn0, line 267: @fisker This doesn't work as expected. Something is wrong inside `hasComment`.

thorn0: I was wrong. Actually, it makes sense. `CommentCheckFlags.Leading | CommentCheckFlags.Trailing` doesn't mean "leading or trailing". It means "leading and trailing".

fisker: Maybe we can add conflicting flags check? eg `Leading`/`Trailing`/`Dangling` and `Dangling`/`Line`

thorn0: Yes, as an assertion removed at build time. Good idea.

PR #8306, opened 2020-05-14
Support multiline shebang

### LINKED ISSUE #7562: Ignore semicolon in shebang
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBiAhAegEYEsqYDOAFgDpSnKUAEmmA3NXAB5xi0CuhATjvpggBu1GIQC0UCABM41MWJjcAhlEIAHXABs4Y6JoCe1SgBIADDRMABSuUiqYIhNQC81AIyn6IADQgIamFxoQmRQJW5uCAB3AAVwhBCUJUEIXCkfEGxlMABrOBgAZTUlMHwAc2RFDjhfYhgAW00AdWJceHUSuAKEttxBNv1kcEIQ33xCOG4YGOUy+qVkADMlTQnfACtCZgAhbLzCpXq4ABl8OCWVtZBN5gLy7QBFDgh4C9WakGLuCe4h7CVsHBNBk1Nx8DAmmkYMRkAAOUy+UEQCZNZRqIaguA-QTnXwAR2e8Bm-kSICU4igcDgMnSvm4cAJuHpMyUcwWSGW718E3quEq3Gq3PucCeL3OHMuHxgAMhUmhyAATL5FEotOUAMIQerzIaSSkZLhwAAqAMSnKugmqAEkoDJYAUwGCAgBBW0FGD6bRvCYAXx9QA)
```sh
--parser babel
```

**Input:**
```jsx
#!/bin/sh
":" //; exec /usr/bin/env ts-node --transpile-only "$0" "$@"
const ten = 10;
```

**Output:**
```jsx
#!/bin/sh
":"; //; exec /usr/bin/env ts-node --transpile-only "$0" "$@"
const ten = 10;

```

**Expected behavior:**
```jsx
#!/bin/sh
":" //; exec /usr/bin/env ts-node --transpile-only "$0" "$@"
const ten = 10;
```

Because env is not portable regarding options, A shebang like above is required. The problem is that when Prettier is configured to insert semicolons, the shell is not always happy.

Some shells will print "./example.ts: 2: ./example.ts: //: Permission denied" before executing the script.
There is no good way to add a prettier-ignore there either because not all shells accept "//" comments.

Because of the above reasons, prettier should not insert a semi colon in this case.

fisker:
Fixes #7562

### DIFF
--- a/changelog_unreleased/javascript/pr-8306.md
+++ b/changelog_unreleased/javascript/pr-8306.md
@@ -1,0 +1,19 @@
+#### Ignore semicolon in multiline shebang ([#8306](https://github.com/prettier/prettier/pull/8306) by [@fisker](https://github.com/fisker))
+
+<!-- prettier-ignore -->
+```js
+// Input
+#!/bin/sh
+":" //; exec /usr/bin/env ts-node --transpile-only "$0" "$@"
+const ten = 10;
+
+// Prettier stable
+#!/bin/sh
+":"; //; exec /usr/bin/env ts-node --transpile-only "$0" "$@"
+const ten = 10;
+
+// Prettier master
+#!/bin/sh
+":" //; exec /usr/bin/env ts-node --transpile-only "$0" "$@"
+const ten = 10;
+```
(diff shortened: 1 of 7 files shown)


fisker:
Can't make it work.

PR #8311, opened 2020-05-15
Build: `parser-flow.js` use strict mode

fisker:
This code is really old, let's see if it works without it.

### DIFF
--- a/scripts/build/config.js
+++ b/scripts/build/config.js
@@ -29,7 +29,6 @@ const parsers = [
   },
   {
     input: "src/language-js/parser-flow.js",
-    strict: false,
   },
   {
     input: "src/language-js/parser-typescript.js",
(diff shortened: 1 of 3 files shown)


scripts/build/config.js
```diff
@@ -28,7 +28,10 @@ const parsers = [
   },
   {
     input: "src/language-js/parser-flow.js",
-    strict: false,
+    replace: {
+      // `flow-parser` use this for `globalThis`, can't work in strictMode
+      "(function(){return this}())": '(new Function("return this")())',
```
evilebottnawi, line 33: Doesn't create CSP problems?

fisker: It could be, but we are using this for `parser-postcss.js` too, and nobody complains.

https://github.com/prettier/prettier/blob/master/scripts/build/bundler.js#L241

evilebottnawi: :+1:

PR #8331, opened 2020-05-17
Playground: autofold for Markdown and other languages

thorn0:


fisker:
I wonder can we highlight code like [astexplorer](https://astexplorer.net/) do

PR #8354, opened 2020-05-20
Fix: don't wrap arrays in `test.each` template strings

### LINKED ISSUE #7653: Don't wrap arrays in `test.each` template strings
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEACdr4GcYDo4CGYAFgAYA6UGGA5gJYBuCAkrHDXAE6oA+q9TKAHkARgCs4YGAEFOnAgE9qyvgIQBlOgFsADgBs4s+QsrLUAEmABGAL5m+lgNrBUOApxhJUAVgA0qBAATLytvVBsAXTsHYEcrXwAmXwBmXwAWXz8ANl8Adl8ADijTZUtbGOdXGHdPH38gkLDI6ItY8hBsGHb-dpxOOigabtR2gDMITmH2nU44GBg6LnbiqlLgZLsVVsq3Dy8-AKhg1FDwqN5tlbNLNJanF13a+MPj5PDfB+q9k4OG1ASAAzvc4VK4YUgACnacAAHgRdAZMHAcMMIcA1FBWPAOJx-BjRBIpEZFHjGBptPpDHJFDYAJSoAC8AD5UMA6ZQOat0J18EQyCVqBiAMIQBGwgmSGTUpR8AUYe6oAghfwiLxvGwfVBgRr+Y6As42WVc+WxT41ZUvNXvM3fUL1I5efUa1lVc0-e16oEaqJG6iQ6FwsVIlEgfxojFY9hcUmCCVE6Ux8li4kKOmMlls2mhkAQHQLaBYZCgdycCAAdwACu4EIWUAQGBA6IFsyJ5GAANZzdQ6IgDGjIGCcACucF8IGIMC0egA6sQ6Nge2A4Ooa-PGPOFMhwFhC2OBlguDAK-IaFoCMhRgQ9Aex2IsDCAEJtzswdTwuAAGQGcAvV5vIDvGFNEGAwAEUhwgeBf2vUcQB7TgD04LcRAIEQ4D0bMZgGGBpybGBiGQAoATHGYIAPad5B0LcZmRLgmGzABHCD4GPXNaxAAgsAAWigOA4ECfjs1mJi6FmY8CFPc8kEvGCxwPLQ6AHYdYKwPswOYn9pL-WDqhEXDAnw5Akg6eQ6D0PsRS0M8tygaAfzHIcDwAFVQ2sZP-BgR1YATYHUMB+jzaQjnUGAFAMaCDxsGwgA)

**Input:**
```jsx
    test.each`
      givenInteger | givenObjectArray         | givenSimpleArray
      ${1}      | ${[{ start: 5, end: 15 }]} | ${[1,2,3,4,5,6,7,8]}
      ${1}| ${[{ start: 5, end: 15 }]} | ${["test", "string", "for", "prettier"]}
      ${3}      | ${[{ start: 5, end: 15 }]} | ${6}
      ${4} | ${[{ start: 1, end: 3 },{ start: 15, end: 20 },]} | ${15.01}
    `("example test", ({givenInteger, givenObjectArray, givenSimpleArray}) => {})
```

**Output:**
```jsx
test.each`
  givenInteger | givenObjectArray           | givenSimpleArray
  ${1}         | ${[{ start: 5, end: 15 }]} | ${[1, 2, 3, 4, 5, 6, 7, 8]}
  ${1}         | ${[{ start: 5, end: 15 }]} | ${["test", "string", "for", "prettier"]}
  ${3}         | ${[{ start: 5, end: 15 }]} | ${6}
  ${4} | ${[
  { start: 1, end: 3 },
  { start: 15, end: 20 }
]} | ${15.01}
`("example test", ({ givenInteger, givenObjectArray, givenSimpleArray }) => {});

```

**Expected behavior:**

I would expect that the `givenObjectArray` containing multiple objects does not get wrapped and stays in one line and that the alignment is working as expected. This example has been taken from an `jest` test.

yogmel:
Fixes #7653

This adds a verification in the "ArrayPattern" in `printer-estree.js` and updates `shouldBreak` variable if a Jest Each Template Literal is identified.

-----------

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1554,7 +1554,8 @@ function printPathNoParens(path, options, print, args) {
         const needsForcedTrailingComma =
           canHaveTrailingComma && lastElem === null;
 
-        let shouldBreak =
+        const shouldBreak =
+          !options.isJestEachTemplateLiteral &&
           n.elements.length > 1 &&
           n.elements.every((element, i, elements) => {
             const elementType = element && element.type;
@@ -1575,18 +1576,6 @@ function printPathNoParens(path, options, print, args) {
 
             return element[itemsKey] && element[itemsKey].length > 1;
           });
-
-        if (n.type === "ArrayExpression") {
-          let i = 0;
-          let node = n;
-          while (node && shouldBreak) {
-            const parent = path.getParentNode(i++);
-            if (parent && isJestEachTemplateLiteral(node, parent)) {
-              shouldBreak = false;
-            }
-            node = parent;
-          }
-        }
 
         parts.push(
           group(
@@ -2447,17 +2436,18 @@ function printPathNoParens(path, options, print, args) {
     case "TemplateElement":
       return join(literalline, n.value.raw.split(/\r?\n/g));
     case "TemplateLiteral": {
-      let expressions = path.map(print, "expressions");
       const parentNode = path.getParentNode();
 
       if (isJestEachTemplateLiteral(n, parentNode)) {
-        const printed = printJestEachTemplateLiteral(n, expressions, options);
+        const printed = printJestEachTemplateLiteral(options, path, print);
         if (printed) {
           return printed;
         }
       }
 
+      let expressions = path.map(print, "expressions");
       const isSimple = isSimpleTemplateLiteral(n);
+
       if (isSimple) {
         expressions = expressions.map(
           (doc) =>
@@ -3885,18 +3875,22 @@ function printMethodInternal(path, options, print) {
   return concat(parts);
 }
 
-function printJestEachTemplateLiteral(node, expressions, options) {
+function printJestEachTemplateLiteral(options, path, print) {
   /**
    * a    | b    | expected
    * ${1} | ${1} | ${2}
    * ${1} | ${2} | ${3}
    * ${2} | ${1} | ${3}
    */
+  const node = path.getNode();
   const headerNames = node.quasis[0].value.raw.trim().split(/\s*\|\s*/);
   if (
     headerNames.length > 1 ||
     headerNames.some((headerName) => headerName.length !== 0)
   ) {
+    options.isJestEachTemplateLiteral = true;
+    const expressions = path.map(print, "expressions");
+    options.isJestEachTemplateLiteral = false;
     const parts = [];
     const stringifiedExpressions = expressions.map(
       (doc) =>
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1558,7 +1558,6 @@ function printPathNoParens(path, options, print, args) {
           n.elements.length > 1 &&
           n.elements.every((element, i, elements) => {
             const elementType = element && element.type;
-
             if (
               elementType !== "ArrayExpression" &&
               elementType !== "ObjectExpression"
@@ -1579,7 +1578,7 @@ function printPathNoParens(path, options, print, args) {
 
         if (n.type === "ArrayExpression") {
           let i = 0;
-          let node = path.getNode(i);
+          let node = n;
           while (node && shouldBreak) {
             const parent = path.getParentNode(i++);
             if (parent && isJestEachTemplateLiteral(node, parent)) {
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -2439,7 +2439,7 @@ function printPathNoParens(path, options, print, args) {
       const parentNode = path.getParentNode();
 
       if (isJestEachTemplateLiteral(n, parentNode)) {
-        const printed = printJestEachTemplateLiteral(options, path, print);
+        const printed = printJestEachTemplateLiteral(path, options, print);
         if (printed) {
           return printed;
         }
@@ -3875,7 +3875,7 @@ function printMethodInternal(path, options, print) {
   return concat(parts);
 }
 
-function printJestEachTemplateLiteral(options, path, print) {
+function printJestEachTemplateLiteral(path, options, print) {
   /**
    * a    | b    | expected
    * ${1} | ${1} | ${2}
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1555,7 +1555,7 @@ function printPathNoParens(path, options, print, args) {
           canHaveTrailingComma && lastElem === null;
 
         const shouldBreak =
-          !options.isJestEachTemplateLiteral &&
+          !options.__inJestEach &&
           n.elements.length > 1 &&
           n.elements.every((element, i, elements) => {
             const elementType = element && element.type;
@@ -3888,9 +3888,9 @@ function printJestEachTemplateLiteral(path, options, print) {
     headerNames.length > 1 ||
     headerNames.some((headerName) => headerName.length !== 0)
   ) {
-    options.isJestEachTemplateLiteral = true;
+    options.__inJestEach = true;
     const expressions = path.map(print, "expressions");
-    options.isJestEachTemplateLiteral = false;
+    options.__inJestEach = false;
     const parts = [];
     const stringifiedExpressions = expressions.map(
       (doc) =>
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1554,7 +1554,7 @@ function printPathNoParens(path, options, print, args) {
         const needsForcedTrailingComma =
           canHaveTrailingComma && lastElem === null;
 
-        const shouldBreak =
+        let shouldBreak =
           n.elements.length > 1 &&
           n.elements.every((element, i, elements) => {
             const elementType = element && element.type;
@@ -1575,6 +1575,19 @@ function printPathNoParens(path, options, print, args) {
 
             return element[itemsKey] && element[itemsKey].length > 1;
           });
+
+        for (let i = 0; i < path.stack.length; i++) {
+          if (path.getParentNode(i) !== null) {
+            if (
+              isJestEachTemplateLiteral(path.getNode(i), path.getParentNode(i))
+            ) {
+              shouldBreak = false;
+              break;
+            }
+          } else {
+            break;
+          }
+        }
 
         parts.push(
           group(
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -1554,22 +1554,10 @@ function printPathNoParens(path, options, print, args) {
         const needsForcedTrailingComma =
           canHaveTrailingComma && lastElem === null;
 
-        const shouldBreak =
+        let shouldBreak =
           n.elements.length > 1 &&
           n.elements.every((element, i, elements) => {
             const elementType = element && element.type;
-
-            if (n.type === "ArrayExpression") {
-              let i = 0;
-              let node = path.getNode(i);
-              while (node) {
-                const parent = path.getParentNode(i++);
-                if (parent && isJestEachTemplateLiteral(node, parent)) {
-                  return false;
-                }
-                node = parent;
-              }
-            }
 
             if (
               elementType !== "ArrayExpression" &&
@@ -1588,6 +1576,18 @@ function printPathNoParens(path, options, print, args) {
 
             return element[itemsKey] && element[itemsKey].length > 1;
           });
+
+        if (n.type === "ArrayExpression") {
+          let i = 0;
+          let node = path.getNode(i);
+          while (node && shouldBreak) {
+            const parent = path.getParentNode(i++);
+            if (parent && isJestEachTemplateLiteral(node, parent)) {
+              shouldBreak = false;
+            }
+            node = parent;
+          }
+        }
 
         parts.push(
           group(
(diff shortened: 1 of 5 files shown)


yogmel:
Hi @thorn0!

This is not my best piece of code and I'm looking forward to improving it.

As you mentioned, I have to test all parent nodes of the current AST Node. I performed a for loop for that, but I'm not sure what to use as a limiting factor.

I have sent this PR as I thought it might be easier to review it with real code.

This works with the cases presented by the reported and also doesn't affect current tests. I didn't find any test cases with `test.each` template strings, so I didn't update any existing snapshot.

thorn0:
Hi there! Using the same indices for `path.stack` and `path.getNode` isn't the right thing to do because `stack` contains not only nodes but also property names. (I recommend to run Prettier in a debugger to explore the structure of `path`.)

Anyways, it's better to view `path.stack` as a private property of `FastPath` and use only the methods. Examples of loops similar to what we need can be found here:

https://github.com/prettier/prettier/blob/765e3b40251abbf7f636580dfcb99f69da3dd42d/src/language-js/needs-parens.js#L214-L227

and here:

https://github.com/prettier/prettier/blob/765e3b40251abbf7f636580dfcb99f69da3dd42d/src/language-js/printer-estree.js#L277-L290

> This works with the cases presented by the reported

Good! Please add those cases to the tests (in `tests/js/test-declarations`). As you can see, the CI has failed because your change isn't covered by tests ("codecov/patch — 66.66% of diff hit (target 80.00%)").

And just for the sake of accuracy, note that the nodes we work with in this case have the type `ArrayExpression` (array literals), not `ArrayPattern` (part of the destructuring syntax).

yogmel:
Hi @thorn0,

Thanks for the information! Those were pretty useful.

I've changed the loop and added the test cases.

Please let me know if it needs more changes!

src/language-js/printer-estree.js
```diff
@@ -1558,6 +1558,19 @@ function printPathNoParens(path, options, print, args) {
           n.elements.length > 1 &&
           n.elements.every((element, i, elements) => {
             const elementType = element && element.type;
+
+            if (n.type === "ArrayExpression") {
+              let i = 0;
+              let node = path.getNode(i);
```
thorn0, line 1564: @yogmel You shouldn't have moved this loop inside the other loop that iterates through the elements of the array literal. Please move it back.

tests/js/test-declarations/jest-each.js
```diff
@@ -60,3 +60,18 @@ test.each([
     expect(Number(a)).toBe(b);
   }
 );
+
+test.each`
```
thorn0, line 64: Please extract these tests to a separate file. When test files get long, their snapshots are difficult to read.

src/language-js/printer-estree.js
```diff
@@ -1558,6 +1558,19 @@ function printPathNoParens(path, options, print, args) {
...
+              while (node) {
+                const parent = path.getParentNode(i++);
+                if (parent && isJestEachTemplateLiteral(node, parent)) {
+                  return false;
+                }
+                node = parent;
```
fisker, line 1570: This loop back to the root, can we break the loop when meet an expression or something safe?

yogmel:
Hi everyone, sorry I am late.

I have extracted the test to another file, but I can't make `yarn test` pass for some reason.

```
Summary of all failing tests
 FAIL  tests/js/test-declarations/jsfmt.spec.js (6.307 s)
  ● template-literal.js › encountered a declaration exception

    Value false is not a valid document

      20 |     !(typeof val === "string" || (val != null && typeof val.type === "string"))
      21 |   ) {
    > 22 |     throw new Error(
         |           ^
      23 |       "Value " + JSON.stringify(val) + " is not a valid document"
      24 |     );
      25 |   }

      at assertDoc (src/document/doc-builders.js:22:11)
          at Array.forEach (<anonymous>)
      at concat (src/document/doc-builders.js:34:11)
      at Object.genericPrint [as print] (src/language-js/printer-estree.js:227:10)
      at callPluginPrintFunction (src/main/ast-to-doc.js:127:18)
      at comments.printComments (src/main/ast-to-doc.js:64:16)
      at Object.printComments (src/main/comments.js:502:19)
      at printGenerically (src/main/ast-to-doc.js:62:22)
      at FastPath.map (src/common/fast-path.js:124:21)
      at printPathNoParens (src/language-js/printer-estree.js:2451:30)
```

src/language-js/printer-estree.js
yogmel: @fisker by root, do you mean moving this logic outside `printPathNoParens()` function?

thorn0:
The test is failing for a natural reason: something isn't right with the code being tested.

src/language-js/printer-estree.js
```diff
@@ -1576,6 +1577,18 @@ function printPathNoParens(path, options, print, args) {
             return element[itemsKey] && element[itemsKey].length > 1;
           });
 
+        if (n.type === "ArrayExpression") {
+          let i = 0;
+          let node = path.getNode(i);
+          while (node) {
```
thorn0, line 1583:
```suggestion
          while (node && shouldBreak) {
```

thorn0:
@yogmel Looks good. 👍 Please add a changelog entry (to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`).

yogmel:
Thanks for your help and patience, @thorn0
I have submitted the last commit with the changelog regarding this PR.

src/language-js/printer-estree.js
```diff
@@ -1576,6 +1577,18 @@ function printPathNoParens(path, options, print, args) {
             return element[itemsKey] && element[itemsKey].length > 1;
           });
 
+        if (n.type === "ArrayExpression") {
+          let i = 0;
+          let node = path.getNode(i);
```
thorn0, line 1582:
```suggestion
          let node = n;
```

fisker:
I still have a concern about performance, I'll have a test.

fisker:
I run format over `prettier@2.0.5` files.

The `isJestEachTemplateLiteral` function is called 2061 times, less than I thought it would be.

@thorn0 Do you think it worse to add a cache? With cache reduce to 1334 times.



<details>
<summary>The cache logic I use to test</summary>

```js
const cache = new WeakMap();
function isInsideJestEachTemplateLiteral(path) {
  const parents = new Set();

  let i = 0;
  let node = path.getValue();
  while (node) {
    if (cache.has(node)) {
      const result = cache.get(node);
      for (const node of parents) {
        cache.set(node, result);
      }
      return result;
    }
    parents.add(node);
    const parent = path.getParentNode(i++);
    if (parent && isJestEachTemplateLiteral(node, parent)) {
      for (const node of parents) {
        cache.set(node, true);
      }
      return true;
    }
    node = parent;
  }

  for (const node of parents) {
    cache.set(node, false);
  }
  return false;
}
```
</details>

thorn0:
@fisker Agree, it should make sense to optimize it.

As we're already inside a tree-traversing recursion, it'd be more efficient to add a temporary flag property to `options` when printing `expressions` here:

https://github.com/prettier/prettier/blob/7659a867ad77b493f6c7f10a4421063c03a2a8ef/src/language-js/printer-estree.js#L2358-L2367

and then replace our check with checking that property. What do you think?

fisker:
That's a better idea!

yogmel:
Hi everyone,
Is this going to be done in this PR? I can do it, but I might need some (more) guidance.
Thanks!

thorn0:
@yogmel Let's do it in this PR. Guidance is not a problem. I'll be glad to help.

yogmel:
Thanks @thorn0 :) I will have more time to have a look at it again this weekend. I really appreciate your mentorship.

yogmel:
From what I understood, I have to pass a new property to `options`, so that when it passes the `case "ArrayPattern"` operations I can use that to verify, instead of doing the while loop.

The problem I face is that, although the "TemplateLiteral" type is on the outer branch of the object tree, the `case "TemplateLiteral"` routine runs **after** the `case "ArrayPattern"`.

image
(_the test case I'm using is [this one](https://deploy-preview-8354--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuE8DOMB0cCGYAWABgDpQ4AEAPuQEbn0MPVikAkwAjAL6P3XsBtYOQw4ATjCTkArABpyCACZSO08lwC6PfsAEdZAJlkBmWQBZZcgGyyA7LIAcWtpy46hImOMkz5SlWqa2uSCxKhwGGHyYRhiAJZQAOZR5GEAZhBiKWEADmJwMDBxcFkgzlDsxjy87sKiElJyClDK5KrqWlQhuuXsZsGCdV4NbX4tUsbqskPeAWOtBgAMU53u5YQAFGFwAB44ALY5ADZw5OgwKRvAOPI08mBcAJTkALwAfOTAT6QgsiAQOSK0DQyFA4jEEAA7gAFcQIEEoHBHSE4ACeIL+NDEeAA1gUAMo5PAJRLIGBiACucD+cH2NDgikUDIAMjgkhScIk4AAxTL7HCFEnIEA4CkwCC-ED4GD7I4AdXwcXQRLAcHx8KVcQAbkrUcKwGgMSAEmgSjBodjEvzkGkkaa-gArNA7ABC2LAeJg+IOcGZCTgNrt1JATp2+JJJwAihSIPBA0d7SAiWJTWJhTQcPSjpK8gkYHK4ooYPhkA5Fn88hBTXLsTlhXkIiUtQG-gBHGPwC0AhEitAAWigcAZDMl+XbcXyFs51qQtoTwdN+ziZMpC4jcGjsYDs6Dfy8NALRZLSCMqGxcSOJIAwhB9jOQBFpJKKaaACqZhFzxNaqkASRaCBemA8SAgAgi0+IwKiJzxqaXBcEAA)_)

By the time `printJestEachTemplateLiteral()` is called, the Array Expression is already formatted.

Sorry if I misunderstood the task. I forgot how daunting it could be to read a whole new codebase.

thorn0:
You didn't misunderstand anything. The expressions in the placeholders of the template literal are printed (converted from AST to Doc) by this call:
```js
let expressions = path.map(print, "expressions"); 
```
Now it indeed happens before the check that detects `test.each`. We need to change the order of how things are called. The processes of printing `test.each` and other template literals don't necessarily have to share the same `expressions` variable. The printing of the expressions for `test.each` can be moved to `printJestEachTemplateLiteral` and `let expressions = ...` in `case "TemplateLiteral"` can end up after the `test.each` check.

yogmel:
Great, thanks for your help! I've sent another update.
This solution looks much better to me. The only part I was not able to solve optimally was this [one](https://github.com/prettier/prettier/pull/8354/commits/46ffe5701512440fdd39539996dc18be3a28f089#diff-3abca41914b4b214a2507e33cc54c28bR3886-R3888).

I thought about passing a parameter to the `print` callback in  `const expressions = path.map(print, "expressions")`, but `printGenerically` does not take the `options` parameter.

thorn0:
Yes, passing a flag via `options` is not exactly elegant, but still it's a good optimization, so let's keep it this way for now. `printGenerically` accepts `args`, but it's passed only one recursion level deep. One day we might implement something like `deepArgs` for passing such things down the tree.

src/language-js/printer-estree.js
```diff
@@ -3873,13 +3876,16 @@ function printMethodInternal(path, options, print) {
   return concat(parts);
 }
 
-function printJestEachTemplateLiteral(node, expressions, options) {
+function printJestEachTemplateLiteral(node, options, path, print) {
```
thorn0, line 3879: Let's reduce the number of parameters. We don't need `node` if we have `path`.

src/language-js/printer-estree.js
```diff
@@ -3873,13 +3876,16 @@ function printMethodInternal(path, options, print) {
...
    * ${1} | ${1} | ${2}
    * ${1} | ${2} | ${3}
    * ${2} | ${1} | ${3}
    */
+  options.isJestTemplateLiteral = true;
+  const expressions = path.map(print, "expressions");
+  options.isJestTemplateLiteral = false;
   const headerNames = node.quasis[0].value.raw.trim().split(/\s*\|\s*/);
   if (
     headerNames.length > 1 ||
```
thorn0, line 3891: Please move printing the expressions inside this `if`. Print results are cached, so if this condition doesn't hold true , the other `print` call for the expressions (in `case "TemplateLiteral"`) will get the same results, which will be wrong.

yogmel:
Ok, thanks again @thorn0! Please let me know if it looks ok.

src/language-js/printer-estree.js
```diff
@@ -3873,18 +3875,22 @@ function printMethodInternal(path, options, print) {
   return concat(parts);
 }
 
-function printJestEachTemplateLiteral(node, expressions, options) {
+function printJestEachTemplateLiteral(options, path, print) {
```
thorn0, line 3878: Please change the order to `path, options, print` to be consistent with similar `print*` functions.

yogmel:
Thanks for the help @thorn0 @fisker !

PR #8360, opened 2020-05-20
LESS: Fix inline comments contains `/*`

### LINKED ISSUE #8130: [Less] Comment deleted for no reason
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAlgWwA4QE4wAEAOiAHRkD0UEAJnAPqZ0CuANnAM6UBmEElAEYBDPGQ6dOpANzEocypUIYo9AB4AKANoByKpBzQEMbgCpTlU+K6cdAGkI6AhFXOUG9ZpdM6Aug55CAF4APkIAAwxDAkcAEmAeMmxhGAALAF94xLg1GHSdaXCASjko3Bi9SgNcKGNuACEWGBhoSkbm6GtJAtKscqJK6qNYbgAxfExKcbxMLttZKDL8Af0IQ1qRygBJKGwm7d2muZ6oRWUEWnRVHLkQOxAIbBh0aE5kUFE8CAB3AAVRBBvFDCNjfYQATze90EeGEYAA1nAYABlZJgK4Ac2QMDwLDg91SMEwbAA6ql0PBOGi4MjART0AA3Cng5DgSR3EBXThwAi-WEYzDCZA8EHc+4AK04anqsIRSORwkwcAAMlc4MLRfiQJK1MjMRwAIosCDwDVsMUgZJ4bl4VkSKGWvBXGAk9C0NLIAAcAAZ7tgvtySbDsKz-VweQz1fcAI7G+B8x5AkDCTgAWlqcHotA5eDgsfQub5wgFQqQIvNWu5mHQ2Nxlf1cCNJvVZc19xgwkErvdqWQACZ27D0GxMQBhNaC1lcACsHJY3IAKp2geWLQy8Tt6LBkWAnU8AIKqZEwcEcM3c9LpIA)
```sh
--parser less
```

**Input:**
```less
@import "../node_modules/foo/bar.less";

// @index(['./components/**/*.less', '!./**/_demo/**'], f => `@import '${f.path}${f.ext}';`)
@import './components/Button/Button.less';
@import './components/Form/Form.less';
@import './components/Input/Input.less';
// @endindex

```

**Output:**
```less
@import "../node_modules/foo/bar.less";

// @index(['./components/**/*.less', '!./**/_demo/**'], f => `@import '${f.path}${f.ext}';`)
@import "./components/Button/Button.less";
@import "./components/Form/Form.less";
@import "./components/Input/Input.less";
@import './components/Form/Form.less';

```

**Expected behavior:**

Don't delete `// @endindex`.

fisker:
Fixes #8130


thorn0:
I would suggest a different solution for this. See `replaceQuotesInInlineComments` in `loc.js`. Replace `/*` there, just like quote marks are replaced.

fisker:
> I would suggest a different solution for this. See `replaceQuotesInInlineComments` in `loc.js`. Replace `/*` there, just like quote marks are replaced.

I'll try

fisker:
@thorn0 This solution seems works, also I simplify the logic a little bit.

src/language-css/loc.js
```diff
@@ -186,27 +184,21 @@ function replaceQuotesInInlineComments(text) {
...
 
-  for (const [start, end] of inlineCommentsToReplace) {
-    text =
-      text.slice(0, start) +
-      text.slice(start, end).replace(/["']/g, " ") +
-      text.slice(end);
-  }
-
-  return text;
+  return chars.join("");
```
thorn0, line 201: I'm worried about the efficiency of this approach. Previously, string manipulations happened only if bug-inducing comments were found. Now they will happen always, plus creating a potentially huge array with `text.split("")` looks memory-unfriendly. Would be good to explore the performance impact with `--debug-benchmark` and a big file.

fisker:
**Prettier pr-8360**
[Playground link](https://deploy-preview-8360--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEABAlgWwA4QE4wAEAOiAIakDcxUA9LQFSEA8AtK4TABboDOhAIzxwyAa34AzfJjIwY6KAHMaGHPiIByARup0mAgDYQwowg1o1GUCK17YyYOIKMnzIADQgI2edF7JQMjw8CAB3AAUghH8UMgNQsgBPf08hB1E4GABlezAFRWQYPABXOE84TAE4ABNqmoAZMiViskU4ADFpWXklZHJimAgPEC4YTAMAdR54Owc4LOj0eQA3JcS+sF4UkAVeOAJwvFaZZAk4vc8AK14ADwAhI5NMrLJMOHqFOFPzspBrm6y+QMcAAisUIPBvgYLiB7Hg9ng+sCtsNsHgFDAJuhqtxkAAOAAMnjRED2EyO2D6aLgCOWX08AEdwfBDt4YuReKwoHAajVhsImehhIdjmQoTC9ph0IUSr9eEDQcyvkgztDfjAyAIsTiuMgAEyeIpkdAGfIAYQgmBOKBpAFZhsU9gAVTUxVUw5alACSUDqsCyYHRPgAgr6sjBEsDxXAAL4xoA)
```sh
--parser less
```

**Input:**
```less
@import "a";
//* <-- this breaks formatting
@import 'b';
/* block */
/*no-space block*/
```

**Output:**
```less
@import "a";
//* <-- this breaks formatting
@import "b";
@import 'b'
/* block */
/*no-s

```

This one still broken, may be replace `*`

thorn0:
> may be replace *

👍

src/language-css/loc.js
fisker: I think replace and concat text multiple times also slow.

How about a new empty string, concat the char when parsing? I'm not sure about performance.

thorn0:
Please add a link to https://github.com/prettier/prettier/issues/8130 to the jsdoc of that function.

src/language-css/loc.js
thorn0: The thing is, often files don't contain such comments at all.

fisker:
```js
const string = '*'.repeat(1e7)

console.time('array');
const a = string.split('');
for (let i = 0; i< 1e4;i++) {
  a[i] = '_'
}
a.join('')
console.timeEnd('array');

console.time('replace');
for (let i = 0; i< 1e4;i++) {
  string.slice(0,i) + string.slice(i,1).replace(/\*/g,'_') + string.slice(i+1)
}
console.timeEnd('replace');
```

Yours faster...

fisker: I'll restore yours

thorn0: You can't really get any meaningful results this way. Something like Benchmark.js needs to be used to run things in a loop. It's even built into Prettier (`--debug-benchmark`).

Anyway, my point is that for most files these string modifications aren't needed at all.

fisker:
So, I just add two `*` to your code? :smile:

thorn0:
Good fixes often are often small like that. 😄

PR #8373, opened 2020-05-21
Enable glimmer printer in the playground

dcyriller:
## Description

https://prettier.io/playground doesn't let select `glimmer` parser (for handlebars) because it is still experimental. With this PR, I propose to enable glimmer parser on the playground. It would simplify the bug reporting on that parser.

AFAICT glimmer is the only parser to be experimental.

## Checklist


dcyriller:
Thank you all!

Issue #8374, opened 2020-05-21
handlebars: `singleQuote` option is not respected for some attributes


dcyriller:
The value of an attribute is not wrapped into the correct quotes when the value is more complex than a bare text.

**Prettier pr-8365**
[Playground link](https://deploy-preview-8365--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAhgAgBYCc4BmAvAOQD0wwArrgDYC+9JAfKmeswDpQY77GcQFanUaDW7LlG688hUjVos2HGVjkCQi8SuYgANCAgAHGAEtoAZ2Sh0uXBADuABTsJrKdLUfoAntcMAI1x0MABrOBgAZWNQsygAc2QYXCo4QzgAW0C4ABNcvIAZdESqdAS4ADEIXEz0GHNE5BB0KhgIAxBsGEzaAHVsM3hLWLA4KPchswA3Id9msEsAkHjLOFwYZxCEuuQCLzXDACtLAA8AIRDwyKj0TLhC+Lg9g-SQE9Oo+ITaOABFKgQeAvWiHECxXBrXDNH5mTL3aGGYy4eIwPpmXIwbDIAAcAAYkQ41n0QsZmsi4FDps9DABHQHwLYmDwtSwAWigcDyeU6+HpZnwW3KuyQ+1BbzWmTMyVSEu+vwBQOeSBSaUMMHQgXRmOxSAATOqQmZaN8AMIQeHoZqUgCsnSoawAKpqPGKwdM0gBJKAFWBRMAo0wAQR9URgvl+ILWjCAA)
```sh
--parser glimmer
--single-quote
```

**Input:**
```hbs
<a href='/{{url}}'></a>
<a href="/{{url}}"></a>

<a href='url'></a>
<a href="url"></a>
```

**Output:**
```hbs
<a href="/{{url}}"></a>
<a href="/{{url}}"></a>

<a href='url'></a>
<a href='url'></a>
```

**Expected behavior:**

PR #8375, opened 2020-05-21
Respect singleQuote option around attribute value more complex than a bare text

dcyriller:
## Description

This would close https://github.com/prettier/prettier/issues/8374

## Checklist


dcyriller:
When the attribute value is a text, a quote would get escaped (it is already implemented). But when the attribute value is more complex, for instance it is composed of a text and a mustache, a quote in the text was / is not escaped. 

It is not pretty at all but it is a valid syntax. Because Handlebars supports the quotes not being escaped. I believe implementing the escape for nested texts into attribute values would be a solid feature (and totally worth it).

Issue #8122, opened 2020-04-22
Refactor comment attaching to not mutate the AST


thorn0:
From https://github.com/prettier/prettier/issues/8083#issuecomment-617424564:

> Prettier mutates the AST. It's in the comment attaching phase that this happens: it adds the comments property to nodes and does other things. But it shouldn't. It's technical debt that has to be refactored, and we don't want to add to it. Mutation makes Prettier less useful in scenarios where ASTs are reused. Now, Prettier doesn't have a public API like prettier.formatAST(ast), but it's going to have it one day. It'd already have it if not for those mutations.

siefkenj:
Would a `Map` or a `WeakMap` be suitable for storing all the associations? I could imagine a proxy object could dynamically insert the `.comments` attribute when backwards compatibility is needed. I'm not sure the performance implications, though.

thorn0:
That should work. Separate maps should be created for every format run as AST nodes can be reused between runs.

As for backwards compatibility, the current logic should be kept and used by default unless the printer has some explicit flag.

siefkenj:
@thorn0 Can you explain what you mean by "Separate maps should be created for every format run"? Doesn't the prettier get called once and then exists when it's returned the formatted file? Are there situations where someone would call prettier a second time without reparsing a file into a new AST?

thorn0:
> Doesn't the prettier get called once and then exists when it's returned the formatted file?

Prettier is not only a command-line tool but also a library used e.g. by editor integrations via its API.

> Are there situations where someone would call prettier a second time without reparsing a file into a new AST?

Prettier doesn't support such situations **now**, but they're possible. E.g. see [this thread](https://twitter.com/sebmck/status/1240443965737861120).

fisker:
Can we make a clone before print?

thorn0:
No, that's the point of this issue.

fisker:
I've done refactor this part, will send PR soon.

Issue #4290, opened 2018-04-09
Add info re: comment handling to the Plugin API documentation


suchipi:
We should document that you can add `handleComments`, `getCommentChildNodes`, `canAttachComment`, and `printComment` to your printer, what they're for, how prettier will use them, and how to use `addLeadingComment`, `addTrailingComment`, and `addDanglingComment`. As part of this, we should explain what it means for a comment to be "leading", "trailing", "dangling", "ownLine", "endOfLine", and "remaining".

cc @mgrip

yxliang01:
@suchipi The documentation is much needed to make plugin development for new developer easier. Hope it can be added soon.

siefkenj:
If someone can explain some of these functions to me, I can take a stab at writing documentation. I'm currently trying to figure it out myself...

thorn0:
@siefkenj Unfortunately, that's a lot to investigate and explain. I might be able to help you if you ask more specific bite-sized questions.

siefkenj:
Okay, here's what I understand so far:

`ownLineComment` is a comment that is on a line by itself. `endOfLineComment` is a comment at the end of a line of code. What is a `remaining` comment? And what is a `dangling` comment?

Also, I don't understand the purpose of `willPrintOwnComment`. Whether it returns true or false, the `printComment` function is still called. What exactly does it buy you? Is the data returned by `printComment` inserted somewhere else if `willPrintOwnComment -> true`?

It seems the general strategy for printing comments is to find which AST node the comment is contained in, call `canAttachComment`, and if it returns true, start calling `handleComments.ownLine`/`endOfLine`/`remaining` based on properties of the raw character stream. Each of those functions either "attaches" the comment to the AST with some properties set and returns `true` or returns `false` and some default behavior is used.

thorn0:
Attaching comments and printing them are separate processes. I've mostly dealt with the former, so I'm going to answer only about that now.

> ownLineComment is a comment that is on a line by itself. endOfLineComment is a comment at the end of a line of code. What is a remaining comment? And what is a dangling comment?

[...]

PR #8388, opened 2020-05-24
Clarified usage of `embed`

siefkenj:
The current documentation doesn't give good guidance about how to use the `embed` function or what the call chain is. This PR explains in more detail.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -184,7 +184,7 @@ function embed(
 ): Doc | null;
 ```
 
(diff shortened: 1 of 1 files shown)


thorn0:
Thanks!

siefkenj:
There are some other things undocumented, like how the `.comments` attribute is handled and the `canAttachComment` and `pringComment` functions. Are these supposed to be used by plugins or are they supposed to be internal?

thorn0:
They're totally intended for use by plugins, see #4290. E.g. [`@prettier/plugin-php` ](https://github.com/prettier/plugin-php) uses them. 

However, they could use some refinement:

* The `add...Comment` methods (`addLeadingComment`, `addTrailingComment`, and `addDanglingComment`) currently mutate the AST, which hopefully will get refactored in the future. See #8122.
* Methods for accessing attached comments (e.g. `hasTrailingComments`) should be part of the plugin API too, but currently they aren't, and every plugin implements them independently, which breaks encapsulation. If the implementation of `add...Comments` changes, those methods will break, which means the old (mutating) way to attach comments still has to be kept as a default backwards compatible mode.

PR #8394, opened 2020-05-25
Support babel `data-type="module"`

fisker:
Ref: https://github.com/babel/babel/pull/11466

### DIFF
--- a/src/language-html/printer-html.js
+++ b/src/language-html/printer-html.js
@@ -72,24 +72,25 @@ function embed(path, print, textToDoc, options) {
             parser === "markdown"
               ? dedentString(node.value.replace(/^[^\S\n]*?\n/, ""))
               : node.value;
+          const textToDocOptions = { parser };
+          if (options.parser === "html" && parser === "babel") {
+            let sourceType = "script";
+            const { attrMap } = node.parent;
+            if (
+              attrMap &&
+              (attrMap.type === "module" ||
+                (attrMap.type === "text/babel" &&
+                  attrMap["data-type"] === "module"))
+            ) {
+              sourceType = "module";
+            }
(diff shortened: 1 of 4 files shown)


src/language-html/printer-html.js
```diff
@@ -72,24 +72,25 @@ function embed(path, print, textToDoc, options) {
...
               breakParent,
               printOpeningTagPrefix(node, options),
-              stripTrailingHardline(
-                textToDoc(
-                  value,
-                  options.parser === "html" && parser === "babel"
-                    ? {
-                        parser,
-                        __babelSourceType:
-                          node.attrMap && node.attrMap.type === "module"
```
fisker, line 86: This is mistake from #8173, should be `node.parent.attrMap `

fisker:
@thorn0 About [this](https://github.com/prettier/prettier/pull/8173#discussion_r422095375), I tried to remove `opts &&` in  `const sourceType = opts && opts.__babelSourceType === "script" ? "script" : "module";` again, and found it's called from [here](https://github.com/prettier/prettier/blob/d420202df4443009d54a6288d65e5d3219aa746f/tests_integration/custom-parsers/custom-rename-parser.js#L4).

thorn0:
A changelog entry is needed. Maybe just modify the one for #8173.

fisker:
The `changelog_unreleased/html` dir contains `pr-8137.md` `pr-8173.md` `pr-8371.md`... What a coincidence!

thorn0:
Yep, I noticed that too. :)

fisker:
Done.

fisker:
Another coincidence, when I see this babel PR yesterday, we both have 627 open issues. 😄

PR #8407, opened 2020-05-25
Updated plugin documentation

siefkenj:
Greatly expand the plugin documentation for Prettier plugin authors. Partially addresses #4290


docs/plugins.md
```diff
@@ -136,25 +136,54 @@ function preprocess(text: string, options: object): string;
 
 Printers convert ASTs into a Prettier intermediate representation, also known as a Doc.
 
-The key must match the `astFormat` that the parser produces. The value contains an object with a `print` function and (optionally) an `embed` function.
+The key must match the `astFormat` that the parser produces. The value contains an object with a `print` function. All other attributes (`embed`, `preprocess`, etc.) are optional.
```
thorn0, line 139: Please call them "properties", not "attributes".

thorn0:
Huge thanks for doing this. The result can be [previewed here](https://deploy-preview-8407--prettier.netlify.app/docs/en/next/plugins.html#utility-functions).

docs/plugins.md
```diff
@@ -184,20 +236,162 @@ function embed(
...
+	ast: AST,
+	// Whether this comment is the last comment
+	isLastComment: boolean
+): boolean
+```
+
+These functions are used to override Prettier's default comment
+assignment algorithm. `ownLine`/`endOfLine`/`remaining` is expected
+to either manually assign a comment to a node (removing it from the
+`astRoot.comments` array) and return `true`, or return a `falsy` value
```
thorn0, line 355: `handleComments.*` don't remove comments from that array.

The reference to that array is removed from the AST root, but this fact isn't directly related to the `handleComments` functions:

https://github.com/prettier/prettier/blob/8718bb4be44f6fd1aed8fcf91717f38931643dd7/src/main/core.js#L56-L60

siefkenj:
Thanks for being helpful with your responses! :-)

docs/plugins.md
```diff
@@ -136,25 +136,54 @@ function preprocess(text: string, options: object): string;
...
 };
 ```
 
-Printing is a recursive process of converting an AST node (represented by a path to that node) into a doc. The doc is constructed using the [builder commands](https://github.com/prettier/prettier/blob/master/commands.md):
+#### The printing process
+
+Prettier uses an intermediate representation, called a
+Doc, which Prettier then turns into a string (based on
+options like `textWidth`). A _printer_'s job is to take
+the AST generated by `parsers[<plugin>].parse` and
```
thorn0, line 165: It's `parsers[<parser name>.parse`. A plugin may contain multiple parsers.

docs/plugins.md
```diff
@@ -184,20 +236,162 @@ function embed(
...
+node (i.e., created new properties) with at least one of `enclosingNode`,
+`precedingNode`, or `followingNode`. These can be used to aid a plugin's
+decision process (of course the entire AST and original text is also
+passed in for making more complicated decisions).
+
+_Manually assigning a comment_
+
+Nodes with comments are expected to have a `comments` property containing
+an array of comments. Each comment is expected to have the following
+properties: `leading`, `trailing`, `printed`. An example `ownLine`
```
thorn0, line 374: These details need to be de-emphasized in favor of `util.add*Comment` as they can change because of https://github.com/prettier/prettier/issues/8122 whereas `util.add*Comment` will stay.

docs/plugins.md
```diff
@@ -184,20 +236,162 @@ function embed(
...
+	options: object,
+	// The AST
+	ast: AST,
+	// Whether this comment is the last comment
+	isLastComment: boolean
+): boolean
+```
+
+These functions are used to override Prettier's default comment
+assignment algorithm. `ownLine`/`endOfLine`/`remaining` is expected
```
thorn0, line 354: So are comments "attached" or "assigned"? Let's choose one (attach) and use it consistently.

docs/plugins.md
```diff
@@ -136,25 +136,54 @@ function preprocess(text: string, options: object): string;
...
 
-Printing is a recursive process of converting an AST node (represented by a path to that node) into a doc. The doc is constructed using the [builder commands](https://github.com/prettier/prettier/blob/master/commands.md):
+#### The printing process
+
+Prettier uses an intermediate representation, called a
+Doc, which Prettier then turns into a string (based on
+options like `printWidth`). A _printer_'s job is to take
+the AST generated by `parsers[<parser name>].parse` and
+return a Doc. A Doc is constructed using [builder
+commands](https://github.com/prettier/prettier/blob/master/commands.md):
```
lipis, line 167: Can we have these paragraphs in one line?! Editors and github are handling that perfectly.. and it's much easier to change in the future, if we need to update the copy.

thorn0: I thought Jason did this intentionally to make it easier to add review comments in the middle of a paragraph. Before merging we'll reformat the file with `--prose-wrap always`.

siefkenj: That's exactly right. I wrapped the text to make it easier to review and add diffs. prosewrap seems like a good solution :-)

thorn0:
I don't think we're done. There are still important points missing here and unanswered questions left in https://github.com/prettier/prettier/issues/4290. I'm still going to answer those.

docs/plugins.md
lipis: Awesome!

siefkenj:
Any updates on merging this? If it's going to take a long time to flesh out the rest of the plugin features, it's probably worth merging this documentation now and doing the rest in a separate PR.

fisker (approved):

evilebottnawi (approved):

fisker:
Note this https://github.com/prettier/prettier/pull/8407#discussion_r431198290

thorn0:
@siefkenj Oops. Sorry, I actually meant `--prose-wrap never`, not `always`.

thorn0 (approved):

fisker:
Update stable docs?

thorn0:
Right, thanks. Did it: c94d1bc014b146eb2

PR #8408, opened 2020-05-25
Add an ability to specify ranges for test snippets

thorn0:
I think I have a fix for #4926, but to proceed with it (to make it easier to review), let's merge this first.

### DIFF
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -184,7 +184,7 @@ global.run_spec = (fixtures, parsers, options) => {
                     )
                   : source,
                 hasEndOfLine ? visualizedOutput : formattedWithCursor,
-                { ...baseOptions, parsers }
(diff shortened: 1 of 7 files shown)


tests_config/run_spec.js
```diff
@@ -111,140 +111,158 @@ global.run_spec = (fixtures, parsers, options) => {
 
   const stringifiedOptions = stringifyOptions(options);
```
thorn0, line 113: This file's diff will look better if you [hide workspace changes](https://github.com/prettier/prettier/pull/8408/files?diff=unified&w=1).

fisker:
Don't you think it's clear?

```js
code: outdent`
  export default  c<<<PRETTIER_RANGE_START>>>lass Foo{
  /**/
  }<<<PRETTIER_RANGE_END>>>
`
```

thorn0:
Maybe you're right. Let's not complicate it with one more way to specify the same thing.

PR #8411, opened 2020-05-26
Refactor: remove use of `escape-string-regexp`

fisker:
`getMaxContinuousCount` has tests covered #7262 .

Other changes are internal.


src/common/util.js
```diff
@@ -682,37 +681,45 @@ function printNumber(rawNumber) {
...
-    new RegExp(`(${escapeStringRegexp(target)})+`, "g")
-  );
-
-  if (matches === null) {
-    return 0;
-  }
+  let max = 0;
+  const targetLength = target.length;
 
   const countPresent = new Map();
```
fisker, line 708: Original implementation use `Map`, not sure about the performance between `Set` and `Map`

evilebottnawi:
Just a question - Why?

fisker:
> Just a question - Why?

I think we used it in simple cases, don't really need it.


````js
	const delimiterRegex = Object.keys(DELIMITER_MAP).map(escape).join("|");
    new RegExp(	
      `^(${delimiterRegex})([^\\n]*)\\n(?:([\\s\\S]*?)\\n)?\\1[^\\n\\S]*(\\n|$)`	
    )
```

vs 

```js
const frontMatterRe = /^(-{3}|\+{3})([^\n]*)\n(?:([\S\s]*?)\n)?\1[^\S\n]*(\n|$)/;
```

Don't you think the second one easier to read?

evilebottnawi:
@fisker both are hard to read :smile: to be honestly we need check perf

fisker:
Yes, we need, but I don't know how 😢

fisker:
Close this as it has been a long time , I assume other member don't agree on this, also there is a `Regexp.escape` proposal, I think we'll be able to use it when runtime supports it.

PR #8420, opened 2020-05-27
Test: visualize the range

fisker:


fisker:
Seems little bit wrong, don't know how to fix...

thorn0:
That's sad. I like the idea.

fisker:
I even tried location calc function from babel repo, still not right.

fisker:
I think I fixed it.

Someone can compare it again?

thorn0:
Looks good.

fisker:
@thorn0  You new tests https://github.com/prettier/prettier/pull/8420/commits/3c88507b5fc702eaff407966f6bae47e462766a2

thorn0:
Beautiful!

fisker:
image

@thorn0 Do we want this?

tests/js/cursor/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -316,12 +319,15 @@ parsers: ["babel", "typescript", "flow"]
 printWidth: 80
                                                                                 | printWidth
```
thorn0, line 320: The printWidth indicator needs to be shifted 6 characters to the right for these tests.

fisker: `.match(/^\s+1 \| /)[0].length` Big file more than `6`, I'll fix it.

thorn0:
> Do we want this?

I don't think it's needed.

fisker:
So excited, this works.

lydell:
It’s such a joy to occasionally visit the Pull requests tabs in this repo and see all the great stuff you all are doing! 👍

fisker:
A memo for someone who want refactor it, the following code works too.

```js
const { SourceCode } = require("eslint");

const locationForRange = (text, rangeStart, rangeEnd) => {
  const sourceCode = new SourceCode(text, {
    tokens: [],
    comments: [],
    range: [],
    loc: {},
  });

  const start = sourceCode.getLocFromIndex(rangeStart);
  const end = sourceCode.getLocFromIndex(rangeEnd);

  start.column += 1;
  if (start.line === end.line) {
    end.column += 1;
  }
};
```

PR #8436, opened 2020-05-28
Babel: support moduleAttributes

fisker:
- https://babeljs.io/blog/2020/05/25/7.10.0#module-attributes-parser-support-10962-https-githubcom-babel-babel-pull-10962
- https://github.com/tc39/proposal-module-attributes/blob/master/README.md


thorn0:
I wonder if we should add some kind of disclaimer about stage 1 and 2 proposals. We don't want to end up needing to support all the versions of experimental syntax at the same time.

fisker:
We can drop old syntax if spec changes.

PR #8475, opened 2020-06-02
Fix plugin auto search dirs

### LINKED ISSUE #8474: Plugin autoload doesn't work for packages installed from GitHub and for Yarn PnP
**Environments:**

- Prettier Version: 2.0.5
- Running Prettier via: CLI
- Runtime: Node.js v14
- Operating System: Windows

**Steps to reproduce:**

```console
yarn add prettier/prettier @prettier/plugin-pug
```

**Expected behavior:**

Can format pug file with `npx prettier test.pug`

**Actual behavior:**

The CLI can't load `pug` plugin automatically

----

Problem should be [here](https://github.com/prettier/prettier/blob/6b4c4066aa1d59e6f806b1646beee314f93103d3/src/common/load-plugins.js#L30), when install from github, `./node_modules/prettier` maybe has own `node_modules` dir, so the `autoLoadDir` will be `./node_modules/prettier`, not expected `./`, so we are looking for `./node_modules/prettier/node_modules/@prettier/plugin-*` not `./node_modules/@prettier/plugin-*` (where the `@prettier/plugin-pug` is installed).

fisker:
Fixes #8474

Not sure how to test it


src/common/load-plugins.js
```diff
@@ -5,18 +5,43 @@ const partition = require("lodash/partition");
...
+    pluginSearchDirs.push(autoLoadDir);
+  }
+
+  // We may have plugins in `prettier/prettier` codebase for test
+  if (isDirectory(path.join(packageRoot, "node_modules"))) {
+    pluginSearchDirs.push(packageRoot);
+  }
+
+  return pluginSearchDirs;
+}
```
evilebottnawi, line 43: What about yarn pnp?

fisker: We don't support it.


Possible solution `--plugin=@prettier/plugin-pug`

evilebottnawi: I think will be great to support them in the future

thorn0: @evilebottnawi I'm interested in your thoughts on https://github.com/prettier/prettier/issues/8474#issuecomment-639013515.

fisker:
Can we merge this first, to make [this](https://github.com/prettier/plugin-pug/issues/5#issuecomment-658651088) work, without additional cli argument?

src/common/load-plugins.js
merceyz: You shouldn't have to reinvent the node resolution, why not use `require.resolve` to find the plugin from the project that depends on prettier / whatever invokes it?

fisker: Because we don't know the plugin name yet.

merceyz: Oh, I assumed you loaded the name from `package.json` which would be the sensible thing to do

fisker: Currently, we search `node_modules`, so some plugins might be "indirect" dependencies.

fisker: Imaging this, you have `@merceyz/prettier` package, include `config` and `plugins`, you may don't want list all plugins in every project.

merceyz: Then I would assume `@merceyz/prettier` declares the plugins name/location in a config and it's loaded from there, basically the same way babel handles plugins/presets

fisker: Yep, that's my proposal https://github.com/prettier/prettier/issues/8474#issuecomment-661543282, but we haven't support it.

fisker:
Plugin search already been removed #14759

PR #8499, opened 2020-06-04
Minor refactor `doc-utils`

fisker:

### DIFF
--- a/src/document/doc-utils.js
+++ b/src/document/doc-utils.js
@@ -1,7 +1,7 @@
 "use strict";
 
-// Using a unique symbol to compare
-const traverseDocOnExitStackMarker = Symbol("traverseDocOnExitStackMarker");
+// Using a unique object to compare by reference.
+const traverseDocOnExitStackMarker = {};
 
 function traverseDoc(doc, onEnter, onExit, shouldTraverseConditionalGroups) {
   const docsStack = [doc];
--- a/src/document/doc-utils.js
+++ b/src/document/doc-utils.js
@@ -1,7 +1,7 @@
 "use strict";
 
-// Using a unique object to compare by reference.
-const traverseDocOnExitStackMarker = {};
+// Using a unique symbol to compare by reference.
+const traverseDocOnExitStackMarker = Symbol("traverseDocOnExitStackMarker");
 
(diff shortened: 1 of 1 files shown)


src/document/doc-utils.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
-// Using a unique object to compare by reference.
-const traverseDocOnExitStackMarker = {};
+// Using a unique symbol to compare
+const traverseDocOnExitStackMarker = Symbol("traverseDocOnExitStackMarker");
```
thorn0, line 4: What is the point in using a symbol if we aren't going to use it as a property key?

fisker: Shouldn't `Symbol` preferred for unique thing?

thorn0: Its main purpose is to be used as a key. Declaring a symbol and not using it as a key might be misleading.

fisker: I can't agree there will be any misleading.

Quote from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol)

> Every symbol value returned from Symbol() is unique.  A symbol value may be used as an identifier for object properties; this is the data type's primary purpose, although other use-cases exist, such as enabling opaque data types, or serving as an implementation-supported unique identifier in general. 

Unique identifier is a good `Symbol` use cases.

fisker: I searched our codebase, seem only the third one is used as a key.

- https://github.com/prettier/prettier/blob/master/src/language-html/printer-html.js#L252
- https://github.com/prettier/prettier/blob/3094ad5aa24ed9bf0826e60a8a36fd997f9259ce/src/language-js/printer-estree.js#L662
- https://github.com/prettier/prettier/blob/master/src/main/comments.js#L25 **
- https://github.com/prettier/prettier/blob/master/src/main/core.js#L23

thorn0: I'm not sure what "implementation-supported unique identifier" means. Anyway "identifier" means roughly the same as "name" or "key", something that points at something else. And that's what symbols are for.

thorn0: Group ids are definitely used as keys.

fisker: Anyway, rest part good?

PR #8500, opened 2020-06-04
Fix crash on `tabWidth: 0`

### LINKED ISSUE #7388: tabWidth=0 cause crash
**Environments:**
- Prettier Version: 1.19.1
- Running Prettier via: CLI
- Runtime: node v12
- Operating System: Linux

**Steps to reproduce:**
1. set `tabWidth=0`
2. run prettier

**Expected behavior:**
print some error message

**Actual behavior:**
out of memory and crash

**debug**
Crash is comes from [Here](https://github.com/prettier/prettier/blob/c68d6e3c2b365a2d9cbdfe353f279d95e22bb911/src/document/doc-builders.js#L218)

**why**
I'm using some tool to create an `.prettierrc.js`, and don't know what happend, it create a config file with `tabWidth=0`.    
Of cource this is caused by third party tool.    
But prettier should not crash because this.    
It's really hard to find out what's happening.

fisker:
Fixes #7388

### DIFF
--- a/src/main/core-options.js
+++ b/src/main/core-options.js
@@ -237,7 +237,7 @@ const options = {
     category: CATEGORY_GLOBAL,
     default: 2,
     description: "Number of spaces per indentation level.",
-    range: { start: 0, end: Infinity, step: 1 },
+    range: { start: 1, end: Infinity, step: 1 },
   },
   useTabs: {
     since: "1.0.0",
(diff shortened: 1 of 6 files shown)


src/main/core-options.js
```diff
@@ -237,7 +237,7 @@ const options = {
     category: CATEGORY_GLOBAL,
     default: 2,
     description: "Number of spaces per indentation level.",
-    range: { start: 0, end: Infinity, step: 1 },
+    range: { start: 1, end: Infinity, step: 1 },
```
fisker, line 240: We can't validate this option?

thorn0: I see some `validate` calls in `options-normalizer.js`, but looks like they don't check this `range` property.

Issue #8204, opened 2020-05-01
Remove `flow` support from `parser: babel`


fisker:
I was trying to improve parse performance on large files, and found it will cost almost half time on some file by removing plugins.

Something interesting come up, we have `flow` and `babel-flow` parser, but we also enabled `flow` syntax in `babel` plugins https://github.com/prettier/prettier/blob/master/src/language-js/parser-babel.js#L101

Shouldn't we encourage people use `flow` or `babel-flow`? People use `typescript` and `babel-ts` for `.ts`, not `babel`.

Can we remove it?

---

Refs: 

Motivation for `babel-flow`  https://github.com/prettier/prettier/issues/5460#issuecomment-441298551 and https://github.com/prettier/prettier/issues/5460#issuecomment-441360123

PR introduce `babel-flow` https://github.com/prettier/prettier/pull/5685

thorn0:
> found it will cost almost half time on some file by removing plugins.

Could you show some benchmark results?

> Can we remove it?

I guess we can if we make Prettier switch to `babel-flow` automatically when the `@flow` pragma is found.

fisker:
> Could you show some benchmark results?

Will do.

> I guess we can if we make Prettier switch to babel-flow automatically when the @flow pragma is found.

My intention is not support flow at all, but check `@flow` is a good idea, so we don't break anything.

fisker:
Run against `src/**/*`

<details>
<summary>Test script</summary>
<pre><code>
const fs = require('fs')
const oldParse = require('./src/language-js/parser-babel-master').parsers.babel.parse
const newParse = require('./src/language-js/parser-babel').parsers.babel.parse
const files = require('fast-glob').sync('src/**/*').map(file => fs.readFileSync(file, 'utf8'))


console.log('first run')
console.time('new')
for (const code of files) {
  newParse(code)
}
console.timeEnd('new')
console.time('old')
for (const code of files) {
  oldParse(code)
}
console.timeEnd('old')

console.log()
console.log('second run')

console.time('old')
for (const code of files) {
  oldParse(code)
}
console.timeEnd('old')


console.time('new')
for (const code of files) {
  newParse(code)
}
console.timeEnd('new')

</code></pre>
</details>

```text
first run
new: 517.260ms
old: 594.656ms

second run
old: 332.977ms
new: 186.684ms
```


-----

Run against this 18M js file from [here](https://github.com/prettier/prettier/issues/5004#issuecomment-459876800)


<details>
[...]

PR #8508, opened 2020-06-06
Quote and unquote number keys

lydell:
Closes https://github.com/prettier/prettier/issues/6064
Closes https://github.com/prettier/prettier/pull/6132

Note: Flow parses number keys, but errors on them during type checking, so we don’t unquote number keys for the `flow` and `babel-flow` parsers.

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzAA8wBeMAb1TDAAYAuMARlQF8g


src/language-js/utils.js
```diff
@@ -734,18 +734,20 @@ function returnArgumentHasLeadingComment(options, argument) {
...
-      node.type === "ClassProperty"
-    )
+    ((isIdentifierName(node.key.value) &&
+      // With `--strictPropertyInitialization`, TS treats properties with quoted names differently than unquoted ones.
+      // See https://github.com/microsoft/TypeScript/pull/20075
+      !(
+        (options.parser === "typescript" || options.parser === "babel-ts") &&
+        node.type === "ClassProperty"
+      )) ||
+      (String(Number(node.key.value)) === node.key.value &&
```
thorn0, line 749: Negative numbers shouldn't be unquoted. `a={-1:0}` is a syntax error.

thorn0:
They also must not be quoted or unquoted in obect types:
**Prettier pr-8508**
[Playground link](https://deploy-preview-8508--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEA3AhgJwAQDMIRLbDYA6IATOUQEYEA2c6U2AvqVCADQgQAOMAJbQAzslBZMEAO4AFLAjEp09aegCeYnjUzowAazgwAynz2CoAc2QxMAVzg84AWxpwAJu48AZZpbvolnAAYhCYzugwQlbIIOh2MBDcIAAWMM70AOopgvAiZmBwxoq5gqi56rFgIlogFiJwmDCyupYRyLgqDTwAViIAHgBCugZGxujOcN4WcB1djiB9-cYWlowAinYQ8HP03SBmmA2YsTDqfHAiYJiCAsl8N7CZgu4wKcgAHAAMPA8QDZldHxYg9Lo1ULMeABHLbwFr8JRxEQAWigcA8HmSmDgMME2JagXaSE6ewWDWcghs9jJqw2sNmxPmPBg6Boz1e7yQFGZukE9FWAGEIM4iSBLgBWZJ2BoAFVZShJ+1QDgAklAvLBjNdbjAAILq4xnRi7BqsVhAA)
```sh
--parser typescript
```

**Input:**
```tsx
var foo: { "2": boolean }

```

**Output:**
```tsx
var foo: { 2: boolean };

```

thorn0:
Hm. I might be wrong about object types...

lydell:
Yeah, as far as I can tell TypeScript doesn’t care about quotes in object types

https://www.typescriptlang.org/play/#code/C4TwDgpgBAglC8UDeUBEAmVAuKAjA9vgDYQCGAdlAL4BQNokUAQgslOjgcWZbTQMb5yAZ2BRSOOIhQcowAE4BXaLUEixuHC2lpMOAGakiwlUA

thorn0:
Not always, but sometimes it does:

https://www.typescriptlang.org/play/#code/DYUwLgBAhhC8EG8IEYBcKA0EBM7sQF8AoUSAIzkQgHJlr1ktrt6dCiiATEAY2CgBOICKWhR0AaxABPAPYAzCGGkAHEAugBuLr35CR4CGTKSZG5Wo1ltRKDHjHtQA

So it's not safe to do in TS at all. Even in object literals.

thorn0:
BigInts can be used as keys too: `{ 2n: 1 }` gives the same result as `{ '2': 1 }`.

lydell:
I wonder if the default "babel" parser is starting to become dangerous … it supports some type syntax, but is not entirely safe to run on Flow and TypeScript code 🤔

thorn0:
There is an issue about this https://github.com/prettier/prettier/issues/8204

src/language-js/printer-estree.js
```diff
@@ -3733,33 +3734,53 @@ function printPropertyKey(path, options, print) {
...
         prop.key &&
-        isStringLiteral(prop.key) &&
-        !isStringPropSafeToCoerceToIdentifier(prop, options)
+        ((isStringLiteral(prop.key) &&
+          !isStringPropSafeToUnquote(prop, options)) ||
+          (isNumericLiteral(prop.key) &&
+            (options.parser === "flow" ||
+              options.parser === "babel-flow" ||
+              options.parser === "typescript" ||
+              options.parser === "babel-ts")))
```
thorn0, line 3743: This condition looks wrong.

**Prettier pr-8508**
[Playground link](https://deploy-preview-8508--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBDAvMAjE1AaVXAXxDxAgAcYBLaAZ2VFQCdmIB3ABRYQZVQA27VAE8GZAEbNUYANZwYAZQozqUAObIYzAK5wycALYS4AE1NmAMqg07U6uADEIzQ6hg0NyNDpgRSIAAWMIYCAOqB1PB0KmBwirxR1ABuUSLeYHTiIGp0cMwwnNLqbsgAZoJ5ZABWdAAeAELScgqKqIZwlmpw5ZX6ILV1imrqAnAAijoQ8L0CVSAqzHnM3jAiFHB0YMzUVAEUO7Bh1KYwgcgAHAAMZAcQeWHSFN4Hm-nJPWQAjlPwRZR8cD0ah0eCIMjMOA-aiQor2UpICpzfp5QzULS6FEjMaTaY9RF9MgwVASY6nc5IABMROk1AEIwAwhBDAiQJsAKwBHR5AAqJL4SPmyT0AEkoBZYIptrsYABBcWKNZjWZ5IhEIA)
```sh
--parser typescript
--quote-props consistent
```

**Input:**
```tsx
a={1:a,a:a}
```

**Output:**
```tsx
a = { "1": a, "a": a };

```

src/language-js/printer-estree.js
```diff
@@ -3734,32 +3735,47 @@ function printPropertyKey(path, options, print) {
...
     // a -> "a"
-    const prop = printString(JSON.stringify(key.name), options);
+    // 1e2 -> "100"
+    // 2n -> "2"
+    const prop = printString(
+      JSON.stringify(
+        key.type === "Identifier" ? key.name : key.value.toString()
+      ),
+      options
+    );
```
thorn0, line 3758: Wow. Surprisingly, this behavior is correct:

**Prettier pr-8508**
[Playground link](https://deploy-preview-8508--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEB5ARgKzmGA6AazgE8BnACmAHIBzAWiqQBYAaABgA90B9dJADn7suvXlCQBmKSwA6UAATzOAQ3ETZCpegCMbPfoNtuhk4e3bxTAL4BKOSBYgIABxgBLaKWShlAJ18QAO4ACn4IXijKADaBymQOIOi+ymBEMADKziluUDTIML4ArnCOcAC26HAAJlXVADKqNIXKNHAAYhC+Zcow7rnIIMqFMBAJABYwZVEA6mNu8KRZYHDp4fNuAG7zxANgpF6OOaRwvjDByTTdyABm0ceOmKQcAELJqXAZymVwdTlwN3cSiBHhx0jkaFE4ABFQoQeAAqL3EBZXzHXwDdDKSpRBLOXw5GDTNxVGBjZD8NiOPEQY7TZLOAZ4uBojb-RwAR1h8HOLgi4E8blI8EQjl8cE5bjF5xaVyQt0RQOOZTc+SKivBkJhcP+csBjhgWKJJLJSAATPrkm4ouCAMIQMqykDMgCsCUKxwAKliIvKkRtigBJKC1WDpMD41wAQWD6RgxEhCOOVisQA)
```sh
--parser babel
--quote-props consistent
```

**Input:**
```jsx
Object.keys({'g-':4,0xb_b:88,0xb_b_bn:333,
  0xan:3,
  0b100000000000_000000000000000011n:4})

```

**Output:**
```jsx
Object.keys({
  "g-": 4,
  "187": 88,
  "0xbbb": 333,
  "0xa": 3,
  "0b100000000000000000000000000011": 4,
});

```

thorn0: Please add this to the tests.

lydell:
Quoting and unquoting can be a little bit “destructive” (but still correct).

```js
x = {
  0b10: null,
};
```

edit ->

```js
x = {
  0b10: null,
  "something-else": null,
};
```

prettier ->

```js
x = {
  "2": null,
  "something-else": null,
};
```

edit ->

```js
x = {
  "2": null,
};
```

prettier ->

```js
x = {
  2: null,
};
```

It’s like Prettier has turned `0b10` into `2`!

Maybe we should only quote/unquote `^\d+$` 🤔 
Or the above case is rare and fine…

thorn0:
Quoting happens only if the user opts in for it using `--quote-props consistent`, so I think it's fine. As for unquoting, I think these cases are confusing and probably shouldn't be unescaped:
**Prettier pr-8508**
[Playground link](https://deploy-preview-8508--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBDAvMA5ARjgahwAYiskcAaLAJjgFpizqBfECkCABxgEtoBnZKFQAnERADuABVEJBKVABsJqAJ6D2AIxGowAazgwAyp108oAc2QwRAVzjs4AW01wAJm-cAZVJduoLOAAxCBEnVBheS2Q0WxgINhAACxgnRQB1JJ54flMwOCM5bJ4AN2zVGLB+DRBzfjgRGCkdC3DkADMlevYAK34ADwAhHX1DI1QnOC9zOA6uhxA+-qNzC0U4AEVbCHg5xW6QUxF6kRjNVFdFRM4Rcxh0njcYJOQADiJ2G4h69J1OGJucBOJVm7AAjtt4M0uPI0Pw6FA4O53IkRHAITw0c0Am0kJ19gt6k4eNY7ITVustjtZnj5uwYBcHk8XkhqPSdDxFKsAMIQJy4kBAgCsiVs9QAKhd5PiDiV7ABJKCeWBGMC3bgAQSVRhgqnWe3qzGYQA)
```sh
--parser babel
```

**Input:**
```jsx
a={'1e+100':1,'2e-10':2}
```

**Output:**
```jsx
a = { 1e100: 1, 2e-10: 2 };

```

lydell:
But I think it’s important that Prettier is “round-trippable”. An addition followed by a deletion should end up with the input code.

```js
a = { 1e100: 1, 2e-10: 2 };
```

edit ->

```js
a = { 1e100: 1, 2e-10: 2, "a-a": 3 };
```

prettier ->

```js
a = { "1e+100": 1, "2e-10": 2, "a-a": 3 };
```

edit ->

```js
a = { "1e+100": 1, "2e-10": 2 };
```

prettier – here we need to unqoute to get back again! ->

```js
a = { 1e100: 1, 2e-10: 2 };
```

But that has the unfortunate side effect of `"1e+100"` being unquoted for `as-needed` users as well.

That’s why I was thinking we should only do stuff with “digits-only” (maybe also floats?) keys – I’m thinking that’s the only likely case we’ll encounter in real code.

src/language-js/printer-estree.js
lydell: You mean this is a bug, right? Because these are the expected keys:

```
> Object.keys({'g-':4,0xb_b:88,0xb_b_bn:333,0xan:3,0b100000000000_000000000000000011n:4})
[ '10', '187', '3003', '536870915', 'g-' ]
```

lydell: Fixed and test added.

thorn0: Where did you copy this from? In my Chrome 83.0.4103.97:
```console
Object.keys({'g-':4,0xb_b:88,0xb_b_bn:333,0xan:3,0b100000000000_000000000000000011n:4})
(5) ["187", "g-", "0xbbb", "0xa", "0b100000000000000000000000000011"]
```

lydell:
Given everything we’ve learned in this PR, it could also make sense to leave things the way the are and close the issue, leaving it up to the user to decide, or to use the ESLint quote-props rule.

src/language-js/printer-estree.js
lydell: Node.js 14.3.0 and Firefox 77. Chrome gives the same result as for you on my machine as well.

lydell: I believe this is a bug in Chrome. It knows how to do `.toString()`, but does not seem to use it for object keys:

```
> 0b10n.toString()
"2"
> ({0b10n: 1})
{0b10: 1}
```

lydell: I asked on Twitter: https://twitter.com/SimonLydell/status/1269565993652715520

thorn0: I can't find anything about bigint keys in the spec: https://tc39.es/ecma262/#prod-LiteralPropertyName

thorn0: So, I read the spec. Definitely a bug in Chrome.

thorn0:
All considered, I vote for "digits only, including floats". Let's play it safe and keep non-decimals, number separators, bigints, etc. as is.

lydell:
Updated to only touch “simple” numbers.

tests/flow/quote-props/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -220,43 +220,78 @@ const d = {
   'd-2': "d2"
 };
 
+// None of these should become quoted, regardless of the quoteProps value.
 const e = {
+  NaN: null,
   1: null,
+  1.5: null,
+  1.0: null,
```
fisker, line 228: Please test `1.` too

src/language-js/utils.js
```diff
@@ -734,21 +734,53 @@ function returnArgumentHasLeadingComment(options, argument) {
...
+        node.type === "ClassProperty"
+      )) ||
+      (isSimpleNumber(node.key.value) &&
+        String(Number(node.key.value)) === node.key.value &&
+        !(
+          options.parser === "flow" ||
+          options.parser === "babel-flow" ||
+          options.parser === "typescript" ||
+          options.parser === "babel-ts"
+        )))
```
thorn0, line 775: Let's explicitly enable it only for `babel`.
Angular can't parse `{ 2: false }` in expressions.
https://stackblitz.com/edit/angular-ivy-qmpjvw?file=src%2Fapp%2Fhello.component.ts

thorn0: Vue supports it though. So alternatively, we need to disable it for `options.parser.startsWith("__ng_")`.

tests/typescript/quote-props/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`types.ts - {"quoteProps":"as-needed"} format 1`] = `
 ====================================options=====================================
-parsers: ["typescript"]
+parsers: ["typescript", "babel-ts"]
```
thorn0, line 5: These changes aren't needed. If a test has `typescript` in `parsers`, it's [automatically verified](https://github.com/prettier/prettier/blob/5d3c137e1bbaec6df50b49d8764ecc6f80977715/tests_config/run_spec.js#L208) against `babel-ts`.

fisker:
> Note: Flow parses number keys, but errors on them during type checking, so we don’t unquote number keys for the flow and babel-flow parsers.

But, `parser=babel` supports `flow`.

thorn0:
That's fine. The `babel` parser isn't safe for Flow anyway: https://prettier.io/blog/2019/01/20/1.16.0.html#flow
I agree that it's a mess. BTW, probably this means that if we proceed with #8024 (parse Flow syntax only if the pragma is present), it won't really be a breaking change as we'd break something that is already broken.

brodybits:
> we don’t unquote number keys for the `flow` and `babel-flow` parsers

Quote & unquote with TypeScript seems to be considered "not safe" as well, as was noted in the comments.

PR #8528, opened 2020-06-08
[WIP] Add option to prevent mutating the AST during printing

siefkenj:
Attempts to address #8122 an allow for printing of an AST without mutating it. Mutation seems to only happen when dealing with comments (the `.comments` property of the root AST is deleted, and
comments are attached to nodes; `.comments` properties are created on nodes if they don't exist).

This PR, instead of mutating the AST, keeps a parallel cache of modifications via a Javascript `Map()` object. When an AST node is requested via `FastPath`, the cache is consulted. If there are cached comments, a new object is constructed containing the contents of the original node and the cached contents. In this way, the original node is never touched.

A new option `preserveAst` is added, which defaults to `false`, to enable the new behavior. This is in case third party plugins rely on Prettier mutating the AST.

This is a WIP because I wanted to see if the community likes this approach before investing more time. It *mostly* works, however, there are some issues I am not sure how to track down.

For example, the code
```
import { //comment
x } from "";
```
will format as itself when `preserveAst === true` but should format as
```
import {
   //comment
   x
} from "";
```
I cannot figure out why/where this is happening. The output from `comments.printComments` is identical with and without `preserveAst`.

### DIFF
--- a/src/common/fast-path.js
+++ b/src/common/fast-path.js
@@ -17,8 +17,19 @@ function getNodeStackIndexHelper(stack, count) {
 }
 
 class FastPath {
-  constructor(value) {
-    this.stack = [value];
+  constructor(value, augmenter) {
+    this.augmenter = augmenter;
+    this.stack = [this._augmentNode(value)];
+  }
+
+  // Sometimes there may be additional information associated with a node
+  // that isn't stored in the actual AST. This function returns a modified
+  // node with extra information (if available).
+  _augmentNode(value) {
+    if (this.augmenter) {
+      return this.augmenter(value);
+    }
+    return value;
   }
 
   // The name of the current property is always the penultimate element of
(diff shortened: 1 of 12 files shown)


fisker:
As I  understand, we are not going to add this option, and we are not going to modify `FastPath`.

What we need to do is change `node.comments` assignment to `map.set(node, comment)`, and change `const foot = node.comments` access to `const  foo = map.get(node)`, hope I didn't get it wrong.

siefkenj:
@fisker Currently printing of comments is done via `path.call(..., "comments")`. If nodes are no longer decorated with a `.comments` property, how will that work (without a modification of fastpath)?

thorn0:
I didn't really look at the code yet, but a toggle to switch between the new behavior and the existing behavior should be on the plugin level (a property of the printer most probably, with `undefined` meaning "old behavior"), not a global option. This way existing plugins won't get broken.

PR #8534, opened 2020-06-09
Fix encoding in postcss-parser build

sibiraj-s:

Fix encoding in postcss-parser build
The encoding in parser-postcss build makes it not possible to load it inside a chrome extension. 

Related issues:

- Fixes: https://github.com/prettier/prettier-chrome-extension/issues/203
- https://github.com/webpack-contrib/terser-webpack-plugin/issues/107

### DIFF
--- a/scripts/build/build.js
+++ b/scripts/build/build.js
@@ -47,7 +47,7 @@ async function createBundle(bundleConfig, cache) {
     // `prettier-chrome-extension` https://github.com/prettier/prettier-chrome-extension
     // details https://github.com/prettier/prettier/pull/8534
     if (target === "universal") {
-      const file = path.join("dist", bundleConfig.output);
+      const file = path.join("dist", output);
       const content = fs.readFileSync(file, "utf8");
       if (content.includes("\ufffe")) {
         throw new Error("Bundled umd file should not has U+FFFE character.");
--- a/scripts/build/build.js
+++ b/scripts/build/build.js
@@ -4,6 +4,7 @@ const chalk = require("chalk");
 const execa = require("execa");
 const minimist = require("minimist");
 const path = require("path");
+const fs = require("fs");
 const stringWidth = require("string-width");
 
 const bundler = require("./bundler");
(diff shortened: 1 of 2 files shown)


thorn0:
@sibiraj-s Could you please first isolate the issue that this PR is supposed to fix? What is the simplest way to reproduce it?

sibiraj-s:
the encoding in `parser-postcss` makes it not possible to load it in the chrome extension.

the simplest way to reproduce this issue would be 

1. clone the repo https://github.com/prettier/prettier-chrome-extension
2. run `yarn start`
3. go to `chrome://extensions`
4. enable developer mode
5. load unpacked -> select `extension` folder

utf-8 eror

fisker:
I think it's okay to add this option, but need add comment explain why we need it, so we won't lost it in future.

thorn0:
I'm sure PostCSS isn't the only source of non-ASCII characters in the bundle. Are we sure this isn't something that can be fixed on the extension's side? Isn't there a way to specify the encoding for scripts (something like `<script charset="utf-8"...>` for script tags)?

sibiraj-s:
I checked the file encoding it is already in UTF-8. it seems just that chrome couldn't parse the contents. the scripts are loaded directly through reference in manifest.json so there is no way place to specify charset manually too. I searched every other possible ways to fix it in the extension side but there seems to be none. 

the only thing could be done is enabling terser in both dev and prod mode which terribly makes the build slower.

sibiraj-s:
> I'm sure PostCSS isn't the only source of non-ASCII characters in the bundle

I didn't dig so deep into this. but i verified all other bundles, the error is caused by the `buffer` module bundled inside the parser-postcss file. it doesn't exist in any other parsers or build files.

fisker:
@thorn0 You are right, search `let g="` in [this file](https://prettier.io/lib/parser-babel.js) lots of non-ASCII characters.

@sibiraj-s Maybe because of the character `U+FFFE` in this file?

sibiraj-s:
> @sibiraj-s Maybe because of the character U+FFFE in this file?

I am not completely sure. I have to dig little deeper

thorn0:
Good point about `U+FFFE`. If it's the culprit, the issue has already been fixed by #8266.

sibiraj-s:
I also checked the same build with rollup, it is working fine (not parser functionality just file contents), the problem seems to be with the webpack build alone.

sibiraj-s:
I will try once with this branch https://github.com/thorn0/prettier/tree/css-value-loc and check

fisker:
> Good point about `U+FFFE`. If it's the culprit, the issue has already been fixed by #8266.

That's not true, you can build on your machine or download artifacts from [this job](https://github.com/prettier/prettier/runs/757497633) search `this.css[0]` there is still `U+FFFE` there.

thorn0:
It's been merged to `master`. Does `master` cause the same trouble as 2.0.5?

sibiraj-sr:
> It's been merged to `master`. Does `master` cause the same trouble as 2.0.5?

yes. I just checked with master. the problem still exists. 

Also, as I mentioned

>I also checked the same build with rollup, it is working fine (not parser functionality just file contents), the problem seems to be with the webpack build alone.

I am checking if there is a difference in config between webpack and rollup. rollup seems to be handling this by default.

thorn0:
@fisker I see, so there is still another instance of U+FFFE, even in master.

@sibiraj-s Can you please try manually removing U+FFFE from the bundle (replace it with something else) and confirm whether it is the culprit?

sibiraj-sr:
> @fisker I see, so there is still another instance of U+FFFE, even in master.
> 
> @sibiraj-s Can you please try manually removing U+FFFE from the bundle (replace it with something else) and confirm whether it is the culprit?

Will check.

sibiraj-s:
@thorn0 I searched, I couldn't find anything. I am pretty much new to this codebase. can you point me incase I missed it?

thorn0:
@sibiraj-s `node_modules/postcss/lib/input.js`

fisker:
I added a check on umd files, failed log https://github.com/prettier/prettier/runs/760104476#step:6:16

PR #8543, opened 2020-06-11
Migrate to native dependabot

fisker:
Ref: https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
Syntax: https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#about-the-dependabotyml-file

Adding this file will enable [this page](https://github.com/prettier/prettier/network/updates)

image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
 version: 2
-update_configs:
+updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 10
 
   - package-ecosystem: "npm"
     directory: "/website"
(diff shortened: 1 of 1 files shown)


evilebottnawi:
Why? What is wrong with default configuration?

fisker:
This is not config for the old dependabot, it's migrate to native github  bot, we don't need go to dependabot.com, there will be logs and `update` button on the github page I post in the description. Consider there are two bots have similar names.

fisker:
I read the docs again, about [`allow`](https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#allow), should we only check direct dependencies? There might too many indirect dependencies PR.

evilebottnawi:
Updating `transient dependencies` is good practice, because you can have regressions in vendor packages, very few projects uses it, but I found it very good and useful, it is my opinion

fisker:
Just afraid it  will be  annoying. Let's keep  it  this  way  and  see?

evilebottnawi:
Yep, let's keep it this way, anyway will be great to hear other opinion here too

fisker:
Checkout this page https://github.com/prettier/prettier/network/updates

Issue #8409, opened 2020-05-26
CSS @supports selector adds unwanted space inside pseudo-selectors


runarberg:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAzgVwA7YgJxnQAJ04AbOMGAgCiQDMIxN0BaANwEt0uAjSgEpiwADpRixPphg0ojZqxHjJkiDPJcocJMSjQ4AbhXEAviZPTZ0BS3bdeAuMomr1MTdt0AmbAA9SCE0AE2ICAEMoAHMjE3MoUxAAGhAIbBguaHRkUHD8fAgAdwAFPIRslHDyQvCAT2yUvnxwsABrOBgAZWwWrSjkGHxMOBSACxgAW3IAdVGueHQesDhO8vmubhha5HB0BpAtMkJi5qiJ8OQGKrIUgCt0PwAhZraOzvCJuAAZLThL65GIHufk6fUoAEVMBB4P9yDcQD18EcdmA9skEfgtDBplxgjBRsgABwABhS2AKZGmzWwO3JcCOHD+KQAjlD4Cc0hUQOF2No4MF+ej8HBWVxhSdwmcLowASkyBMuAMhoDeNEIWy-jK4YCYOE+Di8QSkN4UoNwlxPFEAMIQCbnHb0gCs6NYcAAKnqKldtaZTEA)
```sh
--parser css
```

**Input:**
```css
@supports selector(:focus-visible) {
  button:focus {
    outline: none;
  }
  
  button:focus-visible {
    outline: 2px solid orange;
  }
}
```

**Output:**
Notice the space between the colon (`:`) and the `focus-visible`, making the pseudo-selector invalid.
```css
@supports selector(: focus-visible) {
  button:focus {
    outline: none;
  }

  button:focus-visible {
    outline: 2px solid orange;
  }
}
```

**Expected behavior:**
No space between the colon and the pseudo selector name.
```css
@supports selector(:focus-visible) {
  button:focus {
    outline: none;
  }

  button:focus-visible {
    outline: 2px solid orange;
  }
}
```

teawithlemon:
(I am also getting this on `:nth-child` on this example:
``` &: nth-child(2) {
    border: 3px solid purple;
  }```

Only with `&` though, not if it's by itself.

Presuming it's part of the same problem.

thorn0:
[...]

PR #8545, opened 2020-06-11
Fix: at rule supports selector

boyenn:
Fixes https://github.com/prettier/prettier/issues/8409.

Parses `<complex-selector>` in `@supports selector(<complex-selector>)` as a selector.

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -14,12 +14,12 @@ const {
 } = require("./utils");
 const { calculateLoc, replaceQuotesInInlineComments } = require("./loc");
 
-function getHighestAncestor(node) {
-  if (!node.parent) {
-    return node;
+const getHighestAncestor = node => {
+  while (node.parent) {
+    node = node.parent;
   }
-  return getHighestAncestor(node.parent);
-}
+  return node;
+};
 
 function parseValueNode(valueNode, options) {
   const { nodes } = valueNode;
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -14,7 +14,15 @@ const {
 } = require("./utils");
 const { calculateLoc, replaceQuotesInInlineComments } = require("./loc");
 
-function parseValueNodes(nodes, options) {
+function getHighestAncestor(node) {
+  if (!node.parent) {
+    return node;
+  }
+  return getHighestAncestor(node.parent);
+}
+
+function parseValueNode(valueNode, options) {
+  const { nodes } = valueNode;
   let parenGroup = {
     open: null,
     close: null,
@@ -43,6 +51,17 @@ function parseValueNodes(nodes, options) {
       // For example, 50px... is parsed as `50` with unit `px...` already by postcss-values-parser.
       node.value = node.value.slice(0, -1);
       node.unit = "...";
+    }
+
+    if (node.type === "func" && node.value === "selector") {
+      node.group.groups = [
+        parseSelector(
+          getHighestAncestor(valueNode).text.slice(
+            node.group.open.sourceIndex + 1,
+            node.group.close.sourceIndex
+          )
+        ),
+      ];
     }
 
     if (node.type === "func" && node.value === "url") {
@@ -138,13 +157,16 @@ function flattenGroups(node) {
   return node;
 }
 
-function addTypePrefix(node, prefix) {
+function addTypePrefix(node, prefix, skipPrefix) {
   if (node && typeof node === "object") {
     delete node.parent;
     for (const key in node) {
-      addTypePrefix(node[key], prefix);
+      addTypePrefix(node[key], prefix, skipPrefix);
       if (key === "type" && typeof node[key] === "string") {
-        if (!node[key].startsWith(prefix)) {
+        if (
+          !node[key].startsWith(prefix) &&
+          (!skipPrefix || !skipPrefix.test(node[key]))
+        ) {
           node[key] = prefix + node[key];
         }
       }
@@ -168,14 +190,16 @@ function addMissingType(node) {
 
 function parseNestedValue(node, options) {
   if (node && typeof node === "object") {
-    delete node.parent;
     for (const key in node) {
-      parseNestedValue(node[key], options);
-      if (key === "nodes") {
-        node.group = flattenGroups(parseValueNodes(node[key], options));
-        delete node[key];
+      if (key !== "parent") {
+        parseNestedValue(node[key]), options;
+        if (key === "nodes") {
+          node.group = flattenGroups(parseValueNode(node, options));
+          delete node[key];
+        }
       }
     }
+    delete node.parent;
   }
   return node;
 }
@@ -198,7 +222,7 @@ function parseValue(value, options) {
 
   const parsedResult = parseNestedValue(result, options);
 
-  return addTypePrefix(parsedResult, "value-");
+  return addTypePrefix(parsedResult, "value-", /^selector-/);
 }
 
 function parseSelector(selector) {
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -192,7 +192,7 @@ function parseNestedValue(node, options) {
   if (node && typeof node === "object") {
     for (const key in node) {
       if (key !== "parent") {
-        parseNestedValue(node[key]), options;
+        parseNestedValue(node[key], options);
         if (key === "nodes") {
           node.group = flattenGroups(parseValueNode(node, options));
           delete node[key];
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -14,7 +14,7 @@ const {
 } = require("./utils");
 const { calculateLoc, replaceQuotesInInlineComments } = require("./loc");
 
-const getHighestAncestor = node => {
+const getHighestAncestor = (node) => {
   while (node.parent) {
     node = node.parent;
   }
--- a/tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
@@ -4283,6 +4283,17 @@ B
 @supports selector(selector(:focus-visible)) {}
 @supports selector(:::selector(:focus-visible)) {}
 @supports selector(--selector(:focus-visible)) {}
+@supports not (transform-origin: 10em 10em 10em) {}
+@supports not (not (transform-origin: 2px)) {}
+@supports (display: grid) and (not (display: inline-grid)) {}
+@supports (display: table-cell) and (display: list-item) and (display:run-in) {}
+@supports (display: table-cell) and ((display: list-item) and (display:run-in)) {}
+@supports (transform-style: preserve) or (-moz-transform-style: preserve) {}
+@supports (transform-style: preserve) or (-moz-transform-style: preserve) or
+          (-o-transform-style: preserve) or (-webkit-transform-style: preserve) {}
+
+@supports (transform-style: preserve-3d) or ((-moz-transform-style: preserve-3d) or
+          ((-o-transform-style: preserve-3d) or (-webkit-transform-style: preserve-3d))) {}
 
 =====================================output=====================================
 @supports (transform-origin: 5% 5%) {
@@ -4534,6 +4545,31 @@ B
 @supports selector(:::selector(:focus-visible)) {
 }
 @supports selector(--selector(:focus-visible)) {
+}
+@supports not (transform-origin: 10em 10em 10em) {
+}
+@supports not (not (transform-origin: 2px)) {
+}
+@supports (display: grid) and (not (display: inline-grid)) {
+}
+@supports (display: table-cell) and (display: list-item) and (display: run-in) {
+}
+@supports (display: table-cell) and ((display: list-item) and (display: run-in)) {
+}
+@supports (transform-style: preserve) or (-moz-transform-style: preserve) {
+}
+@supports (transform-style: preserve) or (-moz-transform-style: preserve) or
+  (-o-transform-style: preserve) or (-webkit-transform-style: preserve) {
+}
+
+@supports (transform-style: preserve-3d) or
+  (
+    (-moz-transform-style: preserve-3d) or
+      (
+        (-o-transform-style: preserve-3d) or
+          (-webkit-transform-style: preserve-3d)
+      )
+  ) {
 }
 
 ================================================================================
--- a/tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
@@ -4195,6 +4195,92 @@ run-in
 @supports not((text-align-last:justify)and(-moz-text-align-last:justify)){ }
 @supports (--foo:green) {}
 
+@supports selector(:after) {}
+@supports selector(:focus) {}
+@supports selector(:focus-visible) {}
+@supports selector(A>B) {}
+@supports selector(A       >      B) {}
+@supports selector(A       >
+     B) {}
+@supports selector(A
+>
+B
+) {}
+@supports selector(
+A
+<
+B
+) {}
+@supports selector(.a:not(.b)){}
+@supports selector(:not(.b)){}
+@supports not selector(:after){
+}
+@supports (--foo:green) and selector(:after){
+}
+@supports selector(:before) and selector(:focus-visible){
+}
+@supports selector(:not(.a,.b)) and selector(:focus-visible){
+}
+@supports selector(.a,.b){
+
+}
+@supports selector(){
+
+}
+
+@supports selector(:focus-visible) {
+  button:focus {
+    outline: none;
+  }
+
+  button:focus-visible {
+    outline: 2px solid orange;
+  }
+}
+
+@supports selector(:not(.a,.b)) {
+}
+
+@supports selector(:not(.a,.b), :not(.b,.c)) {
+}
+@supports selector(:not(.a,.b), .b, .c) {
+}
+
+@supports selector(
+  :not(
+  .a,.b)
+  ) {
+}
+@supports selector(
+  :not(
+  .a,.b
+  )
+  ) {
+}
+@supports selector(.asdasldaskdhjkashdahsdkjahskdjhakjsdkjahsdhkas .asdasldaskdhjkashdahsdkjahskdjhakjsdkjahsdhkas,
+.asdasldaskdhjkashdahsdkjahskdjhakjsdkjahsdhkas .asdasldaskdhjkashdahsdkjahskdjhakjsdkjahsdhkas){
+}
+@supports selector(.parent>.child ){
+}
+@supports selector(> .child-one){
+}
+@supports selector(.parent  ~  .child){
+}
+@supports selector(ns|* ){
+}
+@supports selector(svg|a ){
+}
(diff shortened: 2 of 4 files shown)


src/language-css/parser-postcss.js
```diff
@@ -474,6 +474,21 @@ function parseNestedCSS(node, options) {
 
         return node;
       }
+      if (name === "supports" && /^selector\(.*\)$/s.test(params)) {
+
+        // FIXME: We should instead only parse <complex-selector> as a selector
```
boyenn, line 479: The wall I ran into

fisker:
Maybe you can try check `type === "value-func" && value === "selector"`, so `selector(.a)` still parse as `value-root`, then parse `.a` as  selector.

boyenn:
> Where is `selector()` CSS spec?

draft: 
https://drafts.csswg.org/css-conditional-4/#typedef-supports-selector-fn

Does prettier even want to supports drafts actually ?

evilebottnawi:
I don't think we should support this draft, just my opinion, do you have postcss-plugin or something else for this stuff, looks it is not work in browsers right now

boyenn:
https://jsfiddle.net/umn21839/13/

works in chrome (Version 83.0.4103.61 (Official Build) (64-bit))

evilebottnawi:
@boyenn sounds good, let's fix it

boyenn:
> Maybe you can try check type === "value-func" && value === "selector", so selector(.a) still parse as value-root, then parse .a as selector.

That seems to kind of work, except it means that all the node types get prefixed with `value-`, since they are parsed as a value.

So it would introduce changes like 
```
case "value-selector-pseudo": // new
    case "selector-pseudo": {
      return concat([
        maybeToLowerCase(node.value),
        node.nodes && node.nodes.length > 0
          ? concat(["(", join(", ", path.map(print, "nodes")), ")"])
          : "",
      ]);
    }
```

all over the place in the `printer`. Is that okay ? Or is there a way to avoid that?

fisker:
No, only pass `.a:focus` to `parseSelector` and over write the `value-func.group`, there won't be `value-` prefix.

boyenn:
@fisker I might be misunderstanding you then.

I tried https://github.com/prettier/prettier/pull/8545/commits/246d0b4c6b86461ea5cf24519965dee656e6c3f2#diff-29bcb0b461c4ae51930e975e53edfdd9R35 , and then the `value-` prefix gets added.

I did it in that function ( parseValue ) since `value-func` doesn't pop up in parseNestedCSS, am I just doing it in the wrong place?

fisker:
I'm not  sure  if  it  will  work , but  I  mean don't check `at-rule`,  check  `node.type === "value-func"` together with `at-rule`  not  inside.

src/language-css/parser-postcss.js
```diff
@@ -443,6 +448,7 @@ function parseNestedCSS(node, options) {
         }
       }
     }
+    console.log(node.type);
```
fisker, line 452:
```suggestion
if (node.type === "value-func" && node.value === "selector") {
```

something like  this. Will this  work?

boyenn: No, node.type is never value-func here.

But I have found another way, will update in a sec.

tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4392,6 +4397,11 @@ B
 }
 @supports selector(:not(.b)) {
 }
+@supports not selector(: after) {
```
boyenn, line 4400: Auch, indeed.

fisker: If `func` can only checked in parseValue, why not try add a  exception, not add `value-` prefix.

boyenn: if I understood you correctly, I did that now :)

fisker: Seems working? Maybe we  don't need that flag, how about check `type` starts with `selector-`? Thoughts?

boyenn: `selector-selector` exists, this would break that.

( assuming you mean 
```
 if (
            !node[key].startsWith(prefix) &&
            !node[key].startsWith("selector")
          ) {
```
)

boyenn: Alternative solution: passing a regex to `addTypePrefix` to test and skip adding prefix if matches.


Like so : 
```
function addTypePrefix(node, prefix, skipPrefix) {
  if (node && typeof node === "object") {
    delete node.parent;
    for (const key in node) {
      addTypePrefix(node[key], prefix, skipPrefix);
      if (key === "type" && typeof node[key] === "string") {
        if (
          !node[key].startsWith(prefix) &&
          (!skipPrefix || !skipPrefix.test(node[key]))
        ) {
          node[key] = prefix + node[key];
        }
      }
    }
  }
  return node;
}
```

called like
```
  return addTypePrefix(parsedResult, "value-", /selector/);
```

in parseValue


I don't have a strong opinion which way is better, so I'd like to know if you have a preference either way.

src/language-css/parser-postcss.js
```diff
@@ -32,6 +32,12 @@ function parseValueNodes(nodes) {
   for (let i = 0; i < nodes.length; ++i) {
     const node = nodes[i];
 
+    if (node.type === "func" && node.value === "selector") {
+      const selector = parseSelector(stringifyNode(node.group.groups[0]));
```
fisker, line 36: This groups[0] not right `selector()` `selector(.a,.b)`

boyenn: Argh, yep, fixed.

fisker:
**Prettier pr-8545**
[Playground link](https://deploy-preview-8545--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEABAzgVwA7YgJxnQAJ04AbOMGAgCiSghloB0pjiA6AQwBpOARgEohxYGwC+IXiAjYYAS2jpkobvnwQA7gAV1CFSm7kt3AJ4qZA-NzABrODADK2WwqgBzZDHyY4MuABbATgAE1CwgBluT0xuDzgAMQJA7hhFT2QQbkwaaRAACxhA8gB1AoV4dFcwOCcDSoUAN0qzLLB0SxB3MkIdGw9U5AAzYzIZACt0AA8AIRt7RyduQLhI9zgRsf8QKemndw9KAEVMJk2kUfJxkFd8XvbO-Ox8dxhShVCYAuQADgAGGQvCBkUo2bBZF5wXpNTYyACOZ3g-Tkhmy6AAtFA4GEwvl8HBEQoCf14kNLtsZGRAgpvL4duhDickRcrjcYNwBB8vj8kAAmGQ+bgKciHADCEEC5JA0IArPlMGQACqcwxsnZNPwASSgEVgTjAr3kAEFdU4YGZKFtrnAJBIgA)
```sh
--parser css
```

**Input:**
```css
@supports selector(:not(
  .a,.b)) {
}
```

**Output:**
```css
@supports selector(:not(
  .a,.b)) {
}

```

src/language-css/parser-postcss.js
```diff
@@ -185,7 +194,7 @@ function parseValue(value) {
 
   const parsedResult = parseNestedValue(result);
 
-  return addTypePrefix(parsedResult, "value-");
+  return addTypePrefix(parsedResult, "value-", /selector/);
```
fisker, line 197:
```suggestion
  return addTypePrefix(parsedResult, "value-", /^selector\-/);
```

boyenn:
I don't get that output, I get 
```
@supports selector(:not(
  .a,.b)) {
}
= > 

@supports selector(:not(.a.b)) {
}
```

Which is however also wrong ( missing the comma ). I'll have a look hopefully tonight ( GMT )

fisker:
Try to get the original text.`:not(.a,.b)`

boyenn:
> :not(.a,.b)

That is not what would be given when working without @supports selector(..) : 
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuZUIwBQDoCGAaLAIwEpgAdKAXxDxAgAcYBLaAZ2VBwCcuIB3AArcE7FDgA2fHAE92tQlxxgA1nBgBlekqZQA5shhcArnFoALGAFtxAdTNN4rLWDjqRDpgDcH05OFZyIDqscFwwAoq6ljjIAGYSIbQAVqwAHgBCiipq6jiWcAAyOnBxCaYgKanqOrricACKRuglSPHiiSBaXCFcfmABNJ1cOjA2TAAmMGbIABwADLT0vCE2ivR+S3A9niW0AI5N8BEMoiA4rAC0UHBw47eDXHAHTI8ROFExrWW0IZZMBsZyqwanVGs1Su1yjAcIQxpNpkgAEy0Qw4JjiGoAYQglmifi2AFZBkYQgAVGGiNodTwmACSUDusHUYGGjAAggz1DBpHUISFKJQgA)
```sh
--parser css
```

**Input:**
```css
:not(.a,.b){
}
```

**Output:**
```css
:not(.a, .b) {
}

```

fisker:
Sorry, I mean, try not to parse groups one by one, pass `:not(.a,.b)` to `parseSelector()`.

boyenn:
> Sorry, I mean, try not to parse groups one by one, pass `:not(.a,.b)` to `parseSelector()`.

I see, yeah I tried that shortly, but the comma is already not there at that point. I'm not sure why, but I hope I'll figure it out tonight.

boyenn:
Can't figure out why the comma is already gone at that point. I'm afraid this is on the postcss side :/

fisker:
`options.orignalText` has the original text, can we get the correct range and slice from it?

boyenn:
> `options.orignalText` has the original text, can we get the correct range and slice from it?

Wasn't able to figure out how. It seemed you were referring to printers, there it seems to be often done that way. This is in the parser however.


The original text is also available in the parser, but it's not passed through all the functions and doing so would be a nightmare.

I instead did it only for parseValue, since that is the level at which I need the original text.

That did fix the issue, but it feels wrong so I'd really like your feedback if there is a cleaner way to do this

fisker:
I think it's okay to pass `text` on this function, but I'm not sure about `selector(` part, can space or comment included? like

```css
@support selector/* comment */(foo) {
}
```

or

```css
@support selector (foo) {
}
```

Are they valid syntax?

boyenn:
Neither seem to work on chrome :
https://jsfiddle.net/769nhzxp/7/
In spec it is defined as 

```
<supports-feature> = <supports-selector-fn> | <supports-decl>
<supports-selector-fn> = selector( <complex-selector> )
````
So I assume we're good here.

**Comment by boyenn** [comment]
As far as I know, you can never put comments or spaces in function calls in css, right ?

**Comment by boyenn** [comment]
https://www.w3.org/TR/css-values-4/#functional-notations

>"The syntax starts with the name of the function immediately followed by a left parenthesis"
> "White space is allowed, but optional, immediately inside the parentheses"

I think `selector(` is the only way to call `selector`.

**Comment by fisker** [comment]
Okay, let's keep it this way. But how about test `indexOf("selector(") !== -1` and `rootValueText[rootValueText.length -1] === ")"` , you know just in case?

Another possible problem, in scss a function or mixin can be called `selector`, right? Do we need do something?

**Comment by thorn0** [comment]
@boyenn `value-*` nodes have the `sourceIndex` property. Can you probably use it instead of `indexOf`?

**Comment by boyenn** [comment]
Just reading these comments made me realise something anyway. 
```
@supports selector(:not(.a,.b)) and selector(:focus-visible){
}
```
Will in no way work, as it will parse `:not(.a,.b)) and selector(:focus-visible` as a selector as parameter to both `selector()`'s 
Currently working my dayjob, but tonight/tomorrow I'll fix it ( Holiday starting tonight ).

File: src/language-css/parser-postcss.js [hunk 6]
```diff
@@ -32,6 +32,17 @@ function parseValueNodes(nodes) {
   for (let i = 0; i < nodes.length; ++i) {
     const node = nodes[i];
 
+    if (node.type === "func" && node.value === "selector") {
+      node.group.groups = [
+        parseSelector(
+          rootValueText.slice(
+            rootValueText.indexOf("selector(") + "selector(".length,
+            -1
+          )
```
fisker, line 41: I think https://github.com/prettier/prettier/pull/8545#issuecomment-645276974 can be fixed if we cut the string from `groups first param start` to `groups last param end`.

boyenn:
> I think #8545 (comment) can be fixed, if we cut the string from groups first param start to groups last param end.

I assumed the same, will have a look at that soon.

> Another possible problem, in scss a function or mixin can be called selector, right? Do we need do something?

You're right, this is an edge case but whatever is between the `()` will probably be formatted as a selector. I honestly have no clue what I can possible do about that. `parseValueNodes` has no access to where the value is used. We'd have to mark the nodes as something like "isInSupportsAtrule" in `parseNestedCSS`, but I don't know if that is clean :P

fisker:
Even if we can add this `isInSupportsAtrule`, I think it's still possible

```scss
$string-supports: "supports";
#{$string-supports} selector(::before) {}
```

Bang! Broken again. So let's ignore it? 

I remember we can rewrite `url()/rgb()` too in SCSS, but we don't make it special.

@evilebottnawi What's your opinion about the second question above?

fisker:
@boyenn [playground](https://deploy-preview-8545--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEABAzgVwA7YgJxnQAIoIZiAKdOAGzjBgMqQCM4AzAuASmINLkqZCtToMm+Fuy75ePBcWAAdKAF8QAGhARsMAJbR0yUAEN8+CAHcACuYTGUp2ldMBPY9tb5TYANZwMADK2L76UADmyDD4mHDacAC27AAmKXApADKmkZimEXAAYgSJpjAGkcggpphMWiAAFjCJtADqDfrw6KFgcEEOnfoAbp1uVWDoniDhNIQ2PhGlyBzONNoAVugAHgBCPv6BQaaJcJnhcMur8SCbW0HhEfQAipjkF0grtGsgofiz45N6th8OEYK19CkYA1kAAOAAM2mBEBorR82CqwLgsyGF20AEdXvB5rpHNV0ABaKBwDIZepyAn6OTzfJLD5XbQ0RL6aKxa7oB7PQnvT7fGCmVjgyHQpAAJm0MVM+loDwAwhBEqyQFiAKz1TA0AAq4scIuuQziAEkoOlYEEwCC9ABBa1BGBueiXL5wNRqIA)

boyenn:
> @boyenn [playground](https://deploy-preview-8545--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEABAzgVwA7YgJxnQAIoIZiAKdOAGzjBgMqQCM4AzAuASmINLkqZCtToMm+Fuy75ePBcWAAdKAF8QAGhARsMAJbR0yUAEN8+CAHcACuYTGUp2ldMBPY9tb5TYANZwMADK2L76UADmyDD4mHDacAC27AAmKXApADKmkZimEXAAYgSJpjAGkcggpphMWiAAFjCJtADqDfrw6KFgcEEOnfoAbp1uVWDoniDhNIQ2PhGlyBzONNoAVugAHgBCPv6BQaaJcJnhcMur8SCbW0HhEfQAipjkF0grtGsgofiz45N6th8OEYK19CkYA1kAAOAAM2mBEBorR82CqwLgsyGF20AEdXvB5rpHNV0ABaKBwDIZepyAn6OTzfJLD5XbQ0RL6aKxa7oB7PQnvT7fGCmVjgyHQpAAJm0MVM+loDwAwhBEqyQFiAKz1TA0AAq4scIuuQziAEkoOlYEEwCC9ABBa1BGBueiXL5wNRqIA)

You're on fire! Thanks!

boyenn:
https://github.com/prettier/prettier/pull/8566 should be conflicting with this one ( just merged).
No idea why this pr shows no conflicts, but I'll solve them in a few hours.

src/language-css/parser-postcss.js
```diff
@@ -14,7 +14,15 @@ const {
 } = require("./utils");
 const { calculateLoc, replaceQuotesInInlineComments } = require("./loc");
 
-function parseValueNodes(nodes, options) {
+function getHighestAncestor(node) {
+  if (!node.parent) {
+    return node;
+  }
+  return getHighestAncestor(node.parent);
+}
```
fisker, line 22:
```suggestion
const getHighestAncestor = node => {
  while (node.parent) {
    node = node.parent
  }
  return node;
};
```

fisker:
I think this will be last one

```css
@supports selector(selector(:focus-visible)) {}
```

boyenn:
> ~I think this will be last one~
> 
> ```css
> @supports selector(selector(:focus-visible)) {}
> ```
> 
> Update: Am I seem outdated playground? Seem not wrong, when I post this, it formats to
> 
> ```css
> @supports selector(selector()) {
> }
> ```
> 
> Anyway, let's add these to tests.
> 
> ```css
> @supports selector(selector(:focus-visible)) {}
> @supports selector(:::selector(:focus-visible)) {}
> @supports selector(--selector(:focus-visible)) {}
> ```
> 
> It's a supports test feature, it should support some possible future selectors / unknown selectors.

probably due to the merge failure :) 
Seems to work indeed, thanks for all your help!

boyenn:
> Even if we can add this `isInSupportsAtrule`, I think it's still possible
> 
> ```css-scss
> $string-supports: supports;
> @#{$string-supports} selector(::before) {}
> ```
> 
> Bang! Broken again. So let's ignore it?
> 
> I remember we can rewrite `url()/rgb()` too in SCSS, but we don't make it special.
> 
> @evilebottnawi What's your opinion about the second question above?

@evilebottnawi please advise

src/language-css/parser-postcss.js
```diff
@@ -45,6 +53,17 @@ function parseValueNodes(nodes, options) {
...
+    if (node.type === "func" && node.value === "selector") {
+      node.group.groups = [
+        parseSelector(
+          getHighestAncestor(valueNode).text.slice(
+            node.group.open.sourceIndex + 1,
+            node.group.close.sourceIndex
+          )
+        ),
+      ];
+    }
```
evilebottnawi, line 65: Very bad workaround, but I think we can't fix it on our side, so it is better than nothing

boyenn:
@fisker @evilebottnawi can this be merged or should something still be done?

tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4535,6 +4546,31 @@ B
...
+
+@supports (transform-style: preserve-3d) or
+  (
+    (-moz-transform-style: preserve-3d) or
+      (
+        (-o-transform-style: preserve-3d) or
+          (-webkit-transform-style: preserve-3d)
+      )
+  ) {
+}
```
evilebottnawi, line 4573: Can you open an issue for improve it, no need to fix it in this PR?

fisker:
Replace `getHighestAncestor(valueNode).text` with range of   `highestAncestor` from original text should be  able to fix it.

fisker:
#8704

tests/css/atrule/__snapshots__/jsfmt.spec.js.snap
fisker: #8705

fisker:
Thank you for fixing this.

PR #8546, opened 2020-06-12
Update `ignore` to v5

fisker:
Previous attempt #6975


src/common/get-file-info.js
```diff
@@ -51,9 +53,10 @@ getFileInfo.sync = function (filePath, opts) {
   const ignorer = createIgnorer.sync(opts.ignorePath, opts.withNodeModules);
   return _getFileInfo({
     ignorer,
-    filePath: normalizeFilePath(filePath, opts.ignorePath),
```
fisker, line 54: I think this old logic is not right, if `opts.ignorePath` is not in same dir with `.prettierrc`, the related file path to `ignorePath` is passed to `resolveConfig()`, I'm passing absolute path to `resolveConfig()`, only use relative path for `ignore` check.

thorn0: Yes, it's a bug.

```console
$ prettier --file-info ./node_modules/.bin/prettier
{ "ignored": true, "inferredParser": "babel" }

$ prettier --file-info ./node_modules/.bin/prettier --ignore-path ../bla
{ "ignored": true, "inferredParser": null }
```

fisker: I'll fix this separately after #8551 merged.

fisker: #8585

thorn0:
I'd like to take a deeper look at this PR, but I need time for that. Please don't merge for now.

fisker:
It's ready long time ago, solve the conflicts, still waiting for @thorn0 review.

thorn0:
Here's what bothers me about this change.

**./foo/ignore.txt**
```
*.js
```

**./bar.js**
```
1
```

```console
prettier --ignore-path foo/ignore.txt .
```

Result with Prettier 2.0.5: `bar.js` is ignored.
Result with this PR: `bar.js` isn't ignored.

So this change is breaking.

fisker:
@thorn0  `ignore@5` works like this, and I think it's reasonable, add I did add [a test](https://github.com/prettier/prettier/pull/8546/files#diff-0f21c65468d0ba6ad9716e52d3e6560cR65) to confirm  that "Should not ignore files in parent dir".

fisker:
FYI, the related file path pass to `ignore.ignores` is `../bar.js`, so `ignore` didn't ignore files in parent dir. (This change was made in #7588)

thorn0:
I understand why it works this way, but this is still a breaking change.

fisker:
If the breaking change is what you worried about, we can merge later.

Most important is, which behavior do you prefer? I think the new one not bad.

thorn0:
The new behavior is aligned with Git. It looks simpler and cleaner to me, especially if support for multiple ignore files is going to be added. But as long as Prettier doesn't have that support and because we don't know how and why people usually use `--ignore-path` with subdirectories, I'd prefer to keep the current behavior to stay on the safe side.

In order to both keep the behavior and update `ignore`, a workaround seems to be possible: a separate ignorer for paths that start with `../`. It should use only patterns that don't have a slash at the beginning or middle and patterns that start with `**/`. It would be safe then to strip `..` from the relative paths passed to it.

fisker:
Use `ignore@5.2.0` with `allowRelativePaths: true` options, we don't need this anymore. See #11990

PR #8548, opened 2020-06-12
Do not inferParser for ignored files

fisker:
`inferredParser` is useless for ignored files.

### DIFF
--- a/src/common/get-file-info.js
+++ b/src/common/get-file-info.js
@@ -67,27 +67,38 @@ function _getFileInfo({
 }) {
   const fileInfo = {
     ignored: ignorer.ignores(filePath),
-    inferredParser: options.inferParser(filePath, plugins) || null,
+    inferredParser: null,
   };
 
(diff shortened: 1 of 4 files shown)


fisker:
Strange, those codes has test covered, but coverage says not.

src/common/get-file-info.js
```diff
@@ -67,27 +67,38 @@ function _getFileInfo({
 }) {
   const fileInfo = {
     ignored: ignorer.ignores(filePath),
-    inferredParser: options.inferParser(filePath, plugins) || null,
+    inferredParser: null,
   };
 
-  if (!fileInfo.inferredParser && resolveConfig) {
```
thorn0, line 73: Looks like a bug. `parser` can be overridden by the config.

fisker: Yes, I noticed this.

fisker: Added a test, I'll fix it in another PR.

PR #8551, opened 2020-06-12
Fix `inferredParser` when calling `getFileInfo()` with `resolveConfig: true`

fisker:
~Contains changes from https://github.com/prettier/prettier/pull/8548, will add more tests and try to DRY a little bit later.~

When `{resolveConfig: true}`, should return parser set by config file.

I have a question, shouldn't this `resolveConfig` value `true` by default?


thorn0:
> shouldn't this resolveConfig value true by default

That's a good point and a breaking change. Let's open an issue for that and mark it for milestone 3.0.

fisker:
Done #8581

Issue #8509, opened 2020-06-06
Docs: Ellipsis is positioned oddly on mac / Better font stack?


lydell:
In the docs, ellipsis (`…`) is positioned oddly (in the middle of the line) on mac:

<img width="832" alt="Screenshot 2020-06-06 at 22 42 43" src="https://user-images.githubusercontent.com/2142817/83954166-1663b900-a847-11ea-87c0-a059b1b4b7df.png">

It seems to be a problem with the font we are using (Apple SD Gothic Neo) according to https://discussions.apple.com/thread/7278554:

> That font is for the Korean language.  It's often a bad idea to use such fonts for English, because they are not designed for it and may well contain bugs like the one you are describing.  Better to use a font intended for the language you are writing in

Should we find a different font stack?

https://github.com/prettier/prettier/blob/a9fd99846a4b87d2590a74072d77fa241ccefd3f/website/static/overrides.css#L2-L3

jeneser:
`Source Sans Pro` is a good choice. It is [licensed](https://fonts.google.com/specimen/Source+Sans+Pro#license) under the Open Font License. Lots of projects are using it. `ESLint`、`Vue.js`...

We can remove `Apple SD Gothic Neo` and `AppleSDGothicNeo`.  Like this:

```
font-family: "Source Sans Pro" -apple-system, 
   BlinkMacSystemFont, Helvetica, Arial, sans-serif; 
```

PR #8560, opened 2020-06-15
feat: use different font stack for website

jeneser:
https://github.com/prettier/prettier/issues/8509

Using different font stack for website. `Apple SD Gothic Neo` and `AppleSDGothicNeo` are for the Korean language. It's often a bad idea to use such fonts for English.

### DIFF
--- a/website/static/overrides.css
+++ b/website/static/overrides.css
@@ -1,6 +1,6 @@
 html body {
-  font-family: "Source Sans Pro", -apple-system, BlinkMacSystemFont, Helvetica,
-    Arial, sans-serif;
+  font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen,
+    Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
 
   overflow-x: hidden;
 }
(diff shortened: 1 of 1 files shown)


thorn0:
It's not enough to add "Source Sans Pro" to `font-family`. The font needs to be loaded. But does this site really need a custom font? To me, on Windows, it looks good enough with Arial.

jeneser:
yeah, not all fonts are available on all computers.`font-family` property specifies a prioritized list.It provides a system of fallbacks.`Source Sans Pro` just a recommend font stack. If we want most users to use `Source Sans Pro`, the font needs to be loaded. Otherwise we don't have to do this.

lydell:
As far as I can tell, neither Windows, Mac nor Linux dists have Source Sans Pro by default?

jeneser:
Yes, you are right. Any suggestions for `Source Sans Pro`? I think we can load it through Google Fonts.

thorn0:
My vote goes for [`font-familty: sans-serif;`](https://css-tricks.com/sans-serif/).

jeneser:
Any other suggestions？

lydell:
Maybe the font stack https://reactjs.org uses?

fisker:
Is there another decl?

image

fisker:
And `overflow-x: hidden` seems useless? Can we just we remove this file.

thorn0:
@fisket 
> Is there another decl?

Those are Docusaurus's default fonts (from `node_modules/docusaurus/lib/static/css/main.css`). A bit different than the proposed override.

fisker:
@sosukesuzuki Your review is requested by author.
@lydell You suggested the font stack of reactjs.org .

fisker:
Thanks!

Issue #8565, opened 2020-06-15
Prettier breaks code when inline lists are used as arbitrary arguments [SCSS]


boyenn:
I don't know why anyone would ever use a list this way, but it is valid code and prettier breaks it. 

**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAjCATAngAmAHShx1QEMwBrAcwCcIBXKDAWkgBsIakcaqyAKAKwAGHCLGiRAOhkBKANyEAviAA0ICAAcYAS2gBnZKFI06AdwAKJhIZSk2Z0lkPrUNchTgwAypvI6oKmQYGno4dQALGABbNgB1CJ14fT8wOG8bJJ0ANySsZHB9FxAA-TgaGAt3KmjSZAAzezL1ACt9AA8AIXdKL29SaLgAGQC4BqbwkDb27wCqNjgARXoIeHG2ZpA-GjKaAv0wIrUtmgCYOJ0MGAjkAA5hdU06Mrj3TQKnuF3ssfUARxW8CqWlsIFI+mYUDgcAwMOONDgAJ0CKqpBqdSQjQ2kzK0R0wVCOLmC2WqzGmIm6hgpFQFyuNyQACYqe4dGw5gBhCDRWoFL6CY70MoAFRptixm2yYQAkkwED4wKdtABBJjeGBYBbrMpKJRAA)
```sh
--parser scss
```

**Input:**
```scss
body {
  background-color: rgba(50 50 50 50...);
}
```

**Output:**
```scss
body {
  background-color: rgba(50 50 50 50..);
}

```

**Expected behaviour:**
Same as input.

PR #8566, opened 2020-06-15
Fix: Prettier breaks code when inline number lists are used as arbitrary arguments [SCSS]

boyenn:
Fixes: https://github.com/prettier/prettier/issues/8565

Not sure if this warrants a changelog ? Might just clutter it ?

### DIFF
--- a/changelog_unreleased/css/pr-8567.md
+++ b/changelog_unreleased/css/pr-8567.md
@@ -1,6 +1,7 @@
 #### Improve Prettier handling of arbitrary arguments ([#8567](https://github.com/prettier/prettier/pull/8567), ([#8566](https://github.com/prettier/prettier/pull/8566) by [@boyenn](https://github.com/boyenn))
 
 Prettier no longer unexpectedly add extra space after function when passing arbitrary arguments.
+
 <!-- prettier-ignore -->
 ```css
 /* Input */
@@ -23,6 +24,7 @@ body {
 ```
 
 Prettier no longer breaks code when inline number lists are used as arbitrary arguments.
+
 <!-- prettier-ignore -->
 ```css
 /* Input */
--- a/changelog_unreleased/css/pr-8567.md
+++ b/changelog_unreleased/css/pr-8567.md
@@ -30,7 +30,6 @@ Prettier no longer breaks code when inline number lists are used as arbitrary ar
 /* Input */
 body {
   background-color: rgba(50 50 50 50...);
-  test: function(return-list($list)...);
 }
 
 /* Prettier stable */
--- a/changelog_unreleased/css/pr-8567.md
+++ b/changelog_unreleased/css/pr-8567.md
@@ -1,44 +1,28 @@
 #### Improve Prettier handling of arbitrary arguments ([#8567](https://github.com/prettier/prettier/pull/8567), [#8566](https://github.com/prettier/prettier/pull/8566) by [@boyenn](https://github.com/boyenn))
 
 Prettier no longer unexpectedly add extra space after function when passing arbitrary arguments.
-
-<!-- prettier-ignore -->
-```css
-/* Input */
-body {
-  test: function($list...);
-  test: function(return-list($list)...);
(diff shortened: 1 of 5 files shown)


boyenn:
@evilebottnawi I've noticed that sometimes you guys skip changelogs on really trivial things.

What are your thoughts on this PR ? I'm worried that adding a changelog here would also just be clutter?

src/language-css/parser-postcss.js
```diff
@@ -32,6 +32,18 @@ function parseValueNodes(nodes) {
   for (let i = 0; i < nodes.length; ++i) {
     const node = nodes[i];
 
+    if (
+      node.type === "number" &&
+      node.unit === ".." &&
+      node.value[node.value.length - 1] === "."
```
evilebottnawi, line 38: Will be great to add `isScssSyntax()` check here, because we don't want to parse `...` for less and css

boyenn: Can't find `isScssSyntax` anywhere in the codebase, do you mean checking options.parser ? 
If so, I don't have access to it in `parseValueNodes`

thorn0:
Yes, it's okay to skip changelogging such entries. Or you can add this PR to the changelog entry of #8567 if you feel they're related closely enough.

boyenn:
@thorn0 Combined it with the previous one. Tried my best to make it as readable as possible, but I'm not sure if I'm deviating from the standard too much ?

boyenn:
@thorn0 Combined it with the previous one. Tried my best to make it as readable as possible, but I'm not sure if I'm deviating from the standard too much ?

src/language-css/parser-postcss.js
boyenn: @evilebottnawi please advise, I had another look but I can't find any way to determine whether or not I'm in scss syntax this far into the callstack

fisker:
```markdown
#### ONE TITLE ([#pr-1](pr-1) by [@author-1](@author-1), [#pr-2](pr-2) by [@author-2](@author-2), [#pr-n](pr-n) by [@author-n](@author-n))
```

boyenn:
Seems I was pretty damn close.

I followed https://github.com/prettier/prettier/pull/8551/files as example, since that is also one with both PR's of same author.

fisker:
Check these two:

- [1](https://prettier.io/blog/2020/03/21/2.0.0.html#typescript-38-7631httpsgithubcomprettierprettierpull7631-by-thorn0httpsgithubcomthorn0-7764httpsgithubcomprettierprettierpull7764-by-sosukesuzukihttpsgithubcomsosukesuzuki-7804httpsgithubcomprettierprettierpull7804-by-sosukesuzukihttpsgithubcomsosukesuzuki)
- [2](https://prettier.io/blog/2020/03/21/2.0.0.html#always-add-a-space-after-the-function-keyword-3903httpsgithubcomprettierprettierpull3903-by-j-f1httpsgithubcomj-f1-josephfrazierhttpsgithubcomjosephfrazier-sosukesuzukihttpsgithubcomsosukesuzuki-thorn0httpsgithubcomthorn0-7516httpsgithubcomprettierprettierpull7516-by-bakkothttpsgithubcombakkot)

Hah

src/language-css/parser-postcss.js
evilebottnawi: Just pass it to `parseNestedCSS` and other nested functions

boyenn: Done!

changelog_unreleased/css/pr-8567.md
```diff
@@ -20,3 +22,23 @@ body {
...
+
+/* Prettier stable */
+body {
+  background-color: rgba(50 50 50 50..);
+}
+
+/* Prettier master */
+body {
+  background-color: rgba(50 50 50 50...);
+}
```
fisker, line 43: Can  we  use  just  one  code  block ?

Issue #7997, opened 2020-04-09
[Sass] Unexpectedly add extra space after function when passing arbitrary arguments


P233:
**Prettier 2.0.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAjCATAngAmAHShx3gGcYkcAzAVyjBgEtoAKAEgBtHyA6PgSgDchYmQrU6DZlBYAnODBqyoAWi7l26mPz48hhAL4gANCAgAHJtFLJQAQ1myIAdwAKDhDZR2Ozu1htTVFk7MABrBQBlc1DGKABzZBhZGjhTAAsYAFsOAHV0xjIYsDhIz0LGADdCrGRwUkCQONI4WRhXEPisu2QqHxbTACtSAA8AIRDwqLssuAAZOLhe-rSQYZHIuPiOOABFGgh4ZY4BkBjZFtk60jAGkzPZOJhcxgwYdOQADgAGU3MnFq5ELmOr-OCXSpLUwARwO8A6Fi8IDspBUUDgcAwmPu8lhjHkHTsXR6SD6J1WLSyjCSKQpWx2+0OS1JK1MMDsqBebw+SAATGyQowuAkAMIQLLdOrggCs9xoLQAKhyvGTTpVUgBJKBY2CRMCPSwAQW1kRgWB2xxaBgMQA)
```sh
--parser scss
```

**Input:**
```scss
body {
  test: function($list...);
  test: function(return-list($list)...);
}
```

**Output:**
```scss
body {
  test: function($list...);
  test: function(return-list($list) ...);
}

```

**Expected behavior:**
No extra space between the `return-list()` function and `...`.

PR #8567, opened 2020-06-15
Fix: [Sass] Unexpectedly add extra space after function when passing arbitrary arguments 

boyenn:
Fixes https://github.com/prettier/prettier/issues/7997


fisker:
Thank you!

Issue #8584, opened 2020-06-17
Prettier formats handlebars components incorrectly inside html tag


Drakeoon:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4D4A6UABEcMAM4AWAhgE5zoD0mUADgK4wC0kAtq9AQwiAG2oAjOCIC8+EABVKcIgGEI-QbDlFs1EeziyQ6OOTC1MrGJmjbM5cgen3HcAL5vCJL6Qo16TCwc3HwCUEJEvBD00qy0EKzkRHDYCNJR9KISUkaKymoa4VogOnpOciZmFlY2UHYOTi4GHoSojFh4IAA0IAnW0OTIoHTxAO4ACnQIgyjU2BCY6N0g4rTUYADWcDAAyqzrLADmyDC0Bj2UMLwiAOqUmPDk+2BwO9MPOA8AnsjgDssschwWgwcZrQ68ajIABmeiBPQAVuQAB4AITWm22O2ovDgABkWHAYXC4IiUTsjiI4ABFdgQeDEkTwkD7WhA2i-Q4iTC8XEcnpxFgwG6LGCUZAADgADAL4kCbmtWL84qZgallgBHOnwMEJGYgajkLjhBgMZb0LWYehg6gQqFIWFM0kgIG8TAnM7O8iUmnaokOkk9GASEXoMXIABMQbWmG5UEOBUhvyggmW7CB8gkM0dzOwBgAklATLAduZLDAAIJFnYwL5UxlAjxAA)
```sh
--arrow-parens avoid
--parser glimmer
--trailing-comma none
```

**Input:**
```hbs
<div>
  {{shared/input-component label="The Component" value="description" issue=issue}}
  
  {{shared/input-component more=props even=more label="The Component" value="description" issue=issue}}
</div>
```

**Output:**
```hbs
<div>
  {{
    shared/input-component label="The Component" value="description" issue=issue
  }}

  {{
    shared/input-component
    more=props
    even=more
    label="The Component"
    value="description"
    issue=issue
  }}
</div>
```

**Expected behavior:**
```hbs
<div>
  {{shared/input-component label="The Component" value="description" issue=issue}}

  {{shared/input-component
    more=props
    even=more
    label="The Component"
    value="description"
    issue=issue
  }}
</div>
```

[...]

PR #7157, opened 2019-12-18
handlebars: Improve MustacheStatement printing

dcyriller:
Here is the issue addressed by this PR:
```
{{!-- input --}}
<p>Hi here is your name, as it will be displayed  {{firstName}}  {{lastName}}   , welcome!</p>

{{!-- stable --}}
<p>
  Hi here is your name, as it will be displayed {{firstName}} {{lastName
  }} , welcome!
</p>

{{!-- master --}}
<p>
  Hi here is your name, as it will be displayed {{firstName}} {{
    lastName
  }} , welcome!
</p>
```

As a result, it will implement https://github.com/prettier/prettier/pull/7051#discussion_r351894842 and https://github.com/prettier/prettier/pull/7051#discussion_r351877770

It depends on [this first](https://github.com/prettier/prettier/pull/7072) and [second](https://github.com/prettier/prettier/pull/7072) PRs (they should be settled first).

- [x] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [x] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #8593, opened 2020-06-18
Handlebars: Fix incorrect classic component syntax formatting

mikoscz:
Resolves:
#8584 
https://github.com/jgwhite/prettier/issues/1#issuecomment-607273019
https://github.com/jgwhite/prettier/issues/1#issuecomment-630412651

- [x] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/handlebars/pr-8593.md` file following `changelog_unreleased/TEMPLATE.md`.

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -118,10 +118,15 @@ function print(path, options, print) {
       const isParentOfSpecifiedTypes = isParentOfSomeType(path, [
         "AttrNode",
         "ConcatStatement",
-        "ElementNode",
       ]);
+
+      const isChildOfElementNodeAndDoesNotHaveParams =
+        isParentOfSomeType(path, ["ElementNode"]) &&
+        doesNotHaveHashParams(n) &&
+        doesNotHavePositionalParams(n);
+
       const shouldBreakOpeningMustache =
-        isParentOfSpecifiedTypes && doesNotHaveHashParams(n) && isNotYield(n);
+        isParentOfSpecifiedTypes || isChildOfElementNodeAndDoesNotHaveParams;
 
       return group(
         concat([
@@ -655,10 +660,8 @@ function doesNotHaveHashParams(node) {
   return node.hash.pairs.length === 0;
 }
 
-function isNotYield(node) {
-  const isYield =
-    node.path.type === "PathExpression" && node.path.original === "yield";
-  return !isYield;
+function doesNotHavePositionalParams(node) {
+  return node.params.length === 0;
 }
 
 module.exports = {
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -121,7 +121,7 @@ function print(path, options, print) {
         "ElementNode",
       ]);
       const shouldBreakOpeningMustache =
-        isParentOfSpecifiedTypes && doesNotHaveHashParams(n);
+        isParentOfSpecifiedTypes && doesNotHaveHashParams(n) && isNotYield(n);
 
       return group(
         concat([
@@ -653,6 +653,12 @@ function locationToOffset(source, line, column) {
 
 function doesNotHaveHashParams(node) {
   return node.hash.pairs.length === 0;
+}
+
+function isNotYield(node) {
+  const isYield =
+    node.path.type === "PathExpression" && node.path.original === "yield";
+  return !isYield;
 }
 
 module.exports = {
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -115,11 +115,13 @@ function print(path, options, print) {
       );
     }
     case "MustacheStatement": {
-      const shouldBreakOpeningMustache = isParentOfSomeType(path, [
+      const isParentOfSpecifiedTypes = isParentOfSomeType(path, [
         "AttrNode",
         "ConcatStatement",
         "ElementNode",
       ]);
+      const shouldBreakOpeningMustache =
+        isParentOfSpecifiedTypes && doesNotHaveHashParams(n);
 
       return group(
         concat([
@@ -647,6 +649,10 @@ function locationToOffset(source, line, column) {
     seenLines += 1;
     seenChars = nextLine + 1;
   }
+}
+
+function doesNotHaveHashParams(node) {
+  return node.hash.pairs.length === 0;
 }
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


fisker:
@mikoscz  Do you know how it broken in v2.0?

src/language-handlebars/printer-glimmer.js
```diff
@@ -649,6 +651,10 @@ function locationToOffset(source, line, column) {
   }
 }
 
+function doesNotHaveHashParams(node) {
+  return node.hash.pairs.length === 0;
```
evilebottnawi, line 655: It is possible when node doesn't have `hash` or `hash.pairts` properites?

src/language-handlebars/printer-glimmer.js
```diff
@@ -115,11 +115,13 @@ function print(path, options, print) {
...
     }
     case "MustacheStatement": {
-      const shouldBreakOpeningMustache = isParentOfSomeType(path, [
+      const isParentOfSpecifiedTypes = isParentOfSomeType(path, [
         "AttrNode",
         "ConcatStatement",
         "ElementNode",
       ]);
+      const shouldBreakOpeningMustache =
+        isParentOfSpecifiedTypes && doesNotHaveHashParams(n);
```
ziw, line 124: hi, thanks for picking up this issue. Can you help confirm if this case in [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuExgEIAEB3ATtAc0wgFcYAHMzAX2oB0oAeAEwEsA3APgc0zQE9WcADbNMACgAWAQwDOknr0wxWMYXAC84yAFty0BDGWr1ASkW9Jcaczi4tu-VEOYrNu+ahLMw6VAIk0gSa2hB6BrA+fgFBcJ68prQMjAD0bFwMDGjocAAe5HBg8GKkFFSsULKstpgsHJxJUAJCoopScgpeSipqIY4RRj1mFq7Wtvah4c6RbuPxSr7+gcEOYU4uizHB8560IAA0IBDkKtCyyKDSuPjYAApXCOco0sLY0vznhwBGuNJgANZwGAAZXIfwqBGQMFwJDgh0kMB0wgA6pJVHBZGCwHBgY9VBxVPxkOBZJ8QBVZHYYLdfgQdNJkAAzF6Uw4AK1kuQAQr8AUDgdIdHAADIVOBMllwkAc3LAiHqACKJAg8AlwlZIDBuEpuGJBGErB0Qt1h3IuAqMGR1RgkmQAA4AAym-CU5G-cjEs0YuzscWHACOyvgNOOTxAcgAtM44LZmAcQLg4IHWImaUF6WqNZSdKwoTCpVV-Iqg+KkMz1VKYNIvlbmDbkAAmQ7Q6SsA3+ADCYQzKAxAFZ4yRKQAVatPcsa9iwgCSUFssGBYHNJwAgnPgTB+OpM3BaEA) is fixed by your PR? It should not break opening mustache with a yield & hash. thank you.

src/language-handlebars/printer-glimmer.js
mikoscz: According to glimmer `MustacheStatement` interface, this is not possible.
https://github.com/glimmerjs/glimmer-vm/blob/cba8afd3e665087ce9aa1457181b1955cfee91f0/packages/%40glimmer/syntax/lib/types/nodes.ts#L92
https://github.com/glimmerjs/glimmer-vm/blob/cba8afd3e665087ce9aa1457181b1955cfee91f0/packages/%40glimmer/syntax/lib/types/nodes.ts#L226

src/language-handlebars/printer-glimmer.js
mikoscz: You're right the yield was also affected with this bug. I've created a fix - do you know if it can be solved in a better way?

dcyriller:
Could you also add a test for:
```
<div>
  {{a-helper "that takes some arguments" this.anotherOne "why not" @aLastOneLongEnoughToBreak}}
</div>
```

tests/handlebars/mustache-statement/basics.hbs
```diff
@@ -28,15 +28,32 @@
 <div
   class="a-first-class
     {{if this.optionOne "optionOne"}}
-    
+
```
dcyriller, line 31: This change could be reverted

mikoscz:
> Could you also add a test for:
> 
> ```
> <div>
>   {{a-helper "that takes some arguments" this.anotherOne "why not" @aLastOneLongEnoughToBreak}}
> </div>
> ```

@dcyriller how it should behave?

```hbs
  {{
    a-helper
    'that takes some arguments'
    this.anotherOne
    'why not'
    @aLastOneLongEnoughToBreak
  }}
```

```hbs
  {{a-helper
    'that takes some arguments'
    this.anotherOne
    'why not'
    @aLastOneLongEnoughToBreak
  }}
```

tests/handlebars/mustache-statement/basics.hbs
mikoscz: I'll revert it at the end - It caused by my editor, thanks for the catch.

dcyriller:
@fisker I believe it's regression introduced in https://github.com/prettier/prettier/pull/7157

src/language-handlebars/printer-glimmer.js
```diff
@@ -115,12 +115,19 @@ function print(path, options, print) {
...
     }
     case "MustacheStatement": {
-      const shouldBreakOpeningMustache = isParentOfSomeType(path, [
+      const isParentOfSpecifiedTypes = isParentOfSomeType(path, [
         "AttrNode",
         "ConcatStatement",
-        "ElementNode",
       ]);
 
+      const isChildOfElementNodeAndDoesNotHaveParams =
```
dcyriller, line 123: RE: (moving to a thread)

> @dcyriller how it should behave?

IMHO it should be the second, as you just did. Because there is no reason to print differently a helper and a component.

dcyriller, line 123: RE: (moving to a thread)

> @dcyriller how it should behave?

IMHO it should be the second, as you just did. Because there is no reason to print differently a helper and a component.

src/language-handlebars/printer-glimmer.js
ziw: thanks for the fix. I'm not sure if there's a cleaner way. Your fix lgtm.

tests/handlebars/mustache-statement/basics.hbs
mikoscz: Done

mikoscz:
@dcyriller Do you think that we need something more to move forward with this one?

PR #8621, opened 2020-06-23
Recognize known html tags

fisker:


tests/html/tags/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3764,6 +3764,195 @@ printWidth: 80
...
+
+=====================================output=====================================
+<noscript>noscript</noscript
+><details><summary>summary long long long long </summary>details</details
+><dialog open>dialog</dialog
+><object data="horse.wav"
+  ><param name="autoplay" value="true" /><param
+    name="autoplay"
+    value="true" /></object
+
```
fisker, line 3825: This should be a bug.

fisker: #8626

src/language-html/constants.evaluate.js
```diff
@@ -23,16 +23,23 @@ const CSS_DISPLAY_TAGS = {
...
-
   button: "inline-block",
 
   // special cases for some css display=none elements
   template: "inline",
   source: "block",
   track: "block",
   script: "block",
+  param: "block",
+  noscript: "none",
```
thorn0, line 34: Why `none`? How is it different from `block`?

fisker: I  need  update tests, current example not good.

fisker: `none` seems same as `inline`, I'll check if `noscript` can set to `block`

fisker: On chrome, `<noscript>` is `inline`

image

fisker:
Forgot snapshot outdated, fixing.

Issue #8626, opened 2020-06-23
Unstable HTML format


fisker:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeCAjAVnMMAEAJgIYzEC8AOiABYQBOAznAHQDuxAbtQHyoAOxesQC2+KKLhUQxAK4wI-ADbEAntXydiS2VOox6u6pSh9BwsRJF6Z8xSvUhN23dINGQJvgHpMOPDwAjCAANCCKMACW0IzIoEL0EGwACkIIsSjaHKqxYRjCYADWcDAAyoJgkVAA5sjucGE0MCJKAOo0kfCMFXCl6Z2RnJ2qyOCMuSBVzPQwycLVIsTIAGbazGFYjAAeAEIFxWWSADJVcCtrDSCbW6VV1UpwAIqyEPDnSusg5tOjTS2hX3oVRgrUihBgNGQAA4AAxhfiJZitYT8UYIuDTThnMIARxe8DmigyMkYAFooHA4IQqQD6HA8ZE6XNiAslkhVh9LswRJE6oYuXcHs9Xmd2RcwmQMKDwZCkAAmCXCSJKO4AYQgIkWowxAFYAbJmAAVYgYDIcz6cXQASSg1NgpTAQP4MAAgrbSjBVA93swAL6+oA)
```sh
--parser html
```

**Input:**
```html
<object data="horse.wav"><param name="autoplay" value="true"
><param name="autoplay" value="true"
></object>1
```

**Output:**
```html
<object data="horse.wav"
  ><param name="autoplay" value="true" /><param
    name="autoplay"
    value="true" /></object

>1

```

**Second Output:**
```html
<object data="horse.wav"
  ><param name="autoplay" value="true" /><param
    name="autoplay"
    value="true" /></object
>1

```

**Expected behavior:**

AS Second Output

fisker:
It don't has to be `<object>`

**Prettier 2.0.5**
[...]

PR #8628, opened 2020-06-23
[DO NOT MERGE] Do not add `/` to HTML void tags

fisker:
Experimental for #5246

Local test `npx fisker/prettier#html-void-tag **/*.html`


fisker:
Failed test is the same issue #8626

PR #8629, opened 2020-06-23
Update `lock-threads`

fisker:
It has [moved to `GitHub Actions`](https://github.com/dessant/lock-threads/commit/eb01523c2455360feaadea2ccf7db8a21532bef4).


fisker:
We lock PR too? I saw [this one](https://github.com/prettier/prettier/pull/6942) is locked

evilebottnawi:
I think we don't need lock PR, some of them have good ideas, just not completely done

Issue #8633, opened 2020-06-23
Support escaping a MustacheStatement (`{{statement}}`) with a backslash


dcyriller:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAdVxgFsCuBnGAQzAAs4Bfc1KddTBAGzgDc4BaIt3A4sy62rUwAnOFjgw8HQl3xFSFciAA0ICAAcYAS2h5koQsOEQA7gAVDCPSkIMThAJ57VAI2HEA1hIDK64lqgAc2QYYRw4VRIYLAYAdRIteDw-MDhvK0StZkSHZHA8ZxAAvDhhGDN3QKxCZAAzWxLVACs8AA8AIXcwLxhvQnEAGQC4OoaIkBbW7wDApgBFHAh4UYZGkD9hEuE82a0scW3VdWEAmFitABMYEmQADgAGI+MS2Pd1POO4LdYVEABHRbwCoaawgQhSKBwOAXaG-UQArSiCqEKo1JD1VbjEpYLQhMJYmbzQEjdFjVREFznK43JAAJnJ7i0DBmAGEIPs0SAvgBWX74OAAFUILmsGLWzHCAEkoDDYN4wCdNABBGXeGAOJgrEqUIA)
```sh
--parser glimmer
```

**Input:**
```hbs
\{{mustache}}
\\{{mustache}}
\\\{{mustache}}
```

**Output:**
```hbs
{{mustache}}
\{{mustache}}
\\{{mustache}}
```

**Expected behavior:**
```hbs
\{{mustache}}
\\{{mustache}}
\\\{{mustache}}
```

PR #8634, opened 2020-06-23
handlebars: Support escaping a mustache with a backslash

dcyriller:
## Description

In handlebars, it is possible to escape a mustache with a backslash. [See handelbars documentation here.](https://handlebarsjs.com/guide/expressions.html#escaping-handlebars-expressions)

This PR propose to add the support for this feature.

This would close https://github.com/prettier/prettier/issues/8633

Initial report: https://github.com/jgwhite/prettier/issues/1#issuecomment-646227623

## Details

This PR would introduce a new file: `src/language-handlebars/preprocess.js`. It is present to make an adjustement to the AST produced by Glimmer parser.

Indeed, given that template:
```hbs
a non-escaped mustache: \\\{{helper}}
```
Glimmer parser will produce two nodes for the template body:
- a TextNode where `node.chars === "a non-escaped mustache: \\\\"` (note the backslash escaping here, so it is actually`"a non-escaped mustache: \\"`, which means that a backslash has been swallowed)
- a MustacheStatement for `{{helper}}`

In order to correctly print the input, we need to adjust the TextNode to `node.chars === "a non-escaped mustache: \\\\\\"` (or `"a non-escaped mustache: \\\"` without the backslash escaping), namely we have to add a backslash that is removed by Glimmer parser. This is the third commit purpose. If you'd like you could checkout [the second commit's CI report](https://github.com/prettier/prettier/runs/800743079). It highlights the need for the third commit (the preprocess file).

## Checklist

### DIFF
--- a/changelog_unreleased/handlebars/pr-8634.md
+++ b/changelog_unreleased/handlebars/pr-8634.md
@@ -1,0 +1,19 @@
+#### Support escaping a mustache with a backslash ([#8634](https://github.com/prettier/prettier/pull/8634) by [@dcyriller](https://github.com/dcyriller)
+
+<!-- prettier-ignore -->
+```hbs
+{{!-- Input --}}
+\{{mustache}}
+\\{{mustache}}
+\\\{{mustache}}
+
+{{!-- Prettier stable --}}
+{{mustache}}
+\{{mustache}}
+\\{{mustache}}
+
+{{!-- Prettier master --}}
+\{{mustache}}
+\\{{mustache}}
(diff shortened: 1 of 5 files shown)


fisker:
Just noticed thee package size increased ~500k, is there better way?

fisker:
`require("@glimmer/syntax/dist/commonjs/es5/lib/traversal/traverse.js")` ?

fisker:
Or  maybe let  it  be  parser `postprocess `?

PR #8637, opened 2020-06-24
Support global installation run local installation

### LINKED ISSUE #8463: CLI: globally installed Prettier should run locally installed one
With [`import-local`](https://github.com/sindresorhus/import-local) we can make the global installed Prettier try to run local one.

Example: [ava project](https://github.com/avajs/ava/blob/master/cli.js).

We should implement this logic.

_Originally posted by @fisker in https://github.com/prettier/prettier/issues/8094#issuecomment-622657645_

fisker:
Fixes #8463

### DIFF
--- a/.github/workflows/prod-test.yml
+++ b/.github/workflows/prod-test.yml
@@ -116,6 +116,9 @@ jobs:
       - name: Install Dependencies
         run: yarn install --frozen-lockfile
 
+      - name: Remove Local Installed Prettier
+        run: yarn remove prettier
+
       - name: Download Artifact
         uses: actions/download-artifact@v2
         with:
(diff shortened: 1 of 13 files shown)


fisker:
Finally make it work. 

But questions:

1. I add extra `./bin-prettier.js`, because if make `./bin/prettier.js` use this logic, when running integration tests, it will run stable version Prettier. And to support installation from GitHub, we need the same filename as the bundled version, unless we hard-code the filename.
2. To make the `yarn test:dist` pass on CI, I have to uninstall the installed stable Prettier, because it will try to run local one. Dev_Test didn't effected because I'm running `./bin/prettier.js` for Dev_Test.
3. Actually it's don't has to be installed globally, if we run Prettier in a dir which has Prettier installed, no matter where this running Prettier installed, it will try local one. I learn this today, and `ava`, `jest`, `xo` do the same, it should not be a problem.  This means, run `node path/to/bin-prettier.js` in prettier codebase(has Prettier installed), no matter where it is, even `./bin-prettier.js`. It will always run the one in `node_modules/prettier/bin-prettier.js`

thorn0:
`yarn test:dist` should work correctly locally too, not only on CI.

fisker:
1. Use yarn link, but seems we can't do this, we need stable version to format files.
2. Use another cli file without `importLocal` to run test.

Suggestion?

fisker:
Global installation is not recommended.

PR #8642, opened 2020-06-25
Support printing handlebars in javascript files though `hbs`

dcyriller:
### `hbs` usages
`hbs` is provided by the [ember-cli-htmlbars library](https://github.com/ember-cli/ember-cli-htmlbars).

The first usage is in tests, where components are rendered:
```js
import { hbs } from 'ember-cli-htmlbars';

await render(hbs`
  <MyComponent />
`);
```

A second usage can be with Storybook:
```js
import { text } from '@storybook/addon-knobs';

import hbs from 'htmlbars-inline-precompile';

export default {
  title: 'Components/Checkbox',
};

export const checkbox = () => {
  return {
    template: hbs`<Checkbox @label={{this.label}} {{on "click" this.toggleCheckbox}} />`,
    context: {
      label: text('title', 'Welcome to storybook'),
      toggleCheckbox() {},
    },
  };
};

checkbox.story = {
  name: '...',
};
```

### technical details
javascript's `TaggedTemplateExpression`s with the tag `hbs` will be printed using glimmer printer.

### credit
This implementation mimicks html's.

It is adapted in two ways:
- `hbs` doesn't support printing <script> tags
- `hbs` doesn't support interpolation with ${}

## Links / Related
- [Support printing inline handlebars in html](https://github.com/prettier/prettier/pull/7306)

## Checklist

- [x] I’ve added tests to confirm my change works.
- [x] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [x] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #8647, opened 2020-06-26
Support printing handlebars in javascript files though `hbs`

dcyriller:
### `hbs` usages
`hbs` is provided by the [ember-cli-htmlbars library](https://github.com/ember-cli/ember-cli-htmlbars).

The first usage is in tests, where components are rendered:
```js
import { hbs } from 'ember-cli-htmlbars';

await render(hbs`
  <MyComponent />
`);
```

A second usage can be with Storybook:
```js
import { text } from '@storybook/addon-knobs';

import hbs from 'htmlbars-inline-precompile';

export default {
  title: 'Components/Checkbox',
};

export const checkbox = () => {
  return {
    template: hbs`<Checkbox @label={{this.label}} {{on "click" this.toggleCheckbox}} />`,
    context: {
      label: text('title', 'Welcome to storybook'),
      toggleCheckbox() {},
    },
  };
};

checkbox.story = {
  name: '...',
};
```

### technical details
javascript's `TaggedTemplateExpression`s with the tag `hbs` will be printed using glimmer printer.

### credit
This implementation mimicks html's.

It is adapted in two ways:
- `hbs` doesn't support printing <script> tags
- `hbs` doesn't support interpolation with ${}

## Links / Related
- second take for https://github.com/prettier/prettier/pull/8642
- [Support printing inline handlebars in html](https://github.com/prettier/prettier/pull/7306)

## Checklist


src/language-js/embed.js
```diff
@@ -648,4 +652,61 @@ function printHtmlTemplateLiteral(path, print, textToDoc, parser, options) {
...
+
+  const text = node.quasis.map((quasi) => quasi.value.cooked).join("");
+
+  const expressionDocs = path.map(print, "expressions");
+
+  if (expressionDocs.length === 0 && text.trim().length === 0) {
+    return "``";
+  }
+
+  const contentDoc = textToDoc(text, { parser: "glimmer" });
```
fisker, line 684: Can  we  call  `stripTrailingHardline`  here ? We  are  planning to  strip all  the finnal new line in  `textToDoc` function.

https://github.com/prettier/prettier/pull/8595


Just  a  question.

dcyriller: Yes we can and we will be able to remove it after #8595. Interesting I didn't notice that discussion.

I'm adding it back.

fisker: No need extra changes? Why finnal new line didn't effect.

dcyriller: Because there is already no final newline in handlebars formatted templates.

A while ago, I wanted to change that (https://github.com/prettier/prettier/pull/6243). But it was a dead end.

dcyriller: An option to change that could be to read the insert_final_newline option from .editorconfig. But it is not yet supported in the prettier package that reads editorconfig and I’m not sure it would be accepted?

fisker: I don't know that, but we always insert one, maybe we should open an issue and discuss about that.

fisker: Good to know, thanks.

fisker:
**Prettier pr-8647**
[Playground link](https://deploy-preview-8647--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEALARgZwAYB1fbYDcuUqMAtgDZ4HEgA0IEADjAJbSbKgCGATvwgB3AAoCE3FLyrDeAT25N0-XmADWcGAGUWa9lADmyGPwCucJnAro4AEzv2AMryNnehuADEI-CrxgOI2QQXjMYCEY0SioAdVR2eEw9MDhtSUT2ADdE+RCwTCUQA0w4fhhRVUN-ZAAzGVKmACtMAA8AIVUNLW1eCjgnAzg6hssQFtbtA0MqOABFMwh4EapGkD1+Uv4Q9F5bKiiWfgMYWPY7GFRkAA4ABiYjiFLY1RYQo7gtrOGmAEdF+CVVhSUKYAC0UDg9nsUX4cH+7DhlQ8NSQ9VWY1KFHYJnMmOmswWS2GaNGTBgezOFyuSAATOTVOwqNMAMIQCiokCfACsUTMpQAKnspOi1lkLABJKCOWDaMDHNgAQWl2hg8lmK1KAF8tUA)
```sh
--parser babel
```

**Input:**
```jsx
hbs`\``;
html`\``;
```

**Output:**
````jsx
hbs```;
html`\``;

````

**Comment by fisker** [comment]
**Prettier pr-8647**
[Playground link](https://deploy-preview-8647--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEALARgZwAYBJgCGAvtgNwA6UqMAtgDZ6EmkgA0IEADjAJbSbJQBAE7CIAdwAKIhAJQE64ggE8B7dMIJgA1nBgBlTlp5QA5shjCArnHZwa6OABMnzgDIEzVgqbgAxCGEaAhheM2QQAisYCDY0WjoAdVQeeEwjMDh9WVSeADdU5QiwTDUQE0w4YRhJTVNg5AAzBUr2ACtMAA8AIU0dPX0CGjg3EzgmltsQDs79E1M6OABFKwh4CbpWkCNhSuEI9AJHOjjOYRMYRJ4nGFRkAA4ABnYziErEzU4Is7g9vPH2ABHVbwWpcOSRTAAWigcGczjiwjgwJ4SNqPgaSGamymlRoPAs1lx80WKzW4yxk3YMCOVxudyQACZqZoeHR5gBhCA0TEgX4AVjiVkqABUjnJsVs8jYAJJQVywfRgc7cACC8v0MGUiw2lSIRCAA)
```sh
--parser babel
```

**Input:**
```jsx
hbs`${a}`;
html`${a}`;
```

**Output:**
```jsx
hbs``;
html`${a}`;

```

dcyriller:
Good catch.
I found out there is no reason not execute `printHtmlTemplateLiteral` function with "glimmer" as a parser.

dcyriller:
@thorn0 I have updated the tests to run them with the option `--embedded-language-formatting="{auto,off}"`

dcyriller:
@thorn0 did I understand your comment correctly?

thorn0:
@dcyriller "The idea is to design a way to use this option to specify the correspondence between template tags, special comments, etc. and embedded syntaxes whereas the default behavior won't be extended anymore." This PR extends the defautl behavior.

dcyriller:
I read carefully the linked PR and the various issues. I understand well the problem with formatting embedded snippets and that the 1st goal of Prettier is correctness.

So I've narrowed this implementation to only format:
```
hbs`some {{handlebars content}} here`
```
Previously it would also format: 
```
const code = /* hbs */ `some {{handlebars content}} here`;
```

If an user selects the option `--embedded-language-formatting=off`, the embedded snippet won't get formatted. If `--embedded-language-formatting=auto`, it will. (The tests show it)

I'm not sure what else I should do. 🤔

thorn0:
What I'm saying is that this PR (as well as #5993 and #9025) is blocked by #5588. We won't move forward with it until we solve that problem.

aogaili:
Any reason why this change is not being merged? 

We're using ember in a large project and would also like to introduce prettier and this is the only stopper.

Alonski:
@dcyriller 👋  Do you still work on this stuff? Any idea if this can be pushed forward somehow? Thanks!

ijlee2:
People can now use https://github.com/ijlee2/prettier-plugin-ember-hbs-tag to format `hbs` tags in JavaScript and TypeScript files.

PR #8510, opened 2020-06-06
Document recommended setup and improve docs overall

lydell:
Fixes #7828. Fixes https://github.com/prettier/prettier/issues/8464

**✨Check it out✨**

- [Landing page](https://deploy-preview-8510--prettier.netlify.app/)
- [Install](https://deploy-preview-8510--prettier.netlify.app/docs/en/next/install.html)
- [Integrating with Linters](https://deploy-preview-8510--prettier.netlify.app/docs/en/next/integrating-with-linters.html)
- [Editor Integration](https://deploy-preview-8510--prettier.netlify.app/docs/en/next/editors.html)
- [Related Projects](https://deploy-preview-8510--prettier.netlify.app/docs/en/next/related-projects.html)

PR #8652, opened 2020-06-28
Update stable docs

lydell:
To get https://github.com/prettier/prettier/pull/8510 out :shipit: 

I ran `yarn update-stable-docs` and then looked through the diff. The only thing I could find that we don’t want to release in the stable docs yet was docs for the new “Embedded Language Formatting” option. I removed that section from version-stable (it will appear again next time we run `yarn update-stable-docs`).


lydell:
Is it OK to merge this?

Issue #6207, opened 2019-06-10
[Glimmer] Unwanted line breaks within dynamic element class name


GavinJoyce:
**Prettier pr-6178**
[Playground link](https://deploy-preview-6178--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIwA2AhgM4kC8AOiABZwEES7DCYBmuA5hBOgEYBPOLhgAnAK5wAvlOasO3XoOFsiBEtNkt2XHvyEiJmuTsX6VajTJMK9yw5JnUAfJSi5cACXqM3qAPRY2K5QIAA0IBAADjCY0CTIoESiohAA7gAKyQgJKETYEJjo4SB8okRgANZwMADKURWYUJzIYpIRNDAAtgQA6jSY8CQNYHC1OYM4gwLI4GQlTRqiMBnlnF1EyKrqcBEAViQAHgBC5VU1tURdcAAyTXBblrsgB4e1TZwEcACK4hDwjx2EQaoiWs0+mC611EJSioiaMF6RRgNGQAA4AAzA1IaXrlKKzOFwJbYB4RACOf3gq2iuRApAAtFA4HB0KySqI4JTMJzVkR1pskNsNBENF1MK0jKKPl9fv8HkKnhEYEQ+Ej0CjkAAmZXlTAED4AYQgUMFICg0DJIHEGgAKqrcsLNEA)
```sh
--parser glimmer
```

**Input:**
```hbs
<div class="hello {{if goodbye true}} {{if goodbye false}} {{if goodbye true}} {{if goodbye false}} {{if goodbye true}}">
  Hello
</div>

```

**Output:**
```hbs
<div
  class="hello
    {{if goodbye true}}

    {{if goodbye false}}

    {{if goodbye true}}

    {{if goodbye false}}

    {{if goodbye true}}"
>
  Hello
</div>
```

**Expected behavior:**

```hbs
<div
  class="hello
    {{if goodbye true}}
    {{if goodbye false}}
    {{if goodbye true}}
    {{if goodbye false}}
    {{if goodbye true}}"
>
  Hello
</div>
```

dcyriller:
On master we now have the following output:
```
<div
	class="hello {{if goodbye true}} {{if goodbye false}} {{if goodbye true}} {{
		if goodbye false
	}} {{if goodbye true}}"
>
	Hello
</div>
```

thorn0:
Fixed by #10145

PR #8667, opened 2020-06-29
Glimmer: fix newlines rendering in non-mustache attributes

### LINKED ISSUE #8648: Handlebars: newline in an attribute incorrectly rendered as `[object Object]`
As reported in https://github.com/jgwhite/prettier/issues/1#issuecomment-640699531

**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAYhCACAAgQwCcBzAXgB0RypKB6APhABoQIAHGAS2gGdlQihCAHcACkQR8U+ADbD8ATz4sARoXxgA1nBgBlNhs5RiyGIQCucFgAsYAWxkB1a53g8DYOLsmvOAN1cFZHAeZRAjHjhCGFF1Yjt8ZAAzWUiWACseAA8AIXUtHV18OzgAGSM4ZNSrEEys3SNiGTgARXMIeCqZNJADQkjCYKbOOxLBljZCIxhHTgATGGtkAA4ABgmhSMd1NmDJuAG-SpYAR3b4WPYpEHweAFooODg55+YQQjgzzg-Y-HjEpApbo1SJ2TimCwgxrNNodSqA6osGD4FSzBZLJAAJiR6k4MkaAGEIKMASADgBWN7mSIAFRRUiBPT8lgAklAXrBdGAphwAILs3QwBTNLqRAC+YqAA)

**Input:**

```
<Foo arg="
"/>
```

**Output:**

```
<Foo arg="[object Object]" />
```

**Expected behavior:**

```
<Foo arg="
" />
```

kangax:
Fixes #8648 

It appears that `hardline` "primitive" can't be rendered within an attribute in Prettier and results in `\n` output as `[object Object]`.

When working on a fix for this, I found out that Glimmer parser treats attribute values with mustaches differently: they're a `ConcatStatement` rather than a single `TextNode` if the attribute value is a plain string.

This PR changes `TextNode` output behavior to only trim it if it's the only node within an attribute (as opposed to `ConcatStatement`).


kangax:
@dcyriller I saw you were an author of this part of code. Let me know what you think! My first PR to prettier so might be rough around the edges :)

fisker:
We don't trim attribute in html

**Prettier pr-8667**
[Playground link](https://deploy-preview-8667--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAhgAhgSxgGzgF4AdEEqdGGAJwrID5UB6dBikAGhAgAddoAZ2Sh0NGhADuABTEJhKdPknoAnsO4AjGujABrODADKvXTigBzZLQCucbnAC2muABNXbgDLpLN9BbgAMQgaRypcS2QQdBsYCC4QAAsYR3wAdUS8OEFTMDgjeTwcADc8VSiwQQ0Qc0E4GhhpHQsw5AAzJTruACtBAA8AIR19QyN0RzhPczh2zvsQXr6jcwtCAEUbCHhZ-C6QUxo6mijk1ITeGnMYNJxXGETkAA4ABm4LiDq0nV4oi+z64ozbgAR028CafAU0UEAFooHA3G4EjQ4KCcCimv5WkgOrt5nVHDhrDQ7NxBCt1mCZji5twYOhNDc7g8kAAmOk6HD4FYAYQgjmxIGyAFYEjY6gAVBkKXF7Yp2ACSUA8sCMYEu-AAgsqjDBVIQdnUAL5GoA)
```sh
--parser html
```

**Input:**
```html
<a title="
attr
"></a>

```

**Output:**
```html
<a
  title="
attr
"
></a>

```

src/language-handlebars/printer-glimmer.js
```diff
@@ -267,9 +267,21 @@ function print(path, options, print) {
...
 
+      const trimmedValue = text
+        .replace(/^\s+/g, leadingSpace)
+        .replace(/\s+$/, trailingSpace);
+
+      // Glimmer parser treats attribute values with mustaches differently
+      // they're a ConcatStatement rather than a single TextNode
+      // if the attribute value is a plain string, we only trim it and avoid inserting hardlines
+      // since they're not rendered correctly within attributes
+      if (isParentOfSomeType(path, ["AttrNode"])) {
```
dcyriller, line 278: This conditionnal will match `<htmlElement attribute="textNode">`'s TextNode but not `<htmlElement attribute="textNodeInConcatStatement {{mustache}}">`'s.

I would suggest to use instead:
```suggestion
      if (path.stack.includes("attributes")) {
```
to match both.

dcyriller: If you apply this suggestion, you'll notice that you can return `return concat([trimmedValue]);` from line 234 and get rid of the big `else` statement.

Doing that, we'd split the TextNode printer in two: one for `TextNode`s nested in attributes, one for other `TextNode`s. We'd be able later to print differently attributes whether they be `class` / `style` / `srcset` / other.

tests/handlebars/whitespace/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -407,6 +407,33 @@ sometimes{{nogaps}}areimportant<Hello></Hello>
 
 </div>
 
+<div class="attr
```
dcyriller, line 410: There is a `tests/handlebars/attr-node` test folder. Maybe we should create an `attributes.hbs` files there?

dcyriller:
@fisker handlebars printer for attributes is not yet splitted: we could rewrite whitespaces in `class` and `style` attributes, print `scrset` in a custom way and do not process other attributes. This is what html printer does? What do you think?

fisker:
@dcyriller I'm not familiar with handlebars, html didn't trim attribute, I think we should not do trim in handlebars either? And yes, html format `style` `class` and `srcset`.

src/language-handlebars/printer-glimmer.js
kangax: > This conditionnal will match <htmlElement attribute="textNode">'s TextNode but not <htmlElement attribute="textNodeInConcatStatement {{mustache}}">'s.

Correct and that's the whole point — presence of `MustacheStatement` makes it `ConcatStatement` which is handled properly.

tests/handlebars/whitespace/__snapshots__/jsfmt.spec.js.snap
kangax: Good idea, I'll move it there.

kangax:
@fisker @dcyriller Ah, I missed the fact that Prettier treats attribute values differently depending on an attribute. I initially skipped trimming entirely but then noticed a failing glimmer test that checked for `class="  smth  "` being trimmed to `class="smth"`. I then saw in the playground that `class` attribute does indeed get trimmed in, at least, html printer.

Looking at [something like this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAllADgVxgHSjABsBDAZ3IF58RCSLzbDyYBPYuGuqAMwghIARqQBOzKOVFhycGNxbTZBHjHQxOC2Os4TSxeXoMSckrHDDpe6OABMtp8ucvW7EgPQA+EABoQELDVocmRQMVEIAHcABTEEEJR9SNI2EL8hUVIwAGs5AGUsLMwAc2QYURw4PwALGABbYgB1avU4Jyy4PPj1dAA3dTZkcEpfEExZURhozOK60mRefVk-ACtyAA8AIUyc-NI6uAAZTDgFpaqQNfW8ks4ARRwIeDPiZZBC0Qmh2obRrFFMDBGuhbDBqsgABwABj8-wgskamSwQ3+bTgol6pz8AEdHvBpgEEiAKABaKBwOx2UaiOC49A06akWbzJCLV4XWR1dBlCoc25wB5PU6s85+GCkITA0HgpAAJjFmXQxBKAGEIHU5kM2gBWUY4WQAFQlCTZb16lQAklBbAgYHkwADAgBBa15dicF6yAC+XqAA):

```html
<input
class="
class"
style="
foo:bar"
srcset="
srcset"
title="
title"
alt="
alt"
unspecified="
unspecified"
/>
```

I see that "html" printer only trims `class` and `style`?

```html
<input
  class="class"
  style="foo: bar;"
  srcset="
srcset"
  title="
title"
  alt="
alt"
  unspecified="
unspecified"
/>
```

`JSX` printer seems to leave everything untouched:

```html
<input
  class="
class"
  style="
foo:bar"
  srcset="
srcset"
  title="
title"
  alt="
alt"
  unspecified="
unspecified"
/>;
```

I can skip trimming for now entirely, so that we get consistent behavior and fix this bug, then work on splitting printer behavior depending on an attribute?

fisker:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAlgWwOYB0pgA2AhgM6kC8uI+RZp1+pMAnoXFTVAGYQRIAjYgCdGUUsLCk4MTk0nSYAWhWq16jZq3ad6gAQBGAB4AafHouWLEqTID6AJgsOjYmOhjs5sD+zHFCWX9AsQBXcQAHODB0bnQ4ABNvcNIomLjEsQB6AD4QExAICPdoUmRQEWEIAHcABREEMpQA6uIWMoKBYWIwAGsZAGUInvQobGQYYVC4AoALGExCAHVZjzhUnrgBxo90ADcPFmRwcnyQUelhGFru7ExiZG4A6QKAK1IjACFuvsHiTDgABlRnBHs8ZiB3kYBqNsOwAIqhCDwMGEF4gYbCS7HeaLM4RYSjGBLdAJGCzZAADgADAUCRBpEtuhFjgT1nBhHtQQUAI5I+A3IpNEBkJRQOCJRJnYRwPnoGU3Yh3B5IJ5oiHSTDoCZTDWwhH80Gq8EFGDEAQkskUpAOU3ddCEWEAYQgmHux3WAFYzqFpAAVc1NNXovbTACSUASCBgAzAhOKAEFIwNWOxUdIAL4ZoA)
```sh
--parser html
```

**Input:**
```html
<img
class="
class"
style="
foo:bar"
srcset="
srcset------------------------------------------------------------ 1x,
        srcset_2  2x"
title="
title"
alt="
alt"
unspecified="
unspecified"
/>
```

**Output:**
```html
<img
  class="class"
  style="foo: bar;"
  srcset="
    srcset------------------------------------------------------------ 1x,
    srcset_2                                                           2x
  "
  title="
title"
  alt="
alt"
  unspecified="
unspecified"
/>

```

kangax:
We can look at `cleanAttributeValue` from my html-minifier (haven't worked on it in a while, forgot the rules) for some inspiration :) https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L257-L328

src/language-handlebars/printer-glimmer.js
dcyriller: It is handled correctly in the sense it doesn’t have the bug. But a TextNode in a ConcatStatement should probably be formatted no differently than a simple TextNode (ie. trimmed / not trimmed)?

kangax: @dcyriller just tried your suggestion and remembered why I didn't do that. It breaks mustache handling in attributes:

```diff
    <div
    -   class='a-first-class
    -   {{if this.optionOne 'optionOne'}}
    -
    -    {{if this.optionTwo 'optionTwo'}}
    -
    -    {{if this.optionThree 'optionThree'}}
    -
    -    {{if this.optionFour 'optionFour'}}
    -
    -    {{if this.optionFive 'optionFive'}}
    -
    -    {{this.class}}'
    +   class='a-first-class {{if this.optionOne 'optionOne'}} {{
    +     if this.optionTwo 'optionTwo'
    +   }} {{if this.optionThree 'optionThree'}} {{
    +     if this.optionFour 'optionFour'
    +   }} {{if this.optionFive 'optionFive'}} {{this.class}}'
        ...attributes
      >
      </div>
```

kangax:
@dcyriller would you like me to wait until you merge your PR and then we rebase this one on top? or should we merge this as is and then follow up with yours?

src/language-handlebars/printer-glimmer.js
dcyriller: This ☝️ would actually be an improvement - if applied to only class names.

src/language-handlebars/printer-glimmer.js
```diff
@@ -267,9 +267,21 @@ function print(path, options, print) {
         text = "\\" + text;
       }
 
+      const trimmedValue = text
+        .replace(/^\s+/g, leadingSpace)
+        .replace(/\s+$/, trailingSpace);
+
+      // Glimmer parser treats attribute values with mustaches differently
```
dcyriller, line 274: This comment might removed - it doesn't explain how prettier printer works (but the undelying ast).

tests/handlebars/whitespace/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -406,7 +406,6 @@ sometimes{{nogaps}}areimportant<Hello></Hello>
 <div class="a list of classes">
 
 </div>
```
dcyriller, line 408: This change might be reverted

dcyriller:
> @dcyriller would you like me to wait until you merge your PR and then we rebase this one on top? or should we merge this as is and then follow up with yours?

@fisker what do you think?

IMHO if this is merged, I can rebase and add more tests on https://github.com/prettier/prettier/pull/8677

fisker:
@dcyriller This PR prints trimmed attribute values, it's not okay right? So we can't merge this one.

dcyriller:
[The other one](https://github.com/prettier/prettier/pull/8677) - to format only class names - is ready 👍.

kangax:
@dcyriller ok, waiting for yours to merge before we go ahead with this one

src/language-handlebars/printer-glimmer.js
kangax: Yeah, this is something I wanted to discuss with you too. In https://github.com/prettier/prettier/issues/6207 they ask to format classes with newlines (but not extra newlines) whereas I feel like ☝️ option is not bad. We could consider that ticket an improvement or bring it up for more discussion.

Alonski:
@kangax This was merged :)
https://github.com/prettier/prettier/pull/8677

kangax:
@Alonski @dcyriller any idea why the tests are failing on CI? everything passes locally, ran `yarn test -u` and so on

fisker:
@kangax  the test fails because you changed the AST, need fix it in  [`clean`](https://github.com/prettier/prettier/blob/0509833f0a6eeca0f57a78857336bf63852c4ef4/src/language-handlebars/clean.js#L9), to test locally, `AST_COMPARE=1&&yarn test`

src/language-handlebars/clean.js
```diff
@@ -6,7 +6,7 @@ module.exports = function (ast, newNode) {
 
   // (Glimmer/HTML) ignore TextNode whitespace
   if (ast.type === "TextNode") {
-    const trimmed = ast.chars.trim();
+    const trimmed = ast.chars.trim().replace(/\s+/g, " ");
```
fisker, line 9: Please, only do nessary change, should check attribute name.

kangax: @fisker actually I don't quite understand why we need to do this. Could you explain? Why are we "cleaning" AST and when does this happen? Also, why does the test fail now if we only changed Prettier output and not the underlying AST? Checking for attribute==class will involve some duplication from Glimmer's printer which I'd rather avoid. I'm also not sure if this place is performance critical?

fisker: This only runs when comparing the AST during test, so performance is  not very  important, but  make  sure the formated / original code  have the same   AST is very important.

fisker: > Also, why does the test fail now if we only changed Prettier output and not the underlying AST?

You  changed the  output, it parse to  different ASTs.

fisker: I'm not  familiar with  glimmer  AST , but I checked failed tests on  CI , the  attribute with  `TextNode` have  `"name": "class"`, can  we  use it ?.

PR #8694, opened 2020-07-01
Suppress “Checking formatting...” message on non-TTY

andersk:
When Prettier is run by a script that combines the output from dozens of separate linters, any progress output is just noise that makes it harder to see whether there are real errors.  Similarly to how we already suppress printing each filename as it’s checked when outputting to a non-TTY, also suppress the “Checking formatting...” and “All matched files use Prettier code style!” messages in this case.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -56,7 +56,7 @@ This will output a human-friendly message and a list of unformatted files, if an
 prettier --check "src/**/*.js"
 ```
 
-Console output if all files are formatted:
+Console output if all files are formatted (these messages will be suppressed when redirecting output to a file or program):
 
 ```console
 Checking formatting...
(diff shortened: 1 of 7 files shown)


thorn0:
Doesn't `--loglevel warn` work for you? https://prettier.io/docs/en/cli.html#--loglevel

andersk:
@thorn0 No, `--loglevel warn` suppresses all output from `--check`, even when there are changes needed.

```console
$ prettier --check --loglevel warn .
$ prettier --check .
Checking formatting...
foo.js
bar.js
Code style issues found in the above file(s). Forgot to run Prettier?
```

I need this message in the changes-needed case, but no output in the no-changes case.

thorn0:
> --loglevel warn suppresses all output from --check, even when there are changes needed.

I think it's this that needs fixing.

andersk:
Okay, that seems fine to me; I opened #8703 for that alternative.

andersk:
Closing in favor of #8703.

PR #8696, opened 2020-07-02
Fix common util import

fisker:
- Use destructuring require.
- Remove import from `common/util-shared.js`


src/language-js/needs-parens.js
```diff
@@ -2,7 +2,12 @@
 
 const assert = require("assert");
 
-const util = require("../common/util");
+const {
+  isBitwiseOperator,
+  startsWithNoLookaheadToken,
+  getPrecedence,
+  shouldFlatten,
+} = require("../common/util");
```
fisker, line 10: https://github.com/prettier/prettier/pull/8697

PR #8703, opened 2020-07-02
Bump loglevel of issues reported by --check to warn

andersk:
When `prettier --check` is run by a script that combines the output from dozens of separate linters, any progress output like “Checking formatting...” and “All matched files use Prettier code style!” is just noise that makes it harder to see whether there are real errors.  Bump the loglevel of issues reported by `--check` from `log` to `warn` so that the noise can be selectively silenced with `--loglevel=warn`.

Alternative to #8694 suggested by @thorn0.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -67,9 +67,9 @@ Console output if some of the files require re-formatting:
 
 ```console
 Checking formatting...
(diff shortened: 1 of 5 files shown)


fisker:
Thank you!

Issue #6452, opened 2019-09-05
How to enable convert empty JSX elements to self-closing elements


dhmacs:
After version `1.18` Prettier does not merge empty tags anymore (https://github.com/prettier/prettier/pull/6127)

Is there an option to turn this back on? This option is cluttering the code and it's very annoying to manually merge empty tags.

evilebottnawi:
It is expected and no options for this and we don't plan to add option

macs91:
So there's no solution for this? Everybody is just merging empty tags manually? 🤔

evilebottnawi:
You can setup linter for this case

macs91:
I think I'll stick with `1.17` an hope you'll change your mind along the way 😅

evilebottnawi:
I don't recommend stick because we fix a lot of bugs, use linter for this case

lydell:
@macs91 You can use the [react/self-closing-comp](https://github.com/yannickcr/eslint-plugin-react/blob/97a9f397d2a8a3ce3b2af893bdbb86bb2c1d4480/docs/rules/self-closing-comp.md) ESLint rule. It has autofix! Have you tried it? Does that work for you?

macs91:
Thanks for the suggestion @lydell ! 
When we migrated to typescript we found eslint didn't bring value to our development process so we removed it from our toolchain. Adding it back just to fulfill this single purpose does not seem to be worth it (and if I'm not mistaken eslint generates warnings with fix suggestions that need to be manually applied to take effect and since we generally have lot of empty tags that would be an annoying process). 
We've been working happily with `v1.17` so we decided that for now we'll stick to this version.

thorn0:
> we generally have lot of empty tags

Lot? Where do you get them from, I wonder? Are you converting HTML to JSX using Prettier?

icyJoseph:
So users of create react app will have to wait for self-closing-comp to be implemented upstream?

evilebottnawi:
> we found eslint didn't bring value to our development process so we removed it from our toolchain

:confused:

thorn0:
@icyJoseph if I get what you mean, you've missed the fact that now CRA [allows to extend the ESLint config](https://create-react-app.dev/docs/setting-up-your-editor#experimental-extending-the-eslint-config).

icyJoseph:
Aha! I tend to leave things labelled experimental out of my sight. Thanks! Good stuff @thorn0.

macs91:
@thorn0 
[...]

PR #8730, opened 2020-07-09
Add an option to enable empty/self-closing tags conversion in JSX

MattiaMalonni:

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -97,6 +97,14 @@ Use single quotes instead of double quotes in JSX.
 | ------- | -------------------- | ------------------------ |
 | `false` | `--jsx-single-quote` | `jsxSingleQuote: <bool>` |
 
+## JSX Self Closing Tags
+
+Convert empty tags to self closing tags.
+
+| Default | CLI Override              | API Override                 |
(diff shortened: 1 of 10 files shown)


MattiaMalonni:
@evilebottnawi May I ask why?

MattiaMalonni:
According to this #6452, seems like a lot of people are asking for that.

evilebottnawi:
Answer is simple - https://prettier.io/docs/en/option-philosophy.html

kachkaev:
See ESLint this rule from [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react): [self-closing-comp](https://github.com/yannickcr/eslint-plugin-react/blob/97a9f397d2a8a3ce3b2af893bdbb86bb2c1d4480/docs/rules/self-closing-comp.md)

I'm using it at work and in home projects, all works well. Switching to self-closing tags changes AST, which would be out of Prettier’s scope even if it still accepted options.

MattiaMalonni:
@kachkaev TY.

@evilebottnawi Oh well, ok then. Should have read that post first..

PR #8734, opened 2020-07-10
Simplify `getPenultimate`

fisker:


fisker:
> This returns undefined instead of null. Does it matter? 🤔

It doesn't matter, it's internal, only used here https://github.com/prettier/prettier/blob/350f641fca83d40aadfc7df8302b861c2f60efbf/src/language-js/print/call-arguments.js#L279-L286

--- 

Similar simplification for `getLast` function

https://github.com/prettier/prettier/pull/7267

Issue #8736, opened 2020-07-10
Adding prettier-ignore comment in JSX doubles the comment


etr2460:
**Prettier 2.0.5**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAEnADYkQDqEATiQCYAUAlPsADpT77VwzrWcGHLlwA8dHADdhIrsAB0i4AF98AegBU+GbLVr8AARgBnALQ4A5lBpx8G-ZvwAHHjFxxq5qzbtrlOkRwoSQBDEhw6AAVqCCcAXjYQYjJKGnpEgIA+AK54YxgA0TUJSWzOfCYAbg5-KBAAGhBY3GhjZFAQ6hiAd0jOhDaUMO6QgE82xoAjahCwAGteAGUnWaCLZBhqdDhGgAsYAFsSCl2cPJWwOEWBs6kz0eRwYwmQIOMPGGiQiwOQ5FQwu9GgArYwADwAQjN5ksQgc4AAZIJwf6AnYgUFgxZrEhwACK6Ag8FRJCBIBW1He1EekxCk1IDXJ1CCMAoERgu2QAA4AAyNFwQd4UGZOR4uOBUyQoxoAR0J8GisUGIBCZigcDgdE1jJ4cpwPC+Pz+SABpPR7wOOA2W3NOPx8pRJrRjRgdLZdA5yAATC6ZjhwlALABhCAHX6PCUAVkZ6HeABU6YNTWTJNsAJJQLWwRZgZlOGAAQUzixgo1xJPeymUQA)
```sh
--parser babel
```

**Input:**
```jsx
function HelloWorld() {
  return (
    <div
      {...{} /* 
      // @ts-ignore */ /* prettier-ignore */}
      invalidProp="HelloWorld"
    >
      test
    </div>
  );
}

```

**Output:**
```jsx
function HelloWorld() {
  return (
    <div
      {
        ...{} /* 
      // @ts-ignore */ /* prettier-ignore */ /* 
      // @ts-ignore */ /* prettier-ignore */
      }
      invalidProp="HelloWorld"
    >
      test
    </div>
  );
}

```

**Expected behavior:**
[...]

PR #8743, opened 2020-07-13
Check comment `printed` before printComment

fisker:
We check not printed comments,

https://github.com/prettier/prettier/blob/350f641fca83d40aadfc7df8302b861c2f60efbf/src/main/core.js#L43-L52

I think make sense to prevent comments print multiple times too.

This can prevent #8736 happens.


fisker:
Turns out this is not easy as I thought, maybe look into it later, closing now.

PR #8753, opened 2020-07-14
Do not expose `composeLoc`

fisker:
`loc.js` has extra function `composeLoc`, this not public api, should not add to parser.


brodybits:
I hope this does not break any third-party applications or plugins. Or am I too pedantic?

fisker:
I don't think we need worry about this.

PR #8745, opened 2020-07-13
refactor typedefs in src/language-js/print/ternary.js

brodybits:
- move the JSDoc typedefs to the beginning of `src/language-js/print/ternary.js`
- add & use JSDoc typedef for `@param {FastPath} path`, which effectively reverts the change in PR #8740
- keep JSDoc typedef for `@typedef {Object} OperatorOptions` in its own comment block

I think these changes should improve the readability.

- ~~I’ve added tests to confirm my change works.~~
- ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~~
- ~~(If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.~~
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #8759, opened 2020-07-15
typecheck updates & fixes in src/language-js

brodybits:
__updated:__

- add some type support packages to devDependencies
- add & update JSDoc type comments
- add `src/language-js/types/estree.d.ts` to facilitate clean parameter types in `src/language-js/utils.js`
- apply some minor code updates to resolve some typecheck issues
- ~~add a single `@ts-ignore` comment, to work around a type issue remaining in `src/language-js/utils.js`~~
- remove `src/language-js` & `src/common/parser-create-error.js` from exclude list in `tsconfig.json`
- move & refactor typedefs in `src/language-js/print/ternary.js`, based on the changes I had originally proposed in PR #8745
- add some blank lines
- minor extra cleanup of `src/main/comments.js`, as described below

(there used to be many more `@ts-ignore` comments, which are now resolved)

P.S. I suspect some of the minor code updates can make the code a little more robust. It would be interesting to see if we can think of any test cases to prove this or not.


brodybits:
I just merged changes from master & replaced TODO with `TBD (...)` in the `@ts-ignore` comments. I would also like to rebase to improve the history message. My one issue is that I hope we can try to avoid using `@ts-ignore` comments to avoid resolving new JSDoc type issues in the future.

brodybits:
I just rebased this proposal with a hopefully cleaner commit history, with some updates:

- add some conditions to resolve JSDoc type issues, following suggestion in <https://github.com/prettier/prettier/pull/8760#issuecomment-658775087>
- add a few more blank lines
- improve `@ts-ignore` comments in `src/language-js/print/call-arguments.js`
- improve a comment in `src/language-js/print/ternary.js`

At this point there are 2 TODO comments for issues that I think should be addressed in the near future:

- error type issue in `src/common/parser-create-error.js`
- Prettier options object typedef in `src/language-js/print/ternary.js`

Other `@ts-ignore` comments are TBD comments - see the updated list in the description.

It looks like the build cancelled itself, would appreciate it if someone could restart the build.

src/language-js/comments.js
```diff
@@ -930,7 +933,7 @@ function isBlockComment(comment) {
   return comment.type === "Block" || comment.type === "CommentBlock";
 }
 
-function hasLeadingComment(node, fn = () => true) {
+function hasLeadingComment(node, fn = (_) => true) {
```
fisker, line 936: Can this done in other way?

brodybits: Yes, good catch.

Here is one way that works for me:

```suggestion
/**
 * @param {Object} node
 * @param {(comment: Object) => boolean} fn
 * @returns boolean
 */
function hasLeadingComment(node, fn = () => true) {
```

In this case, I think we should add a TODO item to use more specific types for the node parameter and comment function argument.

But I am now thinking we can just drop the `fn` argument since `git grep -w hasLeadingComments` shows it never used.

thorn0: It isn't practical to add a TODO in every place where better types are needed. Please don't do that. The type `any` ([AKA `Object`](https://github.com/microsoft/TypeScript/issues/20337)) is already a TODO in itself. You can also use the `?` type (another synonym for `any`, `*` works too) to make it even more explicit if you want: `/** @param {?} node */`.

fisker: I don't know anything about  this, just  wonder if default value of argument function with  5 parameters is  `()=> true`, we  won't list all the  5 parameter, right ?

brodybits: I think `any` would be easiest to search for. Any other ideas?

> if default value of argument function with 5 parameters is `()=> true`

I am not sure if I understand the question right. It does work for me if I would only keep the `@param` comment for the `fn` argument. But I think this would look like sloppy, unfinished work.

As a side point, I think it should not take too much work to typecheck most of the other sources once this update is finished. Then I think it would be nice to start typecheck with `"strict": true` on as many sources as possible, which I think would imply defining all arguments in the JSDoc comments.

thorn0: > we won't list all the 5 parameter, right ?

We will. Adding types mostly means adding types to function signatures.

fisker: You are saying we  might have code like

```js
function hasLeadingComment(node, fn = (_1, _2, _3, _4,_5) => true)
```

?

src/language-js/needs-parens.js
```diff
@@ -1,5 +1,6 @@
 "use strict";
 
+/** @type {import("assert")} */
```
thorn0, line 3: Why is this needed? `assert` is Node's builtin module. Its types should come from`@types/node`. `require("assert");` should just work.

src/language-js/comments.js
fisker: If we  have

```js
function foo(node, fn = returnTure) {}

function bar(node, fn = returnTure) {}
```

and these  two  `fn` takes different parameters, can  it pass types test ?

src/language-js/needs-parens.js
brodybits: I would get these errors without this type import, `yarn add --dev @types/node` does not seem to make any difference here:

```
src/language-js/needs-parens.js:302:13 - error TS2775: Assertions require every name in the call target to be declared with an explicit type annotation.

302             assert.strictEqual(parent.right, node);
                ~~~~~~~~~~~~~~~~~~

  src/language-js/needs-parens.js:3:7
    3 const assert = require("assert");
            ~~~~~~
    'assert' needs an explicit type annotation.
```

I guess this is something with the TypeScript compiler. Any ideas?

src/language-js/comments.js
thorn0: "adding types to function signatures" - I meant the JSDoc comments, not `_1, _2, ...`.

`() => true` *is* assignable to e.g. `(param1: number) => boolean`.

fisker: Thank you!

src/language-js/needs-parens.js
fisker: This throws since this PR I sent https://github.com/prettier/prettier/pull/8702/files

thorn0: This is a bug in TS: https://github.com/microsoft/TypeScript/issues/38379

thorn0: `/** @type {import("assert")} */` can stay for now, but we need to use `@types/node` instead of `@types/assert`. I'm not sure why the latter even exists.

src/language-js/printer-estree.js
```diff
@@ -672,8 +683,9 @@ function printPathNoParens(path, options, print, args) {
             options.locEnd
           );
           return (
+            nextCharacter &&
```
brodybits, line 686:
```suggestion
            nextCharacter !== false &&
```

I wish we had a good way to test this kind of case.

src/language-js/needs-parens.js
brodybits: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/assert/index.d.ts - seems to be for the Browserify assert package

brodybits:
I rebased again, now using `@types/node` instead of `@types/assert`, with a few minor updates:

- fix the mistake I made in `src/language-js/printer-estree.js`
- now using `@typedef {any} Options` rather than `@typedef {Object} Options` in src/language-js/print/ternary.js, with a TBD comment
- replace `/** @type {any} */` from PR #8702 with `/** @type {import("assert")} */` in `src/main/comments.js` & add a few more blank lines

There should now be no TODO comments introduced by this PR. I can think of the following followup items if this PR is merged:

- remove other `src/language-*` directories from the exclude list in `tsconfig.json`, and other `src` items from the exclude list, and continue resolving typecheck issues
- use `git grep` on src to check for `@ts-ignore` items to be resolved
- check for use of `any` type in JSDoc comments, which should ideally become more specific
- consider using `"strict": true` option for strict type checking on src files, starting with a very limited subset

fisker:
Ready?

brodybits:
This proposal is ready on my part. I just updated `@types/node` (again) and merged updates from master.

I had meant before that this proposal is ready for hopefully final review. There are some followup items that I think we should work on *after* this is merged.

There are quite a few commits, which would presumably result in a pretty long squash commit message. I could try to rebase yet again to combine some of the commit message lines, if absolutely necessary.

src/language-js/printer-estree.js
```diff
@@ -481,17 +486,23 @@ function printPathNoParens(path, options, print, args) {
       //   )
 
       const hasJSX = isJSXNode(n.right);
-      const firstGroupIndex = parts.findIndex((part) => part.type === "group");
+
+      const firstGroupIndex = parts.findIndex(
+        (part) => typeof part !== "string" && part.type === "group"
```
fisker, line 491: Do we  have to do  this ? sting.type can't  be `group`

src/language-js/comments.js
```diff
@@ -578,7 +578,7 @@ function handleCommentAfterArrowParams(text, enclosingNode, comment, options) {
     comment,
     options.locEnd
   );
-  if (text.slice(index, index + 2) === "=>") {
+  if (index !== false && text.slice(index, index + 2) === "=>") {
```
fisker, line 581: If we  change  non-exists index  to  `-1` (like `String #indexOf`) will that  a  breaking chang ?

src/language-js/printer-estree.js
brodybits: I added the `typeof part !== "string"` condition to resolve a JSDoc type error.

src/language-js/comments.js
brodybits: I don't know. I would actually favor using `-Infinity` to signify non-exists index since it should be safer against a mistake like code adding 1 to a non-exists index. And I would also favor making this kind of change after we get the JSDoc typecheck working on the affected modules.

fisker: But  in  js  world, `{String,Array}#indexOf` `Array#findIndex` all use  `-1`, I'm okay  with  `false` if it's breaking change .

src/language-js/printer-estree.js
fisker: I  known, I mean,  this  can't do in  other way ?

I  know nothing  about  JSDoc

brodybits:
> [...] mix types and fixes, separate PRs for git history

I think it means "do not mix adding JSDoc types & code fixes".

Moving the code fixes into one or more separate PRs would give us a cleaner git history, but would also imply:

- likely introducing many more `@ts-ignore` comments, which I would rather we avoid in practice moving forward
- the code fixes would be PR(s) that depend on this PR being reviewed & merged, which I think is not really supported so well by GitHub
- take time away from other things I could contribute to Prettier and other projects (feels like "busy work")

I would be happy to do this if absolutely necessary but am not convinced it would be really worth the time cost to me any possibly others.

src/language-js/comments.js
fisker: It's public api `src/common/utils-shared.js`, so we can't  change it, unless we  use `-1` internal and  expose  public one  differently, then  make breaking changes  in  `v3`.

evilebottnawi:
make sense

src/language-js/printer-estree.js
brodybits: I would have to investigate. I guess we could use `@ts-ignore` comment for now.

fisker: Will `part && part.type === "group"` work ? It make sense  to me .

src/language-js/comments.js
brodybits: > use `-1` internal and expose public one differently

Looks like extra complexity not needed  to me. Interesting idea, though.

We could just use `@ts-ignore` comment for now and raise a new PR to fix it with the breaking API changes.

fisker: Personal I prefer that way, but I think we can keep it your way, and we can change it to `!= -1`  when refactoring.

src/language-js/index.js
```diff
@@ -6,12 +6,14 @@ const estreeJsonPrinter = require("./printer-estree-json");
 const options = require("./options");
 
 const languages = [
+  // @ts-ignore - TBD (...)
```
fisker, line 9: I have one question, isn't it just json object? What exactly need to do?

As I said I don't use JSDoc, maybe this is a silly question.

src/language-js/comments.js
brodybits, line 581:
```suggestion

  // @ts-ignore
  if (text.slice(index, index + 2) === "=>") {
```

to back out the extra condition, ignore the error, and add a blank line to improve readability?

src/language-js/printer-estree.js
brodybits, line 491:
```suggestion
      @ts-ignore
      const firstGroupIndex = parts.findIndex((part) => part.type === "group");
```

to back out the change & ignore the type error - ??

src/language-js/comments.js
fisker: No, extra check is good.

src/language-js/printer-estree.js
brodybits: I think not because IMHO `part` could be a string.

fisker: As I said, even if it's string, there is nothing wrong checking `String.type` ...

I'm going to give up, this JSDoc thing is driving me crazy. :smile:

If other contributors think it's okay, I'm okay with it too.

Thank you for explaining this.

src/language-js/index.js
brodybits: I would get the following error:

```
src/language-js/index.js:9:26 - error TS2307: Cannot find module 'linguist-languages/data/JavaScript' or its corresponding type declarations.

9   createLanguage(require("linguist-languages/data/JavaScript"), (data) => ({
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I think it would be a pure JSON-like object, yes.

I am now wondering if there is something we could do to improve the way the language-specific data would be exported by `linguist-languages` and imported by this module.

fisker: This link https://github.com/microsoft/TypeScript/issues/7071  says  json  should not  be  an error, maybe  we  can  add `.json` extension?

src/language-js/printer-estree.js
brodybits: The whole JSDoc thing can be a bit crazy. There used to be very few users, now seem to be more and more. I would be happy to post some resources if you like, maybe someday later. We can see it is basically shoehorning quite a bit of TypeScript capabilities into documentation comments. I do really like that it cleanly separates type checking away from executable JavaScript code, can sure be a bit messy though :)

src/language-js/comments.js
brodybits: Keeping these extra checks then, thanks.

src/language-js/printer-estree.js
fisker: No... thanks , I don't want, I don't use typescript either.

brodybits:
Is there anything else needed from me?

I don't know if we can get this PR and maybe #8778 (typecheck fixes on language-handlebars) into 2.1, would be nice:)

src/language-js/index.js
brodybits: I tried using the JSON extension in all cases except one, hope it will be green.

I am still wondering if we could improve the import of the data from linguist-languages somehow:)

fisker: @brodybits It works? Can you send another PR to fix all `linguist-languages` import? It will be weird if we only import it with `.json` in js-related file, @evilebottnawi and me can have a quick review and merge it.

fisker: Also it would be great to enable eslint rule to force `.json` extension.

fisker: > I am still wondering if we could improve the import of the data from linguist-languages somehow:)

I think import needed JSON file is already the best way?

brodybits: I will contribute the eslint config & fix `linguist-languages` later today. Nice find.

fisker: Great, in that case, I'm going to sleep now. 0:32 in my place :smile:

brodybits: That hint to use eslint-plugin-import did not work for me. I suspect that is because we are using require to import the JSON items from `linguist-languages`. There is actually not such a large amount of work to start using typecheck on the other language sources, so I will work on this instead.

brodybits:
In addition to #8778, I just raised #8904, #8905, and #8906 to get typecheck working with the other language sources, with the JSON import updated as required by @fisker in a code comment. I am hopeful if they could be considered for the upcoming 2.1 release, will completely understand if this is not so practical.

fisker:
@brodybits You might misunderstood, I mean you can send PR to enfore import `.json` file with extensions, so we can merge it first. I didn't expect it related to typecheck. Thanks for all your hard work.

brodybits:
It did not work for me to enforce the import of .json files with the .json
extension, unfortunately. It is probably because we are using require. I
did already try to say this in a comment. But maybe I was doing something
wrong.
-- 
Sent from my mobile

fisker:
Sorry, I missed that. It should work, I'll take a look.

fisker:
You are right, https://github.com/benmosher/eslint-plugin-import/issues/1230

brodybits:
How important is it to enforce the import of JSON with the extension? Should we raise a new issue or maybe a draft PR?

It looks like someone did try to resolve that in <https://github.com/benmosher/eslint-plugin-import/pull/1237> but it has stalled over the past 1-2 years. I will try pinging them on the PR.

And is there anything else I can do to help with this PR?

fisker:
I made a eslint plugin for it #8908

brodybits:
> I made a eslint plugin for it #8908

Wow that was fast:)

I wonder if the solution could help some others outside Prettier maybe? Just an idea.

fisker:
If you think it should unless people use different `resolver` https://github.com/benmosher/eslint-plugin-import#resolvers

brodybits:
I think I got it. The custom plugin is only for require, not for ES6 import, and does not work with pluggable resolver. It is basically a workaround because of the missing functionality in <https://github.com/benmosher/eslint-plugin-import/issues/1230>. Nice work, too bad they don't fix the issue in the more general purpose eslint plugin.

I am leaving this in your hands now. Please give a holler if you need anything else. Thanks for supporting this update.

fisker:
> The custom plugin is only for require, not for ES6 import, and does not work with pluggable resolver. It is basically a workaround because of the missing functionality in benmosher/eslint-plugin-import#1230. 

Yap, that all what we need.

fisker:
I've  merged #8908

brodybits:
I think it should be possible to merge this PR, since we did the same thing with the `.json` extensions both here and in PR #8908. But it would give us extra, useless commit lines in the git history. And then the bigger impact is that I have listed commit messages for every change, to every file, which would probably be a bit too much. So it could take me a little time to group them more smartly.

I wish GitHub had an easy way for me to edit the commit message lines without having to rebase.

I have set this PR back to draft and will work on cleaning up the commit history.

Any other ideas?

fisker:
"Adds button to update a PR from the base branch to ensure it builds correctly before merging the PR itself."

https://github.com/sindresorhus/refined-github#editing-pull-requests

package.json
```diff
@@ -99,8 +100,8 @@
     "eslint-config-prettier": "6.11.0",
     "eslint-formatter-friendly": "7.0.0",
     "eslint-plugin-import": "2.22.0",
-    "eslint-plugin-prettier-internal-rules": "file:scripts/tools/eslint-plugin-prettier-internal-rules",
     "eslint-plugin-jest": "23.20.0",
+    "eslint-plugin-prettier-internal-rules": "file:scripts/tools/eslint-plugin-prettier-internal-rules",
```
brodybits, line 104: This happened when I tried re-adding the latest update of @types/node to devDependencies, would be happy to take it back out if needed.

brodybits:
I rebased a limited number of commits to remove the `.json` extension update not needed, merged from master, and re-applied the update of `@types/node` to `14.0.27`.

I expect and hope it will be a green build. I will work on the other typecheck proposals now.

Please let me know if anything else is needed from me on this PR. Thanks again.

package.json
fisker: It's on me, I named it `eslint-plugin-internal-rules` at first.

src/language-js/utils.js
```diff
@@ -11,6 +11,21 @@ const {
 } = require("../common/util");
 const handleComments = require("./comments");
 
+/**
+ * @typedef {import("estree").BaseNode} BaseNode
+ * @typedef {import("estree").BaseExpression} BaseExpression
+ * @typedef {import("estree").Expression} Expression
+ *
+ * @typedef {Object} MemberExpressionNode
```
fisker, line 19: There should be `MemberExpression` in `estree`

brodybits: I recall having some trouble using `MemberExpression`, will give it another try.

fisker: Maybe because `optional`? Or `typeAnnotation`? In that case, maybe we should use babel or eslint-typescript AST Type

src/language-js/utils.js
```diff
@@ -11,6 +11,20 @@ const {
...
+ * @typedef {import("estree").BaseExpression} BaseExpression
+ * @typedef {import("estree").Expression} Expression
+ * @typedef {import("estree").MemberExpression} MemberExpression
+ *
+ * @typedef {Object} MemberExpressionNode
+ * @mixes MemberExpression
+ *
+ * @typedef {Object} TemplateLiteralNode
+ * @mixes BaseNode
+ * @property {MemberExpressionNode[]} expressions
```
fisker, line 25: TemplateLiteral.expression could be  any  estree  node, I  don't think `MemberExpressionNode[]` is  right.

fisker: https://github.com/estree/estree/blob/master/es2015.md#templateliteral

src/language-js/utils.js
brodybits: Using `MemberExpression` gave me type errors like this:

```
src/language-js/utils.js:912:6 - error TS2367: This condition will always return 'false' since the types '"MemberExpression"' and '"Literal"' have no overlap.

912     (node.type === "Literal" && node.regex && node.regex.pattern) ||
```

I did change it to mix in `MemberExpression` and was able to get rid of the extra Expression properties.

I can give `@babel/types` a try. The eslint-typescript AST type seems to be specific to Typescript, or am I missing something there?

fisker: Both `eslint-typescript` and  `babel` has own AST  type , and  some of them  is  nit  standard estree node.

`eslint-typescript` mean to transform typescript node to  eslint node (which  is  estree, but  also not all the  same, eslint  don't have `StringLiteral`, they  use  `Literal` )

fisker: > Using `MemberExpression` gave me type errors like this:
> 
> ```
> src/language-js/utils.js:912:6 - error TS2367: This condition will always return 'false' since the types '"MemberExpression"' and '"Literal"' have no overlap.
> 
> 912     (node.type === "Literal" && node.regex && node.regex.pattern) ||
> ```
> 
> I did change it to mix in `MemberExpression` and was able to get rid of the extra Expression properties.
> 
> I can give `@babel/types` a try. The eslint-typescript AST type seems to be specific to Typescript, or am I missing something there?

This node is  a  eslint version  `RegExpLiteral` , it's type is  `Literal` and  should come from `eslint-typescript`

fisker: All our  js AST Nodes, none of them are  exactly estree, I'm not sure, but I think we need both  babel/eslint-typescript/flow

brodybits: I've been having trouble importing the types from eslint-typescript. Any pointers?

src/language-js/utils.js
brodybits, line 25: We could do this:

```suggestion
 * @property {Expression[]} expressions
```

But I am thinking it would be better if we can import the right types instead of doing `@typedef {Object}` with `@mixes`. I just set this PR back to draft state for now.

src/language-js/utils.js
fisker: Sorry, I don't have  this knowledge, can't help.

src/language-js/utils.js
```diff
@@ -11,6 +11,20 @@ const {
...
+ * @typedef {import("estree").BaseExpression} BaseExpression
+ * @typedef {import("estree").Expression} Expression
+ * @typedef {import("estree").MemberExpression} MemberExpression
+ *
+ * @typedef {object} MemberExpressionNode
+ * @mixes MemberExpression
+ *
+ * @typedef {object} TemplateLiteralNode
+ * @mixes BaseNode
+ * @property {Expression[]} expressions
```
brodybits, line 25: I have updated these lines to hopefully resolve the comments by @fisker.

I would like to make these types more specific, with help from `@babel/types` or `@typescript-eslint/types`, but still have not been able to get it to work.

Is it OK to continue with these typedefs for now and consider improving them in the future?

src/language-js/utils.js
fisker: https://github.com/typescript-eslint/typescript-eslint/blob/470174ad51fdb12d82129a896559075513f6c912/packages/types/src/ts-estree.ts#L141

This can't import?

src/language-js/utils.js
fisker: I don't think we need `MemberExpression`, why not just  `Node` from all parsers

src/language-js/utils.js
brodybits: This works for me so far:

```diff
- * @typedef {import("estree").BaseNode} BaseNode
+ * @typedef {import("@typescript-eslint/types").TSESTree.Node} Node
+ * @typedef {import("@typescript-eslint/types").TSESTree.BaseNode} BaseNode
```

I will try it for the other JSDoc types, hang on.

fisker: I don't think `BaseNode` is useful, just this `Node`, now import `Node` from `babel` and other parsers, union them all ad our `Node`.

I think it should work.

brodybits:
I was hoping to improve type JSDoc typing in `src/language-js/utils.js`. I have tried importing types with `import("@typescript-eslint/types").TSESTree` and I have tried union of Node from `{import("estree").Node | import("@babel/types").Node | import("@typescript-eslint/types").TSESTree.Node}`, nothing working for me.

I am thinking that there are some minor differences between `estree` types, `@babel/types`, `@typescript-eslint/types`, and the output of our parser that are seen by the `isSimpleTemplateLiteral` & `isSimpleCallArgument` utility functions.

At this point, I think the best solution would be to just declare the `node` parameter as `any` for `isSimpleTemplateLiteral` & `isSimpleCallArgument` and remove the typedefs I added to `src/language-js/utils.js`.

I think it would be ideal if we could go through the code in `src/language-js` and look for a way to resolve these differences. I don't know how long it would take me to go through this, though.

fisker:
I did a little work on this while you are working on this, I was going to commit...

brodybits:
@fisker can you push your branch somewhere? I would definitely like to see what you did.

fisker:
I'll send PR to your branch.

brodybits:
I am looking forward to the PR, may take my rest soon.

src/language-js/utils.js
```diff
@@ -12,39 +12,11 @@ const {
...
- * @typedef {import("@babel/types").OptionalMemberExpression} OptionalMemberExpression
- * @typedef {import("@babel/types").UnaryExpression} UnaryExpression
- * @typedef {import("@babel/types").TSNonNullExpression} TSNonNullExpression
- *
- * @typedef {Literal | BigIntLiteral | RegExpLiteral | DecimalLiteral | BooleanLiteral | NullLiteral | NumericLiteral | StringLiteral | Identifier | ThisExpression | Super | PrivateName | ArgumentPlaceholder | Import | TemplateLiteral | ObjectExpression | ArrayExpression | ImportExpression | CallExpression | OptionalCallExpression | NewExpression | MemberExpression | OptionalMemberExpression | UnaryExpression | TSNonNullExpression} ESNode
+ * @typedef {import("estree").Node} ESTreeNode
+ * @typedef {import("@babel/types").Node} BabelNode
+ * @typedef {import("@typescript-eslint/types").TSESTree.Node} TSNode
+ * @typedef {import("angular-estree-parser/lib/types").NGNode} NGNode
+ * @typedef {ESTreeNode | BabelNode | TSNode | NGNode} Node
```
fisker, line 19: I wounder if we can union every ESNode in `estree.d.ts` and do `import('./estree.d.ts').Node`.

Something like

```js
import * as ESTree from 'estree'
import * as Babel from '@babel/types'
...

export type Node = ESTree.Node | Babel.Node | ...;

// A automatic loop would be nice.

```

brodybits: TBH I am not so sure how much I like the idea of mixing union of Node types from all different estree-like type packages. It feels a little sloppy to me.

I think the obvious root is that *internally* we have some kind of Frankenstein-like hybrid of ES tree node types that come in from the various parsers and go through our language-js processing.

I would favor getting a solution as clean and simple as we can for now and look into how we can improve the formal interface at some point in the future.

I am now working to see if I can resolve the `@ts-ignore` item.

brodybits: With the following change we can resolve & get rid of the `@ts-ignore`:

```diff
- * @typedef {ESTreeNode | BabelNode | TSNode | NGNode} Node
+ * @typedef {BabelNode | TSNode | NGNode} Node
```

I will do this and a little more cleanup now.

fisker: > With the following change we can resolve & get rid of the `@ts-ignore`:
> 
> ```diff
> - * @typedef {ESTreeNode | BabelNode | TSNode | NGNode} Node
> + * @typedef {BabelNode | TSNode | NGNode} Node
> ```
> 
> I will do this and a little more cleanup now.

I don't think it's right, `estree` Literal can have `regex` property

brodybits:
I pushed some minor updates, getting ready to take my rest. Please let me know if there is anything else needed. Thanks!

fisker:
I'm sending a new commit.

brodybits:
+1 (+100) for adding more types

TBH I am not so thrilled about adding a TS file, even a d.ts file to what had been a pure-JavaScript project. I think it would be ideal if we could find a way to extract the d.ts types into a reusable package someday. But maybe this is the best we can do for now. Off to rest soon.

src/language-js/utils.js
```diff
@@ -115,7 +134,7 @@ function getLeftSidePathName(path, node) {
   if (node.expression) {
     return ["expression"];
   }
-  throw new Error("Unexpected node has no left side", node);
+  throw new Error("Unexpected node has no left side.");
```
brodybits, line 137: I just discovered this change and am a little confused by it.

I had thought it would be nice if the exception would show the `node` object that is malformed if we ever get to this point.

src/language-js/utils.js
```diff
@@ -895,6 +1006,7 @@ function isSimpleCallArgument(node, depth) {
   const plusTwo = (node) => isSimpleCallArgument(node, depth + 2);
 
   const regexpPattern =
+    // @ts-ignore
```
brodybits, line 1009: This `@ts-ignore` does not seem to be needed if we do this to `src/language-js/types/estree.d.ts`:

```diff
-export type Node = ESTree.Node | Babel.Node | TSTree.Node | NGTree.NGNode;
+export type Node = Babel.Node | TSTree.Node | NGTree.NGNode;
```

brodybits:
I just pushed some updates to add some blank lines, cleanup comments for 2 functions, and resolve the one remaining `@ts-ignore` item in this PR.

src/language-js/utils.js
fisker: `Error` constructor only take one  argument.

src/language-js/utils.js
fisker: `ESTree.Node` stays see https://github.com/prettier/prettier/pull/8759#discussion_r467690623 , you  might  not  noticed we  didn't add Flow Node, because `estree/flow.d.ts` is  broken as I tried.

brodybits: > see [#8759 (comment)](https://github.com/prettier/prettier/pull/8759#discussion_r467690623)

says this, for the sake of easier reference for myself:

> `estree` Literal can have `regex` property.
> 
> https://github.com/estree/estree/blob/master/es5.md#regexpliteral

Here are the type errors I get if we keep `ESTree.Node`:

```console
src/language-js/utils.js:1013:38 - error TS2339: Property 'regex' does not exist on type 'SimpleLiteral | RegExpLiteral | BigIntLiteral | BooleanLiteral | NumberLiteral | NullLiteral | RegExpLiteral | StringLiteral'.
  Property 'regex' does not exist on type 'SimpleLiteral'.

1013     (node.type === "Literal" && node.regex && node.regex.pattern) ||
                                          ~~~~~

src/language-js/utils.js:1013:52 - error TS2339: Property 'regex' does not exist on type 'SimpleLiteral | RegExpLiteral | BigIntLiteral | BooleanLiteral | NumberLiteral | NullLiteral | RegExpLiteral | StringLiteral'.
  Property 'regex' does not exist on type 'SimpleLiteral'.

1013     (node.type === "Literal" && node.regex && node.regex.pattern) ||
                                                        ~~~~~
```

This type issue sure looks ugly to me.

TBH I still don't really understand the need for `ESTree.Node`, considering the Node types would be mostly if not completely covered by Babel and TSTree. But I can put it back, maybe with a TODO comment, if needed.

One more thing is that I am not sure if I follow what language-js does with Angular. I can see that language-js deals with Angular embedded in a Component decorator factory, am I missing anything else major here?

fisker: > Here are the type errors.

I mean something wrong in other place.

fisker: > I still don't really understand the need for ESTree.Node

Because `ESTree.Node` is `Node`, we'll use it in other places.

fisker: > One more thing is that I am not sure if I follow what language-js does with Angular. I can see that language-js deals with Angular embedded in a Component decorator factory, am I missing anything else major here?

I don't understand this question

brodybits: I just reverted my change and put `ESTree.Node` back, and put the `@ts-ignore` comment back as well. I am very hopeful that we can resolve the type issue someday.

And my question was simply to understand what language-js does with Angular. I can research it another day.

Please let me know if there is anything else needed.

fisker: `angular-estree-parser` is also a js parser https://github.com/prettier/prettier/blob/6833310f87c597f0a204b06f2e84cb1b3b33ff54/src/language-js/parser-angular.js#L7

src/language-js/print/call-arguments.js
```diff
@@ -159,7 +159,7 @@ function printCallArguments(path, options, print) {
       shouldBreakForArrowFunction;
 
     // We want to print the last argument with a special flag
-    let printedExpanded;
+    let printedExpanded = [];
```
brodybits, line 162: Are we sure `printedExpanded` is no longer empty by the time `printedExpanded[0]` is accessed by the code below?

fisker: Yes, https://github.com/prettier/prettier/blob/a66e43a09319bf13beda421d00c04c6228a4ccba/src/language-js/print/call-arguments.js#L38

src/language-js/utils.js
```diff
@@ -26,6 +45,10 @@ const FLOW_SHORTHAND_ANNOTATION = new RegExp(
 );
 const FLOW_ANNOTATION = new RegExp(`^${NON_LINE_TERMINATING_WHITE_SPACE}*::`);
 
+/**
+ * @param {Node & PrettierEsNode} node
```
fisker, line 49: @brodybits Do you think this is the right way? I'm really new to this.

brodybits: I am having a little trouble following some of the types in estree.d.ts. I would personally favor keeping these changes as simple as possible in this PR due to its size.

At some point I would like to try the typecheck in strict mode (set `"strict": true` in compilerOptions in `tsconfig.json`), which could be another major effort across all of the code in Prettier.

I am thinking the typecheck fixes in language-js is the largest set of changes we will need to get typecheck working throughout our source code.

fisker: I don't have experience on typing, please fix what you think is wrong, glad I learn a lot.

brodybits: This change seems to work for me, but I am not sure whether or not it is what we want:

```diff
- * @param {Node & PrettierEsNode} node
+ * @param {PrettierEsNode} node
```

I would probably need some time to look through the types in `eslint.d.ts`, look through the functions in `util.js`, and look through the code where these functions are called to better understand what is going on with these types.

My original purpose was to simply get typecheck working on src/language-js without too many `@ts-ignore` comments. It would be ideal to get typecheck working in strict mode, and declare specific types on all functions, but I think this should be a different effort. I hope we don't have to add anything else to this PR.

brodybits:
I tried updating the `PrettierEsNode`, hope it is OK.

src/language-js/types/estree.d.ts
```diff
@@ -44,12 +45,12 @@ type PrettierEsComment = Comment & {
   leading?: boolean;
 };
 
-export interface PrettierEsNode {
+export type PrettierEsNode = Node & {
```
fisker, line 48: I did try this, but failed on my machine.

fisker: This is good!

fisker: Can we export this as `Node`?

brodybits:
> Can we export this as Node?
>

I tried something like that and encountered some type issues. I am out now,
can give it another try in the next 1-2 hours.

> --
Sent from my mobile

src/language-js/types/estree.d.ts
brodybits: Yes I have done that now, but it probably needs to be fixed better.

src/language-js/types/estree.d.ts
```diff
@@ -0,0 +1,55 @@
...
+  printed?: boolean;
+  trailing?: boolean;
+  leading?: boolean;
+};
+
+export type Node = AnyNode & {
+  extra?: any;
+  comments?: PrettierEsComment[];
+  trailingComments?: any;
+  // leadingComments?: PrettierEsComment[];
```
brodybits, line 52: @fisker I hacked the entries for `trailingComments` & `leadingComments` due to some type issues. I think these need to be fixed and will look into this. I did push these changes in case you may have any other ideas at the same time.

brodybits:
I have updated the exported Node type and renamed the d.ts file to `ast.d.ts`, hope it is OK now.

src/language-js/types/estree.d.ts
fisker: have you try  `Array<PrettierEsComment | Comment>`?

fisker:
Why rename? They are ESTree https://github.com/prettier/prettier/blob/5e71bb37bc085d2b5ac1b6e6374ffaf7f7130c50/src/language-js/parser-babel.js#L218

src/language-js/types/ast.d.ts
```diff
@@ -0,0 +1,55 @@
+import * as ESTree from "estree";
+import * as Babel from "@babel/types";
+import * as TSESTree from "@typescript-eslint/types/dist/ts-estree";
+import * as NGTree from "angular-estree-parser/lib/types";
+
+type AnyNode = ESTree.Node | Babel.Node | TSESTree.Node | NGTree.NGNode;
+
+export type TemplateLiteral =
```
fisker, line 8: Can we add

```js
 & {
  extra?: any;
  comments?: PrettierEsComment[];
  trailingComments?: ReadonlyArray<PrettierEsComment>;
  leadingComments?: ReadonlyArray<PrettierEsComment>;
}
```

to all them

src/language-js/types/estree.d.ts
brodybits: > `Array<PrettierEsComment | Comment>`

I think that would be no different from `Array<PrettierEsComment>`, since all members added to `PrettierEsComment` are optional.

I discovered that using `ReadonlyArray` is needed because of the types in `@babel/types`.

fisker: I  checked our  comment attach logic, seems  we are  not  touching `leadingComments`, can  we  remove?

src/language-js/types/ast.d.ts
brodybits: yes, coming up

src/language-js/types/estree.d.ts
brodybits: yes (I did discover the same thing)

src/language-js/types/ast.d.ts
fisker: I  mean "all"

brodybits: I will work on this, hang on

brodybits:
I have encountered something really weird. If I would update `@typescript-eslint/types` to 3.9.0 in devDependencies, `yarn lint:typecheck` works fine for me.

But if I would merge from master, with update of `@typescript-eslint/typescript-estree` to 3.9.0, then `yarn lint:typecheck` gives me the following errors:

```console
node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/estree-to-ts-node-types.d.ts:112:45 - error TS2694: Namespace 'ts' has no exported member 'NamedTupleMember'.

112     [AST_NODE_TYPES.TSNamedTupleMember]: ts.NamedTupleMember;
                                                ~~~~~~~~~~~~~~~~

node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/estree-to-ts-node-types.d.ts:120:55 - error TS2694: Namespace 'ts' has no exported member 'NamedTupleMember'.

120     [AST_NODE_TYPES.TSRestType]: ts.RestTypeNode | ts.NamedTupleMember;
                                                          ~~~~~~~~~~~~~~~~

node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/ts-nodes.d.ts:3:918 - error TS2694: Namespace 'ts' has no exported member 'NamedTupleMember'.

3 export declare type TSNode = ts.Modifier | ts.Identifier | ts.QualifiedName | ts.ComputedPropertyName | ts.Decorator | ts.TypeParameterDeclaration | ts.CallSignatureDeclaration | ts.ConstructSignatureDeclaration | ts.VariableDeclaration | ts.VariableDeclarationList | ts.ParameterDeclaration | ts.BindingElement | ts.PropertySignature | ts.PropertyDeclaration | ts.PropertyAssignment | ts.ShorthandPropertyAssignment | ts.SpreadAssignment | ts.ObjectBindingPattern | ts.ArrayBindingPattern | ts.FunctionDeclaration | ts.MethodSignature | ts.MethodDeclaration | ts.ConstructorDeclaration | ts.SemicolonClassElement | ts.GetAccessorDeclaration | ts.SetAccessorDeclaration | ts.IndexSignatureDeclaration | ts.KeywordTypeNode | ts.ImportTypeNode | ts.ThisTypeNode | ts.ConstructorTypeNode | ts.FunctionTypeNode | ts.TypeReferenceNode | ts.TypePredicateNode | ts.TypeQueryNode | ts.TypeLiteralNode | ts.ArrayTypeNode | ts.NamedTupleMember | ts.TupleTypeNode | ts.OptionalTypeNode | ts.RestTypeNode | ts.UnionTypeNode | ts.IntersectionTypeNode | ts.ConditionalTypeNode | ts.InferTypeNode | ts.ParenthesizedTypeNode | ts.TypeOperatorNode | ts.IndexedAccessTypeNode | ts.MappedTypeNode | ts.LiteralTypeNode | ts.StringLiteral | ts.OmittedExpression | ts.PartiallyEmittedExpression | ts.PrefixUnaryExpression | ts.PostfixUnaryExpression | ts.NullLiteral | ts.BooleanLiteral | ts.ThisExpression | ts.SuperExpression | ts.ImportExpression | ts.DeleteExpression | ts.TypeOfExpression | ts.VoidExpression | ts.AwaitExpression | ts.YieldExpression | ts.SyntheticExpression | ts.BinaryExpression | ts.ConditionalExpression | ts.FunctionExpression | ts.ArrowFunction | ts.RegularExpressionLiteral | ts.NoSubstitutionTemplateLiteral | ts.NumericLiteral | ts.BigIntLiteral | ts.TemplateHead | ts.TemplateMiddle | ts.TemplateTail | ts.TemplateExpression | ts.TemplateSpan | ts.ParenthesizedExpression | ts.ArrayLiteralExpression | ts.SpreadElement | ts.ObjectLiteralExpression | ts.PropertyAccessExpression | ts.ElementAccessExpression | ts.CallExpression | ts.ExpressionWithTypeArguments | ts.NewExpression | ts.TaggedTemplateExpression | ts.AsExpression | ts.TypeAssertion | ts.NonNullExpression | ts.MetaProperty | ts.JsxElement | ts.JsxOpeningElement | ts.JsxSelfClosingElement | ts.JsxFragment | ts.JsxOpeningFragment | ts.JsxClosingFragment | ts.JsxAttribute | ts.JsxSpreadAttribute | ts.JsxClosingElement | ts.JsxExpression | ts.JsxText | ts.NotEmittedStatement | ts.CommaListExpression | ts.EmptyStatement | ts.DebuggerStatement | ts.MissingDeclaration | ts.Block | ts.VariableStatement | ts.ExpressionStatement | ts.IfStatement | ts.DoStatement | ts.WhileStatement | ts.ForStatement | ts.ForInStatement | ts.ForOfStatement | ts.BreakStatement | ts.ContinueStatement | ts.ReturnStatement | ts.WithStatement | ts.SwitchStatement | ts.CaseBlock | ts.CaseClause | ts.DefaultClause | ts.LabeledStatement | ts.ThrowStatement | ts.TryStatement | ts.CatchClause | ts.ClassDeclaration | ts.ClassExpression | ts.InterfaceDeclaration | ts.HeritageClause | ts.TypeAliasDeclaration | ts.EnumMember | ts.EnumDeclaration | ts.ModuleDeclaration | ts.ModuleBlock | ts.ImportEqualsDeclaration | ts.ExternalModuleReference | ts.ImportDeclaration | ts.ImportClause | ts.NamespaceImport | ts.NamespaceExportDeclaration | ts.ExportDeclaration | ts.NamedImports | ts.NamedExports | ts.ImportSpecifier | ts.ExportSpecifier | ts.ExportAssignment | ts.SourceFile | ts.Bundle | ts.InputFiles | ts.UnparsedSource | ts.JsonMinusNumericLiteral | ts.JSDoc | ts.JSDocTypeExpression | ts.JSDocUnknownTag | ts.JSDocAugmentsTag | ts.JSDocClassTag | ts.JSDocEnumTag | ts.JSDocThisTag | ts.JSDocTemplateTag | ts.JSDocReturnTag | ts.JSDocTypeTag | ts.JSDocTypedefTag | ts.JSDocCallbackTag | ts.JSDocSignature | ts.JSDocPropertyTag | ts.JSDocParameterTag | ts.JSDocTypeLiteral | ts.JSDocFunctionType | ts.JSDocAllType | ts.JSDocUnknownType | ts.JSDocNullableType | ts.JSDocNonNullableType | ts.JSDocOptionalType | ts.JSDocVariadicType | ts.JSDocAuthorTag;
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       ~~~~~~~~~~~~~~~~


Found 3 errors.
```

I think this is an issue with `@typescript-eslint/typescript-estree`, not with this update, but I would consider this to be a major blocker and would like to find a solution before continuing with work on this PR.

This also means that I would not want to merge from master before getting this resolved.

While it would be possible to downgrade to `@typescript-eslint/typescript-estree` 3.8.0, I would not really favor this kind of solution.

src/language-js/types/ast.d.ts
```diff
@@ -0,0 +1,71 @@
+import * as ESTree from "estree";
+import * as Babel from "@babel/types";
+import { TSESTree } from "@typescript-eslint/types";
+import * as NGTree from "angular-estree-parser/lib/types";
```
brodybits, line 4: I think it should be supported to import types like `NGNode` directly from the `angular-estree-parser` package, like I did for `TSESTree` on the line above.

I have an idea how to fix this on `angular-estree-parser`, will give it a try.

fisker: I don't like `import { TSESTree } from`, same  reason  should prefer  `import identity from "lodash/indentity"`  over  `import {identity} from "lodash"`

fisker: Personlly, I even prefer `package-name/path/to/module.mjs`

brodybits: I can revert it, don't understand though.

fisker: That won't be nessary, not important.

fisker:
Saw  this https://github.com/prettier/prettier/pull/8759#issuecomment-672059792?

brodybits:
> Saw this [#8759 (comment)](https://github.com/prettier/prettier/pull/8759#issuecomment-672059792)?

OK so Prettier basically uses its own "estree" format which can handle estree or estree-like AST structure that comes out of the Babel, Flow, or typescript-estree parser modules.

I am wondering about a couple things, though:

- We had to add what look like some non-standard fields to some AST node types to get the function type checks to work properly. And we added these fields to all other exported "estree" AST node types.
- The output of the Babel, Flow and typescript-estree parsers are processed by postprocess.js before returned by the parse function and are ready to be piped into the printer function.

I will probably take my rest pretty soon. Unfortunately I have some other priorities that I had put off for a while and may not be able to contribute much more over the next 1-2 weeks or so. I hope we can finish this PR without too much more work now and consider improving the type check in future PRs.

brodybits:
Latest estree.d.ts updates in 5414a629cbe40cf9489bcc901d0ba26a81dccb18 look good to me, thanks.

src/language-js/types/estree.d.ts
```diff
@@ -0,0 +1,97 @@
...
+import * as Babel from "@babel/types";
+import { TSESTree } from "@typescript-eslint/types";
+import * as NGTree from "angular-estree-parser/lib/types";
+
+// WORKAROUND HACK for typescript-eslint issue:
+// https://github.com/typescript-eslint/typescript-eslint/issues/2388
+// MUST REMOVE to avoid errors with TypeScript 4.0.0:
+import * as ts from "typescript";
+declare module "typescript" {
+  type NamedTupleMember = Node;
```
ExE-Boss, line 11: This will work even with **TypeScript 4.x**:
```suggestion
  export interface NamedTupleMember extends Node {}
```

brodybits: Thanks @ExE-Boss. I would actually favor keeping this workaround as-is, errors can remind us to remove it with the TypeScript 4 update. Nice idea, though.

brodybits: We should be able to remove this entire workaround if <https://github.com/typescript-eslint/typescript-eslint/pull/2405> is merged and released. I hope this will be possible.

fisker:
Good job

brodybits:
Teamwork yeah!

PR #8760, opened 2020-07-15
use strict number JSDoc type in src/common/util.js

brody2consult:
by using `-Infinity` instead of `false`

expected benefits:

- avoid some of the type issues encountered in PR #8759
- may be faster on engines like V8
- may be less likely for weird, unexpected behavior to show up

I would actually like to wait for PR #8759 to be reviewed first. In case PR #8759 is accepted and merged, it should be possible to remove some of the `@ts-ignore` TODO items as well.

- ~~I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).~~

### DIFF
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -21,7 +21,7 @@ function getPenultimate(arr) {
 
 /**
  * @param {string | RegExp} chars
- * @returns {(text: string, index: number | false, opts?: SkipOptions) => number | false}
+ * @returns {(text: string, index: number, opts?: SkipOptions) => number}
  */
 function skip(chars) {
   return (text, index, opts) => {
@@ -29,8 +29,8 @@ function skip(chars) {
 
     // Allow `skip` functions to be threaded together without having
     // to check for failures (did someone say monads?).
-    if (index === false) {
-      return false;
+    if (index < 0) {
+      return -Infinity;
     }
 
     const { length } = text;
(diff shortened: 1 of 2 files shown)


thorn0:
I don't think this should be changed. `number | false` is a safer type than just `number`. It makes the typechecker check that there is no missing checks for the `false` case.

evilebottnawi:
> may be faster on engines like V8

it is micro-optimizations, don't think it has really benefits on real projects

brodybits:
Understood. I did try the following change to work around a type issue in `src/language-js/comments.js`:

```diff
diff --git a/src/language-js/comments.js b/src/language-js/comments.js
index a039e808d..c6e5389d3 100644
--- a/src/language-js/comments.js
+++ b/src/language-js/comments.js
@@ -573,11 +573,18 @@ function handleCommentAfterArrowParams(text, enclosingNode, comment, options) {
     return false;
   }
 
-  const index = getNextNonSpaceNonCommentCharacterIndex(
+  const maybeIndex = getNextNonSpaceNonCommentCharacterIndex(
     text,
     comment,
     options.locEnd
   );
+
+  if (maybeIndex === false) {
+    return false;
+  }
+
+  const index = /** @type{number} */ (maybeIndex);
+
   if (text.slice(index, index + 2) === "=>") {
     addDanglingComment(enclosingNode, comment);
     return true;
```

I guess we would have to make some more changes like this to resolve the type issues with the `number | false` values. Seems to add a little more complexity, though probably a little safer.

fisker:
Some of these are public api, should be careful.

thorn0:
@brodybits The type assertion shouldn't be needed after that `if`. The type should be narrowed automatically. Doesn't that happen?

brodybits:
> Some of these are public api, should be careful.

Got it.

> This change should suffice:

Yes it does, and I will apply it to PR #8759.

So it looks to me like we have a monad-like pattern of using `number | false` type in `src/common/util.js`. Do we know of any other projects using this kind of value type pattern?

fisker:
Looks like maintainers are not happy with this change.

PR #8765, opened 2020-07-15
Refactor: Move `parsers` to language index

fisker:
Move `parsers` to language index for easier maintenance. 

(When I first  try to add a parser, I don't even know the parsers need add to `common/internal-parsers.js`.)

This also make `language-xxx/index.js` a "completed" plugin, previously `index.js` + `parser-xxxx.js` is the whole plugin (standalone build still in this way).
 


tsconfig.json
```diff
@@ -42,6 +42,7 @@
     // in other sources:
     "src/index.js",
     "src/standalone.js",
+    "src/languages.js",
```
brodybits, line 45: - I think we can remove `src/common/internal-plugins.js` from the list.
- I think `src/languages.js` should move a few lines up to the section with "JavaScript sources affected by JSDoc issues in src/language-*"

brodybits:
+1 (+100) overall

> This also make `language-xxx/index.js` a "completed" plugin

:tada:

Do we think the standalone build will be updated someday?

Any interesting effect on the external API?

fisker:
> Do we think the standalone build will be updated someday?

Maybe with `import('./parser-xxx.mjs')`, when we support asyc api and use ESModule.

> Any interesting effect on the external API?

There should not.

tsconfig.json
fisker: Will do.

PR #8783, opened 2020-07-17
Do not trim `options.originalText`

fisker:
Trim `originalText` doesn't make sense to me.


src/language-markdown/utils.js
```diff
@@ -193,7 +193,7 @@ function hasGitDiffFriendlyOrderedList(node, options) {
 // workaround for https://github.com/remarkjs/remark/issues/351
 // leading and trailing newlines are stripped by remark
 function getFencedCodeBlockValue(node, originalText) {
-  const text = originalText.slice(
+  const text = originalText.trimEnd().slice(
```
fisker, line 196: Finally, I understand why we still need this workaround.

fisker: Fixing this in https://github.com/prettier/prettier/pull/8786

thorn0:
Is this superseded by #8786?

fisker:
Ah, closed by accident, I prefer merge this first, #8786 has some spec issue need discuss.

PR #8786, opened 2020-07-18
Markdown: Fix missing blank lines in fenced code block

fisker:


fisker:
Problem in this PR:

For Value of code block without end mark, there are difference between CommonMark and Remark.

In this case.

````markdown
```
a
````

If we add line breaks to the end.

CommonMark:

node.value didn't count the final `\n`.

Remark:

node.value include all `\n`s to EOF.

And `remark` believe the remark AST is right.

More details https://github.com/remarkjs/remark/issues/512

Which one should we follow?

**Comment by fisker** [comment]
Let's keep current behavior (the  CommonMark way)?

**Comment by evilebottnawi** [comment]
I think it is safe to change

**Comment by fisker** [comment]
The real question is "Do we want to change"?

Github also render

````markdown
<!-- end with 1 blank line -->
```
a


````

with no blank line,

and

````markdown
<!-- end with two blank lines -->
```
a



````

with one blank line

evilebottnawi:
To be honest, I have no strong opinion here

fisker:
Me too, so let's follow CommonMark  and GitHub!

evilebottnawi:
:+1:

evilebottnawi (approved):

sosukesuzuki (approved):

PR #8793, opened 2020-07-19
Mention Apheleia in list of Emacs plugins

raxod502:
I am the maintainer of the Emacs package [Apheleia](https://github.com/raxod502/apheleia) which allows users to apply code formatters asychronously on save without (for the most part) moving the cursor when code is reformatted. Prettier is supported out of the box. This pull request adds Apheleia to the list of suggested Emacs plugins that have support for Prettier.

### DIFF
--- a/website/data/editors.yml
+++ b/website/data/editors.yml
@@ -9,7 +9,7 @@
   content: |
     [`prettier-js`](https://github.com/prettier/prettier-emacs)
     [`prettier.el`](https://github.com/jscheid/prettier.el)
-    [Apheleia](https://github.com/raxod502/apheleia)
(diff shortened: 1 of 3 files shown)


website/data/editors.yml
```diff
@@ -9,6 +9,7 @@
   content: |
     [`prettier-js`](https://github.com/prettier/prettier-emacs)
     [`prettier.el`](https://github.com/jscheid/prettier.el)
+    [Apheleia](https://github.com/raxod502/apheleia)
```
fisker, line 12: Other links wrap with `<code>`, please add `` ` `` too

raxod502: No problem!

PR #8805, opened 2020-07-21
Update `@typescript-eslint/typescript-estree` to v3.7.1

fisker:
[`typescript-eslint/typescript-estree@v3.7.0` release note](https://github.com/typescript-eslint/typescript-eslint/releases)

Support type annotations on catch clauses.

Logical assignment seems requires `typescript@4`


fisker:
`TSNullKeyword` seems removed in https://github.com/typescript-eslint/typescript-eslint/commit/cbaac775c66fc74277eeecd2a6f9fcff2a3329c6

`node.type === "Literal" && node.value === null` should be enough for it.

But we always use `isXXXLiteral` to test `Literal`, so I add `isNullLiteral` and also check `NullLiteral`(babel).

src/language-js/printer-estree.js
```diff
@@ -4997,7 +4997,7 @@ function shouldHugType(node) {
         n.type === "VoidTypeAnnotation" ||
         n.type === "TSVoidKeyword" ||
         n.type === "NullLiteralTypeAnnotation" ||
-        n.type === "TSNullKeyword"
```
thorn0, line 5000: `babel-ts` generates `TSNullKeyword` nodes. Also are we sure these changes are expected? I don't see anything about this in the changelog of `typescript-estree`.

fisker: I didn't test locally, waiting for CI result.

This change didn't mentioned, but tests failed. https://github.com/prettier/prettier/runs/892776097#step:7:1263

#8806 also failed.

fisker: You are right, `TSNullKeyword` can't be removed. Restored.

thorn0: @bradzacher It's not expected that the AST format changed for the `null` type literal (TS 3.9.7), is it?

bradzacher: No, but I'm just now realising that I probably introduced a regression for <4.0.0 with the latest version. I'll submit a fix to undo it.

fisker: Will it  change for typescript 4?

fisker: Thank you!

fisker:
AST change reverted, Waiting for `@typescript-eslint/typescript-estree` new release.

bradzacher:
you can test this against the `canary` version that'll be published in a few minutes.
full release will be Monday morning PST

fisker:
Read to merge.

PR #8810, opened 2020-07-21
Simplify `composeLoc`

fisker:


src/language-js/parser-babel.js
```diff
@@ -38,6 +38,7 @@ function babelOptions({ sourceType, extraPlugins = [] }) {
       ...extraPlugins,
     ],
     tokens: true,
+    ranges: true,
```
fisker, line 41: Babel missing `range` in comments, so we still need `node.start`

https://github.com/babel/babel-eslint/blob/d63d594fc6c602d20c4db96c58d728d3938c1552/lib/babylon-to-espree/convertComments.js#L11

PR #8821, opened 2020-07-23
Fix range tests with `TEST_CRLF` and `endOfLine` option

fisker:
Snapshot is a little difference on windows, found in #8811 

CI log https://github.com/prettier/prettier/runs/901140719#step:7:1310


tests_config/run_spec.js
```diff
@@ -187,15 +188,23 @@ global.run_spec = (fixtures, parsers, options) => {
             codeForSnapshot = visualizeRange(codeForSnapshot, baseOptions);
             codeOffset = codeForSnapshot.match(/^>?\s+1 \| /)[0].length;
           }
+          let resultForSnapshot = formattedWithCursor;
 
           if (hasEndOfLine) {
+            // We snapshot the linux version
+            if (mainOptions.endOfLine === "auto" && os.EOL === "\r\n") {
```
thorn0, line 195: This part looks suspicious. `--end-of-line auto` doesn't mean "use `os.EOL`". It means 
> Maintain existing line endings (mixed values within one file are normalised by looking at what’s used after the first line)

fisker: Thanks, I misunderstood.

fisker:
I have another test to fix for #8811,  after this one merge.

PR #8831, opened 2020-07-24
CLI: Exit on invalid options

fisker:
ESLint use `--fix` to fix errors, Jest use `-u` to update snapshots.

Some times, when I was busy, I type the wrong flags, like `npx prettier foo.js --fix`, the CLI will output formatted code.

We should exit on invalid options.

```console
jest --fisker
● Unrecognized CLI Parameter:

  Unrecognized option "fisker". Did you mean "filter"?

  CLI Options Documentation:
  https://jestjs.io/docs/en/cli.html
```

```console
eslint --fisker
Invalid option '--fisker' - perhaps you meant '--parser'?
```

Now, 

```console
node bin/prettier --fisker
Invalid option "--fisker".
```

Eslint use [`optionator`](https://github.com/gkz/optionator) to do this, I haven't check Jest.

We might want print a better message too.

### DIFF
--- a/src/cli/util.js
+++ b/src/cli/util.js
@@ -995,7 +995,14 @@ function normalizeContextArgv(context, keys) {
   const argv = !keys ? context.argv : pick(context.argv, keys);
 
   context.argv = optionsNormalizer.normalizeCliOptions(argv, detailedOptions, {
-    logger: context.logger,
+    logger: {
+      warn(...args) {
+        if (context.logger) {
+          context.logger.error(...args);
+        }
+        process.exit(1);
+      },
+    },
   });
 }
 //------------------------------context-util-end--------------------------------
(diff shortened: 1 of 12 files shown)


fisker:
I'll try later.

src/cli/util.js
```diff
@@ -995,7 +995,14 @@ function normalizeContextArgv(context, keys) {
...
   context.argv = optionsNormalizer.normalizeCliOptions(argv, detailedOptions, {
-    logger: context.logger,
+    logger: {
+      warn(...args) {
+        if (context.logger) {
+          context.logger.error(...args);
+        }
+        process.exit(1);
+      },
+    },
```
fisker, line 1005: @evilebottnawi You know what? Turns out, we only need change this from `warn` to `error`. But those tests..., really a mess to fix.

fisker:
Not done. I'm about to give up.

fisker:
abandoned

PR #8834, opened 2020-07-25
Exclude experimental parsers(glimmer) from `prettier .`

### LINKED ISSUE #8631: `prettier .` shouldn't include `*.handlebars`
Running `prettier --check .` for a folder with an unsupported file type fails with errors saying `No parser could be inferred for file`, while `prettier --write .` finishes successfully in my project.

Please note that issue https://github.com/prettier/prettier/issues/2691 resolved the problem for `--write` option, but `prettier --check` still works the old way.

**Environments:**

- Prettier Version: 2.0.5
- Running Prettier via: CLI
- Operating System: macOS

**Steps to reproduce:**

Create a folder with a .handlebars file (or probably virtually any file extension which is not supported) and run `prettier --check .`

**Expected behavior:**
prettier finishes successfully

**Actual behavior:**
prettier finishes with an error:
`
src/playground/src/views/main.handlebars[error] No parser could be inferred for file: src/playground/src/views/main.handlebars
`

fisker:
I didn't expect `showUnreleased` to `false` to work, but tests all good, maybe because we don't have any other "unreleased things".

Fixes #8631


fisker:
Why this is breaking change, it fixes #8631

fisker:
On a second thought, maybe you are right.

If use have this config.

```js
{
  "overrides": [
    {
      "files": "*.hbs",
	  "options": {
        "parser": "glimmer"
	  }
    }
  ]
}
```

evilebottnawi:
@fisker yep, I'm talking about it

fisker:
@evilebottnawi Let's wait to see what @thorn0 think. And @czerwinskilukasz1 .

czerwinskilukasz1:
> @evilebottnawi Let's wait to see what @thorn0 think. And @czerwinskilukasz1 .

Hi @fisker ,
Thank you for taking care of the issue I raised! 
Well, I am not familiar with prettier internals, so it's hard for me to give specific advice. One point I would like to make is given that `prettier --write .` worked fine for a folder containing a `.handlebars` file, but `prettier --check .` resulted in an error, I would look for differences between how these two commands are run and fix the issue there. As I said, I'm not familiar with the code, so totally no idea if your change fixes this discrepancy.

fisker:
#8631 already closed.

PR #8848, opened 2020-07-27
Change test package name

fisker:
`prettier/local` is designed to avoid moduleNameMapper misconfigure, [here](https://github.com/prettier/prettier/pull/4554#discussion_r191072725), avoid load prettier from stable version.

But now we have `prettier/standalone` and `standalone.js` exists in stable version,  I think changing package names to `prettier-local` and `prettier-standalone` is more safe.


lipis (approved):

lipis:
Seems safe to me! Shall we merge?

PR #8878, opened 2020-07-31
cleanup: remove extra parent.object conditions

brodybits:
from src/language-js/needs-parens.js

(seems to be not needed)

- ~~I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).~~


fisker:
Very bad idea,I'm on cellphone, but I'm almost sure there are needed, just not covered. MemberExpression have `.property` child.

We should add tests.

thorn0 (approved):

fisker:
@thorn0 You think is not needed?

thorn0:
Yes. `name === "object"` and `parent.object === node` check the same thing.

fisker:
Oh, yes, you're right, I didn't check what name is. 

Sorry about this @brodybits

fisker:
@brodybits How did you notice that it's not needed?

fisker (approved):

sosukesuzuki (approved):

brodybits:
I found one case when comparing changes between a couple forks of Prettier, found the other case from a simple search:)

PR #8903, opened 2020-08-04
feat: condense graphql docstring into a single line if possible

joual:
This work allows for GraphQL docstrings to be contained into a single line if there were no linebreaks (and if the line is < than printWidth). 

### DIFF
--- a/src/language-graphql/printer-graphql.js
+++ b/src/language-graphql/printer-graphql.js
@@ -160,9 +160,9 @@ function genericPrint(path, options, print) {
       if (n.block) {
         return concat([
           '"""',
-          hardline,
+          n.value.includes("\n") ? softline : "",
           join(hardline, n.value.replace(/"""/g, "\\$&").split("\n")),
-          hardline,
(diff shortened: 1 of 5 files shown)


joual:
cc @vjeux

kachkaev:
Keeping `"""This is a description"""` as is is debatable, but I'm not sure if it's a good idea to convert `"""\nThis is a description\n"""` to `"""This is a description"""`. I personally would not like to see this diff in my existing projects.

I think we should discuss this topic in an issue before there is a PR.

joual:
@kachkaev I agree, is there any way for me to detect that it was `"""\ndescription\n"""`?

vjeux:
Seems reasonable to me, but I haven’t worked in the graphql ecosystem much, I don’t know what the customs are. @leeb do you have any thoughts?

sixmen:
I hope this PR accepted.

Multiline graphql comments makes the code ugly in general. (And inconsistent with most tools)

PR #8778, opened 2020-07-17
typecheck fixes in src/language-handlebars

brodybits:
- add `@glimmer/reference` to devDependencies, as needed to avoid some "cannot find module" typecheck errors
- quick JSDoc type workaround in `src/language-handlebars/postprocess.js`
- `@ts-ignore` comments, including a comment in `src/common/parser-create-error.js` which is duplicated from PR #8759
- remove `src/language-handlebars` from `exclude` list in `tsconfig.json`

TODO items:

- [x] ~~This update is blocked by a type error in `@glimmer/validator` due to <https://github.com/glimmerjs/glimmer-vm/issues/1116>. I have not found a way to ignore this kind of an error coming from a package dependency, and I would appreciate any ideas or suggestions.~~
- ~~I think PR #8759 should be merged first, if accepted, and remove `src/common/parser-create-error.js` update from this PR~~

I am also not so happy that I had to add an extra Glimmer dependency to resolve some "cannot find module" typecheck errors, hope they can improve the situation between `@glimmer/syntax` and `tsc` (TypeScript compiler).

~~I am raising this PR as a draft PR now to give Glimmer some more context for <https://github.com/glimmerjs/glimmer-vm/issues/1116>.~~

- ~~I’ve added tests to confirm my change works.~~
- ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~~
- ~~(If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.~~
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #8904, opened 2020-08-04
Enable and fix typecheck  in language-html

brodybits:
__updated:__

- remove `src/language-html` from `tsconfig.json`
- ~~import `linguist-languages items` with `.json` extension, as suggested by @fisker in comments on PR #8759~~
- import `assert` type in `src/language-html/printer-html.js`
- apply some quick type fixes & workaround solutions
- ignore other type errors in `src/language-html`
- ~~ignore type issue in `src/common/parser-create-error.js`, which is duplicate of a change in PRs #8759 & #8778~~

~~The one major drawback is that using `@ts-ignore` comments to ignore this many type issues is not such good practice.~~

_At this point there is a limited number of `@ts-ignore` comments._


src/language-html/parser-html.js
```diff
@@ -339,7 +354,7 @@ function createParser({
   normalizeAttributeName = false,
   allowHtmComponentClosingTags = false,
   isTagNameCaseSensitive = false,
-  getTagContentType,
+  getTagContentType = null,
```
fisker, line 357: I don't think this is right, default value of `getTagContentType` should be `undefined`, `null` don't make sense.

brodybits: Done, hope it doesn't break my build

brodybits: Results in this error now:

```
  ✘  unicorn/no-useless-undefined

     Do not use useless `undefined`
```

I am looking into some alternatives now.

brodybits: I just pushed a more radical soution:

```diff
--- a/src/language-html/parser-html.js
+++ b/src/language-html/parser-html.js
@@ -348,24 +348,9 @@ function locEnd(node) {
   return node.sourceSpan.end.offset;
 }
 
-function createParser({
-  recognizeSelfClosing = false,
-  normalizeTagName = false,
-  normalizeAttributeName = false,
-  allowHtmComponentClosingTags = false,
-  isTagNameCaseSensitive = false,
-  getTagContentType,
-} = {}) {
+function createParser(parserOptions = {}) {
   return {
-    parse: (text, parsers, options) =>
-      _parse(text, options, {
-        recognizeSelfClosing,
-        normalizeTagName,
-        normalizeAttributeName,
-        allowHtmComponentClosingTags,
-        isTagNameCaseSensitive,
-        getTagContentType,
-      }),
+    parse: (text, parsers, options) => _parse(text, options, parserOptions),
     hasPragma,
     astFormat: "html",
     locStart,
```

fisker: Current one  is  absolutely not  right , we  may  pass  extra option to  createParser.

This  type check ,to me , looks like  saying `cont  {foo} = {}` , foo  must  have a default value , isn't `foo` already a  `undefined`?

brodybits: I am not sure if I am understanding your comments correctly. The primary purpose of the simplification is to avoid issues between JSDoc typecheck & existing eslint rules. I would also favor not keeping another internal options interface layer if it is not really needed.

Does the proposed simplification look OK, or are there any issues with it?

brodybits:
I made some potentially controversial changes to reduce the number of ignored type issues down to 4 (plus one overlap with PR #8759).

src/language-html/parser-html.js
fisker: 1. `createParser({getTagContentType = null,})` This could break  `angular-html-parser`, if it do `foo({getTagContentType = somethingNotNull})`

2. `createParser({getTagContentType = undefined,})` this `undefined` seems useless

3. `createParser(parserOptions)` is not right, it's not `parserOptions`, it's could be anything, for example we could pass `prepocess` function

brodybits: I just backed these changes out and added a couple more `@ts-ignore` comments. I think it would be nice to improve it someday later.

brodybits:
I just reverted one of the changes and merged from master, git history starting to become ugly again. I hope someone can clean it up if this PR would be accepted and merged.

src/language-html/ast.js
```diff
@@ -7,6 +7,21 @@ const NODES_KEYS = {
...
+    this.index = undefined;
+    this.siblings = undefined;
+    this.prev = undefined;
+    this.next = undefined;
+    this.parent = undefined;
+
+    this.name = undefined;
+    this.namespace = undefined;
+    this.hasExplicitNamespace = undefined;
+    this.children = undefined;
```
fisker, line 23: This looks so silly...

src/language-html/parser-html.js
```diff
@@ -132,21 +133,23 @@ function ngHtmlParser(
...
+        ? "comment"
+        : node instanceof DocType
+        ? "docType"
+        : node instanceof Element
+        ? "element"
+        : node instanceof Text
+        ? "text"
+        : (() => {
+            throw new Error(`Unexpected node ${JSON.stringify(node)}`);
+          })();
```
fisker, line 152: And this

fisker:
What do you mean about git history, we always do `squash and merge`.

src/language-html/ast.js
brodybits: I did put the rationale in the comments. I would like to enable JSDoc type checking without excessive `@ts-ignore` comments and without too much overhaul at this point. Any better ideas?

src/language-html/parser-html.js
brodybits: Rationale is in the comment. Any better ideas?

brodybits:
GitHub seems to put all of the commit messages into squash and merge, looks like you did take it out of the other PR. OK.

src/language-html/ast.js
fisker: I have no idea.

fisker: The constructor seems taking something like `import('angular-html-parser/lib/compiler/src/ml_parser/ast').Node`, but not exactly the same, I don't know how to make it work.

src/language-html/parser-html.js
fisker: i tried to add to this function, seems working

```js
  /**
  * @typedef {import('angular-html-parser/lib/compiler/src/ml_parser/ast').Node} Node
  * @typedef {object} WithType
  * @property {string} type
  * @typedef {Node & WithType} NodeWithType
  * @param {NodeWithType} node
  */
```

, but I also tried

```js
  /**
  * @typedef {object} WithType
  * @property {string} type
  * @param {WithType} node
  */
```

also work, I don't really know what those things mean...

brodybits: I think `@typedef {object}` is very similar to `@typedef {any}`, they may even be the same thing. (I recall @thorn0 said something about this in another review.) I would actually favor using `any`, to make it easier for us to find and improve at some point in the future.

I tried this to be a little more specific but it did not work:

```js
/**
 * @typedef {import('angular-html-parser/lib/compiler/src/ml_parser/ast').Node} WithType
 * @property {string} type
 * @param {WithType} node
 */
```

I am now thinking it would be best to keep the JSDoc typedefs near the beginning of the module, will work on this.

fisker: But I think it's just Node with an extra type, I just don't know how to do it.

fisker: Or maybe it's Element or something, there are several types , one or some together should work

brodybits:
I would say that this PR is held up by <https://github.com/microsoft/TypeScript/issues/26811>.

The thing is that the properties in the ast class are not listed explicitly in the constructor. It should be possible to define these properties using JSDoc tags, but this does not work due to <https://github.com/microsoft/TypeScript/issues/26811>.

It should be possible to define the properties in a d.ts file, but according to the discussion in <https://github.com/microsoft/TypeScript/issues/26811> this works with ES modules, not with CommonJS modules. And I think this should not be the justification for suddenly moving from CommonJS modules to ES modules.

I did try adding some attributes to the constructor, but it results in silly looking code.

I don't know what else we can do now.

brodybits:
I have rebased and pushed some updates to make the properties more explicit in the ast Node constructor, as required due to <https://github.com/microsoft/TypeScript/issues/26811>, and resolve most of the type issues in parser-html.js.

tsconfig.json
```diff
@@ -22,7 +22,6 @@
     "src/main/multiparser.js",
     "src/main/options.js",
     // [TBD] src/language-* source directory trees with known JSDoc type issues:
```
brodybits, line 24:
```suggestion
```

(comment should not be needed with these changes)

src/language-html/printer-html.js
```diff
@@ -1060,14 +1064,10 @@ function printEmbeddedAttributeValue(node, originalTextToDoc, options) {
     const value = getValue();
     if (!value.includes("{{")) {
       return printExpand(
-        textToDoc(
-          value,
-          {
-            parser: "css",
-            __isHTMLStyleAttribute: true,
-          },
-          { stripTrailingHardline: true }
```
fisker, line 1069: Note this

brodybits: Yes: `lint:typecheck` reported that the last argument `{ stripTrailingHardline: true }` was not needed in multiple cases, and this seems to be right for `textToDoc` within the `printEmbeddedAttributeValue` function block. Prettier then reformats the other lines.

fisker: I think tests will fail, after we merge #9094

brodybits: tsc thinks that the last argument `{ stripTrailingHardline: true }` is an extra argument in these function calls; When I looked at the code I that this was right.

I tried the change from #9094 and have not seen any test failures so far in tests/html, tests/css, tests/vue, or tests/angular. I will now try running the full test suite with the change from #9094.

fisker: That's wired

brodybits: I reverted these changes and added `@ts-ignore` comments. I am running out of ideas on my side about these function calls.

thorn0: `textToDoc` takes only 2 arguments there:
https://github.com/prettier/prettier/blob/23004319a01babadc1e7661091538b4b2110047d/src/language-html/printer-html.js#L1048-L1053

fisker: Ah, it's another function...

brodybits: So it looks like `textToDoc` has 2 meanings in `src/language-html/printer-html.js`:

- The internal `embed` function receives `textToDoc` as an argument and calls `textToDoc` with 4 arguments;
- Internal `printEmbeddedAttributeValue` function has an inner `textToDoc` function that expects 2 arguments

Any explanations why this would be the case and any ideas how we can avoid this kind of confusion?

thorn0: To avoid confusion, use Ctrl+click, or F12, or hover, or Ctrl+hover to see the type and/or definition of what is called.

fisker: There is no problem with the original implementation, it's  on me, I was not careful enough.

brodybits: I have now renamed `textToDoc` to `attributeTextToDoc` in the `printEmbeddedAttributeValue` function, to avoid getting us tripped up in the future, and removed the extra parameter again (reformatted with Prettier).

brodybits:
Hi is there anything else I can do to help move this proposal forward? This is blocking the typecheck on `src/cli` files, which I think is needed before we refactor the CLI context in PR #9964.

In terms of `src/language-html/ast.js`, my proposal has ugly workarounds for <https://github.com/microsoft/TypeScript/issues/26811>, and I am not so sure it is the right way to go. Here are some alternatives I can think of:

- ALT-A: keep my ugly workaround for <https://github.com/microsoft/TypeScript/issues/26811>
- ALT-B: see if we can refactor the AST Node object into a simpler object type
- ALT-C: convert this one source module into a TypeScript module
- ALT-D: move the AST Node class into a separate TypeScript package

I think ALT-A and ALT-B would be the simplest.

I would downvote ALT-C since it would add to the build and overall project complexity.

I think ALT-D would be the most elegant. I did also notice that the Node AST class was reused by the Stryker Mutator, likely with some adaptations. It would be interesting to see if anyone else could reuse this Node AST class in the future.

fisker:
I don't know what to  do, but I don't like `ALT-B`, `ALT-C`, and `ALT-D`.  Can we just ignore this file?

brodybits:
> Can we just ignore this file?

I just added **ALT-E** to "just ignore this file". This was my original idea but I was thrown off by this review comment: https://github.com/prettier/prettier/pull/8904#pullrequestreview-461559056

brodybits:
I just rebased, reverted some undesired changes to `src/language-html/ast.js`, and added some `@ts-ignore` comments to that module. Is there anything else needed to finish this PR?

fisker (approved):

brodybits:
I just resolved a couple merge conflicts and added an extra blank line after a `require` statement. Please let me know if there is anything else needed. Thanks!

thorn0 (approved):
I tweaked a couple of things. In particular, `@type {import("assert")}` seems to be not need anymore. Let's merge it this way for a start.

As the next step, I think we have to switch the parser in `.prettierrc` back to `babel` as `meriyah` doesn't support JSDoc type annotations, which in some cases makes adding proper types impossible.

brodybits:
Thanks for integrating this proposal!

PR #8906, opened 2020-08-04
enable & fix typecheck for several languages

brodybits:
__updated__ - now includes updates for `src/language-markdown` from PR #8905

- remove the following entries from `tsconfig.json`:
  - `src/language-css`
  - `src/language-graphql`
  - `src/language-markdown`
  - `src/language-yaml`
- ~~require the `linguist-languages` data with `.json` extension in each affected subdirectory tree~~
- ignore align type issues in `src/language-markdown` & `src/language-yaml/printer-yaml.js` (similar align type issue as in `src/language-markdown` in PR #8905)
- ignore type issue in `src/common/parser-create-error.js` (duplicate of change in PRs #8859, #8878, #8904)

Recommended followup items after this and other language typecheck proposals are reviewed & merged:

- resolve align issues ignored in PR #8905 this proposal
- start moving individual JavaScript modules out of the exclude list in `tsconfig.json`


brodybits:
I have included the changes from PR #8905 (typecheck for language-markdown) and rebased onto master, looking forward to the review.

fisker:
@brodybits I update the `align` param from `number` to `number | string`, and removed all `@ts-ignore`.

brodybits:
Perfect @fisker, thanks. I took the liberty to rename this PR, feel free to rename it again if needed.

fisker:
I took a look at the source, `align` also accept `object` , but I don't know if we should add it and I don't know how.

fisker:
It takes everthing

https://github.com/prettier/prettier/blob/6793811da2d2304b711da08833e3da704074cbbf/src/document/doc-printer.js#L22-L32

src/document/doc-builders.js
```diff
@@ -56,7 +56,7 @@ function indent(contents) {
 }
 
 /**
- * @param {number} n
+ * @param {number | string} n
```
brodybits, line 59: How about this:

```suggestion
 * @param {any} n
```

This is assuming that we would plan to do `git grep` for JSDoc `any` types and make them more specific at some point in the future.

fisker: I send a PR rewrite this part https://github.com/prettier/prettier/pull/8931/files

Do you think you can make it specific in this PR?

brodybits: I think you mean adding specific types to PR #8931, right?

I would like to keep this PR simple, since it is to enable typecheck on several (4) languages.

Maybe we should keep this PR in draft until we fix the types in `align` and `makeAlign`?

fisker: No, I mean this PR, #8931 only mean to make logic cleaner  for you.

fisker: > I would like to keep this PR simple, since it is to enable typecheck on several (4) languages.

But this PR requires a fix for `align` function, and what I did is not good enough.

fisker: > Maybe we should keep this PR in draft until we fix the types in align and makeAlign?

Let's only take care of `align`.

brodybits: After reading and experimenting with some of the code, I think this is what we need:

```suggestion
 * @param {number | string | Doc} n
```

I think this is the one place that the Doc type option is needed for: https://github.com/prettier/prettier/blob/9dad95b35f935edce4c3d6cfa45c79a0b9c82b9f/src/document/doc-builders.js#L104-L106

I am now pushing changes to apply this suggestion and get rid of that old `@ts-ignore` comment.

fisker: What  about  `if(!n)`, this  contains many  possibility. could be  `undefined` `null` `false` `NaN`.

fisker: Let's ignore it for now, I've add `TODO` there.

src/document/doc-builders.js
```diff
@@ -56,7 +56,8 @@ function indent(contents) {
 }
 
 /**
- * @param {number} n
+ * TODO: add missing types for paramter `n`
```
brodybits, line 59: I am not convinced about this TODO. After searching through the source code, I could only see the following types used for `n`: `number`, `string`, `Doc` ... and I only see `Doc` type used in one place for `n`.

And I only see the same thing for the `makeAlign` function.

I had initially missed the `Doc` type option since it was ignored by a `@ts-ignore` comment. I just did another code search and do not find any other `@ts-ignore` comments related to the `align` function call.

fisker: You fogot one fact, it's public.

fisker: This is why we can't change `!n` to `n === 0` or `n === false`

brodybits: > it's public

as documented here, for the sake of extra clarity:

- https://github.com/prettier/prettier/blob/master/commands.md#align

I am now thinking this part of the documentation could use a little improvement, will take another look sometime.

src/document/doc-builders.js
```diff
@@ -56,7 +56,8 @@ function indent(contents) {
...
+ * @param {number | string | Doc} n
```
brodybits, line 60: Could we do something like this to be a little more consistent with the internal `makeAlign` function:

```suggestion
 * @param {number | string | Doc | false} n
```

src/document/doc-builders.js
fisker: I didn't see the doc, so the doc says `number | string`, but we use `{type: "root"}` ourself...I don't know what to do.

src/document/doc-builders.js
fisker: If the  documentation says `number | string` it probably mean `0` not `false`,  we should make it clear, but maybe later?

brodybits: yup

src/document/doc-builders.js
brodybits: > we use `{type: "root"}` ourself

here (to be extra-clear): https://github.com/prettier/prettier/blob/2.0.5/src/document/doc-builders.js#L104-L106

and we know that `{type: "root"}` would satisfy the requirements to be a `Doc` object, which I had defined maybe too loosely here: https://github.com/prettier/prettier/blob/2.0.5/src/document/doc-builders.js#L4-L11

So we could even do this (does not give me type error), though I would probably not recommend it:

```diff
- * @param {number | string | Doc} n
+ * @param {number | Doc} n
```

I would favor that we keep this PR as-is, with the TODO comment, and see if we can improve it someday later.

fisker: I think it's not `Doc`, just a special object with `root: true` let's use `number | string`, keep it same as the  documentation, and restore the one in `markAsRoot`(use  ts-ignore). Make sense?

brodybits: Agreed, I was thinking the same thing about the type.

fisker:
Thanks for taking care of this.

Issue #5322, opened 2018-11-01
Support for XML / SVG


remcohaszing:
It would be great if prettier could also check and format XML and SVG files.

The formatting rules would be exactly the same as for JSX.

Specificly for SVG, the contents of a `<style />` tag would be formatted as CSS.

#3491 was marked as a duplicate of #1882, but #1882 is only about HTML, not XML.

**Input:**

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="120" height="120">
<defs ><style>.foo {color: cyan}</style></defs>
    <rect fill="#e8e8e8" height="108" rx="12" width="108" x="0" y="4"></rect>
  </svg>
```

**Output:**

N/A

**Expected behaviour:**

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="120" height="120">
  <defs>
    <style>
      .foo {
        color: cyan;
      }
    </style>
  </defs>
  <rect fill="#e8e8e8" height="108" rx="12" width="108" x="0" y="4" />
</svg>
```

evilebottnawi:
@remcohaszing :+1: But i think better implement this after officially implementation `html`, i think we can use same parser for xml based syntax

remcohaszing:
I didn’t think about whitespace when I initially created this issue. When this is taken into account, there isn’t really a point in applying prettier on XML documents, unless of course, the meaning of the document is taken into account. To detect the meaning of an XML document, prettier should interpret the xmlns attribute of the root element.

In many cases whitespace is not interpreted. In this case I think nodes should be formatted by prettier the same way as it formats JSX. If it’s unknown how a node is interpreted, prettier should keep the original spacing.

### SVG

I think SVG is by far the most popular human readable XML format. An SVG document is a document which implements the http://www.w3.org/2000/svg (or https) namespace. (Please correct me if I’m wrong.)

A nice list of all known SVG elements can be found on https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg.

I fiddled a bit with SVG spacing in Google Chrome and concluded whitespace is entirely ignored, so it’s safe to format SVG following the exact same rules as for JSX.

There are some special tags that can be interpreted by prettier in a special way.

The `<style>` tag contains CSS. Since prettier can already interpret CSS, its contents can also be formatted by prettier.

[...]

PR #8907, opened 2020-08-04
XML/SVG support

kddeisz:
Closes https://github.com/prettier/prettier/issues/5322. 

This pulls https://github.com/prettier/plugin-xml into core, adding support for XML and SVG.

Not sure if I got everything, so let's work through the PR together!

### DIFF
--- a/fixture.xml
+++ b/fixture.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "https://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+<?xml-model href="project.rnc" type="application/relax-ng-compact-syntax"?>
+<!-- foo -->
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  xmlns:xlink="http://www.w3.org/1999/xlink"
+  width="200"
+  height="100"
+  viewBox="0 0 200 100"
+>
+  <title>Style inheritance and the use element</title>
+  <desc _attr="attr">
+    &anp; &#12345;
+    <![CDATA[
+      foo
+    ]]>
+    bar
+  </desc>
+  <?pagebreak?>
+  <style></style>
+  <style type="text/css">
+    circle          { stroke-opacity: 0.7; }
+    .special circle { stroke: green; }
+    use             { stroke: purple;
+                      fill: orange; }
+  </style>
+  <?pagebreak?>
+    <g class="special" style="fill: blue">
+        <circle id="c" cy="50" cx="50" r="40" stroke-width="20" />
+    </g>
+  <use xlink:href="#c" x="100" />
+  <ignored>
+    <!-- prettier-ignore-start -->
+      <   ignored   />
+    <!-- prettier-ignore-end -->
+  </ignored>
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed at est eget enim consectetur accumsan. Aliquam pretium sodales ipsum quis dignissim. Sed id sem vel diam luctus fringilla. Aliquam quis egestas magna. Curabitur molestie lorem et odio porta, et molestie libero laoreet. Morbi rhoncus sagittis cursus. Nullam vehicula pretium consequat. Praesent porta ante at posuere sollicitudin. Nullam commodo tempor arcu, at condimentum neque elementum ut.
+  </p>
+  <span>
+    content
+  </span>
+  <div>
+    even more
+    <content />
+  </div>
+</svg>
+<!-- bar -->
--- a/src/language-xml/options.js
+++ b/src/language-xml/options.js
@@ -1,0 +1,26 @@
+"use strict";
+
+module.exports = {
+  xmlSelfClosingSpace: {
+    type: "boolean",
+    category: "XML",
+    default: true,
+    description: "Adds a space before self-closing tags.",
+  },
+  xmlWhitespaceSensitivity: {
+    type: "choice",
+    category: "XML",
+    default: "strict",
+    description: "How to handle whitespaces in XML.",
+    choices: [
+      {
+        value: "strict",
+        description: "Whitespaces are considered sensitive in all elements.",
+      },
+      {
+        value: "ignore",
+        description: "Whitespaces are considered insensitive in all elements.",
+      },
+    ],
+  },
+};
--- a/src/main/core-options.js
+++ b/src/main/core-options.js
@@ -156,6 +156,7 @@ const options = {
         since: "1.17.0",
         description: "Lightning Web Components",
       },
+      { value: "xml", since: "0.0.0", description: "XML" },
     ],
   },
   plugins: {
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@glimmer/syntax": "0.55.3",
     "@iarna/toml": "2.2.5",
     "@typescript-eslint/typescript-estree": "3.8.0",
+    "@xml-tools/parser": "1.0.7",
     "angular-estree-parser": "2.1.0",
     "angular-html-parser": "1.7.1",
     "camelcase": "6.0.0",
--- a/scripts/build/config.js
+++ b/scripts/build/config.js
@@ -75,6 +75,9 @@ const parsers = [
   {
     input: "src/language-yaml/parser-yaml.js",
   },
+  {
+    input: "src/language-xml/parser-xml.js",
+  },
 ].map((parser) => ({
   type: "plugin",
   target: "universal",
--- a/src/language-xml/index.js
+++ b/src/language-xml/index.js
@@ -1,0 +1,35 @@
+"use strict";
+
+const createLanguage = require("../utils/create-language");
+const printer = require("./printer-xml");
+const options = require("./options");
+
+const languages = [
+  createLanguage(require("linguist-languages/data/XML"), () => ({
+    since: "0.0.0",
+    parsers: ["xml"],
+    vscodeLanguageIds: ["xml", "forcesourcemanifest"],
+  })),
+  createLanguage(require("linguist-languages/data/SVG"), () => ({
+    since: "0.0.0",
+    parsers: ["xml"],
+    vscodeLanguageIds: ["svg"],
+  })),
+];
+
+const printers = {
+  xml: printer,
+};
+
+const parsers = {
+  get xml() {
+    return require("./parser-xml");
+  },
+};
+
+module.exports = {
+  languages,
+  options,
+  printers,
+  parsers,
+};
--- a/src/language-xml/parser-xml.js
+++ b/src/language-xml/parser-xml.js
@@ -1,0 +1,36 @@
+"use strict";
+
+const { parse: xmlToolsParse } = require("@xml-tools/parser");
+
+const parse = (text) => {
+  const { lexErrors, parseErrors, cst } = xmlToolsParse(text);
+
+  if (lexErrors.length > 0 || parseErrors.length > 0) {
+    throw new Error("Error parsing XML");
+  }
+
+  return cst;
+};
+
+const locStart = (node) => {
+  if (node.location) {
+    return node.location.startOffset;
+  }
+  return node.startOffset;
+};
+
+const locEnd = (node) => {
+  if (node.location) {
+    return node.location.endOffset;
+  }
+  return node.endOffset;
+};
+
+const parser = {
+  parse,
+  astFormat: "xml",
+  locStart,
+  locEnd,
+};
+
+module.exports = parser;
--- a/src/language-xml/printer-xml.js
+++ b/src/language-xml/printer-xml.js
@@ -1,0 +1,386 @@
+"use strict";
+
+const {
+  concat,
+  fill,
+  group,
+  hardline,
+  indent,
+  join,
+  line,
+  literalline,
+  softline
+} = require("../document").builders;
+
+const ignoreStartComment = "<!-- prettier-ignore-start -->";
+const ignoreEndComment = "<!-- prettier-ignore-end -->";
+
+const hasIgnoreRanges = (comments) => {
+  if (!comments || comments.length === 0) {
+    return;
+  }
+
+  comments.sort((left, right) => left.offset - right.offset);
+
+  let startFound = false;
+  for (let idx = 0; idx < comments.length; idx += 1) {
+    if (comments[idx].image === ignoreStartComment) {
+      startFound = true;
+    } else if (startFound && comments[idx].image === ignoreEndComment) {
+      return true;
+    }
+  }
+
+  return false;
+};
+
+const whitespaceIgnorable = (node) => {
+  const { CData, Comment, reference } = node.children;
+
+  return !CData && !reference && !hasIgnoreRanges(Comment);
+};
+
+const printCData = (node) => ({
+  offset: node.startOffset,
+  printed: node.image
+});
+
+const printComment = (node) => ({
+  offset: node.startOffset,
+  printed: node.image
+});
+
+const printCharData = (path, print) => (node, index) => ({
+  offset: node.location.startOffset,
+  printed: path.call(print, "children", "chardata", index)
+});
+
+const printElement = (path, print) => (node, index) => ({
+  offset: node.location.startOffset,
+  printed: path.call(print, "children", "element", index)
+});
+
+const printProcessingInstruction = (node) => ({
+  offset: node.startOffset,
+  printed: node.image
+});
+
+const printReference = (node) => ({
+  offset: node.location.startOffset,
+  printed: (node.children.CharRef || node.children.EntityRef)[0].image
+});
+
+const replaceNewlinesWithLiteralLines = (content) =>
+  concat(
+    content
+      .split(/(\n)/g)
+      .map((value, idx) => (idx % 2 === 0 ? value : literalline))
+  );
+
+const nodes = {
+  attribute: (path) => {
+    const { Name, EQUALS, STRING } = path.getValue().children;
+
+    return concat([Name[0].image, EQUALS[0].image, STRING[0].image]);
+  },
+  chardata: (path) => {
+    const { SEA_WS, TEXT } = path.getValue().children;
+    const [{ image }] = SEA_WS || TEXT;
+
+    return concat(
+      image
+        .split(/(\n)/g)
+        .map((value, index) => (index % 2 === 0 ? value : literalline))
+    );
+  },
+  content: (path, opts, print) => {
+    const {
+      CData = [],
+      Comment = [],
+      chardata = [],
+      element = [],
+      PROCESSING_INSTRUCTION = [],
+      reference = []
+    } = path.getValue().children;
+
+    let children = []
+      .concat(CData.map(printCData))
+      .concat(Comment.map(printComment))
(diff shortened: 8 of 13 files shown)


src/main/core-options.js
```diff
@@ -156,6 +156,7 @@ const options = {
         since: "1.17.0",
         description: "Lightning Web Components",
       },
+      { value: "xml", since: "0.0.0", description: "XML" },
```
MichaelDeBoey, line 159:
```suggestion
      { value: "xml", since: "2.1.0", description: "XML" },
```

fisker:
I don't think we need any files in `language-xml`, the `@prettier/plugin-xml` is already a full functional plugin(language).

Maybe add `require("@prettier/plugin-xml")` to `src/languages.js`, and little adjustments in build script will work.

fixture.xml
```diff
@@ -0,0 +1,50 @@
...
+                      fill: orange; }
+  </style>
+  <?pagebreak?>
+    <g class="special" style="fill: blue">
+        <circle id="c" cy="50" cx="50" r="40" stroke-width="20" />
+    </g>
+  <use xlink:href="#c" x="100" />
+  <ignored>
+    <!-- prettier-ignore-start -->
+      <   ignored   />
```
ExE-Boss, line 36: Won’t `<` not immediately followed by [`NameStartChar`][XML-NT-NameStartChar] be a syntax error?

[XML-NT-NameStartChar]: https://www.w3.org/TR/REC-xml/#NT-NameStartChar

src/language-xml/options.js
```diff
@@ -0,0 +1,26 @@
...
+      {
+        value: "strict",
+        description: "Whitespaces are considered sensitive in all elements.",
+      },
+      {
+        value: "ignore",
+        description: "Whitespaces are considered insensitive in all elements.",
+      },
+    ],
+  },
```
evilebottnawi, line 25: We need discussion on these options

/cc @prettier/core 

I think the `xmlSelfClosingSpace` option is unnecessary

fisker: @evilebottnawi I don't think we need, it's still a plugin, we should let plugin decide. We can just ship it together with Prettier.

evilebottnawi: @fisker SVG is part of web and we don't close this issue in favor plugin because we want to move it into core, integration HTML + SVG will be painful with plugins due architecture problems, fixes in plugins will be painful too

lydell: > I think the xmlSelfClosingSpace option is unnecessary

I agree! That’s exactly the type of option Prettier should _not_ have. We don’t teams arguing over if they should pick true or false. We should always format like we do in HTML in this case.

Btw – I think it’s called “whitespace”, not “whitespaces”.

thorn0:
:-1: I see no point in merging the plugin into core. Besides, SVG is already supported by core as it's part of HTML.

evilebottnawi:
@thorn0 

> -1 I see no point in merging the plugin into core. Besides, SVG is already supported by core as it's part of HTML.

But this is not so. We don't even have tests for this. Also we had problems supporting XHTML.

thorn0:
> But this is not so. We don't even have tests for this.

Let's add tests then. And one more issue: I'm not sure formatting `*.svg` on running e.g. `prettier .` is a good idea. In many projects, SVGs are seen as images ("binary"), not markup ("text"). It might be unexpected if Prettier starts to format them. E.g. even here, in the Prettier project, SVGs are treated like that: they're minimized by `svgo`, and we don't want to pretty-print them.

ExE-Boss:
@thorn0 You can always use `.prettierignore` for that.

evilebottnawi:
@ExE-Boss Agree

@thorn0 Compressed files and source files should be stored in different directories, what if we need to restore the image, or change it?

evilebottnawi:
@thorn0 Also using HTML parser on SVG is not good idea, we will have the same problem like now we have with css/scss/Less, non standard css syntax is parsed, but we should throw an error, because scss/less support it, it is really weird

evilebottnawi:
Anyway, I think we should start with tests, so we can catch edge cases, and find out which parser suits us best

thorn0:
@evilebottnawi 

> HTML parser on SVG (XML) is not good idea, we will have the same problem like now we have with css/scss/Less

SVG is HTML. If it's not formatted correctly with `--parser html`, it's a bug in the HTML formatter, and it's need to be fixed.

evilebottnawi:
> SVG is HTML

I disagree here, SVG is XML, XHTML is HTML with XML rules, we have many logic for HTML tags, but this is not always applicable for SVG, but in same cases yes, I still think that we should have another printer for SVG/XML, mixing logic with HTML printer will reduce the maintainability of the code, not sure about the parser (maybe we can use same for HTML/SVG), but we don't have tests on this

Another question - should we support XML out of box?

thorn0:
> You can always use .prettierignore for that.

> Compressed files and source files should be stored in different directories, what if we need to restore the image, or change it?

On the one hand, makes sense. On the other, SVG is a graphics format and is usually (more often than not) edited in graphics editors, isn't it? Editing its code manually seems to be a less frequent use case. So probably to follow the least surprise principle, it's this use case that should need an additional setting (overrides + `--parser html`).

thorn0:
@evilebottnawi 

> I disagree here,

It's not something to be disagree with. It's a fact. https://html.spec.whatwg.org/#svg-0

evilebottnawi:
> On the one hand, makes sense. On the other, SVG is a graphics format and is usually (more often than not) edited in graphics editors, isn't it? Editing its code manually seems to be a less frequent use case. So probably to follow the least surprise principle, it's this use case that should need an additional setting (overrides + --parser html).

Honestly, I can see the benefit from this, this will allow you to see in git diff what changed, instead of one big change one line. But I propose to postpone this discussion, because it is a little different from that here. 

Formatting SVG in HTML is very good idea. So I would like to focus on support SVG here and postpone dicussion about formatting separates SVG files.

> Another question - should we support XML out of box?

What about this?

thorn0:
> Another question - should we support XML out of box?

It's a good question. My answer is "probably, eventually, yes, but I'm really not sure". Compared to a part of the core, a separate plugin is much more flexible in terms of maintenance, experimenting, and reacting to the demand.

brodybits:
> So probably to follow the least surprise principle, it's this use case that should need an additional setting (overrides + `--parser html`).

+1 for least surprise

> It's a fact. https://html.spec.whatwg.org/#svg-0

Then what should we do about MathML, from the previous section: https://html.spec.whatwg.org/#mathml etc.?

> should we support XML out of box?

+1 since XML is a core language in a similar fashion to JSON, YAML, etc.

evilebottnawi:
I found supporting XML out of box is good idea, it is still popular format

> separate plugin is much more flexible in terms of maintenance, experimenting, and reacting to the demand.

Yes, but for some formats we can do exception, I think this is our exception.

So, summarize:
- Let's start to add tests for SVG
- Improve SVG in HTML (don't think we need options here), 
- Implement formatting XML
- Decide formatting standalone SVG files

Any thoughts?

remcohaszing:
> It's a fact. https://html.spec.whatwg.org/#svg-0

This states the HTML element `<svg>`  can be used to embed SVG into HTML documents and `<foreightObject />` can be used to nest HTML in SVG. It doesn’t state SVG is HTML.

SVG is XML. This is relevant, because XML, and therefor SVG allow any elements to be self closing, whereas HTML has very strict rules about which elements are self closing and which aren’t.

For example, the following is valid XML/SVG, but not valid HTML:

```svg
<circle />
```

The following is valid HTML document, but is not valid XML:

```html
<!doctype html>
<html>
<body>
<p>
<br>
<p>
```

thorn0:
@remcohaszing 

HTML files can contain `svg` elements with SVG inside. Also tags like `html` and `body` are optional in HTML files, which means a "standalone" SVG file (where `svg` is a root element) would be HTML. Technically, such a file can also contain an XML prolog and a non-HTML doctype, but Prettier doesn't seem to have any trouble with those.

> For example, the following is valid XML/SVG, but not valid HTML:
> `<circle />`

XML - yes. SVG - not sure. SVG should have a root `svg` element. If we add it, `<svg><circle/></svg>` will be both XML and HTML.

sayjeyhi:
I need this really, thanks 👍

zulaica:
@thorn0 FWIW:

> SVG is a language for describing two-dimensional graphics in XML [XML10]. 
> … 
> SVG 1.1 is a modularization of SVG 1.0 [SVG10]. See the Document Type Definition appendix for details on how the DTD is structured to allow profiling and composition with other XML languages.

https://www.w3.org/TR/SVG11/intro.html

> The svg element is a container that defines a new coordinate system and viewport. It is used as the outermost element of SVG documents, but it can also be used to embed an SVG fragment inside an SVG or HTML document.

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg

That `svg` can exist in an HTML document does not mean every SVG file is valid HTML. The `svg` tag is simply a container for valid SVG markup.

lcnittl:
If xml gets included in core (+1 for that), then think that inclusion of the new file types in `.pre-commit-hooks.yaml` is still missing in the PR.

https://github.com/prettier/prettier/blob/e3f075b3e75767e9c8099efbe870c85c16189e55/.pre-commit-hooks.yaml#L5-L16

thorn0:
@zulaica Do you have any specific examples of how Prettier with `--parser html` incorrectly formats SVG?

ExE-Boss:
@thorn0 I have encountered <https://github.com/prettier/prettier/issues/9040>.

tjx666:
any progress? Svg support is very useful for svg developer~

thorn0:
@tjx666 Did you read the thread? Prettier already supports SVG. Use `--parser html`.

kylemh:
To be clear, that means the answer is to do:
`prettier "**/*.svg" --parser html` to format all SVGs?

tjx666:
@kylemh Yes, it works for me.

wyuenho:
Just dropping into this thread because I'm looking for an SVG formatter to turn the make parameters readable in an svgo optimized file. I must say, I'm a little surprised by the imprecise language used here to describe the relationships between HTML and SVG.

From an author's perspective, HTML and SVG are entire distinct documents (yes, SVG is a document). From Prettier's perspective, it only needs to concern itself parsing, and all we need to find out is when to switch parsing modes.

An HTML document, it may **inline** Javascript, CSS, SVG, and MathML. It's a containment relationship. HTML != SVG. HTML can have one of two parsing modes - HTML and XHTML. XHTML should be parsed according to XML rules. When an HTML parser sees an XML doctype or namespace, it should switch to an XML parser and parse the document using XML rules. When the HTML parser sees a `<script>` tag, it should switch to the JavaScript parser, and continue until `</script>` is seen. When the HTML parser sees a `<style>` tag, it should switch to the CSS parser, and continue until `</style>` is seen. When the HTML parser sees a `<svg>` tag, it should switch to an XML parser. While parsing the SVG, you may encounter `<foreignObject>` tags. Any descendants of `<foreignObject>` is parsed using (X)HTML rules, unless noted otherwise by their namespace. In addition, elements within a `<foreignObject>` must be **flow content** elements, so the XHTML fragment within it should not contain any `<script>` or `<style>` tags.

In order to have a complete HTML formatter, you **must** support formatting XML, in which XHTML, SVG and MathML are members of.

With that said, that's what an error-intolerant parser should support. Assuming you already have an tolerant HTML somewhat similar to real HTML parsers implemented in browsers, supporting XML should be extremely easy as the parsing rules are far more straight.

As an MVP, I'll be satisfied with anything that will break up single line XML into multiple lines, indenting descendants, and nicely format the SVG paths long `d` attributes.

MichaelDeBoey:
@thorn0 any specific reason why this one's closed? 🤔

lcnittl:
> @zulaica Do you have any specific examples of how Prettier with `--parser html` incorrectly formats SVG?

@thorn0 I do.

When running

```sh
npx prettier --write --parser html demo.svg
```

this:

image

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:cc="http://creativecommons.org/ns#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:xlink="http://www.w3.org/1999/xlink"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   inkscape:version="1.0 (4035a4fb49, 2020-05-01)"
   sodipodi:docname="demo.svg"
   viewBox="0 0 139.84252 98.267719"
   height="26mm"
   width="37mm"
   xml:space="preserve"
   id="svg1308"
   version="1.1"><metadata
     id="metadata1314"><rdf:RDF><cc:Work
         rdf:about=""><dc:format>image/svg+xml</dc:format><dc:type
           rdf:resource="http://purl.org/dc/dcmitype/StillImage" /><dc:title></dc:title></cc:Work></rdf:RDF></metadata><defs
     id="defs1312"><rect
       x="71.196815"
       y="86.192612"
       width="88.697708"
       height="41.896076"
       id="rect918" /><color-profile
       id="color-profile4"
       xlink:href="C:\WINDOWS\system32\spool\drivers\color\evolis_profil_v701.icc"
       name="EVOLIS-" /></defs><sodipodi:namedview
     inkscape:guide-bbox="true"
     showguides="true"
     units="mm"
     inkscape:document-units="mm"
     inkscape:current-layer="g1316"
     inkscape:window-maximized="1"
     inkscape:window-y="-8"
     inkscape:window-x="-8"
     inkscape:cy="112.84947"
     inkscape:cx="181.14523"
     inkscape:zoom="2.8284271"
     showgrid="false"
     id="namedview1310"
     inkscape:window-height="1377"
     inkscape:window-width="2560"
     inkscape:pageshadow="2"
     inkscape:pageopacity="0"
     guidetolerance="10"
     gridtolerance="10"
     objecttolerance="10"
     borderopacity="1"
     bordercolor="#666666"
     pagecolor="#ffffff"
     inkscape:document-rotation="0"
     inkscape:lockguides="false"
     inkscape:snap-bbox="true"
     inkscape:bbox-paths="true"
     inkscape:bbox-nodes="true"
     inkscape:snap-bbox-edge-midpoints="true"
     inkscape:snap-bbox-midpoints="true"
     inkscape:object-paths="true"
     inkscape:snap-intersection-paths="true"
     inkscape:snap-midpoints="true"
     inkscape:snap-smooth-nodes="true"
     inkscape:snap-object-midpoints="true"
     inkscape:snap-center="true"
     inkscape:snap-text-baseline="true"
     inkscape:snap-page="true" /><g
     transform="matrix(1.3333333,0,0,-1.3333333,0,204.01893)"
     inkscape:label="ausweis"
     inkscape:groupmode="layer"
     id="g1316"><text
       y="-112.6013"
       x="105.33331"
       id="text1326"
       style="font-variant:normal;font-weight:bold;font-size:15px;font-family:'PT Sans';-inkscape-font-specification:PTSans-Bold;text-align:end;writing-mode:lr-tb;text-anchor:end;fill:#000000;fill-opacity:1;fill-rule:nonzero;stroke:none"
       transform="scale(1,-1)"><tspan
         y="-112.6013"
         id="tspan1324"
         sodipodi:role="line"
         x="105.33331">Right aligned</tspan></text></g></svg>
```

gets "formatted" – not only code-wise – to this:

image

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
  xmlns:dc="http://purl.org/dc/elements/1.1/"
  xmlns:cc="http://creativecommons.org/ns#"
  xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
  xmlns:svg="http://www.w3.org/2000/svg"
  xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink"
  xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
  xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
  inkscape:version="1.0 (4035a4fb49, 2020-05-01)"
  sodipodi:docname="demo.svg"
  viewBox="0 0 139.84252 98.267719"
  height="26mm"
  width="37mm"
  xml:space="preserve"
  id="svg1308"
  version="1.1"
>
  <metadata id="metadata1314">
    <rdf:RDF
      ><cc:Work rdf:about=""
        ><dc:format>image/svg+xml</dc:format
        ><dc:type
          rdf:resource="http://purl.org/dc/dcmitype/StillImage" /><dc:title></dc:title></cc:Work
    ></rdf:RDF>
  </metadata>
  <defs id="defs1312">
    <rect
      x="71.196815"
      y="86.192612"
      width="88.697708"
      height="41.896076"
      id="rect918"
    />
    <color-profile
      id="color-profile4"
      xlink:href="C:\WINDOWS\system32\spool\drivers\color\evolis_profil_v701.icc"
      name="EVOLIS-"
    />
  </defs>
  <sodipodi:namedview
    inkscape:guide-bbox="true"
    showguides="true"
    units="mm"
    inkscape:document-units="mm"
    inkscape:current-layer="g1316"
    inkscape:window-maximized="1"
    inkscape:window-y="-8"
    inkscape:window-x="-8"
    inkscape:cy="112.84947"
    inkscape:cx="181.14523"
    inkscape:zoom="2.8284271"
    showgrid="false"
    id="namedview1310"
    inkscape:window-height="1377"
    inkscape:window-width="2560"
    inkscape:pageshadow="2"
    inkscape:pageopacity="0"
    guidetolerance="10"
    gridtolerance="10"
    objecttolerance="10"
    borderopacity="1"
    bordercolor="#666666"
    pagecolor="#ffffff"
    inkscape:document-rotation="0"
    inkscape:lockguides="false"
    inkscape:snap-bbox="true"
    inkscape:bbox-paths="true"
    inkscape:bbox-nodes="true"
    inkscape:snap-bbox-edge-midpoints="true"
    inkscape:snap-bbox-midpoints="true"
    inkscape:object-paths="true"
    inkscape:snap-intersection-paths="true"
    inkscape:snap-midpoints="true"
    inkscape:snap-smooth-nodes="true"
    inkscape:snap-object-midpoints="true"
    inkscape:snap-center="true"
    inkscape:snap-text-baseline="true"
    inkscape:snap-page="true"
  />
  <g
    transform="matrix(1.3333333,0,0,-1.3333333,0,204.01893)"
    inkscape:label="ausweis"
    inkscape:groupmode="layer"
    id="g1316"
  >
    <text
      y="-112.6013"
      x="105.33331"
      id="text1326"
      style="
        font-variant: normal;
        font-weight: bold;
        font-size: 15px;
        font-family: 'PT Sans';
        -inkscape-font-specification: PTSans-Bold;
        text-align: end;
        writing-mode: lr-tb;
        text-anchor: end;
        fill: #000000;
        fill-opacity: 1;
        fill-rule: nonzero;
        stroke: none;
      "
      transform="scale(1,-1)"
    >
      <tspan y="-112.6013" id="tspan1324" sodipodi:role="line" x="105.33331">
        Right aligned
      </tspan>
    </text>
  </g>
</svg>
```

I can open up a new issue if you want.

thorn0:
@MichaelDeBoey When we renamed the default branch, this PR got closed because its fork doesn't exist anymore.

@lcnittl Yes, please open an issue. If you could isolate the problem and make the code example shorter, that'd be great.

PR #8909, opened 2020-08-05
Add `fix` & `lint` scripts

fisker:


fisker:
I like `npm-run-all`, and I think it's popular enough.

kachkaev:
How about `fix` instead of `format`? ESLint does  more than just formatting. Besides, there may be other tools you’ll want you use in future in this group of scripts, such as `yarn-deduplicate`.

Example: https://github.com/kachkaev/njt/blob/master/package.json

fisker:
`fix` is good, I remember I used use `fix` but changed to `format`, and I forgot reason.

PR #8923, opened 2020-08-06
Update ESLint settings

fisker:

### DIFF
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -1,3 +1,5 @@
+/* globals prettier prettierPlugins parsersLocation */
+
 "use strict";
 
 self.PRETTIER_DEBUG = true;
@@ -11,9 +13,7 @@ function importScriptOnce(url) {
 }
 
 importScripts("lib/parsers-location.js");
-const { parsersLocation } = self;
 importScripts("lib/standalone.js");
-const { prettier } = self;
 
 // this is required to only load parsers when we need them
 const parsers = Object.create(null);
@@ -24,7 +24,7 @@ for (const file in parsersLocation) {
     Object.defineProperty(parsers, parserName, {
       get() {
         importScriptOnce(url);
-        return self.prettierPlugins[property].parsers[parserName];
+        return prettierPlugins[property].parsers[parserName];
       },
     });
   }
--- a/website/static/worker.js
+++ b/website/static/worker.js
@@ -1,8 +1,8 @@
-/* eslint-env worker */
-/* eslint no-var: off, strict: off */
-/* globals prettier prettierPlugins parsersLocation */
+"use strict";
 
-var imported = Object.create(null);
+self.PRETTIER_DEBUG = true;
+
+const imported = Object.create(null);
 function importScriptOnce(url) {
   if (!imported[url]) {
     imported[url] = true;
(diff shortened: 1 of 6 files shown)


website/static/worker.js
```diff
@@ -1,8 +1,10 @@
-/* eslint-env worker */
-/* eslint no-var: off, strict: off */
 /* globals prettier prettierPlugins parsersLocation */
 
-var imported = Object.create(null);
+"use strict";
+
+self.PRETTIER_DEBUG = true;
```
fisker, line 5: This is functional https://deploy-preview-8923--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAucBnVADAOlAhkgRgG4CcMQAaECABxgEtpVlRcAnNiAdwAV2FmKXABsuuAJ7MqAIza4wAazgwAyjXn0oAc2Qw2AVzhU4AW2lwAJhcsAZXNv24tcAGIQ2J3DAbbkIXPowEJQgABYwJsIA6qH08KjqYHAqAnH0AG5x4n5g6CGaqHBsMDxyWp7IAGYihVQAVqgAHgBCcorKKrgmcDaacFU1RiANjSqaWsJwAIr6EPADwrUg6myFbH7SuObCITRsmjBR9BYwocgAHAAMVHsQhVFyNH57cGvp-VQAjrPwpbSC-lQAFooHBLJYQmw4N96FDSk4KkhqoshoUTPRdAZUeNJjM5v0kYMqDAtkcTmckAAmYlyejCcYAYQgJkRIFeAFYQvpCgAVLaCZFLdKGACSUGssBUYH2dAAguKVDBxJMFoUAL5qoA

website/static/worker.js
```diff
@@ -1,8 +1,10 @@
-/* eslint-env worker */
-/* eslint no-var: off, strict: off */
 /* globals prettier prettierPlugins parsersLocation */
 
-var imported = Object.create(null);
+"use strict";
```
fisker, line 3: We have `parser-flow.js` which is not `strict`, and this test fails when I try to set it `strict` https://github.com/prettier/prettier/runs/676631327#step:7:65, but [this code](https://deploy-preview-8923--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBnADgQzHVACAXjwB4ATASwDcA+AXTwYG1iB6CmkAGhAnRnOipkoTACdREAO4AFMQiEpMAG0mYAnkO4AjUdgDWcGAGUsYclADmyGKICucbnAC2WuKVJuAMpku3MFuAAxCFEnTBh+S2QQTFsYCC4QAAsYJyUAdSTyeAxsOCN5bKpstWiwVE0Qc1Q4URhpXQsw5AAzZRruACtUAA8AIV0wA2NMJzhPczhW9ocQbp6jcwslOABFWwh4aaUOkCxRGtFolqUpRPRRcxh08lIYJOQADgAGbguIGvTddGiL3FrKFNuABHDbwBq8BQxVAAWigcDcbkSojgoPIKIa-maSDaO1mNSc5GsdnxSxW602UxxM24MEwWhudweSAATLTdOQlEsAMIQJzYkC4ACsiVsNQAKvSFLjdpR7ABJKAeWBGMCXPgAQSVRhgahW2xqAF9DUA) seems working on background.


So it should not breaking things.

fisker: > If you put "use strict" at the top of a web worker javascript file, it will not cause strict mode to be applied to any other files, including those imported via importScripts

Quote from https://stackoverflow.com/questions/26855884/is-the-function-form-of-use-strict-necessary-in-a-web-worker

PR #8931, opened 2020-08-07
Rewrite a deep nested ternary `makeAlign`

fisker:

### DIFF
--- a/src/document/doc-printer.js
+++ b/src/document/doc-printer.js
@@ -19,17 +19,27 @@ function makeIndent(ind, options) {
 }
 
 function makeAlign(ind, n, options) {
-  return n === -Infinity
-    ? ind.root || rootIndent()
-    : n < 0
-    ? generateInd(ind, { type: "dedent" }, options)
-    : !n
-    ? ind
-    : n.type === "root"
-    ? { ...ind, root: ind }
-    : typeof n === "string"
-    ? generateInd(ind, { type: "stringAlign", n }, options)
-    : generateInd(ind, { type: "numberAlign", n }, options);
+  if (n === -Infinity) {
+    return ind.root || rootIndent();
+  }
+
+  if (n < 0) {
+    return generateInd(ind, { type: "dedent" }, options);
+  }
+
+  if (!n) {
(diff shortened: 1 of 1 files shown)


src/document/doc-printer.js
```diff
@@ -19,17 +19,27 @@ function makeIndent(ind, options) {
...
+
+  if (n.type === "root") {
+    return { ...ind, root: ind };
+  }
+
+  return generateInd(
+    ind,
+    { type: typeof n === "string" ? "stringAlign" : "numberAlign", n },
+    options
+  );
```
brodybits, line 42: I would probably favor compacting it something like this:

```js
function makeAlign(ind, n, options) {
  if (n === -Infinity) {
    return ind.root || rootIndent();
  } else if (n < 0) {
    return generateInd(ind, { type: "dedent" }, options);
  } else if (!n) {
    return ind;
  } else if (n.type === "root") {
    return { ...ind, root: ind };
  }
  // else
  return generateInd(
    ind,
    { type: typeof n === "string" ? "stringAlign" : "numberAlign", n },
    options
  );
}
```

fisker: Why use useless else?

fisker: The `else if` is the same as the last `else` you omitted.

fisker: Actually with this option https://eslint.org/docs/rules/no-else-return#options-74, your version will be reported.

brodybits: Understood. I was hoping to get this block more compact, somehow. How about this:

```js
function makeAlign(ind, n, options) {
  if (n === -Infinity) return ind.root || rootIndent();

  if (n < 0) return generateInd(ind, { type: "dedent" }, options);

  if (!n) return ind;

  if (n.type === "root") return { ...ind, root: ind };

  return generateInd(
    ind,
    { type: typeof n === "string" ? "stringAlign" : "numberAlign", n },
    options
  );
}
```

fisker: I never use this style.

PR #8940, opened 2020-08-09
Refactor test

fisker:
I moved a lot of stuff into `format()` function. The rest part just variable renaming.

Reason behind this, I'm thinking move `TEST_CRLF` to a seperate part of `run_spec` (or maybe add `\r` test too).
It should not only run on windows, this is misleading.

So `run_spec` will be like

```js
test('format');

if (TEST_CRLF) {
 test('crlf');
}
```


tests_config/run_spec.js
```diff
@@ -113,89 +122,71 @@ global.run_spec = (fixtures, parsers, options) => {
...
+        expect(firstFormatResult.eolVisualizedOutput).toEqual(
+          visualizeEndOfLine(
+            consistentEndOfLine(firstFormatResult.outputWithCursor)
+          )
         );
+
         if (typeof output === "string") {
-          expect(formatted).toEqual(output);
+          expect(firstFormatResult.output).toEqual(output);
         } else {
```
fisker, line 164: I'll change this to `early return` before merge, in order to reduce indention of rest part.

tests_config/run_spec.js
```diff
@@ -229,66 +220,66 @@ global.run_spec = (fixtures, parsers, options) => {
...
-      const isUnstable = unstableTests.get(filename);
-      const isUnstableTest = isUnstable && isUnstable(options || {});
+      const isUnstable = isUnstableTest(filename, formatOptions);
       if (
         DEEP_COMPARE &&
-        (formatted !== input || isUnstableTest) &&
-        typeof rangeStart === "undefined" &&
-        typeof rangeEnd === "undefined" &&
-        typeof cursorOffset === "undefined" &&
-        !TEST_CRLF
```
fisker, line 273: This enables `second format test` on windows. Don't know I why I skip this before.

tests_config/run_spec.js
```diff
@@ -300,15 +291,46 @@ function parse(source, options) {
...
+    })
+    .replace(RANGE_END_PLACEHOLDER, (match, offset) => {
+      options.rangeEnd = offset;
+      return "";
+    });
+
+  const input = inputWithCursor.replace(CURSOR_PLACEHOLDER, (match, offset) => {
+    options.cursorOffset = offset;
+    return "";
   });
```
fisker, line 312: Actually this part aoild not right, if cursor is before `rangeEnd`, it is wrong.

PR #8942, opened 2020-08-09
test private-in on all Babel parsers

brodybits:
in followup to PR #8431

### DIFF
--- a/tests/js/private-in/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/js/private-in/__snapshots__/jsfmt.spec.js.snap
@@ -2,7 +2,7 @@
 
 exports[`private-in.js format 1`] = `
 ====================================options=====================================
-parsers: ["babel"]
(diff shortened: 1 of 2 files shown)


fisker:
I think this won't be nessary, `babel-flow` and `babel-ts` are just babel, `babel-ts` add extra `typescript` plugin.  `babel-flow` even the same as `babel` because `babel` enable `flow` too, unless there are conflicts, they all the same in this test. You can check `parser-babel.js`

fisker:
FYI: other new  syntax I added also not tested.

fisker:
If you really want do something like this, I hava a  suggest

Do it in `tests_config/run_spec.js`.

If `babel` include, add `babel-flow` `babel-ts`.
If `flow` include, add `babel-flow`(I was planning to do this).
Same for `typescript`

I don't what the results, but some tests should fail.

PR #9003, opened 2020-08-20
Improve `requirePragma` on files with BOM

fisker:
- Improve how `format` handle `BOM`, with `requirePragma` option (some edge case might fail)
- Return correct `cursorOffset` for files don't hasPragma


evilebottnawi (commented):
Can we add tests here? Changing something without tests can be potential dangerous

fisker:
The breaking case before this is, use BOM, SHEBANG, Docblock together, very edge case, but I can add one.

evilebottnawi:
@fisker Yes, let's add it to avoid future regressions

evilebottnawi (approved):

PR #9013, opened 2020-08-20
Improve internal `cursorOffset` calculate

fisker:
Similar to #9012, this one is more tricky.

### DIFF
--- a/src/main/core.js
+++ b/src/main/core.js
@@ -199,8 +199,8 @@ function formatRange(text, opts) {
 
   let formatted =
     text.slice(0, rangeStart) + rangeTrimmed + text.slice(rangeEnd);
-  const eol = convertEndOfLineToChars(opts.endOfLine);
-  if (eol !== "\n") {
+  if (opts.endOfLine !== "lf") {
+    const eol = convertEndOfLineToChars(opts.endOfLine);
     if (cursorOffset >= 0 && eol === "\r\n") {
       const before = formatted.slice(0, cursorOffset);
       const match = before.match(/\n/g);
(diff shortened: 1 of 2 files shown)


src/main/core.js
```diff
@@ -171,68 +171,45 @@ function formatRange(text, opts) {
       ...opts,
       rangeStart: 0,
       rangeEnd: Infinity,
-      // track the cursor offset only if it's within our range
+      // Track the cursor offset only if it's within our range
       cursorOffset:
-        opts.cursorOffset >= rangeStart && opts.cursorOffset < rangeEnd
+        opts.cursorOffset > rangeStart && opts.cursorOffset < rangeEnd
```
fisker, line 176: Should not include `=`

src/main/core.js
```diff
@@ -171,68 +171,45 @@ function formatRange(text, opts) {
...
-  const rangeLeft = text.slice(0, rangeStart);
-  const rangeRight = text.slice(rangeEnd);
 
   let { cursorOffset } = opts;
-  if (opts.cursorOffset >= rangeEnd) {
+  if (cursorOffset >= rangeEnd) {
     // handle the case where the cursor was past the end of the range
     cursorOffset =
-      opts.cursorOffset - rangeEnd + (rangeStart + rangeTrimmed.length);
+      opts.cursorOffset + (rangeTrimmed.length - rangeString.length);
```
fisker, line 193: Should be the same, but simpler

src/main/core.js
evilebottnawi: Can we add test on this?

fisker: Sure, this is not breaking anything, just no need include `=`.

I'll  add one tomorrow.

fisker: Done https://github.com/prettier/prettier/pull/9013/commits/055870514fddb9f834e3695b6787501cdfa2f9cf

evilebottnawi (approved):

PR #9025, opened 2020-08-22
Styled-components: Support createGlobalStyle``

wmertens:
Adds support for https://styled-components.com/docs/api#createglobalstyle


thorn0:
Yes, this PR (as well as #5993 and #8647) is blocked by #5588. We won't move forward with it until we solve that problem.

wmertens:
:ok_hand: that looks like a nice solution! I'll close this

Issue #7746, opened 2020-03-10
Unstable return statement format


fisker:
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEARjlAIYBOAngJJQBCJFlAMnKjABQCU+wAOlHz5ycGOnKCOAoUID0s-ACU4pAM4FUEcvgAsAJmkz9hngCp8ADh34A1PgNQAvgIEYsuAsTJVaDb5UUcAHMAC04efkFhUXFJQzkFZTUNLV0HGTSTW0trcwdnKFdMbDxBSCgAExwPMgAbbl5DETEJfCkohKUVdUFNbWMOzKieAH58AEZ8JHsBAqL3UqJGHygAYWgqmtJ6iKaY1vaM+S7k3tSBjIv8M3wAZnwxyen8gRAAGhAIAAca1WRQCjkCAAdwAChQEH8UKQAG4QHAVd4gQjkUhgADWogAyl80SQgsgYOR0HAPmEALa1ADqIWqcFUuLAcCxkOqOBh1UoyHAqj+HxIqjg5BgoNRQXJpGQqG2go+ACtVAAPOiojHY0jkuDMEhwKUy0kgBWKrH42pwACK6Ag8D1tVlIFx5EF5G5hFIhDgtSRX3IJBgVIRMBCyAsAAYPj6IIKqaivtyffShTDdR8AI5W+Ci75QkBqAC0UDgcAqxaRInTOBEotI4slSGldoNgvJOEJxKbpotGd19f1Hxg7oDFSDyD0-dROFq+PW5Il3Kg0BTIHQgoAKu6oQ37TCSbQS7AsWBfT8AIKVLEwShm22CxyOIA)
```sh
--parser babel
```

**Input:**
```jsx
function binaryInBinaryLeft() {
  return (
    // Reason for 42
    42
  ) * 84 + 2
}

function binaryInBinaryRight() {
  return (
    // Reason for 42
    42
  ) + 84 * 2
}

function conditional() {
  return (
    // Reason for 42
    42
  ) ? 1 : 2
}

function binaryInConditional() {
  return (
    // Reason for 42
    42
  ) * 3 ? 1 : 2
}

```

**Output:**
```jsx
function binaryInBinaryLeft() {
  return (
    // Reason for 42
    42 *
      84 +
    2
  );
}

function binaryInBinaryRight() {
  return (
    // Reason for 42
[...]

PR #9026, opened 2020-08-22
add parens around mixed division and addition or subtraction operators for readability

brad-decker:
partially addresses #187 by applying the same logic of the '%' operator
to the '/' operator. One exception is where there are leading comments:

e.g

```js
return (
  // Reason for 42
  42
) / 8 + 2;
```

will not be formatted to avoid mangling the output.


tests/js/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3614,9 +3614,9 @@ function binary() {
 
 function binaryInBinaryLeft() {
   return (
-    // Reason for 42
+    (// Reason for 42
```
evilebottnawi, line 3617: Ugly

brad-decker: Ooh yeah good catch, @evilebottnawi -- seems like the comment is being considered part of the parent operation. I'll see if I can fix this. If you happen to have any advice on where to look i'd appreciate it!

brad-decker: @evilebottnawi this is my first time contributing here and I'm trying to dig into this a bit. It doesn't seem like from the context of needs-parens that you can really specify where the parens should be added, or add a linebreak after a comment. However, running prettier a second time against the input yields the much cleaner:

```js
function binaryInBinaryLeft() {
  return (
    // Reason for 42
    (42 * 84) +
    2
  );
}
```

So I guess there are two options one of which I'm pretty unsure of where to look for examples of how to achieve the effect:

#### Option 1: Skip Nodes with comments
Essentially we'd just look if either the left or right side of the node has comments and return false in needs-parens. This would allow us to care for the vast majority of `mixed-operators` issues. Remaining issues that end users are having can be resolved by moving the comments manually.

#### Option 2: Change the starting line/col of parenthesis

The problem with the output stems from the comment being a part of the expression. I tried poking around in printer-estree with `parts` which is having parenthesis injected into it if needsParens is `true`. However, I'm not sure what the desired output should be or how to achieve it.

```js
function binaryInBinaryLeft() {
  return (
    // Reason for 42
    (42 * 84) + 2
  );
}
```

seems like the right approach, but I'm getting a little lost on the proper way to achieve this result. If anyone reading this has pointers I'm all ears and very eager to have this PR be considered for inclusion in prettier.

brad-decker: I went with the first option for now.

fisker:
I like the idea, but I don't think it's improving readability in  many  cases.

For example

`((2 / 3) * 10) / 2 + 2;` (from our  tests)

VS

`(((2 / 3) * 10) / 2) + 2;`

I would  prefer.
```js
(
  ((2 / 3) * 10) / 2
) + 2;
```

I'm not sure what to do

brad-decker:
Thanks for your response @fisker :D. I obviously have a preference for `(((2 / 3) * 10) / 2) + 2` for a lot of the reasons mentioned in #187 over the original output. I'm indifferent about breaking it into multiple lines. For me adding the parenthesis to separate the division from the addition at the end of the equation is particularly helpful, and I think that how it is formatted in terms of breaking into multiple lines adds less overall value than having the parenthesis.

I would also say that there are those on either side of the `no-mixed-operators` line in prettier, and I think any additional changes to formatting the output to exist across multiple lines, etc, will further complicate that discussion.

It would be my personal preference to first add the parenthesis and tackle any additional formatting in a separate pull request, perhaps after getting some community feedback on the approach?

brad-decker:
@lydell -- you mentioned in #187 that a PR to add parenthesis to mixed operators in this manner would have a high likelihood of being merged.

https://github.com/prettier/prettier/issues/187#issuecomment-628236114

I would greatly appreciate any feedback you can provide that would help this be easier to get approval on and merge. @fisker has some feedback and I tried to respond to feedback from @evilebottnawi but I'm not sure if my approach with skipping expressions with comments interspersed is preferred. Thank you!

lydell:
@brad-decker In my comment you linked to:

> Not sure what to say about multiplication.

It’s _division,_ not multiplication, that I think has a high chance of getting merged.

brad-decker:
Oops, totally jumped to conclusions on the multiplication part. So, just to confirm, @lydell:

```js
(100 / 10) + 2
```

is easier to merge/approve than

```js
(100 * 10) + 2
```

Is the multiplication stuff just more contentious in terms of style/necessity or is there more complexities in it? As long as my comment above is correct I'll go ahead and remove multiplication from this PR and only focus on division and if there is any way to get multiplication considered I'll address it in a future PR. Thanks for your time and attention.

thorn0:
> Is the multiplication stuff just more contentious in terms of style/necessity

Yes, probably. It might even be a cultural thing. For ex-USSR-ians, parens in expressions like `a + (b * c)` might look really bad, even confusing. They raise questions: What do these redundant things do here? Do they carry some additional information that I'm unable to see? For Americans, on the other hand, such parens seem to be okay, even preferable. It probably comes down to how much people are exposed to math in their childhood.

brad-decker:
Well, today I learned. My apologies for my uncultured assumption -- and I imagine that this ties into @fisker 's feedback regarding the breaking of parenthesis to be more legible. Sorry! I've updated this PR to only apply to division.

thorn0:
@brad-decker Sorry if the "cultural" is the wrong word. I used it only in a geographic/ethnographic sense. In the sense that people see these things differently depending on their origin. I totally didn't mean that somebody is more cultured than somebody else. In our schools, such parens are considered a mistake. Seeing people here want to have them in JS was a TIL moment for me too.

brad-decker:
I assure you I am less cultured than others, took no offense to your message, and hopefully did not offend others with mine by assuming my way of looking at math is the correct one. Thanks for calling it out, @thorn0!

brad-decker:
@fisker the switch to focusing on division doesn't actually change anything for your comment/suggestion. The output is the same. Does the reduced scope of the change help to alleviate some of the concerns or should I explore how to break this expression over multiple lines?

thorn0:
Breaking over multiple lines sounds like a separate issue. Similar to what is discussed with regard to function calls in #6921.

fisker:
**Prettier pr-9026**
[Playground link](https://deploy-preview-9026--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBDAvAAgCwCZMD0AVJpALZkIyZEGGYAc2mA1JgIwA6UGO+dTQiXKVYNOm3YgANCAgAHGAEtoAZ2ShUAJy0QA7gAVtCdSlQAbPagCe62QCMtqMAGs4MAMrznSqAHNkGC0AVzhZODJ7OAATaJiAGVR-YNQ-OAAxCC0yVBhlf2Q0YJgIGRAACxgycwB1cqV4VW8wOA8TBqUANwbrQrBVOxBfVTgtGAMnPxzkADMLEdkAK1UADwAhJ1d3D1RKeN84WfmwkGWVj18-czgARWCIeCPzBZBvLRGtQvtUKPMy+S0vhgNSU0Rg5WQDAADLIARARjUnPJCgC4B9OodZABHe7wCYKUxoVQAWigcBiMTKWjgOKU1ImqWmSDmzxOIzISkCITZl2udweh2Zx1kMB+ILBEKQuBFTiU5kuAGEIBRUIU0QBWMrBEYAFR+phZL06oQAklA4rAPGBAYoAILmjwwazXJ4jAC+bqAA)
```sh
--parser babel
```

**Input:**
```jsx
a= 42 /* comment */ / 84 + 1
a= 42 / 84 /* comment */ + 1
```

**Output:**
```jsx
a = 42 /* comment */ / 84 + 1;
a = (42 / 84) /* comment */ + 1;

```

I think we should add parens for first one too.

fisker:
And I don't think we should treat `%` and `/` differently, maybe we can improve the case @evilebottnawi pointed in seperate PR.

**Prettier pr-9026**
[Playground link](https://deploy-preview-9026--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAnOMCuqoAIAUAOnrgPSm4BKcAhgM7S4BmEquALAEzG4fdQBKMrgAcuANS5OAbmLF0WHAR5kK1eoxZsuKnYNwBSURKnSQAGhAQADjACW0OslA1UqCAHcACq4ROUNAA2HjQAnk6WAEaoNGAA1hgAytaxdlAA5sgwqJhwlnAAtpFwACYlpQAyNBmYNOlwAGKsBTQw9hnIIDSYMBAWIAAWMAWBAOoDdvB0KWBwiX6TdgBuk6GdYHQRIGl0cKgwXjHpLchMQbuWAFZ0AB4AQjHxSTQFcBVpcKfneSDXN4lpdKBOAARUwEHgX0CFxAKVQu1QnUiNGKgX61lQaRgozsJRgA2QIgADJYMRBdqMYtZOhi4Ailp9LABHcHwQ42fxdOgAWigcFKpX66BZdnQhzqJyQZ2hP12BTsWRyssBwLBEM+Uu+lhgKJxeIJSE42pidkCgIAwhACpKQHSAKz9TC7AAqKP80phS1yAEkoOVYIkwJjbABBP2JGChYFQ3YAX1jQA)
```sh
--parser babel
```

**Input:**
```jsx
return (
  // Reason for 42
  42
) / 8 + 2;

return (
  // Reason for 42
  42
) % 8 + 2;
```

**Output:**
```jsx
return (
  // Reason for 42
  42 /
    8 +
  2
);

return (
  (// Reason for 42
  42 %
    8) +
  2
);

```

tests/js/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3621,6 +3628,15 @@ function binaryInBinaryLeft() {
...
 }
 
+function binaryInBinaryLeftDivision() {
+  return (
+    // Reason for 42
+    42 /
+      84 +
+    2
+  )
+}
```
evilebottnawi, line 3638: It is really ugly, we should fix it here

fisker: In that case we can open another issue, fix this first

**Prettier 2.1.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEqEEAFAJT7AA6U++ATnDOvbSTXQPSf4BKcAQwDOBIvXwAWAEwdJMqBQCk+ABz4A1PikBuGgF8QAGhAQADrmhDkoAfXoQA7gAVbCKygEAbBwICeV4wAjegEwAGsmAGVTUJwoAHNkGHp0OGM4AFtAuAATHNyAGQEE9AF4uAAxCHoMgRhcBOQQAXQYCCMQAAsYDM8AdU6ceCEYsDhItyGcADch3yawIQCQOKE4ehgnEPja5FQvNeMAKyEADwAhEPCogQy4Ari4PYO0kBPTyLj4zzgARXQIPBnp5DiAYvQ1vQmoEBNlPB1TPQ4jA+jgcjBOsgVAAGYyIiBrPohUxNRFwSHTJ7GACOAPgWzM7maQgAtFA4Llch1GLScIwtmVdkh9iDXmsMjgkikxV8fv9AU9hS9jDBYaj0ZikFIVSEcJ4vgBhCAZIUgckAVg66DWABVYe4RaDpqkAJJQfKwSJgJHmACC7siMF8P2Baz0eiAA)
```sh
--parser babel
```

**Input:**
```jsx
function foo() {
  return (
  // Reason for 42
  42
) % 8 + 2;
}
```

**Output:**
```jsx
function foo() {
  return (
    (// Reason for 42
    42 %
      8) +
    2
  );
}

```

**Second Output:**
```jsx
function foo() {
  return (
    // Reason for 42
    (42 % 8) +
    2
  );
}

```

evilebottnawi: I don't think we should merge this commits with a lot of future issue, instead we should solve it here

fisker: I mean fix `%` one first, then  this.

brad-decker: What would the desired output look like? I tried to walk through the codebase to find the proper place to move/format lines/parens/comments and was kind of lost. I am committed to learning how to work with the prettier codebase to become an active contributor, so I'm going to give it another go. If you wouldn't mind giving me a target to shoot for in the output? @fisker @evilebottnawi

fisker: This one may won't be easy to fix. Format original one 3 times, you'll get this.


```
function foo() {
  return (
    // Reason for 42
    (42 % 8) + 2
  );
}

```

brad-decker: @fisker yeah I had the same result, I tried looking in the parser where needsParens is called and manipulating the insertion point of the parenthesis to be after the comment, but the comment is already printed at that point. I tried copying the pattern that was used for flow type annotations where the comment object is mutated to have `printed = true` and then the comment is manually inserted. I thought that would do the trick but like I said it seems in that logic tree, or with leading comments, it is already printed at the point it gets to inserting the parens?

brad-decker: in the example here, if you move the comment in the input to be above the return statement, the output is as you'd want... This is why I was hoping to manipulate the printing of the comment to be before the inserted opening parens.

thorn0: This is really a separate issue, namely https://github.com/prettier/prettier/issues/7746
I don't think it should be fixed here.

johannesjo:
Is there any help needed with this?

jsg2021:
Would this address [parentheses around sequence expressions](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAE8MDECGYMEATgJ6YC8mAFAJRUB8tDlzNkUaEANnAHQB3fCSg0A5ABEImKBCxh8PHuLoAaWQFdldANwg1ICAAcYAS3TJQIkhEEAFEQjTIQS4WReGARiUIBrOBgAZWNCMygAc2QYEk04QzgAW284ABM09IAZfCjNfEi4XFIk-BhzKNd8TWIDEAALGCSeAHV6sxwwsDhg5w6zADcOslcwNC8QCLQ4Ehh7P0jS5AAzJWnDACs0AA8AIT8wQJD8JLgsiLgVtYSQLe3giMi+AEVNeUukVZ51kDCSaZIrm8+FSPDqxhIERgLTMaRg9WQAA4AAyGCEQaYtPzGVwQuAAgaXQwARze8HmJhcKHwaAAtFA4Ol0nUSHBSWZWfMCktPtdDNMkmYYnEbmhHi8yR8vj8YCCYXCEUgAEyGWL4Mw8R4AYQgSR5IHxAFY6pppgAVEFU6U3AbxACSUEysGCYEhpgAgo7gjAyHwrt84ABfQNAA)?

PR #9053, opened 2020-08-25
Add `loc` back for `typescript` parser

fisker:
Jest `inline snapshot` rely on AST `loc` property, add it back to avoid breaking jest.

I also checked other parsers, we have `loc` in `babel`, because `comment`s in `babel` don't have range info. `flow` parser seems don't have an option for it.

Context https://github.com/facebook/jest/issues/10444

/cc @SimenB @evilebottnawi @thorn0


SimenB:
Can confirm this change fixes the reported regression 👍

evilebottnawi (approved):

thorn0 (approved):

PR #9057, opened 2020-08-25
Setup code style fix action

fisker:
https://github.com/prettier/prettier/pull/9051#issuecomment-680122099


.github/workflows/code-style.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Fix Code Style
+        run: yarn fix
+
+      - name: Commit Changes
+        uses: stefanzweifel/git-auto-commit-action@v2.3.0
+        with:
+          commit_message: Apply Prettier changes
+          branch: ${{ github.head_ref }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
evilebottnawi, line 31: Hope it will not create broken code due some bugs on our side or eslint :smile:

fisker: Only run on pull request, don't worry.

thorn0:
This needs some kind of protection against Prettier's idempotence bugs, otherwise it might end up in an infinite loop.

fisker:
It seems a dead end, can't commit to the fork.

src/language-css/printer-postcss.js
```diff
@@ -345,7 +345,8 @@ function genericPrint(path, options, print) {
     }
     case "media-feature": {
       return maybeToLowerCase(
-        adjustStrings(node.value.replace(/ +/g, " "), options)
+        adjustStrings(
+        	node.value.replace(/[ ]{1,}/g, " "), options)
```
evilebottnawi, line 349: Why it changed?

fisker:
Have you seen this https://github.com/stefanzweifel/git-auto-commit-action#using-the-action-in-forks-from-public-repositories ?

thorn0:
Yep, but as you can see we came very close: `remote: Permission to fisker/prettier.git denied to github-actions[bot].`

> You can only make commits on pull request branches that:
> - are opened in a repository that **you have push access** to and that were created from a fork of that repository
> - are on a user-owned fork
> - have permission granted from the pull request creator
> - don't have branch restrictions that will prevent you from committing

([source](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork))

At this point, we can probably give push access to `github-actions[bot]` or one of us can generate a personal access token and add it to the repo as a secret, but both of these options look extremely insecure to me.

fisker:
I agree.

thorn0:
In this PR, you added a new workflow file and it got immediately executed. Is it because you have push access? Or does it work the same way for anyone's PRs? In the latter case, anybody can steal all the secrets by simply opening a PR.

fisker:
It's not working, that commit caused by opening a new PR on my fork 😄 

https://github.com/fisker/prettier/pull/711

fisker:
> In this PR, you added a new workflow file and it got immediately executed. 

I don't know.

thorn0:
It didn't work because with `pull_request_target`, [the workflow files are taken from the base branch](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks) of the PR (i.e. `master` in our case). We don't have this workflow in `master` yet, that's why it doesn't work.

thorn0:
I guess we should switch our existing workflows from `pull_request` to `pull_request_target`.

thorn0:
Seems to work! Let me open one more PR against the `tmp-9057` branch to test it one more time.

fisker:
Cool!

lipis (approved):

lipis:
Interesting one..! Are we waiting anything for this one? Worst case scenario we can always remove it later.

fisker:
https://github.com/prettier/prettier/pull/9083

fisker:
Close in favor of #15749

PR #9069, opened 2020-08-26
Add support for @use and @forward rules in Scss

AndrewLeedham:
I forked from https://github.com/prettier/prettier/pull/7057 in order to address the PR review, since that PR has been open for about a year.

### DIFF
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -542,6 +542,8 @@ function parseNestedCSS(node, options) {
           "return",
           "define-mixin",
           "add-mixin",
+          "use",
+          "forward"
         ].includes(name)
       ) {
         // Remove unnecessary spaces in SCSS variable arguments
--- a/src/language-css/parser-postcss.js
+++ b/src/language-css/parser-postcss.js
@@ -21,7 +21,7 @@ const getHighestAncestor = (node) => {
   return node;
 };
 
-function parseValueNode(valueNode, options) {
+function parseValueNode(valueNode, options, parent) {
   const { nodes } = valueNode;
   let parenGroup = {
     open: null,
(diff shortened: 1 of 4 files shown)


src/language-css/parser-postcss.js
```diff
@@ -53,6 +55,34 @@ function parseValueNode(valueNode, options) {
       node.unit = "...";
     }
 
+    // Work around postcss bug parsing `list-*` as the word `list-` and operator `*`.
+    // postcss-values-parser flagged this as a wontfix: https://github.com/shellscape/postcss-values-parser/issues/108
```
evilebottnawi, line 59: We really need to migrate on own parser, `postcss-values-parser` is kitchen sink with a lot of bugs, spec violations and etc trash

tests/scss/scss/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4678,6 +4784,40 @@ label {
...
+  (
+    $black: #222,
+    $border-radius: 0.1rem,
+    $left: 10px,
+    $right: 400px,
+    $box-shadow: 0 0.5rem 1rem rgba($black, 0.15),
+    0 0.5rem 1rem rgba($black, 0.15),
+    0 0.5rem 1rem rgba($black, 0.15),
+    0 0.5rem 1rem rgba($black, 0.15)
+  );
```
evilebottnawi, line 4820: Do you think we should move `with` on the next line?

src/language-css/parser-postcss.js
AndrewLeedham: Agree long-term.

tests/scss/scss/__snapshots__/jsfmt.spec.js.snap
AndrewLeedham: Yes I am just not sure how. Still learning the ropes here.

PR #9073, opened 2020-08-26
Update js `languages`

fisker:
I have a question, why multiple parsers? When will the second one be used?


src/language-js/index.js
```diff
@@ -10,23 +10,23 @@ const languages = [
     require("linguist-languages/data/JavaScript.json"),
     (data) => ({
       since: "0.0.0",
-      parsers: ["babel", "flow"],
+      parsers: ["babel"],
```
fisker, line 13: I think we should only use `babel`

src/language-js/index.js
```diff
@@ -10,23 +10,23 @@ const languages = [
...
       vscodeLanguageIds: ["javascript", "mongo"],
-      interpreters: data.interpreters.concat(["nodejs"]),
+      interpreters: [...data.interpreters, "nodejs"],
     })
   ),
   createLanguage(require("linguist-languages/data/JavaScript.json"), () => ({
     name: "Flow",
     since: "0.0.0",
-    parsers: ["babel", "flow"],
+    parsers: ["flow", "babel-flow"],
```
fisker, line 21: Should prefer `flow` parser for `flow`

thorn0:
> I have a question, why multiple parsers? When will the second one be used?

It's up to API consumers to decide how to use this information.

fisker:
Means I can get all possible parsers, and try one by one. In that case, should we list more parsers.

For example, `babel` `flow` `babel-flow` for `js` and `jsx`?

thorn0:
Yes. This might come in handy for example in a Prettier-standalone scenario where only `parser-typescript.js` is loaded.

thorn0 (approved):

evilebottnawi (approved):

PR #9075, opened 2020-08-27
Refactor `RecoveredErrors` rethrow

fisker:
We can access the babel error messages now. 

~I'm thinking instead of listing the errors need rethrow, we may want list the errors we want ignore.
@thorn0~ Sorry, not a good idea

### DIFF
--- a/src/language-js/parser-babel.js
+++ b/src/language-js/parser-babel.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const errorMessages = require("@babel/parser/lib/parser/error-message");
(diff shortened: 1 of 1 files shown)


evilebottnawi (commented):
Can we add tests?

fisker:
We have tests covered.

https://github.com/prettier/prettier/blob/master/tests/misc/errors/js/module-attributes.js

https://github.com/prettier/prettier/blob/master/tests/misc/errors/babel-ts/issue-8041.ts

evilebottnawi (approved):

thorn0 (approved):

PR #9079, opened 2020-08-27
Rename native shims to `.mjs`

fisker:


scripts/build/build.js
```diff
@@ -22,7 +22,7 @@ process.on("unhandledRejection", (err) => {
   process.exit(1);
 });
 
-const CACHE_VERSION = "v28"; // This need update when updating build scripts
+const CACHE_VERSION = "v30"; // This need update when updating build scripts
```
evilebottnawi, line 25: Maybe we can calculate hash here and stop updating this each time :smile: ?

fisker: We do, but only related source files and `yarn.lock` is calculated, (build scripts calculated on CI too), we don't really need update it, just in case.

PR #9083, opened 2020-08-27
Don't merge: another test for #9057

thorn0:
#9057

### DIFF
--- a/src/cli/constant.js
+++ b/src/cli/constant.js
@@ -3,9 +3,7 @@
 const { outdent } = require("outdent");
 const { coreOptions } = require("./prettier-internal");
 
-const categoryOrder = [
-  coreOptions.CATEGORY_OUTPUT,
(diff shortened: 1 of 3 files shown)


thorn0:
Some results of this testing:
* The commit created by the action doesn't trigger the actions again. While I understand this works this way to prevent loops and is intended, it's not what is ideal for us as we'd like to run the tests for this commit too.
* Looks like this commit 90b78adcf17 to the base branch is ignored. Probably the action uses the workflow files from the commit that was the head of the base branch when the PR was created.

fisker:
> The commit created by the action doesn't trigger the actions again.

[`git-auto-commit-action` mentioned this](https://github.com/stefanzweifel/git-auto-commit-action#no-new-workflows-are-triggered-by-the-commit-of-this-action)

> Looks like this commit 90b78ad to the base branch is ignored. 

This seems unacceptable, if we want change this action, we can't do through PR

PR #9093, opened 2020-08-28
Make `isSCSS*` function stricter

fisker:
Followup #7933, these parts should only work for scss parser now.


tests/css/inline-url/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -107,7 +107,7 @@ printWidth: 80
 
 .validUnquotedUrlsThatAreParsedBySass {
   background: url($foo);
-  background: url($foo + $bar);
+  background: url($foo+$bar);
```
fisker, line 110: `css` parser should keep it as-it-is. I've duplicated this test in `tests/scss`

evilebottnawi (approved):

sosukesuzuki (approved):

Issue #1234, opened 2017-04-13
Upgrading to Prettier 1 strips all semis


ekosz:
Just upgraded to prettier 1.0 and it stripped out all of our semi-colons. But looking at the help doc it seems it shouldn't do that by default?

Here's the command we use:

```
prettier --write --print-width 100 --single-quote --trailing-comma all
```

vjeux:
/fail. Just pushed 1.0.2, sorry about this!

ekosz:
Perfect. Just upgraded and we're fixed. Thanks for the super fast turnover!

PR #9104, opened 2020-08-30
Remove the need to add full GitHub links in changelog entry titles

thorn0:
It should be enough to write just `#1234` and `@username` instead of full links. Links can be generated by the blog post script. 


scripts/lint-changelog.js
```diff
@@ -125,12 +125,6 @@ for (const category of CHANGELOG_CATEGORIES) {
       );
     }
 
-    if (title.startsWith(" ")) {
```
thorn0, line 128: Changelog files are formatted and checked by Prettier, so there can't be a space.

fisker (approved):
Good idea

evilebottnawi (approved):

PR #9105, opened 2020-08-31
Update cspell.json

fisker:
- Ignore urls in `()`
- Ignore `png` file (It's matched by cspell)
- Update `test.*` to `test*.*`

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -405,6 +405,7 @@
         "ve{2,}r{2,}y",
         "ve+r+y+long\\w*",
         "\\/.*?\\/",
+        "\\(https?://.*?\\)",
         "author: \".*?\"",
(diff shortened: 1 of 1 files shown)


evilebottnawi (approved):

thorn0:
Why ignore URLs? Why not let cspell catch typos in URLs?

fisker:
It's either right or wrong, why should we care about misspell in it?

thorn0:
We should care about our own typos that can happen in URLs.

fisker:
We use this  "\\/.*?\\/" to ignore regexp, so we  can't really catch a thing except `namee` in `https://github.com/namee`. the only part cspell reportted a lot, and nothing we can do.

thorn0 (approved):

PR #9129, opened 2020-09-02
Adds support for useTabs in Markdown (lists and indentation)

scinos:
Adds partial support for `useTabs` in Markdown files


scinos:
@fisker @sosukesuzuki any chance to get this PR reviewed? Is there anything I can do to make it easier to review?

fisker:
Thank you for your contribution, and sorry for the delay.

Personally, I don't like this idea, and some issue found in current implementation.

1. I don't think we do any align in other languages
1. I don't think this really possible to do it

	```html
	<pre style="tab-size: 4;"><code>
	9999.	123
	10000.	123
	10001.	123
	</code></pre>
	<pre style="tab-size: 2;"><code>
	9999.	123
	10000.	123
	10001.	123
	</code></pre>
	```

1. In html, the numbers align from the right side.

	```html
	<ol start=99>
	<li>123</li>
	<li>123</li>
	</ol>
	```

1. Possible bug [playground](https://deploy-preview-9129--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBOVA6ABARgEwDMAOlDgAxnb4EgA0IEADjAJbQDOyoAhgE68QA7gAU+CTim4AbQdwCenegCNe3MAGs4MAMqM1LKAHNkMXgFc49OAFslcACb2HAGW5Gz3Q3ABiEXte4YViNkEG4zGAg6EAALGGspAHUYlnh2PTA4bXFUlgA3VLlQsHZFEAN2OF4YYVVDAOQAM2lK+gArdgAPACFVDS1tbms4ZwM4JpbLEA7O7QNDKTgARTMIeAmpVpA9XkreUIDedXshKGjGXgMYRJZ7GBjkAA4yeguISsTVRlCLuD288b0ACOq3gtSYEjC7AAtFA4A4HNFeHAQSxkbVPA0kM1NlNKtYWCZzHj5osVmtxtjJvQYNwlDc7g8kHgaaoWFJ5gBhCDWLEgP4AVmiZkqABU6RJTBZ6HkLABJKBOWDaMCXZgAQUV2hgckWG0qAF8DUA)

scinos:
@fisker To be clear, your concern is about the alignment, or using tabs for Markdown in general?

About the alignment, there is some alignment already going on when using spaces, albeit quite confusing. See this example with different tab widths:

```
Input   | w:1         | w:2        | w:3        | w:4        | w:5        | w:6        | w:7        | w:8 or more
-------------------------------------------------------------------------------------------------------------------
8. a    | 8. a        | 8. a       | 8. a       | 8. a       | 8. a       | 8. a       | 8. a       | 8. a
9. a    | 9. a        | 9. a       | 9. a       | 9. a       | 9. a       | 9. a       | 9. a       | 9. a
10. a   | 10. a       | 10. a      | 10.   a    | 10. a      | 10. a      | 10. a      | 10. a      | 10. a
-------------------------------------------------------------------------------------------------------------------
8.  a   | 8. a        | 8.  a      | 8. a       | 8.  a      | 8.   a     | 8.    a    | 8. a       | 8. a
9.  a   | 9. a        | 9.  a      | 9. a       | 9.  a      | 9.   a     | 9.    a    | 9. a       | 9. a
10. a   | 10. a       | 10. a      | 10.   a    | 10. a      | 10.  a     | 10.   a    | 10.    a   | 10. a
```

Looks like when the input is not aligned (first row), the output is not aligned as well (except for `tab-width:3`, which looks like a bug).

When the input is aligned (second row), the behaviour is quite erratic. `tab-width:1` or `8` kills the alignment, `3` and `7` misalign the `10.`, there is no difference between `2` and `4`... Honestly I don't know what's the expected behaviour. I should probably file bugs for those cases and start a separate discussion.

When using tabs, it behaves in a much more predictable way, mainly because it fixes `tab-width` to `4` (tabs indicated with `░` for easier visualization):

```
Input   | w:any
---------------
8. a    | 8.░░a
9. a    | 9.░░a
10. a   | 10.░a
---------------
8.  a   | 8.░░a
9.  a   | 9.░░a
10. a   | 10.░a
```

The main difference is that `use-tabs` introduces alignment where there wasn't any. This was done to avoid this corner case:

```
// Input
1. an item
2. a very very very long item that wraps to the next line

// Output without tabs
1. an item
2. a very very very long item that wraps
   to the next line

// Tabs, option #1: use tabs to align the wrapped line
1. an item
2. a very very very long item that
░░░░wraps to the next line

// Tabs, option #2: use tabs to align the second item
1. an item
2.░░a very very very long item that
░░░░wraps to the next line

// Tabs, option #3: use tabs to align all lines
1.░░an item
2.░░a very very very long item that
░░░░wraps to the next line
```

Option 1 and Option 2 look misaligned, I don't think that's what the user expects. So the only alternative that makes sense in my head is Option 3, but it means automatically adding alignment even when there was none.

findusl:
It would be great if we could get this feature. Some team members got used to the 3 space identation width of VSCode and they don't want to use 4 spaces. But MkDocs does not display lists correctly when indented with 3 spaces. 

Our solution would be indenting with tabs and show them as 3 spaces in VSCode. Everyone is happy. Except prettier cannot format with tabs so we need to switch to a separate linter/formatter for markdown.

PR #9138, opened 2020-09-03
 feat(markdown): option to disable printing spaces between Chinese / Japanese and others

tats-u:
Based on #8526 (Thanks to @rdrgn)

 Differences from #8526:
 - Don't touch on Hangul.
 - Change option name & type & description.

To do:
- [ ] Deal with `proseWrap`
- [ ] Add test cases for `proseWrap`

I want this program to warn me to specify this option when the system locale is Chinese or Japanese and this option is not explicitly specified. How do I do this?


src/language-markdown/options.js
```diff
@@ -1,9 +1,21 @@
...
   singleQuote: commonOptions.singleQuote,
+  insertCjSpacing: {
+    since: "2.1.2",
+    category: CATEGORY_MARKDOWN,
+    type: "boolean",
+    default: true, // transitional measure until a plugin to shoulder this feature becomes available
+    description: outdent`
+      Insert spaces wherever between Chinese or Japanese and non-CJK.
+    `,
+  },
```
evilebottnawi, line 20: Please read answer in the issue

tats-u:
Since adding new options is not allowed, I will try to solve the issues with a different way (including pre-processing).

PR #9156, opened 2020-09-06
Added `Icons8` to user list

visualpharm:
I don't know if it's your thing, but we are popular among coders, we pledged $40k+ to open source, and we open source ourselves. And yeah, we use you a lot.


fisker (approved):

fisker:

evilebottnawi:
@fisker Probably yes, almost everyone is using Prettier

PR #9167, opened 2020-09-08
Fix plugin search on `pnpm`

### LINKED ISSUE #8056: [Bug?] Symlink'd plugins in `node_modules` are not auto detected.
**Environments:**

- Prettier Version: 2.0.4
- Running Prettier via: VSCode, CLI
- Runtime: Node 12.16.2
- Operating System: OS X

**Steps to reproduce:**

1. Use `npm` to install  [prettier-plugin-organize-imports](https://www.npmjs.com/package/prettier-plugin-organize-imports)
2. Run `prettier foo.txs`
3. Observe that imports are correctly organized

This time, use `pnpm` which uses symlinks instead of copying packages to `node_modules`:

1. Use `pnpm` to install [prettier-plugin-organize-imports](https://www.npmjs.com/package/prettier-plugin-organize-imports)
2. Run `prettier foo.tsx` from the command line
3. Observe that imports are *NOT* organized

Now, add the plugin explicitly to `package.json`:

```json
  "prettier": {
    "plugins": [
      "./node_modules/prettier-plugin-organize-imports"
    ]
  }
```

Repeat step 2 above and observe that plugins are now organized.

**Expected behavior:**

Symlink'd plugins in `node_modules` are auto detected.

**Actual behavior:**

Symlink'd plugins in `node_modules` are not auto detected.

fisker:
Local tested, it not easy to test, because this logic need plugin installed in same location with `prettier`.

Fixes #8056

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -7,7 +7,7 @@ const partition = require("lodash/partition");
 const globby = require("globby");
 const mem = require("mem");
 const internalPlugins = require("../languages");
-const thirdParty = require("./third-party");
+const {findParentDir} = require("./third-party");
 const resolve = require("./resolve");
 
 const memoizedLoad = mem(load, { cacheKey: JSON.stringify });
@@ -27,9 +27,17 @@ function load(plugins, pluginSearchDirs) {
   }
   // unless pluginSearchDirs are provided, auto-load plugins from node_modules that are parent to Prettier
   if (!pluginSearchDirs.length) {
-    const autoLoadDir = thirdParty.findParentDir(__dirname, "node_modules");
-    if (autoLoadDir) {
-      pluginSearchDirs = [autoLoadDir];
+    const nodeModulesDir = findParentDir(__dirname, "node_modules");
+    if (nodeModulesDir) {
+      const dotPnpmDir = findParentDir(nodeModulesDir, ".pnpm");
+      if (dotPnpmDir) {
+        const nodeModulesDir = findParentDir(dotPnpmDir, "node_modules");
+        if (nodeModulesDir) {
+          pluginSearchDirs = [nodeModulesDir];
+        }
+      } else {
+        pluginSearchDirs = [nodeModulesDir];
+      }
     }
   }
 
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -7,7 +7,7 @@ const partition = require("lodash/partition");
 const globby = require("globby");
 const mem = require("mem");
 const internalPlugins = require("../languages");
-const {findParentDir} = require("./third-party");
+const { findParentDir } = require("./third-party");
 const resolve = require("./resolve");
 
 const memoizedLoad = mem(load, { cacheKey: JSON.stringify });
@@ -27,17 +27,11 @@ function load(plugins, pluginSearchDirs) {
   }
   // unless pluginSearchDirs are provided, auto-load plugins from node_modules that are parent to Prettier
   if (!pluginSearchDirs.length) {
-    const nodeModulesDir = findParentDir(__dirname, "node_modules");
-    if (nodeModulesDir) {
-      const dotPnpmDir = findParentDir(nodeModulesDir, ".pnpm");
-      if (dotPnpmDir) {
-        const nodeModulesDir = findParentDir(dotPnpmDir, "node_modules");
-        if (nodeModulesDir) {
-          pluginSearchDirs = [nodeModulesDir];
-        }
-      } else {
-        pluginSearchDirs = [nodeModulesDir];
-      }
+    const autoLoadDir =
+      findParentDir(__dirname, "node_modules/.pnpm") ||
+      findParentDir(__dirname, "node_modules");
+    if (autoLoadDir) {
+      pluginSearchDirs = [autoLoadDir];
     }
   }
 
--- a/changelog_unreleased/api/pr-9167.md
+++ b/changelog_unreleased/api/pr-9167.md
@@ -1,0 +1,3 @@
+#### Fix plugins search with `pnpm` (#9167 by @fisker)
+
+Previously, when install `prettier` and plugins with `pnpm`, the installed plugins can't load automaticlly.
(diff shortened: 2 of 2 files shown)


evilebottnawi (commented):
Can we add simple tests?

fisker:
To setup this test, I need install `pnpm`, init a project, install `prettier` and one plugin, and run prettier from this project.

Seems not worth to do that.

evilebottnawi:
@fisker we can recreate the structure to make sure it works, in the future we can better automate testing, fixes without tests is bad idea,

fisker:
How do we put prettier self into `node_modules/.pnpm/prettier@2.1.1/node_modules/prettier/`?

evilebottnawi:
@fisker we can mock `node_modules` in tests and recreate structure in `test/fixtures/node_modules/.pnpm`

aparajita:
Thanks for this @fisker!

wighawag:
Would be great to have this merged, I switched to npm for my app template because of that.
is the tests the pending issue or is there something else ?

aparajita:
@wighawag I'm using pnpm with prettier. There's an easy workaround, use the `--plugin-search-dir` or `--plugin` command line option.

wighawag:
@aparajita I can make it work from the command line, 
unfortunately it does not work on vscode plugin, but this fix proposed here might not fix the vscode plugin issue neither.
In vscode I also get the issue for eslint when used in conjunction with prettier

thorn0:
FWIW, `plugins` and `pluginSearchDirs` can be specified in the config too. This might help with the VS Code use case.

wighawag:
@thorn0 I tried that but without success, there seems to be other problem with the vscode plugin

fisker:
@wighawag  Are you running a global installed Prettier?

wighawag:
No I use locally installed in the root of the mono repo.
My current workaround is to install prettier via npm in a `_npm` folder and use that path in the prettier vscode config
You can see it in this repo : https://github.com/wighawag/purple-warlock/tree/pnpm

aparajita:
@wighawag You should file an issue with the eslint plugin maintainer.

wighawag:
@aparajita you mean the prettier plugin I suppose

aparajita:
> @aparajita you mean the prettier plugin I suppose

Oops, yes.

src/common/load-plugins.js
```diff
@@ -27,7 +27,9 @@ function load(plugins, pluginSearchDirs) {
   }
   // unless pluginSearchDirs are provided, auto-load plugins from node_modules that are parent to Prettier
   if (!pluginSearchDirs.length) {
-    const autoLoadDir = thirdParty.findParentDir(__dirname, "node_modules");
+    const autoLoadDir =
+      findParentDir(__dirname, "node_modules/.pnpm") ||
+      findParentDir(__dirname, "node_modules");
```
drtyh2o, line 32: I don't believe this will work.`node_modules/.pnpm` is the default pnpm virtual store directory, but can be overridden by the `virtual-store-dir` setting in `.npmrc`.  I think a better solution would be to see if the target path exists and is a directory in `src/common/load-plugins.js`:

**Before**
```js
function isDirectory(dir) {
  try {
    return fs.statSync(dir).isDirectory();
  } catch (e) {
    return false;
  }
}
```
**After**
```js
function isDirectory(dir) {
  try {
    return fs.statSync(fs.realpathSync(dir)).isDirectory()
  } catch (e) {
    return false;
  }
}
```

fisker: Thanks for the information, any suggestion?

drtyh2o: Not sure. It seems the default path (or whatever is specified in `pluginSearchDirs`) is being resolved relative to `process.cwd()` which sits inside the pnpm virtual store. Packages in the virtual store aren't stored hierarchically so the concept of `findParentDir` does not really work. With pnpm, the package hierarchy is captured in the symlink structure, so the plugin searching will need to use the symlink (not the target) somehow. The only way I have gotten plugins to load is by specifying them in `.prettierrc` using a relative path, e.g. `"plugins": ["./node_modules/prettier-plugin-packagejson"]`

drtyh2o: I also use [`eslint` plugins](https://eslint.org/docs/user-guide/configuring/plugins) which seem to work fine with `pnpm` ... Looking at the eslint docs/internals it seems they resolve plugins as if you were to `require` them from the config file.

aminya (approved):

src/common/load-plugins.js
aminya: Could you accept this PR as is and defer fine tuned configuration. Prettier doesn't work with PNPM

lizzi2122 (approved):

src/common/load-plugins.js
drtyh2o: Prettier does work with pnpm, it just won’t auto load plugins. You can load them manually in the .prettierrc, for example:  `"plugins": ["./node_modules/prettier-plugin-packagejson"]`

adamhl8:
For anyone following this thread, I've created a PR that hopefully addresses this issue.

https://github.com/prettier/prettier/pull/11248

syed-haroon:
I am switching from PNPM to NPM :(

fz6m:
Rename `.prettierrc` to `.prettierrc.js`, use absolute plugin path to resolve this problem.

```js
// .prettierrc.js

module.exports = {
  plugins: [
    require.resolve('prettier-plugin-packagejson'),
    // ...
  ]
}
```

yavko:
> Rename `.prettierrc` to `.prettierrc.js`, use absolute plugin path to resolve this problem.
> 
> ```js
> // .prettierrc.js
> 
> module.exports = {
>   plugins: [
>     require.resolve('prettier-plugin-packagejson'),
>     // ...
>   ]
> }
> ```

this doesn't work for me

thonkinator:
any updates on this? it's been open for over 2 years now

fz6m:
Currently the vscode prettier plugin supports resolve pnpm symlinks.

So you don't need `require.resolve` anymore.

```ts
// .prettierrc
{
  "plugins": ["some-plugin"]
}
```

thonkinator:
@fz6m really? for me it doesn't detect pnpm packages either

henrikvilhelmberglund:
This was broken for me and I had to migrate from pnpm to npm (loading plugins explicitly didn't work)

PR #9180, opened 2020-09-10
CSS: format camelCase props to kebab-case

zweimach:
First, I want to thank you for your works on Prettier.

I add a little tweak for the CSS printer that will transform camelCase to kebab-case, pretty useful when I copy CSS-in-JS rules:
```css
.fromCssInJsToCss {
  backgroundAttachment: fixed;
}
```
and Prettier can transform that to:
```css
.fromCssInJsToCss {
  background-attachment: fixed;
}
```

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -32,6 +32,7 @@ const {
   getAncestorNode,
   getPropOfDeclNode,
   maybeToLowerCase,
+  maybeToKebabOrLowerCase,
   insideValueFunctionNode,
   insideICSSRuleNode,
   insideAtRuleNode,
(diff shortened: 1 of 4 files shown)


zweimach:
Feel free to close this PR if it violates Prettier philosophy.

fisker:
I don't think the Prettier team would accept this, but you can make a plugin for it.

https://prettier.io/docs/en/plugins.html#developing-plugins

evilebottnawi (changes requested):
Output of scope prettier, please use stylelint and create plugin for this

zweimach:
Hi @fisker, thanks for your suggestion. I think I'll do that instead.
However, I still believe Prettier should treat camelCased props differently instead of just blindly lowercasing them.
Again, thank you for your works on Prettier that makes the JavaScript world a much better place.

Issue #4129, opened 2018-03-11
Reduce duplication of format and check scripts


nickmccurdy:
Most projects I've seen using Prettier have local configs to support editor plugins and build scripts using `--write` and `--list-different` for local formatting and CI checks respectively. For example, here's what Redux uses in its `package.json` scripts:

```json
"format": "prettier --write \"{src,test}/**/*.js\"",
"format:check": "prettier --list-different \"{src,test}/**/*.js\"",
```

I see this duplication of file arguments often because prettier configs don't have file lists, and I'd like to find a way to make this use case cleaner.

## Options
1. Adjust the `--write` flag to behave like `--list-different` when the `CI` environment variable is true. Most CI environments set this variable, and it's already used by various tools to prevent writing to files in CI environments (for example, Jest snapshots aren't written on CI environments by default). Then pairs of scripts like these could simply be replaced with a single `--write` script. This could technically break existing CI builds that accidentally reformatted code.
2. Same as 1 but use a new flag like `--auto` to keep back compatible behavior in CI.
3. Provide a separate config file allowing non-formatting options so you can define the file list once.
4. Allow the `--list-different` and `--write` flags to disable each other (enabling whichever one was specific last) to allow scripts to be composed by appending overriding arguments.

evilebottnawi:
@nickmccurdy who don't use?
```
"format": "npm run format:check -- --write",
"format:check": "prettier --list-different \"{src,test}/**/*.js\"",
```

nickmccurdy:
Doesn't that still fail when files are unformatted? And either way, I still like the idea of only having to add a single command to `package.json` scripts because this use case is so common. Prettier should act similarly across projects, right now users have to memorize the setups their packages are using and the scripts often differ significantly.

duailibe:
Yeah I'm currently thinking an API that makes more sense is to remove `--list-different` and make the default be that. Users can then add to `package.json`:
```
"prettier": "prettier \"{src,test}/**/*.js\"",
```

And just `npm run prettier` to list files that changed and `npm run prettier -- --write` to write files. The `--list-different` is really weird.

(I also think this is being discussed already in some issue but I can't remember)

nickmccurdy:
Alternatively, I would prefer to have `--write` as the default normally and `--list-different` as the default in CI when paths are given without either flag. If neither flag nor any paths are given, it uses stdout as with the current default behavior to prevent integrations from breaking.

j-f1:
[...]

PR #9181, opened 2020-09-10
feat: add --force-error-on-written-files

thomasruiz:
Add --force-error-on-written-files flag. When running `prettier --check -w` in a git hook, the commit still works but the formatted files are not staged (unless you do `prettier --check -w; git add .`, but using that command can lead to other issues).


evilebottnawi (changes requested):
O can't understand the problem, why you need this flag?

fisker:
I think the problem @thomasruiz facing is:

Use `prettier --write .` to format files, but git hook won't commit the unstaged files, so have to use `git add .`, but this "can lead to other issues".

I think the problem is about how you use Prettier, 

If you only want run prettier on staged files, you should `prettier --write staged_file_1` `git add staged_file_1`.

If you only want run prettier on all files, you should run `prettier --write .` manually, before you commit.

thomasruiz:
That's exactly it. Here is an example :

```
$ echo 'var    unformatted =    ''; > index.js
$ cat package.json
{
    "license": "MIT",
    "dependencies": {
        "husky": "^4.3.0",
        "prettier": "^2.1.1"
    },
    "husky": {
        "hooks": {
            "pre-commit": "npx prettier -w --check index.js" 
        }
    }
}
$ git add .
$ git commit -m 'test'
husky > pre-commit (node v14.9.0)
Checking formatting...
[warn] index.js
[warn] Code style issues fixed in the above file(s).
[master b346303] test
 1 file changed, 1 insertion(+), 1 deletion(-)
$ git status
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   index.js

no changes added to commit (use "git add" and/or "git commit -a")
$ git diff
diff --git i/index.js w/index.js
index 30346da..b72385d 100644
--- i/index.js
+++ w/index.js
@@ -1 +1 @@
-var    unformatted =    '';
+var unformatted = '';
```

fisker:
What you need is https://github.com/okonet/lint-staged, not an option here.

thomasruiz:
I'm not keen to add yet another package to help maintain a proper codebase. Why is it a problem to add this option to prettier?

thomasruiz:
@fisker @evilebottnawi any update on this?

fisker:
I'm sorry, I'm afraid the Prettier team have no interest to add this option.

jasonkarns:
This isn't only a limitation necessary for `lint-staged`.

`--check` is supposed to exit nonzero when files need formatted.
Passing `--check` and `--write` together gives the best possible output (lists only the files that were modified). But passing both options and suddenly prettier exits with a successful status code, even when files were written!

If `--check` + `--write` together would exit per the normal rules (ie, files were in violation=nonzero), then users could do away with the nonsense of having two separate lint scripts.  A single `lint: "prettier --write --check ."` would work for writing files locally, and function as a proper posttest script on CI.

It would cover this issue as well as #4129

thorn0:
Actually, I have no idea why the undocumented combo `--check --write` currently works the way it does and whether the successful status code after changing files is intended. Changing that (returning non-zero) makes sense to me.

@fisker @alexander-akait Any idea on the reasons for the current behavior? I can't find where it was discussed.

fisker:
I don't know the history. But I think we should keep current behavior. 

1. Passing `--write` means files will be fixed, so `--check` passes. Run `eslint --fix`, if all problems are fixable, there will be no error.
2. If we change exit code, we won't able to do `prettier . --write && git commit -am`
3. If we fobid use `--check` and `--write` together, the will be problem too, see this line https://github.com/prettier/prettier/blob/5682d2f4405d843ed77d8c13b8048d7b7dfd8b49/package.json#L158

jasonkarns:
> If we change exit code, we won't able to do prettier . --write && git commit -am

That would become `prettier . --write --check || git commit -am` which is, IMO, more clear. The git action should be explicitly dependent upon prettier making a change.

thorn0:
> If we change exit code, we won't able to do prettier . --write && git commit -am

What I want to discuss is changing the exit code only for `--check --write`, not for `--write` without `--check`.

fisker:
> What I want to discuss is changing the exit code only for --check --write

Yes, I mean this, I know no one will use `prettier . --write --check && git commit -am`, but still


```json
{
	"scripts": {
		"check": "prettier a_long_long_long_long_long_pattern --check",
		"fix": "prettier a_long_long_long_long_long_pattern --write"
	}
}
```

`yarn check --write` is nicer, no need repeat the long arguments in `check` script.

@jasonkarns I think run `--check --write` in `husky` will auto turn in to `--write --check && git add`?

PR #9209, opened 2020-09-16
Update `postcss` to v8

fisker:


src/language-css/parser-postcss.js
```diff
@@ -300,6 +300,33 @@ function parseNestedCSS(node, options) {
...
+            node.source.end.offset + 1
+          );
+        const fakeContent = textBefore.replace(/[^\n]/g, " ") + nodeText;
+        node.value = {
+          type: "css-rule",
+          nodes: parseCss(fakeContent, [], { ...options }).nodes[0].nodes,
+        };
+        return node;
+      }
+    }
```
evilebottnawi, line 328: I think we should remove them in the future major release, it is invalid

tests/css/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -887,11 +925,11 @@ article /* comment 168 */ :--heading /* comment 169 */ + /* comment 170 */ p /*
 /* custom properties set & @apply rule */
 :root {
   /* comments 192 */
-  --centered /* comments 193 */ : /* comments 194 */ {
+  --centered/* comments 193 */ : /* comments 194 */ {
     display: flex;
     align-items: center;
     justify-content: center;
-  }
+  };
```
evilebottnawi, line 932: Regression?

src/language-css/parser-postcss.js
fisker: I was going to parse is as `value-unknown`, but seems it is a proposal, but abandoned

http://tabatkins.github.io/specs/css-apply-rule/#defining

tests/css/comments/__snapshots__/jsfmt.spec.js.snap
fisker: Acutally, I think it's bugfix.

`--centered` is a prop, it should add `;`, the space before `/* comments 193 */` is gone, it's because we print `decl` like this, [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDABMAOld7IA2EATugPQBU+EAtjQjOhWUugMQBmADFwNzo4AviAA0ICAAcYAS2gBnZKFTFiEAO4AFZQgUpUBNagCeCsQCNiqMAGs4MAMoSr0qAHNkMYgFc4YuDTM4ABMg4IAZVDcvVFc4ADESGlQYGTdkNC8YCFEQAAsYGgIAdVzpeDknMDh7HTLpADcyo3SwOVMQFzk4YhgNS1ck5A59LrEAKzkADwAhSxs7e1R6MJc4IZHfEAnJ+xdXAjgARS8IeHWCUZAnYi7iFraciWIXGCLpIJhc5AAOLjEniBdIqWCTpJ5wW71NZiACOJ3gfUkujQcgAtFA4MFgjliHA4dJcX0YoMkMMLpsujRpB5vBS9gdjqc1qSNmIYKgzG8Pl8kAAmNmWaQEPYAYVoJJAEIArDkvF0ACoc3Rky71HwASSgoVg9jAzykAEEtfYYEYDucuoJBEA).

There is a space, because we treat as `rule`'s `selector` before, [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDABAegFTsgW3wRnW03WAB0p08IAbCAJyXQGIAzABi4G51qAXxAAaEBAAOMAJbQAzslComTCAHcACsoQKUqemtQBPBWIBGTVGADWcGAGUJV6VADmyGEwCucMXHxmcAAmQcEAMqhuXqiucABizPioMDJuyGheMBCiIAAWMPj0AOq50vByTmBw9jpl0gBuZUbpYHKmIC5ycEwwGpauScgc+l1iAFZyAB4AQpY2dvaoRGEucEMjviATk-YurvRwAIpeEPDr9KMgTkxdTC1tORJMLjBF0kEwucgAHFxiTxAukVLBJ0k84Ld6msxABHE7wPqSXRoOQAWigcGCwRyTDgcOkuL6MUGSGGF02XXw0g83gpewOx1Oa1JGzEMFQZjeHy+SAATGzLNJ6HsAMIQQiodIQgCsOS8XQAKhzdGTLvUfABJKChWD2MDPKQAQW19hgRgO5y6gkEQA).

fisker: And the test is added by you, the input has `;` https://github.com/prettier/prettier/blob/e857c4e88ffbf0c2fb213ee7172fbe9431ab87a8/tests/css/comments/selectors.css#L156

fisker: I think we should print `less` and `scss` like this too.

src/language-css/parser-postcss.js
evilebottnawi: Yes, but it is violate CSS spec, so prefer to avoid this, because it can break some future CSS features

tests/css/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: scss (need test on less) supports nested declarations https://www.sassmeister.com/, so prefer keep it as is

fisker: https://www.sassmeister.com/ add ';' too.

fisker: http://lesscss.org/less-preview/

Can parse

```less
:root {
    /* comments 192 */ --centered /* comments 193 */ : /* comments 194 */ {
        display: flex;
        align-items: center;
        justify-content: center;
    };
}
```

But can't parse without `;`

```less
:root {
    /* comments 192 */ --centered /* comments 193 */ : /* comments 194 */ {
        display: flex;
        align-items: center;
        justify-content: center;
    }
}
```

evilebottnawi: Remove `;` and no `;` in output

evilebottnawi: For less we should not change broken output, it is AST modification, so for less and sass better to keep input as is

tests/css/postcss-8-improment/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -32,7 +78,7 @@ https://github.com/postcss/postcss/releases/tag/8.0.0
 */
 
 :root {
-  --empty:  ;
+  --empty: ;
```
evilebottnawi, line 81: Good catch :+1:

evilebottnawi: `--empty:;` is invalid syntax
`--empty: ;` is valid syntax

So both should not be changed

tests/css/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Can we add test without `;`?

tests/css/postcss-8-improment/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,71 @@
...
+  --JSON: [1, "2", {"three": {"a": 1}}, [4]];
+  --javascript: function(rule) {console.log(rule)};
+}
+
+@supports (--element(".minwidth", {"minWidth": 300})) {
+  [--self] {
+    background: greenyellow;
+  }
+}
```
evilebottnawi, line 68: Can we add `fixes` to the PR message and add changelog? Or you can do it in the separate PR?

evilebottnawi:
@fisker Ready for review?

fisker:
Yes, please review.

tests/css/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: and we should not fix invalid syntax, it is ast modification

fisker: @evilebottnawi I only changed the output for css,  it's custom property, it should add `;` if it's last one, like other properties. The output for less and scss didn't change. I've add tests.

The following one is not adding `;` 

**Prettier pr-9209**
[Playground link](https://deploy-preview-9209--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuZAnCEYAJgB0pZYC0RAZhkngYVpADYSpJYDEpADOwNz6EC+vYkQCGUTAAs4TVp1I8ofEABoQEAA4wAltADOyUMNToA7gAVDCPSmF1jwgJ56VAI1TCwAazgwAymveaUADmyDCoAK5wKnAAts5wACYJiQAyokHhwkFwAGKMMcIwWsHIIMLhMBDKIOIwMXQA6uKa8Dr+YHA+li2aAG4t9qVgOk4ggTpSMKZuQQXIpDYTKgBWOgAeAEJunt4+wjFwKYFw84tRIKtrPoFBdHAAiuGYJ0gLdEsg-qgTqEMj1WpUIEYA1NAkYOJkAAOdgqQEQCYNNxqUqAuA-XonFQARye8Gm6isZR0RCgcESiWqqDguM01OmWTmrzOKgmMU0oQi5x0Nzuj2ep3e5xgwmcoPBkKQACYVGFhJo6DcAMIQGJMkDogCs1XCEwAKqKrG8Pr1IgBJKDJWA+MBAjQAQUtPhg9jugomfD4QA)
```sh
--parser css
```

**Input:**
```css
:root {
  --foo:{
    color: #f00;
  }
  --another: #00f;
}
```

**Output:**
```css
:root {
  --foo: {color: #f00;} --another: #00f;
}

``` 
WDYT?

evilebottnawi: I think we should not put it in one file, it is like declarations, so better to have each declaration on own line

fisker: The whole part after `--foo` is parsed as value. And sent to `value-parser`, how do we tell this should not parse as `value`?

fisker: Starts with `{`?

evilebottnawi: It we have `--` at the start and `: {`

fisker: Okay, I'll change this  case to  let the value parse  as  `value-unknown`

fisker: Fixed, https://github.com/prettier/prettier/pull/9209/commits/c5e3bdd0e1f41dba79d1fab3866283c634b0d8e6.

evilebottnawi (approved):
Looks good, in the next major release we should print `--var: { }` as is do not touch this, because it is violate CSS spec

sosukesuzuki (approved):

PR #9275, opened 2020-09-26
Add support for [[wiki-style]] links in markdown

iamrecursion:
1.  Support for [[wiki-style]] links ensures that they do not get broken
    by the prettier formatting by being wrapped over multiple lines.
2.  The contents of the link (between the `[[]]` brackets) are treated
    as raw text. This is because the various tools that rely on the
    `[[]]` syntax do not agree on what is allowable between them.

Closes #9275.


fisker (approved):
LGTM

iamrecursion:
Thanks so much! Is there anything that needs doing before merge?

fisker:
Maybe add more tests like

```markdown
[[a
b]]
```

```markdown
[[[a]]]
```

```markdown
[[[a]]
```

```markdown
[[a]]]
```

```markdown
[[a[b](http://www.example.com/)]]
```

iamrecursion:
Done!

iamrecursion:
Any ideas as to the failure on Node 14? I can't reproduce it locally as I'm on macOS, but I can't see any rhyme or reason for it succeeding on macOS and Windows but not on Linux.

fisker:
`FULL_TEST=1 yarn test`

fisker:
The test failed because

```markdown
[[a
b]]
```

ast changed.

iamrecursion:
But isn't that expected in this case?

fisker:
@iamrecursion  

If I understand this correctly, 

```js
[[a
b]]
```

is not a `wikiLink`, right?

iamrecursion:
Ah, I see what you mean. In all honesty I'm not sure _what_ the most sensible behaviour is here, so following whatever happens with a standard `[]()` link is probably best. Can you point me to how this is handled for standard links?

fisker:
I'm not sure if current output is expected, if it is, this is easy to fix, we just convert the first AST from `linkReference` in to `wikiLink`.

fisker:
You can add something like

```js
  // `linkReference` with line breaks maybe covert to `wikiLink` after print
  if (ast.type === "linkReference" && ast.label.includes('\n')) {
    return {
		type: "wikiLink",
		value: ast.label
	      .trim()
	      .replace(/[\t\n ]+/g, " ")
	      .toLowerCase()
	};
  }
```
[here](https://github.com/prettier/prettier/blob/d096a34194f64e0bcf75e83769850de23e1b07c5/src/language-markdown/printer-markdown.js#L931) to pass test.

fisker:
> If prettier doesn't turn the following into a `link`, then it shouldn't turn the test case into a `wikiLink`.
> 
> ```
> [foo bar
> baz](https://example.com)
> ```

This is not the failed case, [this one](https://github.com/prettier/prettier/pull/9275#issuecomment-699841614) is.

iamrecursion:
> This is not the failed case, [this one](https://github.com/prettier/prettier/pull/9275#issuecomment-699841614) is.

I understand that, but my thinking is that the `wikiLink` test case should match the behaviour of of a standard link [like this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBtAhgAgEYCcIDWCAOlADYCWUBAugBQBmEEO6uAlCADQgQAOMCtADOyUG3wB3AApsEolOjKT0AT1E886MERgBlPtqoBzZDFwBXODzgBbbHAAmjpwBl0UYxfTG4AMQhcW3QYQU9kEHQLGAhuEAALGFsyAHV4inhhQzA4PXkMigA3DNUIsGENECphOFwYaVwfYOQGJRqeACthAA8AIUadOH10WzhXKjgWtusQLu69EzI4AEULCHgpsnaQQ1wa3Ajg3AJHCEkoOL5cKhgUikcYeOQADgAGHiuIGpTGvgiruD7QqTHgARzW8Aa-AUkWEAFooHAnE44rg4OCKGiGk10JttjVbBQzJYZsJFisIZMkK0tjMYOhsHcHk8kAAmHjmdAUSieADCEFszRQgIArHELDUACoMhQ07aFKwASSgLlgejA1wEAEEVXoYKolni4ABfY1AA) where it doesn't get formatted.

iamrecursion:
Okay, I've finally had a chance to look into this, and while the proposed band-aid fixes the aforementioned new test case, it causes breakage of `example-518.md` in the markdown tests. Instead I've settled for allowing the wikiLink regex to match over multiple lines, and stripping out the line breaks during parsing.

Please let me know what your thoughts are!

iamrecursion:
Hey @fisker is there anything else I need to do to have this merged?

fisker (commented):
LGTM

fisker (approved):
LGTM

iamrecursion:
As I don't have write access you'll have to hit merge yourself I think @fisker!

src/language-markdown/parser-markdown.js
```diff
@@ -96,6 +97,34 @@ function liquid() {
   };
 }
 
+function wikiLink() {
+  const entityType = "wikiLink";
+  const wikiLinkRegex = /^\[\[(?<linkContents>.+?)]]/s;
```
evilebottnawi, line 102: Hope we don't break content using regexp here

thorn0 (changes requested):
That's not what I expect from `--prose-wrap never`:

**Prettier pr-9275**
[Playground link](https://deploy-preview-9275--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAzAzgEwwAgNp4bYA6U6AhtgLpUgA0IEADjAJbTrKjkBOPEAdwAKvBJxTkANgPIBPTgwBGPcmADWcGAGUmq1lADmyGDwCucBnAC2iuJkx2AMuUOnyBuADEIPK+Rhshsgg5KYwEPQgABYwVpIA6lGs8Oi6YHBaYsmsAG7JssFg6Aog+uhwPDBCKgZ+yKhS5QwAVugAHgBCKuqaWuRWcI76cPWNFiCtbVr6BpJwAIqmEPCjkk0gujzlPMF+PGqYglCRTDz6MPGsmDBRyAAcAAwMpxDl8SpMwVBwORWRAI5LeDVZjiELoAC03zsdkiPDggNY8Oq7jqSAaa3G5SsrGMZixMzmi2WI3RYwYMHIiku11uSAATBSVKxJDMAMIQKxokBwdAAVkipnKABUqeIMescuYAJJQBywLRgM4sACCcq0MFkc1W5QAvrqgA)
```sh
--parser markdown
--prose-wrap never
```

**Input:**
```markdown
fsdfs [[fsdf
sadf]]
```

**Output:**
```markdown
fsdfs [[fsdf sadf]]

```

src/language-markdown/parser-markdown.js
iamrecursion: Can you clarify more what you're worried about?

iamrecursion:
This is what I was a little confused by. I'll look into it - sorry for missing that!

iamrecursion:
@thorn0 I think this should address your concerns!

thorn0 (approved):
LGTM

iamrecursion:
Great! What's the policy for merging things?

thorn0:
> What's the policy for merging things?

PRs like this one need 3 approvals from the maintainers. For simpler changes 2 is enough.

iamrecursion:
That makes perfect sense! Thanks. ^^

evilebottnawi (approved):

iamrecursion:
Thank you to all of you for guiding me through this! Contributing to prettier has been a really good experience.

Issue #8832, opened 2020-07-24
CLI should not print output when multiple files matched


fisker:
When formatting files but forgot `--write` flag, the cli will printed all the codes. It's annoying, and not much useful I think.

Suggest only print code, when one `filepath`(not pattern) or `stdin`.

WDYT?

brodybits:
What would be the new behavior? Exit with an error message or just do nothing? (I would favor doing `process.exit(1)` with an error code.)

fisker:
```console
prettier foo.js
```

check file exists
    a) not exist, throw NotFoundError, process.exit(1)
    b) found, keep current behavior

```console
prettier foo.js bar.js
```

throw MultipleInputNotAllowError , process.exit(1)

```console
prettier [foo].js
```

Same as `foo.js`(do not treat as glob pattern)

brodybits:
> throw MultipleInputNotAllowError , process.exit(1)

In that case, I would favor an explanation that check or write option is needed for multiple input files.

> ```
> prettier [foo].js
> ```

I thought that would match `f.js` & `o.js`, or am I mistaken?

fisker:
I don't think we  need treat it as glob, `prettier [foo].js` we just check `[foo].js`.


So `prettier index*.js` only format file named exactly `index*.js`, won't match `index1.js` even only one file matched.

PR #9297, opened 2020-09-30
CLI: Only print formatted code when one filepath passed

fisker:
This is a quick implementation for #8832, I can't image why would someone what print multiple files output.
And we don't need support glob pattern.


src/cli/util.js
```diff
@@ -433,7 +437,31 @@ function formatFiles(context) {
...
-  for (const pathOrError of expandPatterns(context)) {
+  const willWriteOutput =
+    !context.argv.check &&
+    !context.argv["list-different"] &&
+    !context.argv["debug-check"];
+  let files;
+  if (willWriteOutput) {
+    if (context.filePatterns.length !== 1) {
+      context.logger.error(
+        `Expect only one filepath when format to stdout, ${context.filePatterns.length} given.`
```
lydell, line 448: If you pass _one_ pattern (a glob) that matches multiple files, will this say “expected 1 filepath, 1 given”?

fisker: No, glob is not supported at all.

fisker:
Any objection? @prettier/core

PR #9303, opened 2020-10-01
test Flow with babel-flow parser

brodybits:
I noticed that PR #8508 removed some testing of Flow with the `babel-flow` parser.

This PR adds testing with `babel-flow` to almost all tests in `tests/flow-repo` and `tests/flow`. The one exception is `tests/flow-repo/generics`.

The test from `tests/flow-repo/union_new/issue-815.js` is moved into its own subdirectory since it is the only one that gets a different snapshot for `babel-flow` vs `flow` parser.


fisker:
> The test from tests/flow-repo/union_new/issue-815.js is moved into its own subdirector

We can't move files in `flow-repo`, you can copy it in to `flow` and run test from there.

fisker:
superseded by https://github.com/prettier/prettier/pull/14245

PR #9304, opened 2020-10-01
test Angular with numbered object keys

brodybits:
Coming from some code changed in PR #9000, I could not see that the Angular parser was tested with numbered object keys.

I have to say that some of the actual formatting behavior does not appear to be super-consistent.


fisker:
You can test this parsers https://github.com/prettier/prettier/blob/7c5a7d3c747c73b19f9f44757a7a75e9169afce4/src/language-js/index.js#L82 directly, example https://github.com/prettier/prettier/tree/master/tests/angular/interpolation

brodybits:
I did not understand until now that the `angular` parser is really implemented by `src/language-html/parser-html.js`, using internal `__ng_*` parsers in `src/language-js` with `astFormat: "estree"` ... printed by `src/language-js/printer-estree.js`. Makes sense. I may need some more time to understand this better and do the testing on all internal `__ng_*` parsers.

fisker:
Just put some in https://github.com/prettier/prettier/tree/main/tests/angular/interpolation or maybe need some other `ng-*` parser, I'm not really sure

PR #9329, opened 2020-10-06
Style: no function reference in iterator methods

fisker:
Fix errors caught by [`unicorn/no-fn-reference-in-iterator`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-fn-reference-in-iterator.md), unfortunately it can't distinguish `FastPath#map` and `Array#map`, so we can't enable this rule.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -64,7 +64,6 @@ rules:
     - exceptRange: true
   unicorn/better-regex: error
   unicorn/new-for-builtins: error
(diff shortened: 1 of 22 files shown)


brodybits:
Hi this seems to make the code a little harder to read and seems to add a little overhead in some cases.

While this change may be safer against some utility function updates, I would think the unit testing should pick up on this kind of potential breakage as well.

I should have made this comment in the discussion in PR #9298.

Issue #2277, opened 2017-06-26
[Feature Request] Alpha sort imports


AaronVasquez:
I miss this feature from gofmt

vjeux:
JavaScript imports can have side effects so it's not safe to do that unfortunately. Also, we've taken the stance with prettier not to do those kind of transforms that change the ast.

We recommend that those would be implemented as eslint rules with autofix.

Thanks for asking :)

steida:
I believe the order of imports should be enforced. Fail fast fail cheap. Otherwise, it can happen the developer looking for the "right" order ends up in the cycle. We can add some flag for that. Order for imports is important for a quick scan (or OCD whatever). Please reconsider it. Thank you.

steida:
@vjeux Or via plugin?

vjeux:
@steida: at Facebook we're doing it via a eslint plugin with autofix. I would recommend going that route instead of inside of prettier.

steida:
During dev or after build? Not sure which workflow I should choose for Next.js. Thank you.

vjeux:
We show eslint warnings and autofix within Nuclide and before you send the equivalent of a PR. So it's both.

steida:
I learned how to transform Eslint errors to warnings for dev hour ago :-)
https://github.com/este/este/commit/8865ee80338531ff5bae7cccbffc3f600855dbd6

Thank you again, you are saving souls.

alexeagle:
For what it's worth, clang-format does alpha-sort imports in TypeScript/JavaScript code. We haven't had a problem with it at Google.

suchipi:
If you have modules that modify global state, or circular dependencies, then rearranging imports is not safe. For most cases, it is, but since there are some cases where it is not, prettier is going to err on the side of caution so that we don't break user's code.

lydell:
I wish something like [prettier-transform](https://github.com/azz/prettier-transform) will take off, so people can pick and choose such transforms (100% safe or not). I also want import sorting, but haven't find a really easy to use solution yet.

alexeagle:
I totally understand why prettier doesn't do this - if you start introducing unsafe options then you blur the line with linting, and pollute the mental model of "always accept what the formatter does".

TSLint has a fixer for this too. https://palantir.github.io/tslint/rules/ordered-imports/
If I find time, I'll try switching the main Angular repository to prettier and use that tslint check with `--fix` to fill in this bit.

PR #9417, opened 2020-10-15
Removed experimental plugins

ntotten:
Removing plugins that abuse the plugin api.


lipis (approved):

thorn0:
Somebody has to check other plugins on this page too. E.g. `prettier-plugin-package` is very similar to `prettier-plugin-organize-imports` in how it monkeypatches a built-in parser to reorder things.

hosseinmd:
@thorn0 In this commit [Remove more plugins that use the API in a non-intended way](https://github.com/prettier/prettier/pull/9417/commits/75f5587a9be9871a9f5d5e8d0d9e5f86bbe7ba5f)

I want to know what part of that is non-intended?

simonhaenisch:
lol just found out my plugin got removed from the website a long time ago for some interesting reasons...

- I changed my plugin to be compatible with other plugins that also set `preprocess` in March 2020 already
- Not sure why you'd expose a `preprocess` hook in the plugin API and then call it "abuse" and "unintended" when someone uses it to do some preprocessing with it 🤷🏻‍♂️
- You talk about monkey-patching like it's a bad thing but the whole purpose of Prettier's plugin API is so that Prettier can be [monkey-patched](https://en.wikipedia.org/wiki/Monkey_patch) 🤔 

I mean it's kind of your website so you can put on it whatever you want but it's also more like it's the Prettier community's website so it's kind a lame that one person can just remove whatever from it.

I still think a list of non-official community plugins on the website would be useful for people.

thorn0:
@simonhaenisch The point is that the plugin API is intended for adding support for more languages to Prettier, not to change how the existing formatters work. I have to admit I use your plugin myself and find it practical, but it really should've been implemented as an ESLint rule, not a Prettier plugin. Simply because Prettier is a formatter, not a linter. We don't want to encourage creation of new linter-like plugins by promoting such a plugin on the website. Sorry.

simonhaenisch:
I've used existing implementations for sorting imports via ESLint auto-fix and it also works well, but IMO it's arguable whether this is a linter or a formatter's responsibility, it can fit both. That's the reason for this plugin's existence... I don't always use ESLint but I always use Prettier.

Anyway, thanks for the clarification, makes sense.

thorn0:
Reordering things and especially removing unused things is much closer to ESLint's "auto-fixable" rules than to formatting. Prettier never does that kind of things. See also #2277, https://prettier.io/docs/en/rationale.html#what-prettier-is-_not_-concerned-about

The current plugin API is for adding new parsers and printers to Prettier, not for changing the existing ones. Plugins that change existing parts of Prettier are relying on implementation details. This is not a proper usage of the public API.

JounQin:
@thorn0 

I just notice this PR, this is very interesting.

> The current plugin API is for adding new parsers and printers to Prettier, not for changing the existing ones.

The plugin API is unlimited, or you should forbid current monkey-patched behavior for plugins.

> What Prettier is not concerned about

This is definitely fine for `prettier` core itself, but why it should enforce the 3rd-party plugins? I don't get it.

---

For my own use case, [`prettier-plugin-pkg`](https://github.com/un-ts/prettier/tree/master/packages/pkg) relies on `pretter`'s built-in parser, which `ESLint` does not have by default (there does have a [`jsonc-eslint-parser`](https://github.com/ota-meshi/jsonc-eslint-parser) from community now), that says using `prettier`'s `json` parser is much easier for developers than adding `eslint` + `jsonc-eslint-parser` + [`eslint-plugin-jsonc`](https://github.com/ota-meshi/eslint-plugin-jsonc) + another custom rule.

These plugins are added into document previously, but you suddenly removed them without any discussion, this seems incorrect to me at least. Whether listing here is not a big deal to me, but the statement

> This is not a proper usage of the public API.

is absolutely wrong for me personally.

The public API is just *`public`*.

thorn0:
Monkey patching and the inner workings of the built-in parsers are not public API. If you do it, keep in mind that it's not supported and can break in any version

JounQin:
> Monkey patching and the inner workings of the built-in parsers are not public API. If you do it, keep in mind that it's not supported and can break in any version

It's still the plugin authors' choice. Monkey patching is very commonly used in JavaScript, and it doesn't need to be documented as public API, it **is** public already.

thorn0:
Monkey patching means hacking on undocumented implementation details. A very opposite thing to what "public API" normally means. Anybody is free to hack things however they like, but it doesn't mean Prettier's site should promote these unsupported practices. If we want to let plugins augment the built-in parsers and printers, a proper API should be added for that.

JounQin:
I understand your concern and totally fine with removing them from the official document, but I still think it should not be discouraged to add features for their own or community usage based on current API by your statement.

We, developers, are just making incredible prettier plugins, it should not be considered as abuse.

PR #9432, opened 2020-10-16
Support Flow Enums with unknown members

gkz:
This PR adds support for Flow Enums with unknown members.

E.g. 

```
enum E {
  A,
  B,
  ...
}
```


fisker (approved):
LGTM

sosukesuzuki (approved):

evilebottnawi (approved):

tests/flow/enums-unknown-members/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,47 @@
...
+                                                                                | printWidth
+=====================================input======================================
+// @flow
+
+enum E {...}
+
+=====================================output=====================================
+// @flow
+
+enum E {...}
```
thorn0, line 44: The `bracketSpacing` option (`true` by default) should be respected here. Cf. `type E = { ... };`

thorn0 (changes requested):

tests/flow/enums-unknown-members/__snapshots__/jsfmt.spec.js.snap
gkz: Good point. Actually, if we have the `...` the enum shouldn't be all on one line. I'll update the PR.

thorn0 (approved):

PR #9464, opened 2020-10-21
Enable `unicorn/prefer-optional-catch-binding`

fisker:


evilebottnawi (approved):
Good idea

fisker:
The build script is failing, I'll try to fix it later.

fisker:
#10249 will unblock this.

Issue #907, opened 2017-03-06
Omitted parentheses in ForStatement with `in` in initializer


bakkot:
The grouping parentheses are necessary in all of the below, but prettier removes them:

```js
for (let a = (b in c);;);
```

```js
for (a && (b in c);;);
```

```js
for (a => (b in c);;);
```

```js
function* g() {
  for (yield (a in b);;);
}
```

```js
async function f() {
  for (await (a in b);;);
}
```

vjeux:
Thanks! I think that we should just crawl up the tree and if `in` is inside the first element of a for, then parenthesis is.

bakkot:
Always parenthesizing would be overkill. You don't need to bother with the parentheses in a lot of cases, e.g. if the `in` is within an object, array, template, or function literal (other than the body of an expression-bodied arrow function literal), or a parameter to a call, or a computed member access. Or if the initializer is going to be parenthesized anyway, e.g. in an assignment expression after #862, or in e.g. `!(a in b)`.

vjeux:
We found this edge case during fuzz testing and it doesn't appear in any of our millions of lines of js at Facebook. So I think it is fine having a bit of over-parenthesing for such a rare edge case just to make sure that it doesn't output something invalid :)

PR #9485, opened 2020-10-23
Refactor `needs-parens.js`

fisker:
Similar to @brodybits did in #8878, remove unnecessary conditions, move name check to the first, to avoid access node properties.

### DIFF
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -211,18 +211,14 @@ function needsParens(path, options) {
       const isLeftOfAForStatement = (node) => {
         let i = 0;
         while (node) {
-          const parent = path.getParentNode(i++);
-          if (!parent) {
+          const ancestor = path.getParentNode(i++);
+          if (!ancestor) {
             return false;
           }
-          if (
-            parent.type === "ForStatement" &&
-            // eslint-disable-next-line prettier-internal-rules/better-parent-property-check-in-needs-parens
-            parent.init === node
-          ) {
+          if (ancestor.type === "ForStatement" && ancestor.init === node) {
             return true;
           }
-          node = parent;
+          node = ancestor;
         }
         return false;
       };
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -208,21 +208,9 @@ function needsParens(path, options) {
         return true;
       }
 
-      const isLeftOfAForStatement = (node) => {
-        let i = 0;
-        while (node) {
-          const ancestor = path.getParentNode(i++);
-          if (!ancestor) {
-            return false;
-          }
-          if (ancestor.type === "ForStatement" && ancestor.init === node) {
-            return true;
-          }
-          node = ancestor;
-        }
-        return false;
-      };
-      if (node.operator === "in" && isLeftOfAForStatement(node)) {
+      // We add parentheses to all `a in b` inside `ForStatement` initializer
+      // https://github.com/prettier/prettier/issues/907#issuecomment-284304321
+      if (node.operator === "in" && isPathInForStatementInitializer(path)) {
         return true;
       }
       if (node.operator === "|>" && node.extra && node.extra.parenthesized) {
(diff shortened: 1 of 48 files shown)


src/language-js/needs-parens.js
```diff
@@ -531,7 +531,13 @@ function needsParens(path, options) {
         (grandParent.init === parent || grandParent.update === parent)
       ) {
         return false;
-      } else if (name === "value" && parent.type === "Property") {
+      } else if (
+        name === "value" &&
+        parent.type === "Property" &&
+        grandParent &&
+        grandParent.type === "ObjectPattern" &&
+        grandParent.properties.includes(parent)
```
thorn0, line 539: BTW, `ObjectPattern` doesn't have any other child nodes, only `properties`, so this `includes` seems unneeded.

fisker: I know that, just for safety.

fisker: By the way, do you think adopt `esquery` is a good idea?

So we can do something like

```js
isNodeMatch(node, 'ObjectPattern > Property.properties > AssignmentExpression.value');
```
I'm a little worry about performance

thorn0: Looks interesting, but I have doubts about the performance too.

fisker: Another off topic one, I have been thinking about use esquery selector in `embeddedLanguageFormatting` option few weeks ago, :smile:

thorn0 (approved):

fisker:
Wrote a eslint rule for it, and found more :smile:

evilebottnawi (approved):

src/language-js/needs-parens.js
```diff
@@ -488,35 +495,32 @@ function needsParens(path, options) {
...
+        name === "key" &&
         parent.type === "ClassProperty" &&
-        parent.key === node &&
         parent.computed
       ) {
         return false;
-      } else if (
-        parent.type === "TSPropertySignature" &&
-        parent.name === node
-      ) {
```
thorn0, line 510: This seems to be an artifact left from older versions of `typescript-eslint`. `TSPropertySignature` doesn't have the `name` property anymore. See https://github.com/eslint/typescript-eslint-parser/commit/5421ceae59a10212562eedbafbd4bea116ca2b6e

src/language-js/needs-parens.js
```diff
@@ -215,7 +215,11 @@ function needsParens(path, options) {
           if (!parent) {
             return false;
           }
-          if (parent.type === "ForStatement" && parent.init === node) {
+          if (
+            parent.type === "ForStatement" &&
+            // eslint-disable-next-line prettier-internal-rules/better-parent-property-check-in-needs-parens
```
thorn0, line 220: How about renaming the `parent` variable to `ancestor` to get rid of `eslint-disable`?

src/language-js/needs-parens.js
fisker: Nice catch, I thought tests not covered.

src/language-js/needs-parens.js
fisker: This function is not right

**Prettier 2.1.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAZgCnRCACVKPAIwEMAnASgG48B6OvAHgFoW8p8AHChAZwA6UITnJ5MkKHxh4A+ngC84zJUUA+PMCF4CWHPkIkKNeo1bse5fkIC+lFbVqVgNoSAA0ICFxipofZFAKcggAdwAFXilkEFIAG1DSAE8Az2JyUjAAazgYAGUeMEIAc2QYcgBXOE84AFtiOAATRqaAGVIoYorSYrgAMQhyWtIYX06Y0gqYCA8QAAsYWriAdTnUeD5CuDz+ddQAN3WkmLA+VLQpOHIYcIzi4eR0eL5qkAArPgAPACEM7Ny8qRanBWoQ4I9nq8Pp88iU4nAAIoVCDwCFxF6eSwvcgxMgNOKzLjkQgwZaoRowObIAAcAAZMSEXssMlwYkS4Nj9uDPABHZHwW7eAIoUh8FhQOBNJqzKx81BWW49B5IJ7o14vWqoMqVdVwxH88EqyGeGCkYhkilUpAAJhNGVQcRKAGEILVlSAOQBWWYVF4AFTNwtVGJA+yqAEkoC1YHkwMSfABBKN5GBJeFol42GxAA)
```sh
--parser babel
```

**Input:**
```jsx
if(foo in bar); // <-- no parens

for (const _ = (() => {
  if(foo in bar); // <-- parens
})(); ; ){}

```

**Output:**
```jsx
if (foo in bar); // <-- no parens

for (
  const _ = (() => {
    if ((foo in bar)); // <-- parens
  })();
  ;

) {}

```

fisker: https://github.com/prettier/prettier/issues/907#issuecomment-284304321

sosukesuzuki (approved):

PR #9498, opened 2020-10-26
Refactor `needs-parens.js` (part 2)

fisker:
Follow up #9485

### DIFF
--- a/src/language-js/needs-parens.js
+++ b/src/language-js/needs-parens.js
@@ -1,8 +1,5 @@
 "use strict";
 
-/** @type {import("assert")} */
-const assert = require("assert");
-
 const {
   getFunctionParameters,
   getLeftSidePathName,
@@ -285,7 +282,6 @@ function needsParens(path, options) {
           }
 
           if (name === "right" && parentPrecedence === precedence) {
-            assert.strictEqual(parent.right, node);
             return true;
           }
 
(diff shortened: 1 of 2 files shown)


src/language-js/needs-parens.js
```diff
@@ -24,13 +21,6 @@ function needsParens(path, options) {
   const name = path.getName();
   const node = path.getNode();
 
-  // If the value of this path is some child of a Node and not a Node
-  // itself, then it doesn't need parentheses. Only Node objects (in
-  // fact, only Expression nodes) need parentheses.
-  if (path.getValue() !== node) {
-    return false;
-  }
```
fisker, line 32: @thorn0 Do you think this part is useless? `getNode()` always the same as `getValue()`, `getValue()` seems designed to be faster than `getNode()`? 

This part [exists from 2016](https://github.com/azz/prettier/commit/9b4535e9f8a7040f2fbc65b22c9713632f3731be#diff-e133c8cd5b77cadc0f3918edd7a7c071f8fad2e511c7291c3f25f350d8f41ca6R188)

thorn0: From the code, I can see that arrays are considered "values" but not "nodes". Arrays aren't passed to `needsParens`, so this `if` indeed can be removed.

fisker: Oh, I read the `FastPath#getNode` again, I was wrong about `getNode()` and `getValue()`, I thought they are the same, `getNode()` skip "array". 

Thanks.

thorn0 (approved):

evilebottnawi (approved):

PR #9533, opened 2020-10-29
Use `-1` instead of `false` when index out of bounds

fisker:
Idea from https://github.com/prettier/prettier/pull/8759/files#r465118376

Align this with built-in methods like `String#indexOf()` `Array#findIndex()` etc.
To prevent breaking changes, I add several simple wrapper to make public api behave the old way.

cc @brodybits

### DIFF
--- a/src/common/util-shared.js
+++ b/src/common/util-shared.js
@@ -26,28 +26,48 @@ const {
   addTrailingComment,
 } = require("./util");
 
+const returnFalseIfNegativeOne = (fn) => (...args) => {
+  const result = fn(...args);
+  return result === -1 ? false : result;
+};
+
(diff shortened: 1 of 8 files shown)


thorn0:
`false` is strictly better from the type-checking point of view. E.g. a type-checker won't allow to pass a value of type `number | false` to `.slice(...)`.

fisker:
I agree it's better for the type-checking, but I think most people expect `getXXXIndex()` to return `-1`, functions in `lodash` do this too.

thorn0:
I don't see a point in sticking to this convention inherited from Java. It's enough to document the return type for people to know how to use these functions.

PR #9534, opened 2020-10-29
Don't remove empty values in massageAST

fisker:
```js
> require('.').__debug.parse('[,1]', {parser: 'babel'}, true).ast.program.body[0].expression.elements
[ null, { type: 'NumericLiteral', value: 1 } ]
> require('prettier').__debug.parse('[,1]', {parser: 'babel'}, true).ast.program.body[0].expression.elements
[ { type: 'NumericLiteral', value: 1 } ]
```

We should make sure array elements didn't change.

### DIFF
--- a/src/main/massage-ast.js
+++ b/src/main/massage-ast.js
@@ -2,7 +2,7 @@
 
 function massageAST(ast, options, parent) {
   if (Array.isArray(ast)) {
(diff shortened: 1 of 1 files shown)


fisker:
Not a good one, will find another solution.

PR #9545, opened 2020-10-30
fix: add jsdoc plugin for prettier source

hosseinmd:
I added prettier-plugin-jsdoc and resolved some jsdoc issues.

For some reason, some docs shouldn't format so for ignoring them I added them to commentLine.


hosseinmd:
@fisker I have trouble with tests. I don't know what exactly happened.

fisker:
The tests seems running with the plugin, but I'm not sure how to fix it.

Try add `{"plugins": []}` to `.prettierrc`?

hosseinmd:
> The tests seems running with the plugin, but I'm not sure how to fix it.
> 
> Try add `{"plugins": []}` to `.prettierrc`?

That does not work. This plugin doesn't need to be added to the plugins list (I think this is a bug). When installed in a project it is in prettier plugins.

hosseinmd:
But `"plugins": []` doesn't work. Needed another way to prevent prettier search for it

PR #7948, opened 2020-04-03
Ugly formatting of nested conditional types [TypeScript]

mmiszy:
Fixes https://github.com/prettier/prettier/issues/7940

My intention was to only affect nested conditional types in TypeScript and not to introduce any changes to how ternary expressions are formatted. This allows for a smoother transition, however, it might yield seemingly inconsistent results (see [this example](https://deploy-preview-7948--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAKpuAOQEMBbOAHlwD5sBeAHSmz2zgA94oATAZ214wATgEsoAc2wB+bAxCDREudiRMWuNpwR9sUAK6kARnCHTZIfUZPLVzVhy47DECABs4xZjLnO3HqDZq9lo8-Ho8cABmYnDcZnLh3FEx3IF2Gg7a-ABi4WAwItDxIJF5BdBp6pqO-ACCQkLEaBSeaLTeIMQNTZXBNbp6rq7F+kO9GSE6vGhGbsXTs669chCGAFZw+XIA3ExMkFCC2AAa9EEadHQCwmKSHQq34-RXlsamHa-WICrnz9i+7k8xQB-iel2wiWSUFixUh0WhqW+tiq4NyUHyhS85lK6PKASRv3B9UaaGKXRJYJeg2GH2plIEM2cNPMCyZyxAqw2WxA2xAABoORg8bxkKByRAAO4ABS6CBFKGIrglTRFAsMjTAAGs4DAAMoYYhgW7IYR6OACgAWMFIrgA6haRPBeAawHBdXLHSIAG6OtDIcC8VUgMS8EwwKWNcSkYjISKK0MCta8dgAIQ12r1ZDgABkYrH4+aQEn2Lrbu4AIp6CDwfOuBMgA1CUNCf3oLC8MCiIX8huKGC2kTcGAW5AADgADAKMEIIKHbY0MP7p3Bm164D2AI5V+ARiAYeWdXgAWmhsViPaEcC3IkvEeIUZjSDjdcLodIIhNQjNAt4ZbglerdcnwLAUYGIQwByHEckAAJlAxoRFcW4AGEIFIaN-RXABWHs9FDXBwPlZ96y9M0AEkIlgXVOxEIVah4XV0HcWtQwAX1YoA)).

- [x] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
[...]

Issue #4198, opened 2018-03-23
TypeScript: nested union types + useTabs


thorn0:
**Prettier 1.11.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgILro7AA6UOOADgE4CWAbuvDlBNQLbqq0BecAynGr0hAUWrU2ACmroA7uMnUkZChUww6UAOY4APqrXB0SDVu16oAV3YAjIXuOnNtHQF89AJTlDMAEwAzfwD6AA8wemC-OB9qWLk3fTVklNS0tUMKHQA6AEkACRgYSgAFSXZaTDgAYS5UW3QwAGs8am0AHky1PRwzV20uih7gHCElJF6XHSTFNlyoALZOGFpoHDdBpJGAWThsdG04Cb6dAG4cUXC4ShXoABUAT0ojyfN1roA+QwBKYi7IKCYCCoODZVAQbQyeSzajfU6GDZQNwgAA0IAgN1WgOQoHQEggchKeIQmGQIHQ9AgtD8qPJGmQAS4VTRtlkzTgMH4lEa-WQmiscDRfggYAZTMFIFcVWoMDKB04YtQzJAACtMKEAEJspoc-jodhwAAyrjgiuVECsxUtACYzRLudRpWSYE89mA6DdaTRXDAAOrUmAAC2QAA4AAxomgQKq+2SUMk0PZCES06hwACOVloabl2gVSEZSolVQqfOoArRmH6IIAilYIPA7WiYOhbP6-EHkAAWZuyWjcHTVCDsfMgVhQU1oqxVO6t0lIflwNxuIA)
```sh
--parser typescript
--tab-width 4
--use-tabs
```

**Input:**
```tsx
class Aaa {
  private normalizeServerError(rawError:
    string |
    {a:string|number|{a:string}|Rwersdfsdfvxcvsdfdewerwerw} |                            
    ng.IHttpPromiseCallbackArg<
      | string
      | { error: string | ErrorInformation }
      | { Message: string; ExceptionType: string }
    >
  ) {
    console.log(rawError);
  }
}
```

**Output:**
```tsx
class Aaa {
	private normalizeServerError(
		rawError:
			| string
			| {
					a:
						| string
						| number
						| { a: string }
						| Rwersdfsdfvxcvsdfdewerwerw;
			  }
			| ng.IHttpPromiseCallbackArg<
					| string
					| { error: string | ErrorInformation }
[...]

PR #9552, opened 2020-10-31
Use JSX style for all nested ternaries

### LINKED ISSUE #5814: Nested ternary formatting - add indents back
Edit: I will no longer be actively participating in this thread - my reasons are explained [in comment below](https://github.com/prettier/prettier/issues/5814#issuecomment-719825938).

Hello,

there are many - including me thinking that change to nested ternary is not best decision.

Eg threads https://github.com/prettier/prettier/issues/5476 https://github.com/prettier/prettier/issues/3018

I know there is some reasoning why nested ternary indent was removed, but it's much harder to read code now.

At least would be good to have option to remain nice formatting (from previous prettier version) - formatting of ternaries with indent and not under each other like:

```
condition
  ? a
  : b
  ? c
  : d
```

instead much easier to understand and read is:

```
condition
  ? a
  : b
    ? c
    : d
```

rattrayalex:
Fixes #5814

This is an alternative to https://github.com/prettier/prettier/pull/9559; **please upvote this PR or https://github.com/prettier/prettier/pull/9559, or a proposal in https://github.com/prettier/prettier/pull/9561**

I think this PR has better behavior since it scales better to long ternary formats, has an extremely clear visual difference between conditionals and consequents, and is easy to refactor. I've also had good experiences and heard good feedback on the JSX style when it is used in JSX.

Headline examples from the tests:

```jsx
const message =
  i % 3 === 0 && i % 5 === 0 ? (
    "fizzbuzz"
  ) : i % 3 === 0 ? (
    "fizz"
  ) : i % 5 === 0 ? (
    "buzz"
  ) : (
    String(i)
  );

const paymentMessage =
  state == "success" ? (
    "Payment completed successfully"
  ) : state == "processing" ? (
    "Payment processing"
  ) : state == "invalid_cvc" ? (
    "There was an issue with your CVC number"
  ) : state == "invalid_expiry" ? (
    "Expiry must be sometime in the past."
  ) : (
    "There was an issue with the payment.  Please contact support."
  );

const StorybookLoader = ({ match }) =>
  match.params.storyId === "button" ? (
    <ButtonStorybook />
  ) : match.params.storyId === "color" ? (
    <ColorBook />
  ) : match.params.storyId === "typography" ? (
    <TypographyBook />
  ) : match.params.storyId === "loading" ? (
    <LoaderStorybook />
  ) : match.params.storyId === "deal-list" ? (
    <DealListStory />
  ) : (
    <Message>
      <Title>{"Missing story book"}</Title>
      <Content>
        <BackButton />
      </Content>
    </Message>
  );
```

and a case I hope to add to the tests:

```jsx
children && !isEmptyChildren(children) ? (
  children
) : props.match ? (
  component ? (
    React.createElement(component, props)
  ) : render ? (
    render(props)
  ) : null
) : null;
```

Note that this PR uses indents for TS Conditional Types, eg:

```ts
type TypeName<T> = T extends string
  ? "string"
  : T extends number
    ? "number"
    : T extends boolean
      ? "boolean"
      : T extends undefined
        ? "undefined"
        : T extends Function
          ? "function"
          : "object";
```
but we could easily back that change out.

I may seek to change TS Conditional Type formatting in a follow-on PR, and am considering this **case-style** format:

```ts
type Animal<T> =
  T extends Quacker ? Duck :
  T extends SuperTall ? Giraffe :
  T extends SingleCelled ? (
    T extends Eukaryotic ? Amoeba :
    Bacteria
  ) :
  T extends Barker ? Dog :
  GenericAnimal;
```

### DIFF
--- a/src/language-js/print/ternary.js
+++ b/src/language-js/print/ternary.js
@@ -231,9 +231,7 @@ function printTernaryOperator(path, options, print, operatorOptions) {
         : "",
       line,
       ": ",
-      alternateNode.type === operatorOptions.conditionalNodeType
-        ? path.call(print, operatorOptions.alternateNodePropertyName)
-        : align(2, path.call(print, operatorOptions.alternateNodePropertyName)),
+      align(2, path.call(print, operatorOptions.alternateNodePropertyName)),
     ]);
     parts.push(
       parent.type !== operatorOptions.conditionalNodeType ||
--- a/src/language-js/print/ternary.js
+++ b/src/language-js/print/ternary.js
@@ -182,13 +182,13 @@ function printTernaryOperator(path, options, print, operatorOptions) {
     lastConditionalParent !== node;
 
   if (
-    // We now print nested ternaries the same way we print JSX ternaries.
-    isTernaryChain ||
-    (operatorOptions.shouldCheckJsx &&
-      (isJSXNode(node[operatorOptions.testNodePropertyNames[0]]) ||
-        isJSXNode(consequentNode) ||
-        isJSXNode(alternateNode) ||
-        conditionalExpressionChainContainsJSX(lastConditionalParent)))
+    operatorOptions.shouldCheckJsx &&
+    // We now print nested ternaries the same way we print JSX ternaries, except for TS Conditional types.
+    (isTernaryChain ||
+      isJSXNode(node[operatorOptions.testNodePropertyNames[0]]) ||
+      isJSXNode(consequentNode) ||
+      isJSXNode(alternateNode) ||
+      conditionalExpressionChainContainsJSX(lastConditionalParent))
   ) {
     jsxMode = true;
     forceNoIndent = true;
--- a/src/language-js/print/ternary.js
+++ b/src/language-js/print/ternary.js
@@ -176,12 +176,19 @@ function printTernaryOperator(path, options, print, operatorOptions) {
   const firstNonConditionalParent = currentParent || parent;
   const lastConditionalParent = previousParent;
 
+  const isTernaryChain =
+    alternateNode.type === operatorOptions.conditionalNodeType ||
+    consequentNode.type === operatorOptions.conditionalNodeType ||
+    lastConditionalParent !== node;
+
   if (
-    operatorOptions.shouldCheckJsx &&
-    (isJSXNode(node[operatorOptions.testNodePropertyNames[0]]) ||
-      isJSXNode(consequentNode) ||
-      isJSXNode(alternateNode) ||
-      conditionalExpressionChainContainsJSX(lastConditionalParent))
+    // We now print nested ternaries the same way we print JSX ternaries.
+    isTernaryChain ||
+    (operatorOptions.shouldCheckJsx &&
+      (isJSXNode(node[operatorOptions.testNodePropertyNames[0]]) ||
+        isJSXNode(consequentNode) ||
+        isJSXNode(alternateNode) ||
+        conditionalExpressionChainContainsJSX(lastConditionalParent)))
   ) {
     jsxMode = true;
     forceNoIndent = true;
--- a/src/language-js/print/ternary.js
+++ b/src/language-js/print/ternary.js
@@ -215,9 +215,15 @@ function printTernaryOperator(path, options, print, operatorOptions) {
 
     parts.push(
       " ? ",
+      consequentNode.type === operatorOptions.conditionalNodeType
+        ? ifBreak("", "(")
+        : "",
       isNil(consequentNode)
         ? path.call(print, operatorOptions.consequentNodePropertyName)
         : wrap(path.call(print, operatorOptions.consequentNodePropertyName)),
+      consequentNode.type === operatorOptions.conditionalNodeType
+        ? ifBreak("", ")")
+        : "",
       " : ",
       alternateNode.type === operatorOptions.conditionalNodeType ||
         isNil(alternateNode)
--- a/tests/js/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/js/comments/__snapshots__/jsfmt.spec.js.snap
@@ -3680,7 +3680,7 @@ function memberInAndOutWithCalls() {
 function excessiveEverything() {
   return (
     // Reason for stuff
-    a.b() * 3 + 4 ? (a\`hi\`, 1) ? 1 : 1 : 1
+    a.b() * 3 + 4 ? ((a\`hi\`, 1) ? 1 : 1) : 1
   )
 }
 
@@ -3942,7 +3942,7 @@ function memberInAndOutWithCalls() {
 function excessiveEverything() {
   return (
     // Reason for stuff
-    a.b() * 3 + 4 ? (a\`hi\`, 1) ? 1 : 1 : 1
+    a.b() * 3 + 4 ? ((a\`hi\`, 1) ? 1 : 1) : 1
   );
 }
 
--- a/tests/js/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/js/comments/__snapshots__/jsfmt.spec.js.snap
@@ -3680,7 +3680,7 @@ function memberInAndOutWithCalls() {
 function excessiveEverything() {
   return (
     // Reason for stuff
-    a.b() * 3 + 4 ? ((a\`hi\`, 1) ? 1 : 1) : 1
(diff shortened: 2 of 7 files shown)


thorn0:
Another reason to prefer this style is tabs. "Use tabs for indentation, spaces for alignment" sounds good in theory, but is confusing in practice. I avoid using tabs now because Prettier works best with spaces, specifically because of how it formats ternaries and union types (see #4198).

Also I totally agree that, if the main problem with the current formatting is not the indentation per se but the lack of a clear separation between conditions and results, then this style is a tidier way to achieve this separation than aggressive indentation.

DScheglov:
@rattrayalex , hey

```js
const message =
  i % 3 === 0 && i % 5 === 0 ? (
    "fizzbuzz"
  ) : i % 3 === 0 ? (
    "fizz"
  ) : i % 5 === 0 ? (
    "buzz"
  ) : (
    String(i)
  );
```

it is much more noisy then ts conditional type style:

```js
const message =
  i % 3 === 0 && i % 5 === 0 ? "fizzbuzz" :
  i % 3 === 0 ? "fizz" :
  i % 5 === 0 ? "buzz" :
  String(i);
```

Could "noise" be considered too?

askuzminov:
Original

```ts
const pickBgColor = (colorName: string) =>
  colorName === 'active'
    ? colorName === 'active'
      ? 'var(--blue2)'
      : colorName === 'inactive'
    : colorName === 'inactive'
    ? 'var(--blue1)'
    : colorName === 'success' || colorName === 'abc'
    ? colorName === 'abc'
      ? 'var(--blue2)'
      : 'var(--blue23)'
    : 'var(--red4)';

const message = (i: number) =>
  i % 3 === 0 && i % 5 === 0
    ? 'fizzbuzzfizzbuzzfizzbuzzfizzbuzz'
    : i % 3 === 0
    ? i % 3 === 0
      ? 'ffizzbuzzfizzbuzzfizzbuzzfizzbuzzizz'
      : i % 5 === 0
    : i % 5 === 0
    ? 'fizzbuzzfizzbuzzfizzbuzzfizzbuzz'
    : String(i);
```

current PR
```ts
const pickBgColor = (colorName: string) =>
  colorName === "active" ? (
    colorName === "active" ? (
      "var(--blue2)"
    ) : (
      colorName === "inactive"
    )
  ) : colorName === "inactive" ? (
    "var(--blue1)"
  ) : colorName === "success" || colorName === "abc" ? (
    colorName === "abc" ? (
      "var(--blue2)"
    ) : (
      "var(--blue23)"
    )
  ) : (
    "var(--red4)"
  );

const message = (i: number) =>
  i % 3 === 0 && i % 5 === 0 ? (
    "fizzbuzzfizzbuzzfizzbuzzfizzbuzz"
  ) : i % 3 === 0 ? (
    i % 3 === 0 ? (
      "ffizzbuzzfizzbuzzfizzbuzzfizzbuzzizz"
    ) : (
      i % 5 === 0
    )
  ) : i % 5 === 0 ? (
    "fizzbuzzfizzbuzzfizzbuzzfizzbuzz"
  ) : (
    String(i)
  );
```

The original is much better. The current PR is terrible.

rattrayalex:
Having examples is helpful, thank you – especially when they are more real-world / have sensible code.

I am now convinced that the "JSX style approach" is too simplistic to work well everywhere, so I will close this PR.

I am however still optimistic (possibilistic?) that a solution may exist which truly improves the readability of nested/chained ternaries everywhere.

@DScheglov , I am also starting to think that **case-style** might be a good starting point in general, eg:

```js
const result = 
  children && !isEmptyChildren(children) ? children : 
  props.match ? (
    component ? React.createElement(component, props) : 
    render ? render(props) : 
    null
  ) : 
  null;
```

I'm also interested in the **if-else-style** approach [outlined here](https://github.com/prettier/prettier/issues/737#issuecomment-370202601) by @JAForbes, eg:

```js
const result =
  children && !isEmptyChildren(children)
    ? children
  : props.match
    ? (component
        ? React.createElement(component, props)
      : render
        ? render(props)
        : null)
    : null;
```

I've opened a separate issue to muse about the possibilities at greater length: https://github.com/prettier/prettier/issues/9561

lydell:
@rattrayalex I draw the opposite conclusion from the provided examples (only one of which seems to be real-world code). In my opinion, the JSX style was easier to read in all cases. Longer, yes. More parens, yes. But not noiser than good ol’ if-else.

I think we should reopen and at least give it a chance.

rattrayalex:
@lydell fair, done!

Though I gotta say, I really like the case-style when things are simple, and it naturally turns right into ~JSX-style when they're not:

```jsx
const message =
  i % 3 === 0 && i % 5 === 0 ? "fizzbuzz" :
  i % 3 === 0 ? "fizz" :
  i % 5 === 0 ? "buzz" :
  String(i);

const message =
  i % 3 === 0 && i % 5 === 0 ? (
    // pretend this is big and complex
    "fizz" + "buzz" 
  ) :
  i % 3 === 0 ? (
    // this too
    "fizz" 
  ) :
  i % 5 === 0 ? (
    // such a long line wow
    "buzz" 
  ) : (
    // and this
    String(i)
  );
```

rattrayalex:
@lydell btw, a big part of my motivation in opening #9561 is handling TS Conditional types nicely, which IMO look really bad with JSX-style:

```ts
type TypeName<T> = 
  T extends string ? (
    "string"
  ) : T extends number ? (
    "number"
  ) : T extends boolean ? (
    "boolean"
  ) : T extends undefined ? (
    "undefined"
  ) : T extends Function ? (
    "function"
  ) : (
    "object"
  );

type Unpacked<T> =
  T extends (infer U)[] ? (
    U
  ) : T extends (...args: any[]) => infer U ? (
    SomeVeryLongNameOfSomeKind<U>
  ) : T extends Promise<infer U> ? (
    U
  ) : (
    T
  );
```

if members of the TS community disagree and find this the most readable option, I'd be quite happy to put this back on the table!

DScheglov:
@rattrayalex , I've posted real-world example in the Issue, see the comment:
https://github.com/prettier/prettier/issues/5814#issuecomment-719588547

It is absolutely real.

thorn0:
@rattrayalex WRT conditional types, see #7948 for some good code examples.

rattrayalex:
Just musing aloud, I wonder about an "all-or-nothing" approach to chained ternaries, where if none of the ternaries break, you get this:
```js
const pickBgColor = (colorName) =>
  colorName === 'active' ? 'var(--blue2)' :
  colorName === 'inactive' ? 'var(--blue1)' :
  colorName === 'success' ? 'var(--green4)' :
  'var(--red4)';
```
but if any of them are too long for one line, they all become JSX-style...

thorn0:
@rattrayalex I proposed a similar thing in #7948 for conditional types, but for normal (non-type) conditionals, it might be too diff-unfriendly to reformat an entire multiline expression upon changing only a small part of it.

rattrayalex:
Ah, yes, that's a very good point. Really changes how I think about this to be honest.

DScheglov:
> Ah, yes, that's a very good point. Really changes how I think about this to be honest.

Hey, @rattrayalex , any option we choose (except keeping as is) will be diff-unfriendly, ... 
It seems to be needed an setting.

And one more option -- it is difficult, but possible. Prettier analyzes the source code and defines "current" style and after that checks style-consistency.

rattrayalex:
Sorry, to be clear, the diff changes we are talking about are _while editing code_ once this is released, not "at release of this change". 

Constant large diffs while editing code, in places unrelated to the line you're editing, happen sometimes with prettier but are great to avoid whenever possible.

I appreciate the desire to be helpful, but adding new settings or having prettier "keep existing style" are antithetical to prettier's philosophy. As you can see from countless threads, prettier would rather make some (even many) people unhappy with how their code looks than adopt either approach. I am not a core maintainer; I don't make these rules, I just have to play by them.

thorn0:
Adding a setting also means admitting that all the proposals are equally good/bad, so the whole problem is nothing more but bike shedding, in which case we can as well simply do nothing. I hope it's not the case though.

rattrayalex:
I no longer think this is a good idea, should have closed this a while ago. 

I now believe that postfixed ternaries are the only scalable, readable, consistent alternative, proposed here as "Option B2": https://github.com/prettier/prettier/issues/9561#issuecomment-720262830

PR #9560, opened 2020-11-01
Fix a condition

fisker:
I guess this mean to test `.children.length`, but it's unnecessary too, `Array#every()` always returns `true` on empty array.


sosukesuzuki (commented):
Just question, can `node.children` be undefined?

fisker:
No, https://github.com/glimmerjs/glimmer-vm/blob/70adc10b3b8c14c2957c06f82a53ff15b6aa8172/packages/%40glimmer/syntax/lib/types/nodes.ts#L151 otherwise the old one would failed.

sosukesuzuki (approved):

PR #9569, opened 2020-11-02
Use yarn 2

fisker:


package.json
```diff
@@ -108,7 +108,7 @@
     "eslint-formatter-friendly": "7.0.0",
     "eslint-plugin-import": "2.22.1",
     "eslint-plugin-jest": "24.1.0",
-    "eslint-plugin-prettier-internal-rules": "file:scripts/tools/eslint-plugin-prettier-internal-rules",
+    "eslint-plugin-prettier-internal-rules": "file:scripts/tools/eslint-plugin-prettier-internal-rules/eslint-plugin-prettier-internal-rules-1.0.0.tgz",
```
fisker, line 111: Somehow, yarn calculate different checksum on windows and other os.

package.json
```diff
@@ -23,14 +23,15 @@
...
+    "@fisker/parse-srcset": "1.0.2",
     "@glimmer/syntax": "0.62.4",
     "@iarna/toml": "2.2.5",
     "@typescript-eslint/typescript-estree": "3.10.1",
     "angular-estree-parser": "2.2.1",
     "angular-html-parser": "1.7.1",
     "camelcase": "6.2.0",
     "chalk": "4.1.0",
-    "ci-info": "watson/ci-info#f43f6a1cefff47fb361c88cf4b943fdbcaafe540",
+    "ci-info": "2.0.0",
```
fisker, line 34: Don't know why, but there two can't install in netlify

package.json
fisker: Use `link:` instead of `file:` resolved this problem.

thorn0:
What are the benefits of doing this change?

fisker:
Not really, but yarn suggest to upgrade https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern

fisker:
Close in favor of #12599

PR #9577, opened 2020-11-03
Refactor: Use `p` function to simplify node property print

fisker:
I was going to use `printPathProperty`, but I think `p` is good too, we use `n` for `node`.


fisker:
Close as I have a better idea

PR #9591, opened 2020-11-04
Enable `directly-loc-start-end` in `language-css`

fisker:

### DIFF
--- a/src/language-css/utils.js
+++ b/src/language-css/utils.js
@@ -461,17 +461,7 @@ function isAtWordPlaceholderNode(node) {
   );
 }
 
-function locStart(node) {
-  return node.source.startOffset;
-}
-
-function locEnd(node) {
-  return node.source.endOffset;
-}
-
 module.exports = {
-  locStart,
-  locEnd,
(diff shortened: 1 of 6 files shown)


src/language-css/utils.js
```diff
@@ -461,7 +461,17 @@ function isAtWordPlaceholderNode(node) {
   );
 }
 
+function locStart(node) {
+  return node.source.startOffset;
```
thorn0, line 465: Why not put these to `language-css/loc.js`?

fisker: I was worried about Circular dependency in future.

fisker: In `language-js`, `isBlockComment` function was in `comment.js`,when we  start use functions from `utils.js` in `comment.js`, I have to move it to `utils.js`.

thorn0: On the other hand, files like `utils`, `helpers`, etc. are an antipattern. Usually it's better to move functions to other files with more meaningful names.

thorn0 (approved):

PR #9592, opened 2020-11-04
Add parser check to flow-comment related checks

fisker:
Make related part little safer for other parsers.

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -76,6 +76,7 @@ const {
   hasTrailingComment,
   hasTrailingLineComment,
   identity,
+  isFlowLanguageParser,
   isBinaryish,
   isCallOrOptionalCallExpression,
   isEmptyJSXElement,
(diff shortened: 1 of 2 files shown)


thorn0:
This code path is intended only for `--parser flow`. Babel, the way it's configured now, doesn't parse Flow comments into AST nodes.

fisker:
Are you sure about this? Let me try.

fisker:
No, not only `flow`,

https://github.com/prettier/prettier/blob/357196cf3c0ae248a135ea1268219c4744e16dbb/src/language-js/printer-estree.js#L1008-L1010

This line definitely for `babel`, because with `--parser=flow`, there will be no comment there

thorn0:
I said nothing about that code path. I mean the one that this PR modifies.

fisker:
I added it to it too.

thorn0:
Sorry, overlooked that. I meant the `"case "TypeParameterDeclaration": case "TypeParameterInstantiation"` part. It's intended for `flow` only whereas the other one is for the other parsers, not only for Babel as the original purpose of Flow comments is to keep code JS-compatible:

> Flow supports a comment-based syntax, which makes it possible to use Flow without having to compile your files.

-- https://flow.org/en/docs/types/comments/

thorn0 (approved):

evilebottnawi (approved):

PR #9642, opened 2020-11-11
Add `.values()` and `.entries()` to `FastPath`

fisker:

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -14,6 +14,7 @@ settings:
   import/internal-regex: ^linguist-languages/
 rules:
   prettier-internal-rules/require-json-extensions: error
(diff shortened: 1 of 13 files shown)


fisker:
I know this is very old, but I still hope `for..of` can work. Since we are adding `.index` to `AstPath` in #13448, should we support usage like this?

```js
for (const {index, node} of path.iter("children")) {}
```

//cc @thorn0

thorn0:
Won't it be too many ways to do the same thing? Also the performance needs to be checked.

fisker:
My idea is to replace `.each()`.

thorn0:
I'll need to address https://github.com/prettier/prettier/pull/13224#issuecomment-1228161394 to benchmark this properly.

Issue #8055, opened 2020-04-14
RFC: don't preserve line breaks in object literals if all the properties are shorthand


thorn0:
Let's format
```
const name = {
  firstName,
  lastName,
};
```
into
```
const name = { firstName, lastName };
```

Thoughts? Counterexamples?

evilebottnawi:
:+1: on it

fisker:
Does

```js
const name = {
  ...defaultNames,
  firstName,
  lastName,
};
```

```js
const {
  firstName,
  lastName,
} = name;
```

and `import/export as` the same?

thorn0:
As for spread properties, I think yes, they're in the same basket if the operand is an identifier.

Not sure about destructuring. Let's not discuss it in this issues.

As for imports/exports. Does Prettier preserve line breaks there?

fisker:
Sorry, I thought it's discuss about object-like , I think only object preserve line breaks.

thorn0:
This is only about object literals and about preserving fewer line breaks. 🔥

fisker:
```js
const foo = {...names, firstName, ...location, city};
```
This might harder to read than
```js
const foo = {
  ...names, 
  firstName, 
  ...location, 
  city
};
```

So, maybe only the first/last spread treat as  identifier.

thorn0:
It's okay for arrays:
```js
const foo = [...names, firstName, ...location, city];
```
so should be okay for objects too.

fisker:
What if they are grouped

```js
const foo = {
  ...names, 
  firstName, 

  ...location, 
  county,
  city,

  ...rest
};
```

VS

```js
const foo = {...names, firstName, ...location, county, city, ...rest};
```

[...]

Issue #7927, opened 2020-04-01
Inconsistent formatting with await, destructuring, and object parameter in function call


cprecioso:
**Prettier 2.0.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAExMCMIAmAnpgL6YC8mAhgO7UCWWA5hDABQA6IAFjDAAc0SAPQjaEgHQAnBgDcAtpKgAbEdTAwArnBVwAtAygAzCNIXUYDAxDlxp+yNOjU5DaVrTcANNkxWFOAgtGCRMAFYABmiyAEpOKBBvEAgBK3RkUGppZ1oABWyENGQQahV6ImLk3GkNAGs4GABlAQ0jZmQYDzhkvgUVAHUeJjg0VrA4JqKmeSYiErA0KpAjNHsYPNrmC2RjMrXkgCs0AA8AIVqwBubqQIAZIzhd-Z6QY5Om9r0ARS02J6QexUBxArWka2kJVw1FwuiSoNksAGDAIMB4yAAHJFkgJnGsBrUBCVcaN7HZ4QBHP7wTapYooahofRQOBwAhs+HSOBU9xwTbUbbUZ7A15rBQMTrdZJoL5wX7-YUgmAw5Go9FIABMyS6jBU7QAwhAFDsUKNwvDPHAACow+lAkFyHQASSg7NgTTAsjSAEFXU0YEQ9Iq4KRSEA)
```sh
--parser babel
```

**Input:**
```jsx
const { body } = await got("https://www.rivm.nl/actuele-informatie-over-coronavirus", { timeout: 5000 })

```

**Output:**
```jsx
const {
  body,
} = await got("https://www.rivm.nl/actuele-informatie-over-coronavirus", {
  timeout: 5000,
});

```

**Second Output:**
```jsx
const { body } = await got(
  "https://www.rivm.nl/actuele-informatie-over-coronavirus",
  {
    timeout: 5000,
  }
);

```

**Expected behavior:**
It should correctly format the first time around.

thorn0:
It's not specific to `await`, see an example without `await below. It's the "function call + object literal argument" combination. When I tested 2.0 on different codebases, I noticed that this combination is the worst idempotence offender.

**Prettier 2.0.2**
[...]

PR #9720, opened 2020-11-19
add idempotenceHazard property to groups for willBreak to check

### LINKED ISSUE #9390: Prettier requires 2 passes to format this minified code
**Environments:**

- Prettier Version: v2.1.2
- Running Prettier via: CLI
- Runtime: Node v12.18.4
- Operating System: macOS 10.15.7

**Steps to reproduce:**

The following code snippet requires double passes of prettier to be stable after minifying it.
```javascript
function a() {
  function b() {
    function c() {
      function d() {
        function e() {
          function f() {
            function g() {
              function h() {
                var e;
                0 === (h -= 1) &&
                  (i(),
                  window.addEventListener("resize", o, !1),
                  window.addEventListener("orientationchange", o, !1),
                  window.MutationObserver
                    ? ((e = new MutationObserver(o)).observe(
                        document.documentElement,
                        {
                          childList: !0,
                          subtree: !0,
                          attributes: !0,
                        }
                      ),
                      (i = function () {
                        try {
                          e.disconnect(),
                            window.removeEventListener("resize", o, !1),
                            window.removeEventListener(
                              "orientationchange",
                              o,
                              !1
                            );
                        } catch (e) {}
                      }))
                    : (document.documentElement.addEventListener(
                        "DOMSubtreeModified",
                        o,
                        !1
                      ),
                      (i = function () {
                        document.documentElement.removeEventListener(
                          "DOMSubtreeModified",
                          o,
                          !1
                        ),
                          window.removeEventListener("resize", o, !1),
                          window.removeEventListener(
                            "orientationchange",
                            o,
                            !1
                          );
                      })));
              }
            }
          }
        }
      }
    }
  }
}
```

1. Run it through `terser` (v5.3.4 in this case) and receive the following minified code. `npx terser code.js -o code.js`
```javascript
function a(){function b(){function c(){function d(){function e(){function f(){function g(){function h(){var e;0===(h-=1)&&(i(),window.addEventListener("resize",o,!1),window.addEventListener("orientationchange",o,!1),window.MutationObserver?((e=new MutationObserver(o)).observe(document.documentElement,{childList:!0,subtree:!0,attributes:!0}),i=function(){try{e.disconnect(),window.removeEventListener("resize",o,!1),window.removeEventListener("orientationchange",o,!1)}catch(e){}}):(document.documentElement.addEventListener("DOMSubtreeModified",o,!1),i=function(){document.documentElement.removeEventListener("DOMSubtreeModified",o,!1),window.removeEventListener("resize",o,!1),window.removeEventListener("orientationchange",o,!1)}))}}}}}}}}
```

2. Run prettier v2.1.2 on it once to receive the following code `npx prettier --write code.js`

```javascript
function a() {
  function b() {
    function c() {
      function d() {
        function e() {
          function f() {
            function g() {
              function h() {
                var e;
                0 === (h -= 1) &&
                  (i(),
                  window.addEventListener("resize", o, !1),
                  window.addEventListener("orientationchange", o, !1),
                  window.MutationObserver
                    ? ((e = new MutationObserver(
                        o
                      )).observe(document.documentElement, {
                        childList: !0,
                        subtree: !0,
                        attributes: !0,
                      }),
                      (i = function () {
                        try {
                          e.disconnect(),
                            window.removeEventListener("resize", o, !1),
                            window.removeEventListener(
                              "orientationchange",
                              o,
                              !1
                            );
                        } catch (e) {}
                      }))
                    : (document.documentElement.addEventListener(
                        "DOMSubtreeModified",
                        o,
                        !1
                      ),
                      (i = function () {
                        document.documentElement.removeEventListener(
                          "DOMSubtreeModified",
                          o,
                          !1
                        ),
                          window.removeEventListener("resize", o, !1),
                          window.removeEventListener(
                            "orientationchange",
                            o,
                            !1
                          );
                      })));
              }
            }
          }
        }
      }
    }
  }
}
```

3. Run prettier v2.1.2 on the code again, and observe the following diff `npx prettier --write code.js`
```diff
                   window.addEventListener("resize", o, !1),
                   window.addEventListener("orientationchange", o, !1),
                   window.MutationObserver
-                    ? ((e = new MutationObserver(
-                        o
-                      )).observe(document.documentElement, {
-                        childList: !0,
-                        subtree: !0,
-                        attributes: !0,
-                      }),
+                    ? ((e = new MutationObserver(o)).observe(
+                        document.documentElement,
+                        {
+                          childList: !0,
+                          subtree: !0,
+                          attributes: !0,
+                        }
+                      ),
                       (i = function () {
                         try {
                           e.disconnect(),
```

**Expected behavior:**
There should be no diff on the second round of prettier

**Actual behavior:**

There is a diff after running prettier the second time

thorn0:
If a group is marked as `idempotenceHazard`, it means that while this group isn't necessarily going to break on the current run of Prettier, there is a chance it will break on a subsequent one, so `willBreak` defensively returns `true` for such groups, which prevents idempotence bugs, in particular bugs related to object literals.

This fixes #9390, #7927, partially #2803, and the problem I hit with #8063 (https://github.com/prettier/prettier/pull/7889#issuecomment-614073066).

As you can see from the snapshots, there are some undesired formatting changes. Since it's unlikely we'll find a better solution for the idempotence issues, I suggest that we work around these undesired changes by refraining from preserving line breaks in object literals in some special cases, e.g. see #8055. Object literals in these special cases will be idempotence-safe and thus retain the current behavior with regard to the surrounding code.

### DIFF
--- a/src/document/doc-builders.js
+++ b/src/document/doc-builders.js
@@ -70,7 +70,12 @@ function align(n, contents) {
 
 /**
  * @param {Doc} contents
- * @param {object} [opts] - TBD ???
+ * @param {{
+ *    id?: symbol,
+ *    shouldBreak?: boolean,
+ *    expandedStates?: Doc[],
+ *    idempotenceHazard?: boolean
+ * }} [opts]
  * @returns Doc
  */
 function group(contents, opts) {
@@ -86,6 +91,7 @@ function group(contents, opts) {
     contents,
     break: !!opts.shouldBreak,
     expandedStates: opts.expandedStates,
+    idempotenceHazard: opts.idempotenceHazard,
   };
 }
 
--- a/src/document/doc-utils.js
+++ b/src/document/doc-utils.js
@@ -105,7 +105,7 @@ function isLineNext(doc) {
 }
 
 function willBreakFn(doc) {
-  if (doc.type === "group" && doc.break) {
+  if (doc.type === "group" && (doc.break || doc.idempotenceHazard)) {
     return true;
   }
   if (doc.type === "line" && doc.hard) {
--- a/src/language-js/print/object.js
+++ b/src/language-js/print/object.js
@@ -233,7 +233,7 @@ function printObject(path, options, print) {
     return content;
   }
 
-  return group(content, { shouldBreak });
+  return group(content, { shouldBreak, idempotenceHazard: !shouldBreak });
 }
 
 module.exports = { printObject };
--- a/tests/flow-repo/destructuring/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow-repo/destructuring/__snapshots__/jsfmt.spec.js.snap
@@ -230,8 +230,12 @@ function gen<T>(x: T, { p = x }: { p: T }): T {
 }
 
 // Default values in destructuring unwrap optional types
-obj_prop_fun(({}: { p?: { q?: null } })); // ok
-obj_prop_var(({}: { p?: { q?: null } })); // ok
+obj_prop_fun(
+  ({}: { p?: { q?: null } })
+); // ok
+obj_prop_var(
+  ({}: { p?: { q?: null } })
+); // ok
 
 // union-like upper bounds preserved through destructuring
 function obj_prop_opt({ p }: { p?: string } = { p: 0 }) {}
--- a/tests/flow-repo/facebookisms/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/flow-repo/facebookisms/__snapshots__/jsfmt.spec.js.snap
@@ -75,11 +75,11 @@ let tests = [
   function (copyProperties: Object$Assign) {
     let result = {};
     result.baz = false;
-    (copyProperties(result, { foo: "a" }, { bar: 123 }): {
-      foo: string,
-      bar: number,
-      baz: boolean,
-    });
+    (copyProperties(
+      result,
+      { foo: "a" },
+      { bar: 123 }
+    ): { foo: string, bar: number, baz: boolean });
   },
 
   // module from lib
@@ -234,7 +234,11 @@ let tests = [
   function (mergeInto: $Facebookism$MergeInto) {
     let result = {};
     result.baz = false;
-    (mergeInto(result, { foo: "a" }, { bar: 123 }): void);
+    (mergeInto(
+      result,
+      { foo: "a" },
+      { bar: 123 }
+    ): void);
     (result: { foo: string, bar: number, baz: boolean });
   },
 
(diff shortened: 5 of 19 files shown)


thorn0:
A case from #2803 not fixed by this PR. Looking into it.

**Prettier pr-9720**
[Playground link](https://deploy-preview-9720--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAnOMCuqoAIYCGA5gM4A6euAdAGYCWUAJgKKx0x1wkBCAngCrEAFISIBKCrmoBbAgAchkqbiEAbCOmkBJKCUzSAMhrja5e6QEYxuALwA+JcpXBRAGgBuBVZjhIAUgDKAPIAclRyBKgkcEKe3nBirggcMLxajAC+EpRiANwUIK4gEHIc0CTIoJGoEADuAAqRCBUoXrUEvBVFAEaoBGAA1hgBEWAMRMgwqD5FJt1wjIwLBgRQRJjEcABiGrIwHGvIIASYMBCFIAAWMNKqAOqX7FyjcAHN7HTu7LxHYCRdIAY0VQMHqfSIsmQNC80SKACsSAAPbh9QbDAjSOAGBhwKEwuDwpEBcaqOAARUwEHgeNUsJAESicFQR26BHmqguclQDBgdzojBgl2QAA4AAxFLkQaJ3PpyI5crhM9y4ooAR0p8DBJRaxxIAFooHAFgsLuh1XR0GDiJCkNDaQSQNFpHRJtMHSQSeSNbjbfiioRunyBUKkAAmf19OiqcYAYQg0htIC4AFYLphooJui07XT3D4dEtYAEwNzSgBBJgBVKkmnRDIZIA)
```sh
--parser babel
```

**Input:**
```jsx
return tags
  .findEntitiesByTag(tag)
  .map(
    (loremInsumLoremIpsum1) =>
      ({tag,value:JSON.parse(value),entityId})
  );

```

**Output:**
```jsx
return tags
  .findEntitiesByTag(tag)
  .map((loremInsumLoremIpsum1) => ({
    tag,
    value: JSON.parse(value),
    entityId,
  }));

```

**Second Output:**
```jsx
return tags.findEntitiesByTag(tag).map((loremInsumLoremIpsum1) => ({
  tag,
  value: JSON.parse(value),
  entityId,
}));

```

fisker:
You mean check breaks in 2 rounds? If so, should we  add some thing like a "break priority", so we can check round by round utill all settle?

thorn0:
That's not what I'm experimenting with here, but your idea is interesting too.
The problem is that `willBreak` returns `false` for object literals that are single-line in the input, but then they're sometimes split to multiple lines being printed, so on the next run `willBreak` returns `true` for the same object literals. This makes heuristics that depend on `willBreak` unreliable and leads to those idempotence issues. My idea here is simple: let's make `willBreak` return `true` for object literals on the first run.

thorn0:
Closing for now. I have other ideas for minimizing the occurrence of the object literal related idempotence issues.

PR #9726, opened 2020-11-20
Refactor: Stop mutating comments during `decorateComment`

fisker:
One step to address #8122, stop mutating comments during `decorateComment`, to achieve this, we have to pass decorate info seperately to `handleXXXComment`, I added a flag on printer, but we'll remove it when all done.

### DIFF
--- a/src/language-js/comments.js
+++ b/src/language-js/comments.js
@@ -407,12 +407,11 @@ function handleMethodNameComments(data, text) {
   return false;
 }
 
-function handleFunctionNameComments(comment, text) {
+function handleFunctionNameComments(data, text) {
+  const { comment, precedingNode, enclosingNode } = data;
   if (getNextNonSpaceNonCommentCharacter(text, comment, locEnd) !== "(") {
     return false;
   }
-
-  const { precedingNode, enclosingNode } = comment;
   if (
     precedingNode &&
     enclosingNode &&
--- a/src/language-js/comments.js
+++ b/src/language-js/comments.js
@@ -22,63 +22,60 @@ const {
 } = require("./utils");
 const { locStart, locEnd } = require("./loc");
 
-function handleOwnLineComment(context, text, options, ast, isLastComment) {
-  return (
-    handleIgnoreComments(context) ||
-    handleLastFunctionArgComments(context, text) ||
-    handleMemberExpressionComments(context) ||
-    handleIfStatementComments(context, text) ||
-    handleWhileComments(context, text) ||
-    handleTryStatementComments(context) ||
-    handleClassComments(context) ||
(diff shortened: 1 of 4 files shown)


src/language-js/comments.js
```diff
@@ -22,59 +22,59 @@ const {
 } = require("./utils");
 const { locStart, locEnd } = require("./loc");
 
-function handleOwnLineComment(comment, text, options, ast, isLastComment) {
+function handleOwnLineComment(context, text, options, ast, isLastComment) {
```
thorn0, line 25: Maybe let's move everything to one argument: `{ comment, precedingNode, enclosingNode, followingNode, text, options, ast, isLastComment }`?

fisker: I thought about this too, 

> Rename `data` -> `context` …
> I'm thinking maybe we want add a new `addLeadingComment` function through `context`

https://github.com/prettier/prettier/pull/9726/commits/cf97d08e76a7bb74e19079399cb84ca7ff153674

thorn0: I think `add*Comment` and `is*Comment` should be WeakMap-based. No need to pass them around then.

fisker: Yes, maybe

src/main/comments.js
```diff
@@ -203,30 +204,34 @@ function attach(comments, ast, text, options) {
...
+    };
+
+    let args;
+    if (noMutateComments) {
+      args = [context];
+    } else {
+      comment.enclosingNode = enclosingNode;
+      comment.precedingNode = precedingNode;
+      comment.followingNode = followingNode;
+      args = [comment, text, options, ast, isLastComment];
```
fisker, line 228: The old one seem not used by other language in core, hope I didn't break it.

src/language-js/printer-estree.js
```diff
@@ -2404,6 +2404,8 @@ module.exports = {
   printComment,
   isBlockComment,
   handleComments: {
+    // TODO: Make this as default behavior
+    noMutate: true,
```
thorn0, line 2408: Let's rename to `avoidAstMutation`.

src/language-js/comments.js
```diff
@@ -100,8 +97,8 @@ function addBlockOrNotComment(node, comment) {
   }
 }
 
-function handleClosureTypeCastComments(comment) {
-  const { followingNode } = comment;
+function handleClosureTypeCastComments(context) {
+  const { comment, followingNode } = context;
```
thorn0, line 101: Let's move the destructuring to the signature.

thorn0 (approved):

PR #9731, opened 2020-11-20
Use meriyah

fisker:
It's much faster


src/index.js
```diff
@@ -56,11 +56,8 @@ module.exports = {
     plugins.clearCache();
   },
 
-  getFileInfo: /** @type {typeof getFileInfo} */ (withPlugins(getFileInfo)),
-  getSupportInfo: /** @type {typeof getSupportInfo} */ (withPlugins(
-    getSupportInfo,
-    0
-  )),
+  getFileInfo: withPlugins(getFileInfo),
```
thorn0, line 59: Let's try to keep it typed. This might work:
```suggestion
  /** @type {typeof getFileInfo} */
  getFileInfo: withPlugins(getFileInfo),
```

thorn0 (approved):

PR #9759, opened 2020-11-23
Remove `JSXText` spacial case in comment attaching

fisker:
## Checklist


src/common/util.js
```diff
@@ -565,14 +565,6 @@ function addCommentHelper(node, comment) {
   const comments = node.comments || (node.comments = []);
   comments.push(comment);
   comment.printed = false;
-
-  // For some reason, TypeScript parses `// x` inside of JSXText as a comment
```
thorn0, line 569: Does this still happen though? I can't reproduce.

thorn0: I guess this has been fixed by https://github.com/typescript-eslint/typescript-eslint/pull/703

fisker: I think this comment mean

```js
<Foo>
  text
  // comment
  text
</Foo>
```

`// comment` is parsed as comment? I checked other parsers they parse it as 

```js
{
              "type": "JSXText",
              "start": 5,
              "end": 33,
              "loc": {
                "start": {
                  "line": 1,
                  "column": 5
                },
                "end": {
                  "line": 5,
                  "column": 0
                }
              },
              "extra": {
                "rawValue": "\n  text\n  // comment\n  text\n",
                "raw": "\n  text\n  // comment\n  text\n"
              },
              "value": "\n  text\n  // comment\n  text\n"
            }
```

And TypeScript parses it as

```
          {
            "type": "Literal",
            "value": "\n  text\n  // comment\n  text\n",
            "raw": "\n  text\n  // comment\n  text\n",
            "range": [
              5,
              33
            ]
          }
```

Anyway, I think you are right, it's not need anymore

thorn0: That's a separate thing controlled by the `useJSXTextNode` option of `typescript-estree`. Prettier uses `useJSXTextNode: true`.

thorn0 (approved):

PR #9786, opened 2020-11-28
Use option `vueIndentScriptAndStyle` to dedent 1st level children inside `<html>` tag in HTML files

nthachus:
## Description

This is a patch to dedent 1st level children inside `<html>` tag in HTML files by using `vueIndentScriptAndStyle` option.

Currently, HTML files will be formatted like:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>...</title>
  </head>
  <body>
  </body>
</html>
```

But, by using this patch and enabling `vueIndentScriptAndStyle` option, HTML files will be formatted like:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <title>...</title>
</head>
<body>
  <main>
    <section>...</section>
  </main>
</body>
</html>
```

## Checklist

I don't have time to make tests or documents, so, get this patch and apply by yourself please.

### DIFF
--- a/src/language-html/printer-html.js
+++ b/src/language-html/printer-html.js
@@ -304,6 +304,11 @@ function genericPrint(path, options, print) {
                         options.parser === "vue" &&
                         !options.vueIndentScriptAndStyle
                       ? childrenDoc
+                      : node.type === "element" &&
+                        node.fullName === "html" &&
+                        options.parser === "html" &&
+                        options.vueIndentScriptAndStyle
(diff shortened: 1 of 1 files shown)


thorn0:
Thanks for the PR. Optional behavior isn't added to Prettier anymore.
Please see https://prettier.io/docs/en/option-philosophy.html
If you want, you can open an issue to discuss changing the current behavior.

PR #9815, opened 2020-12-02
Refactor: Simplify `getSortedChildNodes`

fisker:
## Description

- Simplify logic

## Checklist

### DIFF
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -60,6 +60,8 @@ function getSortedChildNodes(node, options) {
       : getSortedChildNodes(childNode, options);
   });
 
+
+  // Sort by `start` location first, then `end` location
   childNodes.sort(
     (nodeA, nodeB) =>
       locStart(nodeA) - locStart(nodeB) || locEnd(nodeA) - locEnd(nodeB)
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -60,7 +60,6 @@ function getSortedChildNodes(node, options) {
       : getSortedChildNodes(childNode, options);
   });
 
-
   // Sort by `start` location first, then `end` location
   childNodes.sort(
     (nodeA, nodeB) =>
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -18,66 +18,55 @@ import {
   addLeadingComment,
   addDanglingComment,
   addTrailingComment,
+  isNonEmptyArray,
 } from "../common/util.js";
 
 const childNodesCache = new WeakMap();
-function getSortedChildNodes(node, options, resultArray) {
-  if (!node) {
-    return;
-  }
-  const { printer, locStart, locEnd } = options;
-
-  if (resultArray) {
-    if (printer.canAttachComment && printer.canAttachComment(node)) {
-      // This reverse insertion sort almost always takes constant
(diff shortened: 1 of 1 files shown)


thorn0:
The changes don't look safe, and the function became more difficult to follow IMHO.

fisker:
Which part do you think is not safe?

I think this great improvement.

Run format on `src/index.js` file, let's see how many times related functions are called

```
Master
┌──────────────────────┬────────┐
│       (index)        │ Values │
├──────────────────────┼────────┤
│ getSortedChildNodes  │  196   │
│ getCommentChildNodes │  126   │
│   canAttachComment   │  160   │
│       locStart       │  140   │
│        locEnd        │  140   │
└──────────────────────┴────────┘
```

```
This PR
┌──────────────────────┬────────┐
│       (index)        │ Values │
├──────────────────────┼────────┤
│ getSortedChildNodes  │  151   │
│ getCommentChildNodes │   94   │
│   canAttachComment   │  128   │
│       locStart       │  149   │
│        locEnd        │  149   │
└──────────────────────┴────────┘
```

And here `getCommentChildNodes` function was called with argument without `type` property

```
┌─────────┬────────┐
│ (index) │ Values │
├─────────┼────────┤
│ string  │   11   │
│  array  │   15   │
│ number  │   64   │
│ object  │   27   │
└─────────┴────────┘
```

fisker:
I pushed some new changes, it should be much easier to understand the logic.

thorn0:
I'm reluctant to change the sorting algorithm without any benchmarks. And I have trouble with understanding the function you pass to `sort`:

```js
    .sort(
      (aNode, bNode) =>
        locStart(aNode) - locStart(bNode) || locEnd(aNode) - locEnd(bNode)
    );
```

Are you sure that's what you meant? Maybe I need to have another look at it...

Also you changed the behavior of the function a little. Previously, if `printer.canAttachComment(node)` returned `false`, the function was called recursively and the children of the `node` were added to the list. Now they aren't. This doesn't break the built-in languages, but might break plugins. The removed behavior can be a useful way for excluding certain nodes from comment attaching without excluding their children, so I'd prefer to keep it.

fisker:
> Are you sure that's what you meant? 

Yes, it is. Why this hard to understand?

```js
 (aNode, bNode) =>
        locStart(aNode) - locStart(bNode) || locEnd(aNode) - locEnd(bNode)
//      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ If locStart(aNode) < locStart(bNode), aNode first
//                                        If same, will be `0`, sort by `locEnd`
```

> Also you changed the behavior of the function a little. 

Seems you're right, I forgot to consider that `canAttachComment` returns false, I only considered that branch hit `canAttachComment` be `undefined`, my bad.

fisker:
It seems I finally got the logic right in https://github.com/prettier/prettier/pull/13255, it #13255 didn't get merge, I may move code to this one.

fisker:
> if printer.canAttachComment(node) returned false, the function was called recursively and the children of the node were added to the list. Now they aren't.

Somehow, I missed this, I have to read the code again to understand how it works in #13255 😢

fisker:
Included in 13255

Issue #6961, opened 2019-11-15
MDX: line break inserted between elements and text in JSX blocks


thorn0:
After we merge #6949 to fix #6943, we still have an issue with whitespace between JSX elements and text in JSX blocks.

References:
- https://github.com/prettier/prettier/pull/6949#pullrequestreview-317348429
- https://spectrum.chat/mdx/general/whitespace-in-block-level-jsx~38817f54-db82-4971-be9e-584e1c4d20ff

**Prettier pr-6949**
[Playground link](https://deploy-preview-6949--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeAhgZwCYDMD0AfAF4AeYAbiADQgQAOMAltJsqOgE6cQDuAClwRsU6ChCbYaIAEad0YANZwYAZXoKmUAObIYnAK5xaACxgBbADYB1E03iYNYOKuH2mFewE9k4TG1otTDhOGH55bXN0ZFx0S2DaACtMUgAheSUVVXRzOAAZLTgYuISQZNJVLW1LOABFAwh4YvjjEA1OYM5fc2xSaXpOLRhrSRgTZAAOAAZaAYhg63l6XwG4Tooi2gBHBvhwhhEQLABaKDg4bAvpTjgdphvw9EjopFiW2mDzJj1DVswqmr1RpFV4lVowdAyEbYMbIABMtH06CYliqAGEIOYor4oNBNiADMEACqQkRvUoUIwASSgl1gqjAg0YAEFaaoYF4as1ggBfHlAA)
```sh
--parser mdx
```

**Input:**
```mdx
<asdf/>zxcv
```

**Output:**
```mdx
<asdf />
zxcv

```

**Expected behavior:**
```mdx
<asdf/>zxcv
```

PR #9847, opened 2020-12-08
Refactor markdown `paragraph` print

### LINKED ISSUE #9503: Inline React components in MDX cause line lengths to exceed print-width
**Prettier 2.1.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAogD3gJygQwBsAZASygGsACACxzgDMBeAHRBphgAckB6X-AHcAzgDp8AW3wAvaKMgTekACZwupLnELk4vNgD4AggHUAylQDCEVQAUNWnal5ZcBEuQr6qpYVXxVIWHIAVwhg31VtADc4HABPPyhlKnpCOHxhOD9gmAgpGFJoKkycKNIwLJgafBg-UmVfYQkICCqqVS5CCDiJBBgxKgBNMID8KG9YOABzHBqsgGYcZK58HBgE3IhCX3JcxITheC4qCAYqOLCcFK10zKouHAgK4V8IK4vg0fHwrKtbe20UCyGUSVAQygAtLkIeDilscoUoAAaCZgQjBZTkKZUADiNAghyoAEkoMJNGBcld4IdfGNksJgqRqaIQEiQBAuAVoMJkKBVo9BDZVggeSgiIJ8HEeWyAEazMAUOAwUwrMBY5AwHDBOBsuASGVwZSqZTEMZTYL4KZwABib3yBSgU2QIHwOQgrPYMAkhGMNCZcDJ+AqphFTNIZXWzrALw95BKMBssymUmQDCImTZACthJgAELyxXKyRwMhA1PpnUgbOYUxYtIARVC8HL20rKxwJWdEmUmA9DwJcGMsy4zoeAdiMQ9AEcm3BExzRS7hBCgYbDR76DPSPRE5aU0g0622ZkJKQNVrK8I63BG604C2MyAYPgZcZ6lVkAAmNma-CkQFTFYEj7iAAYAKwej8AAqL6ioej5RNqJKqLAphgDgGgwIYSSmOsaQPm2GGwG+ygfkgACMAAMlEAL40UAA)
```sh
--parser mdx
--print-width 100
```

**Input:**
```mdx
<ExternalLink href="http://aws.amazon.com/codepipeline/">AWS CodePipeline</ExternalLink> is a continuous delivery and release automation service that aids smooth deployments. You can integrate 3rd party tools into any step of your release process or you can use CodePipeline as an end-to-end solution, including Ghost Inspector tests and suites.
```

**Output:**
```mdx
<ExternalLink href="http://aws.amazon.com/codepipeline/">AWS CodePipeline</ExternalLink> is a continuous
delivery and release automation service that aids smooth deployments. You can integrate 3rd party tools
into any step of your release process or you can use CodePipeline as an end-to-end solution, including
Ghost Inspector tests and suites.

```

**Expected behavior:**

Here the output optically looks correct, however if you view it in the playground or any editor that has a line length “ruler” you can see that each line is actually longer than 100 characters.

```
❯ echo '<ExternalLink href="http://aws.amazon.com/codepipeline/">AWS CodePipeline</ExternalLink> is a continuous' | wc
       1       6     105
```

This causes a bit of a problem if you have word wrapping on in your editor.

Screen Shot 2020-10-26 at 1 58 54 PM

If I remove the inline component it works fine, so the bug seems to have something to do with inline React components.

Thanks!

fisker:
## Description

Fix #9503

## Checklist


tests/mdx/mdx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -381,7 +382,8 @@ printWidth: 80
 </Hello>
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
+123
```
fisker, line 386: This change may related to #6961

thorn0:
Ooops, I keep forgetting about this one...

thorn0:
The doc for #9503 still doesn't look right:
```js
[
  fill([
    conditionalGroup([
      group([/*...*/]),
      group([/*...*/]),
    ]),
    " ", // <--- should be `line`
    "is",
    line,
// ...
```

src/language-markdown/printer-markdown.js
```diff
@@ -84,15 +84,18 @@ function genericPrint(path, options, print) {
...
-    case "paragraph":
-      return printChildren(path, options, print, {
-        postprocessor: fill,
-      });
+    case "paragraph": {
+      const printed = cleanDoc(printChildren(path, options, print));
+      if (typeof printed === "string") {
+        return printed;
+      }
+      return fill(getDocParts(printed));
```
thorn0, line 97: Passing the result of `cleanDoc` to `fill` is wrong. `cleanDoc` doesn't guarantee that the array has the expected structure.

thorn0:
Yes, this is a big one. My research plan so far:

## MDX: line break inserted between elements and text in JSX blocks

```<asdf/>zxcv```

- Prettier 2.2.1 parses it all as JSX (not Markdown) and adds line breaks the way it normally does for JSX.
- That's probably not the expected output. That's why #6961 was opened.
- Also there is a `fill` bug that causes v2.2.1 to print `<Hello></Hello><Hello></Hello>123` without that break.
  ```js
  [
    fill([
      group(["<Hello", indent([]), ">"]),
      "</Hello>", // <--- Problem! Odd-indexed elements must be line breaks.
      hardlineWithoutBreakParent,
      breakParent,
      group(["<Hello", indent([]), ">"]),
      "</Hello>",
      softline,
      "123"
    ]),
    hardline,
  ]
  ```
- Things to look into:
  1. The `fill` bug
     1. Is the bug in #9503 same or different?
     2. How did it work in older versions?
     3. Was `normalizeDoc` safe before it got extracted to `doc-utils`?
     4. Manipulations with `normalizeDoc` and `cleanDoc` (in particular, PR #9847) look like they might affect more things in Markdown than just MDX. What else can be affected?
  2. What actually is the expected output?
     1. Should the text be parsed as JSX or as Markdown?

src/language-markdown/printer-markdown.js
fisker: Yes, you are  right, and it seems I've done something wrong in other places, I'll check.

Issue #9881, opened 2020-12-13
Introduce `htmlWhitespaceSensitivity` option for Handlebars


dcyriller:
I open an issue to discuss / track the progress on the following feature.

## Introduce the [HTML Whitespace Sensitivity](https://prettier.io/docs/en/options.html#html-whitespace-sensitivity) option for Handlebars

In the Handlebars printer, whitespaces are currently considered insentitive. The goal of this feature is to introduce a mode to treat whitespaces as _sensitive_. It will use html plugin's option: `htmlWhitespaceSensitivity`.

The following commands would then be supported:
```sh
yarn prettier --html-whitespace-sensitivity=ignore template.hbs
yarn prettier --html-whitespace-sensitivity=strict template.hbs
yarn prettier template.hbs # currently defaults to `ignore`
```

## Progress

I slice the feature this way to ease the review:

1. [Add snapshots for `htmlWhitespaceSensitivity=ignore` and  `htmlWhitespaceSensitivity=strict`](https://github.com/prettier/prettier/pull/9883). It might be pure duplication at this point. But it will be useful when a/ we will implement the `strict` mode, b/ we will switch the default to `strict`.
2. [Do a first implementation of `htmlWhitespaceSensitivity=strict` option](https://github.com/prettier/prettier/pull/9885).
3. Update the documentation for the option to take into account how it behaves for Handlebars.
4. [optional] Switch the default to `strict`. (note: this is not urgent, and we will have to go through a beta phase before merging that)

## Out of scope

html also has the possible `css` value (its default). It could be implemented later.

thorn0:
@dcyriller Oops, we forgot to add a changelog entry in #9885. Could you submit a PR adding it?

dcyriller:
Done ✅ in https://github.com/prettier/prettier/pull/10133 (I initially didn't add it because this mode was not user facing).

PR #9882, opened 2020-12-13
handlebars: Add `htmlWhitespaceSensitivity` option (1/6)

dcyriller:
## Description

Step 1 out of 6 of https://github.com/prettier/prettier/issues/9881

The only possible and default value is `ignore` for now.

In further steps, we will add tests and implement a `strict` mode to consider whitespaces sensitive.

I have not updated the `docs/` directory yet. I am planning to do it in a following PR (when `strict` mode is implemented).

## Checklist

### DIFF
--- a/src/language-handlebars/index.js
+++ b/src/language-handlebars/index.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const createLanguage = require("../utils/create-language");
+const options = require("./options");
 const printer = require("./printer-glimmer");
 
 const languages = [
@@ -24,5 +25,6 @@ const parsers = {
 module.exports = {
   languages,
   printers,
+  options,
   parsers,
 };
--- a/src/language-handlebars/options.js
+++ b/src/language-handlebars/options.js
@@ -1,0 +1,24 @@
+"use strict";
+
+const CATEGORY_HANDLEBARS = "Handlebars";
+
+// format based on https://github.com/prettier/prettier/blob/master/src/main/core-options.js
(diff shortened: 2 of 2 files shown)


fisker (approved):

dcyriller:
I just spotted an issue with this implementation.

On the following line:
https://github.com/prettier/prettier/blob/ce6f07efe9fdbbbe21f3c17aa48fe959e4dbbaca/src/main/support.js#L37
the various plugins' options are merged into one object. Then, the option added in this PR is overriden by the html plugin's `htmlWhitespaceSensitivity` option.

**A first option**

First option: we want to stick to the same option for HTML and Handlebars (`htmlWhitespaceSensitivity`). Then the option has to be defined only once. This has a drawback: it either forces us to implement the `css` mode for Handlebars, because it is the default mode in HTML. Or alternatively, we could treat the `css` mode as `ignore` mode in Handlebars. Indeed, `ignore` mode is Handlebars' current default. Also, as a side-node it would feel slightly odd to me that the option is defined in the HTML plugin if it is also used by the Handlebars plugin.

**A second option**

An other option could be to rename this option to `handlebarsWhitespaceSensitivity` or `hbsWhitespaceSensitivity`.

Wdyt?

fisker:
I think we can treat `css` as `ignore` first.

dcyriller:
Then, we can skip this PR as the option is defined in the HTML plugin. : ) I'm closing and focusing on the next one (2 / 6).

PR #9883, opened 2020-12-13
Handlebars: Add tests for `ignore` and `strict` modes (1/3)

dcyriller:
## Description

This PR proposes a refactor of the `whitespace` folder in Handlebars tests.
- it renames the folder from `whitespace` to `mode-default` and does not configure `htmlWhitespaceSensitivity` option
- it copies the `mode-default` folder to `mode-ignore` and configures `htmlWhitespaceSensitivity=ignore`
- it copies the `mode-default` folder to `mode-strict` and configures `htmlWhitespaceSensitivity=strict`

My goal is to propose an implementation of the whitespace sensitive mode in the next PR. I propose to add tests first to have a nice diff. The PR might be reviewed commit by commit.

Step 1 out of 3 of https://github.com/prettier/prettier/issues/9881

## Checklist


fisker:
Do you think better add a dir to test for it?

dcyriller:
100% agreed, let me do that

dcyriller:
The PR is updated, as well as the description (as I repurposed it after your suggestion)

dcyriller:
Do you think it’s better @fisker?

fisker:
I don't think we should move other tests, how about a dir named `html-whitespace-sensitivity`, and test both `ignore` and `strict` in one file.

I think still too many files for this option, shouldn't a few cases enough, but if you think all lines are necessary, I'm also okay with it?

fisker (approved):
LGTM, thank you!

PR #9885, opened 2020-12-13
handlebars: Support `htmlWhitespaceSensitivity=strict` option (2/3)

dcyriller:
## Description

This PR proposes an implementation of the `htmlWhitespaceSensitivity=strict` option for Handlebars.

Step 2 out of 3 of https://github.com/prettier/prettier/issues/9881

## Preview

The git diff is not yet meaningful since the target branch is master. To get an idea of the changes, [see this PR](https://github.com/dcyriller/prettier/pull/1/files) (it targets the branch 2 / 3 on my fork).

## Checklist

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -185,7 +185,7 @@ function print(path, options, print) {
         const whitespacesOnlyRE = /^[\t\n\f\r ]*$/;
 
         if (n.chars.match(whitespacesOnlyRE)) {
-          let breaks = line;
+          let breaks = [line];
 
           const newlines = countNewLines(n.chars);
           if (newlines) {
@@ -193,7 +193,7 @@ function print(path, options, print) {
           }
 
           if (isLastNodeOfSiblings(path)) {
-            breaks = dedent(breaks);
+            breaks = breaks.map((newline) => dedent(newline));
           }
 
           return breaks;
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const {
-  builders: { group, hardline, ifBreak, indent, join, line, softline },
+  builders: { dedent, group, hardline, ifBreak, indent, join, line, softline },
   utils: { getDocParts },
 } = require("../document");
 const { locStart, locEnd } = require("./loc");
@@ -10,6 +10,7 @@ const {
   getNextNode,
   getPreviousNode,
   hasPrettierIgnore,
+  isLastNodeOfSiblings,
   isNextNodeOfSomeType,
   isNodeOfSomeType,
   isParentOfSomeType,
@@ -39,28 +40,44 @@ function print(path, options, print) {
     case "Template": {
       return group(path.map(print, "body"));
     }
+
     case "ElementNode": {
-      // TODO: make it whitespace sensitive
-      const bim = isNextNodeOfSomeType(path, ["ElementNode"]) ? hardline : "";
+      const startingTag = group(printStartingTag(path, print));
+
+      const escapeNextElementNode =
+        options.htmlWhitespaceSensitivity !== "strict" &&
+        isNextNodeOfSomeType(path, ["ElementNode"])
+          ? softline
+          : "";
 
       if (isVoid(n)) {
-        return [group(printStartingTag(path, print)), bim];
+        return [startingTag, escapeNextElementNode];
       }
 
-      const isWhitespaceOnly = n.children.every((n) => isWhitespaceNode(n));
+      const endingTag = ["</", n.tag, ">"];
+
+      if (n.children.length === 0) {
+        return [startingTag, indent(endingTag), escapeNextElementNode];
+      }
+
+      if (options.htmlWhitespaceSensitivity !== "strict") {
+        return [
+          startingTag,
+          indent(printChildren(path, options, print)),
+          hardline,
+          indent(endingTag),
+          escapeNextElementNode,
+        ];
+      }
 
       return [
-        group(printStartingTag(path, print)),
-        group([
-          isWhitespaceOnly ? "" : indent(printChildren(path, options, print)),
-          n.children.length > 0 ? hardline : "",
-          "</",
-          n.tag,
-          ">",
-        ]),
-        bim,
+        startingTag,
+        indent(group(printChildren(path, options, print))),
+        indent(endingTag),
+        escapeNextElementNode,
       ];
     }
+
     case "BlockStatement": {
       const pp = path.getParentNode(1);
 
@@ -74,20 +91,21 @@ function print(path, options, print) {
       if (isElseIf) {
         return [
           printElseIfBlock(path, print),
-          printProgram(path, print),
-          printInverse(path, print),
+          printProgram(path, print, options),
+          printInverse(path, print, options),
         ];
       }
 
       return [
         printOpenBlock(path, print),
         group([
-          printProgram(path, print),
-          printInverse(path, print),
-          printCloseBlock(path, print),
+          printProgram(path, print, options),
+          printInverse(path, print, options),
+          printCloseBlock(path, print, options),
         ]),
       ];
     }
+
     case "ElementModifierStatement": {
       return group(["{{", printPathAndParams(path, print), softline, "}}"]);
     }
@@ -158,6 +176,65 @@ function print(path, options, print) {
       return [n.key, "=", path.call(print, "value")];
     }
     case "TextNode": {
+      const inAttrNode = path.stack.includes("attributes");
+
+      if (options.htmlWhitespaceSensitivity === "strict" && !inAttrNode) {
+        // https://infra.spec.whatwg.org/#ascii-whitespace
+        const leadingWhitespacesRE = /^[\t\n\f\r ]*/;
+        const trailingWhitespacesRE = /[\t\n\f\r ]*$/;
+        const whitespacesOnlyRE = /^[\t\n\f\r ]*$/;
+
+        if (n.chars.match(whitespacesOnlyRE)) {
+          let breaks = line;
+
+          const newlines = countNewLines(n.chars);
+          if (newlines) {
+            breaks = generateHardlines(newlines, 2);
+          }
+
+          if (isLastNodeOfSiblings(path)) {
+            breaks = dedent(breaks);
+          }
+
+          return breaks;
+        }
+
+        const [lead] = n.chars.match(leadingWhitespacesRE);
+        const [tail] = n.chars.match(trailingWhitespacesRE);
+
+        let text = n.chars;
+
+        let leadBreaks = [];
+        if (lead) {
+          leadBreaks = [line];
+
+          const leadingNewlines = countNewLines(lead);
+          if (leadingNewlines) {
+            leadBreaks = generateHardlines(countNewLines(lead) || 1, 2);
+          }
+
+          text = text.replace(leadingWhitespacesRE, "");
+        }
+
+        let trailBreaks = [];
+        if (tail) {
+          trailBreaks = [line];
+
+          const trailingNewlines = countNewLines(tail);
+          if (trailingNewlines) {
+            trailBreaks = generateHardlines(trailingNewlines || 1, 2);
+
+            if (isLastNodeOfSiblings(path)) {
+              trailBreaks = trailBreaks.map((hardline) => dedent(hardline));
+            }
+          }
+
+          text = text.replace(trailingWhitespacesRE, "");
+        }
+
+        return [...leadBreaks, text, ...trailBreaks];
+      }
+
       const maxLineBreaksToPreserve = 2;
       const isFirstElement = !getPreviousNode(path);
       const isLastElement = !getNextNode(path);
@@ -191,7 +268,6 @@ function print(path, options, print) {
         }
       }
 
-      const inAttrNode = path.stack.includes("attributes");
       if (inAttrNode) {
         // TODO: format style and srcset attributes
         // and cleanup concat that is not necessary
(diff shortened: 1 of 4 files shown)


fisker:
#9883 Merged

dcyriller:
Still a draft because I have to refactor / cleanup the code I wrote.

src/language-handlebars/printer-glimmer.js
```diff
@@ -162,6 +194,46 @@ function print(path, options, print) {
...
+
+          if (isLastNodeOfSiblings(path)) {
+            return dedent(concat(breaks));
+          }
+
+          return concat(breaks);
+        }
+
+        const [lead] = n.chars.match(/^\s+/) || [];
+        const [tail] = n.chars.match(/\s+$/) || [];
```
fisker, line 212:
```suggestion
        const [lead] = n.chars.match(/^\s*/);
        const [tail] = n.chars.match(/\s*$/);
```

Will this work?

src/language-handlebars/printer-glimmer.js
```diff
@@ -162,6 +194,46 @@ function print(path, options, print) {
...
+
+        let text = n.chars;
+        const hasLeadingWhitespace = lead && isFirstNodeOfSiblings(path);
+        const hasTrailingWhitespace = tail && isLastNodeOfSiblings(path);
+
+        let leadBreaks = [];
+        if (hasLeadingWhitespace) {
+          const leadLineBreaksCount = countNewLines(lead);
+          text = text.replace(/^\s+/, "");
```
fisker, line 221: I don't think we should use `\s` in all these places, shouldn't it the same as html whitespace ? https://github.com/prettier/prettier/blob/196dfd207a78b11ddbfdee258ce41b55c3a3fa70/src/language-html/utils.js#L17

src/language-handlebars/printer-glimmer.js
dcyriller: It works, nice catch

src/language-handlebars/printer-glimmer.js
dcyriller: this is right, thank you

fisker: Need tests for it, you can use `snippets` to add escaped whitespace https://github.com/prettier/prettier/blob/196dfd207a78b11ddbfdee258ce41b55c3a3fa70/tests/html/whitespace/jsfmt.spec.js#L6

dcyriller: excellent!

src/language-handlebars/printer-glimmer.js
```diff
@@ -162,6 +194,51 @@ function print(path, options, print) {
       return concat([n.key, "=", path.call(print, "value")]);
     }
     case "TextNode": {
+      const inAttrNode = path.stack && path.stack.includes("attributes");
```
fisker, line 196: Isn't `path.stack` always an `array`?

src/language-handlebars/printer-glimmer.js
```diff
@@ -162,6 +193,49 @@ function print(path, options, print) {
...
+        const [lead] = n.chars.match(leadingWhitespacesRE);
+        const [tail] = n.chars.match(trailingWhitespacesRE);
+
+        let text = n.chars;
+
+        let leadBreaks = [];
+        if (lead) {
+          const leadLineBreaksCount = Math.min(countNewLines(lead) || 1, 2);
+          text = text.replace(leadingWhitespacesRE, "");
+          leadBreaks = new Array(leadLineBreaksCount).fill(hardline);
```
fisker, line 224: There is a function `generateHardlines`, can we use it?

dcyriller:
@fisker I added two commits to address your comments (+ I rebased)

src/language-handlebars/printer-glimmer.js
```diff
@@ -438,9 +518,12 @@ function printOpenBlock(path, print) {
...
 }
 
-function printElseBlock(node) {
+function printElseBlock(node, options) {
+  const escape =
+    options.htmlWhitespaceSensitivity !== "strict" ? [hardline] : [];
+
   return concat([
-    hardline,
+    ...escape,
```
fisker, line 526:
```suggestion
  return concat([
    options.htmlWhitespaceSensitivity !== "strict" ? hardline : "",
```

No need use `array`

src/language-handlebars/printer-glimmer.js
```diff
@@ -458,11 +541,18 @@ function printElseIfBlock(path, print) {
...
+  let escape = [];
+  if (options.htmlWhitespaceSensitivity !== "strict") {
+    escape = blockStatementHasOnlyWhitespaceInProgram(node)
+      ? [softline]
+      : [hardline];
+  }
+
   return concat([
-    blockStatementHasOnlyWhitespaceInProgram(node) ? softline : hardline,
+    ...escape,
```
fisker, line 555:
```suggestion
  let escape = '';
  if (options.htmlWhitespaceSensitivity !== "strict") {
    escape = blockStatementHasOnlyWhitespaceInProgram(node)
      ? softline
      : hardline;
  }

  return concat([
    escape,
```

Same as above

src/language-handlebars/printer-glimmer.js
```diff
@@ -489,29 +579,37 @@ function blockStatementHasElse(node) {
...
 
   if (blockStatementHasElseIf(node)) {
-    return parts;
+    return concat(parts);
   }
 
   if (blockStatementHasElse(node)) {
-    return concat([printElseBlock(node), indent(parts)]);
+    return concat([printElseBlock(node, options), indent(concat(parts))]);
   }
```
fisker, line 613:
```suggestion
  const printed =
    options.htmlWhitespaceSensitivity !== "strict"
      ? concat([hardline, inverse])
      : inverse;

  if (blockStatementHasElseIf(node)) {
    return printed;
  }

  if (blockStatementHasElse(node)) {
    return concat([printElseBlock(node, options), indent(printed)]);
  }
```

fisker (changes requested):

src/language-handlebars/printer-glimmer.js
```diff
@@ -40,27 +51,46 @@ function print(path, options, print) {
...
     case "ElementNode": {
-      // TODO: make it whitespace sensitive
-      const bim = isNextNodeOfSomeType(path, ["ElementNode"]) ? hardline : "";
+      const startingTag = group(printStartingTag(path, print));
+
+      const escapeNextElementNode =
+        options.htmlWhitespaceSensitivity !== "strict" &&
+        isNextNodeOfSomeType(path, ["ElementNode"])
+          ? [softline]
+          : [];
```
fisker, line 60: This should be `? softline : ""` too

src/language-handlebars/printer-glimmer.js
```diff
@@ -166,6 +197,48 @@ function print(path, options, print) {
...
+          leadBreaks = generateHardlines(countNewLines(lead) || 1, 2);
+        }
+
+        let trailBreaks = [];
+        if (tail) {
+          text = text.replace(trailingWhitespacesRE, "");
+          trailBreaks = generateHardlines(countNewLines(tail) || 1, 2);
+
+          if (isLastNodeOfSiblings(path)) {
+            trailBreaks = trailBreaks.map((h) => dedent(h));
```
fisker, line 235:
```suggestion
            trailBreaks = trailBreaks.map((hardline) => dedent(hardline));
```

tests/handlebars/html-whitespace-sensitivity/jsfmt.spec.js
```diff
@@ -1,8 +1,83 @@
...
+      ].map((textContent) => ({
+        code: `<div>${textContent}</div>`,
+        name: "content",
+        output: "<div>\n</div>",
+      })),
+
+      // many
+      {
+        code: "<div>\u0009\u000A\u000C\u000D\u0020</div>",
+        output: "<div>\n\n</div>",
```
fisker, line 28: Shouldn't this be `<div>\n</div>` ?

fisker:
Another question: Why use `hardline` instead of `line`? Am I missing something?

src/language-handlebars/printer-glimmer.js
dcyriller: it is a trick to avoid pushing an empty string in the concat args ^^

dcyriller: but if you prefer, I will remove the array

tests/handlebars/html-whitespace-sensitivity/jsfmt.spec.js
dcyriller: `\u0009\u000A\u000C\u000D\u0020` includes 2 newlines, which are preserved. From a browser perspective, `<div>\n</div>` and `<div>\n\n</div>` will be rendered the same way. But from an author perspective, it can be nice sometimes to keep the additional newline for readablility purpose. For instance:

```hbs
<div>
    <div>
    </div>

    <div>
    </div>
</div>
```
wdyt?

dcyriller: please note that in this implementation, 3 (or more) newlines will be reduced to two

src/language-handlebars/printer-glimmer.js
dcyriller: I manually applied your suggestion (it was not possible anymore after the rebase)

src/language-handlebars/printer-glimmer.js
dcyriller: I fixed this :)

src/language-handlebars/printer-glimmer.js
dcyriller: it is changed

dcyriller:
> Another question: Why use hardline instead of line? Am I missing something?

Thanks for the challenge. Indeed, it looks better using `line`.

fisker (approved):
LGTM

dcyriller:
Squashed in one commit to ease the rebase on master

tests/handlebars/html-whitespace-sensitivity/jsfmt.spec.js
thorn0: That's inconsistent with how Prettier treats empty lines everywhere else: https://prettier.io/docs/en/rationale.html#empty-lines

dcyriller: @thorn0 I was not aware of the the [empty lines doc](https://prettier.io/docs/en/rationale.html#empty-lines), I have been somehow lucky :)

thorn0:
@dcyriller Have you intentionally decided to print things differently than Prettier's HTML printer does with the `strict` setting?

**Prettier pr-9885**
[Playground link](https://deploy-preview-9885--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzgBwIZQARgA2266AvADogDuEATgCZUB8GO+RJ5V6AnocQBGhOCza4CxUpRAiY8OngBuEanEIsAlqgD0WXK30cp3WXHlxFkKOmi4YLTLqOs97ClEPtJXGXwHYwqIgXhKc0lRyCgTQtlD2eLHmLOjO7KHGvpHm0SpqGiEArmkG4pkRZhZWsXawLAC2JZ5Nri4gADQgEJgwmrHIoNh0dKoACkMI6Mgg2ITU2LxTnYJ02GAA1uYAyjhgmlAA5sgwdIVwnXD1gnAMDDcAMrgHhdgHcABi9PXY8vtHKNhCjAIB0QAALGD1QgAdTBmng+jAcC2k3hmiU8N403QJ00YBgoP26EsMFGqwO32QADNZsTOgArdAADwAQqsNttsPU4Pd9nBqbTziBGUytn8RABFQoQeACwh0kA4OjEujTCFQ0GYOj7GDQzQMGBg5AADgADJ0tRBidDVphplq4CqlPzOgBHaXwMndKYA9AAWigcBuN1BdDg7s0YbJr0pSBp8qFxPqmmOp0T4rgUpl-Ljgs6MECeoNRqQACZ86tNIQ-gBhCD1WMgR0AVlBhWJABVAj74wqlGcAJJQO6wLZgbU9ACCw62MH4Od7cAAvkugA)
```sh
--html-whitespace-sensitivity strict
--parser html
```

**Input:**
```html
<span class="word"><span class="syllable"><span class="letter vowel">i</span><span class="letter consonant">p</span></span
><span class="syllable"><span class="letter consonant onset">s</span><span class="letter vowel">u</span><span class="letter consonant">m</span></span></span>
```

**Output:**
```html
<span class="word"
  ><span class="syllable"
    ><span class="letter vowel">i</span
    ><span class="letter consonant">p</span></span
  ><span class="syllable"
    ><span class="letter consonant onset">s</span
    ><span class="letter vowel">u</span
    ><span class="letter consonant">m</span></span
  ></span
>

```
**Prettier pr-9885**
[Playground link](https://deploy-preview-9885--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzgBwIZQARgA2266AvADogDuEATgCZUB8GO+RJ5V6AnocQBGhOCza4CxUpRAiY8OngBuEanEIsAlqgD0WXK30cp3WXHlxFkKOmi4YLTLqOs97ClEPtJXGXwHYwqIgXhKc0lRyCgTQtlD2eLHmLOjO7KHGvpHm0SpqGiEArmkG4pkRZhZWsXawLAC2JZ5Nri4gADQgEJgwmrHIoNh0dKoACkMI6Mgg2ITU2LxTnYJ02GAA1uYAyjhgmlAA5sgwdIVwnXD1gnAMDDcAMrgHhdgHcABi9PXY8vtHKNhCjAIB0QAALGD1QgAdTBmng+jAcC2k3hmiU8N403QJ00YBgoP26EsMFGqwO32QADNZsTOgArdAADwAQqsNttsPU4Pd9nBqbTziBGUytn8RABFQoQeACwh0kA4OjEujTA6ETT1bmqzqYOj7GDQzQMGBg5AADgADLqRsToatMNM9XAVUp+Z0AI7S+Bk7pTAHoAC0UDgNxuoLocC9mkjZNelKQNPlQuJ9U0x1OKfFcClMv5icFnRggSNJrNSAATEXVpoNYcAMIQLXYaYugCsoMKxIAKoF-UmFUozgBJKB3WBbMD6noAQTHWxg-HzA7gAF9V0A)
```sh
--html-whitespace-sensitivity strict
--parser glimmer
```

**Output:**
```hbs
<span class="word"><span class="syllable"><span
      class="letter vowel"
    >i</span><span class="letter consonant">p</span></span><span
    class="syllable"
  ><span class="letter consonant onset">s</span><span
      class="letter vowel"
    >u</span><span class="letter consonant">m</span></span></span>
```

dcyriller:
> Have you intentionally decided to print things differently than Prettier's HTML printer does with the strict setting?

@thorn0 I did not decide this intentionally and I agree you have a fair point. The formatting could be prettier and closer to HTML's. But when I tried it, I introduced other issues. My goal with this PR was more to have a whitespace sensitive mode. I did not focus that much on printing like html.

thorn0:
I see, I just wanted to know if this was intentional. Ultimately it'd be good to unify how the two printers work and to use the same test for them, but for now this PR looks good enough.

thorn0 (approved):

PR #9901, opened 2020-12-16
Send array to `FastPath#{each,map}` callback

fisker:
## Description

Similar to [`Array#map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#Parameters), send array as third argument, this is useful when need access array in the callback.

## Checklist


sosukesuzuki (approved):
Please resolve conflicts.

fisker:
I'm going to use this in #9926

PR #9902, opened 2020-12-16
Add `isNonEmptyArray`

fisker:
## Description

Add `isNonEmptyArray` and enforce to use it.

## Checklist


sosukesuzuki (approved):
This PR looks good. Btw, what do you think about adding tests for our internal ESLint rules?

fisker:
Yes, I think it should, when I wrote this one, I was afraid the previous logic will broken.... But maybe later

PR #9936, opened 2020-12-18
Html: Enable jsxBracketSameLine option for html (fixes #5377)

### LINKED ISSUE #5377: [resolved] [HTML] allow corresponding option to jsxBracketSameLine
Note: Read this comment to make sure you’re up-to-speed on what this issue is and is not about: https://github.com/prettier/prettier/issues/5377#issuecomment-628213559

---

For React/JSX - Prettier has [an option](https://prettier.io/docs/en/options.html#jsx-brackets) to allow tags to close on the same line as the declaration, rather then on a new line:
`jsxBracketSameLine: true`:
```
<button
  className="prettier-class"
  id="prettier-id"
  onClick={this.handleClick}>
  Click Here
</button>
```

`jsxBracketSameLine: false`:
```
<button
  className="prettier-class"
  id="prettier-id"
  onClick={this.handleClick}
>
  Click Here
</button>
```
With Prettier 1.15.1, at least for the Angular parser, there is no corresponding option, so all HTML is formatted like this:
Input:
```html
<my-component class="wcd-flex-1 dialog-contents wcd-scroll-vertical"
		      [foo]="bar">
</my-component>
```
Output:
```html
<my-component
    class="wcd-flex-1 dialog-contents wcd-scroll-vertical"
	[foo]="bar"
>
</my-component>
```
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAtgTwLSXQB2gRgAIwAbAQwGdqBeAHRAHcwATbAM3LgA9sAjCTYBLSuQgBzXNHixqJVh2pgAThHLlsANziqYIsOKYMoDGOZJXrAbU4QIAXUYgARpVVMAfKdQB6LBkCIlgvEAAaEAh8A2hqZFAPdWYABQ8EeJRKbQgRNgi3VUowAGs4GABlfGKRKElkGFUAVzhIgAsYdHIAdTaReGpqsDgKjP6RbX7MZHBaAtrqPRgUosl0SmROcUXIgCtqXgAhItLyisp0OAAZWrhN7daQfd4K2skeAEUmiHh78h2QNVVItVDNKHUmlRQZF8KpajBunkYG1kAAOAAMMPUi26RXwM1hcBBugKqjgAEcmiIyStKGsNkgtv9Hot0CIGs0WW9Pt9foyHpEYJRXIi2MjkAAWQVFETkN4AYQg6HWMygRAKTUWABVhZkmYsAL4GoA)

It would be nice to have a flag to set it such that the output of the above would be:
```html
<my-component
    class="wcd-flex-1 dialog-contents wcd-scroll-vertical"
	[foo]="bar">
</my-component>
```

karptonite:
## Description
This is an incomplete first pass at fixing #5377.

Essentially, it makes the `jsxBracketSameLine` option also apply in html files. Separately, this option should be renamed to reflect the new functionality; I suggest that `angleBracketSameLine` is a sensible name.

As you can see in the tests, when `jsxBracketSameLine` is set, this will convert:
```html
<article
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars">...</article>
<span
id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars">...</span>
```
to 
```html
<article
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars">
  ...
</article>
<span
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars"
  >...</span
>
```
Without that setting, this would look like:
```html
<article
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars"
>
  ...
</article>
<span
  id="electriccars"
  data-columns="3"
  data-index-number="12314"
  data-parent="cars"
  >...</span
>
```

Note that because of the whitespace sensitivity of a `span`, the span is formatted the same way for each option.
Also note that nothing is done with the final `>` in the closing `span` tag. This isn't something that comes up with jsx, as I understand it, because of how jsx handles whitespace. I'm not sure what the desirable behavior should be here, but if the setting is called `angleBracketSameLine`, perhaps that should be brought up to the previous line as well. But I'm not sure how to accomplish that, if that is the desired behavior.

Thanks to @adamJLev for pointing me in the right direction regarding this fix.

While I have added a test for the above cases, I have no idea what additional test might be desirable. I'm happy to work with someone from the Prettier team if more tests are needed.

I also haven't touched the docs yet. I assume that would come with a change to the name of the new option, or possibly renaming the old option to reflect both options, but that seems like a much larger job.

## Checklist

### DIFF
--- a/tests/html/last_line/jsfmt.spec.js
+++ b/tests/html/last_line/jsfmt.spec.js
@@ -1,0 +1,2 @@
+run_spec(__dirname, ["html"], { jsxBracketSameLine: true });
+run_spec(__dirname, ["html"], { jsxBracketSameLine: false });
--- a/src/language-html/printer-html.js
+++ b/src/language-html/printer-html.js
@@ -40,6 +40,7 @@ const {
   isVueSfcBindingsAttribute,
   isScriptLikeTag,
   isTextLikeNode,
+  isWhitespaceSensitiveNode,
   preferHardlineAsLeadingSpaces,
   shouldNotPrintClosingTag,
   shouldPreserveContent,
@@ -728,7 +729,13 @@ function printAttributes(path, options, print) {
   ) {
     parts.push(node.isSelfClosing ? " " : "");
   } else {
-    parts.push(node.isSelfClosing ? line : softline);
+    const sensitiveEnd = node.isSelfClosing ? line : softline;
+    const insensitiveEnd = node.isSelfClosing ? " " : "";
+    parts.push(
+      options.jsxBracketSameLine && !isWhitespaceSensitiveNode(node)
+        ? insensitiveEnd
+        : sensitiveEnd
+    );
   }
 
   return concat(parts);
--- a/tests/html/last_line/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/html/last_line/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,76 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`last_line.html - {"jsxBracketSameLine":false} format 1`] = `
+====================================options=====================================
+jsxBracketSameLine: false
+parsers: ["html"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+<article
+  id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars">...</article>
+<span
+id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars">...</span>
+
+=====================================output=====================================
+<article
+  id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars"
+>
+  ...
+</article>
+<span
+  id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars"
+  >...</span
+>
+
+================================================================================
+`;
+
+exports[`last_line.html - {"jsxBracketSameLine":true} format 1`] = `
+====================================options=====================================
+jsxBracketSameLine: true
+parsers: ["html"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+<article
+  id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars">...</article>
+<span
+id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars">...</span>
+
+=====================================output=====================================
+<article
+  id="electriccars"
+  data-columns="3"
+  data-index-number="12314"
+  data-parent="cars">
+  ...
+</article>
+<span
+  id="electriccars"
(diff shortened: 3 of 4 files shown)


tests/html/last_line/jsfmt.spec.js
```diff
@@ -0,0 +1,2 @@
+run_spec(__dirname, ["html"], { jsxBracketSameLine: true });
+run_spec(__dirname, ["html"], { jsxBracketSameLine: false });
```
alexander-akait, line 2: It should be `bracketSameLine` or other name, not `jsxBracketSameLine`, because `jsx` is not `html`, I would advise to do:

First PR.

- Added the new option
- if developer doesn't set `jsxBracketSameLine` we will try value of `bracketSameLine`, if still nothing, use default value

Second PR.

- Improve this for HTML + tests

Third PR.

- Improve this for VUE and etc

karptonite: Thanks, I addressed the naming issue in my post above. The eventual goal is that there is only one option, right? with `jsxBracketSameLine` being either a synonym of `angleBracketSameLine` (or whatever it is called) or deprecated.

alexander-akait: /cc @prettier/core

fisker: We  need rename`jsxBracketSameLine` first. Then apply to  html .

karptonite: Let me look into a different pull request that starts with a rename.

karptonite: Would a pull request for a rename include renaming in all the tests and docs? 
also, @alexander-akait , you wrote:
> if developer doesn't set jsxBracketSameLine we will try value of bracketSameLine, if still nothing, use default value
I think we could simplify that to "If a developer sets either of the settings to true, we print the bracket on the same line." The default is false, and we don't want to inadvertently create a new option here by treating jsxBracketSameLine differently from angleBracketSameLine. We are renaming an old option, not adding a new one.

dummdidumm:
> Note that because of the whitespace sensitivity of a `span`, the span is formatted the same way for each option.
> Also note that nothing is done with the final `>` in the closing `span` tag. This isn't something that comes up with jsx, as I understand it, because of how jsx handles whitespace. I'm not sure what the desirable behavior should be here, but if the setting is called `angleBracketSameLine`, perhaps that should be brought up to the previous line as well. But I'm not sure how to accomplish that, if that is the desired behavior.

If it's done for the `>` of the closing tag aswell, I think it should only be done if there's whitespace after the `>`. Else how would one properly format this?

```html
<span><span>looooooooooong</span>looooooooooonglooooooooooonglooooooooooonglooooooooooonglooooooooooong<span><span>hi</span><span>hi</span></span></span>
```

alexander-akait:
https://github.com/prettier/prettier/pull/11006

PR #9949, opened 2020-12-22
Refactor: faster levenshtein

ka-weihe:
## Description
This PR changes the levensthein distance package from `leven` to `fastest-levenshtein`. [leven](https://github.com/sindresorhus/leven) is one of the slowest levenshtein packages out there and the author is trying to deceive people with shitty benchmarks where he is only comparing it slower packages. Please consider using  [fastest-levenshtein]( https://github.com/ka-weihe/fastest-levenshtein) instead which is 10x faster and much higher quality. 

## Checklist


fisker:
👎 for this, 

1. I have no doubt this lib is the fastest public package, but I don't like something called "fastest", how do you know there will no faster one?
2. > the author is trying to deceive people with shitty benchmarks where he is only comparing it slower packages

	That's rude to say that as the owner of 'fastest-levenshtein'.

Similar project `fast-cartesian`, it claim it's fastest, when I [question it](https://github.com/ehmicky/fast-cartesian/issues/2), he just copied my faster algorithm... WTF!

ka-weihe:
> I have no doubt this lib is the fastest public package, but I don't like something called "fastest", how do you know there will no faster one?

@fisker I will not claim that my package is the fastest one unless it actually is. Don't worry, I will make it very clear that the package is not the fastest, if I find out that it is not.

> the author is trying to deceive people with shitty benchmarks where he is only comparing it slower packages.

It is absolutely not rude. His benchmarks are a special case where his implementation skips the main part of the algorithm completely - you can have a look yourself. That is the same as benchmarking a big heavy function with a special case where it returns from an if-statement right in the beginning. That is just deceiving. He is also only comparing it to slower packages, and I have opened a PR to update his benchmarks by adding other packages, but he ignores... probably because it is much slower. 

> Similar project fast-cartesian, it claim it's fastest, when I question it, he just copied my faster algorithm... WTF!

Not okay of him to copy your code, but lets imagine he did not. Would you prefer that he kept saying that his package was the fastest knowing that your package is faster? ... and would you would you prefer that people kept using his package instead of yours, because he was lying?

fisker:
>  would you prefer that people kept using his package instead of yours, despite him lying?

Imagine `leven` package is that package, imagine that "people" is "babel", he already give you the answer.

ka-weihe:
Yes? Babel changed away from leven?

fisker:
I mean the answer is NO.

ka-weihe:
No to what? I'm guessing no to `"would you prefer that he kept claiming that his package is the fastest knowing that your package is faster?"`, but not sure. In that case, I think we agree

fisker:
Yes, answer to that, I quoted it above my comment.

fisker:
Can you open another PR for this, stop saying how `leven` is slow, just say how good is your package.

I think this is the right way to do.

I'm strongly against the `fastest` term in the package name, I won't approve it, but if other maintainers okay will it, I won't block it.

This one is not going to get approved.

ka-weihe:
Okay... but I don't get what is wrong with saying leven is slow, when it is. It is just facts.

fisker:
If you don't get it, I have to say we are different people.

ka-weihe:
You do you. https://github.com/prettier/prettier/pull/9951

PR #9993, opened 2021-01-04
Refactor: Use `Doc[]` instead of `concat()`

fisker:
## Description

Ref: #9733

## Checklist


src/language-css/printer-postcss.js
```diff
@@ -203,47 +198,45 @@ function genericPrint(path, options, print) {
...
-          ]);
+          ];
         }
 
         if (node.function) {
-          return concat([
+          return [
             node.name,
-            concat([path.call(print, "params")]),
+            [path.call(print, "params")],
```
fisker, line 211: This one is a sigle `Doc`, should not use `concat`.

I'm going to fix this first, it's easier to detect before we change `concat()` to `array`.

fisker: #9994

sosukesuzuki (approved):
Great job

PR #10001, opened 2021-01-04
🎉🎉 💯✖💯➕1️⃣ 🎉🎉

fisker:
🎉🎉 💯✖💯➕1️⃣ 🎉🎉


src/options-argument-index.evaluate.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+// Advanced math expression evaluate algorithm invented by @prettier in 2021
+// Steps:
+//  1. Convert math expression to `boolean`
+//  2. Convert `boolean` result back to `number`
+//  3. Get the result of the math expression
+const evaluate = (expression) => Number(Boolean(expression));
```
fisker, line 8: I don't want add test for this, because I'm sure there is no bug.

fisker:
Abandon

PR #10028, opened 2021-01-08
Update subtitle for version 2.0 release

idunn01:
I'm 9 months late, but I feel like like this was a missed opportunity. Or maybe for version 3.0?

## Description

Update the version 2.0 release post to be titled `Prettier 2.0 “Even Prettier”`


fisker:
`v2.0` has been released for a long time, I don't think we can do this, sorry.

PR #10059, opened 2021-01-15
Improve handling of dangling comments in ForStatement

### LINKED ISSUE #10046: Improve comments inside `for` loop
**Prettier pr-9857**
[Playground link](https://deploy-preview-9857--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAzCAnAFAegFQCGeOA3PkaecQJQkA6U62VOBZhx7FtDTuHpAiy416jTPwojBLWiAA0ICAAcYAS2gBnZKAIYMEAO4AFPQm0oCAG0MEAntsUAjDATABrODADKyt2qgAc2QYDABXOEU4AFsnOAATeISAGQIgsIJAuAAxTGiCGHUg5BACMJgIBRAACxhoqwB1arV4TT8wOG9zFrUANxa7ErBNRxAAzTgMGGNXQPzkVGsJxQArTQAPACFXDy9vAmi4ZIC4BaXIkDX17wDAqzgARTCIeDOrZZA-DAmMEqcCOJWKrKDABGANNTxGDVZAADgADIoQRAJg1XMoSiC4D9eqdFABHZ7wGYqCylTQAWigcASCSqGDghLUDJmmXmSEW7wuE2iahC4W5t3uTxepw550UMABEKhMKQACZJa41FZbgBhCDRdkgbEAViqYQmABUARZOR9ehEAJJQJKwbxgUGqACCtu8MDs9zeEwAvj6gA)
```sh
--parser babel
```

**Input:**
```jsx
for(/*a*/;/*a*/;/*a*/);
for(/*a*/a;/*a*/;/*a*/);
for(/*a*/;a/*a*/;/*a*/);
for(/*a*/;/*a*/;a/*a*/);
```

**Output:**
```jsx
/*a*/
/*a*/
/*a*/
for (;;);
for (/*a*/ a /*a*/ /*a*/; ; );
for (; /*a*/ a /*a*/ /*a*/; );
for (; ; /*a*/ /*a*/ a /*a*/);

```

**Expected behavior:**

Comments should stay at the original position.

We could use a solution like #9857

thorn0:
## Description

Fixes #10046

A better handling for comments like this: `for (/*a*/; /*b*/; /*c*/) ...`

A proof of concept of an approach that doesn't modify the AST unlike an alternative approach to a similar problem in #9857.

## Checklist


fisker:
Seems a good solution.

tests/js/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1275,12 +1275,133 @@ for // comment
...
+
+for (;;
+// ...
+);
+
 =====================================output=====================================
-// comment
-for (;;);
+for (
+  ;// comment
```
thorn0, line 1338: Not good...

src/language-js/comments.js
```diff
@@ -929,6 +929,49 @@ function getCommentChildNodes(node, options) {
...
+    }
+    if (text.charAt(nextIdx) === ";") {
+      return nextIdx;
+    }
+    nextIdx++;
+  }
+}
+
+function createCommentAnchorPseudoNode(ref, marker, loc) {
+  return { type: "Prettier:CommentAnchor", ref, marker, range: [loc, loc] };
```
thorn0, line 974: Oops! `range` is a JS-specific thing...

src/language-js/comments.js
```diff
@@ -929,6 +929,49 @@ function getCommentChildNodes(node, options) {
...
```
fisker, line 974: Can we use `object` for this `marker`? I remeber you already use if for something in class, it should be more clear to understand what it is.

thorn0: It's a string label on a dangling comment needed to differentiate it from other dangling comments and to associate it with a specific place inside the enclosing node. Are you proposing not to use strings for this? I think strings work pretty well here: e.g. if the `init` part of a `for` statement is omitted, we might have dangling comments with the `init` marker there.

>  I remeber you already use if for something in class

For the same thing: to differentiate comments related to `extends` from comments related to `implements`, etc.

thorn0: I'm thinking about renaming `marker` to `anchor` as it seems to convey the purpose better, but let's do that in another PR.

thorn0: I did some renames, see the last commit.

fisker (approved):
Looks  good,  but   I would prefer to use tokens.

thorn0:
@fisker I like tokens too as they're more structured then the text-scanning hacks, but on the other hand depending only on 2 inputs (AST and original text) is better than on 3 (AST, original text, tokens). The resulting code is a bit more universal.

src/language-js/comments.js
```diff
@@ -929,6 +929,56 @@ function getCommentChildNodes(node, options) {
...
+    if (text.charAt(nextIdx) === char) {
+      return nextIdx;
+    }
+    nextIdx++;
+  }
+}
+
+function createCommentAnchorPseudoNode(enclosingNode, name, loc) {
+  return {
+    type: "Prettier:CommentAnchor",
```
fisker, line 977: What do you think  `::CommentAnchor`, like a  css pseudo element. 😄

thorn0: I like it :+1:

PR #10064, opened 2021-01-16
Remove incorrect statement about ESLint plugin speed

hubgit:
Remove the sentence in the docs which claims that running Prettier directly is faster than running it as an ESLint plugin.

## Description

* In this repository, running `yarn eslint src --ext=.js` takes 3.4s and running `yarn prettier 'src/**/*.js' --check` takes 2.7s, so in total they take 6.1s.
* In this repository, running `yarn eslint src --ext=.js` after adding the `prettier` ESLint rules (by adding `plugin:prettier/recommended` into the `extends` list in `.eslintrc.yml`) takes 5.6s, which is faster than running the tools independently.

Therefore, it's incorrect to say that running the Prettier rules in ESLint is slower than running Prettier directly.

### DIFF
--- a/docs/integrating-with-linters.md
+++ b/docs/integrating-with-linters.md
@@ -28,7 +28,6 @@ These plugins were especially useful when Prettier was new. By running Prettier 
 The downsides of those plugins are:
 
 - You end up with a lot of red squiggly lines in your editor, which gets annoying. Prettier is supposed to make you forget about formatting – and not be in your face about it!
-- They are slower than running Prettier directly.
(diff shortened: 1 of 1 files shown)


thorn0:
What you did is not the only way to compare speed. E.g., my experience is that on saving big files in VS Code, Prettier formatting called via ESLint autofix is much slower than Prettier called directly via the Prettier extension.

hubgit:
It does seem that - unlike Prettier run by itself - when a large file contains lots of errors ESLint with the prettier plugin gets slower with each error.

Issue #4210, opened 2018-03-27
No line breaks for very long arrays?


cangoektas:
Just wanted to kick off a discussion about how Prettier prints long arrays. 😊 I think in some cases it might make sense to have Prettier format a little differently.

Let's take the following as an example:

```js
const userIds = [
  941,
  722,
  312,
  933,
  841,
  166,
  843,
  978,
  436,
  687,
  269,
  573,
  913,
  873,
  748,
  354,
  153,
  181,
  947,
  563,
  324,
  952,
  885,
  911,
  698,
  248,
  764,
  431,
  811,
  344,
  855,
  430,
  396,
  981,
  594,
  131,
  489,
  552,
  116,
  833,
  521,
  252,
  503,
  540,
  893,
  736,
  969,
  145,
  925,
  417,
  344,
  108,
  161,
  777,
  971,
  159,
  158,
  908,
  511,
  876,
  768,
  284,
  640,
  309,
  651,
  292,
  898,
  284,
  201,
  506,
  654,
  840,
  334,
  246,
  376,
  398,
  714,
  134,
  435,
  181,
  980,
  594,
  396,
  100,
  743,
  812,
  583,
  622,
  800,
  310,
  111,
  537,
  751,
  920,
  872,
  700,
  702,
  655,
  515,
  298
];
```

Lately, I find myself inserting `// prettier-ignore` above arrays like this and adjusting the formatting manually to something like this:

```js
// prettier-ignore
const userIds = [
  941, 722, 312, 933, 841, 166, 843, 978, 436, 687, 269, 573, 913, 873, 748,
  354, 153, 181, 947, 563, 324, 952, 885, 911, 698, 248, 764, 431, 811, 344,
  855, 430, 396, 981, 594, 131, 489, 552, 116, 833, 521, 252, 503, 540, 893,
  736, 969, 145, 925, 417, 344, 108, 161, 777, 971, 159, 158, 908, 511, 876,
  768, 284, 640, 309, 651, 292, 898, 284, 201, 506, 654, 840, 334, 246, 376,
  398, 714, 134, 435, 181, 980, 594, 396, 100, 743, 812, 583, 622, 800, 310,
  111, 537, 751, 920, 872, 700, 702, 655, 515, 298
];
```

I think that in cases like this, it doesn't really help to have each item on a new line. It just makes the code harder to scroll through and harder to read. Line breaks definitely make sense for shorter arrays, but there could be some sort of a thresh-hold after which we could format differently.

[...]

PR #10107, opened 2021-01-22
RFC: use fill to print arrays with numbers & short strings

brodybits:
## Description

__updated__

Based on PR #10106, implements <https://github.com/prettier/prettier/issues/4210#issuecomment-376500821> for both numbers and strings up to 8 characters.

__This includes some quick testing.__

I hope this can help @breck7 as requested in <https://github.com/prettier/prettier/issues/4210#issuecomment-765584985>.

TODO:

- [x] update affected test snapshots
- [ ] resolve issue with cursorOffset & full-width characters in `tests_integration/__tests__/cursor-offset.js`
- [ ] add changelog

## Checklist


src/language-js/print/array.js
```diff
@@ -75,11 +78,25 @@ function printArray(path, options, print) {
         return element[itemsKey] && element[itemsKey].length > 1;
       });
 
+    const shouldUseFill =
+      n.elements.length > 1 &&
+      n.elements.every(
+        (element) =>
+          element &&
+          (isNumericLiteral(element) ||
+            (isStringLiteral(element) && element.value.length < 9))
```
fisker, line 87: Should  use  `raw` instead of `value`

fisker: And 8 is  not  enough for color `#RRGGBBAA`

brodybits: Using `raw` caused an error with `espree`. And this change did not work on `estree` or `meriyah`:

```suggestion
            (isStringLiteral(element) && (element.raw || element.value).length < 9))
```

It also did not resolve my issue with cursorOffset & full-width characters in `tests_integration/__tests__/cursor-offset.js`.

fisker: I mean https://github.com/prettier/prettier/blob/33078c61eb7388afa80a5057968cd656f7b7f0c8/src/language-js/utils.js#L968

fisker:
Why did you close this?

brodybits:
I felt as if I was going in a different direction from PR #10106, which seems to be the "accepted solution" ref: https://github.com/prettier/prettier/issues/4210#issuecomment-765670858

No problem if you think we should reopen this, but I think it should be considered in followup to PR #10106.

PR #10112, opened 2021-01-23
Enable `unicorn/no-array-for-each`

fisker:
## Checklist


src/language-markdown/utils.js
```diff
@@ -50,76 +50,71 @@ function splitText(text, options) {
...
-                }
-          );
-        });
-    });
+      );
+    }
+  }
 
+  // @ts-ignore
   return nodes;
```
fisker, line 118: I don't know how to fix this

```text
src/language-markdown/utils.js:117:3 - error TS2322: Type '{ type: string; value: string; }[]' is not assignable to type '({ type: "whitespace"; value: "" | "\n" | " "; } | { type: "word"; value: string; })[]'.
  Type '{ type: string; value: string; }' is not assignable to type '{ type: "whitespace"; value: "" | "\n" | " "; } | { type: "word"; value: string; }'.
    Type '{ type: string; value: string; }' is not assignable to type '{ type: "word"; value: string; }'.
      Types of property 'type' are incompatible.
        Type 'string' is not assignable to type '"word"'.

117   return nodes;
```

sosukesuzuki (approved):

thorn0 (approved):
I'm not a huge fan of replacing one-liners (`arr.forEach(fn)`) with multiple lines, but also I'm not a fan of arguing about such things.

PR #10122, opened 2021-01-23
Fix lineSuffixBoundary

thorn0:
There is a bug in the implementation of the `lineSuffixBoundary` command: the [`fits` function][fits] in the doc printer doesn't consider `lineSuffixBoundary` to be a potential line break. That's why `lineSuffixBoundary` doesn't break groups, the line length limit is calculated incorrectly after a line break caused by `lineSuffixBoundary`, etc. This PR fixes this bug.

[fits]: https://github.com/prettier/prettier/blob/dd606d9ce1810f970c57e09f2f8cde5adcfbdb01/src/document/doc-printer.js#L147

I couldn't find any open issues that this PR would directly fix. This is because `lineSuffixBoundary` is very little used. It hasn't really been useful because of this bug. I got ugly results every time I tried to use it. But now it should become a great help in solving comments-related issues.

For example, consider this doc:
```js
    group([
      "let foo = [",
      indent([
        softline,
        [lineSuffixBoundary, "item1,"],
        line,
        [lineSuffixBoundary, "item2,", lineSuffix(" // comment")],
        line,
        [lineSuffixBoundary, "item3"],
      ]),
      softline,
      "];",
    ])
```
That's how it is printed now:
```js
let foo = [item1, item2,  // comment
  item3];
```
And with the fix:
```js
let foo = [
  item1,
  item2, // comment
  item3
];
```

## Checklist

### DIFF
--- a/changelog_unreleased/api/10122.md
+++ b/changelog_unreleased/api/10122.md
@@ -1,0 +1,34 @@
+#### Fix `lineSuffixBoundary` (#10122 by @thorn0)
+
+There was a bug in the implementation of the [`lineSuffixBoundary`][lsb] command that significantly limited its usefulness: the printer algorithm didn't correctly consider it a potential line break. Now that the bug has been fixed, we urge plugin authors to give this command another try and see if it can help them simplify printing of trailing comments.
+
+[lsb]: https://github.com/prettier/prettier/blob/main/commands.md#linesuffixboundary
+
+<!-- prettier-ignore -->
+```jsx
+// Input
+group([
+  "let foo = [",
+  indent([
+    softline,
+    [lineSuffixBoundary, "item1,"],
+    line,
+    [lineSuffixBoundary, "item2,", lineSuffix(" // comment")],
+    line,
+    [lineSuffixBoundary, "item3"],
+  ]),
+  softline,
+  "];",
+])
+
+// Prettier stable
+let foo = [item1, item2,  // comment
+  item3];
+
+// Prettier main
+let foo = [
+  item1,
+  item2, // comment
+  item3
+];
+```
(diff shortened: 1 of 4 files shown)


changelog_unreleased/api/10122.md
```diff
@@ -0,0 +1,34 @@
+#### Fix `lineSuffixBoundary` (#10122 by @thorn0)
+
+There was a bug in the implementation of the [`lineSuffixBoundary`][lsb] command that significantly limited its usefulness: the printer algorithm didn't correctly consider it a potential line break. Now that the bug has been fixed, we urge plugin authors to give this command another try and see if it can help them simplify printing of trailing comments.
```
fisker, line 3: Why not `` [`lineSuffixBoundary`][https://github.com/prettier/prettier/blob/main/commands.md#linesuffixboundary] ``

thorn0: Just for readability/editability of non-rendered Markdown

tests_integration/__tests__/line-suffix-boundary.js
```diff
@@ -0,0 +1,41 @@
...
+const {
+  group,
+  indent,
+  line,
+  lineSuffix,
+  lineSuffixBoundary,
+  softline,
+} = prettier.doc.builders;
+
+const printDoc = require("../printDoc");
```
fisker, line 15: We can test this with `prettier.__debug.printDocToString`

thorn0: It's not the same. E.g., `propagateBreaks` isn't called in `__debug.printDocToString`. Even though `propagateBreaks` isn't strictly needed for this specific test, in general it's safer to test docs by printing them the exact way it really happens in production. I hope we'll add more tests for doc printing.

fisker (approved):

alexander-akait (approved):

PR #10124, opened 2021-01-24
Add CLI flag for debugging comment attachment issues

thorn0:
A new `--debug-print-comments` CLI flag and corresponding functionality for the Playground.

## Checklist

### DIFF
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -557,6 +557,7 @@ function addCommentHelper(node, comment) {
   const comments = node.comments || (node.comments = []);
   comments.push(comment);
   comment.printed = false;
+  comment.nodeDescription = describeNodeForDebugging(node);
 }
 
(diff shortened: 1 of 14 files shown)


src/common/util.js
```diff
@@ -561,6 +561,7 @@ function addCommentHelper(node, comment) {
   const comments = node.comments || (node.comments = []);
   comments.push(comment);
   comment.printed = false;
+  comment.nodeDescription = describeNodeForDebugging(node);
```
fisker, line 564: Do you think it better be a getter?

thorn0: It's difficult for me to evaluate the impact and understand which way is better. Strings at least don't hold references to the node.

fisker (approved):

sosukesuzuki (approved):
Thanks!

PR #10125, opened 2021-01-24
Minor spellchecking and typechecking fixes

thorn0:


src/language-js/utils.js
```diff
@@ -1250,27 +1250,24 @@ function hasIgnoreComment(path) {
...
+  /** Check comment is the first attached comment */
   First: 1 << 7,
-  /** @type {number} Check comment is the last attched comment */
+  /** Check comment is the last attached comment */
   Last: 1 << 8,
 };
 
-/**
- * @returns {function}
- */
```
thorn0, line 1273: This kind of type annotation is redundant. TS infers a better type without it.

cspell.json
```diff
@@ -411,17 +416,16 @@
...
         "Zosel"
     ],
     "ignoreRegExpList": [
         "<!-- prettier-ignore -->\\n(`{3,})\\w*\\n[\\s\\S]+?\\1",
         "\\[(\\*{2})?@[-\\w]+?\\1\\]",
         "by @[-\\w]+(?:, @[-\\w]+)?",
-        "\\[`\\w+`\\]",
         "ve{2,}r{2,}y",
         "ve+r+y+long\\w*",
-        "\\/.*?\\/",
```
thorn0, line 424: Because of this pattern, cspell ignored `/* comments */`.

src/language-js/utils.js
```diff
@@ -1304,11 +1301,7 @@ function hasComment(node, flags, fn) {
     return false;
   }
   const test = getCommentTestFunction(flags, fn);
-  return test
-    ? node.comments.some((comment, index, comments) =>
-        test(comment, index, comments)
-      )
-    : true;
+  return test ? node.comments.some(test) : true;
```
fisker, line 1304: Can we revert this change? https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-callback-reference.md

thorn0: I don't approve that rule at all. The issue that it's supposed to prevent happens too seldom. With TS checks, even if the type coverage is as poor as we have it in Prettier, it happens even more seldom. But the added verbosity is ugly.

fisker (approved):

PR #10126, opened 2021-01-24
Document that Prettier never unquotes numeric properties in TS and Flow

### LINKED ISSUE #10123: Number properties can't be unquoted in TypeScript
Prettier usually removes unnecessary quotes from properties in TypeScript and JavaScript. However, when a numeric property is used in TypeScript, they keep their original quoting.

I tried to see if the quotes make TypeScript behave differently using TypeScript 4.1.3, but it does not.

In the linked playground, the diffence can be observed when toggling the parser between babel and typescript.

**Prettier 2.2.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLWcBOAzAhmOAAgDEIJDgAdKQwgcgBYAmOpQgZxk3QHMBuagF9q1SFE6FsZQgF4K1WoxZs6dIQKggANCAgAHGKmjtkoXJkwQA7gAVzCEylwAbK7gCeJnQCNM+ANZwMADKevi8yFwArnA6cAC23nAAJskpADK4UDxRuDxwpJjxuDCG2cgguFEwENogABYw8c4A6vWo8OxhBMEOHagAbh3uFWDsXmjiWDA2fjzFyHjO7LEgAFbsAB4AQn5ggSG48XDp6HCLLis6G5vBvM5wAIpREPAXy6thmCuYFTDuejg7DA3AMdT03FgLVQyRg9WQAA4AAw6CEQFYtPx6CoQoFYAbnHQARxe8Fm+kclXYAFooHAUik6pg4CTUMzZnkFkgllcQCt4qhIpgYjp2PcnqTztzLqsYLhvNDYfCkEwdFxcKhnLwAMIQeJckBAgCsdSiKwAKvLHDzVgMYgBJKBpWDBEGoAwAQSdwX+D3eK0EgiAA)
```sh
--parser typescript
```

**Input:**
```tsx
interface Foo {
  '42': string;
}

const foo = {
  '42': ''
};

```

**Output:**
```tsx
interface Foo {
  "42": string;
}

const foo = {
  "42": "",
};

```

**Expected behavior:**
```tsx
interface Foo {
  42: string;
}

const foo = {
  42: "",
};

```

thorn0:
## Description

Fixes #10123

## Checklist

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -89,7 +89,12 @@ Valid options:
 | ------------- | -------------------------------------------------------------------- | -------------------------------------------------------------------- |
 | `"as-needed"` | <code>--quote-props <as-needed&#124;consistent&#124;preserve></code> | <code>quoteProps: "<as-needed&#124;consistent&#124;preserve>"</code> |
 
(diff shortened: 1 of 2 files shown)


brodybits:
Prettier also never unquotes numeric properties for Angular or Vue, see file comments starting here: https://github.com/prettier/prettier/pull/9000#discussion_r497723764

thorn0:
I added Angular. As for Vue, if I'm not mistaken, it's actually safe to unquote, but before fixing that it'd be good to confirm that for sure.

fisker (approved):
Let's keep it quoted for vue, unless someone confirm it can be unquoted.

PR #10129, opened 2021-01-25
Enable `unicorn/prefer-spread`

fisker:
## Description

Enfore use `...` instead of `Array#concat()`, in some place, it's very unclear the argument is an array or not.
This also prevent unexpected spreading the new `concat` doc (array).

## Checklist

### DIFF
--- a/src/config/resolve-config.js
+++ b/src/config/resolve-config.js
@@ -155,8 +155,8 @@ function mergeOverrides(configResult, filePath) {
 
 // Based on eslint: https://github.com/eslint/eslint/blob/master/lib/config/config-ops.js
 function pathMatchesGlobs(filePath, patterns, excludedPatterns) {
-  const patternList = [].concat(patterns);
-  const excludedPatternList = [].concat(excludedPatterns || []);
+  const patternList = [...patterns];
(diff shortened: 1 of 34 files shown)


src/config/resolve-config.js
```diff
@@ -154,9 +154,11 @@ function mergeOverrides(configResult, filePath) {
...
 
 // Based on eslint: https://github.com/eslint/eslint/blob/master/lib/config/config-ops.js
-function pathMatchesGlobs(filePath, patterns, excludedPatterns) {
-  const patternList = [].concat(patterns);
-  const excludedPatternList = [].concat(excludedPatterns || []);
+function pathMatchesGlobs(filePath, patterns, excludedPatterns = []) {
+  const patternList = Array.isArray(patterns) ? patterns : [patterns];
+  const excludedPatternList = Array.isArray(excludedPatterns)
+    ? excludedPatterns
+    : [excludedPatterns];
```
fisker, line 161: According to doc https://prettier.io/docs/en/configuration.html#configuration-overrides, these two all `string | string[]`

thorn0 (approved):

PR #10135, opened 2021-01-25
Use default parameters where possible and other microrefactoring

thorn0:
## Description

Just refactoring.

## Checklist

### DIFF
--- a/src/language-markdown/printer-markdown.js
+++ b/src/language-markdown/printer-markdown.js
@@ -503,9 +503,7 @@ function getNthListSiblingIndex(node, parentNode) {
   );
 }
 
-function getNthSiblingIndex(node, parentNode, condition) {
-  condition = condition || (() => true);
-
+function getNthSiblingIndex(node, parentNode, condition = (_) => true) {
   let index = -1;
 
   for (const childNode of parentNode.children) {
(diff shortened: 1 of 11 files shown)


fisker:
[`prefer-default-parameters`](
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-default-parameters.md)
, this rule maybe not very safe , but  can help .

src/language-markdown/printer-markdown.js
```diff
@@ -503,9 +503,7 @@ function getNthListSiblingIndex(node, parentNode) {
   );
 }
 
-function getNthSiblingIndex(node, parentNode, condition) {
-  condition = condition || (() => true);
-
+function getNthSiblingIndex(node, parentNode, condition = (_) => true) {
```
fisker, line 506: This default value not needed. Only used in this file, and it passed the argument.

src/language-markdown/utils.js
```diff
@@ -214,7 +212,7 @@ function mapAst(ast, handler) {
     }
 
     return newNode;
-  })(ast, null, null);
+  })(ast, null);
```
fisker, line 215: Let's pass an empty array?

tests_integration/runPrettier.js
```diff
@@ -115,9 +115,7 @@ function runPrettier(dir, args = [], options = {}) {
 
   const result = { status, stdout, stderr, write };
 
-  const testResult = (testOptions) => {
-    testOptions = testOptions || {};
-
+  const testResult = (testOptions = {}) => {
```
fisker, line 118: Can this be removed? I didn't see any test without it.

fisker (approved):
Left few comments.

sosukesuzuki (approved):

PR #10142, opened 2021-01-26
website: bigger font for <h5> in blog

thorn0:
## Description

Now it doesn't look like a heading at all:

> image

## Checklist

### DIFF
--- a/website/static/overrides.css
+++ b/website/static/overrides.css
@@ -513,6 +513,10 @@ a {
   margin-top: 10px;
 }
 
+.mainContainer .wrapper .post h5 {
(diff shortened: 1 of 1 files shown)


fisker:
Isn't `<h5>` `<h6>` normally just `1em`?

thorn0:
There are some other styles there that change that.

fisker:
I suggested to remove categories in highlights before last version release, so we can use `<h4>`

thorn0:
And I think I supported keeping them as is. Besides, `h5` can be used in non-highlights too.

fisker (approved):

PR #7865, opened 2020-03-24
Smarter formatting of HTML class attribute

thorn0:
Break classes onto multiple lines, but try keeping consecutive classes with the same prefix on one line.

fixes #7863 

- [x] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)
- [x] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/pr-XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://deploy-preview-7865--prettier.netlify.com/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIwA2AhgM4kC8AOiAAoBOE2m6cdAsnKQK51wkDC0GEUxRWAfXEALTligBzALQMA7pSi5c6RQDMCcAB7rNew4sgEuAWw2mDigktXHcAKy4kYmHQE9zQhBhFTyI6GDcPL19-WECHJQAjOBgVOAQXIgJMeShFTHgrEmDhMNxM7Nz8uEL480DWagA+ADEICFQAeixsRpAAGhAIAAcvaBJkUFDVGlCEcZRMlSIfcYGEuiIwAGtkgGUhzdF5ZBg6LjgBqRgrAgB1KSqSA7A4Xbn8nHyfZHAyfpBRCRWDB6ER5FYiMgdJkgQNXCQDAAhDbbPZEKxwAAyojgUJhFxA8IMuyO+gAilwIPA8QRYSADnQgXQflcbv8hnRRDBbswYFJkAAOAAMAw5ECBtw2Qx+HL4rGwuIGAEdKfB6MN5iBSIoxHAWOh-rwVZheKDwZCkNDaQSgVZMCczjbSXAKVTcZb8QNhAkeeg+cgAExejaYLIKQRWCE-PgAVn+HjgABUiAl5la6dhzgBJKAsWC7MCckYAQVzuxgPn0NKBAF8a0A)✨**

PR #10145, opened 2021-01-26
Handlebars: fix formatting of attributes

### LINKED ISSUE #8648: Handlebars: newline in an attribute incorrectly rendered as `[object Object]`

thorn0:
## Description

This PR
* fixes escaping of `{{` in attributes and text
* fixes the choice between `'` and `"` for attributes with interpolations
* fixes the bug with `[object Object]` printed in `class`
* implements simple formatting for `class`, like in HTML in the current stable version (before #7865)

Closes #8648
Closes #8667

cc @dcyriller @kangax 

## Checklist


src/language-handlebars/printer-glimmer.js
```diff
@@ -674,6 +685,11 @@ function generateHardlines(number = 0, max = 0) {
  * @param {object} options - the prettier options object
  */
 function printStringLiteral(stringLiteral, options) {
+  const { quote, regex } = chooseEnclosingQuote(options, stringLiteral);
+  return [quote, stringLiteral.replace(regex, `\\${quote}`), quote];
+}
+
+function chooseEnclosingQuote(options, stringLiteral) {
```
dcyriller, line 692: 👍

changelog_unreleased/handlebars/10145.md
```diff
@@ -0,0 +1,25 @@
+#### Fix formatting of attributes (#10145 by @thorn0)
+
+- fix escaping of `{{` in attributes and text
```
dcyriller, line 3: nice catch, I had this refactor / fix on a branch too

dcyriller (approved):
Left one small comment about `n.chars` that might be transformed into `text`. Sounds like a great improvement overall.

Do you think we should use `src/language-html/syntax-attribute.js` functions to format handlebars attributes in the future?

thorn0:
@dcyriller 

> Do you think we should use src/language-html/syntax-attribute.js functions to format handlebars attributes in the future?

What can be reused should definitely be reused, but we need to figure out a way to share that code between the two printers. Currently `language-html` and `language-handlebars` are built as separate bundles.

thorn0:
@dcyriller Just realized that I was wrong here:

> Currently language-html and language-handlebars are built as separate bundles

Only the parsers are extracted to separate bundles. The printers all go in one bundle, so it's okay to `require` things directly from `language-html` in the printer.

PR #10147, opened 2021-01-27
Add parallel-prettier to "Related Projects"

thorn0:
## Description

see https://github.com/microsoft/parallel-prettier

## Checklist

### DIFF
--- a/docs/related-projects.md
+++ b/docs/related-projects.md
@@ -28,6 +28,7 @@ title: Related Projects
 
 ## Misc
 
+- [parallel-prettier](https://github.com/microsoft/parallel-prettier) is an alternative CLI that formats files in parallel to speed up large projects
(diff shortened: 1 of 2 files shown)


sosukesuzuki (approved):

fisker:
I wounder why the project is under microsoft, but package in not.

fisker (approved):

fisker:
And I think we should support `parallel` in out CLI too!

connor4312:
just saw this as a linked issue from the original thread -- to answer your question

>  I wounder why the project is under microsoft, but package in not.

This was originally written during my time at [Mixer](https://en.wikipedia.org/wiki/Mixer_(service)), which had a separate github org and operated somewhat more independently than other MS businesses. I think we also made our npm org a long while before there was an official `@microsoft` one.

PR #10160, opened 2021-01-27
Fix regression after pr-10106

### LINKED ISSUE #10159: Investigate/debug how ifBreak with group id works inside group with that id
https://github.com/prettier/prettier/pull/10160#issuecomment-769469692

thorn0:
## Description

Fixes #10159

## Checklist

### DIFF
--- a/src/language-js/print/array.js
+++ b/src/language-js/print/array.js
@@ -104,9 +104,9 @@ function printArray(path, options, print) {
       ? ","
       : !shouldPrintComma(options)
       ? ""
-      : ifBreak(",", "", {
-          groupId: shouldUseConciseFormatting ? groupId : undefined,
-        });
+      : shouldUseConciseFormatting
+      ? ifBreak(",", "", groupId)
+      : ifBreak(",");
 
     parts.push(
       group(
--- a/src/language-js/print/array.js
+++ b/src/language-js/print/array.js
@@ -57,14 +57,6 @@ function printArray(path, options, print) {
 
     const groupId = Symbol("array");
 
-    const trailingComma = !canHaveTrailingComma
-      ? ""
-      : needsForcedTrailingComma
-      ? ","
-      : !shouldPrintComma(options)
-      ? ""
-      : ifBreak(",", "", { groupId });
-
     const shouldBreak =
       !options.__inJestEach &&
       n.elements.length > 1 &&
@@ -105,6 +97,16 @@ function printArray(path, options, print) {
               })
           )
       );
+
+    const trailingComma = !canHaveTrailingComma
+      ? ""
+      : needsForcedTrailingComma
+      ? ","
+      : !shouldPrintComma(options)
+      ? ""
+      : ifBreak(",", "", {
+          groupId: shouldUseConciseFormatting ? groupId : undefined,
+        });
 
     parts.push(
       group(
--- a/src/language-js/print/array.js
+++ b/src/language-js/print/array.js
@@ -105,7 +105,7 @@ function printArray(path, options, print) {
       : !shouldPrintComma(options)
       ? ""
       : shouldUseConciseFormatting
-      ? ifBreak(",", "", groupId)
+      ? ifBreak(",", "", { groupId })
       : ifBreak(",");
 
     parts.push(
--- a/changelog_unreleased/javascript/10106.md
+++ b/changelog_unreleased/javascript/10106.md
@@ -1,4 +1,4 @@
-#### Concise formatting of number-only arrays (#10106 by @thorn0)
+#### Concise formatting of number-only arrays (#10106, #10160 by @thorn0)
 
 While in general, Prettier avoids this kind of formatting because it's not diff-friendly, in this special case we decided that the benefits outweigh the risks.
 
--- a/tests/js/arrays/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/js/arrays/__snapshots__/jsfmt.spec.js.snap
@@ -41,12 +41,30 @@ parsers: ["babel", "flow", "typescript"]
 printWidth: 80
                                                                                 | printWidth
 =====================================input======================================
-function isSourceErrorModule(currentFilePath, targetModulePath, src) {
-for (const srcPath of [src, \`\${src}.js\`, \`\${src}/index\`, \`\${src}/index.js\`]) {}}
+{for (const srcPath of [src, \`\${src}.js\`, \`\${src}/index\`, \`\${src}/index.js\`]) {}}
+{for (const srcPath of [123, 123_123_123, 123_123_123_1, 13_123_3123_31_43]) {}}
+{for (const srcPath of [123, 123_123_123, 123_123_123_1, 13_123_3123_31_432]) {}}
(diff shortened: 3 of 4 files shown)


src/language-js/print/array.js
```diff
@@ -106,6 +98,16 @@ function printArray(path, options, print) {
...
 
+    const trailingComma = !canHaveTrailingComma
+      ? ""
+      : needsForcedTrailingComma
+      ? ","
+      : !shouldPrintComma(options)
+      ? ""
+      : ifBreak(",", "", {
+          groupId: shouldUseConciseFormatting ? groupId : undefined,
+        });
```
fisker, line 109:
```suggestion
let trailingComma = "";
if (canHaveTrailingComma) {
  if (needsForcedTrailingComma) {
    trailingComma = ",";
  } else if (shouldPrintComma(options)) {
    trailingComma = ifBreak(",", "", {
      groupId: shouldUseConciseFormatting ? groupId : undefined,
    });
  }
}
```

thorn0:
I fixed the issue for non-concise formatting, but it still exists for number arrays:

**Prettier pr-10160**
[Playground link](https://deploy-preview-10160--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEwA6UAEmBmEBOmAFJFAM4yZn5gAKAhjABaYQ6YDaAjAEwDMAGky8+AfRHj+QiTP7jpYiXyVdRAFj48AugEpM6LJgC+GIyAEgIABxgBLaGWSh6+fBADuDfAkcp6AG3d6AE9HCwAjfHowAGs4GABlK2jbKABzZBh8AFc4CzgAW3C4ABMS0oAZenTs+jS4ADECAsY7dOQQemyYCHMQJhgC-wB1Jlt4MmSwOASfcdsAN3HgjrAyMJBUsjh8GFootJbkHADtiwArMgAPACEo2PiE+gK4CtS4Y9O8kEurhNS0v44ABFbIQeCffxnEDJfDbfAdcL0Yr+PpWfCpGDDWwlZjIAAcAAYLOiINthlErB10XB4QsPhYAI5g+D7ay+TpkAC0UDgpVKfW8zNs3n2dSOSBOUO+2wKtkyORlAKBoPBH0lXwsMGR2NxTGQPC1UVs-gBAGEIAUJSBaQBWPrZbYAFWRvil0IWuQAklByrAEmAMTYAIK+hIwYJAyHbIxGIA)
```sh
--parser babel
```

**Input:**
```jsx
{
  for (const srcPath of [123, 123_123_123, 123_123_123_1, 13_123_3123_31_432]) {
  }
}
```

**Output:**
```jsx
{
  for (const srcPath of [
    123, 123_123_123, 123_123_123_1, 13_123_3123_31_432,
  ]) {
  }
}

```

src/language-js/print/array.js
thorn0: My version looks better to me :) Maybe it's better to extract `shouldUseConciseFormatting` to the outer conditional chain though.

thorn0: It reads like a chain of simple single-line yes/no questions. I see nothing wrong with it.

fisker: Honestly, I never used a ternary nested more than 2 levels, so I never care about https://github.com/prettier/prettier/pull/9559

I can read it, but a little slow than `if`.

thorn0:
I have to admit I don't know how to fix https://github.com/prettier/prettier/pull/10160/files#issuecomment-768334756
Let's merge this as is -so that non-number arrays won't get reformatted - and keep the issue open, shall we?

src/language-js/print/array.js
fisker: Another thing I don't like in your version, there are two negated conditions, I usually avoid it.

thorn0: When I see nested `if`s like that, it gives me an impression that something tricky and intimidating (=== "slower to read") is going on there that can't be expressed as a simple yes/no chain. That said, I prefer nested `if`s when conditions are more complex than in this case. I really appreciate that you're a fan of ESLint (thanks for all those internal rules), but IMHO you're too strict on innocent code structures that are perfectly okay.

thorn0:
Turns out the thing I can't fix (https://github.com/prettier/prettier/pull/10160/files#issuecomment-768334756) happens only in `for-of`, which isn't a common case for number arrays.

**Prettier pr-10160**
[Playground link](https://deploy-preview-10160--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBDABAXnQbQKwD6AjAEwDMANOqWceXZdfTQ8VUbS2V0QQCxkS7TvW68x-QQF0A3AB0owBenQAbODHQAjLLjyF9hGu2anGLC82Gtuo3gKHVBDWQoC+ICiAgAHGAEtoAGdkUFQAJ3CIAHcABQiEEJRUVWjUAE8Qry1w1DAAaw0AZR88-ygAc2QYcIBXOC84AFstOAATNvaAGVRK2tQKuAAxCHCm1BgAyuQ0WpgITxAACxgm1QB1Jf94INKwOCLE7f8AN230mbAgrJByoLhwmFjcivHkADMU+68AKyCADwAQrkCsVUE04F1ynAPl8GiA-v8iuUKuoAIq1CDwWGqb4gUrhe7hGZaVCtVSLHzhcowdb+NowJbIAAcAAYvFSIPd1rkfDMqXAiScYV4AI6Y+DPXxJNBBAC0UDg7Xai3CcHF-jVzwGbyQn1x8PuTX81TqhpR6IlML1cK8MDJdIZTKQQhANVQ-lUKIAwhAmrqQIK8ItavcACpkpL6vEneoASSgnVgRTA1L8AEFE0UYOl1Dj7m43EA)
```sh
--parser babel
```

**Input:**
```jsx
a = [5_123, 123_123_123, 123_123_123_1, 13_123_3123_31_432, 13_123_311_311_432];
{
  let b = [55_555_123, 123_123_123, 123_123_123_1, 13_123_3123_31_432, 132_123];
}
```

**Output:**
```jsx
a = [5_123, 123_123_123, 123_123_123_1, 13_123_3123_31_432, 13_123_311_311_432];
{
  let b = [55_555_123, 123_123_123, 123_123_123_1, 13_123_3123_31_432, 132_123];
}

```

thorn0:
Isolated a doc to reproduce the issue. If I replace `group(["const srcPath"])` with `["const srcPath"]` the behavior becomes as expected. Looks like using `ifBreak` with a group id inside a group marked with that id, is an undefined behavior.

```js
[
  "for (",
  group(["const srcPath"]),
  " of ",
  group([
    "[",
    indent([
      softline,
      fill([
        [["123_456_789_123_451_123_123_456"], ","],
        line,
        [["113_123_3123_131_432"], ifBreak(",", "", { groupId: 1 })]
      ]),
    ]),
    softline,
    "]",
  ], {id: 1}),
  ") {",
  hardline,
  "}",
]
```

thorn0:
I hope my comments above don't create a scary impression that this PR is unsafe to merge.

Here what's left unsolved here:

"If a `for-of` loop has a number-only array as iterable, then this line breaks one char earlier than it should."

Not a big deal at all as you can see. Please review.

fisker (approved):

PR #10168, opened 2021-01-28
Refactor CLI

fisker:
## Description

1. Move `logger` out `Context`, so we don't need a extra `Context#initContext()` call.
2. Parse `plugin`/`plugin-search-dir` argument in a later stage, so we can use logger to log errors.

stable:

```console
npx prettier --plugin=ninjia
Cannot find module 'ninjia'
Require stack:
- ...
```

this pr:

```console
node bin/prettier --plugin=ninjia
[error] Cannot find module 'ninjia'
[error] Require stack:
[error] - ...
```

stable:

```console
npx prettier --plugin-search-dir=ninjia
ninjia does not exist or is not a directory
```

this pr:
```console
node bin/prettier --plugin-search-dir=ninjia
[error] ninjia does not exist or is not a directory
```

3. Fix a possible bug 

This line https://github.com/prettier/prettier/blob/main/src/cli/index.js#L65 , should not use `context.args["stdin-filepath"]`, `context.args`(renamed to `context.rawArguments` in this PR) is an array, should be `context.argv`

4. Simplified a lot

## Checklist

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -9,19 +9,11 @@ const prettier = require("../index");
 const core = require("./core");
 
 function run(rawArguments) {
-  const {
-    loglevel: logLevel,
-    plugin: plugins,
-    "plugin-search-dir": pluginSearchDirs,
-  } = core.parseArgvWithoutPlugin(rawArguments, [
-    "loglevel",
-    "plugin",
-    "plugin-search-dir",
-  ]);
-  const logger = core.createLogger(logLevel);
-  const options = { rawArguments, plugins, pluginSearchDirs, logger };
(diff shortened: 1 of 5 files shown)


src/cli/index.js
```diff
@@ -7,78 +7,81 @@ const stringify = require("fast-json-stable-stringify");
...
 const prettier = require("../index");
 const core = require("./core");
+const { createLogger } = require("./logger");
+const { init } = require("./context");
 
 function run(args) {
-  const context = new core.Context(args);
-
+  const result = init(args);
+  const logger = createLogger(result.argv.loglevel);
```
fisker, line 15: @sosukesuzuki I got the logger out of `Context`! Will do more refactoring.

sosukesuzuki (approved):
thanks!

PR #10169, opened 2021-01-28
Round-trippable --debug-print-doc

thorn0:
## Description

The idea is to make the output of `--debug-print-doc` closer to actual code for generating docs. Ideally, it should be possible for it to work without modification after copy-pasting into a JS file. That ideal probably hasn't been reached by this PR, but it's pretty close. I believe this will make `--debug-print-doc` (and the corresponding part of the Playground) more useful.

[Before:](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEwA64NIATqt7DAIy2wCY0oBfKkAGhAgAcYBLaAZ2VAEMAnPhADuABX4IuKHgBshPAJ5cGRPjzABrODADKTNaygBzZDD4BXOAzgBbInAAm9hwBkeRsz0NwAYhD7WeGDYjZBAeMxgIehAACxhraQB1GNZ4Dj0wOG0JVNYAN1T5ULAOJRADDjg+GBFVQwDkADMZSoYAKw4ADwAhVQ0tbR5rOGcDOCaWyxAOzu0DQ2k4AEUzCHgJ6VaQPT5KvlCO6GimPgMYRNZ7GBjkAA4ABgYTiErE1SZQk7g9vPGGAEdVvBasxJGEOABaKBwBwOaJ8OCA1gI2qeBpIZqbKaVaysEzmbHzRYrNbjDGTBgwHhEC5XG5IMiU1SsaTzADCEGs6JA3wArNEzJUACrUySYrZ5CwASSgTlg2jApxYAEFZdoYPJFhtKjQgA)
```js
[
  wrappedGroup([
    "{",
    indent([
      line,
      wrappedGroup([
        wrappedGroup([
          '"c"',
          ":",
          " ",
          wrappedGroup([
            "{",
            indent([line, group([group(['"b"', ":", " ", "2"])])]),
            ifBreak(""),
            line,
            "}",
          ]),
        ]),
      ]),
    ]),
    ifBreak(""),
    line,
    "}",
  ]),
  hardline,
  breakParent,
];
```

[After:](https://deploy-preview-10169--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEwA64NIATqt7DAIy2wCY0oBfKkAGhAgAcYBLaAZ2VAEMAnPhADuABX4IuKHgBshPAJ5cGRPjzABrODADKTNaygBzZDD4BXOAzgBbInAAm9hwBkeRsz0NwAYhD7WeGDYjZBAeMxgIehAACxhraQB1GNZ4Dj0wOG0JVNYAN1T5ULAOJRADDjg+GBFVQwDkADMZSoYAKw4ADwAhVQ0tbR5rOGcDOCaWyxAOzu0DQ2k4AEUzCHgJ6VaQPT5KvlCO6GimPgMYRNZ7GBjkAA4ABgYTiErE1SZQk7g9vPGGAEdVvBasxJGEOABaKBwBwOaJ8OCA1gI2qeBpIZqbKaVaysEzmbHzRYrNbjDGTBgwHhEC5XG5IMiU1SsaTzADCEGs6JA3wArNEzJUACrUySYrZ5CwASSgTlg2jApxYAEFZdoYPJFhtKjQgA)
```js
[
  group([
    "{",
    indent([
      line,
      group([
        group([
          '"c"',
          ":",
          " ",
          group(
            [
              "{",
              indent([line, group([group(['"b"', ":", " ", "2"])])]),
              ifBreak(""),
              line,
              "}",
            ],
            { break: true }
          ),
        ]),
      ]),
    ]),
    ifBreak(""),
    line,
    "}",
  ]),
  hardline,
];

```

## Checklist

### DIFF
--- a/src/document/doc-builders.js
+++ b/src/document/doc-builders.js
@@ -186,13 +186,15 @@ function lineSuffix(contents) {
 const lineSuffixBoundary = { type: "line-suffix-boundary" };
 const breakParent = { type: "break-parent" };
 const trim = { type: "trim" };
+
+const hardlineNoBreak = { type: "line", hard: true };
+const literallineNoBreak = { type: "line", hard: true, literal: true };
+
 const line = { type: "line" };
 const softline = { type: "line", soft: true };
-const hardline = concat([{ type: "line", hard: true }, breakParent]);
-const literalline = concat([
-  { type: "line", hard: true, literal: true },
-  breakParent,
-]);
+const hardline = concat([hardlineNoBreak, breakParent]);
+const literalline = concat([literallineNoBreak, breakParent]);
+
 const cursor = { type: "cursor", placeholder: Symbol("cursor") };
 
 /**
@@ -257,4 +259,6 @@ module.exports = {
   markAsRoot,
   dedentToRoot,
   dedent,
+  hardlineNoBreak,
+  literallineNoBreak,
 };
--- a/src/document/doc-builders.js
+++ b/src/document/doc-builders.js
@@ -187,13 +187,17 @@ const lineSuffixBoundary = { type: "line-suffix-boundary" };
 const breakParent = { type: "break-parent" };
 const trim = { type: "trim" };
 
-const hardlineNoBreak = { type: "line", hard: true };
-const literallineNoBreak = { type: "line", hard: true, literal: true };
+const hardlineWithoutBreakParent = { type: "line", hard: true };
+const literallineWithoutBreakParent = {
+  type: "line",
+  hard: true,
+  literal: true,
+};
 
 const line = { type: "line" };
 const softline = { type: "line", soft: true };
-const hardline = concat([hardlineNoBreak, breakParent]);
-const literalline = concat([literallineNoBreak, breakParent]);
+const hardline = concat([hardlineWithoutBreakParent, breakParent]);
+const literalline = concat([literallineWithoutBreakParent, breakParent]);
 
 const cursor = { type: "cursor", placeholder: Symbol("cursor") };
 
@@ -259,6 +263,6 @@ module.exports = {
   markAsRoot,
   dedentToRoot,
   dedent,
-  hardlineNoBreak,
-  literallineNoBreak,
+  hardlineWithoutBreakParent,
+  literallineWithoutBreakParent,
 };
--- a/src/document/doc-debug.js
+++ b/src/document/doc-debug.js
@@ -41,136 +41,139 @@ function flattenDoc(doc) {
   return doc;
 }
 
-function printDoc(doc, index, parentParts) {
-  if (typeof doc === "string") {
-    return JSON.stringify(doc);
-  }
-
-  if (isConcat(doc)) {
-    return `[${getDocParts(doc).map(printDoc).filter(Boolean).join(", ")}]`;
-  }
-
-  if (doc.type === "line") {
-    const withBreakParent =
-      Array.isArray(parentParts) &&
-      parentParts[index + 1] &&
-      parentParts[index + 1].type === "break-parent";
-    if (doc.literal) {
-      return withBreakParent ? "literalline" : "literallineWithoutBreakParent";
-    }
-    if (doc.hard) {
-      return withBreakParent ? "hardline" : "hardlineWithoutBreakParent";
-    }
-    if (doc.soft) {
-      return "softline";
-    }
-    return "line";
-  }
-
-  if (doc.type === "break-parent") {
-    const afterHardline =
-      Array.isArray(parentParts) &&
-      parentParts[index - 1] &&
-      parentParts[index - 1].type === "line" &&
-      parentParts[index - 1].hard;
-    return afterHardline ? undefined : "breakParent";
-  }
-
-  if (doc.type === "trim") {
-    return "trim";
-  }
-
-  if (doc.type === "indent") {
-    return "indent(" + printDoc(doc.contents) + ")";
-  }
-
-  if (doc.type === "align") {
-    return doc.n === Number.NEGATIVE_INFINITY
-      ? "dedentToRoot(" + printDoc(doc.contents) + ")"
-      : doc.n < 0
-      ? "dedent(" + printDoc(doc.contents) + ")"
-      : doc.n.type === "root"
-      ? "markAsRoot(" + printDoc(doc.contents) + ")"
-      : "align(" + JSON.stringify(doc.n) + ", " + printDoc(doc.contents) + ")";
-  }
-
-  if (doc.type === "if-break") {
-    return (
-      "ifBreak(" +
-      printDoc(doc.breakContents) +
-      (doc.flatContents ? ", " + printDoc(doc.flatContents) : "") +
-      (doc.groupId
-        ? (!doc.flatContents ? ', ""' : "") +
-          `, { groupId: ${JSON.stringify(printGroupId(doc.groupId))} }`
-        : "") +
-      ")"
-    );
-  }
-
-  if (doc.type === "group") {
-    const optionsParts = [];
-
-    if (doc.break && doc.break !== "propagated") {
-      optionsParts.push("break: true");
-    }
-
-    if (doc.id) {
-      optionsParts.push(`id: ${JSON.stringify(printGroupId(doc.id))}`);
-    }
-
-    const options =
-      optionsParts.length > 0 ? `, { ${optionsParts.join(", ")} }` : "";
-
-    if (doc.expandedStates) {
-      return `conditionalGroup([${doc.expandedStates
-        .map(printDoc)
-        .join(",")}]${options})`;
-    }
-
-    return `group(${printDoc(doc.contents)}${options})`;
-  }
-
-  if (doc.type === "fill") {
-    return `fill([${doc.parts.map(printDoc).join(", ")}])`;
-  }
-
-  if (doc.type === "line-suffix") {
-    return "lineSuffix(" + printDoc(doc.contents) + ")";
-  }
-
-  if (doc.type === "line-suffix-boundary") {
-    return "lineSuffixBoundary";
-  }
-
-  throw new Error("Unknown doc type " + doc.type);
-}
-
-let symbolMap;
-let usedStrings;
-
-function printGroupId(id) {
-  if (typeof id !== "symbol") {
-    return JSON.stringify(String(id));
-  }
-
-  if (id in symbolMap) {
-    return symbolMap[id];
-  }
-
-  const prefix = id.description || "symbol";
-  for (let counter = 0; ; counter++) {
-    const string = prefix + (counter > 0 ? ` #${counter}` : "");
-    if (!usedStrings.has(string)) {
-      usedStrings.add(string);
-      return (symbolMap[id] = `Symbol.for(${JSON.stringify(string)})`);
-    }
-  }
-}
-
 function printDocToDebug(doc) {
-  symbolMap = Object.create(null);
-  usedStrings = new Set();
+  const symbolMap = Object.create(null);
+  const usedStrings = new Set();
   return printDoc(flattenDoc(doc));
+
+  function printDoc(doc, index, parentParts) {
+    if (typeof doc === "string") {
+      return JSON.stringify(doc);
+    }
+
+    if (isConcat(doc)) {
+      return `[${getDocParts(doc).map(printDoc).filter(Boolean).join(", ")}]`;
+    }
+
+    if (doc.type === "line") {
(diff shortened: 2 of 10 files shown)


src/document/doc-debug.js
```diff
@@ -90,30 +99,43 @@ function printDoc(doc) {
       "ifBreak(" +
       printDoc(doc.breakContents) +
       (doc.flatContents ? ", " + printDoc(doc.flatContents) : "") +
+      (doc.groupId
+        ? (!doc.flatContents ? ', ""' : "") +
+          (", { groupId: " + JSON.stringify(printGroupId(doc.groupId)) + " }")
```
lipis, line 104:
```suggestion
          (`, { groupId:  ${JSON.stringify(printGroupId(doc.groupId))} }`)
```

src/document/doc-debug.js
```diff
@@ -90,30 +99,43 @@ function printDoc(doc) {
...
+    if (doc.break && doc.break !== "propagated") {
+      optionsParts.push("break: true");
+    }
+
+    if (doc.id) {
+      optionsParts.push("id: " + JSON.stringify(printGroupId(doc.id)));
+    }
+
+    const options =
+      optionsParts.length > 0 ? ", { " + optionsParts.join(", ") + " }" : "";
```
lipis, line 122:
```suggestion
      optionsParts.length > 0 ? `, { ${optionsParts.join(", ")} }` : "";
```

thorn0: Okay, I see. I'll convert them all. :)

src/document/doc-debug.js
lipis: Maybe for another PR though

lipis (approved):

src/document/doc-debug.js
```diff
@@ -127,8 +145,33 @@ function printDoc(doc) {
...
+    return String(id);
+  }
+
+  if (id in symbolMap) {
+    return symbolMap[id];
+  }
+
+  const prefix = id.description || "symbol";
+  for (let i = 0; ; i++) {
+    const string = prefix + (i > 0 ? " #" + i : "");
```
lipis, line 162:
```suggestion
  for (let index = 0; ; index++) {
    const string = prefix + (index > 0 ? ` #${index}` : "");
```

tests_integration/__tests__/debug-api.js
```diff
@@ -45,4 +42,15 @@ describe("API", () => {
   test("prettier.printDocToString", () => {
     expect(stringFromDoc).toBe(formatted);
   });
+
+  const doc2 = new Function(
+    `{ ${Object.keys(builders)} }`,
+    "return " + formatResultFromDoc
```
lipis, line 48:
```suggestion
    `return ${formatResultFromDoc}`
```

src/document/doc-builders.js
```diff
@@ -257,4 +259,6 @@ module.exports = {
   markAsRoot,
   dedentToRoot,
   dedent,
+  hardlineNoBreak,
```
fisker, line 262: These is going public, better add comment why they are added, so we can remove if we don't need anymore.

Another thing I'm not sure if you know, this is actually used in codebase https://github.com/prettier/prettier/blob/1df287acc802939edfd5eed3ffabf3979a8cc3ae/src/language-markdown/printer-markdown.js#L560,  did you consider this when you make changes for the debug-print

lipis:
@thorn0  #10175  :)

src/document/doc-builders.js
thorn0: The fact that they're used is exactly the reason why we need to be able to print them in debug-print.

thorn0: > we can remove if we don't need anymore

Why would that happen?

thorn0: Have a look at `show doc` [here](https://deploy-preview-10169--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAFAlgGwjAPgQXQCcAHCIvACQgFciBnAHSgFoXck2lcvmApCAAsoAAgAiEOLgDKAaQDiuAIwBmFQA5+AQyhwRAJUm4+AMVm4VSgCzMQAGhAQSMdNHrJQWokQgB3VF4I7ihamL5aAJ7uDgBGRFpgANZwMNIkCehQAObIMEQ0cA5wALYxcAAm5RUAMjpZNFpZcCbkxVowLtnIIFo0MBD2IIIwxZgA6oLo8PTpYHDSQVPoAG5TEd1g9NEgmfRwFKjxWW3IAGahew4AVvQAHgBC8Ukp0lrFcNWZcGcXhSA3t2kmSymDgAEUaDhvkhzphLiB0gx9t02kREuU-FBBiQiJkYGN0OUYIJkOoAAwOHEQPZjeIkbo4uB7IjLb4OACOkPghycwR69BYugqFUGRDgnOIcEOjROMN+Dj2xXQuXyf3owNBEKhPzhfxgWhiBKJJKQACYHHktFhgQBhCDFWUgJkAVkGND2ABUDcFYfDlgUAJJQKqwaRgXHOfDB6QwCKgnV7AC+iaAA).

fisker: > Why would that happen?

Point is, adding things are  easy, but  hard 
to  remove things , it's not documented. Have you ever imagine we are removing `concat()`? Things change. All I ask it add a  comment mention it's not intend to let people use.

thorn0: It's used already for Markdown and at least in the Ruby plugin. So I'd rather document it and let people use it directly, not via `getDocParts`.

fisker: Fine to me.

src/document/doc-debug.js
```diff
@@ -64,7 +68,12 @@ function printDoc(doc) {
   }
 
   if (doc.type === "break-parent") {
-    return "breakParent";
+    const afterHardline =
+      Array.isArray(parentParts) &&
+      parentParts[index - 1] &&
+      parentParts[index - 1].type === "line" &&
+      parentParts[index - 1].hard;
+    return afterHardline ? undefined : "breakParent";
```
fisker, line 76: Do you think it's easier to move this part to `flattenDoc`, add a function to replace `hardline + breakParent` as the new `parts`

https://github.com/thorn0/prettier/blob/2fad7569a9a9205ac222878bf21d5a53ff32cb50/src/document/doc-debug.js#L19
And
https://github.com/thorn0/prettier/blob/2fad7569a9a9205ac222878bf21d5a53ff32cb50/src/document/doc-debug.js#L37

fisker, line 76: Do you think it's easier to move this part to `flattenDoc`, add a function to replace `hardline + breakParent` as the new `parts`

https://github.com/thorn0/prettier/blob/2fad7569a9a9205ac222878bf21d5a53ff32cb50/src/document/doc-debug.js#L19
And
https://github.com/thorn0/prettier/blob/2fad7569a9a9205ac222878bf21d5a53ff32cb50/src/document/doc-debug.js#L37

thorn0: The current solution seems to be good enough. Also I'd like to keep it possible to easily comment out the `flattenDoc` call in order to see how the doc looks without it.

fisker: https://github.com/prettier/prettier/pull/10169/files#diff-d5a6031cf2b293b180f03d67c8c402c7c13eb7358ccf04aa3cc4f1ad81f2fc4aR54

This part won't needed too.

We can replace `hardlineWithoutBreakParent + breakParent` with a new `type` while only use in this file, or mark the `breakParent` is after `hardlineWithoutBreakParent`, I think both okay, we won't need do the same check in two places.

thorn0: > mark the breakParent is after hardlineWithoutBreakParent

You mean in doc-builders? That didn't cross my mind. I'm okay with that.

fisker: > keep it possible to easily comment out the flattenDoc call in order to see how the doc looks without it.

If that's what you need, we can mark the `breakParent` is after `hardline` when doing the check, so we don't need check twice.

thorn0: On the other hand, what if somebody in some plugin compares `breakParent` by reference? I'd rather keep things the way they are.

fisker: Okay, good one.

src/document/doc-debug.js
```diff
@@ -127,8 +145,33 @@ function printDoc(doc) {
...
+  if (id in symbolMap) {
+    return symbolMap[id];
+  }
+
+  const prefix = id.description || "symbol";
+  for (let counter = 0; ; counter++) {
+    const string = prefix + (counter > 0 ? ` #${counter}` : "");
+    if (!usedStrings.has(string)) {
+      symbolMap[id] = string;
+      usedStrings.add(string);
```
fisker, line 165: I don't understand this part, why need two object save the same thing?

thorn0: Symbols can't made round-trippable, so we need to generate unique strings instead.
`symbolMap` is used to look up by a symbol, whether it already has an associated string.
`usedStrings` - to look up by a string, to check whether the given string is already in use for some other symbol.

fisker: And use a shared variable look unusual to me, I would do

```js
function printDoc(doc) {
  const symbolMap, usedStrings

  function theRealPrintToRecursion() {}

  return theRealPrintToRecursion(doc)
}
```

thorn0: Just realized I can generate `Symbol.for("unique string")` instead of just "unique string". I'm going to fix this.

thorn0: > shared variable look unusual to me

It's fine as long as the function is sync and doesn't reenter itself, we have this in `doc-printer` too: see `groupModeMap`. Probably I should've fixed `lineSuffixBoundary` the same way, by extracting the `lineSuffix` variable to the outer scope instead of adding parameters to `fits`.

> I would do

Agree. Actually, I did that too, but then I decided to minimize the diff to make this PR easier to review. :)

fisker: Right, I saw `groupModeMap`, don't like it.

thorn0: Okay, I'll wrap these variables in a function scope this PR. And let's do the same with `groupModeMap` and `lineSuffix` in another (Do you want to give it a shot?).

fisker: All yours, it's late in China, and I have a slow net connection.

thorn0: As you wish. It doesn't have to be today anyway.

src/document/doc-debug.js
```diff
@@ -35,100 +33,147 @@ function flattenDoc(doc) {
...
+    if (id in symbolMap) {
+      return symbolMap[id];
+    }
+
+    const prefix = id.description || "symbol";
+    for (let counter = 0; ; counter++) {
+      const string = prefix + (counter > 0 ? ` #${counter}` : "");
+      if (!usedStrings.has(string)) {
+        usedStrings.add(string);
+        return (symbolMap[id] = `Symbol.for(${JSON.stringify(string)})`);
```
fisker, line 173: I used like `return a = b` and `a = b = c`, long time no see...

fisker: Aren't `[...usedStrings]` deepEqual `Object.keys(symbolMap)` ? I mean before you use `Symbol.for`.

So we can also use `string in symbolMap` as condition, right?

thorn0: No, both before and after it's `Object.value(symbolMap)`.

fisker: Oh, right.

fisker: SO, question is, is there a better way to check `Object.values(symbolMap).includes(string)`, I guess `Set` is a better choice.

fisker (approved):

src/document/doc-debug.js
thorn0: Better than what? I use a set already.

thorn0: JS doesn't have [bimaps](https://github.com/google/guava/wiki/NewCollectionTypesExplained#bimap)...

fisker: I know, so you already have the best solution.

fisker: Is in faster then `Set`? You can also use `Set` replace that `in`. If you add both key and value in it... Not sure about the performance.

thorn0: This code is a debugging tool :)

fisker: Haha, but isn't it an interesting idea? 😄

PR #10173, opened 2021-01-29
CLI: Refactor `createLogger` to `Logger` class

fisker:
## Description

Switch to `Logger` class, so we can dynamic change the `logLevel`, after we merge #10168, we'll only have `loglevel` argument parse process without a `logger`, I'm planning add use a default `logger` during `loglevel` argument parse, after we get the correct `loglevel`, change it to correct level, if we don't use this class, we may need `createLogger` twice.

## Checklist


fisker:
Create `logger` twice in #10207, I think it won't cost too much, let's keep the old way.

PR #10175, opened 2021-01-29
style: Add new ESLint rule `prefer-template`

lipis:
It's an autofixable rule and keeps everything consistent.

### DIFF
--- a/scripts/build-docs.js
+++ b/scripts/build-docs.js
@@ -31,7 +31,7 @@ if (isPullRequest) {
   const newPkg = { ...pkg, version: `999.999.999-pr.${process.env.REVIEW_ID}` };
   pipe(JSON.stringify(newPkg, null, 2)).to("package.json");
   shell.exec("yarn build --playground");
-  pipe(JSON.stringify(pkg, null, 2) + "\n").to("package.json"); // restore
+  pipe(`${JSON.stringify(pkg, null, 2)}\n`).to("package.json"); // restore
 }
 
 shell.cp(`${prettierPath}/standalone.js`, `${docs}/`);
--- a/src/cli/expand-patterns.js
+++ b/src/cli/expand-patterns.js
@@ -57,7 +57,7 @@ function* expandPatternsInternal(context) {
     dot: true,
     ignore: Object.entries(silentlyIgnoredDirs)
       .filter(([, ignored]) => ignored)
-      .map(([dir]) => "**/" + dir),
+      .map(([dir]) => `**/${dir}`),
   };
 
   let supportedFilesGlob;
@@ -84,10 +84,9 @@ function* expandPatternsInternal(context) {
       } else if (stat.isDirectory()) {
         entries.push({
           type: "dir",
-          glob:
-            escapePathForGlob(fixWindowsSlashes(pattern)) +
-            "/" +
-            getSupportedFilesGlob(),
+          glob: `${escapePathForGlob(
+            fixWindowsSlashes(pattern)
+          )}/${getSupportedFilesGlob()}`,
           input: pattern,
         });
       }
@@ -133,7 +132,7 @@ function* expandPatternsInternal(context) {
         context.languages.map((lang) => lang.filenames || [])
       );
       supportedFilesGlob = `**/{${[
-        ...extensions.map((ext) => "*" + (ext[0] === "." ? ext : "." + ext)),
+        ...extensions.map((ext) => `*${ext[0] === "." ? ext : `.${ext}`}`),
         ...filenames,
       ]}}`;
     }
--- a/tests_integration/__tests__/patterns-dirs.js
+++ b/tests_integration/__tests__/patterns-dirs.js
@@ -120,12 +120,11 @@ if (path.sep === "/") {
 }
 
 function testPatterns(namePrefix, cliArgs, expected = {}) {
-  const testName =
-    (namePrefix ? namePrefix + ": " : "") +
-    "prettier " +
-    cliArgs
-      .map((arg) => (/^[\w./=-]+$/.test(arg) ? arg : `'${arg}'`))
-      .join(" ");
+  const testName = `${
+    namePrefix ? `${namePrefix}: ` : ""
+  }prettier ${cliArgs
+    .map((arg) => (/^[\w./=-]+$/.test(arg) ? arg : `'${arg}'`))
+    .join(" ")}`;
 
   describe(testName, () => {
     runPrettier("cli/patterns-dirs", [...cliArgs, "-l"]).test({
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -54,6 +54,7 @@ rules:
   prefer-object-spread: error
   prefer-rest-params: error
   prefer-spread: error
+  prefer-template: error
   prettier-internal-rules/jsx-identifier-case: error
   prettier-internal-rules/require-json-extensions: error
   quotes:
--- a/scripts/build/build.js
+++ b/scripts/build/build.js
@@ -102,7 +102,7 @@ async function createBundle(bundleConfig, cache, options) {
 
     console.log(status.DONE);
   } catch (error) {
-    console.log(status.FAIL + "\n");
+    console.log(`${status.FAIL}\n`);
     handleError(error);
   }
 }
--- a/scripts/build/cache.js
+++ b/scripts/build/cache.js
@@ -125,7 +125,7 @@ class Cache {
 }
 
 function required(name) {
-  throw new Error(name + " is required");
+  throw new Error(`${name} is required`);
 }
 
 function hashString(string) {
--- a/scripts/build/rollup-plugins/executable.js
+++ b/scripts/build/rollup-plugins/executable.js
@@ -36,7 +36,7 @@ module.exports = function () {
 
     renderChunk(code) {
       if (banner) {
-        return { code: banner + "\n" + code };
+        return { code: `${banner}\n${code}` };
       }
     },
 
--- a/scripts/build/rollup-plugins/native-shims.js
+++ b/scripts/build/rollup-plugins/native-shims.js
@@ -18,7 +18,7 @@ module.exports = function (dir) {
         return null;
       }
 
-      const shim = path.resolve(dir, importee + ".mjs");
+      const shim = path.resolve(dir, `${importee}.mjs`);
       if (fs.existsSync(shim)) {
         return shim;
       }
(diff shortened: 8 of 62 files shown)


fisker (approved):

fisker:
I don't think it's `feat:`

alexander-akait:
`style:`

tests_integration/__tests__/patterns-dirs.js
```diff
@@ -120,12 +120,11 @@ if (path.sep === "/") {
...
-    (namePrefix ? namePrefix + ": " : "") +
-    "prettier " +
-    cliArgs
-      .map((arg) => (/^[\w./=-]+$/.test(arg) ? arg : `'${arg}'`))
-      .join(" ");
+  const testName = `${
+    namePrefix ? `${namePrefix}: ` : ""
+  }prettier ${cliArgs
+    .map((arg) => (/^[\w./=-]+$/.test(arg) ? arg : `'${arg}'`))
+    .join(" ")}`;
```
fisker, line 127: This is so ugly.

lipis: Make it _Prettier_ in another PR :)

thorn0: That's why I'm against enforcing lint rules like this. Often enough concatenation is more readable than a template literal, and that can't be fixed by changing the formatting of the template literal.

"Linters are here to help people find bugs." or "People are here to make linters happy."? What's your camp? ;)

lipis: Of course can't win them all.. but string literals are generally better. I'm sure in this case could make it better if using some variables

thorn0: Often better but now always. E.g. for simple things like `someLongExpression(withArg) + "\n"` concatenation is more clear too because `+` expresses the same idea of addition as in `expr + 1`: "add one to a number", "add `\n` to a string". Template literals are clear winners only if the static part of the template is bigger than the expressions.

scripts/build-docs.js
```diff
@@ -31,7 +31,7 @@ if (isPullRequest) {
   const newPkg = { ...pkg, version: `999.999.999-pr.${process.env.REVIEW_ID}` };
   pipe(JSON.stringify(newPkg, null, 2)).to("package.json");
   shell.exec("yarn build --playground");
-  pipe(JSON.stringify(pkg, null, 2) + "\n").to("package.json"); // restore
+  pipe(`${JSON.stringify(pkg, null, 2)}\n`).to("package.json"); // restore
```
thorn0, line 34: Seriously, this rule is too aggressive.

thorn0:
The fact that we have template literals in the language now doesn't mean that `+` is somehow deprecated or is now a "bad part" of the language. It's not like `var` vs `let/const`. I blame ESLint and its inflexible rules like this (there is a reason it's not included in `eslint:recommended` though) for spreading this wrong impression. While many changes in this PR are okay, the rest of them is controversial at the very least.

scripts/build-docs.js
thorn0: @fisker Seriously, what's wrong with `+ '\n'`?

fisker: Nothing wrong, but also nothing wrong use template literal, think this way, this rule is opinionated,  just live with it.

Sometimes like Prettier self, in many places, user don't have a choice, have to accept what ever Prettier choose.

src/cli/expand-patterns.js
```diff
@@ -84,10 +84,9 @@ function* expandPatternsInternal(context) {
       } else if (stat.isDirectory()) {
         entries.push({
           type: "dir",
-          glob:
-            escapePathForGlob(fixWindowsSlashes(pattern)) +
-            "/" +
-            getSupportedFilesGlob(),
+          glob: `${escapePathForGlob(
+            fixWindowsSlashes(pattern)
+          )}/${getSupportedFilesGlob()}`,
```
alexander-akait, line 89: But we really should improve prettier here...

alexander-akait:
I think this rule will allow to us catch problems with templates, many developers complain that it is ugly, and here we seem to be several examples of this

scripts/build-docs.js
thorn0: In the case of Prettier the user gets a lot of benifits in return. I don't see enough benifits in this case.

thorn0: All those opinionated decisions in Prettier aren't needed just for the sake of themselves. They're needed to make Prettier work. Prettier won't work at all without those decisions. They're a neccessity. Nothing like these ESLint rules that are just linting for the sake of linting.

alexander-akait: @thorn0 What about this https://github.com/prettier/prettier/pull/10175#issuecomment-769771050?

fisker: You said yourself, many changes are okay, shouldn't that enough?

thorn0:
Those developers complain [exactly because someone in their projects decided that enabling this ESLint rule is a good idea](https://github.com/prettier/prettier/issues/3368#issuecomment-596980743). And now they would like to use `+` instead of templates in some cases but they can't.

src/cli/expand-patterns.js
thorn0: How exactly?

scripts/build-docs.js
thorn0: No, it's not enough. It means the rule isn't good enough. It needs an improvement, a less aggressive mode that won't uglify simple binary expressions like `something + '\n'`

src/cli/expand-patterns.js
alexander-akait: More smart logic for breaks, we have a circular problem - prettier format templates ugly in some cases, therefore developers complain that the `eslint` rule do code ugly and disable it. I agree it can be less aggressive, but prettier is not good too. And I agree that it is not so easy, but I am sure that we can find a solution.

alexander-akait: We can't just say `eslint` rule is bad don't use it, in theory even if we discuss with eslint developers and do it less aggressive, there are still a lot of ugly cases, we can look at them in our issues and some snapshots

scripts/build-docs.js
thorn0: @fisker I'd like to emphasize again if my point wasn't clear that you can't compare things like this to Prettier. They're incomparable.

The benefits you get in return when you accept Prettier's opinionated decisions: Prettier itself with all its benefits.
The benefits you get in return when you accept ugliness that comes from enforcing `prefer-templates`: some expressions now look better. Ehm... ⚖

src/cli/expand-patterns.js
thorn0: Some code just can't look pretty no matter how you try to format it. Luckily the language has alternative constructions that can help in such cases. E.g. `+` for concatenation.

alexander-akait: But in most cases it can look better, we should get more benefits than ugly, that's why we are working on `prettier`

tests_integration/__tests__/patterns-dirs.js
alexander-akait: Not always, for example:
```js
const links = `Docs: ${links.map((link) => colors.bold(link)).join('\n')};
```

Expression is more, yep it can be refactored on separate variables, but why?

thorn0:
A tweet I wholeheartedly agree with: https://mobile.twitter.com/dan_abramov/status/1086215771401920512

scripts/build-docs.js
fisker: Checkout this PR, revert lines you don't like, turn off the rule, commit. Tools suppose to help us, not force us. I offen turn on some rules I usually don't use, just to check possible problems and see if improvements.

What do you say @lipis?

alexander-akait:
But this rule allows to find ugly templates :smile: Honestly, I see more benefits, just my opinion

src/cli/expand-patterns.js
thorn0: What you're saying is very abstract. Let's be more specific and discuss expression with this structure: `<long expression 1> + "/" + <long expression 2>`. I say that no matter how you (or Prettier) try to format such expressions, in general case concatenation would look better than a template literal.

This specific case you commented
```js
glob: `${escapePathForGlob(
  fixWindowsSlashes(pattern)
)}/${getSupportedFilesGlob()}`
```
can be improved by not wrapping the expression at all:
```js
glob: `${escapePathForGlob(fixWindowsSlashes(pattern))}/${getSupportedFilesGlob()}`
```
with spaces it can become a bit better probably (looks unusual though):
```js
glob: `${ escapePathForGlob(fixWindowsSlashes(pattern)) }/${ getSupportedFilesGlob() }`
```
But this strategy won't work at all if the expressions are longer. So we need to break lines. But how exactly? The current version of line breaking doesn't look good to you, right? How else can it be done?
```js
// 1
glob: `${escapePathForGlob(fixWindowsSlashes(pattern))}/${
  getSupportedFilesGlob()}`

// 2
glob: `${
  escapePathForGlob(fixWindowsSlashes(pattern))
}/${
  getSupportedFilesGlob()
}`
```

Do these look better?

tests_integration/__tests__/patterns-dirs.js
thorn0: I mean in those situations they're **clear** winners. In other situations things aren't as clear-cut.

src/cli/expand-patterns.js
alexander-akait: Honestly I prefer do not break here, i.e.
```js
glob: `${escapePathForGlob(fixWindowsSlashes(pattern))}/${getSupportedFilesGlob()}`
```

But I agree this example show how different can it be formatted, I don't have strong opinion here, but moving `fixWindowsSlashes(pattern)` argument for `escapePathForGlob` call the the next line looks quite ugly

tests_integration/__tests__/patterns-dirs.js
thorn0: Overall, for strings with leading/trailing spaces, templates are usually better for this reason:
```js
const links = "Docs: " + links.map((link) => colors.bold(link)).join('\n');
//                  ^^^ 👎
const links = `Docs: ${links.map((link) => colors.bold(link)).join('\n')}`;
//                  ^ 👍
```

src/cli/expand-patterns.js
thorn0: > Honestly I prefer do not break here

Me too, but as I wrote, it's not a generic solution. What if the expressions are longer? Are we going to have 1000 characters long lines? Prettier needs generic solutions that can used algorithmically.

fisker:
Close this since it's old, we can re-enable the rule in another PR if we still want.

PR #10179, opened 2021-01-30
handlebars: Split text content on multiple lines

dcyriller:
## Description

```handlebars
{{!-- Input --}}
<div>
  A long enough string to trigger a line break that would prevent wrapping more and more.
</div>

{{!-- Prettier stable --}}
<div>
  A long enough string to trigger a line break that would prevent wrapping more and more.
</div>

{{!-- Prettier main --}}
<div>
  A long enough string to trigger a line break that would prevent wrapping more
  and more.
</div>
```

## Checklist

### DIFF
--- a/src/language-handlebars/clean.js
+++ b/src/language-handlebars/clean.js
@@ -3,7 +3,11 @@
 function clean(ast, newNode /*, parent*/) {
   // (Glimmer/HTML) ignore TextNode
   if (ast.type === "TextNode") {
-    return null;
+    const trimmed = ast.chars.trim();
+    if (!trimmed) {
+      return null;
+    }
+    newNode.chars = trimmed.replace(/[\t\n\f\r ]+/g, " ");
   }
 
   // `class` is reformatted
--- a/src/language-handlebars/clean.js
+++ b/src/language-handlebars/clean.js
@@ -1,13 +1,9 @@
 "use strict";
 
 function clean(ast, newNode /*, parent*/) {
-  // (Glimmer/HTML) ignore TextNode whitespace
+  // (Glimmer/HTML) ignore TextNode
   if (ast.type === "TextNode") {
-    const trimmed = ast.chars.trim();
-    if (!trimmed) {
-      return null;
-    }
-    newNode.chars = trimmed;
+    return null;
   }
 
   // `class` is reformatted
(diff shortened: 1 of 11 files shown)


src/language-handlebars/clean.js
```diff
@@ -1,20 +1,17 @@
...
 function clean(ast, newNode /*, parent*/) {
-  // (Glimmer/HTML) ignore TextNode whitespace
+  // (Glimmer/HTML) ignore TextNode
   if (ast.type === "TextNode") {
-    const trimmed = ast.chars.trim();
-    if (!trimmed) {
-      return null;
-    }
-    newNode.chars = trimmed;
+    return null;
```
fisker, line 6:
```suggestion
    const trimmed = ast.chars.trim();
    if (!trimmed) {
      return null;
    }
    newNode.chars = trimmed.replace(/[\t\n\f\r ]+/g, '');
```

fisker: We only change these characters, let's replace it with a ` `.

src/language-handlebars/printer-glimmer.js
```diff
@@ -345,9 +356,11 @@ function print(path, options, print) {
         trailingSpace = "";
       }
 
+      text = text.replace(/^\s+/g, leadingSpace).replace(/\s+$/, trailingSpace);
```
fisker, line 359: As you're on this, let's change `\s` to html whitespace?

src/language-handlebars/printer-glimmer.js
```diff
@@ -356,7 +356,9 @@ function print(path, options, print) {
         trailingSpace = "";
       }
 
-      text = text.replace(/^\s+/g, leadingSpace).replace(/\s+$/, trailingSpace);
+      text = text
+        .replace(/^[\t\n\f\r ]+/g, leadingSpace)
```
fisker, line 360: Let's see  if any test fails, if not, please add some test.

dcyriller: I modified the git history to:
- add a test, I pushed this commit to the repo, it should ✅
- modify the printer to only replace html whitespaces (and not vertical tabulations), as a result CI should ❌
- record the tests, CI ✅

dcyriller:
This one should be ready to review

fisker (approved):

thorn0 (approved):

PR #10181, opened 2021-01-30
docs: revise commands.md

thorn0:
## Description

* added some important unobvious details
* minor edits

BTW, isn't it time to move this info to https://prettier.io/docs/en/technical-details.html?

## Checklist

### DIFF
--- a/commands.md
+++ b/commands.md
@@ -1,4 +1,4 @@
-The core of the algorithm is implemented in `src/document/doc-{printer,builders,utils,debug}.js`. The printer uses the basic formatting abstractions provided to construct a format when printing a node.
+The core of the algorithm is implemented in `src/document/doc-{printer,builders,utils}.js`. The printer uses the basic formatting abstractions provided to construct a format when printing a node.
 
 ## Prettier's intermediate representation: `Doc`
 
@@ -10,7 +10,7 @@ type Doc = string | Doc[] | DocCommand;
 
 - _strings_ are printed directly as is (however for the algorithm to work properly they shouldn't contain line break characters)
 - _arrays_ are used to concatenate a list of docs to be printed sequentially into a single doc
-- `DocCommand` is one of the following:
+- `DocCommand` is any of the following:
 
 ### group
 
--- a/commands.md
+++ b/commands.md
@@ -86,7 +86,11 @@ Expects the `docs` argument to be an array of alternating content and line break
 ### ifBreak
 
 ```ts
-declare function ifBreak(ifBreak: Doc, noBreak?: Doc, groupId?: symbol): Doc;
+declare function ifBreak(
+  ifBreak: Doc,
+  noBreak?: Doc,
+  options?: { groupId?: symbol }
+): Doc;
 ```
 
 Print something if the current `group` or the current element of `fill` breaks and something else if it doesn't.
--- a/commands.md
+++ b/commands.md
@@ -252,7 +252,7 @@ Decrease the current indentation to the root marked by [`markAsRoot`](#markAsRoo
 declare const trim: Doc;
 ```
 
-Trim any whitespace or tab character on the current line. This can be used for preprocessor directives.
+Trim all the indentation on the current line. This can be used for preprocessor directives. Should be placed after a line break.
 
 ### hardlineWithoutBreakParent and literallineWithoutBreakParent
 
--- a/commands.md
+++ b/commands.md
@@ -89,7 +89,7 @@ Expects the `docs` argument to be an array of alternating content and line break
 declare function ifBreak(ifBreak: Doc, noBreak?: Doc, groupId?: symbol): Doc;
 ```
 
-Print something if the current group breaks and something else if it doesn't.
+Print something if the current `group` or `fill` breaks and something else if it doesn't.
 
 ```js
 ifBreak(";", " ");
--- a/commands.md
+++ b/commands.md
@@ -89,7 +89,7 @@ Expects the `docs` argument to be an array of alternating content and line break
 declare function ifBreak(ifBreak: Doc, noBreak?: Doc, groupId?: symbol): Doc;
 ```
 
-Print something if the current `group` or `fill` breaks and something else if it doesn't.
+Print something if the current `group` or the current element of `fill` breaks and something else if it doesn't.
 
 ```js
 ifBreak(";", " ");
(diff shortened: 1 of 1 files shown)


commands.md
```diff
@@ -1,4 +1,4 @@
-The core of the algorithm is implemented in `doc-{printer,builders,utils,debug}.js`. The printer should use the basic formatting abstractions provided to construct a format when printing a node. Parts of the API only exist to be compatible with recast's previous API to ease migration, but over time we can clean it up.
+The core of the algorithm is implemented in `doc-{printer,builders,utils,debug}.js`. The printer uses the basic formatting abstractions provided to construct a format when printing a node.
```
fisker, line 1:
```suggestion
The core of the algorithm is implemented in `src/document/doc-{printer,builders,utils,debug}.js`. The printer uses the basic formatting abstractions provided to construct a format when printing a node.
```

commands.md
```diff
@@ -232,28 +234,45 @@ This will dedent the current indentation to the root marked by `markAsRoot`.
...
+declare const literallineWithoutBreakParent: Doc;
+```
+
+These are used very rarely, for advanced formatting tricks. Unlike their "normal" counterparts, they don't include an implicit `breakParent`.
 
 Examples:
 
 - `hardlineWithoutBreakParent` is used for printing tables in Prettier's Markdown printer. With `proseWrap` set to `never`, the columns are aligned only if none of the rows exceeds `printWidth`.
-- `literallineWithoutBreakParent` (called `literalLineNoBreak` [there](https://github.com/prettier/plugin-ruby/blob/9e658db13d3f0524d816528e060b1381e0431559/src/ruby/nodes/heredocs.js)) is used in the [Ruby plugin](https://github.com/prettier/plugin-ruby) for printing heredoc syntax.
+- `literallineWithoutBreakParent` is used in the [Ruby plugin](https://github.com/prettier/plugin-ruby) for [printing heredoc syntax](https://github.com/prettier/plugin-ruby/blob/master/src/ruby/nodes/heredocs.js).
```
fisker, line 256:
```suggestion
- `literallineWithoutBreakParent` is used in the [Ruby plugin](https://github.com/prettier/plugin-ruby) for [printing heredoc syntax](https://github.com/prettier/plugin-ruby/blob/b6e7bd6bc3f70de8f146aa58ad0c8310518bf467/src/ruby/nodes/heredocs.js).
```

Better use a permalink.

fisker: Or maybe link to line 2?

thorn0: The file is small enough and contains good comments

fisker:
Though this file is about commands, do you think we should mention `string` and `Doc[]`?
They are doc but not build with command.

thorn0:
Maybe this way?

```diff
-The following commands are available:
+## Prettier's intermediate representation: `Doc`
+
+```ts
+type Doc = string | Doc[] | DocCommand;
+```
+
+where `DocCommand` is the union of return value types of the following functions (commands):
```

fisker:
We should also document parameters of these functions.

fisker:
Oh, it's there, let's split these function commands and other `Doc` command?

fisker:
@thorn0 Do you think we should freeze those builtin docs? If some plugin modify it, it's dangerous.

thorn0:
Re parameters: actually, I forgot I was going to add the missing `id` for `group` / `groupId` for `ifBreak`.

Re freezing: sounds like a good idea, but I don't know the performance implications.

thorn0:
> `DocCommand` is the union of return value types of the following functions (commands)

That's a bad wording. Too complicated. So I changed to this:

> `DocCommand` is the result of calling any of the following functions

And now I remembered not all of them are functions. Hm...

fisker:
> but I don't know the performance implications.

Isn't it one time thing? I don't think there would be any performance issue.

thorn0:
Still, I'd prefer to set up some kind of benchmark on CI first before proceeding with such experiments.

sosukesuzuki (approved):

lipis (approved):

commands.md
```diff
@@ -1,20 +1,30 @@
...
+
+```ts
+type Doc = string | Doc[] | DocCommand;
+```
+
+- _strings_ are printed directly as is (however for the algorithm to work properly they shouldn't contain line break characters)
+- _arrays_ are used to concatenate a list of docs to be printed sequentially into a single doc
+- `DocCommand` is any of the following:
 
 ### group
```
lipis, line 15:
```suggestion
### `group`
```

commands.md
```diff
@@ -36,17 +46,16 @@ However, if any of the items inside the array have a hard break, the array will
 
 Functions always break after the opening curly brace no matter what, so the array breaks as well for consistent formatting. See [the implementation of `ArrayExpression`](#example) for an example.
 
+The `id` option can be used in [`ifBreak`](#ifBreak) checks.
+
 ### conditionalGroup
```
lipis, line 51:
```suggestion
### `conditionalGroup`
```

commands.md
```diff
@@ -62,31 +71,39 @@ conditionalGroup([a, b, c]);
...
+This is an alternative type of group which behaves like text layout: it's going to add a break whenever the next element doesn't fit in the line anymore. The difference with [`group`](#group) is that it's not going to break all the separators, just the ones that are at the end of lines.
 
 ```js
-fill(["I", line, "love", line, "prettier"]);
+fill(["I", line, "love", line, "Prettier"]);
 ```
 
+Expects the `docs` argument to be an array of alternating content and line breaks. In other words, elements with odd indices must be line breaks (e.g., [`softline`](#softline)).
+
 ### ifBreak
```
lipis, line 82:
```suggestion
### `ifBreak`
```

commands.md
```diff
@@ -62,31 +71,39 @@ conditionalGroup([a, b, c]);
...
-Prints something if the current group breaks and something else if it doesn't.
+Print something if the current `group` or the current element of `fill` breaks and something else if it doesn't.
 
 ```js
 ifBreak(";", " ");
 ```
 
+`groupId` can be used to check another _already printed_ group instead of the current group.
+
 ### breakParent
```
lipis, line 100:
```suggestion
### `breakParent`
```

etc.. since it's code words

thorn0: @lipis Won't this affect existing links to this file?

thorn0: Turns out it won't. Done.

PR #10184, opened 2021-01-31
Playground: Allow hide `input` and `output`

fisker:
## Description

When `show doc`, I think user mean to focus checking `Doc`, don't care "output" that much.

## Checklist

### DIFF
--- a/website/playground/Playground.js
+++ b/website/playground/Playground.js
@@ -242,6 +242,16 @@ class Playground extends React.Component {
                           checked={editorState.showSecondFormat}
                           onChange={editorState.toggleSecondFormat}
                         />
+                        <Checkbox
+                          label="show output"
+                          checked={editorState.showOutput}
+                          onChange={editorState.toggleOutput}
+                        />
+                        <Checkbox
+                          label="show input"
+                          checked={editorState.showInput}
+                          onChange={editorState.toggleInput}
+                        />
                         {editorState.showDoc && debug.doc && (
                           <ClipboardButton
                             copy={() => this.getMarkdown({ doc: debug.doc })}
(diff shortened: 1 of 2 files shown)


thorn0:
It's often important to see what output the doc generates. The user changes the input and immediately can see how it affects both the doc and the output. That's how I often use it. Sometimes small changes in the doc, trigger not very expected changes in the output, and it's better to see that. If you want to hide the output, add an independent checkbox for that.

thorn0:
Is it for mobile too?

fisker:
I just want more space for the panel I'm focusing, yes better on mobile too .

website/playground/Playground.js
```diff
@@ -242,6 +242,16 @@ class Playground extends React.Component {
...
+                        <Checkbox
+                          label="show output"
+                          checked={editorState.showOutput}
+                          onChange={editorState.toggleOutput}
+                        />
+                        <Checkbox
+                          label="show input"
+                          checked={editorState.showInput}
+                          onChange={editorState.toggleInput}
+                        />
```
thorn0, line 254: Maybe let's change the order of the checkboxes to match the order of the panels?

fisker:
I was thinking that these two buttons won't be used as 
 much as others.

website/playground/Playground.js
fisker: I was thinking that these two buttons won't be used as
much as others.

thorn0: That's true, but the order looks strange. Let's change it.

fisker: I'm happen to be an expert to  cure OCD, want help? I can give you discount. 😄

fisker: Can you edit it? I'm not near my laptop.

PR #10235, opened 2021-02-03
Safer check before passing to `fill`

fisker:
## Description

Ref https://github.com/prettier/prettier/pull/9847#discussion_r567485818

## Checklist


thorn0:
What about this part

> Expects the docs argument to be an array of alternating content and line breaks. In other words, elements with odd indices must be line breaks

?

fisker:
In that case, all `concat` doc parts can't turn into `fill`, no matter it from `normalizeDoc` or `cleanDoc`.

fisker:
I don't know what to do, let's make sure they are `array` first?

thorn0:
> In that case, all concat doc parts can't turn into fill

Sometimes (e.g. the result of `join(line, docs)`), it's safe. But overall - yes, it's not safe.

> no matter it from normalizeDoc or cleanDoc.

Before being passed to `normalizeDoc` or `cleanDoc`, the array might have been constructed in a special way for `fill`.
E.g., something like `[['a', 'b'], line, ['c', 'd']]`.

thorn0 (approved):
> let's make sure they are array first

OK

sosukesuzuki (approved):

PR #10245, opened 2021-02-04
Refactor `ChainElement` related parts

fisker:
## Description

- Rename `isCallOrOptionalCallExpression` to `isCallExpression`, they are all `CallExpression` in ESTree
- Add `isMemberExpression`
- Replace related checks

## Checklist

### DIFF
--- a/src/language-js/utils.js
+++ b/src/language-js/utils.js
@@ -102,8 +102,7 @@ function hasNakedLeftSide(node) {
     node.type === "NGPipeExpression" ||
     node.type === "ConditionalExpression" ||
     isCallExpression(node) ||
-    node.type === "MemberExpression" ||
-    node.type === "OptionalMemberExpression" ||
+    isMemberExpression(node) ||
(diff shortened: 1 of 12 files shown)


thorn0 (approved):

src/language-js/utils.js
```diff
@@ -352,15 +349,14 @@ function isTheOnlyJsxElementInMarkdown(options, path) {
  * @returns {boolean}
  */
 function isMemberExpressionChain(node) {
```
thorn0, line 351: Do we want to consider `TSNonNullExpression` part of "member expression chain" in this function?
E.g.: `a.b!.c`
If we do, is a simple expression like `a!`? I guess no. But what about `a.b.c!` (a chain _wrapped_ in a non-null assertion)?

thorn0: This function is used only for assignments, so you can ignore my comment above. I've solved that in #10222.

sosukesuzuki (approved):

PR #10248, opened 2021-02-05
Check `fill` parts

fisker:
## Description

> Expects the docs argument to be an array of alternating content and line breaks. In other words, elements with odd indices must be line breaks

Ref: https://github.com/prettier/prettier/pull/10235#issuecomment-773209893

## Checklist

### DIFF
--- a/src/document/doc-builders.js
+++ b/src/document/doc-builders.js
@@ -140,8 +140,15 @@ function conditionalGroup(states, opts) {
  */
 function fill(parts) {
   if (process.env.NODE_ENV !== "production") {
-    for (const part of parts) {
+    for (const [index, part] of parts.entries()) {
       assertDoc(part);
+      if (index % 2 === 1 && part.type !== "line") {
+        const type = part.type || (Array.isArray(part) ? "array" : "string");
+        console.log(parts);
+        throw new Error(
+          `elements with odd indices expected to be line breaks, got "${type}".`
+        );
+      }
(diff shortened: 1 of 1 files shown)


src/document/doc-builders.js
```diff
@@ -140,8 +140,15 @@ function conditionalGroup(states, opts) {
  */
 function fill(parts) {
   if (process.env.NODE_ENV !== "production") {
-    for (const part of parts) {
+    for (const [index, part] of parts.entries()) {
       assertDoc(part);
+      if (index % 2 === 1 && part.type !== "line") {
```
thorn0, line 145: It's not that simple. It also can be something like `["{' '}", softlline]`. The idea is that it contains a line break.

fisker: Did you mean `["{", "}", softline]`?

thorn0: Anything with a line break inside

fisker: Oh, you mean "elements with odd indices" only need a line break. 

```js
{
	  type: "fill", 
	  parts: [
		    "a", 
		    ["{' '}", softlline], // <- this
		    "b"
	  ]
}
```

Got it.

thorn0: Yes. Even-indexed ones can contain anything, but odd-indexed must contain at least one line break.

PR #10256, opened 2021-02-05
Don't skip file paths contains special words like `constructor`

ashlkv:
## Description

After updating to v2 prettier cli started throwing a "No matching files" error for paths like "js/constructor/*.js" in my team's project. It turned out that the `containsIgnoredPathSegment` check was returning a positive for the paths containing "constructor" segment, and was ignored.

This pull request fixes the issue by additionally checking the path segments against the silently ignored path dictionary with a `hasOwnProperty`.

## Checklist


src/cli/expand-patterns.js
```diff
@@ -163,7 +163,11 @@ function containsIgnoredPathSegment(absolutePath, cwd, ignoredDirectories) {
   return path
     .relative(cwd, absolutePath)
     .split(path.sep)
-    .some((dir) => ignoredDirectories[dir]);
+    .some(
```
fisker, line 166: Can you pass a `Set` to this function instead of `object`?

fisker: I mean turn `silentlyIgnoredDirs` to `Set`

fisker:
This bug is hilarious! 👍

fisker:
@thorn0 I checked your recently committed code https://github.com/prettier/prettier/blob/a8f35961c066de440ceb31c9b7414c88cf91e7af/src/document/doc-debug.js#L184, glad you use `Object.create(null)`, haha

thorn0:
@fisker The CLI part of the project isn't my cup of tea. :)

fisker:
You are not alone, but this bug is on you, [it was `array`](https://github.com/prettier/prettier/pull/7660/files#diff-a1489d380e2e3986584e73d3f58312a0cb45e28853116312abc1d801474bd6b3L439), 😄

src/cli/expand-patterns.js
ashlkv: Done in https://github.com/prettier/prettier/pull/10256/commits/8fbf36f72c2d1eb98f57039de54ffe1f83acfad3

fisker (approved):

thorn0:
Changelog entry?

PR #10273, opened 2021-02-08
Test: Skip install for standalone test

fisker:
## Description

standalone test eval from code, no need install, just read file from `dist/`

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,8 +10,8 @@ const TEST_STANDALONE = Boolean(process.env.TEST_STANDALONE);
 const INSTALL_PACKAGE = Boolean(process.env.INSTALL_PACKAGE);
 
 let PRETTIER_DIR = isProduction
-  ? PROJECT_ROOT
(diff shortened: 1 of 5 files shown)


thorn0:
Why do we need the install for non-standalone production tests?

fisker:
I don't have an answer, I thought about this before, maybe prevent unexpected things like plugin searching?

Sometimes I even disable install when I run production test on my machine.

thorn0 (approved):

duailibe (approved):

duailibe:
> Why do we need the install for non-standalone production tests?

We had a bug that after the "bundle everything into a single file" step, there were some `require("../../file")` left, that would still work when running inside the Prettier folder. So we started to test the production build this way (install the final package, in a tmp folder)

fisker:
> there were some require("../../file") left, that would still work when running inside the Prettier folder.

Wow, yes, this make sense.

Issue #9719, opened 2020-11-19
`options` problems


fisker:
1. Options not varified before passing to `loadPlugins`

	https://github.com/prettier/prettier/blob/7acaa554e92330fabae038b73048efc5384df65f/src/index.js#L21
	
	```js
	require('.').format('a', {pluginSearchDirs: {length: 2}})
	Uncaught TypeError: pluginSearchDirs.map is not a function
	```
	
	```js
	require('.').format('a', {plugins: {length: 2}})
	Uncaught TypeError: Cannot read property 'languages' of undefined
	```
	
	```js
	require('.').format('a', {pluginSearchDirs: [1]})
	Uncaught:
	TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type number (1)`
	```

2. `pluginSearchDirs` should not allow object

	https://github.com/prettier/prettier/blob/7acaa554e92330fabae038b73048efc5384df65f/src/main/core-options.js#L188

	```js
	require('.').format('a', {pluginSearchDirs: [{}]})
	Uncaught:
	TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Object
	```

3. There is no way to disable plugins auto-search

	Maybe we can use `{pluginSearchDirs:[]}` or `{pluginSearchDirs: false}` to disable, but empty array maybe can't use for CLI, but I don't know how to make [this option](https://github.com/prettier/prettier/blob/7acaa554e92330fabae038b73048efc5384df65f/src/main/core-options.js#L176-L191) accept `boolean`

	A workaround for now `{pluginSearchDirs:['/']}`

PR #10274, opened 2021-02-08
Add `--no-plugin-search`

fisker:
## Description

Ref: #9719

Add `--no-plugin-search` to disable plugin search, current implementation is in a dirty way.

~And I don't know how to test it.~ tested.

## Checklist


docs/plugins.md
```diff
@@ -33,7 +35,7 @@ When plugins cannot be found automatically, you can load them with:
 
 Prettier expects each of `pluginSearchDirs` to contain `node_modules` subdirectory, where `@prettier/plugin-*`, `@*/prettier-plugin-*` and `prettier-plugin-*` will be searched. For instance, this can be your project directory or the location of global npm modules.
 
-Providing at least one path to `--plugin-search-dir`/`pluginSearchDirs` turns off plugin autoloading in the default directory (i.e. `node_modules` above `prettier` binary).
+To turns off plugin autoloading, use `{pluginSearchDirs: false}`.
```
kachkaev, line 38:
```suggestion
To turn off plugin autoloading, use `{ pluginSearchDirs: false }`.
```

kachkaev (commented):
Would it be possible to accept `--plugin-search-dir=` as ‘turn off’? I guess we‘ll be getting `[""]`, which can be treated as special case. In theory, we can also accept `["false"]` via `--plugin-search-dir=false` but I’m ess sure about this option. Non-CLI Prettier can accept literal `false`, which will be treated just as `[""]`.

But maybe I’m missing something 🤔

fisker:
We use `--no-xxx` for all `false` flags, `--no-color`, `--no-config`, `--no-editorconfig`, `--no-error-on-unmatched-pattern`, `--no-bracket-spacing`, `--no-semi`

kachkaev (approved):
👏

alexander-akait:
For me the naming a little misleading too, I understand why you gave that name and how it is implemented, but for DX it is sounds very misleading, maybe just `--no-plugins`?

kachkaev:
I guess `--no-plugin-search-dir` still allows adding `--plugin=./foo-plugin`. This makes sense to me. Maybe we can cover this scenario with a test to avoid possible regressions in the future?

fisker:
> I guess --no-plugin-search-dir still allows adding --plugin=./foo-plugin.

Yes, still allow.

fisker:
Maybe `--no-plugin-search`?

alexander-akait:
Sounds better

fisker:
> Maybe we can cover this scenario with a test to avoid possible regressions in the future?

Tested https://github.com/prettier/prettier/pull/10274/commits/28bb4352871aa07cbd4bcfdfcb17a7e7e1dbedcf

fisker:
Rename to `--no-plugin-search` is much harder than I thought, I'll try later.

fisker:
Our CLI part of code like a mess... But seems working now.

docs/plugins.md
```diff
@@ -33,7 +35,7 @@ When plugins cannot be found automatically, you can load them with:
 
 Prettier expects each of `pluginSearchDirs` to contain `node_modules` subdirectory, where `@prettier/plugin-*`, `@*/prettier-plugin-*` and `prettier-plugin-*` will be searched. For instance, this can be your project directory or the location of global npm modules.
 
-Providing at least one path to `--plugin-search-dir`/`pluginSearchDirs` turns off plugin autoloading in the default directory (i.e. `node_modules` above `prettier` binary).
+To turn off plugin autoloading, use `{ pluginSearchDirs: false }`.
```
kachkaev, line 38:
```suggestion
To turn off plugin autoloading, use `{ pluginSearchDirs: false }` in JavaScript code or `--no-plugin-search` when using Prettier CLI.
```

Not sure about exact wording, but maybe it’s a bit better than having `pluginSearchDirs: false` separately from `--no-plugin-search`? WDYT?

kachkaev:
Thanks @fisker! One other edge case that came to my mind is using `--no-plugin-search` together with `--plugin-search-dir`. Maybe we want some early exit in this case, but not sure if it’s worth it or too much. Running `prettier --config=abc --no-config .` does not crash for me 🤔

Either way, I guess we need a CHANGELOG entry for this PR 🙂

fisker:
> One other edge case that came to my mind is using --no-plugin-search together with --plugin-search-dir.

I did try to add this check at first place, but due to our weird CLI arguments parsing. It's not easy to add, anyway https://github.com/prettier/prettier/pull/10274/commits/7862a53dfcb4821a1fd9f26b1783e7e2b74506c2

changelog_unreleased/cli/10274.md
```diff
@@ -0,0 +1,14 @@
+#### Add `--no-plugin-search` to turn off plugin autoloading (#10274 by @fisker)
+
+To turn off plugin autoloading, use `{ pluginSearchDirs: false }` in JavaScript code or `--no-plugin-search` when using Prettier CLI.
+
+If you don't use any plugins, you can also add `{ pluginSearchDirs: false }` to your config file.
```
kachkaev, line 5:
```suggestion
To turn off plugin autoloading, use `{ pluginSearchDirs: false }` in JavaScript code or `--no-plugin-search` when using Prettier CLI. You can also add `{ pluginSearchDirs: false }` to your config file.
```

Use case:

```js
// prettier.config.js

module.exports = {
  trailingComma: "all",

  plugins: [
    require("prettier-plugin-packagejson"),
    require("prettier-plugin-sh"),
  ],
  pluginSearchDirs: false
};
```

Adding `pluginSearchDirs` does not mean that I ‘don't use any plugins’.

---

Side thought: Perhaps, `no-plugin-search` mode can become default default behaviour in v3 because it’s incompatible with PnP. That’s something to discuss separately I guess.

fisker: > no-plugin-search mode can become default default behaviour in v3 

I don't known, but I'm going to add into my config, and I'm planning to use it in our codebase including tests.

bernardobelchior:
I was just searching for this option, thanks for this implementation 🙏 
Hope it gets merged soon, thank you!

fisker:
@prettier/core Can I get more approves?

changelog_unreleased/cli/10274.md
```diff
@@ -0,0 +1,12 @@
+#### Add `--no-plugin-search` to turn off plugin autoloading (#10274 by @fisker)
+
+To turn off plugin autoloading, use `{ pluginSearchDirs: false }` in JavaScript code or `--no-plugin-search` when using Prettier CLI. You can also add `{ pluginSearchDirs: false }` to your config file.
```
duailibe, line 3: What does ``use `{ pluginSearchDirs: false }` in JavaScript code`` mean?

duailibe (approved):

changelog_unreleased/cli/10274.md
kachkaev: I guess it does the same as `--no-plugin-search`.

duailibe: Oh I meant the "in JavaScript code" part. Because later it says to add that to the config file, so I didn't understand what "JavaScript code" is this

kachkaev: Oh IC IC 😅 Maybe it refers to using Prettier as a JS library, not via CLI?

duailibe: Aaahhhh that makes sense! Maybe something like:

> To turn off plugin autoloading, use `--no-plugin-search` when using Prettier CLI or add `{ pluginSearchDirs: false }` to options in `prettier.format()` or to the config file.

kachkaev (approved):

PR #10275, opened 2021-02-08
RFC: label doc

thorn0:
## Description

Introduces a new doc builder: `label(label: string, doc: Doc): Doc`

It marks a doc with a string label. This doesn't affect how the doc is printed, but can be useful for heuristics based on doc introspection.

E.g., to decide how to print an assignment expression, we might want to know whether its right-hand side has been printed as a method call chain, not as a plain function call. If the method chain printing code uses `label` to mark its result, checking that condition can be as easy as `rightHandSideDoc.label === 'method-chain'`.

## Checklist

### DIFF
--- a/src/language-js/print/member-chain.js
+++ b/src/language-js/print/member-chain.js
@@ -24,7 +24,15 @@ const {
 const { locEnd } = require("../loc");
 
 const {
-  builders: { join, hardline, group, indent, conditionalGroup, breakParent },
+  builders: {
+    join,
+    hardline,
+    group,
+    indent,
+    conditionalGroup,
+    breakParent,
+    label,
+  },
   utils: { willBreak },
 } = require("../../document");
 const printCallArguments = require("./call-arguments");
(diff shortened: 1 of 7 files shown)


duailibe (approved):
I like it

fisker:
Can you check `doc-utils.js`, I think some function need update, example `stripDocTrailingHardlineFromDoc`

src/language-js/print/member-chain.js
```diff
@@ -389,16 +399,18 @@ function printMemberChain(path, options, print) {
...
   }
 
-  return [
-    // We only need to check `oneLine` because if `expanded` is chosen
-    // that means that the parent group has already been broken
-    // naturally
-    willBreak(oneLine) || shouldHaveEmptyLineBeforeIndent ? breakParent : "",
-    conditionalGroup([oneLine, expanded]),
-  ];
+  return label("member-chain", result);
```
fisker, line 413: Do you think `label(result, "member-chain")` better.

thorn0: `result` can be long. On the playground, it'll be more convenient to see the string immediately. Also this order matches `join`.

thorn0: image

fisker (approved):

PR #10276, opened 2021-02-08
Refactor: rename FastPath → AstPath

thorn0:
## Description

`FastPath` is a misleading name. Is there also a `SlowPath`? 

## Checklist

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -177,7 +177,7 @@ The printing process works as follows:
 
 1. `preprocess(ast: AST, options: object): AST`, if available, is called. It is passed the AST from the _parser_. The AST returned by `preprocess` will be used by Prettier. If `preprocess` is not defined, the AST returned from the _parser_ will be used.
 2. Comments are attached to the AST (see _Handling comments in a printer_ for details).
-3. A Doc is recursively constructed from the AST. i) `embed(path: FastPath, print, textToDoc, options: object): Doc | null` is called on each AST node. If `embed` returns a Doc, that Doc is used. ii) If `embed` is undefined or returns a falsy value, `print(path: FastPath, options: object, print): Doc` is called on each AST node.
+3. A Doc is recursively constructed from the AST. i) `embed(path: AstPath, print, textToDoc, options: object): Doc | null` is called on each AST node. If `embed` returns a Doc, that Doc is used. ii) If `embed` is undefined or returns a falsy value, `print(path: AstPath, options: object, print): Doc` is called on each AST node.
 
 #### `print`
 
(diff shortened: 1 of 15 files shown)


fisker:
How about `TreePath`? Some parser may use `cst`

thorn0:
The thing is called "AST" everywhere on Prettier's site and in the codebase. Calling it "tree" here would be inconsistent. Also we have other tree structures, e.g. docs IR, but this class isn't intended for them (it can work with docs, of course, but it's not the intention).

docs/plugins.md
```diff
@@ -177,7 +177,7 @@ The printing process works as follows:
 
 1. `preprocess(ast: AST, options: object): AST`, if available, is called. It is passed the AST from the _parser_. The AST returned by `preprocess` will be used by Prettier. If `preprocess` is not defined, the AST returned from the _parser_ will be used.
 2. Comments are attached to the AST (see _Handling comments in a printer_ for details).
-3. A Doc is recursively constructed from the AST. i) `embed(path: FastPath, print, textToDoc, options: object): Doc | null` is called on each AST node. If `embed` returns a Doc, that Doc is used. ii) If `embed` is undefined or returns a falsy value, `print(path: FastPath, options: object, print): Doc` is called on each AST node.
+3. A Doc is recursively constructed from the AST. i) `embed(path: AstPath, print, textToDoc, options: object): Doc | null` is called on each AST node. If `embed` returns a Doc, that Doc is used. ii) If `embed` is undefined or returns a falsy value, `print(path: AstPath, options: object, print): Doc` is called on each AST node.
```
brodybits, line 180: Could this break any plugins depending on a `FastPath` type?

thorn0: No, it's a local variable name. It doesn't survive bundling.

brodybits: Thanks. I guess `AstPath` would be some kind of an opaque type  from the API perspective, then.

brodybits, line 180: Sorry if this is off-topic but I was wondering if this text should be split up like this, for example:

```suggestion
3. A Doc is recursively constructed from the AST.
  i) `embed(path: AstPath, print, textToDoc, options: object): Doc | null` is called on each AST node. If `embed` returns a Doc, that Doc is used.
  ii) If `embed` is undefined or returns a falsy value, `print(path: AstPath, options: object, print): Doc` is called on each AST node.
```

thorn0: Who knows? At least, "i)" isn't valid syntax for lists in Markdown. Anyway, feel free to reformat it to your heart's content and send a PR.

fisker (approved):

PR #10290, opened 2021-02-10
Move Handlebars support out of alpha to release 

dcyriller:
## Description

🎉

I have added a changelog entry, it is largely stolen from HTML and mardown release blog posts. I need to detail it a bit. Please feel free to modify it or to correct the potential spelling and syntax issues, I'm not an English native speaker.

## Checklist

### DIFF
--- a/changelog_unreleased/handlebars/10290.md
+++ b/changelog_unreleased/handlebars/10290.md
@@ -1,51 +1,18 @@
-#### Move Handlebars support out of alpha to release (#10290 by @dcyriller)
+#### Move Handlebars support out of alpha to release (#10290 by @dcyriller & @thorn0)
 
-Prettier can now format Handlebars files! 🎉
+This started in 2017. Handlebars support has been in Prettier for a while, but it wasn’t released officially as it wasn’t really ready. Its status was “alpha”, “experimental”, “beta”. Then, if you checked older release notes, you could see that after “beta” somehow it was “alpha” again…
 
-We use [@glimmer/syntax](https://www.npmjs.com/package/@glimmer/syntax), the
-Ember Handlebars parser, to parse these Handlebars files so it should be
-compliant with the HTML spec thanks to the Ember team.
+Well, anyway, it finally is happening: Prettier can now officially format Handlebars files! 🎉
 
-A few highlights:
+It uses [Glimmer], [Ember]’s Handlebars parser, so it should be compliant with the HTML spec thanks to the Ember team.
 
-##### Whitespace-sensitive formatting
+The [`--html-whitespace-sensitivity`][hws-option] option is supported and defaults to `strict`, which means that Prettier will always respect the presence or the lack of whitespace around tags and consider it unsafe to add whitespace where there were none and vice versa as this can affect how the document is rendered in the browser. The `css` value is not yet supported (treated as `strict` for now).
 
-As you may notice during daily Handlebars works, the following two cases won't
-produce the same output:
+The feature is called “Ember / Handlebars” and not just “Handlebars” because Glimmer doesn’t support some syntax and use cases of Handlebars. This is mostly due to the fact that [Handlebars], being a template engine (a preprocessor), doesn’t care about the underlying syntax of the content it processes whereas Glimmer parses two syntaxes – HTML and Handlebars – at the same time and combines the result into a single tree, which Prettier can print. This means Prettier won’t format Handlebars files that can’t be parsed into such a tree, either because the underlying syntax isn’t HTML or because template directives and tags overlap in a way that can’t be represented in a tree (e.g., `{{#if foo}}<div>{{/if}`).
 
-|                |         html         |       output       |
-| -------------- | :------------------: | :----------------: |
-| with spaces    | `1<span> 2 </span>3` | 1<span> 2 </span>3 |
-| without spaces |  `1<span>2</span>3`  |  1<span>2</span>3  |
+Hopefully, the formatter will still be useful enough to non-Ember Handlebars users.
 
-This happens because whitespace is significant in inline elements.
-
-For this reason, we cannot safely format
-
-<!-- prettier-ignore -->
-```hbs
-<a href="https://prettier.io/" class={{this.classes}}>Prettier is an opinionated code formatter.</a>
-```
-
-into
-
-<!-- prettier-ignore -->
-```hbs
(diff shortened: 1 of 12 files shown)


thorn0:
@jelhan Hi 👋 Can you probably help with improving text for the release blog post? Any additional important points for the Ember community?

dcyriller:
I'd be more confortable if we could cut another alpha release, to a have a test phase for the whitespace sensitive mode. Then, in a latter release we could merge this.

thorn0:
It totally looks good enough for a release. The first changes in `src/language-handlebars` were committed on 2017-12-29. 😲 It's time to put an end to this alpha thing. Of course, there will be issues, but other languages have them too. It's okay.

jelhan:
> @jelhan Hi 👋 Can you probably help with improving text for the release blog post? Better wording, additional important points for the Ember community, less similarity with previous posts :) Any help is appreciated.

Sure. Let me see what I can do. 👍

Is there a deadline? I'm not sure how much time I will have the next days.

thorn0:
Cool. No deadline for now. The team needs time too to tie up some loose ends.

thorn0:
An important point to mention in the blog post: we support only a subset of Handlebars. See https://github.com/prettier/prettier/issues/5340

jelhan:
> An important point to mention in the blog post: we support only a subset of Handlebars. See #5340

What is implemented is a parser for the template engine used by Ember. Ember stopped calling its template engine Handlebars in its own documentation. In December 2018 the page _Handlebars Basics_ in _Templates_ section was renamed to _Templating Basics_. (https://github.com/ember-learn/guides-source/pull/324/files) Since then the term Handlebars is not prominent anymore in Ember's documentation itself. 

The template syntax used Ember's Glimmer VM and the [Handlebars.js](https://handlebarsjs.com/) diverged a lot over the last years. I guess this is reflected by less using the term.

I wonder if marketing the new language as _Handlebars_ is very helpful long term. Maybe we should call it better _Glimmer templates_?

There has been similar discussions at many places in the Ember community. So far there wasn't a big need to give the templates syntax an explicit name. The community just stayed with `.hbs` extension but stopped talking about it as Handlebars. At least that's my feeling. I guess for marketing the Prettier feature we need to settle on a name.

I can ask for input of Ember learning team on the question if you like. It's very likely that there are some discussions I'm not aware of.

thorn0:
@jelhan It boils down to how useful the formatter can be to non-Ember Handlebars users. If they can't really benefit from it, and there is no chance for this to be fixed (within reason, I don't mean crazy tag soup cases like `{{#if foo}}<div>{{/if}`), then the feature should be just called "Ember", similarly to Vue and Angular. Otherwise, it might be something like "Ember/Handlebars". Anyway, the Glimmer name is probably too little-known.

> I can ask for input of Ember learning team on the question if you like. It's very likely that there are some discussions I'm not aware of.

Sounds good. Let's ask them.

jelhan:
> The template syntax used Ember's Glimmer VM and the [Handlebars.js](https://handlebarsjs.com/) diverged a lot over the last years.

I fear I was wrong at this point. I talked with @locks from Ember learning team on Ember Discord. He explained me that Ember's template syntax is still valid handlebars. If I got it right Glimmer parser is even build on top of handlebars parser. There were some fundamental changes to Ember's template syntax. But it kept to be valid handlebars syntax.

dcyriller:
As far as I understand it...

Handlebars.js (the library) doesn’t parse HTML. Instead it treats HTLM elements as bare text.

On some Handlebars syntax, Glimmer VM will throw errors because they are not supported. For instance: blocks partial or handlebars decorators.

In an Handlebars file, Glimmer VM will parse Handlebars syntax with Handlebars.js (and crash on non-supported features). And it will parse HTML syntax with simple-html-tokenizer.

So I think a better name for this language supported by Prettier could be ember-template. We could rename src/language-handlebars to src/language-ember-template.
By the way, ember-template-lint is the name of main linter for these files.

(sorry for the missing links I’m on my phone)

thorn0:
Glimmer [indeed][1] uses the Handlebars parser under the hood. This leaves a chance that support for unsupported things - at least for the most used ones - still can be added, albeit this chance is slim and might involve forking. If somebody from the non-Ember Handlebars community decides to add that support, it'd be much less work than creating a solution from scratch. Looks like the two most wanted missing pieces are these:

* partials, see https://github.com/glimmerjs/glimmer-vm/issues/1233
* blocks around attributes, https://github.com/glimmerjs/glimmer-vm/issues/1251

So my vote is for "Ember / Handlebars", with an explanation of this nuance in the blog post.

> Handlebars.js (the library) doesn’t parse HTML. Instead it treats HTLM elements as bare text.

That's why I mentioned "tag soup". Formatting of files that can't be parsed into a syntax tree will, naturally, never be supported by Prettier. I believe it's not a deal breaker.

> better name for this language supported by Prettier could be ember-template

If "Ember" gets opted out for, it should be just "Ember". For simplicity and because that's how it's been done for Angular and Vue. We refer to the corresponding features simply as "Angular support" and "Vue support".

image


[1]: https://github.com/glimmerjs/glimmer-vm/blob/1a54cba8a6b4175cb2d5a617ff857b616cf6f0b7/package.json#L46

thorn0:
@jelhan Please see my updates in this PR. What would you add to the release notes?

thorn0:
I'm going to merge this as is. Feel free to open a new PR to propose change to the release notes.

jelhan:
> @jelhan Please see my updates in this PR. What would you add to the release notes?

I just gave it a read. And it's great. Haven't found anything that I would change.

Only additions would be very Ember specific. Like the available plugin for ember-template-lint. I don't think this belongs in prettier Changelog. Or do you regular mention ecosystem for new languages added?

fisker (approved):
🎉🎉🎉

thorn0:
> Only additions would be very Ember specific. Like the available plugin for ember-template-lint. I don't think this belongs in prettier Changelog. Or do you regular mention ecosystem for new languages added?

There is a paragraph about non-Ember usage. Having a short useful paragraph about Ember would make sense too. E.g. [this fact](https://mobile.twitter.com/jelhan1/status/1347292468212080640) can be mentioned.

Alonski:
So happy to have this merged! Hope we can get a new Prettier release soon. 🎉

PR #10296, opened 2021-02-10
Refactor decorators print

fisker:
## Checklist


src/language-js/printer-estree.js
```diff
@@ -95,108 +91,52 @@ const { printSwitchCaseConsequent } = require("./print/statement");
...
-    needsParens = pathNeedsParens(path, options);
+    return linesWithoutParens;
   }
 
-  const parts = [];
+  const decorators = printDecorators(path, options, printPath);
+  // Nodes with decorators can't have parentheses
+  const needsParens = !decorators && pathNeedsParens(path, options);
+
+  const parts = [linesWithoutParens];
```
thorn0, line 123: Let's avoid returning an array when possible.

fisker: What do you mean? It suppose to be a `concat`, even if it doesn't need parens, it can still be a `concat` returned by another function,  why other type? You can use `if (isLabel(doc, ...)) {}` to simplify your check.

thorn0: In general I agree that the check should be more flexible.

> It suppose to be a concat, even if it doesn't need parens

But I'm not sure about this. Why create an array for each node?

src/language-js/printer-estree.js
```diff
@@ -94,108 +91,53 @@ const { printSwitchCaseConsequent } = require("./print/statement");
...
-    // Nodes with decorators can't have parentheses, so we can avoid
-    // computing pathNeedsParens() except in this case.
-    needsParens = pathNeedsParens(path, options);
+    return printed;
   }
 
-  const parts = [];
+  const decorators = printDecorators(path, options, print);
+  // Nodes with decorators can't have parentheses
+  const needsParens = !decorators && pathNeedsParens(path, options);
```
thorn0, line 121: Note that `decorators` and `needsParens` are mutually exclusive.

fisker: Am I doing something wrong?

src/language-js/printer-estree.js
fisker: Fixed.

src/language-js/print/decorators.js
```diff
@@ -0,0 +1,81 @@
...
+      hasNewlineBetweenOrAfterDecorators(node, options);
+
+    const separator = shouldBreak ? hardline : line;
+
+    path.each((decoratorPath) => {
+      let decorator = decoratorPath.getValue();
+      if (decorator.expression) {
+        decorator = decorator.expression;
+      } else {
+        decorator = decorator.callee;
```
brodybits, line 41: This looks like untested code according to Codecov (??)

fisker: Should be something about

```
@decorator
```

and

```
@decorator()
```

Not sure.

thorn0: This is just unused remnants of some old code.

fisker:
I thought there are more things I can do in this PR, but that's all.

src/language-js/print/decorators.js
fisker: I'm not sure if it can be removed.

brodybits: I'll try to contribute a test for this when I find a spare cycle,

fisker: I checked `flow` `babel` and `typescript` parser, seems they all use `.expression`, I removed this line.

fisker: I checked `flow` `babel` and `typescript` parser, seems they all use `.expression`, I removed this line.

src/language-js/print/decorators.js
```diff
@@ -0,0 +1,73 @@
...
+    const parts = [];
+    if (parentExportDecl) {
+      parts.push(hardline);
+    }
+
+    path.each((decoratorPath) => {
+      parts.push(print(decoratorPath, "expression"), separator);
+    }, "decorators");
+
+    return parts;
```
thorn0, line 42:
```suggestion
    return [
      parentExportDecl ? hardline : shouldBreak ? breakParent : "",
      join(line, path.map(print, "decorators")),
      line,
    ];
```

src/language-js/print/decorators.js
```diff
@@ -0,0 +1,73 @@
...
+      (decoratorPath) => {
+        const decorator = decoratorPath.getValue();
+        const prefix = decorator.type === "Decorator" ? "" : "@";
+        parts.push(prefix, print(decoratorPath), hardline);
+      },
+      "declaration",
+      "decorators"
+    );
+
+    return parts;
```
thorn0, line 67:
```suggestion
    // Export declarations are responsible for printing any decorators
    // that logically apply to node.declaration.
    return [
      join(hardline, path.map(print, "declaration", "decorators")),
      hardline,
    ];
```

src/language-js/print/decorators.js
thorn0: Printing `line` in `case "Decorator":` in `printer-estree.js` would be even better.

src/language-js/print/decorators.js
fisker: No need check `decorator.type`?

fisker: I guess `decorator.type` was used togethor with `decorator.callee`, indeed I did't see this in Type Definition

thorn0: Yes, leftovers from old times

fisker:
There is another `printDecorators` function in `class.js`, I'll see if we can merge it.

thorn0:
Why is it still draft? You think we could have broken something?

thorn0 (approved):

src/language-js/print/decorators.js
```diff
@@ -0,0 +1,93 @@
...
+    join(line, path.map(print, "decorators")),
+    line,
+  ];
+}
+
+function hasNewlineBetweenOrAfterDecorators(node, options) {
+  const start = locEnd(node.decorators[0]);
+  const end = locEnd(getLast(node.decorators));
+  return (
+    hasNewlineInRange(options.originalText, start, end) ||
```
fisker, line 66: @thorn0 I think this function is not right,

**Prettier 2.2.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgII7AA6UOOAAgCZyQBO6MEtAFMAL4CUOADrRN3FowAnjgC8OAIwBuEmxIkwGbPkIkyVGkwZNW6sjn2cefAUNESZckABoQ-GAEtomZKHS0+AdwAKHhK4o6Khe6MKudgBG9GAA1nAwAMrc6GCOUADmyDC0AK5wdnAAtpFwlNSUADLombnoGXAAYkxFDE6ZyCDouYy2IAAWMEWoAOr9jvCYKWBwiQETjgBuE8KdYNh96ZiCMD70Ga3IAGbB23YAVpgAHgBCMfFJ6EVwlelwx6cFIJdXiekZqDgAEVchB4B9UGcQClaNtaJ1IuhSqg+rx0jARo5KDB+sgABwABjsvAg2xG9G4nV4cDhi3edgAjqD4Ht+IEupgALRQOBlMp9WhwJmOQV7eqHJAnSFfbZFRzZPIy-6AkFg96Sz52GBIzHY3FIABMWvojlQ-wAwhAihKQDSAKx9XLbAAqSMCUqhi3yAEkoNRYIkwLRHNwYHg-YkRICIds2GwgA)

```sh
--parser babel
```

**Input:**

```jsx
class A {
  @decorator({}) property = 1;
}

class A {
  @decorator({
    
  }) property = 1;
}
```

**Output:**

```jsx
class A {
  @decorator({}) property = 1;
}

class A {
  @decorator({})
  property = 1;
}

```

thorn0 (approved):

PR #10297, opened 2021-02-11
RFC: break method chains in assignments if they don't fit after operator

thorn0:
## Description

**Proposal**: if the right-hand side of an assignment is a member chain, then:
- If the entire assignment fits on one line, keep it that way.
- If the member chain doesn't fit after the operator, it's moved to the next line *and is always split onto multiple lines* at that.

**Input:**
```jsx
const kochabCooieGameOnOboleUnweaveWayWay = someLongString.split("jest version =").pop().split(EOL)[0].trim();
```

**Current output ([playground](https://deploy-preview-10222--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuc0DOMAEBrCYAWAhgEYDCEEAlnAOKEC2cA8lE8RADZwCqUA7nEIA3OAHVCAT3ETMAXkxoIjADLQA5gGUYAJ0pQ1AOjQAHDpRgAKADogAVnAyYR2tJWhybASgPGIxi94mZpYAokzKngDaAAwAugY6lPQBANwgADQgfjBuUGjIoITa2hB8AApFCPkohBx8kvmZxNqEYNhwMBrGrXpqyDoArnCZcPTEcAAmE5PKhPoDhGpwAGIQ2vSEMDn6yCCEAzAQGSD4MPQcovjmDt1gcBpV5pRC5hK7YGiNIHpocNowZRaag2yAAZrVfplbGgAB4AIRabQ6GgYcGUejgYIhwzssI0vS4AEUBhB4FiOJCQN0XH9dsQSHAOMdjLpYKJKBMYPhkAAOaKZFkQX6iFrGXYshx-ETHACOJPggL81T2aAAtFA4JNJsdtHA5ZRdYDFiCkOCKTjfvRKP1tENMq59ET5ZjTdjMjASOzOdykAAmd0tShmfTkegmkAOACsxwGvwAKiRqmbKUIhgBJKDTWAaMC6YwwACCma0Ei45N+AF8K0A)):**
```jsx
const kochabCooieGameOnOboleUnweaveWayWay = someLongString
  .split("jest version =")
  .pop()
  .split(EOL)[0]
  .trim();
```

[...]

Issue #5610, opened 2018-12-07
Call Wrapping: Keep function name and arguments together


davidaurelio:
Function call wrapping is not optimal for readability: Function names tend to get pushed to the right, whereas arguments end up on the left, creating a wide gap that the eyes have to cross in order to read the call.

It would be great if we could make this better, e.g. as suggested below. I will be happy to make the necessary code changes if there is consensus on this suggestion.

**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEwCOAzAdwDc8BrAgKwBMcBzALwBpMAjMGNATwBsKcBXMA2YAPfhCoiAFgFtaBAgAcCAX0wBeTAB0omTDhZSYYfvREE4VegApOAJwIBLGQVIU4cR-2ZQKxCPSOtFSkeI7MAIbcREQUtKQyUDJSVACUANwgjCAQijCO6MigEXZ2EAQACiUIaMggEUQQjlRZICx2EWCkcDAAyoqdjlC0yDB2-HDZRjLcAOpSjvBoA2BwvTWLjkSLnHVgaLXZQ2hwdjAVHbQyEch4USfZFGgiAEIdXT29ETJwADJDcFu90mICeIl6Q1o3DgAEVxPAgdwHiABnYTnY6iwIiweNBWoo7EMYLNmjApMgABwABmyBIgJ1mHUUdQJcHRREB2TscAEjm5FwiVxuSDuSJBJxkjlG43FkOhcIgCJFwOyMGxJKoZOQACZVR1HNxIQBhCAya51KDQTkgfgnAAq2NqyrFKhUQA)
```sh
--parser babylon
```

**Input:**
```jsx
const {qfwvfkwjdqgz, bctsyljqucgz, xuodxhmgwwpw} = 
  qbhtcuzxwedz(yrwimwkjeeiu, njwvozigdkfi, alvvjgkmnmhd);
```

**Output:**
```jsx
const { qfwvfkwjdqgz, bctsyljqucgz, xuodxhmgwwpw } = qbhtcuzxwedz(
  yrwimwkjeeiu,
  njwvozigdkfi,
  alvvjgkmnmhd
);

```

**Expected behavior:**
```jsx
const { qfwvfkwjdqgz, bctsyljqucgz, xuodxhmgwwpw } = 
  qbhtcuzxwedz(
    yrwimwkjeeiu,
    njwvozigdkfi,
[...]

PR #10298, opened 2021-02-11
DRAFT RFC: simpler solution for consistent assignment formatting - DRAFT WIP

brodybits:
## Description

This is a much simpler solution than PR #10222 to format assignments more consistently for issue #2482. It should also solve the call wrapping issue in #5610 and break assignments with method chains as proposed in #10297.

In short, this is a proposal to *always* put the RHS into a group for the sake of simplicity and consistency.

Unfortunately I think there may too much collateral damage to take this proposal seriously, but I think the point is to look for a way to simplify the assignment formatting, with fewer exceptions.

## Checklist


brodybits:
My apologies, this does not seem to format the cases right.

PR #10313, opened 2021-02-14
chore: Check for updates every Sunday

lipis:
## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,15 +3,20 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
-    open-pull-requests-limit: 20
+      interval: weekly
+      day: sunday
+      time: "01:00"
 
   - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


thorn0:
It was @fisker who set it to "daily". I guess he likes its being this way, so why not let it be? On the other hand, I see how the stream of bot notifications can deter contributors.

lipis:
> I see how the stream of bot notifications can deter contributors.

That's the main reason.. we can give it a try and if @fisker agrees we can merge :)

fisker:
Personaly I prefer deps update ASAP so the change will be small.

If you feel it's annoying, I'm fine with this change.

One question, does dependabot work on weekend? As far as I know , it doesn't work on weekend if we set to "daily".

lipis:
I think it's also going to work better for you, as this way it might skip a few versions if the release cycle is faster than a week for some libraries.. Yes it works on Sundays if you say so.. not sure when set to daily though. My vote is to give it a shot!

fisker (approved):

PR #10324, opened 2021-02-15
Refine CONTRIBUTING.md

thorn0:

**✨[Preview](https://github.com/thorn0/prettier/blob/contributing-md/CONTRIBUTING.md)✨**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ To debug Prettier locally, you can either debug it in Node (recommended) or the 
 
 ## Other
 
-Run `yarn fix` to lint and format all files.
+The project uses ESLint for linting and Prettier for formatting. If your editor isn't set up to work with them, you can lint and format all files from the command line using `yarn fix`.
 
 After opening a PR, describe your changes in a file in the `changelog_unreleased` directory following the template [`changelog_unreleased/TEMPLATE.md`](changelog_unreleased/TEMPLATE.md) and commit this file to your PR.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,8 @@ Take a look at [`commands.md`](commands.md) and, if you know Haskell, check out 
 
 If you want to know more about Prettier's GitHub labels, see the [Issue Labels](https://github.com/prettier/prettier/wiki/Issue-Labels) page on the Wiki.
 
+# Advanced topics
+
 ## Performance
 
 If you're contributing a performance improvement, the following Prettier CLI options can help:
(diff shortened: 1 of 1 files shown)


fisker:
Can you also simply explain that `run_spec` function accept object as first argument.

fisker:
One more thing, please metion that `babel-ts` parser will be tested when `typescript` appears, and when `babel` parser is tested in `tests/js/` `espree` and `meriyah` will be tested too.

CONTRIBUTING.md
```diff
@@ -4,27 +4,74 @@ To get up and running, install the dependencies and run the tests:
...
+
+```js
+run_spec(__dirname, ["babel"], { trailingComma: "es5" });
+```
+
+Signature:
+
+```ts
+function run_spec(
+  fixtures: { dirname: string; snippets?: string[] } | string,
```
fisker, line 32: `snippets` support object with properties `name`(test name), `filename`, `code`, `output`(direct assert output string)

CONTRIBUTING.md
```diff
@@ -4,27 +4,63 @@ To get up and running, install the dependencies and run the tests:
...
+## Debugging
+
+To debug Prettier locally, you can either debug it in Node (recommended) or the browser.
+
+- The easiest way to debug it in Node is to create a local test file with some example code you want formatted and either run it in an editor like VS Code or run it directly via `./bin/prettier.js <your_test_file>`.
+- The easiest way to debug it in the browser is to build Prettier's website locally (see [`website/README.md`](website/README.md)).
+
+## Other
+
+Run `yarn fix` to lint and format all files.
```
fisker, line 65: I think  better suggest use `yarn lint`, if there is bug  in those eslint rules,        they may lost their work.

thorn0: If you're not sure about some rules, let's better disable them. Many people (me included) use autofix on save. I don't want to lose work too.

fisker: I think enabled rules don't have known issues, in editor if you found something not right you can undo, but if editor already closed and haven't committed, run `yarn fix` you can't find your original code back.

CONTRIBUTING.md
```diff
@@ -4,30 +4,76 @@ To get up and running, install the dependencies and run the tests:
...
+- **`parsers`**: A list of parser names. The tests verify that the parsers in this list produce the same output. If the list includes `typescript`, then `babel-ts` is included implicitly. If the list includes `babel`, and the current directory is inside `tests/js`, then `espree` and `meriyah` are included implicitly.
+- **`options`**: In addition to Prettier's formatting option, can contain the `errors` property to specify that it's expected that the formatting shouldn't be successful and an error should be thrown for all (`errors: true`) or some combinations of input entries and parsers.
+
+The implementation of `run_spec` can be found in `tests_config/run_spec.js`.
```
fisker, line 54: Link to `tests_config/run_spec.js` file.

CONTRIBUTING.md
thorn0: I just might not notice the editor did that.

fisker (approved):
Good job.

PR #10338, opened 2021-02-18
Style: Disallow `==`

fisker:
## Checklist

### DIFF
--- a/src/main/options.js
+++ b/src/main/options.js
@@ -84,7 +84,7 @@ function normalize(options, opts = {}) {
   const mixedDefaults = { ...defaults, ...pluginDefaults };
 
   for (const [k, value] of Object.entries(mixedDefaults)) {
-    if (rawOptions[k] == null) {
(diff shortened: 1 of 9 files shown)


src/main/options.js
```diff
@@ -84,7 +84,7 @@ function normalize(options, opts = {}) {
   const mixedDefaults = { ...defaults, ...pluginDefaults };
 
   for (const [k, value] of Object.entries(mixedDefaults)) {
-    if (rawOptions[k] == null) {
+    if (rawOptions[k] === null || typeof rawOptions[k] === "undefined") {
```
thorn0, line 87: Normal `undefined` is used everywhere else throughout the code base.

thorn0 (approved):
Not sure about this one either, but it's not very obtrusive. If you feel like this achieves something, so be it.

fisker:
The `eqeqeq` rule options `{"null": "ignore"}` has been deprecated, then I thought better enfore use `===`.

PR #10339, opened 2021-02-18
Enable `prefer-named-capture-group`

fisker:
## Checklist

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -54,6 +54,7 @@ rules:
         array: false
         object: false
     - enforceForRenamedProperties: false
+  prefer-named-capture-group: error
   prefer-object-spread: error
   prefer-rest-params: error
   prefer-spread: error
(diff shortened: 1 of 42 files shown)


thorn0:
Sorry, I'm beating my drum again, but this rule doesn't look like an improvement to me.
* It makes short simple regexps longer and long expressions even longer.
* Is there any data on performance of named groups vs usual groups?
I think named capturing groups should be used sparingly, only where they really help readability. E.g., using them in big regexps assembled from parts (like `frontMatterRegex`) is definitely a good thing.

fisker:
> Is there any data on performance of named groups vs usual groups?

I don't think named groups have performance improvements, it's only about readability. And I think some part indeed improve the readability, for example https://github.com/prettier/prettier/pull/10339/files#diff-5cb80a623d9d5444ca337d3df7b41b10cff0c6426f6ca25005e2fead0994d6b2R35.

`non-capturing group` should improve performance, and there are quite a few places don't really need capture in this PR.

PR #10349, opened 2021-02-19
Expose different doc objects

fisker:
## Description

Expose different doc objects for public APIs, so we can get ride of `concat` for internal use.

## Checklist


thorn0:
Looks like an unneeded indirection to me. We'll get rid of `concat` anyway, just not today. Is it causing any trouble now?

fisker:
The doc utils should not exported directly, like https://github.com/prettier/prettier/blob/main/src/common/util-shared.js vs https://github.com/prettier/prettier/blob/main/src/common/util.js

thorn0:
What's public and what's not is a matter of documentation. Let's keep things simple.

fisker:
> What's public and what's not is a matter of documentation. 

I don't think documentation is the only thing, private function can do breaking change when needed, while public things should take compatibility into consideration.

fisker:
Outdated.

PR #10354, opened 2021-02-20
Add `flat-ast-path-call` rule

fisker:
## Description

Enfore `path.call(print, "a", "b")` over `path.call((childPath) => childPath.call(print, "b"), "a")`.

This rule only handle simple cases, we can add more complicated cases in feature.

It only caught one place in codebase 😄 

---

I remember there was unneeded `.map()` in `.call()` somewhere, but couldn't find it.

## Checklist

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -117,6 +117,7 @@ overrides:
       prettier-internal-rules/better-parent-property-check-in-needs-parens: error
   - files: src/**/*.js
     rules:
+      prettier-internal-rules/flat-ast-path-call: error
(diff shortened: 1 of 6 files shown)


thorn0 (approved):
Okay. Let's add this one too. Did you consider adding a rule for `getLast`, BTW?

fisker:
> Did you consider adding a rule for getLast, BTW?

I can do it, it will be a simpler version of [this rule](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/983) I'm going to implement recently.

PR #10391, opened 2021-02-22
JS: Split `locStart` into two functions

fisker:
## Description

Remove the `opt` parameter, I'm planning use `locStartWithoutDecorator` for `Node`s that can't have a decorator, eg: `Comment`

## Checklist


thorn0:
> I'm planning use locStartWithoutDecorator for Nodes that can't have a decorator, eg: Comment

Please show some benchmark numbers confirming that such a microoptimization would really make sense.

fisker:
This part of logic already changed.

PR #10394, opened 2021-02-22
Add `css` version `printString`

fisker:
## Description

One step to get ride of `parser` check in `printString` util.

## Checklist


thorn0:
> to get ride of parser check in printString util

Why is that a good thing?

fisker:
It's public util, I don't think it should related to our parsers.

It should not even receive `options`, IMHO it can have `preferredQuote` and `unescapeUnnecessaryEscapes` option (also `unescapeUnnecessaryEscapes` should default to `false`, since it's only JS thing).

thorn0:
> It's public util

Is it?

fisker:
> Is it?

Sorry, I thought it is... I'll rethink about this.

PR #10402, opened 2021-02-24
Consistent `print*`  function parameters order

fisker:
## Description

Consistent `print*` function parameters order, match [this call](https://github.com/prettier/prettier/blob/5682d2f4405d843ed77d8c13b8048d7b7dfd8b49/src/main/ast-to-doc.js#L50), WDYT?

If it's okay, I'll start fix them.

Log on CI, https://github.com/prettier/prettier/pull/10402/checks?check_run_id=1968654916#step:7:5

## Checklist


fisker:
@sosukesuzuki

I'm going to bring this up again, since we already have inconsistent forcing us uses a simple wrapper to reuse logic https://github.com/prettier/prettier/blob/9dac37c60804c38fe4e6e8888768b39a097d865b/src/language-js/print/enum.js#L4, and I plan to add a new function reuse logic from `printObject` to fix https://github.com/prettier/prettier/issues/17071. 

Accept?

PR #10407, opened 2021-02-26
Simplify JS parser handling  Hashbang

fisker:
## Description

Use the way how ESLint handle this.

## Checklist


src/language-js/parse-postprocess.js
```diff
@@ -3,21 +3,12 @@
 const {
   getLast,
   getNextNonSpaceNonCommentCharacter,
-  getShebang,
```
thorn0, line 6: Didn't you forget to delete this function in `common/util`? Or is it still used?

fisker: Needed, https://github.com/prettier/prettier/blob/fc396277fc64ca6d056960fb43e9b1ad5296f597/src/language-js/pragma.js#L8 and   https://github.com/prettier/prettier/blob/fc396277fc64ca6d056960fb43e9b1ad5296f597/src/language-js/parser-babel.js#L61

fisker:
Need approve here.

thorn0 (approved):

PR #10415, opened 2021-02-27
Restructure test files

fisker:
### Motivation

I want add a new directory for #10410, and I think better move everything in to `tests/`.
We may also want add unit tests later, so we can put in `tests/unit`

### Changes

Move test files.

- `tests/` -> `tests/format`
- `tests_config` -> `tests/config`
- `tests_integration` -> `tests/integration`

## Checklist


thorn0 (approved):
I'm okay with this change. But it's better to merge other PRs first (yours and Sosuke's) to avoid conflicts.

brodybits:
I have to say that moving this many files around makes it harder for forks to keep up. And forks are needed due to the Prettier option philosophy.

fisker:
@brodybits Thanks for your feedback, but I think it's nessary, we already have 3 directories for tests, adding more directories for test to the project root is bad idea IMHO.  Also this change is not that huge, your fork can easily move them.

sosukesuzuki (commented):
Sorry for delay, I'm good with this change.

PR #10437, opened 2021-02-28
Simplify babel parser

fisker:
## Description

Use `optionsCombinations` instead of `pluginCombinations`, makes logic easier to understand.

## Checklist

### DIFF
--- a/src/language-js/parser-babel.js
+++ b/src/language-js/parser-babel.js
@@ -49,6 +49,10 @@ const pipelineOperatorPlugins = [
   ["pipelineOperator", { proposal: "minimal" }],
   ["pipelineOperator", { proposal: "fsharp" }],
 ];
+const appendPlugins = (plugins) => ({
+  ...parseOptions,
+  plugins: [...parseOptions.plugins, ...plugins],
+});
 
 // Similar to babel
 // https://github.com/babel/babel/pull/7934/files#diff-a739835084910b0ee3ea649df5a4d223R67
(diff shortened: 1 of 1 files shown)


thorn0:
Kind of not obvious that this is a simplification. It simplifies some lines of code and makes other ones less simple. The net effect seems to be zero.

fisker:
Yes this not simplified too much, but easier to pass different parse options, main reason behind this is https://github.com/prettier/prettier/issues/10406, I remember some where mentioned `eslint` plugin should be the first one in `plugins`(I haven't test, and not sure), then I was thinking previous version should be hard to add some plugin prefer the plugins list, but in current implement, it's very easy.

```js
const parseEslint = createParse("parse", {...parseOptions, plugins: ['eslint', ....parseOptions.plugins]});
```

Same thing, if we still keep `json` parser in this file

```js
const parseJson = createParse('parseExpression', {ranges: true, tokens: true});
```

The old version can't do this.

brodybits:
> previous version

By "previous version" and "old version" do you mean "prettier-stable" in `main` or do you mean something else?

> but in current implement, it's very easy.
> 
> ```js
> const parseEslint = createParse("parse", {...parseOptions, plugins: ['eslint', ....parseOptions.plugins]});
> ```

Wow that would be awesome. Can you maybe show this in a PR?

fisker:
> By "previous version" and "old version" do you mean "prettier-stable" in main or do you mean something else?

No, old versiob of `createParse` function, the deleted code in this PR.

> Can you maybe show this in a PR?

`appendPlugins` function is doing this.

fisker:
@brodybits This is internal change, no effect to the public api,  some part  in this function is just old, I've done several changes to make it clear, it was hard to understand what happened in the parsing process before.

thorn0 (approved):

PR #10455, opened 2021-03-01
Format svg tags in the same way html tags are formatted

Artur-:
## Description

lit-html uses `svg` template tag for rendering svgs or embedding svgs inside html. This change enables prettier to format the contents of the svg tag.

## Checklist

- [X] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).

### DIFF
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -20,7 +20,7 @@ function getLanguage(path) {
     return "graphql";
   }
 
-  if (isHtml(path)) {
+  if (isHtml(path) || isSvg(path)) {
     return "html";
   }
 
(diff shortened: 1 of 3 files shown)


thorn0:
Unfortunately, changes like this are currently blocked by #5588. Until a solution for that is found and implemented, we can't move forward with adding more embedded syntaxes since they are inherently an unsafe transformation. Thus the more Prettier adds them the less reliable it gets. There were already cases where such changes had to be reverted. We can't know who else and how uses the `svg` template tag.

Similar blocked PRs: #8647, #5993

PR #10457, opened 2021-03-02
Remove hacks for single-element unions/intersections

thorn0:
## Description

Thanks to https://github.com/babel/babel/pull/12758 now we can remove tricky workarounds for single-element unions and intersections.

## Checklist


src/language-js/comments.js
```diff
@@ -980,7 +964,6 @@ module.exports = {
   handleEndOfLineComment,
   handleRemainingComment,
   isTypeCastComment,
-  getGapRegex,
```
fisker, line 983: I believe this only used in js, should we  remove it from comment attach?

thorn0: I'd rather keep it. It's a reasonable extension point. The default regex (`/^[\s(]*$/`) is JS-specific. It might work well for languages that use parens in a way similar to how JS uses them, but it's also possible that other languages might need to override this.

fisker (approved):

Issue #5004, opened 2018-08-22
Prettier crashes on formating large files


trzecieu:
Hi!
I'm working with large ASM.js files generated by Emscripten, when I debug those I use tools like Prettier to make code at lease a little bit more readable. 

In case of prettier I'm getting: 
```

<--- Last few GCs --->

[17699:0x2702f10]    20187 ms: Mark-sweep 1397.2 (1424.1) -> 1396.4 (1423.6) MB, 596.2 / 0.0 ms  (average mu = 0.145, current mu = 0.011) allocation failure scavenge might not succeed
[17699:0x2702f10]    20833 ms: Mark-sweep 1397.4 (1424.1) -> 1396.7 (1424.1) MB, 643.6 / 0.0 ms  (average mu = 0.076, current mu = 0.004) allocation failure scavenge might not succeed


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x2957d235c01d]
    1: StubFrame [pc: 0x2957d2313490]
Security context: 0x16bb6979e549 <JSObject>
    2: willBreak(aka willBreak) [0xccf8baf92a9] [/home/[moderated]/node_modules/prettier/bin-prettier.js:~9173] [pc=0x2957d2724356](this=0x37817e8826f1 <undefined>,doc=0x0388ed06f201 <Object map = 0x2804a52f5a01>)
    3: arguments adaptor frame: 3->1
    4: printArgumentsList(aka printArgumentsList) [0xccf8bafbd69...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x8c02c0 node::Abort() [node]
 2: 0x8c030c  [node]
 3: 0xad15de v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xad1814 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xebe752  [node]
 6: 0xebe858 v8::internal::Heap::CheckIneffectiveMarkCompact(unsigned long, double) [node]
 7: 0xeca982 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [node]
[...]

PR #10492, opened 2021-03-09
JSON: Forbid parenthesized expressions

fisker:
## Checklist


thorn0:
We should research how much memory tokens take before adding solutions that rely on them. I was looking into https://github.com/prettier/prettier/issues/5004#issuecomment-459876800 and saw that the process actually runs out of memory during parsing. I used Meriyah. After disabling tokens and locs (lines/columns), it started to go further than the parsing phase.

PR #10503, opened 2021-03-10
Playground: Use class methods

fisker:
## Checklist

### DIFF
--- a/website/playground/Playground.js
+++ b/website/playground/Playground.js
@@ -58,6 +58,7 @@ class Playground extends React.Component {
       props.availableOptions,
       ENABLED_OPTIONS
     );
+    this.defaultOptions = defaultOptions;
 
     const options = Object.assign(defaultOptions, original.options);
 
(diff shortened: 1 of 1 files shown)


website/playground/Playground.js
```diff
@@ -154,6 +131,7 @@ class Playground extends React.Component {
   render() {
     const { worker, version } = this.props;
     const { content, options } = this.state;
+    const handleOptionValueChange = this.handleOptionValueChange.bind(this);
```
thorn0, line 134: This shouldn't be done in `render`.

website/playground/Playground.js
```diff
@@ -368,6 +348,40 @@ class Playground extends React.Component {
...
     );
   }
+
+  setContent(content) {
+    this.setState({ content });
+  }
+
+  clearContent() {
+    this.setContent("");
+  }
```
thorn0, line 358: It'd be good to enable class fields in the Babel config and replace these with arrow functions assigned to fields, but I remember I had trouble with configuring that.

website/playground/Playground.js
thorn0: Both aren't recommended. This creates a new function value on every render, which triggers re-rendering of components this function passed to.

fisker: I know, but arrow functions used in other part of this `render`, I don't think it cost much.

PR #10504, opened 2021-03-10
Rename `n` to `node`

fisker:
## Checklist

### DIFF
--- a/src/document/doc-builders.js
+++ b/src/document/doc-builders.js
@@ -73,6 +73,7 @@ function indent(contents) {
  * @param {Doc} contents
  * @returns Doc
  */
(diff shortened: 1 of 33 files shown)


src/document/doc-builders.js
```diff
@@ -73,6 +73,7 @@ function indent(contents) {
  * @param {Doc} contents
  * @returns Doc
  */
+// eslint-disable-next-line prettier-internal-rules/no-identifier-n
 function align(n, contents) {
```
thorn0, line 77: Let's rename `n` to `widthOrString` here and in `doc-printer`. Just the variables, not the property.

thorn0 (approved):

sosukesuzuki (approved):

PR #10511, opened 2021-03-12
JS: Simplify print

fisker:
## Description

Added a simpler version `print`, ideally this change should apply to [this function](https://github.com/prettier/prettier/blob/1ce6b8dc3d3a21155bd95cb56824fe8e720837b7/src/main/ast-to-doc.js#L42), but I think this may break some edge case in plugins, so I added it in JS first.

The problem in that `printGenerically`, we meaninglessly passing the same `AstPath` over and over, and it's also missleading, when I first saw code like `path.map(childPath => print(childPath), 'key')`, I thought `path.map(childPath => childPath, 'key').map(print)` should work too, but it's not.

It's not done yet, I want confirm this is the right direction first.

## Checklist

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -142,4 +142,4 @@ overrides:
         - "src/language-js/parser-meriyah.js"
         - "src/language-js/pragma.js"
         - "src/language-js/parser/json.js"
-      prettier-internal-rules/no-ast-path-call: error
+      prettier-internal-rules/simplified-print: error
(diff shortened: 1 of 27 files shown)


fisker:
@sosukesuzuki @thorn0 Need your opinion.

thorn0:
The direction looks interesting. The implementation that creates a new `print` function on every step looks doubtful. But I understand it's far from final. Would be interesting to see it done in `printGenerically`,

fisker:
Do you think it's possible someone use `printGenerically` to print another `AstPath`? 

If not we can do it in `printGenerically` 😄

thorn0:
> Do you think it's possible someone use `printGenerically` to print another `AstPath`?

Its constructor isn't public API. Don't tell me they could do `new path.constructor(...)` :)

fisker:
I'll try to change `printGenerically` then.

fisker:
Close this for now, I'll reopen when it's ready.

PR #10528, opened 2021-03-14
Improve detection of AMD define

### LINKED ISSUE #10422: Formatting issue when using a function called 'define' (no AMD/RequireJS)
**Prettier 2.2.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAmHwBM4AzVKOfAX3zJwnXwB0QBnJuAWgBtUARjgCGOAJ5sA3CygzIUdgU7o4ANVGphA3tQC8JcpTgAKGflYcu+RTkoBzfP3gjebADRn8wqBBgALOBx1Wy0dDyhzMTgYAEEff0DgzW04GQBKSRA3EAhMGFRodmRQUUYAdwAFUQQilGFeMuExIuyhYTAAa2iAZUx2+2QYHABXOGy4dAE4YlJiABlvO2HhOzgAMVx0YRh8qDtkEGFhmAgskD8YdF4AdT9UeHY+sDhumvvUADd7sQOwdha0ApAjAKiI7FtkGR6uwxiAAFbsAAeACERJ0esIVHMjJDobCEYjuvYdABFYa+OC43gw7J9HAwnAHARaOC8M6YWywa6oYj+ZAADgADLTGDDriJMAcOXAGR9KdkAI7k+Cg3K1Q7sbhUabTM44OBK1D60ErCFIKHU2Ew9CoQYjK3EuBkilUmkgGBabm8vzIABM2SGwlQ-D2AGEmGaQDKAKxnYYwgAqWlqFrdH1GAEkoKRYN0wLY8nFiN0YGIdK64DQaEA)

```sh
--parser babel
```
[EDIT: Same issue with TypeScript]

**Input:**

```jsx
import { define } from "some-library";

const someVariable = define(
  "some string literal",
  anotherVariable,
  yetAnotherVariable
);
```

**Output:**

```jsx
import { define } from "some-library";

const someVariable = define("some string literal", anotherVariable, yetAnotherVariable);

```

**Expected behavior:**
Should format like in the input snippet above (while width in "Output" is more than 80 chars, which seems to be a bug).

**Remarks**
I know that Prettier treats functions that are called `define` in a special way (because of the AMD/RequireJS `define` function) and I also know that I could use `// prettier-ignore` (which would be quite sad if this was the only solution).

thorn0:
## Description

Closes #10422

## Checklist

### DIFF
--- a/src/language-js/print/call-expression.js
+++ b/src/language-js/print/call-expression.js
@@ -113,7 +113,7 @@ function isCommonsJsOrAmdCall(node, parentNode) {
   if (node.callee.name === "define") {
     const args = getCallArguments(node);
     return (
-      (!parentNode || parentNode.type === "ExpressionStatement") &&
+      parentNode.type === "ExpressionStatement" &&
       (args.length === 1 ||
         (args.length === 2 && args[0].type === "ArrayExpression") ||
         (args.length === 3 &&
--- a/src/language-js/print/call-expression.js
+++ b/src/language-js/print/call-expression.js
@@ -19,27 +19,25 @@ const { printOptionalToken, printFunctionTypeParameters } = require("./misc");
 
 function printCallExpression(path, options, print) {
   const node = path.getValue();
+  const parentNode = path.getParentNode();
   const isNew = node.type === "NewExpression";
   const isDynamicImport = node.type === "ImportExpression";
 
   const optional = printOptionalToken(path);
   const args = getCallArguments(node);
   if (
-    // Dangling comments not handled, all these special cases should has argument #9668
+    // Dangling comments are not handled, all these special cases should have arguments #9668
     args.length > 0 &&
     // We want to keep CommonJS- and AMD-style require calls, and AMD-style
     // define calls, as a unit.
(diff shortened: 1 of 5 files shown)


src/language-js/print/call-expression.js
```diff
@@ -101,12 +102,27 @@ function printCallExpression(path, options, print) {
...
+  }
+
+  if (node.callee.name === "require") {
+    return true;
+  }
+
+  if (node.callee.name === "define") {
+    const args = getCallArguments(node);
+    return (
+      (!parentNode || parentNode.type === "ExpressionStatement") &&
```
fisker, line 116: I  don't think it's possible that CallExpression don't have a parent.

ChuVienTran (approved):

src/language-js/print/call-expression.js
thorn0: I had big doubts about that too, but wasn't sure. Okay, let's remove this check.

fisker (approved):

PR #10538, opened 2021-03-14
Remove stale plugins from home page and docs

thorn0:
## Description

:(

cc @sirlantis @benjie

## Checklist


sirlantis:
Thanks. Absolutely right thing to do. 👌

benjie:
I’ve actually just started working on prettier-plugin-pg again thanks to a new postgres parser surfacing. No guarantees it’ll result in anything though, so remove it if you want; we can always add it back later 👍

PR #10543, opened 2021-03-15
Fix edge case with trailing comments in arrow chains

thorn0:
## Description

Fixing this:

**Prettier pr-10511**
[Playground link](https://deploy-preview-10511--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAPABAXnQCgIYCUWAfLjgEZGanAA6U66YOBA3PQL7oD0AVOhAgAHAM7pe3AiAA0IYTACW0EclB4ATuogB3AAoaEKlHgA22vAE8Vs8urxgA1nBgBlIfYVQA5shjqArnCycAC25HAAJhGRADJ43v54XnAAYhDqIXgwit7IIHj+MBAyIAAWMCEmAOqlCvAi7mBwLoZ1CgBudRZ5YCLWIJ4icOowunZemcgAZqZDsgBWIqgAQnaOzi54IXAxnnDTs0Egi6gunl4mcACK-hDwByZzIO7qQ+p55HjhJiVC6p4wKoKCIwUrIAAcAAZZH8IEMqnYhHk-nA3u19rIAI63eBjYRGfIiAC0UDgkUiJXUcGxCipYySkyQM0eRyGIQUvgCrPOlxud32TMOshgXyBILBSAATMK7AoTOcAMIQEKMkCogCsJX8QwAKl8jMynu1AgBJKDRWAuMD-IQwACC5pcMAslweQw4HCAA)

```sh
--parser babel
```

**Input:**

```jsx
x = (a) => ((b) => {
  c();
} /* oops */)
```

**Output:**

```jsx
x = (a) => (b) /* oops */ => {
  c();
};

```

## Checklist

### DIFF
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -540,23 +540,18 @@ function printDanglingComments(path, options, sameIndent, filter) {
   return indent([hardline, join(hardline, parts)]);
 }
 
-function prependCursorPlaceholder(path, options, printed) {
-  if (path.getNode() === options.cursorNode && path.getValue()) {
-    return [cursor, printed, cursor];
-  }
-  return printed;
-}
-
-function printComments(path, printed, options, needsSemi) {
+function printCommentsSeparately(path, options) {
   const value = path.getValue();
-  const comments = value && value.comments;
+  const hasComments = isNonEmptyArray(value && value.comments);
+  const isCursorNode = Boolean(path.getNode() === options.cursorNode && value);
 
-  if (!isNonEmptyArray(comments)) {
-    return prependCursorPlaceholder(path, options, printed);
+  if (!hasComments) {
+    const maybeCursor = isCursorNode ? cursor : "";
+    return { leading: maybeCursor, trailing: maybeCursor };
   }
 
-  const leadingParts = [];
-  const trailingParts = [needsSemi ? ";" : "", printed];
+  const leadingDocs = [];
+  const trailingDocs = [];
 
   path.each((commentPath) => {
     const comment = commentPath.getValue();
@@ -568,7 +563,7 @@ function printComments(path, printed, options, needsSemi) {
       if (!contents) {
         return;
       }
-      leadingParts.push(contents);
+      leadingDocs.push(contents);
 
       const text = options.originalText;
       const index = skipNewline(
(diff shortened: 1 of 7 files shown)


src/main/comments.js
```diff
@@ -583,10 +578,20 @@ function printComments(path, printed, options, needsSemi) {
...
-  ]);
+  if (isCursorNode) {
+    leadingParts.unshift(cursor);
+    trailingParts.push(cursor);
+  }
+
+  return { leading: leadingParts, trailing: trailingParts };
+}
+
+function printComments(path, doc, options, needsSemi) {
```
fisker, line 589: As I can see `needsSemi` is not used at all, and I don't think plugin use it, lets remove it.

fisker: Oh, just one place `printStatementSequence` use it, but it can print semi itself.

thorn0: Done!

src/main/comments.js
```diff
@@ -540,23 +540,18 @@ function printDanglingComments(path, options, sameIndent, filter) {
...
-  }
-  return printed;
-}
-
-function printComments(path, printed, options, needsSemi) {
+function printCommentsSeparately(path, options) {
   const value = path.getValue();
-  const comments = value && value.comments;
+  const hasComments = isNonEmptyArray(value && value.comments);
+  const isCursorNode = Boolean(path.getNode() === options.cursorNode && value);
```
fisker, line 546: Why the old `prependCursorPlaceholder` use both `path.getNode` and `path.getValue`? Can we use `value` here?

src/main/comments.js
```diff
@@ -583,10 +578,20 @@ function printComments(path, printed, options, needsSemi) {
...
+
+  return { leading: leadingParts, trailing: trailingParts };
+}
+
+function printComments(path, doc, options) {
+  const { leading, trailing } = printCommentsSeparately(path, options);
+  if (!leading && !trailing) {
+    return doc;
+  }
+  return [leading || "", doc, trailing || ""];
```
fisker, line 594: `leading` and `trailing` can only be `cursor`, `Array`, and `""`? So we should use `leading` directly instead of `leading || ""`.

src/main/comments.js
fisker: Can't it done here https://github.com/prettier/prettier/blob/863eb079dc0aa6478b24842f69bef2b99408c948/src/language-js/print/statement.js#L49 ?

thorn0: In that case we will need a way to say `print` that it shouldn't print comments. How would you do that?

src/main/comments.js
thorn0: No idea. I think we can.

src/main/comments.js
fisker: You are right.

fisker (approved):

src/main/comments.js
fisker: After second thought, we can delete comments then print without comments :smile: But it still won't work, because statements may need parens, semicolon will be inside.

PR #4848, opened 2018-07-16
fix(perf): convert traverseDoc from recursion to stack (#4776)

sompylasar:
In addition to a tiny performance improvement outlined below,
the CPU profile of traverseDoc is now more readable.

Also anonymous arrow functions changed to named regular functions
so that they are properly displayed in the CPU profile,
and moved to outer scope where there's no closure
so that they aren't re-created (this change's performance is dependent
on JS engine implementation and optimization details).

Before (profile):
```
7129.9 ms       5.43 %    13349.9 ms     10.18 %    traverseDocRec
7067.4 ms       5.39 %    11285.5 ms      8.60 %      traverseDocRec
  31.5 ms       0.02 %     1031.9 ms      0.79 %      traverseDoc
  23.6 ms       0.02 %    12313.4 ms      9.39 %      traverseDoc
   2.6 ms       0.00 %       0.3 ms       0.00 %      (anonymous)
   1.7 ms       0.00 %       1.7 ms       0.00 %      call
   1.6 ms       0.00 %       1.6 ms       0.00 %      call
   0.5 ms       0.00 %       0.5 ms       0.00 %      conditionalGroup
   0.4 ms       0.00 %       0.4 ms       0.00 %      printDocToString$1
   0.1 ms       0.00 %       0.1 ms       0.00 %      printGenerically
   0.1 ms       0.00 %       0.1 ms       0.00 %      t
   0.1 ms       0.00 %       0.1 ms       0.00 %      ifBreak
   0.1 ms       0.00 %       0.1 ms       0.00 %      (anonymous)
     0 ms          0 %       0.1 ms       0.00 %      forEach

```

After (profile):
```
6937.9 ms       5.37 %    12872.5 ms      9.97 %    traverseDoc
5944.0 ms       4.60 %    11047.3 ms      8.55 %      propagateBreaks
735.7 ms        0.57 %     1358.3 ms      1.05 %      findInDoc
257.9 ms        0.20 %      466.7 ms      0.36 %      findInDoc
0.1 ms          0.00 %        0.1 ms      0.00 %      has
0.1 ms          0.00 %        0.1 ms      0.00 %      printArgumentsList
```

Before (performance):
```
cat ../LspLanguageService.js | NODE_ENV=production node --inspect-brk ./dist/bin-prettier.js --stdin-filepath LspLanguageService.js --loglevel debug --debug-repeat 1000 > /dev/null
[...]

PR #10557, opened 2021-03-17
Simplified `printGenerically`

fisker:
## Description

I can't reopen #10511, see previous discuss.

## Checklist


docs/plugins.md
```diff
@@ -176,21 +176,28 @@ function print(
   path: AstPath,
   options: object,
   // Recursively print a child node
-  print: (path: AstPath) => Doc
+  printGenerically: (
```
thorn0, line 179: If we want to change this in the docs, we need a better name than `printGenerically`.

src/main/ast-to-doc.js
```diff
@@ -38,36 +38,70 @@ function printAstToDoc(ast, options, alignmentSize = 0) {
...
+    // TODO: Remove support for passing `path` in next major version
+    if (nameOrNames === path || typeof nameOrNames === "undefined") {
+      nameOrNames = [];
     }
 
-    if (shouldCache) {
-      cache.set(node, doc);
-    }
+    const names = Array.isArray(nameOrNames) ? nameOrNames : [nameOrNames];
+    const doc = path.call(() => {
```
thorn0, line 54: Again, a new function instance is created on every `printGenerically` call. Can it be extracted to an outer scope?

src/language-yaml/printer-yaml.js
```diff
@@ -326,18 +323,15 @@ function printNode(node, parentNode, path, options, print) {
...
+      return ["- ", alignWithSpaces(2, !node.content ? "" : print("content"))];
     case "mappingKey":
-      return !node.content ? "" : path.call(print, "content");
+      return !node.content ? "" : print("content");
     case "mappingValue":
-      return !node.content ? "" : path.call(print, "content");
+      return !node.content ? "" : print("content");
     case "mapping":
-      return join(hardline, path.map(print, "children"));
+      return join(hardline, print("children"));
```
thorn0, line 334: Changes like this look bad to me. `path.map` is more type-safe. Previously we knew for sure that `map` would return an array. Now we can only hope it will do so.

fisker: I was not sure about this change too, but we can't use `path.map(print, "children")` anymore, we need use `path.map(() => print(), "children")`, because `print` don't accept `path`. This is the reason I made this change.

fisker: FYI: The old `path.map(print, "children")` is also not good, it send `index` as `args`, I didn't realize this when I add `index` to `AstPath.map`.

src/main/ast-to-doc.js
fisker: https://github.com/prettier/prettier/pull/10557/commits/fe74439ab36459b44b875c6f877fc1f9943c7013

docs/plugins.md
fisker: How about `printProperty`? Any better idea?

src/language-yaml/printer-yaml.js
thorn0: It wasn't you. It did that from the beginning. Which led to disabled caching, which probably was a good thing. Difficult to tell without performance testing.

fisker: Oh, I remember it incorrectly, I added the array to callback #9901.

Anyway, if print don't handle `array` automatically, I think we'll need write this in 

1.

```js
path.map(() => print(), "children");
```

2.

```js
for (let index = 0; index < node.children.length; index ++) {
	parts.push(print(["children", index]))
}
```

All less easier than `print("children")`

fisker: The `index` is on me... #9499

thorn0: I'd really prefer if `path.map(print, "children");` kept working.

fisker: What if we use `print.array("children")`?

docs/plugins.md
thorn0: It's passed around so much that I believe `print` is the best name, even though this name coincides with `parser.print` (I guess that's the reason why you want to rename it).

src/main/ast-to-doc.js
thorn0: That's bad. I remember getting rid of arrow functions led to a performance boost in `doc-utils`. We should be really careful with them in hot paths.

thorn0: Found that PR: https://github.com/prettier/prettier/pull/4848
Getting rid of arrow functions wasn't the critical part of that optimization, but we need to benchmarks different implementations. It might make sense to assign `args` to an outer-scope variable too.

src/language-yaml/printer-yaml.js
fisker: > I'd really prefer if path.map(print, "children"); kept working.

If I remove [this line](https://github.com/prettier/prettier/pull/10557/files#diff-9e781c3963d9712c66a4747944b15dbc98d3e504f3b126537a746051946f1131R87), it will still work the old way. But it makes `printGenerically` very suspicious.

src/language-js/print/function.js
```diff
@@ -277,12 +277,12 @@ function printArrowFunctionExpression(path, options, print, args) {
       node.body.type !== "ArrowFunctionExpression" ||
       (args && args.expandLastArg)
     ) {
-      body.unshift(path.call((bodyPath) => print(bodyPath, args), "body"));
+      body.unshift(print("body", args));
     } else {
       node = node.body;
-      path.call(rec, "body");
+      path.call(() => rec(), "body");
```
thorn0, line 283: If you could stop doing this, that'd be great. Thanks. 🙏

src/language-yaml/printer-yaml.js
thorn0: We should keep supporting the old way anyway. Why make plugin authors change this? What for?

src/main/ast-to-doc.js
fisker: I doubt the performance impact, of course it will be faster a little bit if we don't create a new function but it should be tiny difference, V8 can do a good job on optimize this.


Anyway, I added https://github.com/prettier/prettier/pull/10557/commits/e24087bb3332f5bfa5eabb08a53800b054b314ec (without creating a new arrow function).

I'm not sure how to do the benchmark test, can you help me to test?

https://github.com/prettier/prettier/pull/10557/commits/e24087bb3332f5bfa5eabb08a53800b054b314ec (Callback without args)
https://github.com/prettier/prettier/pull/10557/commits/fe74439ab36459b44b875c6f877fc1f9943c7013 (Extract callback)
https://github.com/prettier/prettier/pull/10557/commits/1328bb87f8a51194f7de9ba40ad6222f65e3ddbe (Before extract callback)
https://github.com/prettier/prettier/commit/edce20a822dd87f437d9c55b557e9b945586db9e (Main branch)

fisker: Here is the benchmark result

```text
# flow_parser.js from `flow-parser@0.145.0`, 2.9MB

e24087bb3332f5bfa5eabb08a53800b054b314ec (Callback without args)

[debug]   "benchmark": "format x 0.06 ops/sec ±5.73% (5 runs sampled)",
[debug]   "hz": 0.06252269581673485,
[debug]   "ms": 15994.19198

fe74439ab36459b44b875c6f877fc1f9943c7013 (Extract callback)

[debug]   "benchmark": "format x 0.06 ops/sec ±3.23% (5 runs sampled)",
[debug]   "hz": 0.06215052078506828,
[debug]   "ms": 16089.969759999998

1328bb87f8a51194f7de9ba40ad6222f65e3ddbe (Before extract callback)

[debug]   "benchmark": "format x 0.05 ops/sec ±16.00% (5 runs sampled)",
[debug]   "hz": 0.05297934041270051,
[debug]   "ms": 18875.282180000002

edce20a822dd87f437d9c55b557e9b945586db9e (Main branch)

[debug]   "benchmark": "format x 0.05 ops/sec ±31.26% (5 runs sampled)",
[debug]   "hz": 0.04564014458775898,
[debug]   "ms": 21910.535320000003

# doc.js from `prettier@2.2.1`, 115KB


e24087bb3332f5bfa5eabb08a53800b054b314ec (Callback without args)

[debug]   "benchmark": "format x 5.17 ops/sec ±27.92% (20 runs sampled)",
[debug]   "hz": 5.169637385866361,
[debug]   "ms": 193.43716500000002

fe74439ab36459b44b875c6f877fc1f9943c7013 (Extract callback)

[debug]   "benchmark": "format x 5.13 ops/sec ±26.42% (20 runs sampled)",
[debug]   "hz": 5.134979746484509,
[debug]   "ms": 194.742735

1328bb87f8a51194f7de9ba40ad6222f65e3ddbe (Before extract callback)

[debug]   "benchmark": "format x 5.24 ops/sec ±28.54% (20 runs sampled)",
[debug]   "hz": 5.241439713419043,
[debug]   "ms": 190.78727499999997

edce20a822dd87f437d9c55b557e9b945586db9e (Main branch)

[debug]   "benchmark": "format x 6.39 ops/sec ±24.34% (23 runs sampled)",
[debug]   "hz": 6.388479671274391,
[debug]   "ms": 156.5317652173913
```

I really can't tell which one is faster... so I run the e24087bb3332f5bfa5eabb08a53800b054b314ec (Callback without args) on flow_parser.js again, result

```
[debug]   "benchmark": "format x 0.05 ops/sec ±48.52% (5 runs sampled)",
[debug]   "hz": 0.05393845863083009,
[debug]   "ms": 18539.64732
```

thorn0: `flow_parser.js` is a strange file. A big part of it consists of huge strings.

thorn0: `main` performed consistently better in my tests, but only because nodes weren't cached (18cb7fb72).
"Before extract callback" with the cache fix had the same performance as `main`.
Looks like this arrow function optimization doesn't make sense. Except perhaps for reducing the depth of the call stack.

fisker: Thank you very much, in that case I prefer callback with args.

thorn0:
@fisker I added a couple of commits. Do they look good to you?

src/main/ast-to-doc.js
```diff
@@ -40,62 +39,53 @@ function printAstToDoc(ast, options, alignmentSize = 0) {
...
+  }
+
+  propagateBreaks(doc);
+
+  return doc;
+
+  function mainPrint(selector, args) {
+    if (typeof selector === "string") {
+      return path.call(() => mainPrintInternal(args), selector);
+    }
```
fisker: This is not right, it could be number.

```js
path.call(() => {
  print(0);
}, "array")
```

src/main/ast-to-doc.js
```diff
@@ -40,62 +39,53 @@ function printAstToDoc(ast, options, alignmentSize = 0) {
...
+    return mainPrintInternal(args);
+  }
+
+  function mainPrintInternal(args) {
     const value = path.getValue();
 
+    // for path.map(print, ...)
+    if (typeof args === "number") {
+      args = undefined;
+    }
```
fisker, line 72: Let's not change it in this PR.

src/main/ast-to-doc.js
fisker: Or even `symbol`

fisker:
All looks good, thank you for the help ❤️

PR #10563, opened 2021-03-18
Fix cache in `path.map(print, ...)`

fisker:
## Description

@thorn0 WDYT?

## Checklist

### DIFF
--- a/src/main/ast-to-doc.js
+++ b/src/main/ast-to-doc.js
@@ -42,6 +42,11 @@ function printAstToDoc(ast, options, alignmentSize = 0) {
   function printGenerically(path, args) {
     const node = path.getValue();
 
+    // When `printGenerically` is used as callback of `AstPath.map()`, it will get `index` as `args`
+    if (printGenerically.isAstPathMapCallback) {
+      args = undefined;
+    }
+
     const shouldCache = node && typeof node === "object" && args === undefined;
     if (shouldCache && cache.has(node)) {
       return cache.get(node);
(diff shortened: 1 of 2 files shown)


src/main/ast-to-doc.js
```diff
@@ -42,6 +42,11 @@ function printAstToDoc(ast, options, alignmentSize = 0) {
   function printGenerically(path, args) {
     const node = path.getValue();
 
+    // When `printGenerically` is used as callback of `AstPath.map()`, it will get `index` as `args`
+    if (printGenerically.isAstPathMapCallback) {
+      args = undefined;
```
thorn0, line 47: `printGenerically` is recursive. To me, this looks like `args` will be discarded for all nested `printGenerically` calls inside a `path.map` call. I'm surprised tests didn't fail.

fisker: Because I restore the old `isAstPathMapCallback` value back after `.map` call, not set to `false`. So it will only set to `true` when calling `.map`, the only way to break it is to call `path.map(printGenerically)` in `printGenerically` function self.

thorn0: E.g. format this code with and without this change:
```
bar =
  (foo,
  (module.exports =
    (generate, generateRandomOptions) =>
    ({ code, sourceType, reproductionData = undefined }) => {}));
```

thorn0: Or this
```
`${function* mySagas() {
  yield effects.takeEvery(rexpress.actionTypes.REQUEST_START, function* ({
    id
  }) {
    console.log("pong");
  });
}}
```

fisker: You are right. It can be recursive, this is wrong.

fisker:
Sorry, bad idea.

PR #10565, opened 2021-03-18
Add `AstPath#{mapValue,eachValue}()`

fisker:
## Description

We don't need pass `path` around, `mapValue` and `eachValue` pass `value` to callback.

## Checklist


thorn0:
What if we add `node`, `name`, `parent` getters to `AstPath` and destructure the path object in callbacks:
```js
path.map(({ node }) => {
  ...
}, "prop");
```
?

fisker:
That's a good idea.

fisker:
Close in favor of #13448

PR #10566, opened 2021-03-18
Fix print cache in `path.map(print, ...)`

fisker:
## Checklist

### DIFF
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -17,8 +17,9 @@ function getNodeStackIndexHelper(stack, count) {
 }
 
 class AstPath {
-  constructor(value) {
+  constructor(value, print) {
     this.stack = [value];
+    this.print = print;
   }
 
   // The name of the current property is always the penultimate element of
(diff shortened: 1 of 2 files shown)


thorn0:
```js
    // for path.map(print, ...)
    if (typeof args === "number") {
      args = undefined;
    }
```
What if we just add this at the beginning of `printGenerically` and call it a day? Let's not overcomplicate things.

fisker:
Why this overcomplicate things? It is the only exception, checking callback is `print` is the safest way. We can also remove the first argument after we merge #10557

thorn0:
Because of coupling. `AstPath` is nicely abstracted from the rest of the code. Adding exceptions to it is like adding exceptions to `Array.prototype.map`.

>  We can also remove the first argument after we merge #10557

Let's not do that. Callbacks won't always be functions that have `path` in their scope.

fisker:
> Callbacks won't always be functions that have path in their scope.

I mean map call with `print`

thorn0:
One more possible approach: https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L6829

PR #10567, opened 2021-03-18
Use `WeakMap` for print cache

fisker:
## Checklist


thorn0:
What is the point?

fisker:
I think there is performance difference, but I havn't got time to run test yet, that's why this PR is draft.

PR #10604, opened 2021-03-26
docs: Typos and wording in integrating-with-linters.md

eduardopelitti:
This is just some minor grammar stuff I picked up as I was reading the docs. I think these changes makes the docs clearer.

Thanks for an awesome library!

## Description

Wording, typos and grammar changes for the Integrating with Linters section of the docs.

## Checklist

### DIFF
--- a/docs/integrating-with-linters.md
+++ b/docs/integrating-with-linters.md
@@ -17,7 +17,7 @@ Check out the above links for instructions on how to install and set things up.
 
 When searching for both Prettier and your linter on the Internet you’ll probably find more related projects. These are **generally not recommended,** but can be useful in certain circumstances.
 
-First, we have plugins that let you run Prettier as if it was a linter rule:
+First, we have plugins that let you run Prettier as if it were a linter rule:
 
 - [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier)
 - [tslint-plugin-prettier](https://github.com/ikatyang/tslint-plugin-prettier)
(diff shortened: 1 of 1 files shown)


docs/integrating-with-linters.md
```diff
@@ -29,12 +29,12 @@ The downsides of those plugins are:
...
 
-Finally, we have tools that runs `prettier` and then immediately for example `eslint --fix` on files.
+Finally, we have tools that run `prettier` and then immediately apply `eslint --fix` on files.
 
 - [prettier-eslint](https://github.com/prettier/prettier-eslint)
 - [prettier-tslint](https://github.com/azz/prettier-tslint)
 - [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint)
 
-Those are useful if some aspect of Prettier’s output makes Prettier completely unusable to you. Then you can have for example `eslint --fix` fix that up for you. The downside is that these tools are much slower than just running Prettier.
+These are useful if some aspect of Prettier’s output makes Prettier completely unusable to you, then you can have `eslint --fix` fix that up for you. The downside is that these tools are much slower than just running Prettier.
```
thorn0, line 40: Removing "for example" isn't the right thing to do. Not all of these tools run ESLint.
Also, the comma makes the sentence strange, like it's a conditional sentence with two main clauses. "Then ..." is a separate thought here.

eduardopelitti: Yeah you're right, the comma does make it weird. It seems like an odd phrase to me, I'll try to work with it for a bit. And I'll restore the "for example" in the previous line as well.

thorn0: The general idea of those tools is that they run extra code transformations - namely linter autofixers - **after** Prettier.

thorn0:
Closing for now, please feel free to reopen.

PR #10658, opened 2021-04-06
docs(precommit): update husky installation

typicode:
## Description

This updates installation instructions to match current version of husky and mrm lint-staged.

## Checklist

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -15,7 +15,7 @@ _Make sure Prettier is installed and is in your [`devDependencies`](https://docs
 npx mrm lint-staged
 ```
 
-This will install [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) and [lint-staged](https://github.com/okonet/lint-staged), then add a configuration to the project’s `package.json` that will automatically format supported files in a pre-commit hook.
+This will install [husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged), then add a configuration to the project’s `package.json` that will automatically format supported files in a pre-commit hook.
 
 Read more at the [lint-staged](https://github.com/okonet/lint-staged#configuration) repo.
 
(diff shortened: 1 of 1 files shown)


docs/precommit.md
```diff
@@ -15,7 +15,7 @@ _Make sure Prettier is installed and is in your [`devDependencies`](https://docs
 npx mrm lint-staged
 ```
 
-This will install [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) and [lint-staged](https://github.com/okonet/lint-staged), then add a configuration to the project’s `package.json` that will automatically format supported files in a pre-commit hook.
+This will install [husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged), then add a configuration to the project’s `package.json` that will automatically format supported files in a pre-commit hook.
```
brodybits, line 18: This seems to overwrite recent PR #10601.

I would favor documenting lint-staged with Husky as a separate option or sub-option.

alexander-akait:
Honestly, the license changes were disastrous, a lot of repos migrate and rewrite own docs on `simple-git-hooks`, just my onion, so I think now we need two sections for `simple-git-hooks` and `husky`

docs/precommit.md
typicode: `mrm-task-lint-staged` now installs husky 6, so this PR reflects only the current state.
https://github.com/sapegin/mrm/compare/mrm-task-lint-staged@4.0.0...mrm-task-lint-staged@5.0.0

thorn0 (approved):

typicode:
@alexander-akait Sorry if it impacted you, that wasn't the intent. I did the license change in the nicest way that I can think of:
- I continued to maintain husky 4 during the early-access
- Companies were excused for 30 days for unknowingly installing the new version and had a 30 days free trial as well

So no one was forced to upgrade or would be in trouble with the change.

I couldn't find a lot of doc changes but maybe I'm missing something: https://github.com/search?l=&p=3&q=simple-git-hooks+extension%3Amd&ref=advsearch&type=Code

docs/precommit.md
```diff
@@ -29,29 +29,21 @@ Install it along with [husky](https://github.com/typicode/husky):
 <!--npm-->
 
 ```bash
-npm install --save-dev pretty-quick husky
+npx husky-init
+npm install --save-dev pretty-quick
+npx husky set .husky/pre-commit "pretty-quick --staged"
```
fisker, line 34: We don't need install `husky`?

brodybits: I think it is now common practice for people to use `npx` to run CLI tools such as `husky` or `create-react-app` without having to install first. It looks to me like the first step is to use `husky-init` exactly once, then do setup to get Husky working with `pretty-quick`. I must admit I don't personally understand this whole solution so well.

brodybits:
@typicode I did not realize until just now that Husky went through the license changes, and it took me some research and reading to understand that the new Husky version 6 is now back to MIT license (<https://github.com/typicode/husky/issues/857#issuecomment-809342520>).

To clarify a couple more things (as an outside contributor):

My understanding is that Prettier is **not** using Husky or any other git hooks itself, it is only documenting *how* these tools can be used by other projects.

The only doc change I found related to Husky was in PR #10601, which removed Husky in only one place but kept documentation for using Husky together with `pretty-quick`.

docs/precommit.md
fisker: `husky-init` is one time thing, it don't need install, but `husky` runs every time you commit (at least I think it work this way before v5)

fisker: Oh, `husky-init` [installs `husky`](https://github.com/typicode/husky/blob/a66271d21fdf798b291d009c213ed48737dae227/packages/husky-init/src/index.ts#L20) make sense.

PR #10666, opened 2021-04-09
JS: Fix duplicated `prettier-ignore` comments

### LINKED ISSUE #10661: Comment is duplicated when prettier-ignore is used
**Prettier 2.2.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDAdAIwDpQAR4D0heADgE5wwwCWc5AtDQOZQSW4HF4DKEAtnDwQYAC3p5I-QbDziO+POjAgANCAila0AM7JQqcuQgB3AAqGEelKgA2J1AE896zOVRgA1lR6kPNKGZkGHIAVzh1OH5MOAATWLiAGVRA0NRmOAAxdn5UagCgm1CYCDUQURh+WwB1URp4HT8wOB4repoAN3rHZHAdFxAAnXoYM3dmXOQAMzth9QArHQAPACF3Lx9UQUSAuGnZiJBFpZ4C2zgARVCRPaQZ2zmQP3Jh8l7MVBjbMooAmGqaLExMgABwABnUFAgw2q7lIvQocFeHT26gAjtd4GNNNY0DoGFA4HE4mVKBiaJQxulJncDuphvwaMEwocdGdLpjbvdHjBPgCgaJkAAmdQhVA0WwFADCAhpICRAFYyqFhgAVT7WbmHDrhACSUASsB4YHINC0AEEDTwYI5zvsHnAAL6OoA)

```sh
--parser babel
```

**Input:**

```jsx
verylongidentifierthatwillwrap123123123123123(
  a.b
    // prettier-ignore
    // Some other comment here
    .c
);
```

**Output:**

```jsx
verylongidentifierthatwillwrap123123123123123(
  // prettier-ignore
  // Some other comment here
  a.b
    // prettier-ignore
    // Some other comment here
    .c
);
```

If you feed the output into Prettier again, it duplicates the comment again:

**Input:**

```jsx
verylongidentifierthatwillwrap123123123123123(
  // prettier-ignore
  // Some other comment here
  a.b
    // prettier-ignore
    // Some other comment here
    .c
);
```

**Output:**

```jsx
verylongidentifierthatwillwrap123123123123123(
  // prettier-ignore
  // Some other comment here
  // prettier-ignore
  // Some other comment here
  a.b
    // prettier-ignore
    // Some other comment here
    .c
);
```

This means that every time I save the file, the comment gets duplicated again above the node 😄

**Expected behavior:**
Comment should not be duplicated

fisker:
## Description

Fixes #10661

## Checklist

### DIFF
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -539,15 +539,15 @@ function printDanglingComments(path, options, sameIndent, filter) {
   return indent([hardline, join(hardline, parts)]);
 }
 
-function printCommentsSeparately(path, options, skip) {
+function printCommentsSeparately(path, options, ignored) {
   const value = path.getValue();
   if (!value) {
     return {};
   }
 
   let comments = value.comments || [];
-  if (skip) {
-    comments = comments.filter((comment) => !skip.has(comment));
+  if (ignored) {
+    comments = comments.filter((comment) => !ignored.has(comment));
   }
   const isCursorNode = value === options.cursorNode;
 
@@ -561,7 +561,7 @@ function printCommentsSeparately(path, options, skip) {
 
   path.each((commentPath) => {
     const comment = commentPath.getValue();
-    if (skip && skip.has(comment)) {
+    if (ignored && ignored.has(comment)) {
       return;
     }
 
@@ -596,8 +596,8 @@ function printCommentsSeparately(path, options, skip) {
   return { leading: leadingParts, trailing: trailingParts };
 }
 
-function printComments(path, doc, options, skip) {
-  const { leading, trailing } = printCommentsSeparately(path, options, skip);
+function printComments(path, doc, options, ignored) {
+  const { leading, trailing } = printCommentsSeparately(path, options, ignored);
   if (!leading && !trailing) {
     return doc;
   }
--- a/src/main/comments.js
+++ b/src/main/comments.js
@@ -15,7 +15,6 @@ const {
   addLeadingComment,
   addDanglingComment,
   addTrailingComment,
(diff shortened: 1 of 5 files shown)


thorn0:
Can we skip printing comments that are marked as `printed`? Did you consider this solution?

fisker:
@thorn0 I tried that long time ago, comments not only printed once, so we can't do that.

fisker:
I guess removing comments failed for the same reason, some node is printed multiple times.

thorn0:
Rather because the condition `hasComment(node, CommentCheckFlags.PrettierIgnore)` doesn't hold after the comment is removed.

src/main/comments.js
```diff
@@ -555,6 +561,10 @@ function printCommentsSeparately(path, options) {
 
   path.each((commentPath) => {
     const comment = commentPath.getValue();
+    if (skip && skip.has(comment)) {
+      return;
```
thorn0, line 565: Looks good. Can you add a test to cover this line?

src/main/comments.js
```diff
@@ -540,12 +539,19 @@ function printDanglingComments(path, options, sameIndent, filter) {
   return indent([hardline, join(hardline, parts)]);
 }
 
-function printCommentsSeparately(path, options) {
+function printCommentsSeparately(path, options, skip) {
```
thorn0, line 542: `skip` is a verb. Only functions should be named with verbs. How about `ignored`?

thorn0 (approved):

PR #10672, opened 2021-04-11
Improve lone argument detection

thorn0:
## Description

Another continuation of #10222

Motivation: undesired diffs like the following
```diff
     position2orSize(channel, config.mark) || {
-      [vgChannel]: pointPositionDefaultRef({
-        model,
-        defaultPos,
-        channel,
-        scaleName,
-        scale
-      })()
+      [vgChannel]:
+        pointPositionDefaultRef({
+          model,
+          defaultPos,
+          channel,
+          scaleName,
+          scale
+        })()
     }
```

## Checklist


src/language-js/print/assignment.js
```diff
@@ -230,41 +242,86 @@ function isAssignmentOrVariableDeclarator(node) {
...
+  if (hasComment(node)) {
     return false;
   }
-  const args = getCallArguments(node);
-  return (
-    args.length === 0 || (args.length === 1 && isSimpleCallArgument(args[0], 1))
-  );
+
+  if (
+    node.type === "ThisExpression" ||
```
fisker, line 298: I don't really understand what this function do, but I  guess it's checking short stuff, so `super`? `import.meta`? `new.target`?

fisker: Or `await short`?

thorn0:
```js
const bifornCringerMoshedPerplexSawderGlyphsHa =
  someBigFunctionName("foo")("bar");
```
If each call in RHS is either without arguments or has a single argument, for which this function returns `true`, then breaking after `=` has priority over breaking inside RHS. This way we decrease the likelihood of formatting like this:
```js
const bifornCringerMoshedPerplexSawderGlyphsHa = someBigFunctionName(
  "foo"
)("bar");
```
Yes, there is room for improvement in this function, but it already covers the most important cases and looks good enough. At least for the projects we are testing Prettier on. Lone numbers, short strings, and short identifiers look the worst. It can be polished it later

thorn0: #10222 included similar logic. That's why this specific example is already formatted the preferred way. But the heuristic in #10222 reused `isSimpleCallArgument` from the member chain breaking heuristic, which wasn't exactly a good fit for this case and lead to diffs like the one shown in the [first comment](https://github.com/prettier/prettier/pull/10672#issue-612985282) here.

fisker (approved):

src/language-js/print/assignment.js
thorn0: As for `await short`, `import.meta`,  `new.target`, for now I decided to match only one-word things. `super` can't be passed as a call argument.

fisker: Right.

PR #10700, opened 2021-04-13
Add `formatWithConfig`

### LINKED ISSUE #10698: (Feature request) `resolveConfigFile`, `resolveConfig` and `format` as all-in-one function.
Hello 👋

Thank you for this project! I love it and frankly; I am addicted. Recently, I've dipped my toe in more code generation, and I find myself using Prettier to handle all formatting.

More often than not, I pack Prettier with my code generation utility/package. This way I don't have to care much about the code style, just the actual generation. Less complexity, fewer things to go wrong 🤷 

Anyway, what I find lacking with the current Prettier API is a way to resolve the config file path, config and format in one function. I tend to write this function all the time:

```typescript
import * as prettier from 'prettier';

async function formatWithPrettier(source: string): Promise<string> {
	const configPath = await prettier.resolveConfigFile();
	const config = await prettier.resolveConfig(configPath);
	return prettier.format(source, { parser: 'typescript', ...config });
}
```

Maybe this is something we can bring into the Prettier API?

Maybe something like this?

```typescript
async function formatWithConfig(source: string, parser: BuiltInParserName): Promise<string> {
	const configPath = await resolveConfigFile();
	const config = await resolveConfig(configPath);
	return format(source, { parser: 'typescript', ...config });
}
```

cobraz:
fixes #10698

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -43,8 +43,7 @@ module.exports = {
   },
 
   async formatWithConfig(text, parser) {
-    const configPath = await config.resolveConfigFile();
-    const config = await config.resolveConfig(configPath);
+    const config = await config.resolveConfig(resolveConfig(process.cwd()););
     return formatWithCursor(text, { parser, ...config });
   },
 
--- a/src/index.js
+++ b/src/index.js
@@ -42,6 +42,12 @@ module.exports = {
     return formatWithCursor(text, opts).formatted;
   },
 
+  async formatWithConfig(text, parser) {
+    const configPath = await config.resolveConfigFile();
+    const config = await config.resolveConfig(configPath);
+    return formatWithCursor(text, { parser, ...config });
+  },
+
   check(text, opts) {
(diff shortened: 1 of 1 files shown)


eggsy (changes requested):

duailibe:
Hi @cobraz 

`prettier.resolveConfig(process.cwd())` is not correctly used here. The expected argument of `resolveConfig()` is the path of the file whose contents you'll format. Passing the `cwd` will not necessarily respect overrides.

Also, the recommended usage is to always let Prettier infer the parser to use, and never pass the parser.

A much more useful function would be:

```js
async function format(filepath) {
  const config = await prettier.resolveConfig(filepath);
  return prettier.format(await fs.readFile(filepath, "utf-8"), config);
}
```

But we're not interested in adding that to Prettier, you're welcome to use that in user-land (and even publish as package if it's convenient).

PR #10710, opened 2021-04-15
[Docs] Fix pre-commit hook setup command

bokub:
## Description

In the provided example, `pretty-quick` is installed locally, not globally, which may cause the hook to fail:

```
❯ git commit
.husky/pre-commit: 4: pretty-quick: not found
husky - pre-commit hook exited with code 127 (error)
```
This problem can be fixed by running `"npx pretty-quick --staged"` instead of just `"pretty-quick --staged"` 

Maybe @typicode can confirm?
____

**N.B** I've made changes to documentation only, so everything below is not applicable

## Checklist

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -31,7 +31,7 @@ Install it along with [husky](https://github.com/typicode/husky):
 ```bash
 npx husky-init
 npm install --save-dev pretty-quick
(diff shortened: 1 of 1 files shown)


thorn0 (approved):
Confirmed: https://typicode.github.io/husky/#/?id=locally-installed-binaries

PR #10759, opened 2021-04-22
WIP: Use insertFinalNewline option from editorconfig for Handlebars / Ember files

dcyriller:
## Description

By default, Prettier will remove the final newline from Handlebars / Ember files. Why? Because otherwise it would add an extra DOM text node and in most cases this would be unnecessary. This is the recommended practise in the Ember ecosystem ([source](https://github.com/prettier/prettier/pull/6243#issuecomment-506720622)).

In this PR I propose a small adjustement to this default behavior. Prettier would read `insertFinalNewline` option from editorconfig (for Handelbars / Ember files only). When the option is set to true, Prettier would add a final newline. When set to false, the final newline would be removed.

This PR is still a draft. Indeed I'd like this option to be read only from editorconfig (not user configurable). But I don't know how to do that and I would have to spend some time on it. So, I prefer to discuss the PR first not to engage too much work in case you don't want to go down that path. Please let me know if you are okay with the direction it is taking.

## Links

- [a PR (closed)](https://github.com/prettier/prettier/pull/6243) to add final newlines
- [related discussion](https://github.com/prettier/prettier/pull/9067)

## Checklist


thorn0:
It's difficult to agree with the Ember team's reasoning about doing the newline stripping on the editor's side.

From https://github.com/ember-cli/ember-cli/pull/3440#issuecomment-77604430:

> That could break cases where a component's white-space is pre

Isn't this an edge case that happens really rarely though? Besides, it looks like the problem can be solved by putting a comment at the end of such a template to control how many significant trailing blank lines it should have.

fisker:
I alway like the idea repeat `editorconfig`, not just `handlebars`. Every editor I use take precedence over editor settings, why shouldn't we?

thorn0:
@fisker So are you okay with adding an option controlled only by EditorConfig?

fisker:
Yes.

Rada8732 (approved):

tylerbecks:
Any updates on this? I'm being affected and would love to see this merge!

Issue #10738, opened 2021-04-20
Add a section on ignoreRevsFile to the docs


thorn0:
https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt

It might be a section about updating Prettier in general.

Blakelist7:
@thorn0  Hey! new to Open-Source as well as prettier. Could you help me to request my first PR in prettier in docs?

thorn0:
@Blakelist7 No. Unfortunately, I've got a lot on my plate at the moment.

Blakelist7:
@thorn0 no worries I'll try to figure it out on my own

PR #10805, opened 2021-04-30
Added a Section for ignore-revs-file[docs]

Blakelist7:
## Description

Added a Section for ignore-revs-file from https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt
#10738


thorn0:
Hi, thanks for the PR, but I'm afraid I don't think this really makes the docs more helpful.

PR #10859, opened 2021-05-12
Use fsPromise in `getFileContentOrNull`

fisker:
## Checklist

### DIFF
--- a/src/utils/get-file-content-or-null.js
+++ b/src/utils/get-file-content-or-null.js
@@ -1,21 +1,18 @@
 "use strict";
 
 const fs = require("fs");
+const fsAsync = fs.promises;
 
 /**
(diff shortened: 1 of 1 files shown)


kachkaev:
There's a weird issue on performance of `fs.promises` in the Node.js repo: https://github.com/nodejs/node/issues/37583. Can it affect Prettier given how `fs.readFile` is used?

fisker:
Good to know that, seems already fixed in latest version, and will backport to v12 and v14. Not sure about the performance.  #10852 also use `readFle`

sosukesuzuki (approved):

fisker:
I'm going to merge this, since the `fs.promises` already used in other places, we'll need replace all them if we found it's necessary.

PR #10863, opened 2021-05-12
Reduce use of `Array#reduce()`

fisker:
## Checklist


src/language-html/print-preprocess.js
```diff
@@ -330,9 +330,9 @@ function extractWhitespaces(ast /*, options*/) {
       isIndentationSensitive,
       children: node.children
         // extract whitespace nodes
-        .reduce((newChildren, child) => {
+        .flatMap((child) => {
           if (child.type !== "text" || isWhitespaceSensitive) {
-            return [...newChildren, child];
+            return [child];
```
thorn0, line 335:
```suggestion
            return child;
```

thorn0 (approved):

PR #10864, opened 2021-05-12
Style: Use `dotAll` regexp

fisker:
## Description

~This contains changes from https://github.com/prettier/prettier/pull/10861~

## Checklist


src/language-markdown/print-preprocess.js
```diff
@@ -5,7 +5,7 @@ const { getOrderedListItemInfo, mapAst, splitText } = require("./utils");
 
 // 0x0 ~ 0x10ffff
 // eslint-disable-next-line no-control-regex
-const isSingleCharRegex = /^([\u0000-\uffff]|[\ud800-\udbff][\udc00-\udfff])$/;
+const isSingleCharRegex = /^(.|[\ud800-\udbff][\udc00-\udfff])$/s;
```
fisker, line 8: I'm not familiar with this one, but seems [regexp-tree](https://github.com/DmitrySoshnikov/regexp-tree/blob/500c689d0e6e291ca3a71fa652c2282b5ff425e3/src/compat-transpiler/transforms/compat-dotall-s-transform.js#L9) matches exactly the same range.

thorn0: The entire expression can probably be replaced with `/^.$/us`, but I'm not sure either :)

fisker: https://github.com/DmitrySoshnikov/regexp-tree/blob/500c689d0e6e291ca3a71fa652c2282b5ff425e3/src/compat-transpiler/transforms/compat-dotall-s-transform.js#L14

This line says `.` in regex with `u` means `0x0 ~  \u{10FFFF} ` different range.

fisker: https://github.com/mathiasbynens/regexpu-core/blob/ffd8fff2e31f4597f6fdfee75d5ac1c5c8111ec3/rewrite-pattern.js#L11
Here , the same.

thorn0: Or probably we can use `Array.from(string.slice(0, 2)).length === 1` instead of this regex.

thorn0: No, that won't work for `'🤪🤪'`. It should be `Array.from(string.slice(0, 3)).length === 1` instead.

thorn0: > says . in regex with u means 0x0 ~ \u{10FFFF} different range

That's what this regexp is supposed to match see the line # 6: `// 0x0 ~ 0x10ffff`

fisker: I thought there are 3 ranges...

src/utils/front-matter/parse.js
```diff
@@ -4,12 +4,13 @@ const frontMatterRegex = new RegExp(
   "^(?<startDelimiter>-{3}|\\+{3})" +
     // trailing spaces after delimiters are allowed
     "(?<language>[^\\n]*)" +
-    "\\n(?:|(?<value>[\\S\\s]*?)\\n)" +
+    "\\n(?:|(?<value>.*?)\\n)" +
     // In some markdown processors such as pandoc,
     // "..." can be used as the end delimiter for YAML front-matter.
     // Adding `\.{3}` make the regex matches `+++\n...`, but we'll exclude it later
     "(?<endDelimiter>\\k<startDelimiter>|\\.{3})" +
```
thorn0, line 11: Adding the `s` flag changes the meaning of `.` in this line.

thorn0 (approved):

src/language-markdown/print-preprocess.js
thorn0: Why not check ourselves if there is any difference?
```js
const regex1 = /^([\u0000-\uffff]|[\ud800-\udbff][\udc00-\udfff])$/
const regex2 = /^.$/us;
for (let i = 0; i <= 0xffff; i++) {
  const char1 = String.fromCharCode(i);
  for (let j = -1; j <= 0xffff; j++) {
    const char2 = j === -1 ? '' : String.fromCharCode(j);
    const string = char1 + char2;
    const result1 = regex1.test(string);
    const result2 = regex2.test(string);
    if (result1 !== result2) {
      console.log('😱', i, j, result1, result2);
    }
  }
}
console.log('done');
```

PR #10869, opened 2021-05-12
Remove dead code

fisker:
## Description

According to [coverage](https://codecov.io/gh/prettier/prettier) these types and properties are not covered.

They are still in `node_modules/@typescript-eslint/types/dist/ts-estree.d.ts`, but not in `node_modules/@babel/types/lib/index.d.ts`, someone can confirm they are removed or know how to test?

## Checklist

### DIFF
--- a/src/language-js/print/typescript.js
+++ b/src/language-js/print/typescript.js
@@ -141,10 +141,6 @@ function printTypescript(path, options, print) {
       return "abstract";
     case "TSAsyncKeyword":
       return "async";
-    case "TSDeclareKeyword":
-      return "declare";
-    case "TSExportKeyword":
-      return "export";
     case "TSNeverKeyword":
       return "never";
     case "TSObjectKeyword":
@@ -179,15 +175,6 @@ function printTypescript(path, options, print) {
     case "TSArrayType":
       return [print("elementType"), "[]"];
     case "TSPropertySignature": {
-      if (node.export) {
-        parts.push("export ");
-      }
-      if (node.accessibility) {
-        parts.push(node.accessibility + " ");
-      }
-      if (node.static) {
-        parts.push("static ");
-      }
       if (node.readonly) {
         parts.push("readonly ");
       }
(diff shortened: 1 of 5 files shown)


thorn0:
Looks like these types of nodes don't make it into the AST anymore, but it's difficult to be sure. 
`TSJSDocFunctionType` was blocked from appearing in the AST by https://github.com/typescript-eslint/typescript-eslint/pull/1525
As for `TS...Keyword` nodes, how about adding generic printing logic for them (`node.type.slice(2, -7).toLowerCase()`)?

fisker:
> how about adding generic printing logic for the

Sounds good.  Maybe we can also print unknown `TSJSDoc....Type` with `originalText.slice(...node.range)`?

thorn0:
> Maybe we can also print unknown

Maybe, but at the moment it'll be dead code.

src/language-js/print/typescript.js
```diff
@@ -52,9 +52,24 @@ const {
 
 function printTypescript(path, options, print) {
   const node = path.getValue();
+
+  // TypeScript nodes always starts with `TS`
+  if (!node.type.startsWith("TS")) {
+    return;
+  }
+
+  if (/^TS[A-Z][a-z]+Keyword$/.test(node.type)) {
```
thorn0, line 61:
```suggestion
  if (node.type.endsWith("Keyword")) {
```

fisker: I was hoping to only matches ONE word keyword, not things like `TSBigIntKeyword`. We don't know future two words keyword should be camel case or not.

thorn0: There is no camel case keywords in JS and TS. Things like `instanceof` are written all lowercase.

fisker: Edited in the web editor https://github.com/prettier/prettier/pull/10869/commits/f801f5824c43466c058388b6706f94332067d6fb

thorn0 (approved):

PR #10874, opened 2021-05-12
Fix #5599, long yaml keys trigger explicit mapping

### LINKED ISSUE #5599: YAML: long property names/keys get prefixed with question mark `?`, and colon `:` gets wrapped onto the next line
**Prettier 1.15.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA3OAnAngfQDbQDmOUAhgLZw4CWUOAZqWDDS9QM75GY4wAWpOgA4ADDjACMTeBk4BGAEwBmRSuWqN6rWqVIAOlAAEh+hAhJDekACNSGKyAA0ICAAcY1aO2Sg7GCADuAAp2CN4opKgQ1AAmTjZSYADWcDAAyq5MtITIMBgArnDOfDDkeADqfNTw7JlgcGlh1dSo1VjI4OzezrTsmDBBUoTkpMiMeH3OAFbsAB4AQokp6RRwADK0cGOkE0UgM7Np2XhwAIr5EPDbu86ZspgdWBR48a4YtDDlsfzIorf+fXKUlcHTecD6GHQ8QwcAAjvlqDDBqRhqMkONJiA+uRqLkCnt2MczhcruidpiYKRrF8Yj8kApnHlSNQ8NkAMIQcgjDpQaBbZz5PoAFSp4QxcAAvhKgA)
```sh
--parser yaml
```

**Input:**
```yaml
very_long_name_in_fact_it_is_longer_than_80_characters_123123123123123123123123123:
  foo: "bar"
```

**Output:**
```yaml
? very_long_name_in_fact_it_is_longer_than_80_characters_123123123123123123123123123
: foo: "bar"

```

**Expected behavior:**

Doesn't wrap the `:` to the second line, i.e. leaves the whole thing intact.

```yaml
very_long_name_in_fact_it_is_longer_than_80_characters_123123123123123123123123123:
  foo: "bar"
```

This issue looks simple to fix, where would I look for the related code if I wanted to make a PR?

pdavies:
## Description

Hi, this PR fixes #5599, in which long YAML keys cause us to print using yaml's explicit mapping syntax. The intent of the code was to use explicit mapping syntax to accommodate multiline keys, but it had the unintended effect of using explicit mapping syntax for long single-line keys too. This is bad because:

* Doing so has the effect of further _increasing_ the line length
* Explicit mappings are obscure and most people have never seen them. It causes confusion to introduce them needlessly, as shown in the issue

As others in the issue have noted, this is a pretty non-obvious area of code, so I've added some comments and used what I believe are more helpful variable names.

### Aside
It might be interesting to someone with more experience to see the (pre-existing) (ab)use of `conditionalGroup` here. Based on the documentation for `conditionalGroup`, it sounds like there shouldn't be any value in calling it only to give it an array with a single `Doc` in it. `conditionalGroup` should just pick the only option you've given it. But actually, whoever wrote this yaml code seems to have discovered that `conditionalGroup` is a crafty way to force printing with `MODE_FLAT`, and the yaml formatter depends extensively on this fact. This is the only use of `conditionalGroup` anywhere in prettier where it's used for this purpose and not for any conditioning. It is perhaps worth considering whether prettier could expose a function to print with `MODE_FLAT` by design, but that would be separate from this PR.

## Checklist

~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~

### DIFF
--- a/changelog_unreleased/yaml/10874.md
+++ b/changelog_unreleased/yaml/10874.md
@@ -1,24 +1,9 @@
-<!--
-
-
-4. Fill in a title, the PR number and your user name.
-
-5. Optionally write a description. Many times it’s enough with just sample code.
-
-6. Change ```jsx to your language. For example, ```yaml.
-
-7. Change the `// Input` and `// Prettier` comments to the comment syntax of your language. For example, `# Input`.
-
-8. Choose some nice input example code. Paste it along with the output before and after your PR.
-
--->
-
 #### Fix #5599, long yaml keys trigger explicit mapping (#10874 by @pdavies)
 
 Fixes #5599, in which long YAML keys cause us to print using yaml's explicit mapping syntax. The intent of the code was to use explicit mapping syntax to accommodate multiline keys, but it had the unintended effect of using explicit mapping syntax for long single-line keys too. This is bad because:
 
-* Doing so has the effect of further increasing the line length
-* Explicit mappings are obscure and most people have never seen them. It causes confusion to introduce them needlessly, as shown in the issue
+- Doing so has the effect of further increasing the line length
+- Explicit mappings are obscure and most people have never seen them. It causes confusion to introduce them needlessly, as shown in the issue
 
 <!-- prettier-ignore -->
 ```yaml
--- a/changelog_unreleased/yaml/10874.md
+++ b/changelog_unreleased/yaml/10874.md
@@ -1,0 +1,36 @@
+<!--
+
+
+4. Fill in a title, the PR number and your user name.
+
+5. Optionally write a description. Many times it’s enough with just sample code.
+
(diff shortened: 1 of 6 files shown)


thorn0:
Thanks for the PR. I haven't check the code yet, but I have some comments already:

> Doing so has the effect of further increasing the line length

Only if the `proseWrap` isn't set to `always`. With `proseWrap: always`, the key can become multiline, so in that case the current logic should be preserved I guess. Does your PR take that into account?

thorn0 (approved):
Looks good.

Extra thanks for pointing out that interesting use of `conditionalGroup`. I'm going to investigate it.

Also, while testing this PR, I found another bug: #10879

sosukesuzuki (approved):

PR #10881, opened 2021-05-15
Merge Tips section with Another Tips section

induratized:
It was confusing at first 'Another tip! If your project isn’t ready to format, say, HTML files yet, add `*.html`.' . as to where `adding *.html` files was required. 
Merging these two tips in one section would appear second tip dependent on first tip. (👍 )

## Checklist

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -37,7 +37,6 @@ coverage
 ```
 
 > Tip! Base your .prettierignore on .gitignore and .eslintignore (if you have one).
-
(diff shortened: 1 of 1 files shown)


thorn0:
> second tip dependent on first tip

It's not depended though. They're two separate tips.

thorn0:
Thanks for the PR, but I'm afraid I don't think this really makes the docs more helpful.

PR #10921, opened 2021-05-21
Replace `lock-threads` with `stale`

fisker:
## Description

`stale` is a Github application

https://github.com/marketplace/stale

---

Update: I think it needs to install first, but I can't.

## Checklist


thorn0:
They do different things. We don't want to **close** issues automatically, only to lock closed issues.

fisker:
Ah, sorry. You are right.

PR #10926, opened 2021-05-22
[YAML] Allow to set amount of spaces between comments and content

tumido:
## Description

YAML linters often by default expect more than 1 space between content on line and a trailing comment. E.g. [default comment settings for `yamllint`](https://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.comments). This PR allows user to specify the amount of spaces inserted.

PR is adding a new YAML specific, option which defaults to maintain the current behavior - a single space.

Supports requested features for VSCode YAML plugin https://github.com/redhat-developer/vscode-yaml/issues/433

## Checklist


fisker:
I'm sorry, but we are not going to accept this,  see https://prettier.io/docs/en/option-philosophy.html

ssbarnea:
@fisker While I can understand the philosophy and the reasoning behind, I also wonder if prettier has a process for fixing what could be called a historical mistake? Maybe it would worth evaluating if the solution for this problem is not to just ensure that inline comments are always using two space indentation instead of just one space.

There are good reasons why python decided to use double space indentation for all in-line comments, is accessibility and visibility as it makes it easier to spot comments and to not confuse them with the rest of the line. This can become quite tricky for the eye if a line could look like `foo: bar * 3 # 4`. Once extra space makes a bit difference when looking at a file.

kachkaev:
> Once extra space makes a bit difference when looking at a file.

Hmmm not sure I see how. Here’s a screenshot from this PR’s diff:

Screenshot 2022-01-17 at 18 52 55

I can take my glasses off and still clearly see comments in both _Before_ and _After_. Syntax highlighting does a great job here because _color_ visual variable is one of the strongest ones.

---

🟪🟪🟪🟪🟪🟧🟧

↑ It is obvious that this line has two categories of squares (and I used zero spaces)

akumria:
@fisker The philosphy you quote says:

> Some were added for compatibility reasons. 👍

The intent behind this change is to support compatibility with the Python ecosystem of formatters (namely Black) and yamllint which both use two spaces before a comment mark.

Determining why two different formatting systems are conflicting with each other (prettier vs yamllint) is really a significant time-sink that could be avoided with this modification.

I'm just spent ~2 hours this week doing this *because* prettier is incompatible with other systems.

ssbarnea:
I am in the middle here and using both projects considerably.

It worth mentioning that I kinda see the prietter project considerably more inclusive and considerate regarding external contributions. You can check the previous contributors on both and you will realise that yamllint is one-man-project and he is not willing to compromise. Basically yamllint defaults are author views about how things should be, not result of a group of people consensus. Some might even guess it from the URL of the project, is not an org.

* https://github.com/adrienverge/yamllint/graphs/contributors
* https://github.com/prettier/prettier/graphs/contributors

Prettier paints a different picture. While it fights adding more complexity, it does this for genuine reasons (similar to python black). I would say that adding options like this should involve acceptance or rejection from more than one major contributor to the project, mainly because they know that maintenance burden cost better than anyone else.

Meanwhile you could try to use `ansible-lint` as a generic yaml linter, which has defaults compatible with prettier. It uses yamllint itself.

conclusionlogic:
This feature would have been really nice...

JuryA:
I fully understand your philosophy, but you are wrong! I do not care if Yamllint is a one-man project - this is irrelevant! Two spaces before the inline comment are based on this: https://peps.python.org/pep-0008/#inline-comments and from this, it was adopted by flake8 and others! Prettier - maybe - has nice philosophy but this is not about philosophy, it's about understanding who is and who is not competent to make decisions about such questions! I still do not understand why you cannot simply explain why Prettier insists on one space and argument that "our philosophy want to not support style wars, etc." 

With this disrespect, Prettier is sending a signal that "the community around Prettier wants it this way and we will enforce it!" This is not the right way to gain natural authority in enforcing such decisions! Then your entire philosophy is completely irrelevant, because you cannot demand respect from others for a philosophy that does not respect their philosophy! I'm really upset with this! Again ... Prettier autoformatted bunch of YAML configuration files which linter do not accept because of some Prettier philosophy! No offense guys! That is reality!

Issue #4723, opened 2018-06-21
[YAML] indentation style for sequence in mapping


ikatyang:
I'd like to avoid things like #4251 in the future, let's see how people think about it. We'll decide which style to use before releasing 1.14 based on the 👍/👎 and change the style if one's vote (👍s - 👎s) is greater than another one about 30%, otherwise stay as-is (i.e., do indent).

<table>
    <tbody>
        <tr>
            <th>do indent</th>
            <th>do not indent</th>
        </tr>
        <tr>
            <td align="center">better readability</td>
            <td align="center">no unnecessary indentation</td>
        </tr>
        <tr>
            <td><pre>a:
  - b
  - c</pre><pre>before_script:
  - command
  - command
  - command
stages:
  - test
  - test
  - test
other:
  - other
  - other
  - other</pre></td>
            <td><pre>a:
- b
- c</pre><pre>before_script:
- command
- command
- command
stages:
- test
- test
- test
other:
- other
- other
- other</pre></td>
        </tr>
    </tbody>
</table>

ikatyang:
Please 👍/👎 this comment for the `do indent` style

ikatyang:
Please 👍/👎 this comment for the `do not indent` style

jaydenseric:
At first glance, I like the indented look. Consistency with Markdown might be a consideration, where the top list level is not indented:

<img width="148" alt="screen shot 2018-06-21 at 1 14 24 pm" src="https://user-images.githubusercontent.com/1754873/41696022-3f336a78-7555-11e8-8709-3a5dc058f8a1.png">

I guess the YAML dashed list items are actually second level though, depending how you look at it.

Issue #9853, opened 2020-12-09
YAML - Flag for Optional Indentation on Hyphenated List


timduhenchanter:
In reference to https://github.com/prettier/prettier/issues/4723 and https://github.com/redhat-developer/vscode-yaml/issues/172

Due to the Red Hat VSCode extensions dependency on Prettier and formatting standards at my company as well as other editors is it possible to have an optionally flag for the additional indentation of YAML lists?

smyrman:
To give a perhaps better context. A number of tools, at least in the world of Kubernetes, auto-formats YAML to omit additional spaces in sequence lists. I.e. the opposite of what was decided in #4723. As far as I understand this issue, it requests adding an _option_ to not indent _sequences_ in YAML. E.g. `-no-indent-sequence` or something like that.

**Prettier 2.3.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzCEkB0oFoAEARgIYBOOBJAXiADQgQAOMAltAM7KhmkQDuABTIJOKYgBs+xAJ6d6hUsTABrODADKjJSygBzZDFIBXOPTgBbQnAAm1mwBlieo8V1wAYhFLniMVnuQQYiMYCDoQAAsYc3EAdQiWeHYtMDh1EUSWADdE6UCwdjkQHXY4UhgBRV0fZFQJUvoAK3YADwAhRRU1dWJzOHsdOFr60xBmlvUdXXE4AEUjCHhh8QaQLVJS0kDpXvFwxlIdGFiWaxgI5AAOAAZ6A4hS2MVGQIO4Tayh+gBHBfhKpiiILsXBQOA2GzhUhwX4saGVVw1JB1FajUrmFgGYxoqYzeaLIbIkb0GDEQgnM4XJAAJhJihY4imAGEIOYkSB3gBWcJGUoAFTJohRqyyJgAklA7LB1GBDswAIKS9QwaQzZalAC+GqAA)

**Input:**

```yaml
foo:
- bar
- baz
```

**Output:**

```yaml
foo:
  - bar
  - baz
```

**Expected behavior:**

```yaml
foo:
- bar
- baz
```


While running prettier _after_ each tool command is an option, it quickly becomes a nuance.

I appreciate that Prettier is an opinionated formater, and that adding options should not be done without a justifiably reason.

[...]

PR #10927, opened 2021-05-22
[YAML] Add option for do-not-indent style for sequence in mapping

tumido:
## Description

This PR goes **against** the #4723. With extensive popularity of `do-not-indent` style among Kubernetes and golang projects I feel it's worth at least adding an option for users to enable this behavior.

This PR adds a `--do-not-indent` YAML specific option. This option is off by default, which maintains the current behavior.

This supports a quite heavily requested feature by YAML VSCode extension users: https://github.com/redhat-developer/vscode-yaml/issues/172

## Checklist


ludusrusso:
Lgtm

fisker:
I'm sorry but we are not going to accept https://prettier.io/docs/en/option-philosophy.html

smyrman:
Is there an option to _detect_ Kubernetes object files and format them differently as discussed in #9853?

May involve matching files that match one or more of of:
1. Contain `apiVersion` and `kind` as top-level keys.
2. Are named `kustomization.yaml` or `kustomization.yml`.

Alternatively ignoring rule number 2 as kustomization files can also be written with an `apiVersion` and `kind` key.

tumido:
I think it is technically possible, sure. However I'm not convinced that it's an acceptable solution:

1. It would make Prettier behave differently for the same language in different context. Is there any precedent for such behavior in Prettier? Would that be acceptable to maintainers?
2. How should we handle multidocument YAMLs, where one doc can be detected as a Kubernetes resource and another document in the same file (due to typo for example) is not? Do we format each document to a different standard?

I'm not sure if introducing a context aware dual behavior of the formatter is the right response to "new CLI flags are forbidden". I think @fisker is very polite about this but the decision seems final and I respect that.

I'm personally more inclined to opening this PR against [`prettierx`](https://github.com/brodybits/prettierx) and trying to convince redhat-developer/vscode-yaml to replace `prettier` with it. I know it's not an ideal solution either...

dex4er:
@fisker I can't use Prettier for Kubernetes YAML files right now because Prettier is not compatible with the default output from kubectl commands and other tools from this landscape. This option would be very helpful and could make that I would introduce Prettier as a default formatter.

I think this particular case should be a legit exception for your "no options" policy.

fisker:
> Is there any precedent for such behavior in Prettier?

Actually, we do, we format `.prettierrc` and `.stylelintrc` differently from other YAML files , #8105 and #10924.

> Would that be acceptable to maintainers?

Personally, I think it's fine.

tumido:
Ok, in that case I've changed this PR to define the `doNotIndent` flag internally only - it's not user facing anymore, but inferred for any Kubernetes resource. @fisker wdyt about it now? :slightly_smiling_face:

ggirard07:
Not so sure go-yaml v3 will change anything about this to the current landscape, they are already adding workaround for backward compatibility!
https://github.com/go-yaml/yaml/pull/753

shadiramadan:
[thorn0](https://github.com/thorn0) I disagree.

When querying anything from kubernetes with yaml output it is in dedented format.  This is not simply a 'vscode plugin issue'

e.g. `kubectl get deployment -o yaml`

Declarative yaml is the core of kubernetes.  It doesn't make sense to me to force prettier's opinion on it.

RafalSkolasinski:
For Kubernetes de-indented format on list in yamls is de-facto standard. The multi-level manifests are otherwise just unreadable. 

Ideal would be if one could just treat it as separate syntax, e.g. `k8s-yaml` instead of just `yaml`.

kachkaev:
I _think_ Prettier does a similar distinction between _any JSON_ and `package.json`. So I guess that formatting k8s YAMLS differently based on the presence of `apiVersion` & `kind` would be OK 🤔 There’s a clear benefit for k8s users. Can we imagine any negative side effects for other devs?

shadiramadan:
@RafalSkolasinski a lot of the tooling around k8s is actually pretty good about also using apiVersion/kind

e.g.
https://skaffold.dev/docs/references/yaml/
https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/

shadiramadan:
@thorn0 I went down a rabbit hole for this point
> Now that go-yaml v3 has switched to consistent indentation, there is a considerable chance that the tools that use that library will switch to that style too.

https://github.com/kubernetes-sigs/kustomize/issues/4033

I don't think that will happen- go-yaml was forked just to prevent this indentation change until it becomes an option upstream.

I really don't see this changing and it seems silly to to ignore the "3.9 million Kubernetes developers worldwide" who work with k8s yaml.

https://www.cncf.io/wp-content/uploads/2022/02/CNCF-AR_FINAL-edits-15.2.21.pdf

shadiramadan:
@thorn0 Oh wow your profile says Kyiv.  Best of luck, hope you are safe.

shadiramadan:
@fisker Do you know what it would take to get this in?

shadiramadan:
Are there any more discussion / requirements required from the prettier team for this?  What are the blockers / questions to getting an approval?  Is there something wrong with the PR code?

fisker (changes requested):

nkkowa:
If an exception is being made for Kubernetes resources, should the same be made for Helm `Chart.yaml` and `values.yaml` files? `Chart.yaml` has certain keys that could be checked, though `values.yaml` is fairly free-form, Values files could also be named `xyz-values.yaml` by convention, such as `production-values.yaml` and `staging-values.yaml`.

mikedld:
Looking at #4723, I'd say the counts are 1) slightly rigged and 2) not representative.

For p. 1, getting the full lists of voters there's an intersection, meaning some people voted for more than a single option:

<details><summary>Recount details</summary>

User | Do indent (398956397) | Do not indent (398956407) | Date
-|-|-|-
@JPinkney | 👍 | 👎 ❌ | 2018-06-21
@TheDancingCode | 👍 | | 2018-06-26
@ThomOrlo | 👍 | 👎 ❌ | 2018-06-22
@aaronjameslang | 👎 ❌ | 👍 | 2018-06-21
@alaister | 👍 | | 2018-06-21
@alexander-akait | 👍 | 👎 ❌ | 2018-06-21
@alexanderbird | 👍 ❌ | | 2018-09-15
@andrewrod33 | 👍 | 👎 ❌ | 2018-06-26
@ascorbic | 👍 | 👎 ❌ | 2018-06-21
@cggruszka | 👍 ❌ | | 2018-10-03
@azz 👨‍💻 | | 👍 | 2018-06-21
@diescake | 👍 | 👎 ❌ | 2018-07-17
@drewvolz | 👍 | | 2018-06-21
@eemeli | 👍 | 👎 ❌ | 2018-06-25
@hawkrives | 👍 | | 2018-06-21
@hrobertson | 👍 ❌ | | 2018-10-11
@ikatyang 👨‍💻 | 👍 | | 2018-06-21
@ikorolev93 | 👍 | 👎 ❌ | 2018-07-02
@j-f1 👨‍💻 | 👍 | 👎 ❌ | 2018-06-21
@jaydenseric | 👍 | | 2018-06-21
@lipis 👨‍💻 | 👍 | | 2018-06-21
@listepo-alterpost | 👍 | | 2018-07-02
@lydell 👨‍💻 | 👍 | 👎 ❌ | 2018-06-22
@ndxbn | 👍 | 👎 ❌ | 2018-06-24
@ostrokach | 👍 | | 2018-07-12
@suchipi 👨‍💻 | 👍 | 👎 ❌ | 2018-06-21
@thorn0 👨‍💻 | 👍 | 👎 ❌ | 2018-07-04
@vivian-farrell | 👍 | 👎 ❌ | 2018-06-21
@vvs | 👍 | 👎 ❌ | 2018-06-21

</details>

Voting was closed on 2018-07-17, so I'm not counting 3 votes that came in later:

Vote | Official | Recount | Recount (sans contributors) | Recount (contributors only)
-|-|-|-|-
Do indent | 23 (???) | 24 (92%) | 18 (95%) | 6 (86%)
Do not indent | -12 (???) | 2 (8%) | 1 (5%) | 1 (14%)

This means 26 (and not 41) people took part, and the difference is 22, not 35.

Now you may think "well, `do indent` still wins, what's your point?" For p. 2, I don't have the counts at the time of the vote, but right now there're 43.3k stargazers, 412 watchers, and 3.4k forks. To me, taking 26 (or 19, if you don't count contributors) votes as a good measure of public opinion doesn't feel right. The voting somehow passed by on the down low, not sure why.

Limiting non-indented formatting for K8s files only isn't that helpful to me. I just happen to prefer this style in all my yaml files.

CH-DrewWatson:
How did this get so stale?

RafalSkolasinski:
> if Prettier maintainers are against this option being part of their project. shrug

It should be a separate flavour of yaml: `k8s-yaml` or `ansible-yaml`... 
That should not go against the philosophy to have opinionated yaml formatting.

pre:
I would love to see this happen. Our team already has Prettier&co in VSCode for the JS projects, but the same formatter can’t be used in our CD git-repo of Kubernetes manifests. 

It’s too bad the default formatting is in contrast with Prettier’s opinion. Changing the kubernetes formatting is not possible because tools such as Kustomize, Helm and even kubectl -o yaml, produce yaml without indenting the array sequences. 

Luckily there’s another (limited) kubernetes yaml formatter for vscode. But that’s feels like a workaround for lack of Prettier’s support for this option.

tumido:
I'm closing this PR now, the discussions seem to stall, people found other workarounds and decided not to use prettier for kube YAML formatting... Let's move on to something more fruitful. 🙂 

If somebody wants to pick this work up, feel free to reuse my code, you have my full support and I wish you much better outcome than I accomplished. 🙂 ❤️ 

See discussion on https://github.com/redhat-developer/vscode-yaml/issues/172 for possible workarounds. Sorry. 😞

PR #10948, opened 2021-05-25
Special format `cspell.json` file

fisker:
## Description

Our first file don't have a final new line. 😄 

## Checklist


sosukesuzuki (commented):
I understand the issue, but I don't think it is the good to support special formatting for `cspell.json`. `cspell.json` is not a frequently updated file. So I think it's enough if it's in `.prettierignore`. Alternatively, we should fix https://github.com/streetsidesoftware/vscode-spell-checker/issues/391 on the cspell side.

alexander-akait:
> Alternatively, we should fix streetsidesoftware/vscode-spell-checker#391 on the cspell side.

I agree too

PR #10954, opened 2021-05-27
Minor tweak to js printer

fisker:
## Description

- Handle `NG-*` node in `angular.js`
- Rename `printFunctionDeclaration` to `printFunction` to be more correct, it handles expressions too.
- Move `expandArg` logic to `printFunction`
- Rename `printReturnAndThrowArgument` to `printReturnOrThrowArgument`
- Move `returnArgumentHasLeadingComment` from `utils.js` to `function.js`

`returnArgumentHasLeadingComment` should have a better name, because it also checks `throw` argument, but `returnOrThrowArgumentHasLeadingComment` seems too long, any idea? 

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -71,7 +71,7 @@ module.exports = {
     "prettier-local": "<rootDir>/tests/config/require-prettier.js",
     "prettier-standalone": "<rootDir>/tests/config/require-standalone.js",
   },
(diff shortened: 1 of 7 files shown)


thorn0 (approved):
> should have a better name, because it also checks throw argument

It can stay as is because it's really important to get comments right only for `return`.

sosukesuzuki (approved):

PR #11004, opened 2021-06-02
Fix parenthesis for optional function as return type

vjeux:
## Description

Parenthesis is missing after prettier which produces incorrect result and a syntax error.

## Checklist


fisker:
The tests are failing.

tests/format/flow/return-arrow/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -136,7 +136,7 @@ function f5(): string => string {}
 
 =====================================output=====================================
 const f1 = (): ((string) => string) => {};
-const f2 = (): (?(y) => { a: (b) => c }) => (0: any);
+const f2 = (): (?((y) => { a: (b) => c })) => (0: any);
```
fisker, line 139: @vjeux Is this expected?

vjeux: I’m not sure. I haven’t dug into whether this was correct with this change or not.

fisker: I don't think extra parens will break anything, let's merge as it is.

If we  found it's moot, we can improve later. WDYT?

fisker (approved):

tests/format/flow/return-arrow/__snapshots__/jsfmt.spec.js.snap
fisker: Updated to add to `NullableTypeAnnotation` instead https://github.com/prettier/prettier/pull/11004/commits/df35bb6dd3601ac195d6b469287a738b8dcfffa8

PR #11007, opened 2021-06-03
Keep editor from removing trailing whitespace in format test files and snapshots

adalinesimonian:
Keep editor from removing trailing whitespace in format test files and snapshots.

## Description

Configures editorconfig to avoid trimming EOL whitespace in format test files and snapshots.

This change was originally proposed downstream in https://github.com/brodybits/prettierx/pull/549 but was determined to be a better candidate for an upstream change. (cc @brodybits)

## Checklist

- [x] N/A ~~I’ve added tests to confirm my change works.~~
- [x] N/A ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~~
- [x] N/A ~~(If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.~~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,3 +14,9 @@ insert_final_newline = false
 
 [website/blog/*.md]
 trim_trailing_whitespace = false
+
+[tests/{**/__snapshots__/*, tests/format/**/*}]
+trim_trailing_whitespace = false
+
+[tests/format/**/jsfmt.spec.js]
+trim_trailing_whitespace = true


fisker:
I suggest use `snippets` if you want test trailing whitespace https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md#tests

adalinesimonian:
> I suggest use `snippets` if you want test trailing whitespace https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md#tests

Do you mean with respect to prettier's own .editorconfig configuration? This change doesn't change any of prettier's code.

fisker:
I don't think we need this change, we normally don't care about trailing whitespace, if we need to test it, we do

```js
run_spec(
  {
    dirname: __dirname,
    snippets: [
      'foo;     \nbar;'
	]
  },
  parsers: ['babel']
);
```

This is how we test all kinds of whitespace https://github.com/prettier/prettier/blob/main/tests/format/html/whitespace/jsfmt.spec.js

adalinesimonian:
> I don't think we need this change, we normally don't care about trailing whitespace, but if we need to test it, we do
> 
> ```js
> run_spec(
>   {
>     dirname: __dirname,
>     snippets: [
>       'foo;     \nbar;'
> 	]
>   },
>   parsers: ['babel']
> );
> ```
> 
> Here is how we test all kinds of whitespace https://github.com/prettier/prettier/blob/main/tests/format/html/whitespace/jsfmt.spec.js

There are currently 279 files in `tests/` with trailing whitespace in the main branch, all of which will have that whitespace deleted when saved using any editor that supports editorconfig. Deleting the whitespace in some of these files, certainly, as you point out, doesn't really affect anything, but in other files, it does.

In merging changes between prettier 2.0 and 2.3 into prettierx, we encountered multiple test suites where the elimination of EOL whitespace, whether in a snapshot or a test file, resulted in test errors. Of course, these failures were resolved, either by running `jest -u` or by editing .editorconfig to allow the editor to save EOL whitespace and reverted afterwards. But these four additional lines in the root .editorconfig file would have avoided this whole corner case entirely and saved time spent head-scratching and making corrections.

I believe this change is worth it since it is minuscule and doesn't affect any other part of the codebase.

fisker:
>There are currently 279 files in tests/ with trailing whitespace in the main branch 

I didn't know that. But adding this may just increase this number. Personally I prefer to remove them from the fixtures, but may need time. So I guess I'm fine with this change, but definitely worth to spend some time to remove them or move them to `snippets` in future.

fisker (approved):

adalinesimonian:
> >There are currently 279 files in tests/ with trailing whitespace in the main branch 
> 
> I didn't know that. But adding this may just increase this number. Personally I prefer to remove them from the fixtures, but may need time. So I guess I'm fine with this change, but definitely worth to spend some time to remove them or move them to `snippets` in future.

I completely agree – you are right that there is the risk that this change may increase the number of tests with this issue. Perhaps sorting those tests out and making tests involving EOL whitespace use the mechanism you've described, or making a more specific rule in .editorconfig, may be an action item after this PR.

thorn0 (approved):

adalinesimonian:
I thought I'd check in to see if any issues are blocking this from being merged that I may need to address?

PR #11038, opened 2021-06-07
Fix lint:deps script

adalinesimonian:
## Description

The `check-deps` script is failing because of a mistake from #10867 in creating the URL to read `package.json`. This failure can be [seen in CI runs of lint:deps](https://github.com/prettier/prettier/runs/2572587849#step:5:6), but for whatever reason, it is not causing the `Lint` action to fail.

This PR [allows the script to pass](https://github.com/prettier/prettier/pull/11038/checks?check_run_id=2767164627#step:5:5).

## Checklist

- [x] N/A ~~I’ve added tests to confirm my change works.~~
- [x] N/A ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~~
- [x] N/A ~~(If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.~~

### DIFF
--- a/scripts/check-deps.mjs
+++ b/scripts/check-deps.mjs
@@ -8,7 +8,10 @@ import chalk from "chalk";
 
   validateDependencyObject(packageJson.dependencies);
   validateDependencyObject(packageJson.devDependencies);
-})();
+})().catch((error) => {
+  console.error(error);
+  process.exit(1);
+});
 
 function validateDependencyObject(object) {
   for (const key of Object.keys(object)) {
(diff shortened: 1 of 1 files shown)


adalinesimonian:
Failing action [seems to be unrelated to changes in this PR](https://github.com/prettier/prettier/pull/11038/checks?check_run_id=2767164621#step:7:53).

fisker (approved):
Thanks for fixing my mistake, and other mistakes.

fisker:
I was going to use top level await at first, but forgot to add `.catch` when refactoring.

PR #11044, opened 2021-06-11
Revert "Revert features for minor"

fisker:
Reverts prettier/prettier#11016

Restore changes in #10924

We may want keep this open until we decide not to do path release for v2.3


sosukesuzuki (approved):
Thank you

PR #11139, opened 2021-06-30
Update plugins.md

kamiazya:
## Description

Set the return value of the embed function to null so that there are no type errors.

## Checklist

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -251,7 +251,7 @@ function embed(path, print, textToDoc, options) {
   if (node.type === "javascript") {
     return textToDoc(node.javaScriptText, { ...options, parser: "babel" });
   }
-  return false;
+  return null;
 }
 ```
 


fisker:
Can we update the signature? And remove this `return`, we didn't do `return null` in our codebase, except [this one](https://github.com/prettier/prettier/blob/57f8eec8e65290a0dbb35b4551e152bb4cd48e77/src/language-markdown/embed.js#L66).

kamiazya:
If you change the TypeScript signature to remove the return, we need to modify it to allow `undefined` as below.

```ts
    embed?: (
        path: AstPath<T>,
        print: (path: AstPath<T>) => Doc,
        textToDoc: (text: string, options: Options) => Doc,
        options: ParserOptions<T>,
    ) => Doc | null | undefined;
```

[source](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/prettier/index.d.ts#L203-L208)

We could do the following, but it would affect the existing TypeScript codebase.

```ts
    embed?: (...) => Doc | undefined;
```

PR #11241, opened 2021-07-22
chore: drop mem dependency

43081j:
## Description

drop mem entirely to reduce the amount of bloat in the dependency chain.

@fisker as you said, "it's about 350 lines in the bundle", here we have it in a lot less. if you really want to save a few more lines, im sure you could reduce a few of my conditionals etc further

@sosukesuzuki 4 less dependencies, less code in the build, faster CI install. you had concerns in #11230 you never managed to explain to us, presumably you closed to due to a downside we're not aware of so it would be much appreciated if you could share what that is.

## Checklist

- [] I’ve added tests to confirm my change works.

cc @alexander-akait

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -5,31 +5,41 @@ const path = require("path");
 const uniqBy = require("lodash/uniqBy");
 const partition = require("lodash/partition");
 const globby = require("globby");
-const mem = require("mem");
 const internalPlugins = require("../languages.js");
 const thirdParty = require("./third-party.js");
 const resolve = require("./resolve.js");
+const { isNonEmptyArray } = require("./util.js");
 
-const memoizedLoad = mem(load, { cacheKey: JSON.stringify });
-const memoizedSearch = mem(findPluginsInNodeModules);
+const nodeModulesPluginCache = new Map();
+let loadCache = new WeakMap();
+
 const clearCache = () => {
-  mem.clear(memoizedLoad);
-  mem.clear(memoizedSearch);
+  nodeModulesPluginCache.clear();
+  loadCache = new WeakMap();
 };
+const defaultPluginSearchDirs = [];
+const defaultPlugins = [];
 
-function load(plugins, pluginSearchDirs) {
+function loadPlugins(plugins, pluginSearchDirs) {
   if (!plugins) {
-    plugins = [];
+    plugins = defaultPlugins;
   }
 
-  if (!pluginSearchDirs) {
-    pluginSearchDirs = [];
+  if (!isNonEmptyArray(pluginSearchDirs)) {
+    pluginSearchDirs = defaultPluginSearchDirs;
   }
+
+  const cachedSearchDir = loadCache.get(pluginSearchDirs) ?? new WeakMap();
+  const cachedPlugins = cachedSearchDir.get(plugins);
+  if (cachedPlugins) {
+    return cachedPlugins;
+  }
+
   // unless pluginSearchDirs are provided, auto-load plugins from node_modules that are parent to Prettier
   if (pluginSearchDirs.length === 0) {
     const autoLoadDir = thirdParty.findParentDir(__dirname, "node_modules");
     if (autoLoadDir) {
-      pluginSearchDirs = [autoLoadDir];
+      pluginSearchDirs.push(autoLoadDir);
     }
   }
 
@@ -80,7 +90,7 @@ function load(plugins, pluginSearchDirs) {
         );
       }
 
-      return memoizedSearch(nodeModulesDir).map((pluginName) => ({
+      return findPluginsInNodeModules(nodeModulesDir).map((pluginName) => ({
         name: pluginName,
         requirePath: resolve(pluginName, { paths: [resolvedPluginSearchDir] }),
       }));
@@ -98,10 +108,19 @@ function load(plugins, pluginSearchDirs) {
     ...externalPluginInstances,
   ];
 
-  return [...internalPlugins, ...externalPlugins];
+  const result = [...internalPlugins, ...externalPlugins];
+
+  cachedSearchDir.set(plugins, result);
+  loadCache.set(pluginSearchDirs, cachedSearchDir);
+
+  return result;
 }
 
 function findPluginsInNodeModules(nodeModulesDir) {
+  if (nodeModulesPluginCache.has(nodeModulesDir)) {
+    return nodeModulesPluginCache.get(nodeModulesDir);
+  }
+
   const pluginPackageJsonPaths = globby.sync(
     [
       "prettier-plugin-*/package.json",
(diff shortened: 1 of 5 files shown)


fisker:
> drop mem entirely to reduce the amount of bloat in the dependency chain.

Production Prettier has no dependencies at all, we bundle everything. https://unpkg.com/browse/prettier@2.3.2/package.json

43081j:
of course, it doesn't have dependencies as they are already bundled. here we've saved 4 dependencies worth of code from production prettier, which originated from dependencies.

meanwhile, the more of these bloated micro-packages we remove, the faster CI builds will be. absolutely no reason pulling 4 dependencies in for what can be done with your own map IMO.

if there's a downside to this, please do let me know. as far as i can see, your bundle will be smaller, your CI will be faster.

src/config/resolve-config-editorconfig.js
```diff
@@ -3,11 +3,9 @@
 const path = require("path");
 
 const editorconfig = require("editorconfig");
-const mem = require("mem");
 const editorConfigToPrettier = require("editorconfig-to-prettier");
 const findProjectRoot = require("./find-project-root.js");
-
-const jsonStringifyMem = (fn) => mem(fn, { cacheKey: JSON.stringify });
+const configCache = new Map();
```
fisker, line 8: This changes the behavior, you are sharing cache between async and sync version, which is good, but I think it's not this simple.

The old async version not caching the result, but the promise, this will cause `Promise.all()` to search multiple times.

43081j: i think i found it. wouldn't it mean a potential race condition more than anything? if we call it twice, quickly, neither will have cached yet and both will run whereas before it would've waited on one promise.

i can change to having two caches again if needed

fisker: It's not about two caches, the async version need cache the promise, not result.

43081j: yes, which means two caches in this case (we're keying by the same thing). or a tuple per cache entry (async / sync). unless you have a better suggestion?

src/config/resolve-config-editorconfig.js
```diff
@@ -3,11 +3,9 @@
 const path = require("path");
 
 const editorconfig = require("editorconfig");
-const mem = require("mem");
 const editorConfigToPrettier = require("editorconfig-to-prettier");
 const findProjectRoot = require("./find-project-root.js");
-
```
brodybits, line 9: nit: why remove the blank line here?

I would generally keep a blank line between a require and any other code.

43081j: think it was just lost during a few different attempts at solving the problem, happy to put it back

src/config/resolve-config-editorconfig.js
brodybits: > yes, which means two caches in this case (we're keying by the same thing)

Wouldn't the `sync` flag be part of `opts` which we would be keying by in the cache, or am I mistaken here?

brodybits:
Minor nit: shouldn't this be a "refactor" commit?

+1 (+100) for the conventional commit message, which I have found to be extremely helpful in all kinds of projects.

fisker:
To reduce dependencies, I send  one PR to remove `escape-string-regexp` one year ago, this module only used one line (there are two places last year). Didn't get approved yet... https://github.com/prettier/prettier/pull/8411


To reduce the bundle size , there are `parser-flow.js` and `parser-typescript.js` they are ~3M minified, I'm sure optimize them are more effective.

I really don't see a point to doing this.

43081j:
The point of doing this is:

* It is right here, near enough ready to go
* It will reduce CI time (installs)
* It will reduce build output

Why would anyone turn down reduced package size and reduced build times? As of yet, nobody has presented a good reason for not doing it, its all positives from here on as far as i can tell. 

I'm sure it isn't the only unnecessary dependency we've got, a bit at a time is how you make progress.

fisker:
> It is right here, near enough ready to go

Don't you think it's obviously wrong to use `WeakMap`? The `opts` is not a reference most time. We use `JSON.stringify` as key for a reason. This and pointing out "should cache the promise" are  the main reasons I don't want do this, we are creating problems instead of solving problems. Anyway, once these get resolved I guess I have no reason to object.

> It will reduce CI time (installs)

As the one setup the CI, I even didn't bother to cache `node_module` dir, not because I don't know how to cache it, compare to 1 hour CI running time, install time (couple mins) can be ignored.

43081j:
every little helps, even a small increase in build speed is better than none, especially if you keep improving it over time.

i wasn't aware of `opts` not being a reference, it probably should be but its easy enough to change it to be keyed by the JSON instead 👍

43081j:
both those are resolved now, thanks for the help too, much appreciated.

it may seem like a small improvement in the grand scheme of things but every improvement helps. we're saving ourselves a hundred or so lines of code and this is only one package, there will be other unnecessary ones too.

43081j:
@fisker can you re-review when you get chance? also need someone to approve the CI build so it'll run i think?

43081j:
FYI CI fails because i used a null coalesce but CI uses node 12 which doesn't understand that syntax presumably.

ill update to not use that syntax and try again

43081j:
have replaced the null coalesce, CI should pass if one of you can approve the run again 👍

43081j:
@fisker could you trigger a ci run?

src/common/load-plugins.js
```diff
@@ -5,31 +5,41 @@ const path = require("path");
...
 
-const memoizedLoad = mem(load, { cacheKey: JSON.stringify });
-const memoizedSearch = mem(findPluginsInNodeModules);
 const clearCache = () => {
-  mem.clear(memoizedLoad);
-  mem.clear(memoizedSearch);
+  nodeModulesPluginCache.clear();
+  loadCache = new WeakMap();
 };
+const defaultPluginSearchDirs = [];
```
fisker, line 20: This change looks unsafe to me.

fisker:
I'm not going to review this PR again, but I won't block this PR either.

@alexander-akait seems interest replacing this package, I'll leave the choice to him.

src/common/load-plugins.js
43081j: this is so if we don't pass any plugins, we re-use the same empty array rather than creating a new one each time (which helps keep the caching a lot cleaner). its pretty safe in the current code because we never mutate it, but there is risk someone unaware of this might later mutate it without understanding the side effects.

i could leave a comment to help people understand not to do that

43081j:
no worries, thanks for putting the time in so far, much appreciated.

seems CI is happy too 🥳

43081j:
@alexander-akait mind taking another look at this? would be good to get it merged

alexander-akait (changes requested):
We need tests

43081j:
sure ill figure them out and get some added 👍  was relying on any existing ones so far since it should behave the same from outside

43081j:
sorry this went a little inactive, ill rebase onto master.

we don't seem to have any tests around this directly. it would be super helpful if someone could point me in the right direction for adding some...

i can have another dig through the test files today but iirc i got a bit lost last time

PR #11243, opened 2021-07-23
Use setup-node cache yarn dependencies

fisker:
## Description

To cache dependencies, it's much simpler now. The installation step is actually fast, normally just seconds, bring this up because mentioned in  https://github.com/prettier/prettier/pull/11241 CI installation can be faster.

## Checklist


fisker:
Without cache https://github.com/prettier/prettier/pull/11243/checks?check_run_id=3139724322#step:3:11
With cache https://github.com/prettier/prettier/pull/11243/checks?check_run_id=3139741098#step:3:15

sosukesuzuki (approved):

PR #11248, opened 2021-07-23
Improve plugin resolution (fixes pnpm compatibility)

adamhl8:
## Description

Fixes https://github.com/prettier/prettier/issues/8056

**== See this reply for an explanation of the updated fix: https://github.com/prettier/prettier/pull/11248#issuecomment-886130669 ==**

~~Previously, plugins were loaded from the `node_modules` directory that was parent to the current running instance of Prettier. Loading relative to the cwd ensures that the right `node_modules` is used (for example, when using Prettier with pnpm).~~

~~I added [pkg-dir](https://github.com/sindresorhus/pkg-dir) to handle finding the project directory. I initially wrote my own way to find it, but I figured using this package is probably better and safer than anything I could come up with. This package simply starts from the cwd and traverses up until it finds a path where `package.json` exists. Hopefully this wouldn't break plugin loading for anyone. I can't think of any project setups where this would.~~

~~I've tested this with my own projects that use pnpm and Prettier loads all my plugins without issue.~~

## Checklist

I could be completely wrong, but I don't think additional tests need to be written for this change. All current tests pass. `cwd` will always be a path in the actual project that is being formatted, so the right `node_modules` will be found.

I can update the docs if needed, but it seems like the way `plugins.md` is currently worded still makes sense.

```
Providing at least one path to `--plugin-search-dir`/`pluginSearchDirs`
turns off plugin autoloading in the default directory (i.e. `node_modules` above `prettier` binary).
```

The default directory is still the `node_modules` above the `prettier` binary, we're just getting there a different way.

### DIFF
--- a/changelog_unreleased/css/13577.md
+++ b/changelog_unreleased/css/13577.md
@@ -1,0 +1,32 @@
+#### Fix formatting of long `:is`, `:where`, and `:not` selectors (#13577 by @j-f1)
+
+Pseudo-selectors like `:is`, `:where`, and `:not` that can take multiple selectors as arguments are now formatted like function calls are in other languages. Previously, no special significance was attached to the commas between their “arguments,” leading to confusing wrapping behavior. There are likely still improvements to be made here — please open an issue with some example code if you find something that doesn’t look as expected.
+
+<!-- prettier-ignore -->
+```css
+/* Input */
+:where(
+  label > input:valid,
+  label > textarea:not(:empty),
+  label > button[disabled]
+) ~ .errors > .error { display: none; }
+
+/* Prettier stable */
+:where(label > input:valid, label > textarea:not(:empty), label
+    > button[disabled])
+  ~ .errors
+  > .error {
+  display: none;
+}
+
+/* Prettier main */
+:where(
+    label > input:valid,
+    label > textarea:not(:empty),
+    label > button[disabled]
+  )
+  ~ .errors
+  > .error {
+  display: none;
+}
+```
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -461,7 +461,12 @@ function genericPrint(path, options, print) {
       return [
         maybeToLowerCase(node.value),
         isNonEmptyArray(node.nodes)
-          ? ["(", join(", ", path.map(print, "nodes")), ")"]
+          ? group([
+              "(",
+              indent([softline, join([",", line], path.map(print, "nodes"))]),
+              softline,
+              ")",
+            ])
           : "",
       ];
     }
--- a/tests/format/css/indent/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/css/indent/__snapshots__/jsfmt.spec.js.snap
@@ -42,7 +42,9 @@ a {
 
 =====================================output=====================================
 a {
-  ~ .Pagination-itemWrapper:not(.is-separator):not([data-priority^="#{$priority}"])
+  ~ .Pagination-itemWrapper:not(.is-separator):not(
+      [data-priority^="#{$priority}"]
+    )
     ~ .Pagination-itemWrapper.is-separator[data-priority^="#{$priority}"] {
     display: flex;
   }
--- a/tests/format/css/pseudo-call/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/css/pseudo-call/__snapshots__/jsfmt.spec.js.snap
@@ -1,5 +1,86 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`is.css format 1`] = `
+====================================options=====================================
+parsers: ["css"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+:is(ol, ul, menu:unsupported) :is(ol, ul) {
+    color: green;
+}
+
+:is(ol, ul) :is(ol, ul) ol {
+    list-style-type: lower-greek;
+    color: chocolate;
+}
+
+:is(ol, ul, menu, dir) :is(ol, ul, menu, dir) :is(ul, menu, dir) {
+  list-style-type: square;
+}
+
+/* Level 0 */
+h1 {
+  font-size: 30px;
+}
+/* Level 1 */
+:is(section, article, aside, nav) h1 {
+  font-size: 25px;
+}
+/* Level 2 */
+:is(section, article, aside, nav) :is(section, article, aside, nav) h1 {
+  font-size: 20px;
+}
+/* Level 3 */
+:is(section, article, aside, nav)  :is(section, article, aside, nav)  :is(section, article, aside, nav)  h1 {
+  font-size: 15px;
+}
+
+some-element:is(::before, ::after) {
+  display: block;
+}
+
+=====================================output=====================================
+:is(ol, ul, menu:unsupported) :is(ol, ul) {
+  color: green;
+}
+
+:is(ol, ul) :is(ol, ul) ol {
+  list-style-type: lower-greek;
+  color: chocolate;
+}
+
+:is(ol, ul, menu, dir) :is(ol, ul, menu, dir) :is(ul, menu, dir) {
+  list-style-type: square;
+}
+
+/* Level 0 */
+h1 {
+  font-size: 30px;
+}
+/* Level 1 */
+:is(section, article, aside, nav) h1 {
+  font-size: 25px;
+}
+/* Level 2 */
+:is(section, article, aside, nav) :is(section, article, aside, nav) h1 {
+  font-size: 20px;
+}
+/* Level 3 */
+:is(section, article, aside, nav)
+  :is(section, article, aside, nav)
+  :is(section, article, aside, nav)
+  h1 {
+  font-size: 15px;
+}
+
+some-element:is(::before, ::after) {
+  display: block;
+}
+
+================================================================================
+`;
+
 exports[`pseudo_call.css format 1`] = `
 ====================================options=====================================
 parsers: ["css"]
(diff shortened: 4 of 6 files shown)


fisker:
Thanks for working on this.

1. We search project root differently in other place, see [this](https://github.com/prettier/prettier/blob/main/src/config/find-project-root.js)
2. If we decide to change search logic, what do you think just search from `process.cwd()`?
3. Big problem, we still support global installation, this solution will break that, I'm not sure what's right to do.

fisker:
> This package simply starts from the cwd and traverses up until it finds a path where package.json exists. Hopefully this wouldn't break plugin loading for anyone. I can't think of any project setups where this would.

This will break monorepo.

adamhl8:
I've come up with a solution that fixes compatibility with pnpm and shouldn't (hopefully) break any project setups (monorepo, global install, etc). It also doesn't care about what the virtual store is called or where it is.

I added tests and updated the plugins documentation page as well. I'm not very familiar with jest so I'd appreciate if someone could check those over.

There are comments in the code that hopefully makes it clear, but I'll somewhat briefly explain here:

Essentially, instead of trying to find pnpm's `node_modules` directory (which contains the symlinks), we just use the virtual store directly since this is where Prettier is being run from. (The virtual store contains hardlinks to the *actual* packages but that's irrelevant in this case.)

All packages in the virtual store have a directory structure that looks like this:

```
.pnpm/ (virtual store)
├── <package>@<version>/ (e.g. prettier@2.3.2)
│   └── node_modules/ (contains package + its dependencies)
│       ├── prettier/ (dir where prettier is run from)
│       │   ├── index.js
│       │   └── ...
│       ├── dependency1/
│       └── ...
├── prettier-plugin-foo@1.2.3
│   └── node_modules
│       └── prettier-plugin-foo
└── ...
```

All packages in the virtual store are appended with `@<version>`. We can check for that to determine if we're in the virtual store, move up into the parent directory (virtual store), and get the path for each plugin directory. Because each plugin directory will have its own `node_modules`, we can simply pass each path to the existing plugin resolution function.

Additionally, you can now pass `--plugin-search-dir` any arbitrary directory that contains Prettier plugins. This shouldn't affect anyone already loading plugins this way. i.e. you can pass `node_modules` directly or a directory that is the direct parent to plugins:

```
somePluginsDir/ (or node_modules)
├── prettier-plugin-foo/
└── prettier-plugin-bar/
```

adamhl8:
@fisker Any thoughts on this?

malekjaroslav:
@fisker @adamhl8 Any updates on this, it's a very annoying bug not being able to use Prettier plugins with pnpm

adamhl8:
> It seems there are lots of unrelated changes.

What do you mean?

adamhl8:
> It seems there are lots of unrelated changes.

Sorry for the slow response.
Unrelated to what? I feel like all the changes are within the "scope" of this PR. I realize that "Improve plugin resolution" is a bit broad, but that's what this PR aims to do. And a side effect of that being it fixes plugin compatibility with pnpm.

rburgst:
any updates on this? I would love to be able to use pnpm

Shinigami92:
> any updates on this? I would love to be able to use pnpm

What's stopping you?
Just configure it like in this project: https://github.com/prettier/plugin-pug/blob/3c6dd7222b65fea5c2ba69ddbb00dbed56d70b86/.prettierrc.cjs#L8

ZeroAurora:
> > any updates on this? I would love to be able to use pnpm
> 
> What's stopping you? Just configure it like in this project: https://github.com/prettier/plugin-pug/blob/3c6dd7222b65fea5c2ba69ddbb00dbed56d70b86/.prettierrc.cjs#L8

But it's something that should work out of box instead of manual configuring. Not to mention those who prefer to use JSON as the configuration format (like me).

TicTak21:
Any updates?

adamhl8:
I'd appreciate if I could get some more detailed feedback on what is preventing this PR from moving forward. @fisker

fisker:
Can you explain to me how this change https://github.com/prettier/prettier/pull/11248/files#r724610453 is related?

adamhl8:
I updated/cleaned up the plugins documentation to reflect the [changes](https://github.com/prettier/prettier/pull/11248#issuecomment-886130669) I made. I'm honestly not trying to be difficult; I don't understand how any of these changes are unrelated. I can remove the changes I made to the docs, but then the docs wouldn't accurately reflect what you can do with the plugin CLI flags or the config file.

docs/plugins.md
```diff
@@ -3,35 +3,44 @@ id: plugins
 title: Plugins
 ---
 
-Plugins are ways of adding new languages to Prettier. Prettier’s own implementations of all languages are expressed using the plugin API. The core `prettier` package contains JavaScript and other web-focused languages built in. For additional languages you’ll need to install a plugin.
+Plugins are ways of adding new languages or formatting rules to Prettier. Prettier’s own implementations of all languages are expressed using the plugin API. The core `prettier` package contains JavaScript and other web-focused languages built in. For additional languages you’ll need to install a plugin.
 
 ## Using Plugins
 
-Plugins are automatically loaded if you have them installed in the same `node_modules` directory where `prettier` is located. Plugin package names must start with `@prettier/plugin-` or `prettier-plugin-` or `@<scope>/prettier-plugin-` to be registered.
+Plugins are automatically loaded if you have them installed in the same `node_modules` directory where `prettier` is located. Plugin package names must start with `prettier-plugin-` or `@prettier/plugin-` or `@<scope>/prettier-plugin-` to be registered.
```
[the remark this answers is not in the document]
adamhl8: This is just a change of the ordering of the text to make things a bit more consistent throughout the docs.

docs/plugins.md
```diff
@@ -3,35 +3,44 @@ id: plugins
...
 
-- Or the [API](api.md), via the `plugins` and `pluginSearchDirs` options:
+- The [API](api.md), via the `plugins` and `pluginSearchDirs` options:
 
   ```js
   prettier.format("code", {
     parser: "foo",
-    pluginSearchDirs: ["./dir-with-plugins"],
     plugins: ["./foo-plugin"],
+    pluginSearchDirs: ["./dir-with-plugins"],
```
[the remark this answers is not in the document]
adamhl8: This is just a change of the ordering of the text to make things a bit more consistent throughout the docs.

docs/plugins.md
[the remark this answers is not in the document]
fisker: Do this in a seperate PR.

adamhl8:
Removed changes to the docs as requested and rebased on main.

src/common/load-plugins.js
```diff
@@ -25,11 +26,48 @@ function load(plugins, pluginSearchDirs) {
...
+      ? path.resolve(autoLoadDirectory)
+      : ""; // Make sure autoLoadDir isn't null.
+
+    const isProjectRoot = findProjectRoot(__dirname) === autoLoadDirectory; // Make sure we don't go outside projectRoot.
+
+    // Get the parent directory of autoLoadDir.
+    // It will be the pnpm virtual store if we're not at the project root and if "prettier@" is found.
+    const pnpmVirtualStore =
+      !isProjectRoot &&
+      autoLoadDirectory.includes("prettier@") &&
```
fisker, line 58: `path.dirname(...).startsWith("prettier@")`

src/common/load-plugins.js
```diff
@@ -25,11 +26,48 @@ function load(plugins, pluginSearchDirs) {
...
-    const autoLoadDir = thirdParty.findParentDir(__dirname, "node_modules");
-    if (autoLoadDir) {
-      pluginSearchDirs = [autoLoadDir];
+    // This is likely the project root. In projects using pnpm, this directory is named "prettier@<version>".
+    let autoLoadDirectory = thirdParty.findParentDir(__dirname, "node_modules");
+    autoLoadDirectory = autoLoadDirectory
+      ? path.resolve(autoLoadDirectory)
+      : ""; // Make sure autoLoadDir isn't null.
+
+    const isProjectRoot = findProjectRoot(__dirname) === autoLoadDirectory; // Make sure we don't go outside projectRoot.
```
fisker, line 52: Though it's not recommend, but we support global installation, I think editors use global one when Prettier is not installed locally.

src/common/load-plugins.js
adamhl8: @fisker `pnpmVirtualStore` gets set to the path returned by `path.dirname()` (assuming the first two checks pass). It's not supposed to be `true`. I might be misunderstanding your comment though.

I'll change `autoLoadDirectory.includes("prettier@")` to `autoLoadDirectory.startWith("prettier@")`.

src/common/load-plugins.js
adamhl8: That shouldn't affect global install. `isProjectRoot` is only used to check if we're in the pnpm virtual store (line 57).

src/common/load-plugins.js
fisker: Isn't it a absolute path? Are we testing the dir name is `pretty@2.5.1`?

adamhl8: When using pnpm, the parent directory of `autoLoadDirectory` is the pnpm virtual store. We check to see if `autoLoadDirectory` starts with `prettier@` so we know that pnpm is being used. Then we grab the parent directory with `path.dirname`.

fisker: I understand that. `autoLoadDirectory` is a absolute path, `.startsWith()` will never be true, you need get the dirname.

fisker: We can know the version of Prettier by reading `package.json` so we can test the dirname is exactly `prettier@<version>`

adamhl8: Ah, gotcha, sorry. Pushed a commit that should fix that. Would it not be better/less fragile to just check for `startsWith("prettier@")`. (For example, if pnpm changes how it appends the version number.)

adamhl8:
@fisker Was hoping you could take a look at this again. I fixed the merge conflicts and all tests pass.

adamhl8:
Added proper tests to make codecov happy. Are there any issues that are preventing this from being merged? @fisker

fisker:
Sorry for the delay, I haven't take a deep look. But the plugin related part already changed in `next` branch. Would like to change target base to `next`?

adamhl8:
@fisker I closed this by mistake. Are you able to reopen it and change the base to `next`?

I updated my changes/tests to work with `next`.

adamhl8:
Ah weird. New PR is here: #13583

PR #11313, opened 2021-08-06
Disable comment attach in babel parser

fisker:
## Description

Ref https://github.com/babel/babel/pull/13229

## Checklist


fisker:
This seems not easy to disable, we use `leadingComments` in parser's postprocess, I tried to move it into printer's preprocess, then it breaks comment attach... Maybe take a deep look another time.

Issue #6885, opened 2019-11-08
Show Detailed Infos on CI linke circleCI


muescha:
more people submit PR for documents in MD files or other things via Github web editor. if they make a mistake then the CI fails but there is no change to inform the user whats wrong. 

i see some infos should be with the `--debug-check` but currently there is no output with this option.

**Environments:**
- Prettier Version: 1.18.2
- Running Prettier via: cli
- Runtime: node?
- Operating System: linux

**Steps to reproduce:**

prettier "**/*.{md,css,scss,yaml,yml,ts}" "--check"

**Expected behavior:**

show each error and the changes like with --debug-check here in #4415 

**Actual behavior:**

show nothing about the errors:

```
> @ prettier /home/circleci/project
> prettier "**/*.{md,css,scss,yaml,yml,ts}" "--check"

Checking formatting...
docs/contributing/environment.md
Code style issues found in the above file(s). Forgot to run Prettier?
```

also this are no infos:

prettier "**/*.{md,css,scss,yaml,yml,ts}" "--debug-check"
prettier "**/*.{md,css,scss,yaml,yml,ts}" "--debug-check" "--check"

dylannz-sailthru:
"Forgot to run Prettier?" is really not a very helpful message...

thorn0:
The message contains the file names. What else do you need to run Prettier on those files? ¯\_(ツ)_/¯

lydell:
I personally like the output of `--check` very much.

_CI:_ "Forgot to run Prettier?"

_Me:_ "Oh yeah, I guess I did!" _Runs Prettier._

_CI:_ "All good!"

muescha:
I like to point again to:

> more people submit PR for documents in MD files or other things via Github web editor.

That means they have no local IDE installed. That's why they *can not* run prettier on the file.

muescha:
Expected behavior would be a detailed error message or a diff 

```
[error] src/index.js: ast(input) !== ast(prettier(input))
[error] Index:
[error] ===================================================================
[error] ---
[error] +++
[error] @@ -17,6 +17,6 @@
[error]                    "method": false,
[error]                    "key": {
[error] -                    "type": "StringLiteral",
[error] -                    "value": "asd"
[error] +                    "type": "Identifier",
[error] +                    "name": "asd"
[error]                    },
[error]                    "computed": false,
[error]
[error] Index:
[error] ===================================================================
[error] ---
[error] +++
[error] @@ -1,1 +1,1 @@
[...]

PR #11369, opened 2021-08-20
Style: updated failure message to be more formal

webark:
## Description

The term "Forgot to run Prettier?" can be offputting for some (as documented in #6885). I believe that informing what needs to be done, rather than assuming someone did something "wrong", no matter how fine that "wrong" thing is, will provide a more even and calming message.

I debated adding more information (like a link to the logs or an example command to run) but I thought that was a separate initiative that could be picked up regardless of this change.

All I was after was attempting to remove some language that some find offputting if not condescending. 

Thanks!

## Checklist

### DIFF
--- a/changelog_unreleased/cli/11369.md
+++ b/changelog_unreleased/cli/11369.md
@@ -1,0 +1,7 @@
+#### Updated failure message to be more informative (#11369 by @webark)
+
+Updated the "Forgot to run Prettier?" to "Run Prettier to fix."
+
+This keeps the same spirit of the message, but is less likely to be
+missinterpretted as it's a more formal message rather than being
+somewhat familial.
--- a/changelog_unreleased/cli/11369.md
+++ b/changelog_unreleased/cli/11369.md
@@ -3,5 +3,5 @@
 Updated the "Forgot to run Prettier?" to "Run Prettier to fix."
 
 This keeps the same spirit of the message, but is less likely to be
-missinterpretted as it's a more formal message rather than being
+misinterpreted as it's a more formal message rather than being
 somewhat familial.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -69,7 +69,7 @@ Console output if some of the files require re-formatting:
 Checking formatting...
 [warn] src/fileA.js
 [warn] src/fileB.js
-[warn] Code style issues found in 2 files. Forgot to run Prettier?
+[warn] Code style issues found in 2 files. Run Prettier to fix.
 ```
 
 The command will return exit code `1` in the second case, which is helpful inside the CI pipelines.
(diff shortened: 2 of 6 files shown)


lydell:
@kachkaev What do you think about this change?

kachkaev:
Thanks for the PR @webark! _‘Forgot to run Prettier?’_ originates from #5629, which also contains the discussion of the phrasing. I wish we could be specific about what command to run as a fix, but it does not look possible given a variety of setup options.

I’m not a native speaker so can’t tell if the message is offputting or not – I trust you that it can be. Sorry for missing this detail when implementing the `--check` flag!

I’m all for the change if it helps improve tone and make the tool more welcoming to people. I wonder if we can remove the second sentence completely because we can’t really tell what _exactly_ to do:

```diff
  Checking formatting...
  [warn] src/fileA.js
  [warn] src/fileB.js
- [warn] Code style issues found in the above file(s). Forgot to run Prettier?
+ [warn] Code style issues found in the above file(s).
```

There’s one thing to keep in mind though: by changing the phrasing we can break someone’s dev flows such as CI jobs. A [search for ‘forgot to run prettier’ on GitHub](https://github.com/search?q="forgot+to+run+prettier"&type=code) suggests that people might use the phrase as a trigger of some sort. Perhaps, we can implement the change [in 3.0](https://github.com/prettier/prettier/milestone/14) for some extra safety?

lydell (approved):
I went through the search results @kachkaev posted. Most of them are:

- Snapshot tests. Should be easy to update for folks!
- Wrapper tools around Prettier that print a similar message.

But [two results](https://github.com/search?l=Shell&q=%22forgot+to+run+prettier%22&type=Code) are shell scripts that grep for that exact output. That’s bad. This output is for human and was never intended to be relied on – people should use the exit code.

It’s hard to say how many private scripts rely on the exact wording. We _could_ say that people have been relying on an implementation details, but it might be safer to wait for 3.0.

Either way, I like the wording change in this PR so I’m adding a :heavy_check_mark:

sosukesuzuki (approved):
I agree with @lydell. I think we should switch base branch of this PR to `next` branch.

thorn0 (approved):

changelog_unreleased/cli/11369.md
```diff
@@ -0,0 +1,7 @@
+#### Updated failure message to be more informative (#11369 by @webark)
+
+Updated the "Forgot to run Prettier?" to "Run Prettier to fix."
+
+This keeps the same spirit of the message, but is less likely to be
+misinterpreted as it's a more formal message rather than being
+somewhat familial.
```
fisker, line 7: I forgot how we wrap this normally, should the on one line?

fisker (approved):
One comment.

changelog_unreleased/cli/11369.md
thorn0: If we wanted to enforce wrapping, we'd have `proseWrap: always` in `.prettierrc`.

PR #11401, opened 2021-08-26
Prettier should not re-escape a backslash in a string literal

kiwiupover:
## Description
Formatting handlebars prettier is escaping a backslash with another backslash.

Currently this PR only adds a breaking test. 
 
For this example
```hbs
<DateFormat @pathFormat="dateFormat:YYYY-MM-DD hh\:mm\:ss" />
```

Current output
```hbs
<DateFormat @pathFormat="dateFormat:YYYY-MM-DD hh\\:mm\\:ss" />
```

Expected output
```hbs
<DateFormat @pathFormat="dateFormat:YYYY-MM-DD hh\:mm\:ss" />
```

## Checklist

### DIFF
--- a/tests/format/handlebars/escape/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/handlebars/escape/__snapshots__/jsfmt.spec.js.snap
@@ -1,5 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`backslash.hbs format 1`] = `
+====================================options=====================================
+parsers: ["glimmer"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+<DateFormat @pathFormat="dateFormat:YYYY-MM-DD hh\:mm\:ss" />
+
+
+=====================================output=====================================
+<DateFormat @pathFormat="dateFormat:YYYY-MM-DD hh\:mm\:ss" />
+================================================================================
+`;
+
 exports[`html-entities.hbs format 1`] = `
 ====================================options=====================================
 parsers: ["glimmer"]
--- a/tests/format/handlebars/escape/backslash.hbs
+++ b/tests/format/handlebars/escape/backslash.hbs
@@ -1,0 +1,2 @@
+<DateFormat @pathFormat="dateFormat:YYYY-MM-DD hh\:mm\:ss" />
+


fisker:
Can you open an issue instead?

kiwiupover:
I have tried to fix it. But I was in able to fix it. 

@fisker Is it worth adding a failing test and a issue.

fisker:
Have you try unescape [here](https://github.com/prettier/prettier/blob/2380e113624300bfee65cc9ea1cffae44e8208be/src/language-handlebars/printer-glimmer.js#L194)?

fisker:
Or maybe it's bug in the glimmer parser?

jelhan:
This function seems to be responsible for duplicating the first backslash (`\`): https://github.com/prettier/prettier/blob/c9c6d3364f8cf5140c3c16b44b92943464be7873/src/language-handlebars/parser-glimmer.js#L7-L20

`node.chars` is as expected before but have the duplicated slash afterwards. Not a big surprise if looking at that code.

It is run as a plugin for Glimmer parser. https://github.com/prettier/prettier/blob/c9c6d3364f8cf5140c3c16b44b92943464be7873/src/language-handlebars/parser-glimmer.js#L43-L46

If I remove that plugin for the Glimmer parser the following tests start to fail:

```
 FAIL  tests/format/handlebars/mustache-statement/jsfmt.spec.js
  ● escaped.hbs › format

    expect(received).toMatchSnapshot()

    Snapshot name: `escaped.hbs format 1`

    - Snapshot  - 2
    + Received  + 2

    @@ -11,8 +11,8 @@
      \\\{{my-component}}

      an escaped mustache:
      \{{my-component}} a non-escaped mustache:
    - \\{{my-component}}
    - another non-escaped mustache: \\\{{my-component}}
    + \{{my-component}}
    + another non-escaped mustache: \\{{my-component}}
      ================================================================================

      283 |         CURSOR_PLACEHOLDER,
      284 |       })
    > 285 |     ).toMatchSnapshot();
          |       ^
      286 |   });
      287 |
      288 |   if (!FULL_TEST) {

      at Object.<anonymous> (tests/config/format-test.js:285:7)

  ● escaped.hbs - {"singleQuote":true} › format

    expect(received).toMatchSnapshot()

    Snapshot name: `escaped.hbs - {"singleQuote":true} format 1`

    - Snapshot  - 2
    + Received  + 2

    @@ -12,8 +12,8 @@
      \\\{{my-component}}

      an escaped mustache:
      \{{my-component}} a non-escaped mustache:
    - \\{{my-component}}
    - another non-escaped mustache: \\\{{my-component}}
    + \{{my-component}}
    + another non-escaped mustache: \\{{my-component}}
      ================================================================================

      283 |         CURSOR_PLACEHOLDER,
      284 |       })
    > 285 |     ).toMatchSnapshot();
          |       ^
      286 |   });
      287 |
      288 |   if (!FULL_TEST) {

      at Object.<anonymous> (tests/config/format-test.js:285:7)

 › 2 snapshots failed.
```

So it seems to be still needed to work-a-round another bug in the Glimmer parser.

To be honest until today I have never seen that escape syntax. Not sure if it is even officially supported by Glimmer at all.

Issue #4547, opened 2018-05-25
Ignore commonly generated folders


astorije:
At the moment, Prettier ignores `node_modules`. In my experience, folders `coverage` and `dist` are commonly used to contain generated artifacts and could be ignored accordingly, as a sensible default.

If someone _really_ needs to include those, adding `!coverage` and/or `!dist` in `.prettierignore` should re-include them.

Thoughts? :)

BuckyMaler:
I think all entries in `.gitignore` should be ignored by Prettier without the need to specify `--ignore-path .gitignore`. The `.gitignore` should be a constant with the option to additionally use a `.prettierignore`. I think this is sensible because I cannot think of a reason as to why a file that is not version controlled would need to be formatted. With this, many projects may not even need to add a `.prettierignore`. A `.prettierignore` is particularly useful for ignoring files that are version controlled because that cannot be done with a `.gitignore`.

danihodovic:
Any updates on this?

evilebottnawi:
We can implement:
```
--ignore-pattern [String]      Pattern of files to ignore (in addition to those in .prettierignore)
```

eslint has this flag too

danihodovic:
Is it possible to specify this in .prettierrc.yml instead of passing it as a CLI option every time?

lydell:
@danihodovic Sounds like you’d be interested in https://github.com/prettier/prettier/issues/7073

lydell:
Similar issue: #4081

jdufresne:
> I think all entries in .gitignore should be ignored by Prettier without the need to specify --ignore-path .gitignore. 

Came here to make the same suggestion. I think that would reduce the need to configure prettier in the common case.

Other opinionated tools, such as [black](https://github.com/psf/black/commit/e9d4e7b67f359114ba57ecf0c1300f7eb4f08950), started doing this and it has worked great.

alexilyaev:
I think it's safe to say that eventually Prettier will face the same requests from the community as ESLint did.
So I suggest following ESLint's ignore features:
https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories

### Specifically:
- Supporting [--ignore-pattern](https://eslint.org/docs/user-guide/command-line-interface#ignore-pattern) CLI command (can be used multiple times) to add custom ignore patterns on top of the ones from `--ignore-path`.
  - This is useful for running Prettier without needing to add config files.
[...]

PR #11435, opened 2021-09-04
Add file-list option to CLI

backbone87:
## Description
this option allows to specify one or more commands to obtain file lists from. only these listed files may then be
formatted.

this is useful for example to only format/check files in
git, e.g.: `prettier --file-list 'git ls-files' ./**/*.*`

fixes https://github.com/prettier/prettier/issues/4547

## Checklist

### DIFF
--- a/src/cli/constant.js
+++ b/src/cli/constant.js
@@ -166,6 +166,16 @@ const options = {
       * inferredParser (string | null) - name of parser inferred from file path
     `,
   },
+  "file-list": {
+    type: "string",
+    category: coreOptions.CATEGORY_CONFIG,
+    description: outdent`TBD`,
+    array: true,
+  },
+  git: {
+    type: "boolean",
+    description: outdent`TBD`,
+  },
   help: {
     type: "flag",
     alias: "h",
--- a/src/cli/format.js
+++ b/src/cli/format.js
@@ -1,6 +1,8 @@
 "use strict";
 
 const { promises: fs } = require("fs");
+const os = require("os");
+const { execSync } = require("child_process");
 const path = require("path");
 
 const chalk = require("chalk");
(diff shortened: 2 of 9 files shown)


sosukesuzuki (commented):
Thanks for this PR but I don't think this is a good way. I prefer https://github.com/prettier/prettier/issues/4547#issuecomment-628637558 as solution for #4547

backbone87:
What's wrong with this feature? it solves a real problem ("run prettier only for version-controlled files") and is minimal code. https://github.com/prettier/prettier/issues/4547#issuecomment-628637558 is still possible to do.

kachkaev:
I haven’t seen other CLI tools using this pattern. @backbone87 do you have any examples to share?

backbone87:
I don't have another example. But I wish eslint would have the same feature. Reimplementing the whole git ignore algorithm with global, user and project local ignore seems far too complex. This approach should also work with almost any VCS. It is also very lightweight and completely opt-in.

kachkaev:
If your main goal is to fully match the list of git-tracked files, how about making `.prettierignore` a symlink of `.gitignore`?

Also note that by using `git ls-files` in your example, Prettier will omit untracked files, while still try to format just-deleted files and fail. That does not look like the behaviour you look for 🤔

backbone87:
ignore-path is not enough. There are things like nested gitignore and global/user/repo local ignores that all get respected by git but not prettier (or eslint for that matter).

As implemented file-list is not the source of files to be formatted but just another filter so deleted files are not a problem. Also you can specify file-list multiple times to add both `git ls-files` and `git ls-files --others --exclude-standard`.

backbone87:
any reconsiderations here?

PR #11485, opened 2021-09-09
Add alignColumns

### LINKED ISSUE #6767: Table formatting assumes a monospace font
**Prettier 1.18.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBlAnrAhgDwAQA+eAInAM5gBOAlgA4zXQA6UAtOwe6ywBJxYATOJUJ4AKtRgAbOCwAKWSlgDmS2gAtRYuDhggANCAj1GUMslCLKEAO4LKCcyiwA3CNQEGQAIyVgA1nAwqLRYYNRQysgwlACucIbqMAC2UgDq6pLkoWBwqI6S1C6S6MjgZOaGEWTCMHJKyslYyABmWFI1hgBWZDgAQn6BwVjJcAAyEXCt7Z0gPTioEcoyAIqxEPDTHQkgoZQ1lGVNlP4CtlBetDSwaR4w6sgAHAAMhlcQNWlqZVfkwi5TQwAR3W8HqxicICwZFYUDgcCEnkMDhB1Ac9RUTS2sxqyWo0TiOzIS1WoKmSDa20MMCw3luAnuyAATNSlNQpEsAMIQZJYlBQaCAkCxGpiWlOSk1AC+UqAA)
```sh
--parser markdown
```

**Input:**
```markdown
Syntax | Description
---|---
Header | Title
Paragraph | Text
```

**Output:**
```markdown
| Syntax    | Description |
| --------- | ----------- |
| Header    | Title       |
| Paragraph | Text        |
```
This looks fine with a monospace font, but not a proportional font:
> \| Syntax    \| Description \|
> \| --------- \| ----------- \|
> \| Header    \| Title       \|
> \| Paragraph \| Text        \|

**Expected behavior:**
Should be the same as the input in this case to prevent issues with misalignment when using proportional fonts (non-monospace fonts where characters are different width, and therefore assumptions cannot be made about code alignment). If spaces are missing next to pipes (`|`), they should still be added, as Prettier does currently. Columns should not be aligned.

**Explanation:**
Besides making tables less visually appealing in proportional fonts, this issue makes tables significantly more difficult to edit if they're wider than the editor. For example, let's say you're editing a table with some very long content cells, but very short heading cells. Using Prettier's formatting with a monospace font is fine because you can always scroll up to see the current column's heading aligned above. However, when using a proportional font, the heading row will often be invisible because it doesn't align with the current content cell, and you won't even know whether to scroll left to right as it depends on the width of each row. Currently the workaround is to disable Prettier or change your font, which I think is too limiting.

Overall, I feel that Prettier formatting should not cause code to be less readable in any situation, even if it means not making it more readable in some situations.

Austaras:
## Description
closes #6767 

## Checklist


kachkaev:
This PR might be hard to get this PR through because new options are not in line with [Prettier’s Option Philosophy](https://prettier.io/docs/en/option-philosophy.html) 😔

Austaras:
> This PR might be hard to get this PR through because new options are not in line with [Prettier’s Option Philosophy](https://prettier.io/docs/en/option-philosophy.html) pensive

You could see previous discussion in related issue

nickmccurdy:
If we can't add an option for this I would prefer to disable alignment by default personally. However most users seem to use monospace fonts so this could be a good exception.

Austaras:
I have rebased this pr, anything I could do to make any progression since #6644 is merged?

Also I don't believe make it a default behavior will be a good thing -- it will benefit only a small potion of people, but will change lots of existing code.

Issue #10419, opened 2021-02-27
Glimmer: Double quotes removed even if single quotes inside


bmaehr:
**Prettier 2.2.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIaYwA2cAvADojDAy4DkREA5plAHQCuAznAE5QBDALZw26OFzC9MABwLQ6AX0WVcAiiEoB6AHwgANCAhzM0LslADevCAHcAClYTmUAorYEBPc4YBGvATAAazgYAGUZQNYmZBheDjhDOCFfOHRxdAAZASgmDgEmOAAxCF4hARgCXOQQAQ4YCAMQAAsYISIAdWbCCUiwODDnQhxCTxqwLh8QVh5eGHsApnLkADM3HkMAKy4ADwAhAODQsOE4TNY4VfXEkG2dsOiSAEUOCHgrog2QSN5ZmqYiJghCJeE0ZNJYB1MOgYM1kAAOAAMhnBEB4HQCMhq4IkfGwl0MAEdXvAFsYXLUuABaKBwNJpJq8ODEzBMhYFZZINafG48ISYWLxXmPOAvN6XJBxBKGGACXxQmFwpAAJhlAUwgNyAGEIMCBDUJABWJrcOAAFTlLm5X2wCQAklBxLAwlJZDAAIKOsIwTwkD48ZRAA)

```sh
# Options (if any):
--single-quote
--parser=glimmer
```

**Input:**

```jsx
<div title="{{t 'login.username.description'}}" />
```

**Output:**

```jsx
<div title='{{t 'login.username.description'}}'></div>
```

**Expected behavior:**

Behavior like parser=html

```jsx
<div title="{{t 'login.username.description'}}" />
```

Even if the output can be parsed by glimmer it is not readable with only the single quotes.

nraju99:
Hi! I would like to make this contribution!

dcyriller:
@nraju99 you have test cases already [in this file](https://github.com/prettier/prettier/blob/2.3.0/tests/format/handlebars/single-quote-option/attributes.hbs#L9-L12)

The code to adjust to get this issue fixed is probably here:
https://github.com/prettier/prettier/blob/2afc3b9ae674ede93489b55ab3f90ba2bcdd27c5/src/language-handlebars/printer-glimmer.js#L156-L166

Alonski:
@nraju99 Want to try fixing this please? :)

[...]

PR #11524, opened 2021-09-15
Fix mustache quotes inside attribute (#10419)

bmaehr:
## Description

Uses the opposite quote type for quotes inside mustache statements in attributes

## Checklist


tests/format/handlebars/single-quote-option/attributes.hbs
```diff
@@ -1,8 +1,12 @@
 <div title='My title'></div>
 <div title="My title"></div>
 <div title='My "title"'></div>
+<div title="My 'title'"></div>
```
dcyriller, line 4: 👍  then, we can remove the test case on line 5 which is a duplicate

tests/format/handlebars/single-quote-option/attributes.hbs
```diff
@@ -1,8 +1,12 @@
...
 <div title="My other 'title'"></div>
 
+<div title="{{t 'my.title'}}" />
+<div title='{{t "my.title"}}' />
+
 <a href="/{{url}}/{{url}}"></a>
 <a href='/{{url}}/{{url}}'></a>
```
dcyriller, line 12: May we add another more complicated test case, with nested SubExpressions? (note that it passes with the current implementation)
`<div class="padding {{if foo (if fooAgain "bar" (if fooAgainAgain "bar" "foo"))}} baz" />`

tests/format/handlebars/single-quote-option/attributes.hbs
```diff
@@ -1,8 +1,12 @@
 <div title='My title'></div>
 <div title="My title"></div>
 <div title='My "title"'></div>
+<div title="My 'title'"></div>
 <div title="My other 'title'"></div>
 
+<div title="{{t 'my.title'}}" />
+<div title='{{t "my.title"}}' />
```
dcyriller, line 8: this test case was missing, thanks

src/language-handlebars/utils.js
```diff
@@ -58,6 +58,18 @@ function isParentOfSomeType(path, types) {
...
+  let index = 0;
+  let parentNode = path.getParentNode(index);
+  while (parentNode && !isNodeOfSomeType(parentNode, types)) {
+    parentNode = path.getParentNode(++index);
+  }
+  if (parentNode) {
+    return index;
+  }
+  return -1;
+}
```
dcyriller, line 71: I am a little concerned by the performance of the while loop because `parentNode && !isNodeOfSomeType(parentNode, types)` would be relatively lax. To address this, we could invert the condition to `parentNode && isNodeOfSomeType(parentNode, types)`. This is possible if we make the function more specific to our usecase. For instance, the function could be akin:
```js
function needsOppositeQuote(path) {
  let index = 0;
  let parentNode = path.getParentNode(index);
  while (parentNode && isNodeOfSomeType(parentNode, ["SubExpression"])) {
    ++index;
    parentNode = path.getParentNode(index);
  }

  if (
    isNodeOfSomeType(path.getParentNode(index + 1), ["ConcatStatement"])
    && isNodeOfSomeType(path.getParentNode(index + 2), ["AttrNode"])) {
    return true;
  }

  return false;
}
```

In that case, we could move the util function from the `utils.js` file (that holds generic utils) to the main `printer-glimmer.js` ([here for example](https://github.com/prettier/prettier/blob/7ced9e66549b8d5a8e3a74473af462f86fb0245c/src/language-handlebars/printer-glimmer.js#L692)).

wdyt?

src/language-handlebars/printer-glimmer.js
```diff
@@ -397,6 +398,17 @@ function print(path, options, print) {
...
     }
     case "StringLiteral": {
+      const mustacheIndex = findParentOfSomeType(path, ["MustacheStatement"]);
+      if (
+        mustacheIndex >= 0 &&
+        isNodeOfSomeType(path.getParentNode(mustacheIndex + 1), [
+          "ConcatStatement",
+        ]) &&
+        isNodeOfSomeType(path.getParentNode(mustacheIndex + 2), ["AttrNode"])
+      ) {
```
dcyriller, line 408:
```suggestion
      if (needsOppositeQuote(path)) {
```

dcyriller (commented):
This is a nice fix @bmaehr, thank you for working on this!

tests/format/handlebars/single-quote-option/attributes.hbs
bmaehr: Removed line 4 because 5 was there before.

tests/format/handlebars/single-quote-option/attributes.hbs
bmaehr: Added

src/language-handlebars/utils.js
bmaehr: Good idea. I had the also concerns about the performance but was not aware of the fact all inside are SubExpressions.

bmaehr:
@dcyriller Sorry for the delay, I was ill. Would be nice if you could check again and perhaps merge.

changelog_unreleased/handlebars/11524.md
```diff
@@ -0,0 +1,45 @@
...
+{{!-- Prettier main --}}
+<div title="{{t 'login.username.description'}}" />
+
+{{!-- Prettier stable --}}
+<div title="{{t "login.username.description"}}" />
+
+```
+
+<!-- prettier-ignore -->
+```hbs
```
dcyriller, line 17: I'm not sure it is necessary to have a second example. You could sophisticate slightly the first one to:
```hbs
<div title="foo {{t 'login.username.description'}}" />
```

dcyriller (approved):

changelog_unreleased/handlebars/11524.md
bmaehr: My idea for the first sample was the usual usage and the second one was one of the existing tests where the behavior really changed. Should I change it the way you suggested even if you already approved it?
I still have the message "First-time contributors need a maintainer to approve running workflows" - Could you do/handle this? Are there any further steps necessary from my side? Sorry for the questions but it is my first PR for this project.

dcyriller: I am not a maintainer ;)

dcyriller: The goal of having this changelog entry is to present the users with the recent changes. So I guess the more concise and precise the changelog is, the best!

bmaehr: Ok, I removed the second example. Adding foo to the first one ist not necessary because from the user perspective only the change is interesting - the user expects that it works with simple and with complex cases and it is not necessary to some kind of "proove" that it also works for complex cases in the changelog (other than in a test).

fisker:
@dcyriller Is this included in #11559?

fisker (approved):

sosukesuzuki (approved):

PR #11525, opened 2021-09-15
[core] reuse `attachComment` logic and export from core API

pieterv:
## Description

Move all comment attachment setup logic to a central `attachComments` function. We can then reuse this within `multiparsers.js`. Additionally export the `attachComments` API under the `prettier.utils` object.

General notes:
- `attachComments` now returns the cached comments from `options` along with the newly attached comments, this enables that `ensureAllCommentsPrinted` to correctly check attached comments are printed even if attachment is manually triggered outside of the prettier call.
- `attachComments` is merged with the `utils` to avoid a circular dependency between `comments` and `utils`.

## Checklist

   - No will do this as a followup or after agreeing on the direction.


thorn0:
Did you see #8122? Eventually we'd like to stop mutating the AST. So I'm not comfortable with exposing this API.

thorn0:
On second thought, this change doesn't really expose how exactly comments are associated with nodes. If the `comments` property gets replaced with weak maps, it won't break. So it seems like my concern is irrelevant.

pieterv:
> Did you see #8122?

No i hadn't, that was interesting, thanks for the link. I would love to see a `prettier.formatAST(ast)` style API, the current `parser` style API is pretty clunky as you cannot cancel the printing after parsing (which is an important use case for codemods).

> If the comments property gets replaced with weak maps, it won't break.

Yes this makes sense to me. I guess the only caveat is when mutating the AST tools should use the prettier helpers for mutating comments rather than mutating the `comments` array (likely the same change you will need to make internally to prettier). However we might need remove + read versions of `util.addTrailingComment/addLeadingComment/addDanglingComment` to make this possible e.g. moving comments from one node to another is a comment operation.

pieterv:
Added `attachComments` API to `standalone.js` to fix the failing tests.

pieterv:
Derp, actually look into and the fix failing test. Verified via `yarn test tests/integration/__tests__/util-shared.js`, i just removed the check i added since `attachComments` is added to utils in `index.js` and verified in the other test i added.

pieterv:
Just following up here. Is there any changes needed to this? Happy to keep iterating if needed.

Issue #10706, opened 2021-04-15
MDX Comments don't seem to work


jeffreymeng:
It seems like prettier isn't parsing MDX comments properly (instead, it seems to parse the `*This is a comment*` part separately as italicized text.

(This isn't a spec, but here's where it seems to say that this type of comment should be supported: https://github.com/mdx-js/mdx/issues/1042)

**Prettier 2.2.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBiABACTgGxxAGnQHUIAnHAEwEIAdKdde4AegCoAVACwEsBndP3QBDdJAC24hDHRsWAXxAEQEAA4we0PslDCyZCAHcACnoTaUwnIeEBPbcoBGZYWADWcGAGVVrnlABzZBgyAFc4ZThxRzhKSliAGWFA0OEAuAAxcnFhGA1A5BBhUJgIJRAuGHEcYl54Pl8wOC9zHg0ANzbbQrA+BxB-PjgyGGMXAJzkADMrIeUAKz4ADwAhF3dPL2EpBP84admIkEWlr38AnDgARVCIeAOcOZBfMiGyQvFKJfLVMn8YYg8SgwLjIAAcAAZlL8IENiC5VIVfnA3u19soAI63eBjNQWIp8AC0UDgsVi5TIcCxPEpYzSkyQM0eRyG4h4wTCLPOlxud32jMOyhgwkcgOBoKQACYhS4eDhzgBhCCSYSFFEAVnKoSGHBFFiZT3a4QAklB4rAvGA-uoAIJmrwwWyXB5DeTyIA)

```sh
--parser mdx
```

**Input:**

```mdx
# Hello, World!
  
{/*This is a comment */}
```

**Output:**

```mdx
# Hello, World!

{/_This is a comment _/}

```

**Expected behavior:**
Input should already be properly formatted, and should not be modified.

thorn0:
Tried your example on the MDX playground https://mdxjs.com/playground and it's rendered as italics there, but it might be using an older version of MDX.

jeffreymeng:
Hmm that's interesting. The syntax highlighting on the playground also grays out the source code, as if it was a comment.

thecodingwizard:
[xdm](https://github.com/wooorm/xdm) treats `{/* */}` as a comment. I believe mdx v2 will also treat `{/* */}` as a comment (the playground may run on MDX v1, which might not consider that as a comment).

[...]

Issue #10157, opened 2021-01-27
Drop support for Node.js 10, switch to ES Module


fisker:
Node.js 10 will go end of life in April 2021, Node.js 12 has full native modules support.

alexander-akait:
> switch to ES Module

Not sure it will be good idea, because many tools and extensions for IDE, linter plugins/boilerplates and etc/etc/etc, whice have wide support Node.js (they have EOL which do not match with EOL of Node.js), don't forget we very popular, maybe we should do dual publish (cjs and esm), I am not against esm, but I think will be great to keep cjs version possibly until the end of the life of Node.js v12. We must not forget about our users, even if they are in legacy world

fisker:
We can convert codebase to esm, this will stop "direct-install-from-codebase" support for Node.js 10, and we can ship both bundled `cjs` and `esm`.

thorn0:
What about publishing separate packages for cjs and esm? Would it make sense? Or is it considered a bad practice?
Some users complain the package size has become too big.

fisker:
I'm not sure about cjs and esm, but we can at least publish seperate package for `standalone`.

thorn0:
Standalone is only 514 kB whereas esm has doubled the size of the package.

fisker:
I was  thinking `standalone` package with all umd/esm files, Node.js package have different cjs/esm file.  But I guess this won't solve package size problem. Sorry, my bad.

alexander-akait:
@fisker 

> We can convert codebase to esm, this will stop "direct-install-from-codebase" support for Node.js 10, and we can ship both bundled cjs and esm for a while.

Yep, sounds good :+1: 

> What about publishing separate packages for cjs and esm? Would it make sense? Or is it considered a bad practice?
> Some users complain the package size has become too big.

hm, can you provide link? Don't think it is real problem...

thorn0:
@alexander-akait Can't find it now. Hopefully it is indeed a non-issue

fisker:
Maybe we can publish pre language? Also prevent unexpected embed format 😄

alexander-akait:
> Maybe we can publish pre language? Also prevent unexpected embed format smile

Sounds good and bad :smile: Good for developers which known what they want, bad for new developers and DX

To be honestly I don't think size is real big issue for dev dependencies, I know many developers say I am wrong, but here two things:
1. good DX and big bundle - just install and all works
2. bad DX, a lot of packages - experience to understand what you need and extra setup

[...]

PR #11563, opened 2021-09-21
Add support for MDX 2 comments

wooorm:
## Description

This adds support for MDX 2 comments (which are the same as other
comments inside JSX) in addition to MDX 1 comments (HTML-like
constructs).

Please note that this is my first PR to prettier and I don’t know the codebase well.

Related to GH-10706.
Related to wooorm/xdm#61.

/cc @thorn0 @fisker

## Checklist


src/language-markdown/embed.js
```diff
@@ -52,6 +52,16 @@ function embed(path, print, textToDoc, options) {
...
+    case "esComment":
+      return (
+        "{" +
+        textToDoc(
+          "/* " + node.value + " */",
+          { parser: "babel" },
+          { stripTrailingHardline: true }
+        ) +
+        "}"
+      );
```
wooorm, line 64: I’m not sure why the MDX stuff is in `embed`, but that’s why I added it here

src/language-markdown/pragma.js
```diff
@@ -9,6 +9,7 @@ function startWithPragma(text) {
   const regex = new RegExp(
     [
       `<!--\\s*${pragma}\\s*-->`,
+      `{\\s*\\/\\*\\s*${pragma}\\s*\\*\\/\\s*}`,
       `<!--.*\r?\n[\\s\\S]*(^|\n)[^\\S\n]*${pragma}[^\\S\n]*($|\n)[\\s\\S]*\n.*-->`,
```
wooorm, line 13: This other regex looks complex, which is why I didn’t port that to an es-comment either

tests/format/mdx/mdx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -316,6 +324,14 @@ semi: false
 | -------- | ------------------- |
 | Text     | <Hello>Text</Hello> |
 
+---
+
+A
+
+{/* JS-style comment */} comment.
```
wooorm, line 331: Anyone any clue why there’s a blank line inserted before the comment (but not after?). Ideally there would be no blank lines

fisker: Seems inserted [here](https://github.com/prettier/prettier/blob/b2fa604e77dab3c9e33e832f639703e7d22f128f/src/language-markdown/printer-markdown.js#L719)

wooorm: thanks @fisker, fixed!

src/language-markdown/embed.js
fisker: I think this is not the right place to add it, it should be here https://github.com/prettier/prettier/blob/5aabd709505e6e858396586d2811b98fb06e40fc/src/language-markdown/printer-markdown.js#L75

When we disable embed language, it will throw.

[Playground link](https://deploy-preview-11563--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBBABMA9AKnQKQGUBaAZxgE8AbOdSAW3oRnRywF86JHmA6AHSiDseVFAgwAFnABOBEuWq0GTWKw4gANCAgAHGAEtopZKACGMmRADuABQsITKM1WtmKJ7QCMZZsAGs4GEIzJgAZAyg4ZAAzF1I4b18AoMJdP0iAc2QYGQBXRJA4ei84ABMy8rCzKEy8s0y4ADEIGXozGENa5B0YmK0QSRh6KgB1SQN4UnSwOEJHSYMAN0mKHrBSTxBIhJkYW19M9tj4woArUgAPQiyaAEU8iWikOKoE7XSZXZ76MsuB3QySIwUYGMpSZAADgADB8rAlRr5dD1AXBdktotoAI6PeAHPROEBmUjEKLlcoDGRwHEGKkHBrHF6nbQJegGHL5QqkW5wB5PE5vQowMxeUHgyTIABM2lyZgMVCyAGFuIyiqQAKwDPIJAAqIqcr3eICWBQAklBKrBCGAgfoxGVCJQaAKEux2EA)

wooorm: Pushed a fix to also add it to the normal printer. It looks like `embed` is for formatting embedded languages (so JS in MD), but not always available, hence it’s in two places?

fisker: Given they print the same result, I think we can remove this one.

Please add a test in https://github.com/prettier/prettier/tree/main/tests/format/mdx/embedded-language-formatting

wooorm: Does the embedded parser perhaps break lines in comments or in other ways format them? 🤔

src/language-markdown/printer-markdown.js
```diff
@@ -758,13 +760,15 @@ function getLastDescendantNode(node) {
...
 /** @return {false | 'next' | 'start' | 'end'} */
 function isPrettierIgnore(node) {
-  if (node.type !== "html") {
-    return false;
+  let match;
+
+  if (node.type === "html") {
+    match = node.value.match(/^<!--\s*prettier-ignore(?:-(start|end))?\s*-->$/);
+  } else if (node.type === "esComment") {
+    match = node.value.match(/^prettier-ignore(?:-(start|end))?$/);
```
fisker, line 768: It seems we don't have a test for `prettier-ignore` and `pragma` in mdx. Can you add some?

src/language-markdown/embed.js
fisker: Your current implementation just call the babel parser to print a JS block comment, and we print single comment as-it-is in JS printer.

The real question is does this comment support multiple line? And does it support comments that all lines starts with `*`? [We called it "IndentableBlockComment"](https://github.com/prettier/prettier/blob/5aabd709505e6e858396586d2811b98fb06e40fc/src/language-js/print/comment.js#L53), if it does, there will be differences (you didn't add test for it).

wooorm: Hmm okay. I removed it.

I think the primary goal right now is to add support for it. Then in the future, we can improve the parser, including the MDX part, by updating everything to latest remark?

fisker: Okay, I was going to ask you about that... I understand that's too much for you, but I failed to update `remark-parse`(#9397). It will be great, if you can help us for the upgrade sometime later.

src/language-markdown/printer-markdown.js
wooorm: added a test for `prettier-ignore` (which caught a bug in my code).

fisker:
@JounQin Can you please help review this?

src/language-markdown/printer-markdown.js
wooorm: Added a pragma test

src/language-markdown/embed.js
wooorm: My goal is to get MDX to an RC in 2 weeks (essentially porting xdm back into it and adding docs). So we do have some time (before but also probably after it).

I can help, yeah. All of unified is now ESM-only. I believe prettier is bundling dependencies so that shouldn’t be a problem for end users. Can ESM be setup internally in prettier/src/language-markdown?

fisker: Unfortunately, no. We still don't the problem, but we can't make ES Module work for the tests, even a simple esm file...  #11235

I guess we'll have to wait until Jest have better support for modules.

sosukesuzuki: Also, our migration to Pure ESM can be tracked at #10157

wooorm: can the markdown language be compiled first before tests?

JounQin:
Will prettier support `mdx@1` and `mdx@2` at the same time in the future? Or is this just a temporary workaround?

wooorm:
I see this as a workaround to at least not format MDX 2 comments.

I recommend to, as soon as possible, support MDX 2 only.

tests/format/mdx/embedded-language-formatting/pull-11563.mdx
```diff
@@ -0,0 +1,7 @@
+# title
+
+{ /* Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. */ }
+
+{/* Some more. */}
+
+{/* Some more. */}
```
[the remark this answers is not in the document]
wooorm: Remove the duplicate.
Stars are not yet supported, as old remark sees them as lists. Would be nice to have that, but it’s a lot of work, which would also be solved by updating to use latest remark and remark-mdx!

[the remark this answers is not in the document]
JounQin: A changelog should be added at https://github.com/prettier/prettier/tree/main/changelog_unreleased/mdx and mention this limitation.

sosukesuzuki (approved):
I'm not familiar with MDX, but looks good to me

tests/format/mdx/embedded-language-formatting/pull-11563.mdx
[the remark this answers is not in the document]
wooorm: How should I do this? I have no experience with this

Note that there are lots of limitations currently. xdm/mdx-js@2 allow any JS expression in there, so there are many limitations anyway, but comments (like this) solve a common use case, even though it's not perfect

fisker:
@wooorm https://github.com/prettier/prettier/pull/11563#discussion_r713967512

wooorm:
@fisker done. How does this look?

@JounQin What would you suggest for this changelog entry?

changelog_unreleased/mdx/11563.md
```diff
@@ -0,0 +1,18 @@
+#### Add support for MDX 2 comments (#11563 by @wooorm)
+
+This adds support for MDX 2 comments (which are the same as other comments
+inside JSX) in addition to existing MDX 1 comments (HTML-like constructs).
```
JounQin, line 4:
```suggestion
This adds basic support for MDX 2 comments (which are the same as other comments
inside JSX) in addition to existing MDX 1 comments (HTML-like constructs), only single star is supported right now.
```

How about notice the limitation here?

wooorm: I’m definitely :+1: on `basic`.

Not sure about `, only single star is supported right now.`.
Because there are many limitations to expressions/comments, not just single-line.
And I’m not sure what “single-star” means? What is “single star” about `{/* asd */}`?

JounQin: I saw `Stars are not yet supported` at https://github.com/prettier/prettier/pull/11563#discussion_r713963012 so I was thinking `single star` is the current supported. I'm not a native English speaker, so maybe you explain it in a better way. Anyway, my point is that a simple description about the limitation should be noticed here.

wooorm: Ah, my bad, that was confusingly worded, sorry!

Thanks Joun. I wrote some more about it. I think this is good?

JounQin (approved):

ChristianMurphy (approved):

sosukesuzuki:
@fisker https://github.com/prettier/prettier/pull/11563#issuecomment-927302855

wooorm:
Awesome!
Is there a rough time known for when that would be?

wooorm:
Thanks for merging this @sosukesuzuki, and everyone for the help!

kachkaev:
General MDX 2 issues: https://github.com/prettier/prettier/issues/12209

gutenye:
Does not support block comment

```mdx
{/*

hello

*/}
```

txp1035:
+1

PR #11559, opened 2021-09-20
Respect single-quote option in HTML attributes

dcyriller:
## Description

Would close https://github.com/prettier/prettier/issues/5828

This PR implements the single-quote option in all HTML attributes ~~but `class` and `srcset` (and this might be an issue)~~.

## Changes

```sh
--parser html
--single-quote
```

**Input:**

```html
<div title='My title'></div>
<div title="My title"></div>
```

**Prettier stable:**

```html
<div title="My title"></div>
<div title="My title"></div>
```

**Prettier main:**

```html
<div title='My title'></div>
<div title='My title'></div>
```

## Checklist

- [x] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).
- [x] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #11569, opened 2021-09-22
Refactor and use `getPreferredQuote` in both jsx and handlebars printers

dcyriller:
## Description

Later on we will be able to use it in HTML's printer (https://github.com/prettier/prettier/pull/11559#discussion_r713525169).

The changes to the `getPreferredQuote`'s arguments and return statement would be possible because it is not exposed (as it is not in the file `src/common/util-shared.js`).

The review should be easier commit by commit.

## Checklist

### DIFF
--- a/src/language-js/print/jsx.js
+++ b/src/language-js/print/jsx.js
@@ -481,8 +481,9 @@ function printJsxAttribute(path, options, print) {
       const raw = rawText(node.value);
       // Unescape all quotes so we get an accurate preferred quote
       let final = raw.replace(/&apos;/g, "'").replace(/&quot;/g, '"');
+
       const quote = getPreferredQuote(
-        final,
+        final.slice(1, -1), // remove enclosing quotes
         options.jsxSingleQuote ? "'" : '"'
       );
       const escape = quote === "'" ? "&apos;" : "&quot;";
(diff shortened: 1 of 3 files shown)


src/language-js/print/jsx.js
```diff
@@ -480,13 +480,15 @@ function printJsxAttribute(path, options, print) {
     if (isStringLiteral(node.value)) {
       const raw = rawText(node.value);
       // Unescape all quotes so we get an accurate preferred quote
-      let final = raw.replace(/&apos;/g, "'").replace(/&quot;/g, '"');
-      const quote = getPreferredQuote(
+      let final = raw
+        .replace(/&apos;/g, "'")
+        .replace(/&quot;/g, '"')
+        .slice(1, -1); // remove enclosing quotes
```
fisker, line 486: I think this should slice first.

fisker (approved):
Nice refactoring!

There might be further improvements we can do.

1. We are only unescaping `&apos;`, but not `&#39;` and other valid html entities. We can add functions like `unescapeQuote` and `escapeQuote` to do the job.
2. Maybe we can also add function like `smartQuote`, `smartQuoteHtml`, so we don't need manually replace quotes in languages.

These are out scope of this PR, just ideas.

src/language-js/print/jsx.js
dcyriller: Makes sense - it happend to work because the attributes' enclosing quotes might never be escaped.

sosukesuzuki (approved):
Thanks for doing this!

PR #11595, opened 2021-09-27
Update eslint config

fisker:
## Description

- Update config
- Enable some new `unicorn` rules

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,14 +4,11 @@ const { isCI } = require("ci-info");
 module.exports = {
   root: true,
   env: {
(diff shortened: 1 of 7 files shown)


sosukesuzuki (commented):
All new rules that is added by this PR make sense :+1: if errors caused by `import/extensions` are resolved, LGTM.

fisker:
Seems bug in `eslint-plugin-import`, it considers `node:fs/promises` as core module when running Node.js 16, but differently when running Node.js 14. I don't want to report it, updated config to "ignorePackages`.

Issue #10918, opened 2021-05-20
Disabling "smart formatting" of HTML class attribute


luukdv:
_"We get a lot of requests for adding options, but Prettier is built on the principle of being opinionated about code formatting."_

I totally understand this, and so I apologize beforehand for opening this issue. The reason I think this is a valid issue is because the opinion about this seems to have _changed_ from the perspective of Prettier, even though it was mentioned that there's no community interest for this change (and a lot of pushback) in the [introducing pull request](https://github.com/prettier/prettier/pull/7865).

**The issue**
When using frameworks like [Tailwind](https://github.com/tailwindlabs/tailwindcss), readability has lowered drastically since this feature:

<img width="320" src="https://user-images.githubusercontent.com/1065835/117627441-13b4c680-b170-11eb-824b-d22f07456e84.png" />

**References**
- [#836345253](https://github.com/prettier/prettier/pull/7865#issuecomment-836345253)
- [#836536543](https://github.com/prettier/prettier/pull/7865#issuecomment-836536543)
- [#836652940](https://github.com/prettier/prettier/pull/7865#issuecomment-836652940)
- [#839911172](https://github.com/prettier/prettier/pull/7865#issuecomment-839911172)
- [#841662012](https://github.com/prettier/prettier/pull/7865#issuecomment-841662012)

I'd love to still be able to use Prettier, but this is a blocking issue for me and many others.

**The solution**
Respecting the `class` attribute to stay on a single line, and only applying formatting it when a user already has a multi-line approach for `class`. Alternatively, an option to disable this feature.

Thanks in advance for considering this!

thorn0:
> Respecting the class attribute to stay on a single line, and only applying formatting it when a user already has a multi-line approach for class. 

This sounds like a good trade-off to me, but probably it can be improved. I guess developers who don't use Tailwind and have longer class names would prefer to keep the new behavior. So can Prettier probably somehow detect whether a single-line `class` attribute is "Tailwind-like"?

alexander-akait:
@luukdv 

> Respecting the class attribute to stay on a single line, and only applying formatting it when a user already has a multi-line approach for class. Alternatively, an option to disable this feature.

just interesting, what is the problem with multi line, I found it more readable

tkalmar:
I find the single line approach problematic when having many classes (like in a tailwind app with multiple breakpoints)
The big advantage i see is diffability (is this even a word?) i see clearly which classes are changed when i look at the diff.
[...]

Issue #7863, opened 2020-03-24
P2.x - Group HTML classes by prefix instead of forcing onto one line


pburkindine:
Unfortunately we cannot upgrade to 2.x until the change mentioned here: https://prettier.io/blog/2020/03/21/2.0.0.html#format-value-of-html-class-attribute-7555httpsgithubcomprettierprettierpull7555-by-fiskerhttpsgithubcomfisker

... is parameterized and made optional.  This would seriously affect the readability and maintainability of our templates.

The reasoning behind the change given here: https://github.com/prettier/prettier/issues/5482 is:

```
An example of such an attribute is class, where class=" a b c " should be semantically equivalent to class="a b c". However, note that if the user is parsing element.className directly, this change is not safe. 
```

We do not parse className directly; we do, however, make extensive use of long BEM class names and of bootstrap.

Here's a typical example:
```

<div class="ProviderMeasuresContainer__heading-row
            d-flex
            flex-column flex-lg-row
            justify-content-start justify-content-lg-between
            align-items-start align-items-lg-center"
...
```

vs.
```
<div class="ProviderMeasuresContainer__heading-row d-flex flex-column flex-lg-row justify-content-start justify-content-lg-between align-items-start align-items-lg-center"
```

You can see this greatly reduces the readability of the element.  This leads to longer development and debugging cycles as information in some arbitrary position to the right of the viewing screen is the missing piece on why XYZ is styled incorrectly.

thorn0:
I ask you again. Please use the Playground to report issue.

pburkindine:
Sure, I was able to for this one, still trying on the other:

[...]

PR #11827, opened 2021-11-17
Reverts #7865, keeping class names on the same line

jlongster:
## Description

This reverts #7865 which introduced automatically breaking class names across multiple lines. This has introduced significant community churn, particularly in CSS frameworks like Tailwind where it's typical to have a lot of class names.

I think this was on overstep on Prettier's part. The consequence of breaking this line is large and there's not a strong enough reason for it. While these decisions are somewhat subjective, I still don't see strong arguments to implement this feature. It's too opinionated. It's a "nice" thing for people that want it, but it wreaks havoc on code for people that don't want it. Given that balance, I think we should remove it.

If we merge this PR and someone really wants class names to be formatted, look into suggestions like here: https://github.com/prettier/prettier/issues/10918#issuecomment-971055862. I think it'd be fine to simply break the class names across lines if there is newline in there. It's not fully reversable, but we have the same behavior for objects. I don't 100% love it all the time, but it is a clever technique that balances the tradeoffs between users. I think it'd be fine here even if I don't want that behavior for everything.

## Checklist


tests/format/html/attributes/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -253,29 +253,17 @@ printWidth: 80
 
 =====================================output=====================================
 <div
-  class="
-    ProviderMeasuresContainer__heading-row
-    d-flex
-    flex-column flex-lg-row
-    justify-content-start justify-content-lg-between
-    align-items-start align-items-lg-center
-  "
```
j-f1, line 262: Can you add this output as a new test input to make sure that existing code with line breaks in the `class` attribute continues to be formatted correctly?

jlongster-stripe: This PR does not implement the behavior where it keeps it if there is already a line break. So they will get collapsed. 

I just realized why I never noticed this behavior before: In JSX, we just keep the class name as it exactly. It seems like the JSX and HTML behavior should be exactly the same. We should just keep the class name as is, so that means that all existing users won't be effected.

jlongster: This PR does not implement the behavior where it keeps it if there is already a line break. So they will get collapsed.

I just realized why I never noticed this behavior before: In JSX, we just keep the class name as it exactly. It seems like the JSX and HTML behavior should be exactly the same. We should just keep the class name as is, so that means that all existing users won't be effected.

azz (approved):

tests/format/html/attributes/__snapshots__/jsfmt.spec.js.snap
fisker: > In JSX, we just keep the class name as it exactly. It seems like the JSX and HTML behavior should be exactly the same. 

Formatting for class names in html was introduced in #7555  https://prettier.io/blog/2020/03/21/2.0.0.html#format-value-of-html-class-attribute-7555httpsgithubcomprettierprettierpull7555-by-fiskerhttpsgithubcomfisker

fisker (approved):
This will reopen #7863

sosukesuzuki (approved):

AndrewBogdanovTSS:
I can say that this is super weird behavior. Is there any chance to have this available at least behind an option, something similar to  proseWrap? Maybe classWrap?

alicercedigital:
Someone know how to solve this?
We need an formater to long classNames. There is a way to activate this without downgrade Prettier?

AndrewBogdanovTSS:
@alicercedigital I don't think it's possible at the moment, but highly needed in my opinion

PR #11855, opened 2021-11-23
Optimize visitNode by specialising by node type

joux3:
## Description

The new array path avoids creation of an `entries` iterator. The object path is something that has been
benchmarked to be a fast way to iterate over objects.

For an example case I had (about 2000 line .js file) this change brought
preprocess time down from 50ms to about 20ms.

Here are the `--debug-repeat` results for 100 iterations on Node.js v16.13.0. Before:
```
[debug] '--debug-repeat' option found, running formatWithCursor 100 times.
[debug] '--debug-repeat' measurements for formatWithCursor: {
[debug]   "repeat": 100,
[debug]   "hz": 7.541478129713425,
[debug]   "ms": 132.6
[debug] }
```
After:
```
[debug] '--debug-repeat' option found, running formatWithCursor 100 times.
[debug] '--debug-repeat' measurements for formatWithCursor: {
[debug]   "repeat": 100,
[debug]   "hz": 9.142439202779302,
[debug]   "ms": 109.38
[debug] }
```

## Checklist

### DIFF
--- a/src/language-js/parse/postprocess/visitNode.js
+++ b/src/language-js/parse/postprocess/visitNode.js
@@ -1,29 +1,27 @@
 "use strict";
 
 function visitNode(node, fn) {
-  let entries;
-
   if (Array.isArray(node)) {
-    entries = node.entries();
-  } else if (
-    node &&
-    typeof node === "object" &&
-    typeof node.type === "string"
-  ) {
-    entries = Object.entries(node);
-  } else {
+    // As of Node.js 16 using raw for loop over Array.entries provides a
+    // measurable difference in performance. Array.entries returns an iterator
+    // of arrays.
+    for (let i = 0; i < node.length; i++) {
+      node[i] = visitNode(node[i], fn);
+    }
     return node;
   }
-
-  for (const [key, child] of entries) {
-    node[key] = visitNode(child, fn);
+  if (node && typeof node === "object" && typeof node.type === "string") {
+    // As of Node.js 16 this is benchmarked to be faster over Object.entries.
+    // Object.entries returns an array of arrays. There are multiple ways to
+    // iterate over objects but the Object.keys combined with a for loop
+    // benchmarks well.
+    const keys = Object.keys(node);
+    for (let i = 0; i < keys.length; i++) {
+      node[keys[i]] = visitNode(node[keys[i]], fn);
+    }
+    return fn(node) || node;
   }
-
-  if (Array.isArray(node)) {
-    return node;
-  }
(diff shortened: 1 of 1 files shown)


sosukesuzuki (approved):

fisker:
We use `for...of` to iterate array and object everywhere, I don't think we want change them to `for` loop. If the benchmark result is true, there must be something wrong in the runtime or babel config.

joux3:
I did some more testing (this time without creating a bundle to speed up the feedback loop) with `--debug-benchmark`:

- the code in `main` branch: 6.9 ops/sec
- the PR code as of 8dafe050c0b610ddf90128768645c4de13d83d4a: 8.4 - 8.6 ops/sec
- the PR code but using `for of + Array.entries` in the array code path: 7.9 - 8.3 ops/sec (this varies a lot more than the previous one)
- the PR code but using `for of + Object.entries` in the object code path: 7.3 ops/sec
- the PR code but using `for of + Array.entries` & `for of + Object.entries` (so pretty much the original code reorganized): 6.9 ops/sec

The API of `Array.entries` is probably more forgiving to performance than `Object.entries` is as the first one gives out an iterator while the second one gives out an array of arrays.

Note: an unscientific observation was that using the idiomatic approach in any way increases the variability of the results a bit. Most probably those start performing way better after JIT.

Note2: benchmarking is always a hassle, feel free to test the changes on a real project with real amounts of code to see how your mileage varies.

Whether breaking the code base & idiomatic conventions makes sense for this hot loop is probably something I can't comment on.

fisker:
These two links also says the `Object.entries()` is slow [1](https://medium.com/hackernoon/5-techniques-to-iterate-over-javascript-object-entries-and-their-performance-6602dcb708a8), [2](https://gists.cwidanage.com/2018/06/how-to-iterate-over-object-entries-in.html)

lipis (approved):

joux3:
(Sorry @lipis, I had a miss click on the github UI with a mobile phone. I did not mean to request another review)

duailibe (approved):

fisker:
@joux3 Can you add a short comment to explain why we should not use `Object.entries()`? So people should not try to rewrite in future.

joux3:
@fisker done. Please verify whether this matches with what you had in mind.

(I also took the liberty of fixing the commit message which said that `Array.entries` returns an array of arrays. Actually it returns an iterator of arrays)

fisker (approved):
Thank you, LGTM

PR #11900, opened 2021-12-01
[handlebars] Named blocks can't be self closing

### LINKED ISSUE #11899: Glimmer: Named blocks without content are self-closed on format
**Prettier 2.5.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAwhAtgB2gmAdQEsYALAOQEMs4ATAIQBsIwBrAZwD4AdKAAn6okAM2IAnDjAC0UGvWkAjFuy6oA9KIlTZ8ukpVteAoUg5xIUfXNr7lrI0M3nL1vQYfHBwmwvvtpAHdSMmlILFpYLmBgAEJ+ABEIfigIGH5LADc4cXSYZKoBOFwYAE9+OCZiggBfGo0kXztDIJCw7EiYL1MmjwDg8nbYAi4ACUqWBt7-NlbBy3govg1MXHxYEnJqW2YHbj4+GPjU8SwqJgqqzo5+QIhxNn4qG7gADxwLeDp+OuW6YkyanU-0BBygR34AEl0uZijd8vwRBAmCxAvxyHB+Bx5PwWABzYhgJ43cJ4KAEG7qfgABWeYHO6Ge9Eu1VgHAA-D8astVmSCJsyBAAK4wTDDKIrbB8jYhYWi6CLLrLMgTCBAlUotVgkAAGhAEBwMGI0A4yFAVHE4gggVp4gQppQ50CVFKpr1inEVHYcBgAGV5AAZYjk5Aic7md2e71+nBe4N40PhuB64qKeh0egBwp4oVUPFwABi9zOMCNUATjpFEF1IDIMCwTEIZFIcA4sbAcF99tIANIpWQ4A4bpAwfMuWpnrxZ0TTAjIAAVhxXr741UAIpCtJwGdz2OSHIDvFMYgRA96nDiYNEYh0cjIAAcAAZz1bzIRPTgB06XcOAI6b+AJwNB0QGeWQ4Hoegaztf8JDgCc82nUQkz1cwsGIZAYHEIVkxADhVzgDctx3XCYCoRQSFvMhkAAJj1LCqGIY9y1WJCQFbABWGshXMAAVciHTDWdcMyHDISsAhfTAS9DQAQSsX0yiqEi9UXV4GCjNgfX9WggxDZDhLqIA)

```sh
--parser glimmer
--no-bracket-spacing
--prose-wrap always
```

**Input:**

```hbs
<ComponentWithNamedBlocks>
  <:first-named-block></:first-named-block>
  <:second-named-block> </:second-named-block>
  <:named-block-with-comment>{{! Do not convert to an empty element}}</:named-block-with-comment>
  <:named-block-with-content>Hello</:named-block-with-content>
</ComponentWithNamedBlocks>

{{! normal elements work as expected }}
<div></div>

{{! It seems to follow the same logic as components / PascalCased elements? }}
<ComponentWithoutContent></ComponentWithoutContent>
<hello></hello>


```

**Output:**

```hbs
<ComponentWithNamedBlocks>
  <:first-named-block />
  <:second-named-block />
  <:named-block-with-comment
  >{{! Do not convert to an empty element}}</:named-block-with-comment>
  <:named-block-with-content>Hello</:named-block-with-content>
</ComponentWithNamedBlocks>

{{! normal elements work as expected }}
<div></div>

{{! It seems to follow the same logic as components / PascalCased elements? }}
<ComponentWithoutContent />
<hello></hello>
```

**Expected behavior:**
```hbs
<ComponentWithNamedBlocks>
  <:first-named-block></:first-named-block>
  <:second-named-block> </:second-named-block>
  <:named-block-with-comment
  >{{! Do not convert to an empty element}}</:named-block-with-comment>
  <:named-block-with-content>Hello</:named-block-with-content>
</ComponentWithNamedBlocks>

{{! normal elements work as expected }}
<div></div>

{{! It seems to follow the same logic as components / PascalCased elements? }}
<ComponentWithoutContent />
<hello></hello>
```
Ember 3.25 enabled the yieldable named blocks feature which is explained in [this RFC](https://emberjs.github.io/rfcs/0460-yieldable-named-blocks.html).

Prettier does support the syntax already, but there is one case that doesn't work as expected: If a named block doesn't have any content Prettier will convert it to its self-closing equivalent (which seems to mimic the behavior for components in the Angle Bracket Invocation). The problem is that the Glimmer template compiler doesn't support self-closing named blocks and throws an error.

`<:named-block-name/> is not a valid named block: named blocks cannot be self-closing`.

So I think the named blocks should be treated the same as normal HTML elements instead of as components. The easiest way around this is to add a comment to the named block so that it will be not formatted as a self-closing element.

duailibe:
## Description

Closes #11899 

## Checklist

### DIFF
--- a/changelog_unreleased/handlebars/11899.md
+++ b/changelog_unreleased/handlebars/11899.md
@@ -1,0 +1,23 @@
+#### Support Glimmer's named blocks (#11899 by @duailibe)
+
+Prettier already supported this feature, but it converted empty named blocks to self-closing, which is not supported by the Glimmer compiler.
+
+See: [Glimmer's named blocks](https://emberjs.github.io/rfcs/0460-yieldable-named-blocks.html).
+
+<!-- prettier-ignore -->
+```jsx
+// Input
+<Component>
+  <:named></:named>
+</Component>
+
+// Prettier stable
+<Component>
+  <:named />
+</Component>
+
+// Prettier main
+<Component>
+  <:named></:named>
+</Component>
+```
--- a/changelog_unreleased/handlebars/11899.md
+++ b/changelog_unreleased/handlebars/11899.md
@@ -5,7 +5,7 @@ Prettier already supported this feature, but it converted empty named blocks to 
 See: [Glimmer's named blocks](https://emberjs.github.io/rfcs/0460-yieldable-named-blocks.html).
 
 <!-- prettier-ignore -->
-```jsx
+```hbs
 // Input
 <Component>
   <:named></:named>
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -25,13 +25,14 @@ function isLastNodeOfSiblings(path) {
 }
 
 function isUppercase(string) {
-  return /^[A-Z]+$/.test(string);
+  return string.toUpperCase() === string;
 }
 
 function isGlimmerComponent(node) {
   return (
     isNodeOfSomeType(node, ["ElementNode"]) &&
     typeof node.tag === "string" &&
+    node.tag[0] !== ":" &&
     (isUppercase(node.tag[0]) || node.tag.includes("."))
   );
 }
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -25,6 +25,7 @@ function isLastNodeOfSiblings(path) {
 }
 
 function isUppercase(string) {
+  return /^[A-Z]+$/.test(string);
   return string.toUpperCase() === string;
 }
 
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -26,7 +26,6 @@ function isLastNodeOfSiblings(path) {
 
 function isUppercase(string) {
   return /^[A-Z]+$/.test(string);
-  return string.toUpperCase() === string;
 }
 
 function isGlimmerComponent(node) {
--- a/tests/format/handlebars/basics/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/handlebars/basics/__snapshots__/jsfmt.spec.js.snap
@@ -93,6 +93,30 @@ printWidth: 80
 ================================================================================
 `;
 
+exports[`named-block.hbs format 1`] = `
+====================================options=====================================
+parsers: ["glimmer"]
+printWidth: 80
+                                                                                | printWidth
+=====================================input======================================
+<ComponentWithNamedBlocks>
+  <:first-named-block></:first-named-block>
+  <:second-named-block> </:second-named-block>
+  <:named-block-with-comment>{{! Do not convert to an empty element}}</:named-block-with-comment>
+  <:named-block-with-content>Hello</:named-block-with-content>
+</ComponentWithNamedBlocks>
+
+=====================================output=====================================
+<ComponentWithNamedBlocks>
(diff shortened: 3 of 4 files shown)


changelog_unreleased/handlebars/11899.md
```diff
@@ -0,0 +1,23 @@
+#### Support Glimmer's named blocks (#11899 by @duailibe)
+
+Prettier already supported this feature, but it converted empty named blocks to self-closing, which is not supported by the Glimmer compiler.
+
+See: [Glimmer's named blocks](https://emberjs.github.io/rfcs/0460-yieldable-named-blocks.html).
+
+<!-- prettier-ignore -->
+```jsx
```
fisker, line 8: Is there a better language for this? I don't know about this, just a question.

duailibe: oops, that was actually an oversight.. there's  `hbs` I think

src/language-handlebars/utils.js
```diff
@@ -25,7 +25,7 @@ function isLastNodeOfSiblings(path) {
 }
 
 function isUppercase(string) {
-  return string.toUpperCase() === string;
+  return /^[A-Z]+$/.test(string);
```
fisker, line 28: If it contains a number, how should we print it? Is there a test for it? Or maybe contains `-`.

changelog_unreleased/handlebars/11899.md
fisker: We can check how we use in the blog post before.

src/language-handlebars/utils.js
duailibe: This is how 2.5 prints it: https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAjAMwhABAegD4AdKVAWmwiNNSprIDEciQAaECABxgEtoAzslABDAE5iIAdwAK4hEJQiANlJEBPIRwBGYkWADWcGAGURAWzgAZXlDjJMKgXB17Dxk1322A5shhiAK4uIHDm2nAAJpFRViJQPoEiPnDMYuYiMHwJyCAigTAQ7CAAFjDmygDqJbzwAl5gcCYKtbwAbrXquWACWiC2zmIwMno+GQ5OIQBWAgAeJr7KcACKgRDwE8rOHF5ig7k+yrzmlmLFXGK2MJW8kTAlyAAcAAw7ks6Vely5F3CDbfYOABHNbwEbcRR5ATkOxRKLFMRwEG8REjZLjJCOLYhZzmXj+II4xYrUH2TGTDgwETaG53B5IABMlL0vCOCQAwhATiJcn8AKzFQLOAAq1MUWO2IDawQAklAYrATGBLjwAILykwwdRLTbOAC+eqAA

duailibe: For clarification: this is checking whether the tag is a Glimmer component or not. So to be strict, a named block `:foo` isn't a Glimmer component so we shouldn't print as one

Things starting with a number or `-` aren't even valid tags so I think it's out of the scope of this PR

duailibe: Ok, got it now. I'll change the implementation

duailibe: @fisker let me know if it looks better now

fisker (approved):

PR #11903, opened 2021-12-02
Migrate CLI to ES Module

fisker:
## Description

Part of #10157

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
   "scripts": {
     "prepublishOnly": "echo \"Error: must publish from dist/\" && exit 1",
     "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest",
-    "test:integration": "yarn test --config=jest-integration-test.config.mjs",
+    "test:integration": "echo \"Disabled temporary\" && exit 1",
     "test:format": "yarn test --config=jest-format-test.config.mjs",
     "test:dev-package": "cross-env INSTALL_PACKAGE=1 yarn test",
     "test:dist": "cross-env NODE_ENV=production yarn test",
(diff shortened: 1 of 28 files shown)


fisker:
Any idea to make this work?

fisker:
@SimenB Little help here? This PR changed `require()` to `import()`,  I can't make tests pass, what's the possible reason?

SimenB:
Does it happen if you only run a single test?

If it only happens when running multiple tests it's most probably due to faulty caching in V8 and tracked in https://github.com/nodejs/node/issues/37648

fisker:
> Does it happen if you only run a single test?

No.


> it's most probably due to faulty caching in V8 

Thanks.

fisker:
Temporary disabled "Integration Tests" on Dev version, I successfully convert CLI part into ESM.

Note: "Integration Tests" still running on production.

Do you think we can merge this? We are blocking by Jest forever...

package.json
```diff
@@ -143,7 +143,7 @@
   "scripts": {
     "prepublishOnly": "echo \"Error: must publish from dist/\" && exit 1",
     "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest",
-    "test:integration": "yarn test --config=jest-integration-test.config.mjs",
+    "test:integration": "echo \"Disabled temporary\" && exit 1",
```
SimenB, line 146: does it work if you run in band?

fisker: Can't run at all, we are `require()`ing, not `import()`, and `import()`+ runInBand not work too.

fisker: Want try? Maybe you can find something.

fisker: See #12271

fisker:
I'm going to merge this, so I can continue work on API migration.

package.json
SimenB: 👍

PR #11937, opened 2021-12-07
Update netlify config

fisker:
## Description

"Ubuntu Focal 20.04" is the default build image, https://app.netlify.com/sites/prettier/settings/deploys, I tried to switch it before, but fails because of  https://github.com/netlify/build-image/issues/599, now that issue has been closed, let's see if it works.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,4 +3,4 @@
 
 [build.environment]
   NODE_VERSION = "lts/*"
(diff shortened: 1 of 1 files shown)


fisker:
Everything seems to be working fine. I'm going to merge this, `NODE_VERSION = "lts/*"` should always keep us up to date without modify this config file.

PR #12005, opened 2021-12-28
docs(configuration): linkfix

simon04:
https://eslint.org/docs/user-guide/configuring/#example-configuration does no longer exist.

Is https://eslint.org/docs/user-guide/configuring/configuration-files an adequate replacement?


fisker:
How about link to https://eslint.org/docs/user-guide/configuring/?

fisker:
We need update this file as well https://github.com/prettier/prettier/blob/main/website/versioned_docs/version-stable/configuration.md#configuration-overrides

fisker (approved):

fisker:
Thanks for working on this

PR #12043, opened 2022-01-04
Enable `unicorn/prefer-json-parse-buffer` and `unicorn/no-useless-promise-resolve-reject` rule

fisker:
## Checklist


sosukesuzuki (approved):
Next time you create a PR like this, it would be nice if you could post a link to the rules document.

Issue #12078, opened 2022-01-13
Useless newlines between function parameters are kept if there is one empty line between parameters


bradzacher:
**Prettier 2.5.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgGIQQ7AA6UOOAtnDABYQAmAFGRRegDRmtsBGX5HAEpiAXx490AJykQA7jgC8OFoPYCeFfjxGKAfGLLioZAGYBXKGBgBLaDnSr1mtq57aoI4MbKQomGBxeJRVJDUEPXQNvAG4QDhAIAAdbaExkUGlZOQAFaQR0lHRUOXQAT3SE3il0MABrGgBldGoAGRsoOGRTYsw4Kpr6pqTajoBzZBgpc36QOEpeOAYGJdb0KDHzdDG4AilKdBhbDeQQdHMYCHiQWhhKVAB1Wht4TBGwOEaCl5sANxeyqcwNhrh0+lIYDkamMDt1erMAFaYAAejXGqDgAEVzBB4HDUH0EiMpODTrx0ItUNcklIOjAHjYGHRkAAOAAMRNkfQeNSSpxpcHBvy6CQAjjj4FDkoUzpgALSdJZLa5SODimyqqHbWFIHoE2Z9Sg2SbTA3orESrq6+EJGAUhlM2jIABMtpqNlQ4wAwhBKDq5pgAKzXcx9AAqFMKesJIF+MwAklAVrBGmBaSkAIJJxowMoY-F9USiIA)

```sh
--parser babel
```

**Input:**

```jsx
class Foo {
  method(
    a,

    b,
  ) {}
  
  arrow = (
    a,

    b,
  ) => {}
}

function a(
  a,
          
  b,
) {}

const b = (
  a,

  b,
) => {};
```

**Output:**

```jsx
class Foo {
  method(
    a,

    b
  ) {}

  arrow = (
    a,

    b
  ) => {};
}

function a(
  a,

  b
) {}

const b = (
  a,

  b
) => {};

```

**Expected behavior:**

Should format to:
```jsx
class Foo {
  method(a, b) {}

  arrow = (a, b) => {};
}

function a(a, b) {}

const b = (a, b) => {};

```

kachkaev:
👋 @bradzacher! This looks like expected behaviour to me – see  https://prettier.io/docs/en/rationale.html#empty-lines. WDYT?

bradzacher:
[...]

PR #12079, opened 2022-01-13
docs: link preserving empty lines to preserving comments

brodybits:
## Description

Coming from discussion in issue #12078

## Checklist

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -22,7 +22,7 @@ Prettier maintains the way your string is escaped. For example, `"🙂"` won’t
 
 ### Empty lines
 
-It turns out that empty lines are very hard to automatically generate. The approach that Prettier takes is to preserve empty lines the way they were in the original source code. There are two additional rules:
+It turns out that empty lines are very hard to automatically generate. The approach that Prettier takes is to preserve empty lines the way they were in the original source code, in a similar fashion to preserving [comments](#comments). There are two additional rules:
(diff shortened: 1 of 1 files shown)


docs/rationale.md
```diff
@@ -22,7 +22,7 @@ Prettier maintains the way your string is escaped. For example, `"🙂"` won’t
 
 ### Empty lines
 
-It turns out that empty lines are very hard to automatically generate. The approach that Prettier takes is to preserve empty lines the way they were in the original source code. There are two additional rules:
+It turns out that empty lines are very hard to automatically generate. The approach that Prettier takes is to preserve empty lines the way they were in the original source code, in a similar fashion to preserving [comments](#comments). There are two additional rules:
```
brodybits, line 25:
```suggestion
It turns out that empty lines are very hard to automatically generate. The approach that Prettier takes is to preserve the existence of empty lines the way they were in the original source code, in a similar fashion to preserving [comments](#comments). There are two additional rules:
```

adding "the existence of" to *hopefully* add clarification

brodybits (commented):
I am not yet convinced that I got the wording as clear and concise as it really should be ... suggesting a possible clarification that hopefully makes some sense here

docs/rationale.md
FuzzyDark: Has this issue been resolved?

Issue #12144, opened 2022-01-25
Reduce browser bundle size to be comparable with alternative formatters


TimvdLippe:
**Environments:**

- Prettier Version: 2.5.1
- Running Prettier via: Browser API
- Runtime: Chrome 99
- Operating System: n/a
- Prettier plugins (if any): parser-babel

**Steps to reproduce:**

I would like to include Prettier as part of a browser web page. In this case, I am looking at integrating Prettier into Chrome DevTools, which has a feature to automatically format web content in our sources panel. This feature is currently implemented using custom-written formatters ([example](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/entrypoints/formatter_worker/JavaScriptFormatter.ts;l=134;drc=12bcadba35805a2209e10bed6dbd8ee19b2558a8)).

The current formatters rely on the CodeMirror 5 tokenizers, which I am working on removing: https://bugs.chromium.org/p/chromium/issues/detail?id=1287519 As part of that work, I am analyzing the state of affairs of formatters used on the web and which ones we could potentially use. Since Prettier is an active and popular project, it is one of the first I am looking at.

I have a basic version of Prettier integration working, but I quickly discovered a blocker for us: the bundle size of Prettier greatly exceeds the size budget I have available. At Chromium, we have strict size restrictions for our projects, as users install Chrome on low-space devices. If I include only the standalone bundle + babel parser (required for JS), it clocks in at ~900Kb. I am working with a budget of around 100Kb, as we have both our CodeMirror 5 tokenizers as well as custom implementation of formatting.

At the same time, I looked at alternatives such as js-beautify, which is 40Kb in total (25Kb main bundle, 15Kb for css + html), which is well within our budget. However, js-beautify is less active than Prettier, which is why I initially would prefer to adopt Prettier.

[...]

PR #12145, opened 2022-01-25
Add assert to utils

TimvdLippe:
## Description

By implementing a custom function for assert, we no longer rely on the
Node global polyfill for the assert module. As part of #12144, I
discovered that the node global polyfill is a large contributor to the
file size.

Currently, on main I obtained the following numbers:

```
> yarn build --file=standalone.js --no-minify
> perl -le "map { \$sum += -s } @ARGV; print \$sum" dist/esm/standalone.mjs
< 1207295

> yarn build --file=standalone.js
> perl -le "map { \$sum += -s } @ARGV; print \$sum" dist/esm/standalone.mjs
< 562931
```

With this PR, the numbers change to:

```
> yarn build --file=standalone.js --no-minify
> perl -le "map { \$sum += -s } @ARGV; print \$sum" dist/esm/standalone.mjs
< 1136137

> yarn build --file=standalone.js
> perl -le "map { \$sum += -s } @ARGV; print \$sum" dist/esm/standalone.mjs
< 530682
```

This is a 5.89% and 5.72% size reduction for respectively unminified and
minified standalone bundle output.

## Checklist

### DIFF
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -10,6 +10,15 @@ const notAsciiRegex = /[^\x20-\x7F]/;
 const getPenultimate = (arr) => arr[arr.length - 2];
 
 /**
+ * @param {boolean} condition Condition that needs to be truthy
+ */
+function assert(condition) {
+  if (!condition) {
+    throw new Error("Assertion error");
+  }
+}
+
+/**
  * @typedef {{backwards?: boolean}} SkipOptions
  */
 
@@ -631,6 +640,7 @@ function describeNodeForDebugging(node) {
 }
 
 module.exports = {
+  assert,
   inferParserByLanguage,
   getStringWidth,
   getMaxContinuousCount,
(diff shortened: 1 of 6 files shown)


fisker:
https://github.com/prettier/prettier/pull/8165#discussion_r415812680

TimvdLippe:
> [#8165 (comment)](https://github.com/prettier/prettier/pull/8165#discussion_r415812680), I think this comment make sense, we should config build system to do thisq.

For my understanding, do you consider that a blocker of this PR or would you be okay to do that in a follow-up change? Given the impact of this PR on the overall bundle size, I think it would be worthwhile to land it as-is and then we can further prune the build size by removing them altogether.

duailibe:
I agree that should be the end goal, but there's still value in this PR while we don't do that.

In fact, this actually fixes a regression, because that shim was there exactly to prevent polyfilling the assert.

fisker:
Let me see if I can fix the build script quickly.

sosukesuzuki (approved):

fisker:
#12147

PR #12147, opened 2022-01-25
Exclude `assert` shim

fisker:
## Description

Closes #12145 
Closes #8165

## Checklist


fisker:
@TimvdLippe Can you take a look at this?

TimvdLippe (approved):
Yes I can confirm that with this PR, the file size becomes 530750, which is a similar improvement compared to #12145, thanks!

duailibe (approved):

duailibe:
I'm fine with either this or #12145. I do not think one is necessarily a better solution than the other, so feel free to decide on this @fisker 🙂

sosukesuzuki (approved):
#12145 is great, however it is better to solve this problem at build time. LGTM!

PR #12187, opened 2022-01-29
Build: Add `--compare-size` flag

fisker:
## Description

I was going to print

```text
index.js....................................1.53 MB, -191 kB (compare to v2.5.1)    DONE
```

But there is no enough room :(

image

## Checklist


sosukesuzuki (approved):
LGTM. But ideally, I think it is useful to compare to bundled `main`.

fisker:
Though it doesn't work well , but we already have https://github.com/prettier/prettier/blob/54df16cdf5c2d7839c1caaba9cd52526a67fd383/.github/workflows/prod-test.yml#L37, this is added to show sizechanges when doing release.

PR #12213, opened 2022-02-02
Fix printing of CSS selector() function

### LINKED ISSUE #10291: scss mixin breaks on format if name is "selector"
If a SCSS mixin is named "selector" (maybe other keywords also) and has a default value for a parameter, each format removes the last character of the default value.

**Prettier 2.2.1**

```sh
# Options:
--single-quote
--trailing-comma all
--no-semi
--print-width 120
```

**Input:**

```scss
@mixin selector($param: 'value') {
}
```

**Output:**

```scss
@mixin selector($param: 'value) {
}
```

**Expected behavior:**
Don't remove the last character of the default value...

duailibe:
## Description

Fixes #10291, this is an improvement on #8707

What happens here is that when there's a newline before the closing `)`, the parser reports the `endOffset` of that close group is equal to `startOffset + 1`, which is why the code using `locEnd(node.end) - 1` worked. But when there's no newline the parser reports `startOffset == endOffset`, so using `locEnd(node.end) - 1` gets part of the content of the selector.

The fix is to simply use `locStart(node.end)` which works and is also what we really want (get the slice of original text until the *start* of the close group)

## Checklist

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -493,7 +493,7 @@ function genericPrint(path, options, print) {
         grandParent.type === "value-func" &&
         grandParent.value === "selector"
       ) {
-        const start = locStart(parentNode.open) + 1;
+        const start = locEnd(parentNode.open) + 1;
         const end = locStart(parentNode.close);
         const selector = options.originalText.slice(start, end).trim();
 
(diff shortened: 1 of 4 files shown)


src/language-css/printer-postcss.js
```diff
@@ -493,8 +493,8 @@ function genericPrint(path, options, print) {
         grandParent.type === "value-func" &&
         grandParent.value === "selector"
       ) {
-        const start = locStart(parentNode.open) + 1;
-        const end = locEnd(parentNode.close) - 1;
+        const start = locEnd(parentNode.open) + 1;
```
duailibe, line 496: This wasn't really necessary but I think it makes sense to align `locEnd(open)` and `locStart(close)`

sosukesuzuki (approved):

PR #12221, opened 2022-02-04
Forbid use `--config` and `--no-config` together

### LINKED ISSUE #12012: Forbid use `--config` and `--no-config` together
Running `prettier --config=abc --no-config .` does not crash for me 🤔

_Originally posted by @kachkaev in https://github.com/prettier/prettier/issues/10274#issuecomment-1003180368_

It should be easy to fix after we merge #10274, I added `__raw` to cli argument parse result there.

Balastrong:
## Description

This PR fixes #12012 by throwing an error if both arguments are passed.

## Checklist


fisker:
It's better to handle those conflicting flags here. https://github.com/prettier/prettier/blob/543aa9abf194b73f95a7d90117ed1ce5dfefe874/src/cli/index.js#L45

Balastrong:
Thanks for the feedback!
I also wanted to put the error there as it seems more logical, however at that point the normalizer has already took action and the config value is `false` (I think it just takes the last value, in the example between abc and false).

I could manually look on `rawArgs` if I see both arguments as strings... Its probably less clean but I agree it's a more reasonable location in the code.

fisker:
This is why I said

> It should be easy to fix after we merge https://github.com/prettier/prettier/pull/10274, I added __raw to cli argument parse result there.

in the issue.

Maybe you can copy that part? Or just wait for #10274

Balastrong:
Ah right, maybe is better to wait for #10274 first :)

Balastrong:
I updated the PR with the new code from  #10274!

However, now I feel a little code smell on that copy paste... maybe we should add a more generic function (or even structure) to validate all cases of mutually exclusive options? Or is it overkill? :)

fisker (approved):
Sorry for the long delay, thanks for contributing!

PR #12225, opened 2022-02-05
Replace `minimatch` with `micromatch`

TrySound:
## Description

Micromatch already used inside fast-glob. It makes more sense to reuse
it. Though its behaviour slightly different I found a way to match
minimatch.

```
du -ck dist
```
before: 17260 kB
after: 17232

## Checklist

### DIFF
--- a/src/config/resolve-config.js
+++ b/src/config/resolve-config.js
@@ -168,7 +168,7 @@ function mergeOverrides(configResult, filePath) {
 function pathMatchesGlobs(filePath, patterns, excludedPatterns) {
   const patternList = Array.isArray(patterns) ? patterns : [patterns];
   // micromatch always matches against basename when the option is enabled
-  // use only patterns without slashes with it to match minimatch bahavior
+  // use only patterns without slashes with it to match minimatch behavior
   const withSlashes = [];
   const withoutSlashes = [];
   for (const pattern of patternList) {
--- a/src/config/resolve-config.js
+++ b/src/config/resolve-config.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const path = require("path");
-const minimatch = require("minimatch");
+const micromatch = require("micromatch");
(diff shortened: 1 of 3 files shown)


fisker:
> Support for directories with [] characters is something that minimatch has had issues with since 2014

This was mentioned in #7726

TrySound:
Is this good @fisker?

fisker:
I haven't look yet, is this going to be a problem? https://github.com/prettier/prettier/pull/12225#issuecomment-1031051496

duailibe:
I think this will actually fix a supposed bug on Windows when formatting a path with `[]` in it.

src/config/resolve-config.js
```diff
@@ -165,18 +165,30 @@ function mergeOverrides(configResult, filePath) {
...
+  // use only patterns without slashes with it to match minimatch behavior
+  const withSlashes = [];
+  const withoutSlashes = [];
+  for (const pattern of patternList) {
+    if (pattern.includes("/")) {
+      withSlashes.push(pattern);
+    } else {
+      withoutSlashes.push(pattern);
+    }
+  }
```
fisker, line 180: Will it increase bundle size if we use `lodash.partition()`? It's already used in other place.

TrySound: I would honestly try to move away from lodash. It handles a lot of unused cases with tone of code. And readability may be affected as well. Caz I have not idea how partition works and will need to use docs. While here everybody aware of for/of and arrays. I just don't see the value in using lodash with modern js.

fisker: I agree with you, I suggest this only because we already used `lodash.partition()`.

fisker:
Can you fix the conflicts?

duailibe (approved):

fisker (approved):

PR #12250, opened 2022-02-06
Include vendors/ when installing from GitHub

j-f1:
## Description

This includes the `vendors` directory in the package when installing from Github `(i.e. npm install prettier/prettier`).

Previously, Prettier would not be able to run because npm removes everything not listed in `files` when installing from GitHub. I’ve confirmed on a local project that installing Prettier from this branch does work correctly.

Closes #12253.

## Checklist

- [N/A] I’ve added tests to confirm my change works.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -70,7 +70,7 @@ module.exports = {
   moduleNameMapper: {
     "prettier-local": "<rootDir>/tests/config/require-prettier.js",
     "prettier-standalone": "<rootDir>/tests/config/require-standalone.js",
-    "#(.*)": "<rootDir>/vendors/$1",
+    "#(.*)": PRETTIER_DIR + "/vendors/$1",
   },
   modulePathIgnorePatterns: [
     "<rootDir>/dist",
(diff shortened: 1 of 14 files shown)


fisker (dismissed):

fisker:
Why this test didn't fail? https://github.com/prettier/prettier/blob/main/.github/workflows/dev-package-test.yml

duailibe (approved):

j-f1:
> Why this test didn't fail?

I ran that test locally on `main`, and added these lines of code to `get-string-width.js`:

```js
console.log(`__filename = "${__filename}"`);
console.log(
  `require.resolve("#string-width") = "${require.resolve("#string-width")}"`
);
```

Here’s the output from the test run:

```
__filename = "/private/var/folders/[...]/node_modules/prettier/src/utils/get-string-width.js"
require.resolve("#string-width") = "<cwd>/vendors/string-width.js"
```

It looks like the `imports` key in the package.json is resolved relative to the cwd (rather than the `package.json` location, for some reason. Perhaps this is a misconfiguration on our end, or a bug in Node.js?

fisker:
I run a test, it seems Node.js doesn't work like that, I think, it's caused by this https://github.com/prettier/prettier/blob/f19604a07ef7397072c66de42833419a87296178/jest.config.js#L73

fisker (approved):

sosukesuzuki (approved):
Thank you!

fisker:
Maybe better just import directly from `vendors/`?

duailibe:
> Maybe better just import directly from `vendors/`?

I think that's a good idea

j-f1:
Can I get one more ✅ on the latest changes?

sosukesuzuki (approved):

fisker (approved):

j-f1:
Thanks!

PR #12251, opened 2022-02-06
docs(webstorm): Add image for Prettier as External Tool

bennycode:
## Description

Currently the image on [Using Prettier as External Tool](https://prettier.io/docs/en/webstorm.html#using-prettier-as-external-tool) is broken because it got accidentally removed with #11909. That's why I am bringing it back.

## Checklist


fisker (approved):

fisker:
Thanks

PR #12271, opened 2022-02-11
[DEMO] Migrate CLI to ES Module 

fisker:
## Description

In #11903, we skipped dev version "Integration Tests", we are testing bundled cli(in cjs) with `require()`

What we want is,  use `import()` to run tests, so we can test both source(esm), and bundle(cjs) CLI

//cc @SimenB

Checkout this branch, do `yarn && yarn test:integration`


fisker:
@SimenB Found something?

fisker:
Close since it's old.

PR #12280, opened 2022-02-12
Disable wrapping JSXFragment in parens

j-f1:
## Description

Since `JSXFragment` nodes can only be `<>` and `</>`, this PR removes the wrapping parens around them in almost all cases. This reduces the indention of child elements by one step, and matches the UX of using parentheses around a single JSX element.

```jsx
// Before
return (
  <>
    <div className="Content">
      <p>Hello, world!</p>
    </div>
  </>
);

// After
return <>
  <div className="Content">
    <p>Hello, world!</p>
  </div>
</>;
```

## Checklist

### DIFF
--- a/changelog_unreleased/javascript/12280.md
+++ b/changelog_unreleased/javascript/12280.md
@@ -1,4 +1,4 @@
-#### Disable wrapping JSX fragments in parens (#12280 by @user)
+#### Disable wrapping JSX fragments in parens (#12280 by @j-f1)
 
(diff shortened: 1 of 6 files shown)


tu4mo:
Not a fan of this change. I think it looks prettier with parens. JSX start/end tags, whether it's a fragment or not, looks cleaner aligned together.

alexander-akait:
I agree, better to have `(` and )`

PR #12303, opened 2022-02-14
Migrate Markdown language to ES Module

fisker:
## Description

Based on #12278, should merge that one first.

Part of #10157

## Checklist

### DIFF
--- a/src/language-markdown/parsers.js
+++ b/src/language-markdown/parsers.js
@@ -1,14 +1,16 @@
-"use strict";
+import markdownParser from "./parser-markdown.js";
 
-module.exports = {
+const parsers = {
   /* istanbul ignore next */
   get remark() {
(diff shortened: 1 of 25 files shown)


src/language-markdown/parsers.js
```diff
@@ -1,14 +1,16 @@
-"use strict";
+import markdownParser from "./parser-markdown.js";
```
fisker, line 1: This became a top-level require in bundle, there is no way to make it work for both source and bundle. This will cause `index.js` slower to load (see #12262), I'll find a solution later.

A ESLint rule is disabled temporarily https://github.com/prettier/prettier/pull/12303/commits/839c531ad7554c81f15efdf2d56503dabdc5e45d

sosukesuzuki (approved):

PR #12345, opened 2022-02-21
Migrate `src/{document,utils}/**` to ES Modules

fisker:
## Description

~Include changes from #12344, will rebase after it merge.~

Part of #10157

This is the last part for the source code.

We'll decide if we want ship ESM bundle later.

## Checklist


fisker:
@sosukesuzuki I didn't wait for you to review other PRs since they are blocking each other, it will be great if you can take a quick look.

- #12278 
- #12303 (You already reviewed, list here just in case someone else want take another look)
- #12321 (Same as above)
- #12339
- #12341 
- #12340 
- #12342 
- #12343

Feel feel to comment or fix if you find mistakes.

PR #12346, opened 2022-02-21
Fix plugin Issue when node_modules not at root level within Intellij

szynkowskij:
## Description

When utilizing prettier within Intellij and the node modules is not at the root of the project, resolve will throw an error if a plugin is defined within the .prettierrc. The pluginSearchDirs finds the appropriate node_modules, but it is not used in the externalManualPluginInfos.

## Checklist

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -49,7 +49,9 @@ function load(plugins, pluginSearchDirs) {
         requirePath = resolve(path.resolve(process.cwd(), pluginName));
       } catch {
         // try node modules
-        requirePath = resolve(pluginName, { paths: [process.cwd()] });
+        requirePath = resolve(pluginName, {
+          paths: [process.cwd(), ...pluginSearchDirs],
+        });
       }
 
       return {


fisker:
This change makes sense to me, @thorn0 WDYT?

Note: we need to update [this part](https://github.com/prettier/prettier/blob/2cde2dbf9d280260f8c499039d89419b21fa4dc0/src/common/resolve.js#L11), if we decide to accept.

szynkowskij:
@fisker sorry you are correct. I didn't realize that I shouldn't be defining the plugins name within the configuration if it lived within the same directory and then should be auto loaded. 

If I define a `pluginName` and a `pluginSearchDirs` within my configuration though, it should be using the directory I defined. Looking at the method I updated, it only ever looks at the `process.cwd()` if I'm understanding this correctly. Which would mean, that it never actually uses that directory in order to find the plugin. For my scenario, I had tried defining just the `pluginName` and also tried `pluginName` and `pluginSearchDirs` which makes no difference for this block as long as `pluginName` was defined, but prettier would crash due to the resolve method in the catch also throwing an exception since it's only ever looking at the working directory. In Intellij, my structure is  is like:

>working directory 
>>java web app src
>>angular_app 
>>>node_modules

So based on the logic here, if I define a plugin in my configuration,  it will only load that plugin if it's within the current working directory. If it can't find it there, the resolve will throw an exception. I believe I should either be allowing me to define the directory where my manual plugin is (i.e. my change), or it should not be throwing an exception and try to see if it can auto load it within the following method. If it's defined and not able to be loaded ever, then it should be throwing the exception that I'm receiving.

Again this works fine if I run prettier from the root of my angular app and it finds the node_modules just fine, but since my angular app and my java back end are in the same project, I need the ability to define work my angular app is since Intellij is running prettier from the root of the project.

thorn0:
At the moment, `plugins` and `pluginSearchDirs` are independent options, idependent pieces of logic. Not sure if changing this will make the API simpler, not more complicated. The way @fisker describer should work with a little modification:
```js
// prettier.config.js
const path = require('path')
const paths = [path.resolve(__dirname, 'angular_app')]
module.exports = {
	plugins: [
		'prettier-plugin-foo'
	].map(plugin => require.resolve(plugin, { paths }))
}
```

PR #12347, opened 2022-02-21
Add prettier-plugin-jsonata to the Plugins page in docs

asennikov:
## Description

Adding a link to our plugin for [JSONata language](https://jsonata.org/) to the plugins page.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -159,6 +159,7 @@
         "jscodefmt",
         "jsesc",
         "jsfmt",
+        "jsonata",
         "jsonl",
         "judgements",
         "kalmanb",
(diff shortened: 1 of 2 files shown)


kachkaev:
Thanks @asennikov! Could you please move the new plugin up a bit so that the list was in the alphabetical order? I guess it won’t be wrong to also place `prettier-plugin-sh` before `prettier-plugin-solidity` – that’s the only other inconsistency I see in the list.

asennikov:
> Could you please move the new plugin up a bit so that the list was in the alphabetical order? I guess it won’t be wrong to also place `prettier-plugin-sh` before `prettier-plugin-solidity` – that’s the only other inconsistency I see in the list.

sure thing - sorted the list properly in my last commit!

kachkaev (approved):

fisker (approved):

Issue #12324, opened 2022-02-18
CSS values are transformed into lower case


alexander-rymasheusky:
**Prettier 2.5.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAbgQwDYFc4AEALABoAymATgOaEBMAHAPQDMAbAA4AeA3ADpQBAOngBnGAWACCBAGbQYSYuSq0CACTjZ0cGAEswmflAC+AkABoQEDvuijkoKpQgB3AApUEDlDleYATwcrACNKTDAAa10AZUwAWzgyPSg4ZFkcUThQ8KjYjgiU6mQYSnwrOHiQuAATGtqKKGpcTFoAMQhKeMwYfSbkEExcGAhLEAALGHjsAHVxvTECsDgY7wW9dAWAgbBRYJAUrMoYd3DqbvTM7JAAK1EuGKLsOABFXAh4S+wsqwLKI52ezGHEoKRgMz0NRg42QDAADL8XFkZuEOAMQXAjjoxgBHd7wU42HyDUQAWlStVqY0ocDxehpp1aFyQGW+1yy8T0JTK7Ker3xaRZVysMEwIQhUJhSDoIvCemwRQAwhB4syQJiAKxjXBZAAqYp8rJ+IHQ+AAklB6rAYmBQbYAIKWmIwALPL5ZEwmIA)

```sh
# Options (if any):
(no options)
```

**Input:**

```css
@value 4XLarge 28/36px;

.test {
  font: 4XLarge Helvetica;
}
```

**Output:**

```css
@value 4XLarge 28/36px;

.test {
  font: 4xlarge Helvetica;
}
```

**Expected behavior:**
```css
@value 4XLarge 28/36px;

.test {
  font: 4XLarge Helvetica;
}
```

The project uses PostCSS with its @value and `@value...from` syntax extensions.  Typography rules are defined using constants like shown above; renaming the values is not an option.  Changing the case of the values is an error (violates the "Correctness" principle).

fisker:
The problem caused by this https://github.com/prettier/prettier/blob/888db50d9ff7e1f2be877494c6016de8ec5970da/src/language-css/printer-postcss.js#L972, and I have already built [css-units-list](https://github.com/fisker/css-units-list) for this, but forgot to fix.

niklasvatn:
Hi,
Looks like a bug I could try to fix.
Is it only when the unit is a CSS unit we want to make it lowercase?

[...]

PR #12393, opened 2022-03-05
Fix 12324

niklasvatn:
## Description

Using [css-unit-lists](https://github.com/fisker/css-units-list) to check whether a provided unit should be formatted or not. This solves the problem of lowercasing PostCSS-values (#12324) when they start with a digit.

**NOTE:** This PR adds a new dependency. 

## Checklist

- [x] I’ve added tests to confirm my change works..


fisker:
Looks good, but maybe we should use a `Set` instead of array.

niklasvatn:
Cool, I thought about it as well to improve the performance
Is it ok to create this file on top of the changed file or would you prefer it be created elsewhere?

fisker:
We can use a seperate file like `css-unit.evaluate.js` so we can evaluate to object when bundling.

niklasvatn:
> Looks good, but maybe we should use a ~`Set` instead of array.~ Sorry I mean `Map` or an object.

I first implemented it using Map, but some of the checks said that `Map.get` was not a function. Changed it to an `Object` and now it seems to work.

src/language-css/css-units.evaluate.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const cssUnits = require("css-units-list");
+
+const CSS_UNITS = new Map(cssUnits.map((unit) => [unit.toLowerCase(), unit]));
```
fisker: This will evaluate to JSON, it can't be a `Map`.

fisker: Sorry.Commented from the notify. Outdated.

fisker (approved):

src/language-css/utils/css-units.evaluate.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const cssUnits = require("css-units-list");
```
alexander-akait, line 3: @fisker Looks good, one missing unit here https://github.com/fisker/css-units-list/blob/main/index.js#L39 - `x` https://www.w3.org/TR/css-values-4/#resolution

alexander-akait (approved):
Let's fix `x` and we can merge

niklasvatn:
Updated the version of `css-units-list` and added `x` to the test case. I think it is ready to merge now.

PR #12428, opened 2022-03-10
Add CSharpier to List of Community Plugins

Kurt-von-Laven:
CSharpier is a standalone tool for C# based on a port of Prettier.

## Description

Add a link to CSharpier on the Prettier landing page.

## Checklist

### DIFF
--- a/website/data/languages.yml
+++ b/website/data/languages.yml
@@ -44,6 +44,7 @@
   image: /images/languages/tools_wip.svg
   variants:
     - "[Apex](https://github.com/dangmai/prettier-plugin-apex)"
+    - "[C#](https://github.com/belav/csharpier/) (CSharpier is a standalone tool based on a port of Prettier)"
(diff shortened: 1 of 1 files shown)


fisker:
I don't think that's a prettier plugin. Maybe https://prettier.io/docs/en/related-projects.html?

fisker:
Ha, it's already there.

Kurt-von-Laven:
Yeah, CSharpier is definitely not a plugin. It does seem though that the more probable user need would be "how can I get the same or similar formatting for language _X_?" and not "is there a Prettier plugin for language _X_?" In other words, as a user of Prettier, I would ideally want to be able to see a list that includes all ports, plugins, etc. sorted by language in one place and not have to know that I need to look at the related projects page for ports if the language I care about isn't listed among the plugins.

Harry-Hopkinson (approved):

PR #12430, opened 2022-03-10
Add JS-in-JS formatting

j-f1:
## Description

This PR adds basic support for formatting JS in JS template strings. It adds support for the following syntaxes:

```js
<script>{`
  // JS goes here!
  // You can’t use interpolations, and this must
  // be the only child of the <script> tag.
`}</script>;

const foo = /* JS */ `
  // JS goes here!
  // You can’t use interpolations.
`;
```

I’ve also added support for all of the languages supported by `<script>` tags in HTML.

## Checklist


fisker:
https://github.com/prettier/prettier/pull/5993#issuecomment-478573220

j-f1:
Would it be reasonable to add just the `<script>` changes for symmetry with HTML? And/or has the rationale changed? While this is fairly rare, it is annoying that there is no way to write JS code inside of JS and have it be formatted without adding additional syntax (like `(() => {...}).toString()`).

MichaelDeBoey (commented):
I would like to suggest to also support `js`, `jsx`, `ts` & `tsx` literal template tags in both JS & TS, just like we already do for `css`, `markdown`, `graphql`, ...

shaodahong:
Can this feature be a plugin? It would be useful for people who want to add it.

PR #12446, opened 2022-03-13
Update `chalk` to v5

fisker:
## Description

Closes #12417

Bundled `cli.js` file reduces ~28kb

## Checklist


alexander-akait (commented):
Just interesting - do we really need this big package, maybe we can replace or avoid it?

fisker:
It's 17kb, not that big. But I'm fine to replace it with another alternative module.

fisker:
I'm going to merge this, but feel free to replace it.

PR #12463, opened 2022-03-15
Build: Add a plugin to throw unexpected warnings

fisker:
## Description

Quote from https://github.com/evanw/esbuild/issues/2103#issuecomment-1067919506

> Most warnings for files inside node_modules are debug messages instead of warnings. One reason is because the warning isn't immediately actionable, since the person seeing the warning has no way of fixing the problem (only the package author can fix the problem). Another reason is because every time I add a warning that triggers for files inside node_modules, someone ends up filing an issue telling me to remove it. So I have been making them debug messages instead of warnings.

Add a plugin to throw unexpected warnings.

## Checklist


fisker:
I'm going to use this plugin to make sure #12461 is safe to merge.

Issue #12408, opened 2022-03-06
Inconsistency between object property values with quotes and template string


B2o5T:
**Prettier 2.5.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzCFMC8mwAOlJpgPSWYBGATnAIYDWmM+UcA7pgDYBLLpgAmAVziYAjmIjw0ZCrgi0mSTAHIAbnHoBPftADmmDPSEmd+w1BNmLmKwb7HTMc7ce7nr+560aADSKVDQwABYCaJhQcnSMrKIS7HAAtgAOfEzwbh5GIcqq6gAGTjZ27g5lLp5+lt7luVUNNRV5jsXB5KHxzGwcMTz8QpIiEHDRsVjhTOnpCJgC2Jjp9BDz9DAGUEypkpCwTEKYADxEACyYYDP0TGDw9NGAfBuAKLsFeLTq2i2+lf4-tT+9WsrSa-xBv3aWiYfAkQTIAF8QIEQOsYAJ0MhQEx6GtuAAFHEINDIEAw7hMPQklEMO4sOAwADKuzgABkRshsDC0HAabcwPSmek7hZkO4JCi0rQ4CIRDLWUxbGImEY4AAxCD0VLZdG2UlMMQcZEgcIwVJ8ADqkXkwrAcEZxIE6K0Tr0pLAaGpICEPM2+NuRm1nO5vJAACs0AAPRkWPhwACKsngwb4PJRwseulJqmlfGNqyEMAtAhEEWQAA4AAzptY8i23dKk1YTXQ6Y0yORwf3rEkoJhoAC0XBlMuNjBkAkY-pVQaQXNToZ5qQEYvoEpAaFjCaTcFX65gTFoxdL4WQACYUe4joJbABhCCpWcgCYAVmNYh5ABVD7352mQFoEgAJJQHKsCMmA5jpDAACCoGMlscYpjyCIIkAA)

```sh
--parser babel
--single-quote
```

**Input:**

```jsx
const foo = {
  // break to new line due quotes
  fooba: 'very long string very long string very long string very long string v',
  // this not break due template string
  fooba: `very long string very long string very long string very long string v`,
  // break to new line does not happen if property name contain <= 4 characters 🤔
  foob: 'very long string very long string very long string very long string value',
}
```

**Output:**

```jsx
const foo = {
  // break to new line due quotes
  fooba:
    'very long string very long string very long string very long string v',
[...]

PR #12465, opened 2022-03-15
[JavaScript] Make consistency between object property values with string literal and template string

B2o5T:
## Description

Fixes https://github.com/prettier/prettier/issues/12408

## Checklist


B2o5T:
@fisker @sosukesuzuki Hey guys 👋, it's been over a month since I created my small PR that fixes the current inconsistency, can someone review it and say what you think?

fisker:
Sorry for the long delay.

I'm not sure if we want to change it, at least not until a better solution found.
Making an exception here is really bad. People will ask why long identifiers/numbers/bigints don't format like this.
If you want to change the output, add more tests, we'll see if it's fine to accept.

B2o5T:
@fisker which tests would you prefer that I add? Only for properties of an object?

fisker:
> long identifiers/numbers/bigints don't format

B2o5T:
@fisker added more tests

thorn0:
The current behavior for string literals is intentional. Please see https://prettier.io/blog/2021/05/09/2.3.0.html#prevent-wrapping-object-properties-with-short-keys-10335httpsgithubcomprettierprettierpull10335-by-thorn0httpsgithubcomthorn0

If we're going to change something here, that should be the behavior for template literals.

PR #12489, opened 2022-03-18
Build: Combine `replace-module` and `replace-text` plugin

fisker:
## Description

This doesn't change the output, just combine them for easier use.

## Checklist


sosukesuzuki (commented):
Personally, I think it is better to have those plugins separate because they do different things.

fisker:
I don't think they are doing different things, you can achieve `replaceModule` result with `replaceText`.

```js
replaceModule: {foo: {contents: “”}}
```

equals

```js
replaceText: [file: foo, find: /.*/s, replacement: ""]
```

----


```js
replaceModule: {foo: {path: bar}}
```

equals

```js
replaceText: [file: "*", find: `require(${foo})`, replacement: `require(${bar})`,]
```

----

```js
replaceModule: {foo: {path: bar, external: true}}
```

equals

```js
replaceText: [file: "*", find: `require(${foo})`, replacement: `require(${bar})`,],
external： [bar]
```

----

Motion behind this change: Some times I want try different solutions, but options for `replaceModule` is an object, options for `replaceText` is an array, hard to switch between them.

sosukesuzuki (approved):
I got it. Also can you add documentation for this plugin as an another PR? It is getting a little complicated...

fisker:
Sure, I can try.

PR #12492, opened 2022-03-19
Add typeAnnoation test with cursorOffset

### LINKED ISSUE #12491: Wrong cursorOffset with Identifier.typeAnnotation
Since the playground does not have the `cursorOffset` option (from what I could see), I've put it into RunKit: https://runkit.com/gregoor/prettier-cursor-offset-bug

Here is the code:
```ts
const prettier = require("prettier");
prettier.formatWithCursor("let i: T;", { cursorOffset: 5, parser: "typescript" });
```

which erroneously returns a `cursorOffset` of `7`, where I'd expect it to stay `5`.

I might look into a fix myself this weekend, unless someone tells me that my expectation are wrong and this is actually expected behavior.

Gregoor:
## Description

Atm this only adds a test, probably in not-the-best-location but there were no non-snapshot tests for ts yet. So I wasn't sure, should I just add `tests/format/typescript/cursor/jsfmt.spec.js` and move it into it?

I think the bug is in these lines: https://github.com/prettier/prettier/blob/ae080df3e7c38e22ddb47699f71bd2bbca3d822e/src/document/doc-printer.js#L584-L600

But I have not understood them enough yet to fix it and not cause other issues. TBC I hope!

Fixes #12491

## Checklist


fisker:
> should I just add tests/format/typescript/cursor/jsfmt.spec.js and move it into it?

Yes.

Gregoor:
I'm looking at this again and got a bit stuck unraveling the data flow. 

https://github.com/prettier/prettier/blob/3ef82ae06b41574d9c45ab2506553bcc00afc019/src/document/doc-printer.js#L282-L283

This is a case for the `doc.type` switch. What I am failing to discover is where the cursor type is put into the doc. Do you have an idea @fisker, or know someone who could give me a pointer?

fisker:
https://github.com/prettier/prettier/blob/3ef82ae06b41574d9c45ab2506553bcc00afc019/src/main/comments.js#L572-L575

fisker:
It's called here

https://github.com/prettier/prettier/blob/3ef82ae06b41574d9c45ab2506553bcc00afc019/src/main/ast-to-doc.js#L145

Gregoor:
Thanks! It looks like that branch is not hit for my test case though (since `comments.length === 0` an early-return kicks in that function). So I'm still a bit at a loss in that regard.

But then again I did find another fn which calculates the `cursorOffset`. I don't quite get when which is used (both seem to be called for my test case), so I'll try to see what damage I can do there

https://github.com/prettier/prettier/blob/3ef82ae06b41574d9c45ab2506553bcc00afc019/src/main/core.js#L70-L138

fisker:
`comments.length === 0` is not the problem.

I also took a look at this issue, problem is here

https://github.com/prettier/prettier/blob/3ef82ae06b41574d9c45ab2506553bcc00afc019/src/language-js/print/misc.js#L65

`:` is not printed inside `TSTypeAnnotation`, so cursor is insert after `:`, if we move `:` print into `TSTypeAnnotation` will fix the issue, but I'm not sure what's the best way to fix it, because other problems should already exists for other nodes.

Gregoor:
Thanks for digging! True, I'm seeing a lot of other failed tests when I move the colon into `TSTypeAnnotation`. I'll see if I can get to a good place in reasonable time.

I was wondering if there is not a different approach for calculating `cursorOffset`, namely:
1. find and remember the path to the cursor node 
2. calc and remember the offset for the cursor within that node
3. format
4. find the node again, using the path
5. return newNodeStart + oldOffsetWithin
6. Tada?

Example: `a= 42;` with cursor at `4` (middle of the number literal)
1. path: `/program/body/0/expression/right`
2. offset `1`
3. formatted `a = 42;`
4. path is ofc the same
5. new node start is at `4` + offset `1` = `5`
6. Tada indeed?!

I suppose I might be overlooking an edge case or 10, but it seems simpler to me, so I thought I'd at least ask!

fisker:
> path is ofc the same

Unfortunately, we chang AST some times.

eg: remove/insert semi colons, remove/add parens .

Gregoor:
For semis and parens the AST paths should still stay the same, no? I just tried it for my application, and it did the job except for JSX. E.g. `<><></></>;` get formatted in such way that new `JSXText` nodes are inserted, thus the path is indeed not the same.

That said, it did solve this issue for me. And since I'm a bit overwhelmed by the amount of failing tests when moving `:` into `TSTypeAnnotation` I'd leave this PR for now. Happy to jump back into it, if there are some other ideas.

fisker:
> For semis and parens the AST paths should still stay the same, no?

```js
[].forEach();

// -> {semi: false}
;[].forEach()
```

```js
foo && (bar && baz)

// ->

foo && bar && baz
```

Gregoor:
> > For semis and parens the AST paths should still stay the same, no?
> 
> ```js
> [].forEach();
> 
> // -> {semi: false}
> ;[].forEach()
> ```

Oh good point, that adds an extra statement which my retain-path algorithm would not be able to account for!

> ```js
> foo && (bar && baz)
> 
> // ->
> 
> foo && bar && baz
> ```

These would work fine though, because parenthesized expressions don't introduce any AST changes, thus the paths would have stayed the same.

Anyway, looks like you got this one covered. Thanks for the exchange and support (and prettier in general)!

PR #12501, opened 2022-03-20
docs: AstPath.match comment.

NaridaL:
## Description

I couldn't find docs describing what the AstPath.match function is supposed to do so I attempted to describe what the code does. Please check if the docs match what it is supposed to do.

## Checklist

### DIFF
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -126,6 +126,19 @@ class AstPath {
   }
 
   /**
+   * Checks whether the nodes at the end of this path match the respective
+   * predicate, in reverse order. Array indexes are ignored.
+   *
+   * For example, if the current stack were
+   *
+   * `[$root, "bar", $barNode, "fooNodes", $array, 2, $fooNode]`
+   *
+   * the following calls would return true:
+   *
+   * - `match(isFooNode)`
+   * - `match(isFooNode, isBarNode)` (arrays are skipped)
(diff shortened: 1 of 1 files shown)


Harry-Hopkinson (approved):

sosukesuzuki (approved):
LGTM. /cc @fisker

fisker:
I don't think the comment correctly explained this method. But I don't know how to express this, forgive my poor English skill. There is nothing to do with the stack, the predicate are testing node ancestors.

NaridaL:
> I don't think the comment correctly explained this method. But I don't know how to express this, forgive my poor English skill. There is nothing to do with the stack, the predicate are testing node ancestors.

But it's checking the stack? `node = this.stack[stackPointer--]` etc

fisker:
But user should not care how it works.

NaridaL:
> But user should not care how it works.

I'm not seeing a better way to describe an example though...

fisker:
Maybe

```js
foo.bar(baz)
```

```js
match(node => node.type === "Identifier" && node.name === "baz");
match(
	node => node.type === "Identifier" && node.name === "baz",
	node => node.type === "MemberExpression" && !node.computed && node.property.name === "bar",
);
```

fisker:
Close due to age. Feel free to ask me to reopen.

PR #12505, opened 2022-03-21
[JavaScript] Add an option to avoid adding line breaks to template literals

JohnnyFangJr:
## Description

Add an option to avoid having Prettier add line breaks in template literal expressions.
There have been many issues opened for template literals in the past, the top issue that's still active is #3368 .
Although that issue's aim is to improve the formatting, the comments are filled with asks to have a configurable option.

I made this PR hastily to test the waters for whether there's a chance this new option would actually get accepted.
After(if) there's initial traction, I'll work my way through the checklist to make this a proper PR.

## Checklist

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -104,6 +104,19 @@ Note that Prettier never unquotes numeric property names in Angular expressions,
 
 If this option is set to `preserve`, `singleQuote` to `false` (default value), and `parser` to `json5`, double quotes are always used for strings. This effectively allows using the `json5` parser for “JSON with comments and trailing commas”.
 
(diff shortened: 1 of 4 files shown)


i1u5:
I guess the `ignoreTemplateLiterals` option falls under the "great demand" category in your philosophy and I think you guys should consider re-opening this PR and let others contribute to it, feel free to keep the default to `off` if you don't like it but at least give us a choice other than `printWidth` because it negatively affects other functions too and I'd prefer if it stayed at 80.

thongxuan:
"Now that Prettier is mature enough", lol

PR #12519, opened 2022-03-23
[graphql] Add support for printing `SchemaExtension` nodes

trevor-scheer:
## Description

`SchemaExtension` nodes are valid graphql but prettier doesn't currently know how to handle them. This adds the ability to print these nodes.

Worth mentioning:
* `SchemaExtension` doesn't support `description` the way `SchemaDefinition` does
* `node.operationTypes` is optional on `SchemaExtension` whereas it's required on a `SchemaDefinition`
* When `node.operationTypes` is empty, no `{ }` should be printed

Relevant bits in the language spec: https://spec.graphql.org/October2021/#sec-Schema-Extension

## Checklist


fisker (approved):
LGTM. Thank you!

sosukesuzuki (approved):

Issue #12540, opened 2022-03-26
Prettier adds redundant semicolon when formatting javascript ranges


rotu:
**Prettier 2.6.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEBLTfMC8AOiAGYQBOAtgIZZVwCEJ+A3CADQgQAOM20AM7JQNChQgB3AApiEwlDVSSaAT2FcARhRpgA1hgDKNBgBlsUOMlJLBcLTv1GeuiwHNkMCgFd7IOFSacAAmwSGmNFBu3jRucABilLQw-FHIIDTeMBCcIAAWMFSoAOp52PCCLmBwhvLl2ABu5arpYIIaIBZ2FDDSOm601rZ+AFaCAB6G7ugAit4Q8EOodlwuFN3ppKhSuTwUFjDF2MEwecgAHAAMqxJ2xTo86Xtw3Q1WXACO8-B9vAoZggAtJYQiFchQ4F9sBC+rFBkgbMs-HYqNhPD5kdM4HMFlYEcMuDAaJojiczkgAEyEnTYVDuADCECo8P8ggArLlvHYACrEhSIlYgBq+ACSUDCsEMYH2fAAguLDDBVOgloKdFEakSesgACxcdVxACi4uQAEZzgBfC1AA)

```sh
--parser flow
--range-start 4
--range-end 18
```

**Input:**

```jsx
let i    ="format me!"   ;
```

**Output:**

```jsx
let i = "format me!";   ;
```

**Expected behavior:**
There should not be an extra semicolon in the output. The output should be either 
```jsx
let i = "format me!";
```
or 
```jsx
let i = "format me!"   ;
```

Harry-Hopkinson:
@rotu fixed in this PR #12542

fisker:
Problem cause by https://github.com/prettier/prettier/blob/0dcb0a1a308e7d751f5f45f54847a90b57009e7a/src/language-js/parse/postprocess/index.js#L74-L81, I'll try to remove that to see what breaks.

fisker:
**Prettier pr-12563**
[...]

PR #12542, opened 2022-03-26
Fix for Prettier adds redundant semicolon when formatting javascript #12540

Harry-Hopkinson:
## Description

Fix for Prettier adds redundant semicolon when formatting javascript #12540.

Before Prettier added an unecessary semi colon with badly formatted code. This PR fixes this issue.

```javascript
let i    ="format me!"   ;
```

This code is now ouputted as...
```javascript
let i = "format me!";
```

Instead of the previous...
```javascript
let i = "format me!";   ;
```

## Checklist

### DIFF
--- a/src/language-js/print/statement.js
+++ b/src/language-js/print/statement.js
@@ -35,7 +35,7 @@ function printStatementSequence(path, options, print, property) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


fisker:
You didn't understand that issue at all!

fisker:
I don't think you can fix that, it's a hard one.

Issue #12553, opened 2022-03-28
add parentesis on a expression with several compare operators


yukulele:
**Prettier 2.6.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDABAXk+gRlnMAbgB0oAGdAHnQDdUAbAVzmvQEZzyiQAaECAAcYAS2gBnZKFQAnGRADuABVkJJKRgtQBPSf1wzUYANZwYAZVQBbOABkRUOMgBmjcXH2GTZ84KMOAc2QYGRZ+OCtcOAATaJjbVCgAplQAuAAxCBkrVBhRJOQ0JhgIPhAACxgrBgB1cpF4cT8wOHM1BpFaBu1CsHE9EAd3GRglQwCclzcPEAArcQAPc0CGOABFJgh4KYZ3fj8ZYcLcVCiGMsEZBxgakWiYcuQADnJ9+Xcaw0FCy7hh2ic-AAjpt4GMhOo0OIALSOGIxMoyOAgkRIsapSZIVy7GbuKwiYKhXErdagpxY6b8GCnW73R5IABMVMMIgYgQAwhArJiQH8AKxlJjuAAqp3U2L2IFoLAAklA4rBzGArsIAILy8wwbSrHbuAC+eqAA)

```sh
--parser babel
```

**Input:**

```jsx
a == b == c;
0 < value < 100;

```

**Output:**

```jsx
(a == b) == c;
0 < value < 100;

```

it may suggest that this code is equivalent to
```jsx
0 < value && value < 100
```

**Expected behavior:**
```jsx
(a == b) == c;
(0 < value) < 100;

```

PR #12564, opened 2022-03-31
Draft: Parenthesize compare sequence

niklasvatn:
## Description

I have made changes in `shouldFlatten` to sequences of compare expressions parenthesized like in #12553. 

## Checklist

As of now, it is not working properly. The test that I added fails when the leftmost operator is less/greater than or equal and the right is less/greater than. My knowledge of the test infrastructure is not good enough to understand what goes wrong there.

Also, some flow code seems to be broken by this and if I am to continue on this I would appreciate some suggestions that I could make to fix this.


tests/format/flow/typeapp-call/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -27,18 +27,18 @@ f?.<T>(e);
 =====================================output=====================================
 //@flow
 f<T>();
-f < T > <U></U>;
+(f < T) > <U></U>;
```
niklasvatn, line 30: This is an issue, does someone have a suggestion on how to distinguish between this scenario and a sequence of comparison?

tests/format/misc/errors/js/async-await/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -15,7 +15,7 @@ exports[`snippet: #0 [acorn] format 2`] = `
 `;
 
 exports[`snippet: #0 [babel] format 1`] = `
-"'await' is only allowed within async functions and at the top levels of modules. (2:20)
+"'await' is not allowed in async function parameters. (2:20)
```
niklasvatn, line 18: These diffs look a bit odd to me.

PR #12578, opened 2022-04-01
Increase the image size

Harry-Hopkinson:
## Description

Increased the State of JS 2021 Image size.

Before:

before

After:

after

## Checklist

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -293,7 +293,11 @@ const UsersSection = ({ language }) => {
             >
               <img
                 src="/images/state_of_js_grey.svg"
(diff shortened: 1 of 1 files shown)


fisker:
So, no.

fisker:
Screenshot_2022-04-01-22-33-23-817_com microsoft emmx canary

So, no.

Harry-Hopkinson:
It is larger than the counter part using phone dimensions though after doing tests on my local machine.

Before:
before

After:
after

fisker:
Sorry, I'm not going to change it.

PR #12599, opened 2022-04-06
Use yarn 3

fisker:
## Checklist


kachkaev:
Do we need to keep multiple `yarn.lock` files? Yarn 3 has an improved ‘focus mode’ for workspaces, if I remember correctly. In theory, there can be just one `yarn.lock` at the top of the repo.

fisker:
> Do we need to keep multiple yarn.lock files?

I think we do, we run release script from `scripts/release` and during the release it removes the one in root. But I'm not sure.

fisker:
> Before we decide to use Yarn 3 in the main branch

Unfortunately, the CI log is telling us we can't, it seems yarn 3 don't support Node.js 10.

kachkaev:
> Yarn 3 don't support Node.js 10

Yeah, that's a shame. We can probably finish this PR after Prettier v3 is out then. We'll have an opportunity to drop Node 10 and 12 I guess.

PR #12612, opened 2022-04-07
chore: disable string escaping in snapshots

SimenB:
## Description

Makes the snapshots more readable. We'll probably change this default in Jest 29, but since you already use `jest-snapshot-serializer-raw` in this repo, I'm thinking this is something you want now?

## Checklist


jest.config.js
```diff
@@ -58,6 +58,9 @@ module.exports = {
     "jest-snapshot-serializer-raw",
     "jest-snapshot-serializer-ansi",
   ],
+  snapshotFormat: {
+    escapeString: false,
+  },
```
SimenB, line 63: this is the only change outside of `yarn test -u`

fisker:
Wow, much better, just to confirm, the `` ` `` character still need escape, right?

SimenB:
Yes, the entire snapshot string is wrapped in backticks, so any `` ` `` within that needs to be escaped

fisker (approved):

sosukesuzuki (approved):
Thank you for doing this!

PR #12626, opened 2022-04-08
Please describe what prettier ignores by default

stopsopa:
## Checklist

### DIFF
--- a/docs/ignore.md
+++ b/docs/ignore.md
@@ -24,9 +24,9 @@ coverage
 
 It’s recommended to have a `.prettierignore` in your project! This way you can run `prettier --write .` to make sure that everything is formatted (without mangling files you don’t want, or choking on generated files). And – your editor will know which files _not_ to format!
 
-(See also the [`--ignore-path` CLI option](cli.md#--ignore-path).)
+By default prettier ignores files in version control systems directories (".git", ".svn" and ".hg") and `node_modules` (if [`--with-node-modules` CLI option](cli.md#--with-node-modules) not specified)
 
-By default pettier ignores what it will find in any .gitignore it finds in the current repository
+(See also the [`--ignore-path` CLI option](cli.md#--ignore-path).)
 
 ## JavaScript
 
--- a/docs/ignore.md
+++ b/docs/ignore.md
@@ -26,6 +26,25 @@ It’s recommended to have a `.prettierignore` in your project! This way you can
 
 By default prettier ignores files in version control systems directories (".git", ".svn" and ".hg") and `node_modules` (if [`--with-node-modules` CLI option](cli.md#--with-node-modules) not specified)
 
+So by defualt it will be
+
+```js
+**/.git
+**/.svn
+**/.hg
+**/node_modules
+```
+
+and
+
+```js
+**/.git
+**/.svn
+**/.hg
+```
+
+if [`--with-node-modules` CLI option](cli.md#--with-node-modules) provided
+
 (See also the [`--ignore-path` CLI option](cli.md#--ignore-path).)
 
 ## JavaScript
--- a/docs/ignore.md
+++ b/docs/ignore.md
@@ -26,6 +26,8 @@ It’s recommended to have a `.prettierignore` in your project! This way you can
 
 (See also the [`--ignore-path` CLI option](cli.md#--ignore-path).)
 
+By default pettier ignores what it will find in any .gitignore it finds in the current repository
+
 ## JavaScript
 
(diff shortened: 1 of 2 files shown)


docs/ignore.md
```diff
@@ -26,6 +26,8 @@ It’s recommended to have a `.prettierignore` in your project! This way you can
 
 (See also the [`--ignore-path` CLI option](cli.md#--ignore-path).)
 
+By default pettier ignores what it will find in any .gitignore it finds in the current repository
```
fisker, line 29: I don't think it's true, we don't read `.gitignore` file.

stopsopa: That's what I thought that it looks for .gitignore files after I did quick dive into the code using github search.
I've stumbled across this DEFAULT_IGNORE in https://github.com/prettier/prettier/blob/529312e9acf8f48bebe67f7526cb629665b5a4f3/vendors/tempy.js#L5696

I might be wrong though... I don't have full day for that search

What I'm trying to highlight here is that I think it's natural question to ask "If prettier ignores some files by default". Because it does seem to not search files in .git or node_modules directories. So naturally you might ask yourself "Is there anything else it's ignoring out of the box?"

Just document that somehow please. What is the default ignoring rules if there are any?

fisker: https://github.com/prettier/prettier/blob/cb997ca7abac937899bede96c3ff87aac16abc83/src/cli/expand-patterns.js#L48-L51

sd-godaddy: Ok, there it is, thank you.
Sooo... Can't it be documented? Just mentioned that this is the case, that it ignores those version control directories and node_module when --with-node-modules is not specified?

It's not difficult to miss the fact that there is flag like --with-node-modules
and even when you find it in cli prettier --help it's doesn't carry information about ignoring .git, .svn and .hg...

fisker: Sure!

sd-godaddy: @fisker check PR now please

fisker (approved):
LGTM, thanks!

sosukesuzuki (approved):

fisker:
@stopsopa @sd-godaddy (Twins?)

Can you copy changes to this directory?https://github.com/prettier/prettier/tree/main/website/versioned_docs/version-stable That's the website uses.

sd-godaddy:
Check now please

sd-godaddy:
Hey, If that was approved maybe it's worth merging then, why wait? waiting might cause future conflicts on merge due to other future changes in that region of doc...

docs/ignore.md
```diff
@@ -24,6 +24,27 @@ coverage
 
 It’s recommended to have a `.prettierignore` in your project! This way you can run `prettier --write .` to make sure that everything is formatted (without mangling files you don’t want, or choking on generated files). And – your editor will know which files _not_ to format!
 
+By default prettier ignores files in version control systems directories (".git", ".svn" and ".hg") and `node_modules` (if [`--with-node-modules` CLI option](cli.md#--with-node-modules) not specified)
+
+So by default it will be
+
+```js
```
kachkaev, line 31: @fisker we can change that to ` ```txt` when there’s an opportunity. Not worth a separate PR I guess. [Current render](https://prettier.io/docs/en/ignore.html):

Screenshot 2022-04-12 at 11 30 13

docs/ignore.md
```diff
@@ -24,6 +24,27 @@ coverage
 
 It’s recommended to have a `.prettierignore` in your project! This way you can run `prettier --write .` to make sure that everything is formatted (without mangling files you don’t want, or choking on generated files). And – your editor will know which files _not_ to format!
 
+By default prettier ignores files in version control systems directories (".git", ".svn" and ".hg") and `node_modules` (if [`--with-node-modules` CLI option](cli.md#--with-node-modules) not specified)
```
kachkaev, line 27: prettier → Prettier (again, probably not worth a separate PR)

docs/ignore.md
```diff
@@ -24,6 +24,27 @@ coverage
...
+
+and
+
+```js
+**/.git
+**/.svn
+**/.hg
+```
+
+if [`--with-node-modules` CLI option](cli.md#--with-node-modules) provided
```
kachkaev, line 46: Missing full stop at the end of the sentence. It might be easier to read this statement if we put this line before the second block:

````txt
and if [`--with-node-modules` CLI option](cli.md#--with-node-modules) is provided:

```txt
...
```
````

kachkaev:
Yeah sorry for seeing this PR a bit too late 😅

I’m a bit short on time these days, but if you want to tweak the wording, happy to review a PR!

PR #12652, opened 2022-04-14
Remove unnecessary folder in testing.

Harry-Hopkinson:
## Description

Before there were unnecessary folders for testing which complicated the testing process. Removing these folders makes testing easier.


Harry-Hopkinson:
I changed lwc/lwc/test files to lwc/test files and misc/misc/test files to misc/test files.

fisker:
Sorry, won't accept. We don't put things in the language root, because we'll need add seperate dirs.

PR #12656, opened 2022-04-16
docs(pre-commit): Use the new syntax

tony:
## Description

The existing doc example raises a warning with precommit:

```
[WARNING] normalizing pre-commit configuration to a top-level map.  support for top level list will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
[WARNING] The 'rev' field of repo 'https://github.com/pre-commit/mirrors-prettier' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.
See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details. 
Hint: `pre-commit autoupdate` often fixes this.
prettier.................................................................Failed
```

- No more top-level list, it uses the `repo` at the root
- Always specify version in the rev

## Checklist

### DIFF
--- a/docs/precommit.md
+++ b/docs/precommit.md
@@ -53,8 +53,9 @@ Read more at the [pretty-quick](https://github.com/azz/pretty-quick) repo.
 Copy the following config into your `.pre-commit-config.yaml` file:
 
 ```yaml
+repos:
 - repo: https://github.com/pre-commit/mirrors-prettier
-  rev: "" # Use the sha or tag you want to point at
+  rev: "v2.6.2" # Use the sha or tag you want to point at
   hooks:
     - id: prettier
 ```


fisker:
[Here](https://github.com/pre-commit/mirrors-prettier#using-prettier-with-pre-commit) still shows we should use the old way.

tony:
> [Here](https://github.com/pre-commit/mirrors-prettier#using-prettier-with-pre-commit) still shows we should use the old way.

@fisker To clarify, what do you propose? I make a PR for that repo also or we close this?

I think that the new format pre-commit wants uses the format in this PR.

fisker:
>  I make a PR for that repo also or we close this?

Make a  PR for that repo first.

tony:
@fisker PR'd at https://github.com/pre-commit/mirrors-prettier/pull/21

tony:
Per https://github.com/pre-commit/mirrors-prettier/pull/21 being closed I'm closing this.

Issue #3460, opened 2017-12-12
prettierIgnore key in package.json?


kylesimmonds:
ESlint currently allows ignore options to be set in `package.json` under the `eslintIgnore` key: https://eslint.org/docs/user-guide/configuring#using-eslintignore-in-packagejson

I noticed there's no mention of a `prettierIgnore` option in the docs (https://prettier.io/docs/en/ignore.html). I also tried adding a similar key for prettier like this, which did nothing:

```
"prettierIgnore": ["build"],
```

At this point I'm pretty sure this isn't possible currently, but I wanted to make sure. Are there plans to add this in or is this intentionally left out?

azz:
It's not currently implemented. We don't currently look at the package.json at all. We use cosmiconfig for configuration which is separate to the ignore file handling.

lipis:
@azz that is not true. You can add the `"prettier"` key in the `package.json` with all the options there.

duailibe:
@lipis Yeah that’s the “we use cosmicconfig” part, but we don’t look the package.json for ignore.. those are two completely separate flows

lipis:
@duailibe I see :)

azz:
What's the incentive for putting this in package.json anyway?

kylesimmonds:
@azz the only reason I can think of is keeping the config format options as close to ESlint as possible because many people use it alongside ESlint and would expect the options to be the same. I could be wrong but not allowing `prettierIgnore` is the only difference from where you can add ESlint settings.

Maybe more people prefer using files to package.json keys (`.eslintrc`, `.eslintignore`, `.prettierrc`, `.prettierignore`) but it seems incomplete that the only thing missing is allowing `prettierIgnore` in package.json

azz:
I don't find "because ESLint does it this way" a particularly compelling reason. I think people like to put all kinds of things in their package.json files that don't really belong, and this seems like one of those things.

kylesimmonds:
@azz you're right that's not a compelling reason, but assuming that it shouldn't belong in package.json is a pretty sweeping dismissal of another dev's potential needs. By that reasoning using package.json shouldn't be an option for `"prettier"` either. And seeing as this is the only config missing in package.json it seems more like an accident than an informed decision.

If you don't want to do it, that's fine, I won't lose sleep over an extra file. But don't attribute it to wanting to keep package.json clean

[...]

PR #12672, opened 2022-04-20
Add support for `prettierIgnore` key in `package.json`

mataha:
## Description

Most of what I've wanted to write about had already been discussed in #3460; to keep it brief:

`package.json` is used more and more to store configuration for various development tools (including their gitignore-like files, if any) in lieu of dotfiles. Prettier provides a way to do that for its options, but not for the list of files to ignore itself.

This PR tries to amend that. Implementation is similar to ESLint - via `prettierIgnore` key.

## Checklist


mataha:
Anyone?

Trim21:
any progress 👀?

src/common/create-ignorer.js
```diff
@@ -21,9 +31,20 @@ async function createIgnorer(ignorePath, withNodeModules) {
...
 createIgnorer.sync = function (ignorePath, withNodeModules) {
-  const ignoreContent = !ignorePath
+  let ignoreContent = !ignorePath
     ? null
     : getFileContentOrNull.sync(path.resolve(ignorePath));
+
+  if (!ignoreContent) {
+    try {
+      ignoreContent = JSON.parse(
+        getFileContentOrNull.sync(path.resolve("package.json"))
```
fisker, line 41: I think we should search `package.json` from the file not `cwd`.

mataha: Is something like [this](https://github.com/prettier/prettier/pull/12672/commits/7256153fb0d9bdfd609de8e87ec3111334c290e1) fine?

fisker: No, not the ignore file, the formatting file.

fisker: Maybe it should be done in `getFileInfo` function.

mataha: Okay, how about [this](https://github.com/prettier/prettier/pull/12672/commits/3b0bd49b4b54ef23db2a85a58c8a9ba3db0508c9) then? I have to admit it's pretty ugly though.

src/common/get-key-content-or-null.js
```diff
@@ -0,0 +1,79 @@
+"use strict";
```
fisker, line 1: Maybe we use cosmiconfig to search for `prettierIgnore`?

```js
cosmiconfig("prettierIgnore", {searchPlaces: ["package.json"]})
```

src/cli/format.js
```diff
@@ -227,8 +227,13 @@ function format(context, input, opt) {
 }
 
 async function createIgnorerFromContextOrDie(context) {
+  const filepath = context.argv.filepath
```
fisker, line 230: Where this filepath come from? There is no `filepath` flag.

mataha: There is - as [`--stdin-filepath`](https://github.com/prettier/prettier/blob/fa3b95c2e1a3a3351dc1b0168aa2c26f0cc4e9b3/src/main/core-options.js#L97-L106) - and it's used [here](https://github.com/prettier/prettier/blob/fdccc0f48e5bd3a17ae1e3500c1b5c1763c6a368/src/cli/index.js#L99) as well.

src/common/get-key-content-or-null.js
mataha: My biggest gripe about this is that cosmiconfig is suited for JSON and YAML configuration files, not lists of file globs with or without negations (which ignore files are).

I will try it, however.

CiiDyR:
I'd love to see this land. It would allow us to have pretterignore settings shared across projects. As we can already share settings, but not ignored files (https://prettier.io/docs/en/configuration.html#sharing-configurations)

Is there anything we could contribute to accelerate this?

fullstackjedi:
I would suggest adding this as an option *within* the configuration file (i.e. `.prettierrc.json`). That way, it can be included in shareable configs.

nickmccurdy:
I don't think sharing ignored files would be a good practice, as ignored files tend to vary more based on a project's tooling usage than its style guide, and we wouldn't want to couple something unrelated to style to a config. For reference, ESLint shareable configs also don't allow sharing ignores. If you do have several projects with the same tooling, you can simply copy the `.prettierignore` (or with this, the `package.json`).

fullstackjedi:
> For reference, ESLint shareable configs also don't allow sharing ignores.

Yes, [they do](https://eslint.org/docs/latest/user-guide/configuring/ignoring-code).

src/common/get-key-content-or-null.js
mataha: I've changed the implementation - performance is almost the same compared to the previous method while it's a bit nicer on the eyes (I think).

mataha:
> I would suggest adding this as an option _within_ the configuration file (i.e. `.prettierrc.json`). That way, it can be included in shareable configs.

That's a different feature for a different PR.

src/common/create-ignorer.js
```diff
@@ -2,40 +2,80 @@
...
 /**
+ * @param {string} filePath
  * @param {string?} ignorePath
  * @param {boolean?} withNodeModules
  */
-createIgnorer.sync = function (ignorePath, withNodeModules) {
-  const ignoreContent = !ignorePath
-    ? null
-    : getFileContentOrNull.sync(path.resolve(ignorePath));
+createIgnorer.sync = function (filePath, ignorePath, withNodeModules) {
```
fisker, line 28: Let's change base to `next` branch, so we don't need support sync version.

mataha: Done.

ivodolenc:
It might not be a bad idea to implement this feature directly in the `prettier.config` file.

So when we want to share the settings we have all the options inside one file. This is exactly what `eslint` have.

For example, I can specify in `package.json` custom path for `eslint.config.js` and `prettier.config.js` and clean up the root as much as possible (no need for additional ignore files and custom ignore options inside package.json):

```json
{
  "name": "package.json",
  "type": "module",

  "prettier": "./.config/prettier.config.cjs",
  "eslintConfig": {
    "extends": "./.config/eslint.config.cjs"
  }
}
```

and in one file I have all settings, including `ignorePatterns` like this:

```js
// .config/eslint.config.js

module.exports = {
  root: true,

  extends: [],

  plugins: [],

  ignorePatterns: [
    '.DS_Store',
    'node_modules'
    // ...
  ]
}
```

This would be a huge win to put `ignore` directly inside `prettier` config:

```js
// .config/prettier.config.js

module.exports = {
  singleQuote: true,
  trailingComma: 'none',
  
  plugins: [],

  ignorePatterns: [
    '.DS_Store',
    'node_modules'
    // ...
  ]
}
```

What do you think?

src/cli/format.js
fisker: `filepath` only exists when format stdin, when formatting files, it doesn't exist.

fisker: See `formatFiles` function, it calls `createIgnorerFromContextOrDie` on first line. I don't think it will work.

kachkaev:
@mataha would you be interested to continue this PR? @ivodolenc's [suggestion](https://github.com/prettier/prettier/pull/12672#issuecomment-1256947760) makes sense – it' be great to see alignment between ESLint and Prettier!

aensley:
I second @ivodolenc's suggestion: https://github.com/prettier/prettier/pull/12672#issuecomment-1256947760

It would be really great to be able to specify `ignorePatterns` in the prettier config.

Standard8:
Note, as this appears to have stalled out, I've started working on adding the options to the configuration. As a first step, I've added command line arguments to echo ESLint - as they were quite easy to do, and I've just posted that in #14223

wallpants:
Is there anything I can do to help get this over the finish line? In a world where we already have a ton of config files in the root of our projects, I'm struggling to understand why we can't specify patterns to ignore in our prettier config file and we're forced to create a new file just for that.

kevinwolfcr:
> Is there anything I can do to help get this over the finish line? In a world where we already have a ton of config files in the root of our projects, I'm struggling to understand why we can't specify patterns to ignore in our prettier config file and we're forced to create a new file just for that.

I think they don't like listening to their users :)

Standard8:
> Is there anything I can do to help get this over the finish line? In a world where we already have a ton of config files in the root of our projects, I'm struggling to understand why we can't specify patterns to ignore in our prettier config file and we're forced to create a new file just for that.

As discussed elsewhere, I think it would be much better to add an option in the `.prettierrc.js` file rather than `package.json`, this would also align it better with ESLint (which I believe was one of the aims).

I tried [doing this previously](https://github.com/prettier/prettier/pull/14223#issuecomment-1401827265) but hit some issues, and despite asking never got any feedback from the developers, even when pinging the devs directly, so I gave up.

Feel free to take my work and try moving it forward though, maybe you'll have more success.

mataha:
I'm in a position in which I can commit some time to this again. I'd like to clarify some things beforehand though.

> @mataha would you be interested to continue this PR? @ivodolenc's [suggestion to use `ignorePatterns`](https://github.com/prettier/prettier/pull/12672#issuecomment-1256947760) makes sense – it' be great to see more alignment between ESLint and Prettier!
> 
> 📖 [Docs on `ingorePatterns` in ESLint config](https://eslint.org/docs/latest/user-guide/configuring/ignoring-code#ignorepatterns-in-config-files)

Again, I believe that's a different feature for a different PR. The core tenet of this is that one can share `prettier` configuration between different projects or even monorepo packages, yet be able to selectively choose which files should be formatted at the project level, e.g. plumbing for package managers, test specifications, generated files... As someone here said, ignorable files tend to be influenced by a project's tooling more than anything - two entities sharing such config might not necessarily use the same package manager, among other things. `ignorePatterns` does not achieve this - it's applied on a configuration level, thus applicable to every package using said config.

That said, the main pain point of this PR continues to be how Prettier should behave upon encountering a `package.json`:

  - stop scanning
  - continue scanning until filesystem root, merge all sections
  - continue scanning if section is malformed or missing

I'll have to rewrite this as several things have changed since then (e.g. ignore facility has moved to `src/utils`).

mataha:
Can't run tests at all due to jestjs/jest#7914.

PR #12674, opened 2022-04-20
Run tests on Node.js v18

fisker:
## Checklist

### DIFF
--- a/.github/workflows/dev-test.yml
+++ b/.github/workflows/dev-test.yml
@@ -18,20 +18,21 @@ jobs:
           - "macos-latest"
           - "windows-latest"
         node:
+          - "18"
           - "16"
           - "14"
         include:
           # only enable coverage on the fastest job
           - os: "ubuntu-latest"
-            node: "16"
+            node: "18"
             ENABLE_CODE_COVERAGE: true
             FULL_TEST: true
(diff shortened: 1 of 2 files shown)


kachkaev (approved):
Makes sense, thanks @fisker!

.github/workflows/dev-test.yml
```diff
@@ -18,20 +18,21 @@ jobs:
...
-        # exclude:
-        #   - os: "macos-latest"
-        #     node: "14"
-        #   - os: "windows-latest"
-        #     node: "14"
+        exclude:
+          - os: "macos-latest"
+            node: "16"
+          - os: "windows-latest"
+            node: "16"
```
kachkaev, line 35: I assume the following thinking process: if it passes on Node 18 and Node 14, it would pass on Node 16 too anyway. I agree with this shortcut as it saves some CI time.

fisker: You are right, we do the same on production test https://github.com/prettier/prettier/blob/dca46b71a7fd4c8bf1f7aac43569012537cbc1b1/.github/workflows/prod-test.yml#L79-L91

sosukesuzuki (approved):

PR #12748, opened 2022-04-29
Support async parser

fisker:
## Checklist


fisker:
Only problem here: I can't simply make `parse` return a Promise, embed parser have to be sync.

Options:

1. Don't support async parser in embed format.
2. Support async `printer.print()` (maybe jest  `printer.embed()` function is enough ), I was not going to support async printer, there are some known issue in docs

Personally I prefer `1`.

@kachkaev @sosukesuzuki  WDYT?

PR #12756, opened 2022-04-30
Provide an example `.editorconfig`

j-f1:
## Description

This PR adds an example `.editorconfig` file to the documentation that makes it easy to align your editor to Prettier’s behavior. [See the deploy preview here](https://deploy-preview-12756--prettier.netlify.app/docs/en/next/configuration.html#editorconfig)

## Checklist

- (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -176,9 +176,35 @@ If you’d like a JSON schema to validate your configuration, one is available h
 
 ## EditorConfig
 
-If `options.editorconfig` is `true` and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc. Currently, the following EditorConfig properties are supported:
+If `options.editorconfig` is `true` and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc.
 
-- `end_of_line`
-- `indent_style`
-- `indent_size`/`tab_width`
-- `max_line_length`
+Here’s an annotated description of how different properties map to Prettier’s behavior:
+
+```toml
+[*]
+# Non-configurable Prettier behaviors
+charset = "utf-8"
+insert_final_newline = true
+# Caveat: Prettier won’t trim trailing whitespace inside template strings, but your editor might.
+trim_trailing_whitespace = true
+
+# Configurable Prettier behaviors
+# (change these if your Prettier config differs)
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -176,9 +176,34 @@ If you’d like a JSON schema to validate your configuration, one is available h
...
-If `options.editorconfig` is `true` and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc. Currently, the following EditorConfig properties are supported:
+If `options.editorconfig` is `true` and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc.
 
-- `end_of_line`
-- `indent_style`
-- `indent_size`/`tab_width`
-- `max_line_length`
+Here’s an annotated description of how different properties map to Prettier’s behavior:
+
+```toml
```
fisker, line 183: I thought .editorconfig is `ini` file.

sosukesuzuki (approved):

docs/configuration.md
```diff
@@ -176,9 +176,34 @@ If you’d like a JSON schema to validate your configuration, one is available h
...
+
+Here’s a copy+paste-ready `.editorconfig` file if you use the default options:
 
-- `end_of_line`
-- `indent_style`
-- `indent_size`/`tab_width`
-- `max_line_length`
+```ini
+[*]
+charset = "utf-8"
```
fisker, line 203: Remove quots? We didn't use it https://github.com/prettier/prettier/blob/9348049311ec1e25ce83b9c03ec32ec14a9751b7/.editorconfig#L4

docs/configuration.md
```diff
@@ -176,9 +176,34 @@ If you’d like a JSON schema to validate your configuration, one is available h
...
+insert_final_newline = true
+end_of_line = "lf"
+indent_style = "space"
+indent_size = 2
+max_line_length = 80
+```
```
fisker, line 209: Should we add this https://github.com/prettier/prettier/blob/9348049311ec1e25ce83b9c03ec32ec14a9751b7/.editorconfig#L1?

j-f1: I don’t think it should be on by default, in case the user wants to set e.g. `trim_trailing_whitespace = true` in their `~/.editorconfig`, we don’t want to stop the editor for looking for parent files

fisker (approved):

PR #12763, opened 2022-05-01
Try searching for manually specified plugins in pluginSearchDir's

zachkirsch:
## Description

If you specify a `--plugin` that lives outside the current tree (i.e. not resolvable by `require.resolve`), the CLI throws with a MODULE_NOT_FOUND error. Locations provided via `--plugin-search-dir` are _not_ searched.

**After this PR:** Locations provided via `--plugin-search-dir` are searched when resolving manually specified plugins.

## Checklist

### DIFF
--- a/src/common/load-plugins.js
+++ b/src/common/load-plugins.js
@@ -48,8 +48,17 @@ function load(plugins, pluginSearchDirs) {
         // try local files
         requirePath = resolve(path.resolve(process.cwd(), pluginName));
       } catch {
-        // try node modules
-        requirePath = resolve(pluginName, { paths: [process.cwd()] });
+        try {
+          // try node modules
+          requirePath = resolve(pluginName, { paths: [process.cwd()] });
+        } catch {
+          // try provided plugin search dirs
+          requirePath = resolve(pluginName, {
+            paths: pluginSearchDirs.map((dir) =>
+              path.resolve(process.cwd(), dir)
+            ),
+          });
+        }
       }
 
       return {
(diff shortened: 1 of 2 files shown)


zachkirsch:
Thanks for taking a look @fisker and @kachkaev! While `pluginSearchDirs` and `plugins` don't require each other, this PR is addressing the case where you choose to use both.

In the link you sent, the first example is

```
prettier --write main.foo --plugin-search-dir=./dir-with-plugins --plugin=prettier-plugin-foo
```

i.e. using a plugin `prettier-plugin-foo` that exists in `./dir-with-plugins`. 

However, if you instead specified a path that doesn't exist inside the current workspace, then prettier fails:
```
prettier --write main.foo --plugin-search-dir=../../../../dir-with-plugins --plugin=prettier-plugin-foo
```

If you checkout the first commit on this branch https://github.com/prettier/prettier/pull/12763/commits/9f8329d60b1b5f79008c9c63d011414dde30bee2 (it includes the test but not the fix) you can see an example

fisker:
You can do `prettier --plugin=../../../../dir-with-plugins/prettier-plugin-foo/index.js`

zachkirsch:
Yep that’s true! Just kind of cumbersome if I have a lot of plugins that live in that directory.

Would you rather this PR not merge? I thought I was fixing a bug but it sounds like the existing behavior is preferable?

fisker:
Close sinc the plugin search feature has been removed.

PR #12791, opened 2022-05-05
Simplify `loadEditorConfig()`

fisker:
## Checklist


sosukesuzuki (commented):
I'm not familiar with tests for `loadEditorConfig`. But I concern codecov comments. Can we ignore those comments?

fisker:
I think it's fine to ignore them for now, everyone uses .editorconfig, if there are problems, we'll know. There are too much work in V3, don't have much time for this, we can add tests in future.

PR #12802, opened 2022-05-07
Simplify `getFileInfo()`

fisker:
## Checklist

### DIFF
--- a/src/common/create-ignorer.js
+++ b/src/common/create-ignorer.js
@@ -1,20 +1,23 @@
 import path from "node:path";
 import ignore from "ignore";
-import getFileContentOrNull from "../utils/get-file-content-or-null.js";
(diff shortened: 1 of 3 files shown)


sosukesuzuki (commented):
Looks good to me. However, shouldn't we not ignore codecov comments?

fisker:
It's not covered only because we disabled tests, https://github.com/prettier/prettier/blob/0946b636a66524d1d16f7996fd01bde2fee9c2d6/package.json#L139, just ignore the coverage.

Reason: https://github.com/prettier/prettier/pull/11903#issuecomment-1034642807

I was hoping we can fix it after https://github.com/prettier/prettier/pull/12705#issue-1215553456

sosukesuzuki (approved):

PR #12866, opened 2022-05-14
Unused-imports for the .eslintrc.js.

Harry-Hopkinson:
## Description

Use the no-unused-imports package for errors when importing unused packages. This eslint package also brings in a no-unused-vars option which identifies unused variables with a higher accuracy than the former option of the built in no-unused-vars option with eslint.

Fixed files in order to comply with the .eslintrc.js rules.

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,11 +18,13 @@ module.exports = {
     "regexp",
     "unicorn",
     "@typescript-eslint",
+    "unused-imports",
   ],
   settings: {
     "import/internal-regex": "^linguist-languages/",
   },
   rules: {
+    "unused-imports/no-unused-imports": "error",
     "@typescript-eslint/prefer-ts-expect-error": "error",
     "arrow-body-style": ["error", "as-needed"],
     curly: "error",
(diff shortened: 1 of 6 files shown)


sosukesuzuki (approved):

fisker:
I can't see how this is rule better.

Harry-Hopkinson:
It generally detects unused variables and imports better then the default options.

fisker:
Any example?

Harry-Hopkinson:
It detected the following example which the default option did not detect and checks for unused imports - even though there are no unused imports in this codebase.

```
const { children, copy, ...rest } = this.props;
```
Where copy is an unused variable.

fisker:
That's excatly the case I'm against this rule. In the case above. `rest` is excluding `children` and `copy`. Removing `copy` is adding an extra property to `rest`. Idealy it should be

```js
const { children, copy: _copy, ...rest } = this.props;
```

but I don't think it's nessary. And we use this pattern in other places https://github.com/prettier/prettier/blob/36c20c51bc4244d185c5e21f49c4fceecf3de5ca/tests/config/utils/create-snapshot.js#L23-L31

Harry-Hopkinson:
We could use the previous ```no-unused-vars``` rule and just use ```"unused-imports/no-unused-imports"```.

fisker:
I don't think we need this rule.

fisker:
Close this since I don't think we want this.

Issue #12922, opened 2022-05-24
--find-ignore-path option


Zamiell:
Prettier currently has:
- `--config` - Specifies the path to the intended configuration file.
- `--ignore-path` - Specifies the path to the intended ignore file.
- `--find-config-path` - Prints the current config file that is used.

However, it is missing `--find-ignore-path`.

I propose that this should be added.

Zamiell:
Counter: Having this would have made diagnosing prettier functionality / bugs much less painful. See #12923.

fisker:
We can add this to debug info, and debug with `--loglevel`.

fisker:
Or, maybe print in `--file-info`?

Harry-Hopkinson:
I think the command is useful if you have a monorepo for example they may use multiple .prettierignore's. This can help with debugging purposes. If that command is implemented I will use it a lot.

Zamiell:
fisker can you unclose this issue?

PR #13048, opened 2022-06-26
Find ignore path command

Harry-Hopkinson:
## Description

A CLI Option that finds the .prettierignore path of a file as requested in #12922.

e.g

```shell
prettier --find-ignore-path ./README.md outputs .prettierignore.
```

The command works I can not implement a test for it due to the Github Actions runners having a different path than my local machine.

## Checklist


fisker:
We havn't decide to accept #12922

PR #13051, opened 2022-06-27
Registers axml file as HTML

mountainmoon:
This is a HTML-like file registered by the vscode extension alipay.minicode. The axml file is used in alipay miniprogram which is a popular web framework in China.

I followed [this doc](https://github.com/prettier/prettier-vscode/wiki/Custom-File-Extension-Configuration#custom-language-identifier) to send the PR for making prettier-vscode support formating axml files in vscode.

## Checklist

### DIFF
--- a/src/language-html/index.js
+++ b/src/language-html/index.js
@@ -17,7 +17,7 @@ const languages = [
   createLanguage(require("linguist-languages/data/HTML.json"), (data) => ({
     since: "1.15.0",
     parsers: ["html"],
-    vscodeLanguageIds: ["html"],
+    vscodeLanguageIds: ["html", "axml"],
     extensions: [
       ...data.extensions,
       ".mjml", // MJML is considered XML in Linguist but it should be formatted as HTML


fisker:
I don't think `.axml` is HTML compatible, it's more like xml.

mountainmoon:
> 

It looks like xml, but in fact there is one feature in axml not compatible with xml, which is `attribute with out a value`.  You can find more info in [page1, is-an-xml-attribute-without-a-value-valid](https://stackoverflow.com/questions/6926442/is-an-xml-attribute-without-a-value-valid) and [page2, are-attributes-without-value-allowed-in-html4](https://stackoverflow.com/questions/21796217/are-attributes-without-value-allowed-in-html4). Formatted with xml `attribute with out a value` will occur an error like this:
```
["INFO" - 11:02:09 AM] Prettier Options:
{
  "filepath": "/Users/xxxx/index.axml",
  "parser": "xml",
  "singleQuote": true
}
["ERROR" - 11:02:09 AM] Error formatting document.
["ERROR" - 11:02:09 AM] Expecting token of type --> EQUALS <-- but found --> 'style' <--
   7 |     <view>
>  8 |       <swiper disable-touch style="height: 100%;"
     |                             ^^^^
```
It's more like html rather than xml not only in `attribute with out a value` but also in so many HTML-like attributes `onXxxEvent`, `class`, `src`, etc. Also, it's more like Web Components.

fisker:
Good point, but I don't think it's HTML either. I think the best way here is to add this

```js
{
	  overrides: [
		  {
			    files: "**/*.axml",
			    options: {
				    parser: "html",
			    },
		   },
	],
}
```

to your config file (I guess you already did).

mountainmoon:
Yes, I tried it. But like [this doc](https://github.com/prettier/prettier-vscode/wiki/Custom-File-Extension-Configuration) said:

> This will work sometimes in the VS Code extension, but the big caveat here is that VS Code must also have the language registered as a type that Prettier has registered to. Put another way, if VS Code has the file *.abc registered as the 'HTML' language, then the configuration above will work and both the CLI and VS Code will format the file as expected. However, if the language is registered as something else in VS Code like my-custom-language, the VS Code extension will not format this document.

the `.axml` is registered as somehting else in the official alipay miniprogram vscode extension [支付宝小程序助手](https://marketplace.visualstudio.com/items?itemName=alipay.minicode).  And the `File Associations` way in the above doc isn't a good way either for the same reason. So I tried the last way in the doc:

> To get these file types working, you need to send a pull request to either prettier or the prettier plugin to register additional vscodeLanguageIds for the particular parser.

Yes, it isn't HTML. But angular and vue are not exactly HTML either, they are both handled as html in https://github.com/prettier/prettier/blob/main/src/language-html/index.js .

fisker:
Try add to your VSCode settings,

```json
  "[axml]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true
  },
```

Please let me know if it works.

mountainmoon:
I tried, and it worked! But I do remember I have tried this before sending the PR. Maybe it related to some low level spelling errors. Whatever~
Thanks!

fisker:
@ntotten Do you think the linked wiki page is outdated? https://github.com/prettier/prettier-vscode/wiki/Custom-File-Extension-Configuration#custom-language-identifier

PR #13060, opened 2022-07-01
Enable `unicorn/prefer-logical-operator-over-ternary` rule

fisker:
## Description

[prefer-logical-operator-over-ternary](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-logical-operator-over-ternary.md)

## Checklist


kachkaev (approved):
I assume that all ternary conditions cannot be equal to `false` / `0` or `""`. Otherwise, we’d have to use `||` for a full backwards compatibility.

sosukesuzuki (approved):

PR #13091, opened 2022-07-11
Bundle CLI as ES Module

fisker:
## Description

This change should not noticeable to users.

## Checklist


fisker:
Can't believe this mistake takes me one month to fix, https://github.com/prettier/prettier/pull/13091/commits/95b7cc5dbbbf37d9af9752a72e9e608675e7b76b

And this proves running "install package" before test is useful. I skipped installation at first.

Issue #13097, opened 2022-07-12
Incorrect color output when redirecting


dspinellis:
**Environments:**

- Prettier Version: 2.7.1
- Running Prettier via: CLI
- Runtime: Node.js v18.5.0
- Operating System: Reproduced on Windows, Linux

**Steps to reproduce:**
Case 1: Color appearing when it shouldn't
```
bin/prettier.js --check xyzzy.js 2>errors
sed -n l errors
```

**Expected behavior:**
```
[error] No files matching the pattern were found: "xyzzy.js".$
```

**Actual behavior:**
```
[\033[31merror\033[39m] No files matching the pattern were found: "xy\
zzy.js".$
```
(The file contains color-changing escape codes.)

Case 2: Color not appearing when it should
```
bin/prettier.js --check xyzzy.js >/dev/null
```

**Expected behavior:**
The word `error` in red.

**Actual behavior:**
The word `error` without color.

I am uploading a fixing PR.

fisker:
Fixed by  #13098

PR #13098, opened 2022-07-12
Use appropriate chalk instance for stderr output

dspinellis:
Chalk auto-detects whether the output channel supports colors, but does that for stderr.  Consequently, colors appear inconsistently (and wrongly) depending on whether stdout or stderr have been redirected.

Issue: #13097

## Description

This change corrects the instance of chalk used, according to the employed output channel (stdout or stderr).

## Checklist

- [ ] I manually tested the result. I also tried adding tests to confirm my change works in three ways, but none is viable. Specifically, I tried the following:
  - via changes to `run-pretties.js` (its existing `process.stdin.isTTY` code is bogus, because `chalk` calls directly `isatty()`),
  - via mocking (it doesn't work, because `chalk` sets the values at the module level), and 
  - by modifying the `chalk` object with `Object.defineProperty` (it doesn't work, because the `configurable` is set to false).

### DIFF
--- a/src/cli/logger.js
+++ b/src/cli/logger.js
@@ -6,7 +6,7 @@ const wcwidth = require("wcwidth");
 // eslint-disable-next-line no-restricted-modules
 const { default: stripAnsi } = require("../../vendors/strip-ansi.js");
 // eslint-disable-next-line no-restricted-modules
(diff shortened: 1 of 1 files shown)


fisker (approved):
Thanks for working on this!

sosukesuzuki (approved):

PR #13110, opened 2022-07-14
Docs: echo '{}' > .prettierrc.json

devinrhode2:
looks less scary


fisker:
This won't work, check the comment above.

devinrhode2:
I a doof lol 😂

PR #13149, opened 2022-07-22
Throw errors for invalid doc

fisker:
## Description

The printer is now async, it can easily forget to `await` it, should refuse to print if `Promise` inside doc.

## Checklist


kachkaev (approved):

thorn0:
How about making a `getDocType` function, which could be reused in `switch` statements in `fits` and `printDocToString`? The function would return `doc.type` or special values: `string`, `array`, `invalid`. `invalid` is returned if the argument isn't a string, an array, or an object with `type`.

fisker:
I was going to do that when we remove `concat`

sosukesuzuki (approved):
I don't disagree with thorn0, but personally I am happy with this PR!

src/document/printer.js
```diff
@@ -541,7 +541,7 @@ function printDocToString(doc, options) {
         break;
       case DOC_TYPE_LINE_SUFFIX_BOUNDARY:
         if (lineSuffix.length > 0) {
-          cmds.push([ind, mode, { type: "line", hard: true }]);
```
thorn0: It should be hardlineWithoutBreakParent

fisker: Thanks

src/document/printer.js
```diff
@@ -292,283 +317,292 @@ function printDocToString(doc, options) {
...
 
-    if (typeof doc === "string") {
-      const formatted = newLine !== "\n" ? doc.replace(/\n/g, newLine) : doc;
-      out.push(formatted);
-      pos += getStringWidth(formatted);
-    } else if (isConcat(doc)) {
-      const parts = getDocParts(doc);
-      for (let i = parts.length - 1; i >= 0; i--) {
-        cmds.push([ind, mode, parts[i]]);
+    switch (type) {
```
thorn0, line 322: Let's do `switch(getDocType(doc))` instead?

fisker: I was going to use type in default branch, but I don't need that anymore.

thorn0 (approved):
I'm not a fan of constants, but that's subjective.

thorn0:
BTW, we're hitting TypeScript's another limitation here. `getDocType`-like functions can't narrow union types. There is an issue about that: https://github.com/microsoft/TypeScript/issues/30698

Issue #4678, opened 2018-06-12
Preprocessor support


warrenseine:
I’m trying to deal with `#if` / `#endif` preprocessor directives in Prettier C#. I would like to leave them untouched.

The easiest way to do that would be to parse them as comments. The good side is I benefit from the intelligent out-of-the-box comment handling of Prettier. The bad side is I inherit from the intelligent out-of-the-box comment handling of Prettier. Typically, the following code:

```
#if DEBUG
extern alias A;
#endif

using System;
```

would get formatted as:

```
#if DEBUG
extern alias A;

#endif
using System;
```

because #endif would get associated with its following node. In the case of preprocessor directives, the logic is a bit different.

I’m aware that there is a flag to indicate that the plugin takes ownership of the rendering of comments. Is this the way I should go? Or should I stop imagining this workaround will work?

Any better idea? Should there be an API for directives as well as comments, since the preprocessor logic is not uncommon?

warrenseine:
@vjeux I know you implemented comments in the Python plugin, so you might have an interesting opinion here.

j-f1:
What would happen if you interpreted them as statements?

suchipi:
@j-f1 they can appear anywhere, including in expressions

j-f1:
Alright, that makes sense. Here’s an example since it took me a minute to figure out how that could happen:

```c
foo = (
  1
#ifdef BAR
  + 2
#endif
);
```

duailibe:
@warrenseine what if you treat everything from `#ifdef` to `#endif` as a block comment?

warrenseine:
Then, there's the [#pragma or #error](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/) case, which are single lines.

warrenseine:
Ok, I found a way with a dirty (but little) Prettier hack. I'm not sure it would work for all situations but it helps here.

In `comments.js`, there's a logic to consider that a single-line comment should always prefer leading comments when a following node is found ([code](https://github.com/prettier/prettier/blob/master/src/main/comments.js#L222)). I've added something like:

```javascript
if (followingNode && precedingNode && options.printer.preferTrailingComment && options.printer.preferTrailingComment(comment)) {
  addTrailingComment(precedingNode, comment); }
```

where `preferTrailingComment` would be an API function, such as:

```javascript
function preferTrailingComment(comment) {
[...]

PR #13154, opened 2022-07-22
Refactor: simplify `fits`

thorn0:
## Description

`fits` currently contains indentation-related logic, but it's not needed. `fits` deals with the remainder of the line when the indentation has been already printed.

## Checklist

### DIFF
--- a/src/document/doc-printer.js
+++ b/src/document/doc-printer.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-sparse-arrays */
 "use strict";
 
 const { convertEndOfLineToChars } = require("../common/end-of-line.js");
@@ -7,11 +6,16 @@ const getStringWidth = require("../utils/get-string-width.js");
 const { fill, cursor, indent } = require("./doc-builders.js");
 const { isConcat, getDocParts } = require("./doc-utils.js");
 
-/** @type {Record<symbol, typeof MODE_BREAK | typeof MODE_FLAT>} */
+/** @typedef {typeof MODE_BREAK | typeof MODE_FLAT} Mode */
+/** @typedef {{ ind: any, doc: any, mode: Mode }} Command */
+
+/** @type {Record<symbol, Mode>} */
 let groupModeMap;
 
-const MODE_BREAK = 1;
-const MODE_FLAT = 2;
+// prettier-ignore
+const MODE_BREAK = /** @type {const} */ (1);
+// prettier-ignore
+const MODE_FLAT = /** @type {const} */ (2);
 
 function rootIndent() {
   return { value: "", length: 0, queue: [] };
@@ -147,8 +151,17 @@ function trim(out) {
   return trimCount;
 }
 
+/**
+ * @param {Command} next
+ * @param {Command[]} restCommands
+ * @param {number} width
+ * @param {boolean} hasLineSuffix
+ * @param {boolean} [mustBeFlat]
+ * @returns {boolean}
+ */
 function fits(next, restCommands, width, hasLineSuffix, mustBeFlat) {
   let restIdx = restCommands.length;
+  /** @type {Array<Omit<Command, 'ind'>>} */
   const cmds = [next];
   // `out` is only used for width counting because `trim` requires to look
   // backwards for space characters.
@@ -162,7 +175,7 @@ function fits(next, restCommands, width, hasLineSuffix, mustBeFlat) {
       continue;
     }
 
-    const [, mode, doc] = cmds.pop();
+    const { mode, doc } = cmds.pop();
 
     if (typeof doc === "string") {
       out.push(doc);
@@ -170,7 +183,7 @@ function fits(next, restCommands, width, hasLineSuffix, mustBeFlat) {
     } else if (isConcat(doc) || doc.type === "fill") {
       const parts = getDocParts(doc);
       for (let i = parts.length - 1; i >= 0; i--) {
-        cmds.push([, mode, parts[i]]);
+        cmds.push({ mode, doc: parts[i] });
       }
     } else {
       switch (doc.type) {
@@ -178,7 +191,7 @@ function fits(next, restCommands, width, hasLineSuffix, mustBeFlat) {
         case "align":
         case "indent-if-break":
         case "label":
-          cmds.push([, mode, doc.contents]);
+          cmds.push({ mode, doc: doc.contents });
           break;
 
         case "trim":
@@ -195,7 +208,7 @@ function fits(next, restCommands, width, hasLineSuffix, mustBeFlat) {
             doc.expandedStates && groupMode === MODE_BREAK
               ? getLast(doc.expandedStates)
               : doc.contents;
-          cmds.push([, groupMode, contents]);
+          cmds.push({ mode: groupMode, doc: contents });
           break;
         }
 
@@ -206,7 +219,7 @@ function fits(next, restCommands, width, hasLineSuffix, mustBeFlat) {
           const contents =
             groupMode === MODE_BREAK ? doc.breakContents : doc.flatContents;
           if (contents) {
-            cmds.push([, mode, contents]);
+            cmds.push({ mode, doc: contents });
           }
           break;
         }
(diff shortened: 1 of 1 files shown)


src/document/doc-printer.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-sparse-arrays */
```
fisker, line 1: I was told that [`HOLEY` array](https://v8.dev/blog/elements-kinds#packed-vs.-holey-kinds) are slower, and [v8 says we should avoid creating holes](https://v8.dev/blog/elements-kinds#avoid-creating-holes), but it's been a while. It will be nice if you can run some performance tests.

thorn0: I guess we can just change the order of elements in commands, so that `ind` will be last.
BTW, looks like it'll be better if this targets the `next` branch instead of `main`, right?

fisker: > I guess we can just change the order of elements in commands, so that ind will be last.

Is it a good idea to use object instead of array?

> BTW, looks like it'll be better if this PR targets the next branch instead of main, right?

Yes, but that won't be a problem, I can rebase, do whatever you like.

thorn0: >Is it a good idea to use object instead of array?

I think so.

src/document/doc-printer.js
```diff
@@ -6,11 +6,16 @@ const getStringWidth = require("../utils/get-string-width.js");
...
+
+/** @type {Record<symbol, Mode>} */
 let groupModeMap;
 
-const MODE_BREAK = 1;
-const MODE_FLAT = 2;
+// prettier-ignore
+const MODE_BREAK = /** @type {const} */ (1);
+// prettier-ignore
+const MODE_FLAT = /** @type {const} */ (2);
```
thorn0, line 18: Without these `@type {const}` casts, variables like this
```js
const next = { ind, mode: MODE_FLAT, doc: doc.contents };`
```
are inferred as incompatible with the `Command` type.

`// prettier-ignore` is needed because Meriyah doesn't support parenthesized expressions.

fisker (approved):

src/document/doc-printer.js
```diff
@@ -575,7 +552,7 @@ function printDocToString(doc, options) {
     // there is no new line after the line-suffix.
     if (cmds.length === 0 && lineSuffix.length > 0) {
       cmds.push(...lineSuffix.reverse());
-      lineSuffix = [];
+      lineSuffix.length = 0;
```
fisker, line 555:
```
console.time('length0')
for (let i = 10000; i--;) {
    const a = Array.from({length: 100}, () => 0)
    const b = Array.from({length: 10}, () => 1)
    a.push(...b.reverse())
    b.length = 0
}
console.timeEnd('length0')

console.time('splice + reverse')
for (let i = 10000; i--;) {
    const a = Array.from({length: 100}, () => 0)
    const b = Array.from({length: 10}, () => 1)
    a.push(...b.splice(0).reverse())
}
console.timeEnd('splice + reverse')

console.time('reverse + splice')
for (let i = 10000; i--;) {
    const a = Array.from({length: 100}, () => 0)
    const b = Array.from({length: 10}, () => 1)
    a.push(...b.reverse().splice(0))
}
console.timeEnd('reverse + splice')

console.time('while')
for (let i = 10000; i--;) {
    const a = Array.from({length: 100}, () => 0)
    const b = Array.from({length: 10}, () => 1)
    while(b.length) a.push(b.pop())
}
console.timeEnd('while')

console.time('length0')
for (let i = 10000; i--;) {
    const a = Array.from({length: 100}, () => 0)
    const b = Array.from({length: 10}, () => 1)
    a.push(...b.reverse())
    b.length = 0
}
console.timeEnd('length0')
```

Ran a test, all seems same speed, 😄

thorn0: It's simpler when you know that it's always the same instance.

fisker: Agree, I was thinking if there is a simpler way, then I came up with `.splice(0).reverse()` and ran a test.

fisker:
I forgot that we already renamed this file, have to copy line by line...https://github.com/prettier/prettier/pull/9583/commits/c5ffd9130f03f1e90bbebdcc6942b383ffdc3a47 + https://github.com/prettier/prettier/pull/9583/commits/f79c8d56acaf17e89fc2045c063604a7d3a43cd0

thorn0:
> I forgot that we already renamed this file, have to copy line by line...[c5ffd91](https://github.com/prettier/prettier/commit/c5ffd9130f03f1e90bbebdcc6942b383ffdc3a47) + [f79c8d5](https://github.com/prettier/prettier/commit/f79c8d56acaf17e89fc2045c063604a7d3a43cd0)

Looks beautiful. One thing that annoys me is `groupModeMap`. What should we do with it?

- Option 1: make `groupModeMap` local to `printDocToString` and pass it to `fits` as an argument
  - 👍 `fits` becomes pure
  - 👎 `fits` gets one parameter more (`groupModeMap`)
- Option 2: make `groupModeMap` and `fits` local to `printDocToString`, then inside `fits`, `groupModeMap`  and `lineSuffix` would be used from the outer scope
  - 👍 `fits` gets one parameter fewer (`hasLineSuffix`)
  - 👎 `printDocToString` becomes much longer

fisker:
I remember you told me this before, I'd go with "Option 1".

fisker:
By the way, I was thinking about maybe someday we want a `printDocToStream` another day. But looks like `trim` is the only thing changes `out` array, am I right?

fisker:
> gets one parameter more

Maybe we can add an internal object to save `cmds`/`width`/`groupModeMap` on it, so we can pass it instead of all those parameters.

fisker:
Also, [this part](https://github.com/prettier/prettier/blob/f79c8d56acaf17e89fc2045c063604a7d3a43cd0/src/document/printer.js#L597-L614) looks can be simplify to me, it should be able to done during print.

thorn0:
> By the way, I was thinking about maybe someday we want a printDocToStream another day. But looks like trim is the only thing changes out array, am I right?

Yes. I wonder if users really use it and if it's really needed. The same thing (formatting of preprocessor instructions) can be probably achieved with `dedentToRoot` or some other manipulation with indentation.

> Maybe we can add an internal object to save cmds/width/groupModeMap on it, so we can pass it instead of all those parameters

Not sure about this way. It might make code appear more complex. Also `fill` passes `[]` instead of `cmds`. I didn't try to understand its implementation yet.

fisker:
The `trim()` looks problematical to me too, at least `cursor` placeholder should be ignored when searching whitespace back.

thorn0:
Just for reference. Discussions that lead to adding `trim`: #4678, #4771, #4772
An example of usage: https://github.com/warrenseine/prettier-plugin-csharp/blob/93ff6a315c8f59df78fa63764cafb4ba40c2ea57/src/printer.js#L4002

thorn0:
Seems to be replaceable with `lineSuffix`: [demo](https://deploy-preview-10183--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBzAThArgBwBQDaAOlAASlEh4BUlANCeQJZQAmCMhj5pANi3AZkelAAro4MGEzjok9buX5RBiiiADyOFk2gBDeKwYghPPgIDKWAGbWmADzzt2sACoQAShAicCACz10VmVBUgByAGIma1IcCTiIMDgAZ2SIdAA6LLCAXQBKPNMeEKLySgBhCHZjUvMVWsoAMXSAWwN4dAVhfKKS7kpqPK78kxAIHGloZORQQMwAd1FAhGmUPV55vQBPaboQACN0PTAAa0kLPRa4ABkBZGt15MEDo9PznGOWVGQYdCxnuAtfZwVjOa56KCoLB6VBwZroNpSL7IEB6LAwCCjPwwFq8ADqfiY8GSHySFhWRKYADciVsUWBUqMWE90DBxDC2vdHs8AFbJewWL68OAARSwPjgXN4Tz2H3QLJRrESAFo4PYcLx0rJRnEWDA8UxWDA-MgABwABllmCeeKOOBRcRSsipkr2AEdxfBxONVqjksqVCCQaMJB6mBJ2ahOUgHtLnk8Wkwfn940LRZ7JTHuXsYHp9gajSakAAmHNHJjKVCVFrRkApACsoywT1cedWsZlICp-wAkmwOBYwOgmBMAIJsCwwLbCqVPAC+c6AA)

thorn0:
I haven't find any alive plugins that use it. How about removing it in v3 with a recommendation to use the `lineSuffix` trick instead?

fisker:
I don't know how them works... But I trust your judgment.

thorn0:
Nope, I was wrong. That trick doesn't solve the core problem: to skip the indentation _after_ a line break was printed.

PR #13172, opened 2022-07-24
docs: Mention `code-tag` helper in Embedded Language Formatting

fregante:
## Description

Template literal functions appear simple, but are deceiving:

```js
const html = ([string]) => string;

html`<title>${document.title}</title>`;
// -> '<title>', not '<title>The document title</title>'
```

So I published [`code-tag`](https://github.com/fregante/code-tag/blob/main/index.ts) to safely mark embedded code.

I'd like to extend Prettier’s documentation to mention the existence of this helper and I'm open suggestions to extending the helper’s own readme to document this obscure _platform_ feature:

- https://github.com/fregante/code-tag/pull/5

Prettier’s own docs only mention this feature in the options page I think, so I avoided adding examples because the sentence is already quite long.

## Checklist

None applicable

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -440,7 +440,7 @@ _First available in v2.1.0_
 
 Control whether Prettier formats quoted code embedded in the file.
 
-When Prettier identifies cases where it looks like you've placed some code it knows how to format within a string in another file, like in a tagged template in JavaScript with a tag named `html` or in code blocks in Markdown, it will by default try to format that code.
+When Prettier identifies cases where it looks like you've placed some code it knows how to format within a string in another file, like in a tagged template in JavaScript with a [tag named `html`](https://github.com/fregante/code-tag#readme) or in code blocks in Markdown, it will by default try to format that code.
 
 Sometimes this behavior is undesirable, particularly in cases where you might not have intended the string to be interpreted as code. This option allows you to switch between the default behavior (`auto`) and disabling this feature entirely (`off`).
 


docs/options.md
```diff
@@ -440,7 +440,7 @@ _First available in v2.1.0_
 
 Control whether Prettier formats quoted code embedded in the file.
 
-When Prettier identifies cases where it looks like you've placed some code it knows how to format within a string in another file, like in a tagged template in JavaScript with a tag named `html` or in code blocks in Markdown, it will by default try to format that code.
+When Prettier identifies cases where it looks like you've placed some code it knows how to format within a string in another file, like in a tagged template in JavaScript with a [tag named `html`](https://github.com/fregante/code-tag#readme) or in code blocks in Markdown, it will by default try to format that code.
```
fisker, line 443: Don't think it's a good idea to link this directly to `code-tag`.  Maybe add a list?

fregante: > so I avoided adding examples because the sentence is already quite long.

> extending the helper’s own readme to document this obscure platform feature

My intention is to link to other in the readme as well

sosukesuzuki (commented):
I don't think this needs to be mentioned in the document. We do not want to recommend using a tagged template literal that does nothing. I think it is enough to add an example.

fregante:
Judging by the formatting of your comment, that doesn’t seem to be as compatible. I’d still include examples to more such templates at least, otherwise  “tag named `html`” is meaningless to someone who doesn’t know what that does or how to make it work (see my first code block in this thread)

fregante:
I extended the readme of `code-tag` to better define this behavior and include more examples

<table>
<tr>
	<td>
<tr>
	<td>Screen Shot 4
<tr>
	<td>
</table>

Josh-Cena:
FWIW, I've yet to see someone using `const html = ([string]) => string;`... A common pitfall is `const html = String.raw`, but that's usually not as terrible as it seems. This is why we want [`String.cooked`](https://github.com/tc39/proposal-string-cooked)

fregante:
Hello could you merge this or improve the docs in some other way? As is, they're really lacking and this is intended to document the feature and offer an option. It's too bad that there has been better docs available in this PR for 9 months and Prettier users haven't had access to it.

fisker:
We format tagged templates with these names because people use libs provide them for certain functionality. I admire they are not well supported, but I don't think we should encourage people install a lib that provide these tags as "noop", it's kind like a hack.

Personally, I think we should drop support for all of them since they are causing problems, that's why we introduced `--embedded-language-formatting=off`. Language comments maybe safe to support, so perhaps we should only support them, but support more languages.

fisker:
I'm going to close this since team members not agree to accept. We'll discuss how to improve embed template support in https://github.com/prettier/prettier/issues/5588

PR #13203, opened 2022-08-01
Remove `prettier.doc.builders.concat`

fisker:
## Description

Ref #13142

## Checklist


fisker:
Should we remove it "softly" `const concat = parts => parts`?

sosukesuzuki (commented):
Update `commands.md` and add changelog?

sosukesuzuki (approved):

fisker:
@thorn0 https://github.com/prettier/prettier/pull/13203#issuecomment-1201056691

thorn0:
> Should we remove it "softly" `const concat = parts => parts`?

Yes, let's do this.

thorn0:
Otherwise for many plugins it might be the only breaking change. With this soft deprecation, there is a good change those plugins won't need to be changed at all.

thorn0 (approved):

fisker:
Should we change this line 

https://github.com/prettier/prettier/blob/e096ea7552c2aaec1883c7fdde4bd9f5283301bd/src/document/constants.js#L2

to anther string?

Otherwise, 

```js
{
	type: 'concat'
}
```

Will pass the doc type check, but fail when calling array methods.

thorn0:
Let's use "array".

PR #13210, opened 2022-08-02
Revert "Switch print/embed to async (#13104)"

fisker:
## Checklist


changelog_unreleased/api/12740.md
```diff
@@ -1,4 +1,4 @@
-#### [BREAKING] standalone version export style changed (#12740 by @fisker)
+#### [Breaking] standalone version export style changed (#12740 by @fisker)
 
 The `esm/standalone.mjs` entry has changed export style from default export to named export.
```
thorn0, line 3: That was fast! :)
My only comment is that there is no need to revert this file.

fisker:
Let's merge and bring yours?

thorn0 (approved):

PR #13211, opened 2022-08-02
Support async parser for embedded languages

thorn0:
## Checklist


src/main/ast-to-doc.js
```diff
@@ -35,6 +35,9 @@ function printAstToDoc(ast, options, alignmentSize = 0) {
   const cache = new Map();
   const path = new AstPath(ast);
 
+  const embeds = new Map();
+  await printEmbeddedLanguages(path, mainPrint, options, printAstToDoc, embeds);
```
fisker, line 39: `const embeds = await printEmbeddedLanguages(...) `?

thorn0: Won't work. `embed` is used inside `mainPrint`, which is called inside `printEmbeddedLanguages`.

src/language-js/embed.js
```diff
@@ -47,11 +47,10 @@ function embed(path, print, textToDoc, options) {
...
   }
 
-  const language = getLanguage(path);
-  if (!language) {
-    return;
-  }
+  return getLanguage(path);
+}
 
+function embed(path, print, textToDoc, options, language) {
```
fisker, line 53: What if we ask `embed` to run print function, so we don't need add an extra detectEmbeddedLanguage?


```js
function embed() {
  const language = ...;
  return () => formatMarkdown(...)
}
```

fisker: And the logic don't need separate, and check language twice.

thorn0: Changing `embed`'s return type from
```ts
Doc | undefined | Promise<Doc | undefined>
```
to
```ts
Doc | undefined | () => Promise<Doc | undefined>
```
is an interesting idea, but I've run out of time for now. :(

fisker: Take your time, no rush. There is no async parser in the world yet.

thorn0: I agree that this would be a better API.

fisker: https://github.com/thorn0/prettier/pull/1

fisker:
I have one more concern, if this get merged, the `print` can't be async. But some plugins call `parse` in `print`, and we already allow `parse` to be async, so the plugin will be broken. I guess the answer is move those parts to `embed`?

[`@prettier/plugin-pug` format embed languages in this way](https://github.com/prettier/plugin-pug/blob/6ed005ec74e3ac2eb1a48699802b3f574f57998b/src/index.ts#L112).

//cc @Shinigami92

thorn0:
I can't see where it calls parse in print

fisker:
https://github.com/prettier/plugin-pug/blob/6ed005ec74e3ac2eb1a48699802b3f574f57998b/src/printer.ts#L650

thorn0:
I guess that should be moved to embed then. Isn't this what embed is for?

fisker:
As mentioned reason here, https://github.com/thorn0/prettier/pull/1#discussion_r937449363, this plugin has to move it to `embed` and add a fallback in `print`.

fisker:
@Shinigami92 ^

Any comment?

Shinigami92:
> @Shinigami92 ^
> 
> Any comment?

Would make things much harder for me (I wrote the whole plugin in an iterative way instead of a AST walker because pug works much better and more performant this way)
But yeah, as this breaking change introduced by prettier I would assume a major version bump
and then I and other plugin maintainers need to deal with the breaking change 🤷

fisker:
As I understand, you don't use `print` either, you traverse yourself?

In that case, you can also do print (async stuff) in parse... Some other plugins do this..

thorn0:
@Shinigami92 Making `printer.print` async lead to a 20% slowdown. See https://github.com/prettier/prettier/issues/13158#issuecomment-1193170276

Shinigami92:
I get the tokens in `parse`: https://github.com/prettier/plugin-pug/blob/6ed005ec74e3ac2eb1a48699802b3f574f57998b/src/index.ts#L69-L71
Pass it to `print` https://github.com/prettier/plugin-pug/blob/6ed005ec74e3ac2eb1a48699802b3f574f57998b/src/index.ts#L105-L107
And process it with a `while`-loop in the `Printer`: https://github.com/prettier/plugin-pug/blob/6ed005ec74e3ac2eb1a48699802b3f574f57998b/src/printer.ts#L325

thorn0:
@fisker `print` can be allowed to be async if it's not used recursively (`await` only the root call), but that would be a really clumsy API. We might want to keep it undocumented.

fisker:
We also have `printer.preprocess()` maybe should allow it to be async?

docs/plugins.md
```diff
@@ -246,36 +249,57 @@ Check out [prettier-python's printer](https://github.com/prettier/prettier-pytho
...
+    return async (textToDoc) => {
+      return [
+        "<script>",
+        await textToDoc(node.javaScriptCode, { ...options, parser: "babel" }),
+        "</script>",
+      ];
+    };
   }
-  return false;
 }
```
fisker, line 298:
```suggestion
function embed(path, options) {
  const node = path.getValue();
  if (node.type === "javascript") {
    return async (textToDoc) => {
      return [
        "<script>",
        hardline,
        await textToDoc(node.javaScriptCode, { parser: "babel" }),
        hardline,
        "</script>",
      ];
    };
  }
}
```

tests/format/misc/plugins/embed-async-parser/with-aync-parser-code-block.md
```diff
@@ -1,5 +1,5 @@
 ```async-parser
```
fisker, line 1: Let's use `markdown` async parser, and remove this test?

https://github.com/prettier/prettier/pull/13129/files

fisker (approved):

fisker:
Your plugin is unstable 😄

thorn0:
Should we print a warning (or error) if `printer.embed.length > 2`?

tests/format/misc/plugins/embed-async-printer/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -16,7 +16,7 @@ This text should be uppercased.
 
 =====================================output=====================================
 \`\`\`async-printer
-/* formatted by async-printer plugin */ this text will be lowercased and prefixed with a comment.
+this text will be lowercased and whitespace-normalized.
```
fisker, line 19: Use this JSON as text, and process them separately, more clear.

```json
{
	"printer.process": "text1",
	"printer.print": "text2"
}
```

fisker:
> Should we print a warning (or error) if `printer.embed.length > 2`?

Maybe throws `error`?

The following still work

```js
embed(path, print, textToDoc, options) {return onlyUsePath(path)}
```

The following may silently lead to unexpected result, `options` not there

```js
embed(path, print, textToDoc, options) {return onlyUsePathAndOptions(path, options?.singleQuote)}
```

thorn0:
On the other hand, what if somebody tries to stay compatible with both v2 and v3 and does stuff like this?
```js
embed(path, printOrOptions, textToDoc, options) {
  return typeof textToDoc === "function"
    ? oldEmbed(path, printOrOptions, textToDoc, options)
    : newEmbed(path, printOrOptions);
}
```

sosukesuzuki (approved):
Thanks @thorn0 and @fisker for the great work!

fisker:
Will you try [this](https://github.com/thorn0/prettier/pull/1#issuecomment-1204093481)?

thorn0:
No, I really doubt it'll do a better job than `ignoredProperties` already does.

thorn0:
How about moving `printer.massageAstNode.ignoredProperties` to `printer.ignoredProperties`? It also should be used in other places where the AST in traversed (e.g., `hasNode`).

fisker:
If we are going to do this, I prefer 

```js
getVistorKeys(node) {}
````

But for now we can just simply use 

```js
const ignored = new Set(['tokens'])
getVistorKeys(node) {
	return Object.keys(node).filter(key => !ignored.has(key))
}
```

thorn0:
That doesn't look performance-friendly. It will need to be called for each node: a function call, creating an array of keys, and iterating it. A simple `set.has` check must be much cheaper.

fisker:
> a function call, creating an array of keys, and iterating it. 

When traverse ast, normally you need "get keys/values", "check keys", " iterating it", aren't they the same?

https://github.com/prettier/prettier/blob/21b4045dfa0e20aa99888b43bc5daacfc5492875/src/main/massage-ast.js#L19
https://github.com/prettier/prettier/blob/21b4045dfa0e20aa99888b43bc5daacfc5492875/src/main/multiparser.js#L77

thorn0:
Yep. Somehow I misread what you're proposing. It might be not that bad.

thorn0:
But why do you think a simple key set won't be good enough? Do you have any specific cases in mind?

fisker:
Not really, just every estree traver I used use `vistorKeys`, maybe we can check what the reason behind it.

fisker:
`acorn` walk add walk for each ast type from beginning, https://github.com/acornjs/acorn/commit/b660ff6d1c462a8e6b77aec96e4d0f76deabd079, I guess to be more effective?

thorn0:
Probably. Also maybe it's important to visit keys in a specific order.

fisker:
Many parser come with traverse function, eg https://github.com/graphql/graphql-js/blob/67aefd9c1daefceacf937363c2da9e1117e550d9/src/language/visitor.ts#L177, if we remove AstPath someday, we may want use this function or vistorKeys.

thorn0:
How are we going to know where we are at the tree without the path?
(Side question: do the tests pass locally for the `next` branch for you? I get 18 failed, seems to be related to color codes.)

fisker:
Link to that I ask meriyah to supply such ithing https://github.com/meriyah/meriyah/issues/180

fisker:
> I get 18 failed, seems to be related to color code

I found it few hours ago , was working on it, but late here.  Caused by https://github.com/prettier/prettier/pull/13208. I changed cli  test to not print color by default, but not working.

thorn0:
Ah, okay. I'm glad that it's not something local with my env.
And another question: https://github.com/prettier/prettier/pull/13211/files#diff-f825b86383b78b8c50baa02327cfd46f6c1a3fee472409d978e5454672a83417R13
Is it expected that plugins persist across `format` calls? Looks like a bug to me.

fisker:
> How are we going to know where we are at the tree without the path? Eslint use `.parent` https://github.com/prettier/prettier/issues/10785

fisker:
> Is it expected that plugins persist across format calls?

Not sure.

Maybe what I occurred in https://github.com/prettier/prettier/pull/13201 is related, but I forgot what the possible bug is,  this "[Revert plugin change to avoid possible bug](https://github.com/prettier/prettier/pull/13201/commits/7a0b1cd23c14ef76dcd4182ca72fe1c11fb170de)" commit fixed test.

thorn0:
> Is it expected that plugins persist across `format` calls?

Tracked it down: https://github.com/prettier/prettier/issues/13235

PR #13216, opened 2022-08-03
Use `try/finally` in `AstPath`

fisker:
## Description

Use `try/finally` in `AstPath`, make them `try/catch` friendly.

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -214,8 +214,8 @@ module.exports = {
     {
       files: [
         "tests/format/**/jsfmt.spec.js",
-        "tests/config/**/*.js",
         "tests/integration/**/*.js",
+        "tests/unit/**/*.js",
         "scripts/release/__tests__/**/*.spec.js",
       ],
       env: {
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -213,6 +213,7 @@ module.exports = {
     },
     {
       files: [
+        "tests/config/**/*.js",
         "tests/format/**/jsfmt.spec.js",
         "tests/integration/**/*.js",
         "tests/unit/**/*.js",
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -47,8 +47,9 @@ const config = {
     printBasicPrototype: false,
   },
   testMatch: [
-    "tests/format/**/jsfmt.spec.js",
-    "tests/integration/__tests__/**/*.js",
+    "<rootDir>/tests/format/**/jsfmt.spec.js",
+    "<rootDir>/tests/integration/__tests__/**/*.js",
+    "<rootDir>/tests/unit/**/*.js",
   ],
   testPathIgnorePatterns,
   collectCoverage: ENABLE_CODE_COVERAGE,
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -46,7 +46,10 @@ const config = {
     escapeString: false,
     printBasicPrototype: false,
   },
-  testRegex: "jsfmt\\.spec\\.js$|__tests__/.*\\.js$",
+  testMatch: [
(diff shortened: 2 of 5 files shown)


thorn0:
This definitely needs thorough benchmarking

fisker:
I never thought this will cause permeance issue.

anyway, 

Before: 

```text
[debug] '--debug-benchmark' option found, measuring formatWithCursor with 'benchmark' module.
[debug] '--debug-benchmark' measurements for formatWithCursor: {
[debug]   "benchmark": "format x 0.13 ops/sec ±40.51% (5 runs sampled)",
[debug]   "hz": 0.1279763236020376,
[debug]   "ms": 7813.94536
[debug] }
```

After:

```text
[debug] '--debug-benchmark' option found, measuring formatWithCursor with 'benchmark' module.
[debug] '--debug-benchmark' measurements for formatWithCursor: {
[debug]   "benchmark": "format x 0.13 ops/sec ±39.04% (5 runs sampled)",
[debug]   "hz": 0.13366986736401562,
[debug]   "ms": 7481.117619999999
[debug] }
```

I don't know if I can trust those numbers, the same version result quite differently on small files. So I find a bigger one.

thorn0:
Try-catch blocks were known to cause deoptimizations in old Node versions. Maybe it's not something to worry about anymore

fisker:
I ran many times on https://gitpod.io/#https://github.com/prettier/prettier/pull/13216 with this

```sh
PRETTIER_PERF_REPEAT=1000 yarn perf:benchmark CHANGELOG.md
```

The result is very unstable, maybe our benchmark script can be trust?

thorn0 (approved):
```
Benchmark 1: node ./bench.mjs ast-path-try-catch serial
  Time (mean ± σ):     52.371 s ±  0.219 s    [User: 64.481 s, System: 3.545 s]
  Range (min … max):   52.127 s … 52.783 s    10 runs

Benchmark 2: node ./bench.mjs ast-path-try-catch parallel
  Time (mean ± σ):     52.813 s ±  0.369 s    [User: 62.492 s, System: 5.370 s]
  Range (min … max):   51.978 s … 53.233 s    10 runs

Benchmark 3: node ./bench.mjs next serial
  Time (mean ± σ):     51.750 s ±  0.269 s    [User: 63.739 s, System: 3.431 s]
  Range (min … max):   51.326 s … 52.067 s    10 runs

Benchmark 4: node ./bench.mjs next parallel
  Time (mean ± σ):     53.110 s ±  0.150 s    [User: 63.534 s, System: 5.411 s]
  Range (min … max):   52.831 s … 53.298 s    10 runs

Summary
  'node ./bench.mjs next serial' ran
    1.01 ± 0.01 times faster than 'node ./bench.mjs ast-path-try-catch serial'
    1.02 ± 0.01 times faster than 'node ./bench.mjs ast-path-try-catch parallel'
    1.03 ± 0.01 times faster than 'node ./bench.mjs next parallel'
```
I've run this benchmark 3 times and the results are consistently like this. 1% is not much, but it's real, not a measurement error.

I guess we can merge this. BTW you can remove the `try` method of `AstPath` now.

fisker:
> BTW you can remove the try method of AstPath now.

No, we can't.
`AstPath#try` backup the whole stack, so whatever happened inside, the stack still correct.
Those methds don't work if error happens when call `.callParent()`.

thorn0:
I don't see what you mean. You have added `finally` in `callParent`.

fisker:
`.match` `.findAncestor` not `try/finnally`, we assume they won't throw.

thorn0:
> `.match` `.findAncestor` not `try/finally`, we assume they won't throw.

They don't change the stack, so it doesn't matter.

PR #13222, opened 2022-08-04
Remove `parser.preprocess`

fisker:
## Description

`parse` function is called immediately after this function, if plugin want do something before parse, it should be done inside `parse`.

The only case I can image is that parser need change originalText for options, or print errors, but I don't think any plugin need that.

Core plugins not using it.

## Checklist

### DIFF
--- a/src/main/core.js
+++ b/src/main/core.js
@@ -32,12 +32,12 @@ function attachComments(text, ast, opts) {
   return astComments;
 }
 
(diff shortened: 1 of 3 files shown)


fisker:
I was wrong about this, it seems plugin do need change text for comment attach, see #3664. It can be done by changing the `parse` function return result, but won't be necessary to  do another break change.

PR #13236, opened 2022-08-06
Reuse `ignoredProperties` in more places

thorn0:
## Description

```
Benchmark 1: node ./bench.mjs next serial
  Time (mean ± σ):     51.332 s ±  0.487 s    [User: 65.975 s, System: 3.728 s]
  Range (min … max):   50.748 s … 52.057 s    10 runs

Benchmark 2: node ./bench.mjs next parallel
  Time (mean ± σ):     51.291 s ±  0.441 s    [User: 64.703 s, System: 5.529 s]
  Range (min … max):   50.716 s … 52.060 s    10 runs

Benchmark 3: node ./bench.mjs ignored-properties serial
  Time (mean ± σ):     46.682 s ±  0.303 s    [User: 60.893 s, System: 3.284 s]
  Range (min … max):   46.249 s … 47.097 s    10 runs

Benchmark 4: node ./bench.mjs ignored-properties parallel
  Time (mean ± σ):     48.246 s ±  0.430 s    [User: 62.322 s, System: 5.515 s]
  Range (min … max):   47.543 s … 49.004 s    10 runs

Summary
  'node ./bench.mjs ignored-properties serial' ran
    1.03 ± 0.01 times faster than 'node ./bench.mjs ignored-properties parallel'
    1.10 ± 0.01 times faster than 'node ./bench.mjs next parallel'
    1.10 ± 0.01 times faster than 'node ./bench.mjs next serial'
```

## Checklist


fisker (approved):

thorn0:
For the latest commit here:
```
Benchmark 1: node ./bench.mjs dd034ec serial
  Time (mean ± σ):     48.498 s ±  0.216 s    [User: 62.927 s, System: 3.243 s]
  Range (min … max):   48.093 s … 48.776 s    10 runs

Benchmark 2: node ./bench.mjs dd034ec parallel
  Time (mean ± σ):     48.727 s ±  0.429 s    [User: 61.628 s, System: 5.037 s]
  Range (min … max):   48.036 s … 49.367 s    10 runs

Benchmark 3: node ./bench.mjs ignored-properties serial
  Time (mean ± σ):     43.798 s ±  0.203 s    [User: 57.627 s, System: 3.167 s]
  Range (min … max):   43.511 s … 44.156 s    10 runs

Benchmark 4: node ./bench.mjs ignored-properties parallel
  Time (mean ± σ):     44.865 s ±  0.288 s    [User: 57.717 s, System: 4.937 s]
  Range (min … max):   44.526 s … 45.384 s    10 runs

Summary
  'node ./bench.mjs ignored-properties serial' ran
    1.02 ± 0.01 times faster than 'node ./bench.mjs ignored-properties parallel'
    1.11 ± 0.01 times faster than 'node ./bench.mjs dd034ec serial'
    1.11 ± 0.01 times faster than 'node ./bench.mjs dd034ec parallel'
```

thorn0:
Tried to use `Object.keys` instead of `for-in`:
```
Benchmark 1: node ./bench.mjs dd034ec serial
  Time (mean ± σ):     48.425 s ±  0.201 s    [User: 62.848 s, System: 3.327 s]
  Range (min … max):   48.237 s … 48.914 s    10 runs

Benchmark 2: node ./bench.mjs dd034ec parallel
  Time (mean ± σ):     48.885 s ±  0.429 s    [User: 62.147 s, System: 4.981 s]
  Range (min … max):   48.264 s … 49.649 s    10 runs

Benchmark 3: node ./bench.mjs ignored-properties serial
  Time (mean ± σ):     43.948 s ±  0.158 s    [User: 57.297 s, System: 3.473 s]
  Range (min … max):   43.707 s … 44.174 s    10 runs

Benchmark 4: node ./bench.mjs ignored-properties parallel
  Time (mean ± σ):     44.407 s ±  0.224 s    [User: 56.884 s, System: 5.030 s]
  Range (min … max):   44.127 s … 44.822 s    10 runs

Benchmark 5: node ./bench.mjs ignored-properties-object-keys serial
  Time (mean ± σ):     44.242 s ±  0.224 s    [User: 58.015 s, System: 3.423 s]
  Range (min … max):   43.965 s … 44.747 s    10 runs

Benchmark 6: node ./bench.mjs ignored-properties-object-keys parallel
  Time (mean ± σ):     45.159 s ±  0.209 s    [User: 58.187 s, System: 5.016 s]
  Range (min … max):   44.910 s … 45.521 s    10 runs

Summary
  'node ./bench.mjs ignored-properties serial' ran
    1.01 ± 0.01 times faster than 'node ./bench.mjs ignored-properties-object-keys serial'
    1.01 ± 0.01 times faster than 'node ./bench.mjs ignored-properties parallel'
    1.03 ± 0.01 times faster than 'node ./bench.mjs ignored-properties-object-keys parallel'
    1.10 ± 0.01 times faster than 'node ./bench.mjs dd034ec serial'
    1.11 ± 0.01 times faster than 'node ./bench.mjs dd034ec parallel'
```

thorn0:
As we can see, `Object.keys` isn't that bad. We can switch to it if we decide to explore the `getVisitorKeys` approach, but for now I'd prefer to keep the most performant and memory-friendly way to iterate properties, that is `for-in`.

fisker:
If we are not going to document this, I'm fine with this.  If we are going to document new property, I prefer `getVisitorKeys`. I don't want end up difficult to remove this.

thorn0:
Yes, let's not document it

fisker:
It seems we don't traverse all plugin AST before. 

I'm little worried about embedded print traverse, you did ignored some common properties, but to be safer, we can add a cache to prevent circular reference, do you think nessary?

thorn0:
Let's use a `visited` cache only if the printer doesn't have `ignoredProperties`.

thorn0:
On the other hand, if nodes happen to have links like `nextSibling` that will ruin the depth-first processing anyway. I'm not sure what to do.

fisker:
Me too, FYI acorn  used use same reference for shorthand  property key/value

fisker:
Let's pretend we don't know, not do anything for now.

thorn0:
Or we can document `ignoredProperties`. If we add `getVisitorKeys`, `ignoredProperties` still can be used when `getVisitorKeys` isn't defined.

fisker:
Okay.

thorn0:
How about renaming `ignoredProperties` to `ignoredKeys`?

fisker:
Fine with both.

thorn0:
`nonTraversableKeys`?

fisker:
https://github.com/prettier/prettier/pull/13211#issuecomment-1207261668  This is not hard, and you said there are no much performance difference, let's implement it? I can do it if you like.

fisker:
We don't even need write test, these lines all should be able to remove https://github.com/prettier/prettier/blob/3e9e9ddec7e50169066c717bc89c3b2a7538d688/src/language-css/parser-postcss.js#L164

https://github.com/prettier/prettier/blob/3e9e9ddec7e50169066c717bc89c3b2a7538d688/src/language-graphql/parser-graphql.js#L29-L32

thorn0:
I said the difference is small if `for-in` is simply replaced with `Object.keys`. But to say something definite about the `getVisitorKeys` approach, we need to benchmark a proper implementation of it, not a fallback to ignored properties. If it turns out that such an implementation doesn't have performance benefits, I don't see why add `getVisitorKeys`.

thorn0:
> We don't even need write test, these lines all should be able to remove

I don't think it's that simple. Better not to do in this PR.

fisker:
>  I don't see why add getVisitorKeys.

I think it's more flexible api.

> we need to benchmark a proper implementation of it,

I'll draft a PR. (maybe in two days, monday always a busy day)

thorn0:
> I think it's more flexible api.

It is flexible, but we don't have use cases for this flexibility, so I'm skeptical.

fisker:
> but we don't have use cases for this flexibility

Maybe not for now, but we will.

thorn0:
...or maybe we will not. Too hypothetical.

fisker:
When I open #10785, I imaged that we can do things like this someday,

```js
traverse(ast, (node) => {
	workers.print(node)
});

await print(ast);
```

thorn0:
I can't imagine how print can reuse the visitor keys

fisker:
Not `print`, `traverse`. I hope traverse won't send unexpected node to `print` (object but not ignored), with explicit visitorKeys, this never going to happen.

thorn0:
Why would `print` need `traverse`?

fisker:
I was talking about "print node in parallel".

Before you add traverse for the `embed`, we can't do parallel print without traverse.

This can be useful, if JS file is big, every root node can be printed to doc separately.

thorn0:
Closing in favor of #13255

PR #13248, opened 2022-08-07
Imporve error message for `InvalidDocError`

fisker:
## Checklist

### DIFF
--- a/src/document/invalid-doc-error.js
+++ b/src/document/invalid-doc-error.js
@@ -40,6 +40,11 @@ const disjunctionListFormat = (list) =>
   [...list.slice(0, -1), `or ${getLast(list)}`].join(", ");
 
 function getDocErrorMessage(doc) {
+  const type = doc === null ? null : typeof doc;
+  if (type !== "string" && type !== "object") {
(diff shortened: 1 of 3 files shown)


fisker:
Main change https://github.com/prettier/prettier/pull/13248/commits/7eac6f3f36e09f384912c64cbaff5157a2ba0e97

src/document/invalid-doc-error.js
```diff
@@ -55,6 +51,17 @@ function getDocErrorMessage(doc) {
     throw new Error("doc is valid.");
   }
 
+  const type = doc === null ? null : typeof doc;
+  if (type !== "string" && type !== "object") {
+    return `Unexpected doc '${type}', \nExpected it to be 'string', 'array', or 'object'.`;
```
thorn0, line 56: It can't be `string` at this point, and "array" isn't a `typeof` value. 🤓

fisker: > It can't be string at this poin

It can be if `new InvalidDocError('foo')`

> and "array" isn't a typeof value.

I think it's fine to use in the message.

thorn0 (approved):

PR #13250, opened 2022-08-07
Remove support for custom parser api

### LINKED ISSUE #13232: Remove custom parser API
We've supported ESM config, ESM plugins, there is one more ESM related thing left.

We support [file path as `parser`](https://github.com/prettier/prettier/blob/2c77149e6a3f6d5a74745a23c38e13f2ffebe28d/src/main/parser.js#L53), but we always treat it as [estree parser](https://github.com/prettier/prettier/blob/21b4045dfa0e20aa99888b43bc5daacfc5492875/src/main/load-parser.js#L11-L23), I really don't think this feature is useful.

Q: Should we remove support for it or load if via `import()`?

fisker:
## Description

Fix #13232

## Checklist


docs/api.md
```diff
@@ -147,4 +149,32 @@ prettier.format("lodash ( )", {
...
+import parserBabel from "prettier/parser-babel.js";
+
+const myCustomPlugin = {
+  parsers: {
+    "my-custom-parser": {
+      parse(text) {
+        const ast = parserBabel.parsers.babel.parse(text);
+        ast.program.body[0].expression.callee.name = "_";
+        return ast;
+      },
```
fisker, line 165:
```suggestion
      async parse(text) {
        const ast = await parserBabel.parsers.babel.parse(text);
        ast.program.body[0].expression.callee.name = "_";
        return ast;
      },
```

Should we suggest async api? we may change `babel.parse` in future.

docs/api.md
```diff
@@ -147,4 +149,32 @@ prettier.format("lodash ( )", {
...
+      astFormat: "estree",
+    },
+  },
+};
+
+format("lodash ( )", {
```
fisker, line 171:
```suggestion
await format("lodash ( )", {
```

thorn0 (approved):

tests/integration/__tests__/parser-api.js
```diff
@@ -1,83 +0,0 @@
...
-      expect(typeof parsers.babel).toBe("function");
-      const ast = parsers.babel(text);
-      ast.program.body[0].expression.callee.name = "bar";
-      return ast;
-    },
-  });
-  expect(output).toBe("bar();\n");
-});
-
-test("parsers should allow omit optional arguments", async () => {
```
fisker, line 29: No wonder I can't find this test.... it's removed. We still need it.

PR #13260, opened 2022-08-08
Always wrap certain CSS prop values (box-shadow, text-shadow, etc)

### LINKED ISSUE #6024: When box-shadow has multiple values (separated by comma) it'd be prettier to put each value on a separate line
**Prettier 1.16.4**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAjCAPAtAZwBYCGAJhAO5IA6UABNQAwAOG1AjE-RwMQBmvANFVqNmAJnZ0ucKQG4qIPiAgMYAS2g5koAgCdtZAAo6EGlAQBuEFUXlptBMAGs4MAMoN7KqAHNkMbQFc4BTwYAFsAGwB1PBV4HHcwOBdjWJUzWIBPZHAcDQVPHDhtGH07L1CCZG4CcMKFACscDAAhO0dnFwJQuAAZTzgqmrqQRowXTy9wuABFfwh4QdqgkHdtQu1ssFybBm1PGEirGDxkAA46BV2IQsi7BmzduHWzAYVtOABHfxV30oJyypIapLBSFUIqXwBZY4CZTWbzAZAobLGAEVCHIjHZAiBR+AgqcITADCEFCFWyUGgrxA-kKABU0SZgYUAL4soA)
```sh
--parser css
```

**Input:**
```css
box-shadow:
  0px 1px 0 0 #fff,
  0px 2px 0 0 #eee;
```

**Output:**
```css
box-shadow: 0px 1px 0 0 #fff, 0px 2px 0 0 #eee;

```

**Expected behavior:**
```css
box-shadow:
  0px 1px 0 0 #fff,
  0px 2px 0 0 #eee;
```

When each value is on a separate line, the whole value feels significantly easier to read.

sciyoshi:
## Description

Always wrap specific CSS property values (box-shadow, transition, background-image, text-shadow, background). Fixes #6024

## Checklist

### DIFF
--- a/src/language-css/utils/index.js
+++ b/src/language-css/utils/index.js
@@ -54,6 +54,17 @@ function getPropOfDeclNode(path) {
   return declAncestorNode?.prop?.toLowerCase();
 }
 
+const wrappedProps = new Set([
+  "box-shadow",
+  "transition",
+  "background-image",
+  "text-shadow",
+  "background",
+]);
+function isWrappedProp(value) {
+  return wrappedProps.has(value.toLowerCase());
+}
+
 const wideKeywords = new Set(["initial", "inherit", "unset", "revert"]);
 function isWideKeywords(value) {
   return wideKeywords.has(value.toLowerCase());
@@ -402,6 +413,7 @@ module.exports = {
   insideAtRuleNode,
   insideURLFunctionInImportAtRuleNode,
   isKeyframeAtRuleKeywords,
+  isWrappedProp,
   isWideKeywords,
   isLastNode,
   isSCSSControlDirectiveNode,
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -37,6 +37,7 @@ const {
   insideAtRuleNode,
   insideURLFunctionInImportAtRuleNode,
   isKeyframeAtRuleKeywords,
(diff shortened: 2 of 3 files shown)


fisker:
👍 For this, we need more tests.

src/language-css/utils/index.js
```diff
@@ -54,6 +54,17 @@ function getPropOfDeclNode(path) {
   return declAncestorNode?.prop?.toLowerCase();
 }
 
+const wrappedProps = new Set([
+  "box-shadow",
+  "transition",
+  "background-image",
```
fisker, line 60: Not only `background-image`, we should treat all longhand `background-*` the same. https://developer.mozilla.org/en-US/docs/Web/CSS/background#constituent_properties

sciyoshi:
> For this, we need more tests.

Definitely - I also have a small thing to fix in the implementation first. Just wanted to make sure that this was a change that would be accepted before fixing up those additional things.

sciyoshi:
@fisker I think this PR needs a design decision before I can move forward - I'd be curious to hear your thoughts on https://github.com/prettier/prettier/issues/6024#issuecomment-1209580521.

bronzegoddess:
This is much needed

PR #13262, opened 2022-08-09
Use `startColumn: 1` in babel parse

fisker:
## Description

Easier to throw error. Local test didn't pass, should be bugs in babel.

## Checklist

### DIFF
--- a/src/language-js/parse/babel.js
+++ b/src/language-js/parse/babel.js
@@ -44,6 +44,7 @@ const parseOptions = {
   ],
   tokens: true,
   ranges: true,
+  startColumn: 1,
 };
 
 /** @type {ParserPlugin} */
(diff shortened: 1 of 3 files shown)


thorn0:
That changes the column number only for the first line.

fisker:
```js
> require('@babel/parser').parse('a\nb',{startColumn: 99, startLine: 99}).program.body.map(n => n.loc)
[
  SourceLocation {
    start: Position { line: 99, column: 99, index: 0 },
    end: Position { line: 99, column: 100, index: 1 },
    filename: undefined,
    identifierName: undefined
  },
  SourceLocation {
    start: Position { line: 100, column: 0, index: 2 },
    end: Position { line: 100, column: 1, index: 3 },
    filename: undefined,
    identifierName: undefined
  }
]
```

Seems it's working that way... But I don't think that's expected. Let's see what babel says.

fisker:
Misunderstood this option.

Issue #13157, opened 2022-07-23
Line breaks in multiline strings aren't replaced with `literalline` and confuse `fits`


thorn0:
Note that it's not about template literals. It's about normal string literals with escaped line breaks.

**Prettier 2.7.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAbCAnOAWwEkAHNAV0IDEIIAhAQwC8m8BGTAXkwHIBZCthgBLbCKhxMGPBIDmAHSVRMqtWsHCxEqTPlKFUXocOQoGHPiJlKNOk1aM8AJm59No8ZOkxZURcrqQR7a3nr+BipBGkKeOj5+csZQIAA0IBCkoujIoE54EADuAApOCGjIIIzYhYwAnhXpAEZ4jGAA1nAwAMqMhHAAMjrIAGbVaHDNrR1d3aRt8si+FJMgRE1wACabWwOM-hSMcnC0eISMMKL+lYwUMBBpIAAWMITYAOpPIvBo82Bw3XK3xEADdvnVKmA0I0QBIJngYMVWnJzqNxqsAFZoAAe3Xk2DgAEUKBB4GjsBN0vM8PDKk1GBtsI9SH4YO8RJsYE9kAAOAAMVIKE3erVIlRZcHhILgjwAjiT4EjMhUUIw0ABaSRbLaPAjykQEJFHVFIMYU1YTQgiJZ4FbpND4okKmWm9HpGAM9mc7lIZzu1qhOQAYQghBNazQAFZHhQJgAVBkqs2UkAglbEKA7WDdMCyLIAQUz3RgdQJ5ImAF8K0A)

```sh
--parser babel
```

**Input:**

```jsx
const loremIpsumFooBazBar1 = 'Multiline string\
         Multiline string\
'

const loremIpsumFooBazBar2 = 'Multiline string\
         Multiline string\
         Multiline string'

```

**Output:**

```jsx
const loremIpsumFooBazBar1 = "Multiline string\
         Multiline string\
";

const loremIpsumFooBazBar2 =
  "Multiline string\
         Multiline string\
         Multiline string";

```

**Expected behavior:**

Formatting style shouldn't change when more lines are added.

Related: #6501

PR #13274, opened 2022-08-11
fix: line breaks in multiline strings, issue 13157

GlebDolzhikov:
## Description

in order to fix https://github.com/prettier/prettier/issues/13157 i've created a function that replaces string literal linebreaks with prettier doc node "literallines", which helps to behave consistently with multiline strings of any length.

## Checklist


changelog_unreleased/javascript/13157.md
```diff
@@ -0,0 +1,31 @@
+#### Fix line breaks in multiline strings aren't replaced with literalline and confuse fits (#13157 by @thorn0 )
```
thorn0, line 1: The thing in parens should the PR number and author, i.e. you. Also the file should be named after the PR.

changelog_unreleased/javascript/13157.md
```diff
@@ -0,0 +1,31 @@
...
+
+<!-- prettier-ignore -->
+Note that it's not about template literals. It's about normal string literals with escaped line breaks.
```
thorn0, line 4: Please follow `template.md` more closely. See the comments in that file.

src/common/util.js
```diff
@@ -292,6 +295,16 @@ function printString(raw, options) {
       options.__embeddedInHtml
     )
   );
+
+  return adjustedString?.includes("\n")
```
thorn0, line 299: `makeString` can't return `null` or `undefined`, so `?.` isn't needed here.

src/common/util.js
```diff
@@ -292,6 +295,16 @@ function printString(raw, options) {
...
+    ? printIndentableString(adjustedString)
```
thorn0, line 300: No need to extract this logic to a function. It fits in one line.

thorn0:
This breaks the CSS printer:
**Prettier pr-13274**
[Playground link](https://deploy-preview-13274--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEATAlgNwNoDMIQC8AOiAEbGVQCGATqQLoAEwTkANhLUk7XKkwC+IADQgIABxjpoAZ2Sg6tCAHcACnQTyU1diuoBPeWLK1qYANZwYAZWoBbOABl0UOMly7ZcE2cvWbCXNXAHNkGFoAVx8QOHsyflR+J2ooEMjqELgAMS57ahhpNOQQakiYCFEQAAsYe3YAdWr0eFkgsDgbLRasFoMSsFljEFdvWhg1MxD8jy8YgCtZAA8bUPY4AEVIiHhZ9m8xINoxgaGqiVpXGAb0VBhq5AAOAAZD5W8GswkSi7gxzHcYgAjtt4JNJNpSrIALRufj8Kp8EHoPiTTIzJCefYxbz2dDhKI4tabUHuTFzMQwahkG53B5IABMlLM6HYoQAwhB7BjYrIAKxVSLeAAq1O0WIOIEw0QAklAkrAbGBLlIAILymwwAzrPbeQSCIA)

```sh
--parser css
```

**Input:**

```css
div[foo="b\
ar"] { color: red }
```

**Output:**

```css
div[foo="[object Object]"] {
  color: red;
}

```

fisker:
Shouldn't this be here?

https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/src/language-js/print/literal.js

thorn0:
CSS and Flow string literal types support this syntax too.

GlebDolzhikov:
try different approaches, for now, left it in printString call because when it calls from https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/src/language-js/print/literal.js it doesn't work for babel parser.
Also for now disable it for CSS parser - probably we should add another issue for multiline CSS, if it makes sense.

thorn0:
If you aren't going to fix CSS in this PR, please open a new issue about (tip: use the Report Issue button on the playground to create an issue).
Also please add tests for Flow and TS literal types.

GlebDolzhikov:
issue https://github.com/prettier/prettier/issues/13299

tests/format/typescript/literal/test.ts
```diff
@@ -0,0 +1,7 @@
+const fooBazBar1:string = 'Multiline string\
```
thorn0, line 1: This is not a literal type. See https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types

fisker:
I just noticed we have `replaceTextEndOfLine` to do the job, and used everywhere.

https://github.com/prettier/prettier/pull/10913

Since we are only going to change the JS part in this PR, let's use it [here](https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/src/language-js/print/literal.js) for now. Otherwise the bundle size will increase, unless we extract the function to a separate file.

GlebDolzhikov:
@fisker we can't move it [here](https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/src/language-js/print/literal.js)
because it will not fix it for babel parser

fisker:
> we can't move it [here](https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/src/language-js/print/literal.js)

Of course I mean both `Literal` and `StringLiteral`.

thorn0:
> I've adjust tests

Please do the same for Flow.

src/language-js/print/literal.js
```diff
@@ -40,14 +43,24 @@ function printLiteral(path, options /*, print*/) {
...
-        return printString(node.raw, options);
+        return printIndentableString(printString(node.raw, options));
       }
 
       return String(value);
     }
   }
 }
 
+function printIndentableString(str) {
```
fisker, line 54: Use `replaceTextEndOfLine` from '/src/document/utils.js'

GlebDolzhikov: Nice function, I use it but left ```printIndentableString``` in order to avoid duplicated multiline check code, what do you think?

fisker: Sorry for not been clear, you can use it directly, no need check `\n`.

GlebDolzhikov: nice, adjust it

fisker (approved):

fisker:
Out of scope of this PR, but @thorn0 do you think we should do something here? https://github.com/prettier/prettier/blob/a2a3849c430b31871db5e89ff1542b4d14f80f68/src/document/doc-printer.js#L272-L274

thorn0:
> do you think we should do something here?

We should. At least a dev env warning. Not that sure about throwing an error. I don't want v3 to break more plugins than it's already going to.

fisker:
> We should. At least a dev env warning. 

I don't think we should warn, we should silently split it. Plugins don't event print doc at all, eg `@prettier/plugin-pug`.

fisker:
>I don't understand why this works for Flow literal types without changing print/flow.js.

I don't know either, maybe because it didn't break before string in tests?

https://github.com/prettier/prettier/blob/a2a3849c430b31871db5e89ff1542b4d14f80f68/src/language-js/print/flow.js#L280-L281

thorn0:
> I don't understand why this works for Flow literal types without changing `print/flow.js`.

It doesn't work, the playground was cached. @GlebDolzhikov Please fix and add this snippet to tests.

**Prettier pr-13274**
[Playground link](https://deploy-preview-13274--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEMCeAHOACANhAJzgFsBJDAZwFdiAxCCAIQEMAvFggRmwF5sByALJVcMAJa4xUHBRgEpAcwA6KqNnUaNw0RKky5i-kqjH0WPIRLlqdBi3bMCAJl4Dt4ydOyz5UZas1A910vH0UVY0DNYM99XwV+EAAaEAgMcWgKZFBHAggAdwAFRwQslGZcfOY0LJSAIwJmMABrOBgAZWZiOAAZPWQAMwqKOHrGlrb2jCbFZDkqUZASOrgAE1W1nuY-KmYFOHoCYmYYcT9kEGYqGAhkkAALGGJcAHV7sXgKabA4dtKPsQANw+aAuYAotRAUhGBBghUaCmOg2GiwAVhQAB7tRS4OAARSoEHgyNwIxS0wIMIuA3w+TuGF8MBeYlWMHuyAAHAAGcl5EYvRoYC4MuAwwFwO4AR0J8HhaTKlwoAFppGs1nciNKxER4XskUghqTFiNiGI5gQFikKDj8TKJQaUSkYMw6szWeykE4nY0QgoAMIQYj6pYUACsdyoIwAKi6yoaySBAQtSFANrB2mB5OkAIKp9roXEkkYAX2LQA)

```sh
--parser flow
```

**Input:**

```jsx
type loremIpsumFooBazBar1 = 'Multiline string\
         Multiline string'

type loremIpsumFooBazBar2 = 'Multiline string\
         Multiline string\
         Multiline string'
```

**Output:**

```jsx
type loremIpsumFooBazBar1 = "Multiline string\
         Multiline string";

type loremIpsumFooBazBar2 =
  "Multiline string\
         Multiline string\
         Multiline string";

```

tests/format/flow/literal/test.js
```diff
@@ -0,0 +1,11 @@
+const fooBazBar1:'Multiline string\
```
thorn0, line 1: Please rename the file for the name to include "multiline" and use the snippet from this comment: https://github.com/prettier/prettier/pull/13274#issuecomment-1222488611

Same for TS.

changelog_unreleased/javascript/13274.md
```diff
@@ -0,0 +1,36 @@
+#### Fix line breaks in multiline strings aren't replaced with literalline and confuse fits (#13274 by @GlebDolzhikov)
```
thorn0, line 1:
```suggestion
#### Fix inconsistent formatting for multiline strings (#13274 by @GlebDolzhikov)
```
Most users have no idea what `fits` and `literalline` are.

tests/format/flow/literal/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,26 @@
...
+         Multiline string\\
+         Multiline string'
+
+=====================================output=====================================
+type loremIpsumFooBazBar1 = "Multiline string\\
+         Multiline string";
+
+type loremIpsumFooBazBar2 =
+  "Multiline string\\
+         Multiline string\\
```
thorn0, line 22: This output isn't right. You haven't fixed Flow string literal types.

thorn0 (changes requested):

fisker:
Thank you for contributing to Prettier! ❤️

Issue #13275, opened 2022-08-12
Parentheses shouldn't be removed from inferred function return types with extends constraints


rossipedia:
**Prettier 2.7.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAYhCAHgBUA+bAXm2OzgA94oATAZ2wAoA6bgQyWx5Q0AbQC6ASkrl2ASygAzOACdsAJVoMErbCxhK5Ac0kB+Ndn5Q4AN2UBuEABoQEDDBnQWyUDyVKIAdwAFHwRPFB4AG38eNE8nACMlHjAAazgYAGUeAFs4ABk5OGR5SJY4BKTU9IyMZMNkPQBXcpA4bPi4JiZOvMEDRp4DOHwlbJ4YNygDZBAeRpgIRxAACxhsiIB1ZZl4FlqwOAzQnZkrHbQZsBY4kDkypRhApIMx4tKWgCsWOgzDCLgAIqNCDwN4RMpOWpKe4zeI8DoRAC0MBuGH0sA2MiYMGWyAAHAAGSF+MobJIYGZouD3GxLACOwPgTxcYVmLERlk6nSWSjgDJkvKeg1eSBK4JaZWyMgaSmaThYf0BjKKoveThg8Mx2NxSAATOqkjIIoYAMIQbIi1osACsS0aZWI8LCYohICszQAkswEJkwPpXABBZgZdD-MFlAC+EaAA)

```sh
--parser babel-ts
```

**Input:**

```tsx
type Foo<T> = T extends (...a: any[]) => (infer R extends string) ? R : never;
```

**Output:**

```tsx
type Foo<T> = T extends (...a: any[]) => infer R extends string ? R : never;

```

**Expected behavior:**

The parentheses around `(infer R extends string)` should be preserved. Removing them results in an error with TypeScript, as shown in the TS Playground ([link](https://www.typescriptlang.org/play?ssl=3&ssc=1&pln=3&pc=79#code/C4TwDgpgBAYg9nAPAFQHxQLxWVCAPYCAOwBMBnKACgDpaBDALijqJAG0BdASk3QEsiAMwgAnKACVcBYuShlgIgQHMoAfglQmRCADdRAbgBQAemPYAFnwoB3OCIDWZBodCQoAIToiU6LDnyEpBQ09Ews7Ny8VALCYpIBMhTyikRKPOqSWroGQA)):

[...]

PR #13289, opened 2022-08-15
fix: paren shouldn't be removed from infer func 

GlebDolzhikov:
## Description

Parentheses shouldn't be removed from inferred function return types with extends constraints
fix for https://github.com/prettier/prettier/issues/13275
## Checklist


jspereiramoura (approved):

tests/format/typescript/conditional-types/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -263,13 +263,13 @@ type Unpacked<T> =
 =====================================output=====================================
 type TestReturnType<T extends (...args: any[]) => any> = T extends (
   ...args: any[]
-) => infer R
+) => (infer R)
```
thorn0, line 266: These shouldn't change.

thorn0 (approved):

sosukesuzuki (approved):

Issue #12597, opened 2022-04-05
Misaligned indent in curried arrow function


spenceryue:
**Prettier 2.6.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEwA6UAEn1ezjffSKAZxkwA9MBeTAQQDoAjACkKP3b0+NszQgAhszAATOADMA5gAsAlgCsA1gBsAtlAgAHAI4AncgFcAbgHdKATwBeI8VLlK1mnQePmrt0RMmCANBy82GL8gnY+jioaWnqGMKYWNuEOClEuse6JXva+IIGcAJS0AHz5RKxwRTTFmAAMZZgFANz5AL4c7VCdIH4gOjDy0KTIoEL6+hBmAApjCMMoQqpmQpbDvcz6QmDKcDAAykLqcAAy8lBwyJKLpHDrm9u7e9pbZ9LIMPpGtyBw6sxwYgkYmOQig0iMQmkcAAYhB9OohDABmDkMIjDAID0QLIYOpVAB1BTwUjPMBwPZzeQDExUyyosCkNYgM43fQwKabaQIy7Xb6KUiUPavVRwACKRgg8B5qhuvWehjg+lRMEs2jgpDA+nk2hgWO0Wtg+PkYhgsmQAEYAEy1OUTG74zbaVH69WKkwXXq6CXwDk6ebCUgAWnOAIBWP0cC98gjHMh3KQVxl3xu6nk70+yeFYu9FwTvN6MBERpNZqQloLm3kqleAGEIOp4z9SABWLFGG4AFRE80TspAJi+AEkoBJYHtNdqYPQR3sVSLpTdWq0gA)

```sh
--parser typescript
--print-width 120
```

**Input:**

```tsx
{
  {
    {
      const x = A.b(
        (
          c = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdef",
          d = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdef"
        ) =>
        (e) => 0
      );
    }
  }
}

```

**Output:**

```tsx
{
  {
    {
      const x = A.b(
        (
            c = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdef",
            d = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdef"
          ) =>
          (e) =>
            0
[...]

PR #13300, opened 2022-08-18
Fix row break in arrow func with default argument value

GlebDolzhikov:
## Description

fixes https://github.com/prettier/prettier/issues/12597

## Checklist

### DIFF
--- a/changelog_unreleased/javascript/13300.md
+++ b/changelog_unreleased/javascript/13300.md
@@ -1,0 +1,17 @@
+#### Fix row break in the arrow func with default argument value (#13300 by @GlebDolzhikov)
+
+<!-- prettier-ignore -->
+```js
+
+// Input
+const fn = (a = 1) => () => 0;
+
+// Prettier stable
+const fn =
+  (a = 1) =>
+  () =>
+    0;
+
+// Prettier main
+const fn = (a = 1) => () => 0;
+```
--- a/src/language-js/print/function.js
+++ b/src/language-js/print/function.js
@@ -305,7 +305,10 @@ function printArrowFunction(path, options, print, args) {
       // Always break the chain if:
       (node.returnType && getFunctionParameters(node).length > 0) ||
       node.typeParameters ||
-      getFunctionParameters(node).some((param) => param.type !== "Identifier");
+      getFunctionParameters(node).some(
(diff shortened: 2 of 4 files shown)


thorn0:
What you changed is intended behavior and doesn't fix the issue.

GlebDolzhikov:
if you agree with this [behavior](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzKBGTAXkwAoBDIzAJgEoiA+TABgG4AdKSKDHKKy8kTx1CjEgCMRjAMztO6LLmkCKxWg1KSNJMFMyyOXHrgAsKvRIu7tAEz0H53RVACs57eOLDt10aTvacPZyRs4AbO5+lj6E0hYBUUHaAObBhgq8AOyRYlpRvmI2sRZJUana2MEgADQgEAAOMACW6MigZABOHRAA7gAKnQhoyCBkADY9ZACew7XiHWRgANZwMADKZAC2cAAyTVBwyNjjaHBzC8ura-WL+8nIMB0ArmcgcJvicDY2XztkUMknmRknAAGIQDqbMgwZoAkZkJ4wCA1EAACxgmzGAHVUU14GgbmA4GshnimgA3PFTEZgNCzED7U4dGB9BbJKFHE6vABWaAAHms7mM4ABFJ4QeCcsanWo3DpMkbiMifMYo+odfYwLFNGwwVHIAAcTFl3VOWIW9RG6rgTPJh1qAEdxfBWQ1higyGgALQHL5fFEdOBOpqB1nAjlIY7S16nTZNB7PGNC0XOw6Rrm1GDK7W6-VIKiZhZNMZ3ADCEE2EbeaBcKKepwAKsr3VGZSByS8AJJQH6wNZgDWNACCPbWMCmwqlpwAvtOgA), let's close this PR

thorn0:
Your examples are contrived. If you find real world code that looks bad because of the current heuristics, let's change this. But the condition should be stricter than just "allow assignments patterns". E.g. "allow assignments patterns if there is only one parameter and the right-hand side of the assignment is a simple (whatever that means) expression".

thorn0:
BTW, you can use https://github.com/prettier/prettier-regression-testing to check the potential impact of your changes.

PR #13309, opened 2022-08-19
Update docs

fisker:
## Checklist


docs/cli.md
```diff
@@ -16,7 +16,7 @@ To format a file in-place, use `--write`. (Note: This overwrites your files!)
 In practice, this may look something like:
 
 ```bash
-prettier --write .
+prettier . --write
```
fisker, line 19: Am I the only one? I always type patterns first.

docs/ignore.md
```diff
@@ -13,13 +13,13 @@ To exclude files from formatting, create a `.prettierignore` file in the root of
 
 Example:
 
-```
```
thorn0, line 16: What's the point of marking it as `text`?

thorn0 (approved):

docs/ignore.md
fisker: To satisfy https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md040---fenced-code-blocks-should-have-a-language-specified, I can remove it.

thorn0: I just was curious.

fisker: I'll keep it then. I'll need update docs in #13530

PR #13335, opened 2022-08-21
Remove common options from plugins

fisker:
## Description

I don't think they should be listed.

## Checklist


fisker:
So, if it's not listed, the option value got filter-out? Don't know how it works...

PR #13346, opened 2022-08-22
Enforce load utilities from the index file

fisker:
## Description

Tree-shaking shouldn't be a problem now.

## Checklist


fisker:
It seems esbuild can't detect side effect correctly.

PR #442, opened 2017-01-24
Refactor traversal

vjeux:
- Introduce `findInDocs` that avoids some annoying boilerplate when you want to get a value while traversing the docs
- Implement a `hasStopped` behavior that stops the traversal when you found the value you needed

It shaves off a few milliseconds so it's negligible, but the code feels right :)

```js
./bin/prettier src/printer.js
```

Before:

screen shot 2017-01-23 at 4 29 19 pm

After:

screen shot 2017-01-23 at 4 29 09 pm

PR #13349, opened 2022-08-23
Add test for `traverseDoc` and `findInDoc`

fisker:
## Checklist


src/document/utils.js
```diff
@@ -158,37 +158,34 @@ function mapDoc(doc, cb) {
   }
 }
 
-function findInDoc(doc, fn, defaultValue) {
-  let result = defaultValue;
-  let hasStopped = false;
```
fisker, line 163: This is useless, `onEnter` returns `false` will stop traverse, must left over during refactoring https://github.com/prettier/prettier/pull/442

src/document/utils.js
```diff
@@ -158,37 +158,34 @@ function mapDoc(doc, cb) {
...
-  function findInDocOnEnterFn(doc) {
+function findInDoc(doc, fn) {
+  let result;
+  traverseDoc(doc, (doc) => {
     const maybeResult = fn(doc);
-    if (maybeResult !== undefined) {
```
fisker, line 166: I don't think we want use `findInDoc` to do find falsely value.

src/document/utils.js
```diff
@@ -158,37 +158,34 @@ function mapDoc(doc, cb) {
...
+        (doc.type === DOC_TYPE_LINE && doc.hard) ||
+        doc.type === DOC_TYPE_BREAK_PARENT
+    )
+  );
 }
 
-function willBreak(doc) {
-  return findInDoc(doc, willBreakFn, false);
+function canBreak(doc) {
+  return Boolean(findInDoc(doc, (doc) => doc.type === DOC_TYPE_LINE));
```
fisker, line 188: This not seems correct to me, line may live in a unbreakable group. But it's out scope of this PR.

src/document/utils.js
fisker: Maybe I'm wrong, adding test.

fisker: return false on `onEnter` doesn't stop traverse. https://github.com/prettier/prettier/pull/13349/commits/e3ff2edc6f778797659dbd8a323a4d26a7be3405

thorn0:
The named functions were used to make profiling more convenient as dev tools show those names in reports. I know we don't profile Prettier much these days, but if we need to do that, we might end up having to return these names

thorn0:
Overall, it's already a very simple function. Also it's exported and may be used by plugins. I'd just remove `hasStopped` and call it a day.

src/document/utils.js
```diff
@@ -161,16 +161,20 @@ function mapDoc(doc, cb) {
 function findInDoc(doc, fn, defaultValue) {
   let result = defaultValue;
   let hasStopped = false;
+
   function findInDocOnEnterFn(doc) {
+    if (hasStopped) {
+      // Stop processing rest docs
+      return false;
```
thorn0, line 168: Doesn't `return false` stop the further processing immediately? Is the `hasStopped` variable really needed?

fisker:
Okay, I reverted change, so this PR only going to add some tests.

A minor improvement, the `findInDoc` previously keep running function on the remaining docs, but it should not. I moved `hasStopped` check to top.

src/document/utils.js
fisker: See https://github.com/prettier/prettier/pull/13349#discussion_r952138588

fisker: I thought it does, but it's not. It will keep processing the siblings.

thorn0: Ah, okay. I see it's needed.

src/document/utils.js
```diff
@@ -161,16 +161,20 @@ function mapDoc(doc, cb) {
 function findInDoc(doc, fn, defaultValue) {
   let result = defaultValue;
   let hasStopped = false;
+
   function findInDocOnEnterFn(doc) {
+    if (hasStopped) {
+      // Stop processing rest docs
```
thorn0, line 167: Yes, the top seems to be a more correct place. Let's just not use the word "stop" in the comment:
```suggestion
      // Skip further processing
```

thorn0:
How about renaming `hasStopped` to `shouldSkipFurtherProcessing` and removing the comment?

PR #13350, opened 2022-08-23
Unwrap `fill` with single part in `cleanDoc`

fisker:
## Description

I tried this in #9736, but somehow it doesn't work.

## Checklist


fisker:
Use it in `printDocToString` nothing is broken. @thorn0 Do you think it's safe?

thorn0:
I didn't play with `fill` much, so I'm not sure.

fisker:
Though this is not related, but I'm planning to improve the `fill` print. Starting with #13315.

Print logic lives in https://github.com/prettier/prettier/blob/4959b7c113c6dd554f7e5f15e98f58563a778157/src/document/printer.js#L427-L500

We are printing the parts 2-by-2, we should measure all parts, and print more at once. I'm not sure if I can make it work, but will try.

Back to this one, if there are only one part in fill, it should work the same as a single doc. Problem I'm not quite sure either. But those tests passed, I guess it's safe. When working on #9736 I tried this but forgot the result.

thorn0:
I know where it lives, of course. I've just never looked deep into it. The doc printer issues I ever worked on never really involved `fill`. I'll have a deeper look if I have time.

Issue #5995, opened 2019-03-20
VSCode shifts errors underlining when Perttier's multiline import formatting applies


DmitryKoterov:
Hi.

There is a bug somewhere when builtin VSCode's source.organizeImports feature is used together with Prettier.

People sometimes prefer to use VSCode's built-in source.organizeImports feature **ON SAVE** which, among other things, places all imports on the same line. And this conflicts with Prettier when it formats imports at multiple lines: if there is an error in the code, the underlining of this error shifts towards beginning of the file.

I recorded a video which shows the effect, pretty easy to reproduce:
https://www.dropbox.com/s/apj5znzlqy8pub6/vscode-prettier.mp4?dl=0

There are two possible ways I imagine:

1. Fix the bug somehow (I tried to find the root case, but failed unfortunately), so organizeImports + Prettier could work together.
2. Introduce an option which allows to turn off imports indentation in Prettier (a quick PR: https://github.com/prettier/prettier/pull/5991) - it's a work-around in some sense, but at the same time, having all imports on the same line is VSCode's default they evangelize, and Prettier in some sense goes against this default.

lydell:
Hi! Thanks for the well written and the video!

VSCode underlining in the wrong place sounds like a bug in VSCode and/or the Prettier VSCode plugin to me – please report this to either/both of those!

That being said, I do think it would be interesting to discuss not breaking `import` statements into multiple lines. I happen to have discussed this thing before, for elm-format (another code formatter): https://github.com/avh4/elm-format/issues/324#issuecomment-290947601

> > I would not mind elm-format automatically splitting long those [long import lines]!

> Hmm … I might have to take that back. I worked on a view file today where the imports ended up taking a whole screen-full. So I tried joining each import to a single line. They became quite long – but it I think I like it better! I don’t really read the imports, so it was nicer to get to see the actual contents of the without scrolling a full page.

It happens quite often to me that files start with more than a window-full of just imports, that I have to scroll past every time I open them. To me it has started to feel more and more that imports are just cruft at the top of the file that I rarely look at and want to spend as little time on as possible. I want the computer to add new imports and remove unused ones, sort them and format them, and otherwise mostly keep them out of the way.

I think not breaking imports onto multiple lines could help with this. It could at least be worth exploring.

DmitryKoterov:
> it has started to feel more and more that imports are just cruft at the top of the file that I rarely look at

[...]

Issue #5086, opened 2018-09-12
No line break insertions in short CSS selectors


janosh:
**Prettier 1.14.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAgGcYBPAGzgBN8AzHCdfAcmPKoFpItoEZCmAOlCEZsefOiqoAhgEUArnBwla9RkwB0GgPRbt8mKjKFtkyjIVKSg4VEhRi+AArS8qMKkzTYhfAF4iUgpKDUI4MENoAAMhfHwvSnMoAHMkfAAmHDh0AG5Y-AAjaTAAa2T6eShKNIASYEx6TF8-AD54xsINGAALbLgNdGlUKAB1btR4AF88qDjzQkwyaRI08tRKGbi1ynZk6Uw0zOzN-G6ARgAaU-Sr7oBmW4AWfGB8uIB3dZ60s4AGX4ApCc4gkkql8L8NABWLKMI65N6FYplCpVWr1Dr+NoNCBNLq9SQaMioZLdGAAcSyCGm+Um+TqZgsimUGkw3R4kxeiPOtxupwep2er1mcVFNGgMHYhFQAC84D9YcD8HSRSq4gypHJmSQutIChQYJzhaL8NtdvsFcduZdrrcBd0hYitjh1pwIGR5OgoD9tOwzkq1cr6cBGVqrANhgAVPUGo3W3l2p5ckUm97jeBSrxgeX4KAQd44fZK52u+jvQ4B2lCGK2OAADzEBEocBo0nkZAILjcHi8PhAFxAuMiDmQoFcZa7CEIyBA0gAbhB1v2QAVC6U4DAAMpZ4bJZAwHCKAfDMJ4JyF5KDZCt4xwAcAK0IdYAQmuShvN9JJAAZYZwa-SLeA5eDgp4zkUBTkNAy4NMMMAjF83TIAAHL8wH0GEIyFpgM4NHAp5zv+A5ZAAjvIqBZOe0iXtIAFASAYToKgdFhAO0opBQCgQPALF3iAMB6ghlA9MgNz8YWRi7gAwgwV4oHmUBESA8hhNGBTTkgN5hJMkxAA)
```sh
--parser babylon
--no-semi
```

[...]

PR #13353, opened 2022-08-24
import multiline optimization

GlebDolzhikov:
## Description

partly fixes https://github.com/prettier/prettier/issues/5995

## Checklist

### DIFF
--- a/src/language-js/print/module.js
+++ b/src/language-js/print/module.js
@@ -1,6 +1,7 @@
 "use strict";
 
-const { isNonEmptyArray } = require("../../common/util.js");
+const { isNonEmptyArray, getLast } = require("../../common/util.js");
+
 const {
   builders: { softline, group, indent, join, line, ifBreak, hardline },
 } = require("../../document/index.js");
@@ -28,6 +29,7 @@ function printImportDeclaration(path, options, print) {
   const node = path.getValue();
   const semi = options.semi ? ";" : "";
   /** @type{Doc[]} */
+
   const parts = [];
 
   const { importKind } = node;
@@ -45,7 +47,11 @@ function printImportDeclaration(path, options, print) {
     semi
   );
 
-  return parts;
+  const lastSpecifier = getLast(node.specifiers);
+  const brakeAfterSpecifiers =
+    lastSpecifier && lastSpecifier.end < options.printWidth;
+
+  return brakeAfterSpecifiers ? group([parts]) : parts;
 }
 
 function printExportDeclaration(path, options, print) {
@@ -211,6 +217,24 @@ function printModuleSpecifiers(path, options, print) {
     if (groupedSpecifiers.length > 0) {
       if (standaloneSpecifiers.length > 0) {
         parts.push(", ");
+      }
+
+      const lastSpecifier = getLast(node.specifiers);
+      if (
+        groupedSpecifiers.length > 1 &&
+        lastSpecifier &&
+        lastSpecifier.end < options.printWidth
+      ) {
+        parts.push([
+          "{",
+          options.bracketSpacing ? " " : "",
+          join([", "], groupedSpecifiers),
+          options.bracketSpacing ? " " : "",
+          "}",
+          softline,
+        ]);
+
+        return parts;
       }
 
       const canBreak =
--- a/tests/format/js/import/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/js/import/__snapshots__/jsfmt.spec.js.snap
@@ -19,6 +19,10 @@ import {
 } from '.';
 import {fitsIn, oneLine} from '.';
 
+
+
+import { GET_DATA, GET_DATA_START, GET_DATA_START2, GET_DATA_START3} from "./test-export";
+
 =====================================output=====================================
 import {
   runTaskForChanged,
@@ -31,6 +35,13 @@ import {
   soWeCanGetItTo80Columns,
 } from ".";
 import {fitsIn, oneLine} from ".";
+
+import {
+  GET_DATA,
+  GET_DATA_START,
+  GET_DATA_START2,
+  GET_DATA_START3,
+} from "./test-export";
 
 ================================================================================
 `;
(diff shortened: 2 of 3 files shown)


GlebDolzhikov:
@fisker @thorn0 please take a look

fisker:
I(Personally) like things vertically one by one, I even use empty comment to prevent Prettier collapse to one line. Same opinion [here](https://github.com/prettier/prettier/issues/5086#issuecomment-523727602)

fisker:
Easier to reorder <kbd>Alt + ↑</kbd>, delete <kbd>Ctrl + X</kbd>

thorn0:
Looks like a reasonable tradeoff to me.

src/language-js/print/module.js
```diff
@@ -38,14 +40,16 @@ function printImportDeclaration(path, options, print) {
...
-    printModuleSpecifiers(path, options, print),
-    printModuleSource(path, options, print),
+    printModuleSpecifiers(path, options, print, groupId),
+    printModuleSource(path, options, print, groupId),
     printImportAssertions(path, options, print),
     semi
   );
 
-  return parts;
+  return group([parts]);
```
thorn0, line 52: The square brackets are redundant here.

src/language-js/print/module.js
```diff
@@ -40,18 +40,16 @@ function printImportDeclaration(path, options, print) {
...
     printImportAssertions(path, options, print),
     semi
   );
 
-  const lastSpecifier = getLast(node.specifiers);
-  const brakeAfterSpecifiers =
-    lastSpecifier && lastSpecifier.end < options.printWidth;
-
-  return brakeAfterSpecifiers ? group([parts]) : parts;
+  return group([parts]);
```
thdomingues-dev, line 52: `parts` already is an Array, you're passing double array to `group`.

```javascript
return group(parts);
```

tests/format/flow-repo/es6modules/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1516,10 +1516,8 @@ import { specifierNumber } from "ES6_Named1"; // Error: Did you mean \`specifier
 ///////////////////////////////////////////////////
 // == Multi \`export *\` should combine exports == //
 ///////////////////////////////////////////////////
-import {
-  numberValue1 as numberValue8,
-  numberValue2 as numberValue9,
-} from "./ES6_ExportAllFromMulti";
+import { numberValue1 as numberValue8, numberValue2 as numberValue9 }
+  from "./ES6_ExportAllFromMulti";
```
kachkaev, line 1521: Hmm not sure I prefer this over the original 🤔

Even where there is no `as`, new formatting makes a bit harder to scan through names. AFAIU, the main benefit we should expect from the change is the reduction of lines of code. But what does it improve in DX? Existing approach optimizes for cognitive ease and clean `git diff` – that will be a shame to lose.

thorn0: Good point about `as`. The current formatting seems to work better for such imports.

As for clean diffs, I think the proposed formatting might be an improvement in that regard. Currently, the diff isn't clean when switching between the single-line and multi-line formatting happens. With the proposed formatting, that switching is seemingly going to happen less often, and the diff for switching single-line⇄double-line is quite clean, definitely cleaner than single⇄multi.

thorn0: Just realized I'm used to diff viewers that highlight in-line diffs like this:
image
And you're probably talking about raw unhighlighted output of `git diff`. That makes me wonder how many people view diffs without in-line highlighting and why they would do that.

thorn0: > But what does it improve in DX?

Less scrolling through something that is edited mostly by tooling (auto-import, linters for sorting imports and removing unused ones), not manually? I wish we had some statistics on how many developers edit imports manually. That'd put this proposal in a good perspective. I guess these people are already in minority, but that's only a guess.

fisker: I never need write code more than 200 lines. Split the file, if it's too long, that's what ES Module for.

fisker: > auto-import

I have these lines in my vscode settings

```js
  "javascript.suggest.autoImports": false,
  "vetur.completion.autoImport": false,
  "volar.completion.autoImportComponent": false,
```

thorn0: > I have these lines in my vscode settings

I'm surprised. To me, this is like formatting code manually instead of using Prettier. The only problem I ever had with auto-import has been [fixed in TS 4.8](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#exclude-globs-auto-import). Do you think there are many people who share you preferences in this regard?

fisker: I don't know.

fisker: [There are few, seach on github](https://github.com/search?q=javascript.suggest.autoImports&type=code)

GlebDolzhikov: also didn't know that someone these days moves imports manually)

shamseen:
> partly fixes #5995

Hi @GlebDolzhikov, what parts of that issue aren't getting fixed? Also, will this be an opt-in feature? I think that's the best middle ground for an otherwise ... hotly debated proposal.

_______
Will this be implemented? There was quite the pushback when someone [stated it likely wouldn't be added](https://github.com/prettier/prettier/issues/5995#issuecomment-477775022). @thorn0 @fisker @kachkaev  @thdomingues-dev

PR #13356, opened 2022-08-26
Minor refactor

fisker:
## Checklist

### DIFF
--- a/src/language-js/printer-estree.js
+++ b/src/language-js/printer-estree.js
@@ -799,7 +799,10 @@ function printPathNoParens(path, options, print, args) {
 
     default:
       /* istanbul ignore next */
-      throw Object.assign(new Error("Unknown node type"), { node });
+      throw Object.assign(
+        new Error("Unknown node type: " + JSON.stringify(node.type)),
+        { node }
+      );
   }
 }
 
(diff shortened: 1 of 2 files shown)


sosukesuzuki (approved):

src/language-js/printer-estree.js
```diff
@@ -811,7 +799,7 @@ function printPathNoParens(path, options, print, args) {
 
     default:
       /* istanbul ignore next */
-      throw new Error("unknown type: " + JSON.stringify(node.type));
+      throw Object.assign(new Error("Unknown node type"), { node });
```
thorn0, line 802: Having `node.type` in the message is convenient. Let's keep it.

thorn0:
While you're at these minor cleanup, I noticed the same function `isTypeCastComment` is included twice in two different files. I guess it's not intended.

thorn0:
BTW, there is a thing I've wanted to ask for a while. Why do we use this style:
```js
function foo() { ... }
export { foo }
```
instead of
```js
export function foo() { ... }
```
?

fisker:
> Why do we use this style

Just personal preference, easier to see what module exported.

thorn0 (approved):

thorn0:
`isTypeCastComment` has been left in the exports :)

fisker:
Stupid me.

PR #13359, opened 2022-08-26
Refactor isSimpleCallArgument

thorn0:
## Description

Refactor `isSimpleCallArgument` and make it a bit stricter

## Checklist

### DIFF
--- a/src/language-js/utils/index.js
+++ b/src/language-js/utils/index.js
@@ -248,6 +248,13 @@ function isStringLiteral(node) {
   );
 }
 
+function isRegExpLiteral(node) {
+  return (
+    node.type === "RegExpLiteral" ||
+    (node.type === "Literal" && Boolean(node.regex))
+  );
+}
+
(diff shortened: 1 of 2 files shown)


src/language-js/utils/index.js
```diff
@@ -870,11 +874,14 @@ function isSimpleCallArgument(node, depth) {
...
-      (node.type === "ImportExpression" ||
-        isSimpleCallArgument(node.callee, depth)) &&
-      getCallArguments(node).every(isChildSimple)
-    );
+    if (
+      node.type === "ImportExpression" ||
+      isSimpleCallArgument(node.callee, depth)
+    ) {
+      const args = getCallArguments(node);
+      return args.length <= depth && args.every(isChildSimple);
```
thorn0, line 882: `args.length <= depth` is the only change in behavior. The rest is refactoring.

fisker (approved):

PR #13361, opened 2022-08-26
Doc printer: always print most expanded state in break mode, refactor

thorn0:
## Checklist


fisker:
Extract to `printer/print-group.js`?

thorn0:
I can't remember why I thought this was the right thing to do.

PR #13392, opened 2022-08-28
Add print pre-check to `handlebars`, `html`, `markdown`, `yaml`

fisker:
## Checklist

### DIFF
--- a/src/language-markdown/printer-markdown.js
+++ b/src/language-markdown/printer-markdown.js
@@ -464,12 +464,7 @@ function genericPrint(path, options, print) {
     case "export": // transformed in to `importExport`
     default:
       /* istanbul ignore next */
-      throw Object.assign(
-        new Error(
-          "Unexpected markdown node type: " + JSON.stringify(node.type)
-        ),
-        { node }
-      );
+      throw new UnexpectedNodeError(node, "Markdown");
   }
 }
 
--- a/src/language-markdown/printer-markdown.js
+++ b/src/language-markdown/printer-markdown.js
@@ -23,6 +23,7 @@ import {
 import { normalizeDoc, replaceEndOfLine } from "../document/utils.js";
 import { printDocToString } from "../document/printer.js";
 import createGetVisitorKeys from "../utils/create-get-visitor-keys.js";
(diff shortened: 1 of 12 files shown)


src/language-markdown/printer-markdown.js
```diff
@@ -456,7 +464,12 @@ function genericPrint(path, options, print) {
     case "export": // transformed in to `importExport`
     default:
       /* istanbul ignore next */
-      throw new Error(`Unknown markdown type ${JSON.stringify(node.type)}`);
+      throw Object.assign(
+        new Error(
+          "Unexpected markdown node type: " + JSON.stringify(node.type)
+        ),
+        { node }
+      );
```
thorn0, line 472: How about extracting this to `UnexpectedNodeTypeError`?

fisker: Used shorter name `UnexpectedNodeError` [`4098457` (#13392)](https://github.com/prettier/prettier/pull/13392/commits/4098457e654d36cc098bb7a225087cc77f2ab427)

thorn0: How do we decide whether a file goes to `src/common` or to `src/utils`? Other errors live in `src/common/errors.js`.

fisker: Mentioned once long time ago, https://github.com/prettier/prettier/pull/5889#pullrequestreview-206353103

My original plan is to move `src/common` into `src/utils`, but things didn't go well https://github.com/prettier/prettier/pull/13346

thorn0 (approved):

src/language-markdown/printer-markdown.js
thorn0: So you wanted to remove `common` completely and have `utils` instead? That would be just a rename, so I don't think that's what you mean. What were you going to keep in `common`?

fisker: > So you wanted to remove common completely and have utils instead?

I mean extract all staff from `src/common/` into `src/utils/`, remove `src/common/` completely, and re-export in `src/utils/index.js`.

PR #13393, opened 2022-08-28
Refactor printObject

thorn0:
## Checklist

### DIFF
--- a/src/language-js/print/object.js
+++ b/src/language-js/print/object.js
@@ -94,7 +94,7 @@ function printObject(path, options, print) {
       hasNewlineInRange(
         options.originalText,
         locStart(node),
-        locStart(propsAndLoc[0].node)
+        propsAndLoc[0].loc
       ));
 
   const separator = isFlowInterfaceLikeBody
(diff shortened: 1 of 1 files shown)


src/language-js/print/object.js
```diff
@@ -80,11 +90,11 @@ function printObject(path, options, print) {
             property.value.type === "ArrayPattern")
       )) ||
     (node.type !== "ObjectPattern" &&
-      firstProperty &&
+      propsAndLoc.length > 0 &&
       hasNewlineInRange(
         options.originalText,
         locStart(node),
-        locStart(firstProperty)
+        locStart(propsAndLoc[0].node)
```
fisker, line 97:
```suggestion
        propsAndLoc[0].loc
```

fisker (approved):
LGTM. Except I don't like nest ternary..

fisker:
https://github.com/prettier/prettier/blob/d545e5ffda6a19e2e6a014c2761043fcea367dfb/src/language-js/print/object.js#L204-L210 Doesn't look right to me, number = 0 doesn't mean it's first argument.

thorn0:
I agree. It looks suspicious.

PR #13397, opened 2022-08-29
Stop passing `undefine` and `null` to printer

fisker:
## Description

Should be safe, things like hole in array should checked by printer like [we did](https://github.com/prettier/prettier/blob/5a627651ff648edd38bd0b5e667210a8a1f4e4a1/src/language-js/printer-estree-json.js#L19)

## Checklist

### DIFF
--- a/changelog_unreleased/api/13397.md
+++ b/changelog_unreleased/api/13397.md
@@ -1,0 +1,19 @@
+#### [BREAKING] `undefined` and `null` are not passing to plugin's `print` function (#13397 by @fisker)
+
+If your plugin happened use `print` to print them before, please check them in the parent node.
+
+```diff
+function print(path, print) {
+-  const value = path.getValue();
+-  if (!value?.type) {
+-    return String(value);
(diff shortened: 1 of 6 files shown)


thorn0:
This potentially breaks plugins, right?

fisker:
> This potentially breaks plugins, right?

In theory, yes.


The only case I can think, if plugin print holes in array like this

```js
function print(path) {
	if (typeof path.getName() === "number" && !path.getValue()) {
		return "null"
	}
}
```

thorn0:
I guess we should mention that in the changelog.

fisker:
Another possible case,

AST: `{type: "Literal", value: 0}, {type: "Literal", value: false} `

```js
function print(path, print) {
	const node = path.getValue()
	if (!node?.type) {
		return String(node)
	}

	if (node.type === "Literal") {
		return print("value")
	}
}
```

Let's only check `undefined` and `null`, that's only we need for core plugins.

PR #13415, opened 2022-09-03
Override version for the trim package to fix Markdown perf

thorn0:
## Description

fixes #13217

```console
# before (2.7.1)
$ time prettier test-slow.md -c
Checking formatting...
All matched files use Prettier code style!

real    0m18.831s
user    0m18.844s
sys     0m0.141s

# after
$ time prettier test-slow.md -c
Checking formatting...
All matched files use Prettier code style!

real    0m1.664s
user    0m1.359s
sys     0m0.375s
```

## Checklist

- [ ] I’ve added tests to confirm my change works.
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).
- [ ] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [ ] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #13426, opened 2022-09-04
Get rid of regexp and handle cursor in trim in doc printer

thorn0:
## Description

see https://github.com/prettier/prettier/pull/13415#issuecomment-1236294897

## Checklist

### DIFF
--- a/src/document/printer.js
+++ b/src/document/printer.js
@@ -142,27 +142,41 @@ function generateInd(ind, newPart, options) {
   }
 }
 
-// TODO: Skip cursor placeholder
 // Trim `Tab(U+0009)` and `Space(U+0020)` at the end of line
 function trim(out) {
   let trimCount = 0;
+  let cursorCount = 0;
+  let outIndex = out.length;
 
-  while (out.length > 0) {
-    const last = getLast(out);
-    if (typeof last !== "string") {
-      break;
+  outer: while (outIndex--) {
+    const last = out[outIndex];
+
+    if (last === CURSOR_PLACEHOLDER) {
+      cursorCount++;
+      continue;
     }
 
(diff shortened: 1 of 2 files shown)


src/document/printer.js
```diff
@@ -142,27 +142,41 @@ function generateInd(ind, newPart, options) {
...
-    const trimmed = last.replace(/[\t ]*$/, "");
-    trimCount += last.length - trimmed.length;
+    if (process.env.NODE_ENV !== "production" && typeof last !== "string") {
+      throw new Error(`Unexpected value in trim: '${typeof last}'`);
+    }
 
-    if (trimmed !== "") {
-      out[out.length - 1] = trimmed;
-      break;
+    for (let charIndex = last.length - 1; charIndex >= 0; charIndex--) {
```
fisker, line 163: How do you think add a function to "get trailing space/tab length" so we don't need use label to break.

src/document/printer.js
```diff
@@ -142,27 +142,41 @@ function generateInd(ind, newPart, options) {
...
 
-  while (out.length > 0) {
-    const last = getLast(out);
-    if (typeof last !== "string") {
-      break;
+  outer: while (outIndex--) {
+    const last = out[outIndex];
+
+    if (last === CURSOR_PLACEHOLDER) {
+      cursorCount++;
```
fisker, line 155: I'd use an array to save cursors, and push them directly after text trimmed. But use number is fine too.

src/document/printer.js
thorn0: If avoiding the label is the only benefit, I'm not convinced.

fisker (approved):
Can you add a comment explain why we don't use `RegExp`?

Issue #11098, opened 2021-06-21
Comment gets wrongly placed after readonly keyword in mapped type


MichaelDeBoey:
**Prettier 2.3.0**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAKpnAXm2AB0ptsB6K7SAW3oRnMoCc4BDAE2gBs02ANq5sASwpQArvQBGcNgF0k2APIwAFgvxYRigNzkAvvpAAaEBAwwx0AM7JQnNmwgB3AArOEDlJz5unGgOFrJsnGAA1nAwAMoYERIA5sgwbFJwFnBycNzcuQAynFBJUpxJcABiEGz0nDA2JcggnFIwEOYgGjD0fADqGmLwdglgcLE+Q2IAbkNozWB2ISASdgowHuFJdcgAZv5rFgBWdgAeAELhUTGxnEwFEnB7B5kgJ6exyXxwAIpSEPBnnxDiAEmw1mxmugsHYwGwxNZOhh4bA+mJuJpkAAOAAMFmREDWfXCGGayLgEOmTwsAEd-vBNlZfC07ABaKBwXK5TocOliDibco7JD7YGvNb0MSpdLir6-elPEUvCwwTiyNEYjTIABMKvCYj4yQAwhBGJxmhSAKydKRrXBq3yikHTDIASSg+VgsThCJgAEEPbF0N8gWsjEYgA)

```sh
--parser typescript
```

**Input:**

```tsx
type Type = {
  // comment
  readonly [T in number]: OtherType[T];
};
```

**Output:**

```tsx
type Type = {
  readonly // comment
  [T in number]: OtherType[T];
};

```

**Expected behavior:**
```tsx
type Type = {
  // comment
  readonly [T in number]: OtherType[T];
};
```

thorn0:
fixed by https://github.com/prettier/prettier/issues/11098

PR #13427, opened 2022-09-04
Fix leading comments in mapped types with `readonly`

thorn0:
## Description

Fixes https://github.com/prettier/prettier/issues/11098

@sosukesuzuki I believe this is a more optimal fix than #11190. There is no need to support comments between the modifier and identifier. I used the tests from your PR.

## Checklist

### DIFF
--- a/changelog_unreleased/typescript/13427.md
+++ b/changelog_unreleased/typescript/13427.md
@@ -1,0 +1,22 @@
+#### Fix leading comments in mapped types with `readonly` (#13427 by @thorn0, @sosukesuzuki)
+
+<!-- prettier-ignore -->
+```tsx
+// Input
+type Type = {
+  // comment
+  readonly [key in Foo];
+};
+
+// Prettier stable
+type Type = {
+  readonly // comment
+  [key in Foo];
+};
+
+// Prettier main
+type Type = {
+  // comment
+  readonly [key in Foo];
+};
+```
--- a/src/language-js/print/type-parameters.js
+++ b/src/language-js/print/type-parameters.js
@@ -16,6 +16,7 @@ import {
   shouldPrintComma,
   getFunctionParameters,
   isObjectType,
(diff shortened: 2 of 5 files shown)


sosukesuzuki (approved):
I forgot about that PR I made myself. Your solution is excellent :+1:

fisker (approved):

bradzacher:
I noticed that this PR got merged into the `next` branch which I assume is 3.0?
If it is 3.0 - is there any way we could cherry-pick this across to the v2 main branch?

It seems a shame to wait until the v3 release to have this bug fixed.

fisker:
Ah, I was reading this part the other day, didn't know it unshipped yet. 

I have to say what a strange AST shape, why `readonly` not on `TSTypeParameter`?

bradzacher:
> why readonly not on TSTypeParameter?

The `TSTypeParameter` specifically represents the bit within the `[]`:
```
{ readonly [ Key in Type ]: Value }
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ TSMappedType
             ^^^^^^^^^^^ .typeParameter -> TSTypeParameter
                            ^^^^^^ .typeAnnotation -> TypeNode
``` 
It's not the parameter that the `readonly` applies to - it's the entire "property". And given that a mapped type may only have one "property" declared in it - really that means the `readonly` applies to the mapped type itself.

It's worth noting that it's also a REALLY weird AST here - `TSTypeParameter` is the *wrong* node for us to emit - it's an abuse of the AST lol. In all other contexts having a `.constraint` on a type parameter means `T extends Constraint`, but in a mapped type instead it means `T in Constraint`.

Should probably change it (filed https://github.com/typescript-eslint/typescript-eslint/issues/6433)

fisker:
@bradzacher released as 2.8.4

PR #13428, opened 2022-09-05
feat(stdin): added ability to prettify std input without pipe

kkyusufk:
## Checklist

### DIFF
--- a/src/cli/format.js
+++ b/src/cli/format.js
@@ -2,6 +2,8 @@
 
 const { promises: fs } = require("fs");
 const path = require("path");
+const readline = require("node:readline");
+const { stdin: input, stdout: output } = require("node:process");
 
 // eslint-disable-next-line no-restricted-modules
 const { default: chalk } = require("../../vendors/chalk.js");
@@ -11,6 +13,8 @@ const prettier = require("../index.js");
 // eslint-disable-next-line no-restricted-modules
 const { getStdin } = require("../common/third-party.js");
 
+// eslint-disable-next-line no-restricted-modules
+const isNonEmptyArray = require("../utils/is-non-empty-array.js");
 const { createIgnorer, errors } = require("./prettier-internal.js");
 const { expandPatterns, fixWindowsSlashes } = require("./expand-patterns.js");
 const getOptionsForFile = require("./options/get-options-for-file.js");
@@ -241,6 +245,21 @@ async function createIgnorerFromContextOrDie(context) {
   }
 }
 
+function getInputFromTerminal() {
+  return new Promise((resolve, reject) => {
+    const inputCode = [];
+    const readLine = readline.createInterface({ input, output });
+    readLine.on("line", (line) => inputCode.push(line));
+    readLine.on("close", () => {
+      if (isNonEmptyArray(inputCode)) {
+        resolve(inputCode.join("\n"));
+      } else {
+        reject("No code inputed");
+      }
+    });
+  });
+}
+
 async function formatStdin(context) {
   const filepath = context.argv.filepath
     ? path.resolve(process.cwd(), context.argv.filepath)
@@ -254,7 +273,7 @@ async function formatStdin(context) {
     : path.relative(process.cwd(), filepath);
 
   try {
-    const input = await getStdin();
+    const input = (await getStdin()) || (await getInputFromTerminal());
 
     if (
       relativeFilepath &&
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -96,7 +96,7 @@ async function main(context) {
 
   const hasFilePatterns = context.filePatterns.length > 0;
   const useStdin =
-    !hasFilePatterns && (!process.stdin.isTTY || context.argv.filePath);
+    !hasFilePatterns && (!process.stdin.isTTY || context.argv.filepath);
 
   if (context.argv.findConfigPath) {
     await logResolvedConfigPathOrDie(context);


src/cli/index.js
```diff
@@ -96,7 +96,7 @@ async function main(context) {
 
   const hasFilePatterns = context.filePatterns.length > 0;
   const useStdin =
-    !hasFilePatterns && (!process.stdin.isTTY || context.argv.filePath);
+    !hasFilePatterns && (!process.stdin.isTTY || context.argv.filepath);
```
kkyusufk, line 99: Out of scope bug fix which was actually causing issues in the if condition on line 105

kkyusufk:
Open question:
With this feature added we can probably make the argument to remove the old way of passing stdin (using the `cat` command). Should we remove it or keep it around?

src/cli/format.js
```diff
@@ -241,6 +245,31 @@ async function createIgnorerFromContextOrDie(context) {
...
+ */
+function getInputFromTerminal() {
+  return new Promise((resolve, reject) => {
+    printToScreen("Start writing your code snippet below");
+    printToScreen("After finishing press ctrl+D to exit the read mode");
+    const inputCode = [];
+    const readLine = readline.createInterface({ input, output });
+    readLine.on("line", (line) => inputCode.push(line));
+    readLine.on("close", () => {
+      if (isNonEmptyArray(inputCode)) {
```
fisker, line 263:
```suggestion
      if (inputCode.lenth > 0) {
```

src/cli/format.js
```diff
@@ -241,6 +245,31 @@ async function createIgnorerFromContextOrDie(context) {
...
+ * on the screen.
+ * @returns {Promise<string>} On success, resolves the input code joined together 
+ * to form a single string. On error, reject with an error message
+ */
+function getInputFromTerminal() {
+  return new Promise((resolve, reject) => {
+    printToScreen("Start writing your code snippet below");
+    printToScreen("After finishing press ctrl+D to exit the read mode");
+    const inputCode = [];
+    const readLine = readline.createInterface({ input, output });
```
fisker, line 260: Are you sure this is correct? Are we piping `stdin` to `stdout`?

src/cli/format.js
```diff
@@ -254,7 +283,7 @@ async function formatStdin(context) {
     : path.relative(process.cwd(), filepath);
 
   try {
-    const input = await getStdin();
+    const input = (await getStdin()) || (await getInputFromTerminal());
```
fisker, line 286: I prefer add another parameter to this function. Use `getInputFromTerminal` only if `--stdin` flag found.

kkyusufk: Will do, my initial approach was using a flag but went with this one for simplicity.
Will add the flag

kkyusufk: add a new cli option `--std-in`

src/cli/format.js
kkyusufk: removed the `output` from the interface.

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,32 @@
+"use strict";
```
kkyusufk, line 1: Removed this function from format file.
For some reason, I was not able to mock this function even if I exported it from `format` file.
Had to extract it to a separate file in order to mock this.

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,32 @@
...
+
+const readline = require("readline");
+const { stdin: input } = require("node:process");
```
fisker, line 4: We should use `stdin` directly.

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,32 @@
...
+function getInputFromTerminal() {
+  return new Promise((resolve, reject) => {
+    printToScreen("Start writing your code snippet below");
+    printToScreen("After finishing press ctrl+D to exit the read mode");
+    const inputCode = [];
+    const readLine = readline.createInterface({ input });
+    readLine.on("line", (line) => inputCode.push(line));
+    readLine.on("close", () => {
+      if (inputCode.length > 0) {
+        printToScreen("\x1b[35m%s\x1b[0m", "\nFormatted code");
```
fisker, line 23: Please use `chalk` is you wand add color.

src/main/core-options.js
```diff
@@ -104,6 +104,16 @@ const options = {
     cliCategory: CATEGORY_OTHER,
     cliDescription: "Path to the file to pretend that stdin comes from.",
   },
+  stdIn: {
```
fisker, line 107: I think we should use `stdin` instead of uppercased

src/cli/index.js
```diff
@@ -55,6 +55,12 @@ async function main(context) {
     }
   }
 
+  if (context.argv.stdIn && !context.argv.filepath) {
```
fisker, line 58: I think use `--parser` is fine too.

fisker (changes requested):
Overall  looks goood, I'll take a deeper look when back to laptop.

src/cli/input-from-terminal.js
kkyusufk: Made all the changes except this one.
I do not understand this comment.

kkyusufk:
@fisker Please take a look again whenever you get the time
Only one comment is left to address which I do not understand 😅

src/cli/input-from-terminal.js
fisker:
```suggestion
const { stdin } = require("node:process");
```

fisker:
https://github.com/prettier/prettier/pull/13428#discussion_r973593191

src/cli/index.js
```diff
@@ -55,6 +55,12 @@ async function main(context) {
     }
   }
 
+  if (context.argv.stdin && !context.argv.filepath && !context.argv.parser) {
+    throw new Error(
+      "Cannot use --stdin without also passing --stdin-filepath or --parser"
```
fisker, line 60:
```suggestion
      "'--stdin-filepath' or '--parser' is required when using '--stdin'"
```

src/main/core-options.js
```diff
@@ -104,6 +104,16 @@ const options = {
     cliCategory: CATEGORY_OTHER,
     cliDescription: "Path to the file to pretend that stdin comes from.",
   },
+  stdin: {
+    since: "2.8.0",
+    category: CATEGORY_SPECIAL,
+    type: "boolean",
+    default: false,
+    description:
+      "Starts accepting std input from terminal. Should be used with stdin-filepath option",
```
fisker, line 113: Please update the description

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,32 @@
...
+function getInputFromTerminal() {
+  return new Promise((resolve, reject) => {
+    printToScreen("Start writing your code snippet below");
+    printToScreen("After finishing press ctrl+D to exit the read mode");
+    const inputCode = [];
+    const readLine = readline.createInterface({ input });
+    readLine.on("line", (line) => inputCode.push(line));
+    readLine.on("close", () => {
+      if (inputCode.length > 0) {
+        printToScreen("Formatted code");
```
fisker, line 23: This should print after the code if formatted.

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * This function starts reading any input which user types on the console.
+ * After the user clicks ctrl+D to exit, the input is formatted and printed
+ * on the screen.
+ * @returns {Promise<string>} On success, resolves the input code joined together
+ * to form a single string. On error, reject with an error message
+ */
+function getInputFromTerminal() {
+  return new Promise((resolve, reject) => {
+    printToScreen("Start writing your code snippet below");
```
fisker, line 16: Let's print the message before calling this function?

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,32 @@
...
+    printToScreen("After finishing press ctrl+D to exit the read mode");
+    const inputCode = [];
+    const readLine = readline.createInterface({ input });
+    readLine.on("line", (line) => inputCode.push(line));
+    readLine.on("close", () => {
+      if (inputCode.length > 0) {
+        printToScreen("Formatted code");
+        resolve(inputCode.join("\n"));
+      } else {
+        reject("Nothing was typed");
```
fisker, line 26: Let's resolve the promise with empty string instead of rejecting.

src/cli/input-from-terminal.js
kkyusufk: sounds good to me.

src/cli/input-from-terminal.js
kkyusufk: so you mean something like this
```console
prettier --stdin --parser babel
Start writing your code snippet below
After finishing press ctrl+D to exit the read mode
hello

hello;
Formatted code
```

kkyusufk: Pushed changes with the above pattern.

fisker: Sorry, I mean formatted, but  before print.

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,28 @@
...
+function getInputFromTerminal() {
+  return new Promise((resolve) => {
+    const inputCode = [];
+    const readLine = readline.createInterface({ input: stdin });
+    readLine.on("line", (line) => inputCode.push(line));
+    readLine.on("close", () => {
+      if (inputCode.length > 0) {
+        resolve(inputCode.join("\n"));
+      } else {
+        resolve("");
```
fisker, line 22: Meaningless if-else

src/cli/input-from-terminal.js
kkyusufk: I am sorry, I still do not get it. Can you drop a suggestion please?

src/cli/input-from-terminal.js
kkyusufk: 🤦🏻 should have seen it. Will update

src/cli/input-from-terminal.js
fisker: https://github.com/prettier/prettier/pull/13428/commits/d856d63a9acfefc12586d402589b8af3cfbda3c6

src/cli/format.js
```diff
@@ -252,14 +253,20 @@ async function formatStdin(context) {
...
     : path.relative(process.cwd(), filepath);
-
   try {
-    const input = await getStdin();
-
+    let input;
+    if (context.argv.stdin) {
+      printToScreen("Start writing your code snippet below");
+      printToScreen("After finishing press ctrl+D to exit the read mode");
+      input = await getInputFromTerminal();
```
fisker, line 261: I'm not near a laptop, will <kbd>Ctrl + C</kbd> reject the Promise? Do we need handle it?

kkyusufk: Do not think we need to handle that scenario. 
Pressing `Ctrl + C` exits the read mode without any error.

kkyusufk: https://user-images.githubusercontent.com/22116082/190865506-23ce198d-4c84-4faf-b411-d72d376f7d54.mov

fisker: 👍

kkyusufk:
@fisker ready for another round. Should be close to approval by now 🤞🏻

src/cli/input-from-terminal.js
```diff
@@ -0,0 +1,26 @@
...
+ * to form a single string. On error, reject with an error message
+ */
+function getInputFromTerminal() {
+  return new Promise((resolve) => {
+    const inputCode = [];
+    const readLine = readline.createInterface({ input: stdin });
+    readLine.on("line", (line) => {
+      inputCode.push(line);
+    });
+    readLine.on("close", () => {
```
fisker, line 20: Can't exit with <kbd>Ctrl</kbd> + <kbd>D</kbd>, I also tried [<kbd>Ctrl</kbd> + <kbd>Z</kbd>](https://stackoverflow.com/questions/16136400/how-to-send-eof-via-windows-terminal), can't either.

src/cli/input-from-terminal.js
fisker: This can't be mocked in production, we need add it to `src/common/third-party.js`(you can rename it to `modules-to-mock.js`) and use it from there.

kkyusufk:
Hey @fisker , apologies for neglecting this PR for a long time.
Please review, I fixed the issue on windows.

src/cli/input-from-terminal.js
kkyusufk: Fixed in [8d2872f6de17859b5d474cff88c1b1806b3895a6](https://github.com/prettier/prettier/pull/13428/commits/8d2872f6de17859b5d474cff88c1b1806b3895a6)

kkyusufk:
@fisker pinging for another review.

fisker:
Sorry for missing your comment. Will try again soon.

fisker:
It seems I have to press <kbd>Ctrl</kbd> + <kbd>D</kbd> on a new line to exit. And <kbd>↑</kbd> <kbd>↓</kbd> doesn't work as expected.

kkyusufk:
@fisker Readline only registers a line after we press enter. 
Seems like this will be a bad DX,
I will try to iterate on this. Let's see if there is anything we can use. 

If not, maybe we just add those instructions,
Pressing Ctrl+D after pressing enter will format.

PR #13429, opened 2022-09-05
Update browser docs

### LINKED ISSUE #13413: Add more functions to browser versions
I'd like to have the following functions in the browser bundle so we can use them in the [Browser Extension](https://github.com/prettier/prettier-browser-extension):

- [`check`](https://prettier.io/docs/en/api.html#prettierchecksource--options)
- [`formatWithCursor`](https://prettier.io/docs/en/api.html#prettierformatwithcursorsource--options)
- [`clearConfigCache`](https://prettier.io/docs/en/api.html#prettierclearconfigcache)
- [`getFileInfo`](https://prettier.io/docs/en/api.html#prettiergetfileinfofilepath--options)
- [`getSupportInfo`](https://prettier.io/docs/en/api.html#prettiergetsupportinfo)

**Environments:**

- Prettier Version: 2.7.1
- Runtime: Chrome/Firefox 
- Operating System: Any 
- Prettier plugins (if any): All

nickmccurdy:
## Description

Adds missing functions to the browser docs to fix #13413 and makes docs more consistent

## Checklist

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -9,9 +9,15 @@ If you want to run Prettier programmatically, check this page out.
 const prettier = require("prettier");
 ```
 
+## Options
+
+- **[`parser`](options.md#parser) (or [`filepath`](options.md#file-path))**: One of these options should be specified for Prettier to know which parser to use. They will be required in a future release.
+
+Other [options](options.md) may be provided to override the defaults.
+
 ## `prettier.format(source [, options])`
 
-`format` is used to format text using Prettier. `options.parser` should be set according to the language you are formatting (see the [list of available parsers](options.md#parser)). Alternatively, `options.filepath` can be specified for Prettier to infer the parser from the file extension. Other [options](options.md) may be provided to override the defaults.
+`format` is used to format text using Prettier.
 
 ```js
 prettier.format("foo ( );", { semi: false, parser: "babel" });
--- a/docs/api.md
+++ b/docs/api.md
@@ -9,9 +9,9 @@ If you want to run Prettier programmatically, check this page out.
 const prettier = require("prettier");
 ```
 
-## `prettier.format(source, options)`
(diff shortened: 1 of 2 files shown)


nickmccurdy:
I thought it would be cleaner to just link from the browser APIs to the Node APIs, since they have the same use cases and similar usage. Alternatively, we could copy the descriptions to the API package so you wouldn't have to go back and forth, but I feel like that would be more effort to maintain.

Maybe in the future we could just have a single API page with details for different versions, like the module section we already have for the browser docs.

docs/api.md
```diff
@@ -33,7 +39,7 @@ prettier.formatWithCursor(" 1", { cursorOffset: 2, parser: "babel" });
 // -> { formatted: '1;\n', cursorOffset: 1 }
 ```
 
-## `prettier.resolveConfig(filePath [, options])`
+## `prettier.resolveConfig([filePath, options])`
```
fisker, line 42: I don't think `filePath` is optional.

docs/api.md
```diff
@@ -9,9 +9,15 @@ If you want to run Prettier programmatically, check this page out.
 const prettier = require("prettier");
 ```
 
-## `prettier.format(source, options)`
+## Options
 
-`format` is used to format text using Prettier. `options.parser` must be set according to the language you are formatting (see the [list of available parsers](options.md#parser)). Alternatively, `options.filepath` can be specified for Prettier to infer the parser from the file extension. Other [options](options.md) may be provided to override the defaults.
+- **[`parser`](options.md#parser) (or [`filepath`](options.md#file-path))**: One of these options should be specified for Prettier to know which parser to use. They will be required in a future release.
```
fisker, line 14: This is not currect, wen can also infer parser from hashbang.

fisker (changes requested):

PR #13445, opened 2022-09-08
Support nestled JSDoc comments

### LINKED ISSUE #12653: Nestled JSDoc comments are still broken after #9672

thorn0:
## Description

Fixes #12653

## Checklist


tests/format/js/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3144,6 +3049,115 @@ function HelloWorld() {
...
+function value(type, value) {
+  if (arguments.length === 2) {
+    return new ConcreteValue(type, value);
+  } else {
+    return new Value(type);
+  }
+}
+
+/** Trailing nestled comment 1
+ */ /** Trailing nestled comment 2 */
```
fisker, line 3156: Should they stick together?

thorn0: Depends on whether we need to support this for single-line comments.

fisker (approved):

fisker:
Maybe easier to merge them into one comment after parse?

thorn0:
The code for printing dangling comments needs to be refined anyway. Now it's very basic. It prints every comment on a new line and doesn't preserve blank lines between comments. Instead its logic should be similar to how trailing comments are printed. But it's something for a separate PR. As for this PR, what if we just skip this feature for dangling comments for now? I'll just add a check that dangling comments can't be considered nestled.

fisker:
Any comment on this idea? https://github.com/prettier/prettier/pull/13445?notification_referrer_id=NT_kwDOAAKiKLE0MzU3NzEzMDk4OjE3MjU4NA#issuecomment-1240689242

thorn0:
IDK. Merging would work too, but it's difficult to say if it's going to make things simpler.

fisker:
> it's difficult to say if it's going to make things simpler.

Shouldn't it work without any change to the print part?

thorn0:
For the print part - yes. I thought merging itself might be a bit tricky, but I may be wrong.

fisker:
Aren't they looks like one comment? Would like to try?

thorn0:
Yes, I'm going to try that approach.

src/language-js/parse/postprocess/index.js
```diff
@@ -154,6 +157,29 @@ function postprocess(ast, options) {
...
+  if (isNonEmptyArray(ast.comments)) {
+    for (let i = ast.comments.length - 2; i >= 0; i--) {
+      const comment = ast.comments[i];
+      const nextComment = ast.comments[i + 1];
+      if (
+        isBlockComment(comment) &&
+        isIndentableBlockComment(comment) &&
+        isBlockComment(nextComment) &&
+        isIndentableBlockComment(nextComment) &&
+        locEnd(comment) === locStart(nextComment)
```
fisker, line 170: `loc` compare should be faster than `isIndentableBlockComment` check, let's move this condition up

fisker:
**Size Change:** +605 kB (+6%) 🔍

**Total Size:** 11.5 MB

| Filename | Size | Change |  |
| :--- | :---: | :---: | :---: |
| `./dist/esm/parser-babel.mjs` | 368 kB | +61.5 kB (+20%) | 🚨 |
| `./dist/esm/parser-espree.mjs` | 212 kB | +61.8 kB (+41%) | 🚨 |
| `./dist/esm/parser-flow.mjs` | 1.11 MB | +63.2 kB (+6%) | 🔍 |
| `./dist/esm/parser-meriyah.mjs` | 221 kB | +61.8 kB (+39%) | 🚨 |
| `./dist/esm/parser-typescript.mjs` | 1.49 MB | +54.4 kB (+4%) |  |
| `./dist/parser-babel.js` | 369 kB | +61.5 kB (+20%) | 🚨 |
| `./dist/parser-espree.js` | 213 kB | +61.8 kB (+41%) | 🚨 |
| `./dist/parser-flow.js` | 1.11 MB | +63.2 kB (+6%) | 🔍 |
| `./dist/parser-meriyah.js` | 221 kB | +61.8 kB (+39%) | 🚨 |
| `./dist/parser-typescript.js` | 1.49 MB | +54.4 kB (+4%) |  |

thorn0:
> Maybe we should extract isIndentableBlockComment

You mean to a separate file?

thorn0:
BTW, I agree with this opinion on default export https://mobile.twitter.com/stefanjudis/status/1566047068510093315 :)

src/language-js/parse/postprocess/index.js
```diff
@@ -154,6 +157,31 @@ function postprocess(ast, options) {
...
+        isIndentableBlockComment(comment) &&
+        isBlockComment(followingComment) &&
+        isIndentableBlockComment(followingComment)
+      ) {
+        ast.comments.splice(i + 1, 1);
+        followingComment = ast.comments[i] = {
+          ...comment,
+          value: comment.value + "*//*" + followingComment.value,
+          range: [locStart(comment), locEnd(followingComment)],
+        };
```
fisker, line 177: We already mutate the array, let's mutate the comment directly?

src/language-js/parse/postprocess/index.js
```diff
@@ -154,6 +157,27 @@ function postprocess(ast, options) {
...
+  if (isNonEmptyArray(ast.comments)) {
+    let followingComment = getLast(ast.comments);
+    for (let i = ast.comments.length - 2; i >= 0; i--) {
+      const comment = ast.comments[i];
+      if (
+        locEnd(comment) === locStart(followingComment) &&
+        isBlockComment(comment) &&
+        isIndentableBlockComment(comment) &&
+        isBlockComment(followingComment) &&
+        isIndentableBlockComment(followingComment)
```
fisker, line 170:
```suggestion
        isBlockComment(comment) &&
        isBlockComment(followingComment) &&
        isIndentableBlockComment(comment) &&
        isIndentableBlockComment(followingComment)
```

Let's move `isBlockComment` check up, to avoid split big comments.

.eslintrc.cjs
```diff
@@ -329,7 +330,6 @@ module.exports = {
             functions: ["hasComment", "getComments"],
           },
           "src/language-js/pragma.js",
-          "src/language-js/parse/postprocess/*.js",
```
thorn0, line 332: This didn't work because the rules doesn't resolve globs.

PR #13447, opened 2022-09-08
Refactor: use path.getValue() everywhere instead of getNode()

thorn0:
## Checklist


fisker:
I say we should remove `getNode` it's confusing. Also  I don't understand why `getParent` skipping falsely values. I call `path.each(() = path.getParent(), "elements")` on array, if it'sa hole, am I getting different result?

fisker (approved):

PR #13448, opened 2022-09-08
Add getters to path object

thorn0:
## Description

This PR adds getters to the path object that can be used, in particular, for convenient destructuring. See https://github.com/prettier/prettier/pull/10565#issuecomment-1231520272. Also, this PR uses the added getters in the Markdown printer.

## Checklist

### DIFF
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -1,6 +1,5 @@
 import assert from "node:assert";
 import getLast from "../utils/get-last.js";
-import { getPenultimate } from "./util.js";
 
 function getNodeHelper(path, count) {
   const stackIndex = getNodeStackIndexHelper(path.stack, count);
@@ -23,7 +22,21 @@ class AstPath {
   }
 
   get key() {
-    return getPenultimate(this.stack) || null;
+    const { stack } = this;
+    const { length } = stack;
+    let key = stack[length - 2];
+    if (typeof key === "number") {
+      assert(Array.isArray(stack[length - 3]));
+      key = stack[length - 4];
+    }
+    return key;
+  }
+
+  get index() {
+    const { stack } = this;
+    const { length } = stack;
+    const index = stack[length - 2];
+    return typeof index === "number" ? index : null;
   }
 
   get node() {
@@ -44,6 +57,14 @@ class AstPath {
 
   get previous() {
     return this.#getSiblingNode(-1);
+  }
+
+  get siblings() {
+    const { stack } = this;
+    const { length } = stack;
+    const array = stack[length - 3];
+    assert(Array.isArray(array));
+    return array;
   }
 
   // The name of the current property is always the penultimate element of
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -1,4 +1,5 @@
 import getLast from "../utils/get-last.js";
+import { getPenultimate } from "./util.js";
 
 function getNodeHelper(path, count) {
   const stackIndex = getNodeStackIndexHelper(path.stack, count);
@@ -18,6 +19,30 @@ function getNodeStackIndexHelper(stack, count) {
 class AstPath {
   constructor(value) {
     this.stack = [value];
+  }
+
+  get name() {
+    return getPenultimate(this.stack) || null;
+  }
+
+  get node() {
+    return getLast(this.stack);
+  }
+
+  get parent() {
+    return this.getNode(1);
+  }
+
+  get grandparent() {
+    return this.getNode(2);
+  }
+
+  get next() {
+    return this.#getSiblingNode(1);
+  }
+
+  get previous() {
+    return this.#getSiblingNode(-1);
   }
 
   // The name of the current property is always the penultimate element of
@@ -46,6 +71,23 @@ class AstPath {
 
   getParentNode(count = 0) {
     return getNodeHelper(this, count + 1);
+  }
+
+  #getSiblingNode(offset) {
+    const { stack } = this;
+    const { length } = stack;
+    if (length < 3) {
+      return null;
+    }
+    const number = stack[length - 2];
+    if (typeof number !== "number") {
+      return null;
+    }
+    const array = stack[length - 3];
+    if (!Array.isArray(array)) {
+      return null;
+    }
+    return array[number + offset];
   }
 
   // Temporarily push properties named by string arguments given after the
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -77,7 +77,7 @@ class AstPath {
   #getSiblingNode(offset) {
     const { stack } = this;
     const { length } = stack;
-    assert(length < 3);
+    assert(length >= 5);
     const index = stack[length - 2];
     assert(typeof index === "number");
     const array = stack[length - 3];
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -1,21 +1,6 @@
 import assert from "node:assert";
 import getLast from "../utils/get-last.js";
 
-function getNodeHelper(path, count) {
-  const stackIndex = getNodeStackIndexHelper(path.stack, count);
-  return stackIndex === -1 ? null : path.stack[stackIndex];
-}
-
-function getNodeStackIndexHelper(stack, count) {
-  for (let i = stack.length - 1; i >= 0; i -= 2) {
-    const value = stack[i];
-    if (value && !Array.isArray(value) && --count < 0) {
-      return i;
-    }
-  }
-  return -1;
-}
-
 class AstPath {
   constructor(value) {
     this.stack = [value];
@@ -52,11 +37,11 @@ class AstPath {
   }
 
   get next() {
-    return this.#getSiblingNode(1);
+    return this.siblings[this.index + 1];
   }
 
   get previous() {
-    return this.#getSiblingNode(-1);
+    return this.siblings[this.index - 1];
   }
 
   get siblings() {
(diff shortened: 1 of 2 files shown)


src/common/ast-path.js
```diff
@@ -48,6 +73,23 @@ class AstPath {
...
 
+  #getSiblingNode(offset) {
+    const { stack } = this;
+    const { length } = stack;
+    if (length < 3) {
+      return null;
+    }
+    const number = stack[length - 2];
+    if (typeof number !== "number") {
+      return null;
```
fisker, line 84: Should we throw errors, when getting siblings if it's not in array?

src/common/ast-path.js
```diff
@@ -20,6 +21,30 @@ class AstPath {
     this.stack = [value];
   }
 
+  get name() {
```
fisker, line 24: I know we used "getName" before, but do you think `key` is a better choice?

src/common/ast-path.js
```diff
@@ -20,6 +21,30 @@ class AstPath {
...
+    return getPenultimate(this.stack) || null;
+  }
+
+  get node() {
```
fisker, line 28: Use `value`?

src/common/ast-path.js
```diff
@@ -20,6 +21,30 @@ class AstPath {
...
+    return getLast(this.stack);
```
fisker, line 29: Node.js 16 should already support `.at(-1)` but I guess esbuild only transform new syntax but don't polyfill methods.

src/common/ast-path.js
thorn0: This distinction value vs node is not practical. Node is a much better word here.

src/common/ast-path.js
thorn0: I like `key` better too. Let's rename it.

src/common/ast-path.js
fisker: If so, let's change remove old `.getValue()` change `.getNode()` to current `.getValue()`?

src/common/ast-path.js
thorn0: Let's add assertions.

thorn0: After all, `map` and `each` don't check `isArray`.

src/common/ast-path.js
```diff
@@ -48,6 +74,17 @@ class AstPath {
     return getNodeHelper(this, count + 1);
   }
 
+  #getSiblingNode(offset) {
+    const { stack } = this;
+    const { length } = stack;
+    assert(length < 3);
```
fisker, line 80: Shouldn't this at least `5`?

src/common/ast-path.js
```diff
@@ -48,6 +74,17 @@ class AstPath {
...
+    const number = stack[length - 2];
```
fisker, line 81: Use `index`?

src/common/ast-path.js
thorn0: IDK. Can't the root of the AST be an array?

fisker:
Can we add a `isInArray()` so we don't need use `typeof key` to do that?

After all you can access elements in array with `array["0"]`.

src/common/ast-path.js
thorn0: `getNode` has a parameter. Also that might break plugins if, unlike Prettier, they happen to point the path to arrays. I thought we would just deprecate `getName` and `getValue`.

fisker: Okay, never mind, I guess we'll stop using them. I try to change them to getters long time ago, that PR was maybe two years ago.

thorn0:
> Can we add a isInArray() so we don't need use typeof key to do that?

I wanted to keep it as lightweight as possible. Do you want to support string keys? How exactly do you see that?

src/common/ast-path.js
thorn0: Okay, changed it to 5 to make the check stricter.

src/common/ast-path.js
thorn0: Found it! https://github.com/prettier/prettier/pull/7505

fisker:
I don't want use string key, but I want have a explicit way to check it's in array. But we can do this in another PR.  Speaking of checking arrays, maybe we can make `.key` getter always return the key accessing from parent, but `.index` getter return the index when it's in array?

fisker:
Maybe also `.siblings` getters, :smile:

src/common/ast-path.js
```diff
@@ -20,6 +21,52 @@ class AstPath {
...
+  get grandparent() {
+    return this.getNode(2);
+  }
+
+  get next() {
+    return this.#getSiblingNode(1);
+  }
+
+  get previous() {
+    return this.#getSiblingNode(-1);
```
fisker, line 59: Use `this.siblings[this.index - 1]`

fisker:
Looks great!

fisker:
Only thing I don't like is we still getting the wrong parent if the "node" is falsely, I hope we change the getter later to behavior differently with `.getParent` before we make those publish.

thorn0:
> we still getting the wrong parent if the "node" is falsely

Do you think this can happen?

fisker:
Yes, https://github.com/prettier/prettier/blob/87d724f418acefbf949415c9ac32c77dc5f59ac9/src/main/create-print-pre-check-function.js#L40

thorn0:
But how can there be a falsy value in the middle of the stack? Falsy values can't have properties.

fisker (approved):
This is kind of big change, let's wait for @sosukesuzuki to review.

fisker:
Middle one can't but last one can `path.call(, "optionalChildKey")`

thorn0:
What if we just remove that truthiness check? How breaking can it be?

fisker:
Another case is `path.call(, "children")` things like array holes, but not sure how this case behave.

fisker:
> What if we just remove that truthiness check? How breaking can it be?

I don't know, but at least we can make `.parent` correct.

thorn0:
Let's do it then.

fisker:
You already did more, why not add `.isInArray` then, unless you don't think it's a good idea.

fisker:
We also check `isLast` a lot... let's add two more? `isFirst` `isLast`

fisker:
`.at` is coming so `this.siblings.at(-1)` should already simple enough, not worth add `.last`

src/common/ast-path.js
```diff
@@ -1,35 +1,79 @@
...
+  get siblings() {
+    const { stack } = this;
+    const { length } = stack;
+    const array = stack[length - 3];
+    assert(Array.isArray(array));
+    return array;
+  }
+
+  get isInArray() {
+    return this.index !== null;
```
fisker, line 56: I was hoping `Array.isArray[this.stack.length - 3]`, and other getters call this one first. More reliable.

src/common/ast-path.js
```diff
@@ -1,35 +1,82 @@
...
-function getNodeStackIndexHelper(stack, count) {
-  for (let i = stack.length - 1; i >= 0; i -= 2) {
-    const value = stack[i];
-    if (value && !Array.isArray(value) && --count < 0) {
-      return i;
+  get key() {
+    const { stack } = this;
+    const { length } = stack;
+    let key = stack[length - 2];
+    if (typeof key === "number") {
```
fisker, line 13: Let's use `.isInArray` here.

src/common/ast-path.js
```diff
@@ -1,35 +1,82 @@
...
-class AstPath {
-  constructor(value) {
-    this.stack = [value];
+  get index() {
+    const { stack } = this;
+    const { length } = stack;
+    const index = stack[length - 2];
+    return typeof index === "number" ? index : null;
+  }
+
```
fisker, line 26: Let's use `.isInArray` here.

fisker (approved):
Great work!

I'm a little bit prefer `undefined` over `null`, but not important.

fisker:
`.isRoot`?

sosukesuzuki (approved):

src/common/ast-path.js
```diff
@@ -1,35 +1,75 @@
...
+  get key() {
+    const { stack, isInArray } = this;
+    return stack[stack.length - (isInArray ? 4 : 2)] ?? null;
+  }
+
+  get index() {
+    assert(this.isInArray);
+    return this.isInArray ? getPenultimate(this.stack) : null;
```
thorn0, line 22: Let's just let it return `null` if not in an array. This might make destructuring simpler: `const { key, index, node } = path;`.

fisker: I was hoping check `.isInArray` first if not sure.

fisker: But your point make sense.

thorn0: That'd make code more verbose. E.g. we want be able to do this:
```js
function hasPrettierIgnore(path) {
  return path.index > 0 && isPrettierIgnore(path.previous) === "next";
}
```

thorn0:
Overall, getters that throw feel unnatural, don't they? E.g. DOM getters never throw. Maybe our getters should return null instead, just like the DOM ones.

thorn0:
Mainly because DOM uses `null`.

thorn0:
On the other hand, DOM can't contain nulls whereas ASTs do have them, so I don't insist.

fisker:
I prefer `undefined`, `typeof` check is also easier. I implemented [`unicorn/no-null`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-null.md) 😄 

But I respect your choice.

fisker:
I'll be on vacation, feel free to make change and merge.

thorn0:
I decided to go with `null` after all because `getParentNode` returns it too.

PR #13474, opened 2022-09-13
Simplify print pre-check with `AstPath` getters

fisker:
## Checklist

### DIFF
--- a/src/common/ast-path.js
+++ b/src/common/ast-path.js
@@ -7,8 +7,8 @@ class AstPath {
   }
 
   get key() {
-    const { stack, siblings } = this;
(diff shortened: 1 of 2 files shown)


src/common/ast-path.js
```diff
@@ -7,8 +7,8 @@ class AstPath {
   }
 
   get key() {
-    const { stack, siblings } = this;
-    return stack[stack.length - (siblings === null ? 4 : 2)] ?? null;
```
thorn0, line 11: The overall style of this file is to prefer micro-optimizations where possible, so let's stay consistent and not change this.

thorn0 (approved):

PR #13511, opened 2022-09-23
Rename `bin-prettier.js` bundle to `bin-prettier.cjs`

fisker:
## Description

My idea is to use 
- `.cjs` for CommonJS modules, `bin-prettier.cjs` `index.cjs`
- `.mjs` for ESM modules, `cli.mjs` `index.mjs` `third-party.mjs`(#13509) `esm/*`
- `.js` for UMD modules, `standalone.js`, `parser-*.js`

FYI:
I'm planning to 

- move `esm/parser-*` and `parser-*` to `plugins/`
- move `third-party.mjs*`  to `interal/`

So there will be

`babel.js` and `babel.mjs` in `plugins/`, should we use `.common.cjs` / `.esm.mjs` / `.umd.js` to be more clear?

## Checklist

### DIFF
--- a/.github/workflows/prod-test.yml
+++ b/.github/workflows/prod-test.yml
@@ -133,4 +133,4 @@ jobs:
         run: node -v | grep "v0.10.48" || exit 1
 
       - name: Run CLI on Node.js v0.10.48
-        run: node dist/bin-prettier --version 2>&1 >/dev/null | grep "prettier requires at least version 14 of Node, please upgrade" || exit 1
+        run: node dist/bin-prettier.cjs --version 2>&1 >/dev/null | grep "prettier requires at least version 14 of Node, please upgrade" || exit 1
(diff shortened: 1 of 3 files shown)


thorn0:
> should we use .common.cjs / .esm.mjs / .umd.js to be more clear?

I don't think this is needed. What's needed is some very short and noticeable documentation that describes what's included in the `prettier` package. Even now it's causing confusion as people don't realize that the package includes a CLI tool, a Node lib, and a standalone lib (bundled in different module formats). Would be good to make it more clear.

thorn0 (approved):

fisker:
Since we don't need the UMD version in Node.js anymore, maybe it's time to remove standalone version to a separate package? The standalone version still need both ESM and UMD version.

thorn0:
Maybe. I thought there was an issue with a discussion about splitting the package, but I can't find it.

fisker:
Closed https://github.com/prettier/prettier/issues/10157

PR #13519, opened 2022-09-25
Fix `parse-srcset` checksum in `yarn.lock`

mataha:
## Description

When trying to install dependencies on the `next` branch, Yarn fails with an error related to one of its dependencies' remote archive not matching the expected checksum in `yarn.lock`. This behaviour occurs because the default for `checksumBehavior` is set to [`throw`](https://yarnpkg.com/configuration/yarnrc#checksumBehavior). Doing a one-time update fixes that (as I don't believe continuous updates should be a thing).

This is equivalent to running `YARN_CHECKSUM_BEHAVIOR=update yarn`.

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6376,7 +6376,7 @@ __metadata:
 "parse-srcset@ikatyang/parse-srcset#54eb9c1cb21db5c62b4d0e275d7249516df6f0ee":
   version: 1.0.2
   resolution: "parse-srcset@https://github.com/ikatyang/parse-srcset.git#commit=54eb9c1cb21db5c62b4d0e275d7249516df6f0ee"
-  checksum: 4f3325a79e8d881d1d29ae4d2857a40b1d9d8d98f093b865a04a4e18ba5b136e7e1d37a317a8fc2ed10a8f473951c396f7c1b0904fe10de430e44811698ce0f1
+  checksum: 1f4b1e48977ed0c47a357e8f484792ca71007c87df33dc3b9c806c1b096162e9e7ba41f5517ba16c2fddce7c6754016692f4b673b4c0960241508671ac8dcebe
   languageName: node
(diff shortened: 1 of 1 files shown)


thorn0:
> When trying to install dependencies on the next branch, Yarn fails with an error related to one of its dependencies' remote archive not matching the expected checksum in yarn.lock.

I can't reproduce this.

fisker:
I can reproduce either, but mentioned in other issue too https://github.com/prettier/prettier/issues/12209#issuecomment-1239572266 not sure what's the reason.

mataha:
> I can't reproduce this.

Do you have `YARN_CHECKSUM_BEHAVIOR` set globally to `ignore`? Alternatively, is there a parent `.yarnrc.yml` file with `checksumBehavior: "ignore"` within it? If that's not the case, then the package might be cached:

```sh
git checkout next
yarn cache clean --all
yarn install
```

One other case that comes to my mind is line ending incompatibility when generating checksums for dependency protocols other than semver (I develop on Windows, so CRLF for me) - is there anyone working on Prettier facing the same problem who's also on Windows? Yarn had a bug similar to that before.

thorn0:
Can't reproduce this on Windows, even with `core.autocrlf` set to `true`.

thorn0:
I guess it's time to republish that fork of `parse-srcset`.

mataha:
What about my branch? Does it install correctly, or are you getting the same errors I do on `next`? For reference, here's what's happening on my end:

https://user-images.githubusercontent.com/7210216/192166681-203d906c-b6d3-442d-9b40-819219029b97.mp4

thorn0:
> or are you getting the same errors I do on next?

this

thorn0:
Superseded by https://github.com/prettier/prettier/pull/13536

PR #13532, opened 2022-09-25
Fix unstable embedded template literals

### LINKED ISSUE #6150: Unstable template literals marked with /* HTML */ and /* GraphQL */
Happens if the template literal initially doesn't contain line breaks.

Another issue about unstable template literals: #5885

### HTML in JS ###

**Prettier 1.17.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzCEAUB6AVAAgAkAVAWQBl9dt8ADAHgEIBaZ-AIwEMAnARn1YA+Dj3xNWI7gCYBzQbQCUAbhAAaEBAAOMAJbQAzslA9uEAO4AFHgkMpOANwg6AJmpDtunMAGs4MAMqaXjpQAObIMNwArnDqABYwALYANgDqcTrw+kFgcP42mTr2mQCeyOD6huoh+nDcMBaeoYmcyKicybXqAFb6AB4AQp4+fv6ciXDkIXBtHV0gvX3+IaHJcACKURDws52xIEHctdzlXOwlydBumtwhMKkuMHHIABwADOo3ELWpnprlNzgx3sM3UAEctvBGlpbCBOPpmFA4HBnCi3Nw4BCdBjGpxmq0kO09upaokdBFovt9Cs1pttjNCXN9jBOOwHs4nsgpOpIpwdMkVgBhCCJFrlKDQUEgKK1Yis2xE2oAXyVQA)
```sh
--parser babylon
```

**Input:**
```jsx
foo(/* HTML */ `<!-- bar1 --> bar <!-- bar2 -->`);
```

**Output:**
```jsx
foo(
  /* HTML */ `
    <!-- bar1 -->
    bar
    <!-- bar2 -->
  `
);

```

**Second Output:**
```jsx
foo(/* HTML */ `
  <!-- bar1 -->
  bar
  <!-- bar2 -->
`);

```

### GraphQL in JS ###

**Prettier 1.17.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzCEAUB6AVAAgHEAnAQwAcALARQBl9dt8ADARwFc5iBPfYfdBD74ARqWL4AvlOYBKANwgANCAjkYAS2gBnZKHHEIAdwAK4hLpSkAbhA0ATZSBFkwAazgwAyuVJgNUADmyDDEnCqUMAC2ADYA6pQa8Nq+YHBeFkka1kncyODauioB2lwwJmSBUaTIqKQxpSoAVtoAHgBCrh7epFFwtAFwtfWNIC2tXgGBMXDU7BDwww1wKr7EpcT5YiLcMdBO5MQBMHEOMJTIABwADKuGpXFk5PmHcBvWQyocC3AVapYgUjaAC0UDgcHsEKcxDgHA0MIqpCqNSQdWWKlKUQ0ITCKxA2imMzmPyWoxgpBEp3s52QACYVKFSBoYlMAMIQKLVfJQaCfEDsUoAFQpljRpUkkiAA)
```sh
--parser babylon
```

**Input:**
```jsx
foo(/* GraphQL */ `query { foo { bar } }`);
```

**Output:**
```jsx
foo(
  /* GraphQL */ `
    query {
      foo {
        bar
      }
    }
  `
);

```

**Second Output:**
```jsx
foo(/* GraphQL */ `
  query {
    foo {
      bar
    }
  }
`);

```

### LINKED ISSUE #11205: Feature request: expand on label builder
With version 2.3.0 the builder [label](https://github.com/prettier/prettier/blob/main/commands.md#label) was introduced.
This has helped a lot for `doc` analysis and making printing choices in our plugin. But the signature says that label should be a `string` and we are using `JSON.stringify()` and `JSON.parse()` to pass additional data.

It would be great if the `label` specification could also expand to support metadata or have a dedicated `metadata` builder that behaves the same as label but receives an object.

**Expected behavior:**

I could call `label` like

```JavaScript
label({ label: 'foo', ...additionalData }, doc);
// or
metadata(additionalData, doc);
```

**Actual behavior:**

I'm only capable to call

```JavaScript
label('foo', doc);
```

thorn0:
## Description

Fixes #6150
Fixes #11205

Template literals with embedded languages are hugged now if they're the only argument and have leading and trailing whitespace. The idempotence issue has been fixed.

## Checklist

**✨[Try the playground for this PR](https://deploy-preview-13532--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAzCEAUAdKACPAegCo8AJAFQFkAZPYwvAAwB4BCAWg7wCMBDAE4BGPFwB8uAgX4DJU9l16CATKI5imuAJQBuEABoQEAA4wAltADOyUIIEQA7gAVBCayj4AbB3wCe1wx4BPjAAazgYAGU+AFs4GjMoOGRUL0s4QOCwiMjjEMSAc2QYAQBXDJA4GJ44ABNaupo+KALSvgK4ADEIARi+GHMW5BA+UpgIAxAACxgYzwB1KbN4SzywOEi3ZbMAN2XfYbBLAJBE9IEYJ2CCvpS0ioArSwAPSMLPOABFUoh4O890oY8gJzsN+DVPJNjAJEjB5mZajApsgABwABiB9nS82CxmG0Lg5x2yUMAEcfvAriZ3CNLBwknU6pMBHByWYWVd2rckKkARV0jEzMUyvz3l8KckefdDDA+Dx4YjkUhlDLgmZPIUAMIQGLckaeSGGUrpChy9y8wEgHblACSUAasEiYBhpgAgvbIjBfB9-ukAL5+oA)✨**


changelog_unreleased/javascript/13532.md
```diff
@@ -0,0 +1,26 @@
...
+
+// Prettier stable (second output)
+foo(/* HTML */ `<!-- bar1 -->
+  bar
+  <!-- bar2 -->`);
+
+// Prettier main (first output)
+foo(/* HTML */ `<!-- bar1 -->
+  bar
+  <!-- bar2 -->`);
```
thorn0, line 25: For this specific case hugging actually doesn't look good to me.
Probably, it'd make sense to distinguish between these two cases and do hugging only for the 2nd:

```js
// 1
foo(
  /* HTML */ `<!-- bar1 -->
    bar
    <!-- bar2 -->`
);

// 2
foo(/* HTML */ `
  <!-- bar1 -->
  bar
  <!-- bar2 -->
`);
``` 

Thoughts on whether this makes sense and on how to implement this?

thorn0:
Using substrings in label checks instead of what clearly should be a list looks clumsy. It'd be more clear to extend the type of label  docs (see https://github.com/prettier/prettier/issues/11205), but I decided not to do that in this PR.

src/language-js/embed/html.js
```diff
@@ -73,17 +79,19 @@ async function embedHtmlLike(parser, textToDoc, print, path, options) {
...
+      : [
+          "`",
+          leadingWhitespace,
+          topLevelCount > 1 ? indent(group(contentDoc)) : group(contentDoc),
+          trailingWhitespace,
+          "`",
+        ]
+  );
+
+  return linebreak ? resultDoc : label("[no-hug]", resultDoc);
```
fisker, line 94: `label` seems not limited to `string`, how about

```js
return label({hug: false}, ...)
```

thorn0: Yes, I'll try that.

fisker (approved):

src/document/builders.js
```diff
@@ -251,7 +251,7 @@ function addAlignmentToDoc(doc, size, tabWidth) {
 }
 
 function label(label, contents) {
-  return { type: DOC_TYPE_LABEL, label, contents };
+  return label ? { type: DOC_TYPE_LABEL, label, contents } : contents;
```
fisker, line 254: Why the label has to be `truthy`?

thorn0: For building code to be more compact. E.g.
```js
label(
  typeof doc.label === "object" && { commented: true, ...doc.label },
  [leading, doc, trailing]
)
```
No need to create an intermediate variable for `[leading, doc, trailing]`.

fisker: If the `doc.label` is `null`, you can still separate in object, they result the same, `0`, `false`, `undefined`, '""' won't pass `typeof` check

thorn0: `doc.label` can't be `null` because the `label` builder doesn't accept falsy values. So this code adds a `type: "label"` doc wrapper only if `doc` is a label itself and its `.label` is an object. This way things like `.label.embed` and `.label.hug` are propagated. Otherwise it returns `[leading, doc, trailing]` without wrapping.

fisker: I'm confusing myself.... you want to use the code above? Then the deleted comment seems more correct.

thorn0: I do use the code above, see the diff.

> you can still spread, they result the same

But I don't want to add the label doc wrapper at all in that case.

fisker: > But I don't want to add the label doc wrapper at all in that case.

Make sense.

But when people call `label()` should expected a "label" doc to return.

thorn0: Compare these two alternatives. The first one looks better to me.
```js
// 1
function printTaggedTemplateLiteral(print) {
  const quasiDoc = print("quasi");
  return label(quasiDoc.label && { tagged: true, ...quasiDoc.label }, [
    print("tag"),
    print("typeParameters"),
    lineSuffixBoundary,
    quasiDoc,
  ]);
}

// 2
function printTaggedTemplateLiteral(print) {
  const quasiDoc = print("quasi");
  const resultDoc = [
    print("tag"),
    print("typeParameters"),
    lineSuffixBoundary,
    quasiDoc,
  ];
  return quasiDoc.label
    ? label({ tagged: true, ...quasiDoc.label }, resultDoc)
    : resultDoc;
}
```

thorn0: > But when people call label() should expected a "label" doc to return

There already are things like `join` among the doc builders that break this expectation. I think it's enough to document this behavior (which I did).

fisker: The second one has more explicit intention to me.

thorn0: That's right, but the `concat` builder had more explicit intention than arrays too. I expect we're going to have more code like this (relying on propagated labels). It's good to be explicit in one-off scenarios, but not in something repeated.

thorn0: As a trade-off for explicitness, a `conditionalLabel` builder could be added, but I'm not convinced it's really needed.

fisker: `concat` is a removed thing.

fisker: Won't block this, just thought it's unnecessary limitation.

fisker (approved):

src/document/builders.js
thorn0: > concat is a removed thing

Right. But it was explicit. It's just another instance of the typical "DRY vs explicit" tradeoff.

> Won't block this, just thought it's unnecessary limitation.

I saw a reusable pattern. I used it at least 3 times only in this PR and expect it to be used more. So I encapsulated it into the function.

thorn0:
Ouch! Sometimes it's not wrong to have more than one changelog file in 1 PR.
```
Run yarn lint:changelog
Duplicate files for #13532 found.
  - changelog_unreleased/api/13532.md
  - changelog_unreleased/javascript/13532.md
Error: Process completed with exit code 1.
```

fisker:
>  not wrong to have more than one changelog file in 1 PR.

My idea is to prevent accidentally "copy file" instead of "move file". You can remove that check.

fisker:
Filename of those files shouldn't used when generating blog post. You can also update https://github.com/prettier/prettier/blob/822f354cebf2956af77582fb32d8679b4f0e1887/scripts/lint-changelog.mjs#L65 to allow `13532-2.md`

PR #13577, opened 2022-10-02
Format pseudo-selector args like a function call

j-f1:
## Description

Currently, pseudo-selector arguments are formatted all on one line (using `", "` as a separator. This results in [undesirable formatting](https://twitter.com/scriptraccoon/status/1576320682245267456):

```css
:where(#p0:checked ~ #play:checked ~ #c1:checked, #p1:checked
    ~ #play:checked
    ~ #c2:checked, #p2:checked ~ #play:checked ~ #cO:checked)
  ~ #result
  > #c {
  display: block;
}
```

My change formats them more like function arguments in other languages (using the `group`/`indent` primitives).

Additionally, tests for `:where`/`:is` were missing from the codebase, so this issue was not apparent in the tests. I’ve added the above code, as well as some examples from MDN to make sure we don’t regress in the future.

The current logic was implemented in #1792 back in 2017, which I believe is before these newer selectors were proposed, and since they didn’t introduce any syntax changes on our end, it seems like nobody ever thought to update the implementation

## Checklist

- (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).


fisker (approved):
LGTM

PR #13583, opened 2022-10-06
Improve plugin resolution (fixes pnpm compatibility)

adamhl8:
Continuation of #11248


tests/integration/__tests__/plugin-resolution.js
```diff
@@ -93,6 +93,55 @@ describe("automatically loads '@prettier/plugin-*' from --plugin-search-dir (dif
...
+  runPrettier("plugins/pnpmVirtualStore/prettier@0.0.0", [
+    "file.txt",
+    "--parser=foobar",
+  ]).test({
+    stdout: "foobar+contents" + EOL,
+    stderr: "",
+    status: 0,
+    write: [],
+  });
+});
```
fisker, line 143: I don't think these tests are correctly testing the added part. The plugin search logic always starts from Prettier cli, if we want test, we'll have to put Prettier into those dirs.

src/common/load-plugins.js
```diff
@@ -113,6 +142,20 @@ async function findPluginsInNodeModules(nodeModulesDir) {
   return pluginPackageJsonPaths.map(path.dirname);
 }
 
+async function findPluginDirsInPnpmVirtualStore(searchDirectory) {
+  const pluginPackageJsonPaths = await fastGlob(
+    ["prettier-plugin-*", "@*/node_modules"],
```
fisker, line 147: Shouldn't plugins also stored like `prettier-plugin-foo@1.0.0`? I don't know how the `@*/plugin-*` stored, but are you sure these patterns correct?

tests/integration/__tests__/plugin-resolution.js
adamhl8: I could be misunderstanding the code, but I think in `run-prettier.js` the cwd gets set properly in `run()`. Either way, if I log `autoLoadDir`, it is set to the expected directory, so it does seem prettier is starting from the right directory.

src/common/load-plugins.js
adamhl8: In the pnpm virtual store, the three possible types of plugins (`@prettier/plugin-*`, and `@<name>/prettier-plugin-*`, `prettier-plugin-*`) are stored like this:

```
.pnpm/
├── @prettier+plugin-foo@<version>/
│   └── node_modules/
│       └── @prettier/
│           └── plugin-foo/
├── @<name>+prettier-plugin-bar@<version>/
│   └── node_modules/
│       └── @<name>/
│           └── prettier-plugin-bar/
└── prettier-plugin-baz@<version>/
    └── node_modules/
        └── prettier-plugin-baz/
```

I don't think the `+package-name` part of the directory used to be there, but now that it is, I can just have the patterns be `["prettier-plugin-*", "@prettier+plugin-*", "@*+prettier-plugin-*"]`.

adamhl8:
@fisker Any chance this will be included in next?

Issue #13485, opened 2022-09-16
Muliple NBSPs in code secontion in markdown string are collapsed


nschonni:
**Prettier 2.7.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAxCAnABHAHgQwFsAHAGzgBosYALOLAZxgwEsoBzLAAwB0R8ABUIBGfLlgAUtDHHoNi+MHAZZhcGAHdZULAH18urPigATPcN0BKRnAUZ88FfsPGzui1kKZ6tY9z74qmKS0HIKSqrqWgh6Bkam5lYAdCAUIBDEMCzQDMig+BgYEBoACgUIuSj4pBr4AJ65acL2YADW6gDKRHAAMmxwyABm1QyUIM2K7TAd4WzsyMwArmNwhGomJnAmPcbsi-jscOgYhA5ZHMj8izAQqSA0MISkAOo0LI7hcB0V7ywAbu86pcwAxGiA2KMMDASvZ2KchiMxgArBi4DpzcgARUWEHgCNIozSdkhl1OGFaJmKUDuxFYsGeLBMtGQAA4AAxEoqjZ72YiXWnKOAYP4DNIARxx8BhGUq-AYAFooLJNiY7jIJSwZDCDvCkMMCWNRoQWAsMMs0gwMXBsbiBnrEWkYPhhAymTRkAAmR32FikOYAYQghF1IGUAFY7otRgAVZ2VfWEkB-ZYASVMCGmYFYmQAgqYOjA6uR8aMAL6loA)

```sh
--parser markdown
```

**Input:**

```markdown
For example, the string `"a   b"` (three spaces between _a_ and _b_) separates _a_ and _b_ more than `"a b"` (one space between _a_ and _b_).
```

**Output:**

```markdown
For example, the string `"a b"` (three spaces between _a_ and _b_) separates _a_ and _b_ more than `"a b"` (one space between _a_ and _b_).

```

**Expected behavior:**
I replaced regular spaces with Non-breaking spaces to prevent the collapsing of the spaces that occurs with regular spaces, but if you look at `"a   b"` it shouldn't be collapsed as it renders correctly.

fisker:
https://github.com/prettier/prettier/blob/cd3e530c2e51fb8296c0fb7738a9afdd3a3a4410/src/language-markdown/print-preprocess.js#L31-L38

kachkaev:
[...]

PR #13590, opened 2022-10-08
Markdown: Preserve multiple spaces in inline code

kachkaev:
## Description

Fixes https://github.com/prettier/prettier/issues/13485

## Checklist

**✨[Try the playground for this PR](https://deploy-preview-13590--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEADABAMwhdAjAQwCd8CAvdVEAGhAgAcYBLaAZ2VGKIgHcAFYgnYoCAGx4EAnu1p4iBMAGs4MAMoEAtnAAyTKHGSYxrOLPlKVq+gr0BzZDCIBXUyDga8cACZfv2glC2TgS2cABiEEQaBDDMgcggBE4wEDQgABYwGqIA6ulM8KzWYHCqQgVMAG4FkglgrDIgeiZEMHzyttGGxq4AVqwAHqp2onAAik4Q8N2iJrTWRC0J0USKXrxQafREejA5TF4w6cgAHAAM89wmOfL0CdtwLZUGtACOk-DtDMKJrAC0+m83jSRDg7yYoPaIS6SCMs1cJg0TAczgRI3GHwMsJ6tBgBDw+0OxyQACZcfImKI7ABhCAaGGJUSiNJOEwAFXxwjhcxAlRcAEkoL5YKowDtGABBIWqGCSUYzEwAX0VQA)✨**


src/language-markdown/print-preprocess.js
```diff
@@ -34,7 +34,13 @@ function transformInlineCode(ast) {
       return node;
     }
 
-    return { ...node, value: node.value.replace(/\s+/g, " ") };
+    return {
+      ...node,
+      value: node.value
+        .replace(/^\s+/, " ")
+        .replace(/\n/g, " ")
+        .replace(/\s+$/, " "),
```
fisker, line 42: Does `/(?:^\s+|\n|\s+$)/g` work? Is it faster?

kachkaev: I‘ve updated the logic a bit. Looks like we need to treat ` ` and other whitespace differently, [according to the spec](https://spec.commonmark.org/0.30/#backtick-string).

tests/format/markdown/spec/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,162 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`commonmark-0.30-example-328.md - {"proseWrap":"always"} format 1`] = `
```
kachkaev, line 3: Context: https://github.com/prettier/prettier/issues/13591

kachkaev:
Hmmm [Example 331](https://spec.commonmark.org/0.30/#example-331) fails second pass:

```
FULL_TEST=true yarn test markdown/spec/jsfmt
```

````txt
[space][space]``[space][space]
↓
[space]``[space]
↓
``
````

Screenshot 2022-10-08 at 21 37 09


Going to take a break now. Any thoughts?

**Comment by fisker** [comment]
example 331 & example 330 has different value, https://github.com/prettier/prettier/blob/62d95b33b515f83187aba245b3d97b937f37914d/src/language-markdown/clean.js#L36

**Comment by kachkaev** [comment]
Example 331 becomes Example 330 after first pass, does not it? Maybe I’m missing something.

File: src/language-markdown/print-preprocess.js [hunk 3]
```diff
@@ -31,7 +31,17 @@ function transformInlineCode(ast) {
       return node;
     }
 
-    return { ...node, value: node.value.replace(/\s+/g, " ") };
+    const valueWithoutNewlines = node.value.replace(/\n/g, " ");
```
thorn0, line 34: Let's preserve new lines when `proseWrap` is set to `preserve'. This PR will supersede #11373 then.

thorn0: #11373 has been merged. Please rebase.

kachkaev:
Pressed the wrong button 😅

I included https://github.com/prettier/prettier/pull/11373. I still don’t fully understand the relationship between `src/language-markdown/print-preprocess.js` → `transformInlineCode` and `src/language-markdown/clean.js` 🤔 The latter does not have access to `options.proseWrap`, but tests still pass. If I remove any transforms from either function, some snapshots fail 🤔

kachkaev:
@thorn0 I merged `next` into my branch, so PR playground includes https://github.com/prettier/prettier/pull/13606 now. Thanks for implementing ‘show preprocessed AST‘!

Do you foresee any other changes we need to make before merging?

src/language-markdown/print-preprocess.js
```diff
@@ -27,11 +27,22 @@ function transformImportExport(ast) {
...
-    return { ...node, value: node.value.replace(/\s+/g, " ") };
+    const value =
+      options.proseWrap === "preserve"
+        ? node.value
+        : node.value.replace(/\n/g, " ");
+
+    return {
+      ...node,
+      value:
+        value.startsWith(" ") && value.endsWith(" ") && value.trim().length > 0
```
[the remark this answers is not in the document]
thorn0: Okay, I see why you did this. But this should be done in `print`, not in `preprocess`.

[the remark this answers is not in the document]
kachkaev: I see, thanks for the fix! 🙌

Do you know why we still need `newObj.value = ast.value.replace(/\n/g, " ");` in `clean.js`? This transform is applied regardless of `proseWrap`. I vaguely remember trying to remove it but tests failed.

changelog_unreleased/markdown/13590.md
```diff
@@ -0,0 +1,15 @@
+#### Preserve multiple spaces in inline code (#13590 by @kachkaev)
```
kachkaev, line 1:
```suggestion
#### Preserve multiple spaces in inline code (#13590 by @kachkaev and @thorn0)
```

@thorn0 feel free to commit this if you want

src/language-markdown/print-preprocess.js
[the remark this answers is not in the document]
thorn0: Those tests are run only with `proseWrap: "always"`, that's why it works. Need to think what to do.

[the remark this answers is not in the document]
thorn0: No, I was wrong. It's some other tests that fail if I add
```js
run_spec(import.meta, ["markdown"], { proseWrap: "preserve" });
run_spec(import.meta, ["markdown"], { proseWrap: "never" });
```
to `jsfmt.spec.js`, not related to code spans. 

As for `clean.js`, see how the `clean` function (AKA `massageAST`) is used:

https://github.com/prettier/prettier/blob/955553bb3972eae532206ffc3a7d4a2383cb7c62/src/cli/format.js#L137-L169

It's applied to both the original AST and to the AST built from Prettier's output, and then the ASTs are checked for deep equality. So it makes sense that this replacement is needed. After it, not matter what `proseWrap` is, the values in the two ASTs are equal.

thorn0 (approved):

src/language-markdown/print-preprocess.js
```diff
@@ -31,7 +31,7 @@ function transformInlineCode(ast, options) {
       return node;
     }
 
-    return { ...node, value: node.value.replace(/\s+/g, " ") };
+    return { ...node, value: node.value.replace(/\n/g, " ") };
```
fisker, line 34: Maybe we can move this to `print`? So we can reduce one `mapAst` call. Can be separate PR.

tests/format/js/multiparser-markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -109,7 +109,7 @@ markdown\`
 
 =====================================output=====================================
 markdown\`
-  const cssString = css\\\` background-color: \\$\\{color('base')\\}\\\`;
+  const cssString = css\\\`  background-color: \\$\\{color('base')\\}\\\`;
```
fisker, line 112: Shouldn't this call `markdown -> js -> css` and remove the leading space? Because it's not valid css?

fisker (approved):
Good job! Just a few unimportant questions.

tests/format/js/multiparser-markdown/__snapshots__/jsfmt.spec.js.snap
thorn0: This is a crazy test. `const cssString = ...` is Markdown here, not JS. The Markdown-in-JS embedder also includes logic for ignoring indentation (no idea what's the story behind that). That's why the output is like that.

fisker: Ah, you explained before... forgot. Sorry.

kachkaev:
Great, thanks for helping with this PR @thorn0 and @fisker! 👏

PR #13606, opened 2022-10-09
Playground: show preprocessed AST

thorn0:
## Description

This can be especially useful for the Markdown printer as its preprocessing is really complex. HTML is interesting too.

## Checklist

### DIFF
--- a/src/main/core.js
+++ b/src/main/core.js
@@ -336,9 +336,9 @@ const prettier = {
       }
       if (devOptions.preprocessForPrint) {
         attachComments(parsed.text, parsed.ast, options);
-        parsed.ast = options.printer.preprocess
-          ? await options.printer.preprocess(parsed.ast, options)
-          : parsed.ast;
+        if (options.printer.preprocess) {
+          parsed.ast = await options.printer.preprocess(parsed.ast, options);
+        }
       }
     }
     return parsed;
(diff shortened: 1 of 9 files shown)


src/main/core.js
```diff
@@ -324,14 +324,22 @@ async function formatWithCursor(originalText, originalOptions) {
...
-    if (massage) {
-      parsed.ast = massageAST(parsed.ast, options);
+    if (devOptions) {
+      if (devOptions.massage) {
+        parsed.ast = massageAST(parsed.ast, options);
+      }
+      if (devOptions.preprocessForPrint) {
+        attachComments(parsed.text, parsed.ast, options);
+        parsed.ast = options.printer.preprocess
+          ? options.printer.preprocess(parsed.ast, options)
```
fisker, line 340:
```suggestion
          ? await options.printer.preprocess(parsed.ast, options)
```

tests/config/require-standalone.cjs
```diff
@@ -56,7 +56,9 @@ module.exports = {
               ...($$$options.plugins || []),
             ],
           };
-          prettier.__debug.parse($$$input, options, ${JSON.stringify(massage)});
+          prettier.__debug.parse($$$input, options, ${JSON.stringify(
+            devOptions
+          )});
         `,
         { $$$input: input, $$$options: options, ...sandbox }
```
fisker, line 63: Let's pass `devOptions` as `$$$devOptions` like other arguments.

fisker (approved):

PR #13632, opened 2022-10-13
Refactor yaml print

fisker:
## Checklist


src/language-yaml/print/mapping-item.js
```diff
@@ -62,9 +63,7 @@ function printMappingItem(node, parentNode, path, print, options) {
       hardline,
       join(
         "",
```
fisker, line 65: Don't know why call `join()` with empty string here, remove it the test still passing. Should I remove?

sosukesuzuki (approved):

Issue #13540, opened 2022-09-27
Update tests


fisker:
We have [`index.cjs`](https://github.com/prettier/prettier/blob/next/src/index.cjs) and `[`index.js`](https://github.com/prettier/prettier/blob/next/src/index.js) now, currently we only tested ESM version, we should run tests on both.

The bundled ESM standalone version also not well tested https://github.com/prettier/prettier/blob/3f21a973e1b508ec7506f9ea1f23ce0be9585878/tests/integration/__tests__/bundle.js#L35, better run full tests too.

Ideally, we should also run standalone version on both Node.js and a REAL browser.

PR #13639, opened 2022-10-14
Fix package.json `exports` field

fisker:
## Description

It seems

```
{
	"./*": "./"
}
```

is incorrect.

https://github.com/prettier/yaml-unist-parser/pull/289

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
       "require": "./src/index.cjs",
       "default": "./src/index.js"
     },
-    "./*": "./"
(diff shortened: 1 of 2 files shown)


kachkaev (approved):
The change makes sense, especially given prior testing in https://github.com/prettier/yaml-unist-parser/pull/289

fisker:
I did install locally, and tested. We need test it in future https://github.com/prettier/prettier/issues/13540

PR #13642, opened 2022-10-14
Sync commonmark spec examples

### LINKED ISSUE #13591: Markdown spec examples need updating in tests
The contents of the [tests/format/markdown/spec](https://github.com/prettier/prettier/tree/008b371e3ed64d31b85f10b6b3b67d3c32f175b6/tests/format/markdown/spec) folder date back to https://github.com/prettier/prettier/pull/2943 (June 2017). [CommonMark spec](https://spec.commonmark.org/0.30/) has been updated since then and enumerates 652 examples instead of 616.

We should probably sync the list of examples to make sure that nothing is missing. Getting the sequencing right should improve DX too. In the meantime, I have inserted a few files named `commonmark-0.30-example-xxx.md` to close the gap. See https://github.com/prettier/prettier/pull/13590.

@ikatyang do you remember using any automation to populate the examples?

fisker:
## Description

Fixes #13591

## Checklist


kachkaev:
WDYT of switching to `commonmark-0.30-example-xxx.md` for file names? If we stick with `example-xxx.md`, git will detect files as modified instead of renamed each time examples are shifted in a new spec. Diff will be cleaner with file prefix.

fisker:
Let's do this after #13676

fisker:
Data already exists in this package https://github.com/wooorm/commonmark.json

fisker:
[create-gfm-fixtures](https://www.npmjs.com/package/create-gfm-fixtures) can be useful too.

PR #13680, opened 2022-10-18
Update `angular-html-parser` to v2.1.0

fisker:
## Description

Pass `angularHtmlParser` around, instead of making everything async.

## Checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -44,7 +44,7 @@
     "src/index.cjs",
     "standalone.js",
     "bin/",
-    "dist/",
+    "dist*/",
     "docs/",
     "scripts/",
     "tests/",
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -44,7 +44,7 @@
     "src/index.cjs",
     "standalone.js",
     "bin/",
-    "dist/",
+    "dist*/",
```
thorn0, line 47: Is this intended?

thorn0:
Why not just import it statically like e.g. remark is loaded in `parser-markdown.js`?

tsconfig.json
fisker: Yes, when debugging build script, I save different `dist` directories, so I ignored `dist*` https://github.com/prettier/prettier/blob/92a6af22e72cf1d937492fecc591bb41edd16306/.gitignore#L14

thorn0 (approved):
Looks good. But it'd be good to investigate what's going on with those dynamic imports.

sosukesuzuki (approved):

PR #13682, opened 2022-10-19
Run tests on Node.js 19

fisker:
## Checklist


kachkaev (commented):
Are changes in `tests/integration/__tests__/__snapshots__/third-party.js.snap` and `tests/integration/__tests__/third-party.js` intended?

fisker:
Yes, the error message is different, so we can't snapshot, https://github.com/prettier/prettier/actions/runs/3278542963/jobs/5397053358#step:5:53. I assert with  `toConatain` now.

sosukesuzuki (approved):

PR #13684, opened 2022-10-19
Build: fix `outdent` entry

fisker:
## Checklist

### DIFF
--- a/scripts/build/build-javascript-module.js
+++ b/scripts/build/build-javascript-module.js
@@ -232,6 +232,8 @@ function getEsbuildOptions({ file, files, shouldCollectLicenses, cliOptions }) {
     logLevel: "error",
     format: file.output.format,
     outfile: path.join(DIST_DIR, cliOptions.saveAs ?? file.output.file),
+    // https://esbuild.github.io/api/#main-fields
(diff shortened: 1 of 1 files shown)


thorn0:
Bundling seemed to work fine. Why is this fix needed?

fisker:
Our source is ESM, I expect esm version to be used, and the universal bundle works as expected. 

Because

> The default main fields depend on the current platform setting and are essentially browser,module,main for the browser and main,module for node. 

https://esbuild.github.io/api/#main-fields

In future, we may have other dual mode modules, I hope they can resolve to ESM version by adding `moduleFields`.

Also, there was a bug in esbuild, when  resolving to esm version [this workaround](https://github.com/prettier/prettier/blob/ebb8e1328261798a96228f9f18cac767cef983eb/scripts/build/build-javascript-module.js#L63) should work. Anyway, maybe esbuild already fixed it somehow.

thorn0 (approved):

fisker:
You may missed file change, [two Node.js bundle changed](https://github.com/prettier/prettier/actions/runs/3280617739/jobs/5401616297#step:4:217)

PR #13686, opened 2022-10-19
Rewrite sentence

HonkingGoose:
## Description

- Rewrite sentence on homepage:
  - Use the verb `click` instead of `press` (actually we should use a input-neutral verb!)
  - Change sentence order

## Checklist

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -96,7 +96,7 @@ const TldrSection = ({ language }) => (
         <div style={{ display: "flex", flexDirection: "column" }}>
           <h2>Why?</h2>
           <ul style={{ flex: "1" }}>
-            <li>Click save and your code is formatted</li>
+            <li>Your code is formatted on save</li>
             <li>No need to discuss style in code review</li>
             <li>Saves you time and energy</li>
             <li>
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -96,7 +96,7 @@ const TldrSection = ({ language }) => (
         <div style={{ display: "flex", flexDirection: "column" }}>
           <h2>Why?</h2>
           <ul style={{ flex: "1" }}>
-            <li>You press save and code is formatted</li>
+            <li>Click save and your code is formatted</li>
             <li>No need to discuss style in code review</li>
             <li>Saves you time and energy</li>
             <li>


sosukesuzuki (commented):
I don't know how natural these English sentences are to me...

website/pages/en/index.js
```diff
@@ -96,7 +96,7 @@ const TldrSection = ({ language }) => (
         <div style={{ display: "flex", flexDirection: "column" }}>
           <h2>Why?</h2>
           <ul style={{ flex: "1" }}>
-            <li>You press save and code is formatted</li>
+            <li>Click save and your code is formatted</li>
```
HonkingGoose, line 99: The [Microsoft Style guide](https://learn.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui) has some good information about describing interactions with UI:

> Customers interact with products using different input methods: keyboard, mouse, touch, voice, and more. So use generic verbs that work with any input method. Don't use input-specific verbs, such as _click_ or _swipe_. Instead, use the verbs listed here.

The document listed above also has a table of good verbs, but none of these fit the Prettier homepage... 😞 

Based on their recommendation, my suggestion to use `click` doesn't fix the problem. 😄 So how about using `save` as the verb? 

```suggestion
            <li>Formats code when saving files</li>
```

kachkaev:
👋 @HonkingGoose! Which practical problem are you trying to address?

HonkingGoose:
## Intro

Hi @kachkaev! 👋

Good question! Let's look at the verbs associated with some input methods:

## Input methods and verbs

| Input method | Verb(s) |
| :--- | :--- |
| mouse/trackball | `click` |
| keyboard | `type` |
| voice commands | `speak` |
| touchpad/table/smartphone | `swipe` / `tap` |
| eyetracking | `look` |
| controller | `move joystick`, `press trigger`, `press button` |

See how each input method has it's own verb?

## Benefits of using input-neutral verbs

Using an input-neutral verb like `select` has these benefits:

- The verb works for each input device, the user already knows how to use their input device to select things
- We use accessible language, as we're assuming that people use a keyboard or mouse
- It's easier to use one single verb, instead of figuring out if the user is `typing`/`clicking`/`swiping`/`tapping` something

## Summary

I'm not solving an practical problem with this PR. The goal here is to use accessible language, and simplify writing about UI interactions. 😉

apmeehan:
@HonkingGoose While I'm a fan of input-device-agnostic language, "_Formats code when saving files_" feels to me like it's missing some of the immediacy and... "snappiness" of the original :)

What about something like "_Save your file and your code is formatted_", or just "_Your code is formatted on save_"?

website/pages/en/index.js
HonkingGoose, line 99:
```suggestion
            <li>Your code is formatted on save</li>
```

HonkingGoose (commented):
I like the suggestion from @apmeehan best.

fisker:
"Your code is formatted on save" souds good.

fisker:
"Your code is formatted on save" souds good.

fisker (approved):

fisker:
Yes, need more review.

PraveenNanda124 (commented):
Good work

HonkingGoose:
Thank you all for reviewing my PR, and for merging it! 😄 ❤️

PR #13690, opened 2022-10-20
handlebars: Stricter `prettier-ignore` check in

fisker:
## Description

It seems we only mean to ignore elements, the previous check is too lossy. I think only elements next to or under the comment should be ignored.

## Checklist

### DIFF
--- a/src/language-handlebars/utils.js
+++ b/src/language-handlebars/utils.js
@@ -26,23 +26,40 @@ function isWhitespaceNode(node) {
   return node.type === "TextNode" && !/\S/.test(node.chars);
 }
 
-function isPrettierIgnoreNode(node) {
+function isPrettierIgnoreComment(node) {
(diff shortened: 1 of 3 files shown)


fisker:
@dcyriller We only have tests about elements. Is there any other case to be ignored?

dcyriller (approved):
> Is there any other case to be ignored?

This is true - only elements will be ignored. The following text and mustache types could be probably be ignored as well:
```hbs
{{! prettier-ignore }}
              ugly
```
```hbs
{{! prettier-ignore }}
{{        ugly}}
```

EDIT: opened https://github.com/prettier/prettier/pull/13693 to precise / highlight which cases are currently supported

fisker:
If so, we can allow them.

sosukesuzuki (approved):

Issue #13624, opened 2022-10-12
Ember handeblars formatting doesn't allow for self-closing tags


ArnaudWeyts:
**Prettier 2.7.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AID0AfADpSpgCuAzjBALYC0kdADtAjAYSADQgQsYmaFWSgAhgCdJEAO4AFKQlEpxAG1niAnqL4AjSeLABrODADK4unAAymKHGQAzdVTj7DJs+ZZH7Ac2QYSQp3EDg6PTh0dGibcSh-CnF-OAAxCEk6cRghROQQcQpaXhAACxg6NQB1Msx4Kl8wOHNlepx6rQKwKl0QezdJGHlDf2znVzCAKyoAD3MAtTgARQoIeAm1Nz5fSUGC-zVMOmtJUpZJexhqzHQYMuQADgAGHZk3asMWAou4QexHHwAI5reAjAQqQpUBgOaLRUqSOAgzCIkYpcZIFxbMJuOiYIIhHGLFagxyYyZ8GDiPQ3O4PJAAJkphkwR0SAGF6BjwlQAKylahwAAq1JUWO2IGwoQAklBYrBzGBLoIAIJy8wwLRLTZuAC+uqAA)

**Input:**

```hbs
<div />
<custom-component />
```

**Output:**

```hbs
<div></div>
<custom-component></custom-component>
```

**Expected behavior:**

```hbs
<div />
<custom-component />
```

[...]

PR #13691, opened 2022-10-20
ember / glimmer: Preserve self-closing tags

dcyriller:
## Description

As reported in https://github.com/prettier/prettier/issues/13624, self-closing tags are supported by Ember ([doc](https://guides.emberjs.com/v4.7.0/components/#toc_self-closing-tags)).

This PR would preserve them. The PR would close https://github.com/prettier/prettier/issues/13624.

## Checklist

- [] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).


src/language-handlebars/utils.js
```diff
@@ -41,6 +41,7 @@ const voidTags = new Set(htmlVoidElements);
 function isVoid(node) {
   return (
     voidTags.has(node.tag) ||
+    node.selfClosing === true ||
```
fisker, line 44: Should we check this in print? So we can keep `<i></i>`?

dcyriller: Do you think `<i/>` should be formatted to `<i></i>`? While `<div />` should remain `<div />`? (inline elements vs others)

(Just pushed more tests and verbose changelog to better describe the changes.)

fisker: No, I mean maybe we should keep it as it is. It's not our job to enforce the self close style.

fisker: `<i></i>` -> `<i></i>` 
`<i/>`  -> `<i/>`

dcyriller: oh I see, thanks for the clarification
I have added more test cases in the PR to show it doesn't enforce the self close style (except for the `<Component />` case, but that was already the case prior to this PR, for a reason I don't remember)

fisker: Ah, sorry, I misunderstood.

thorn0 (approved):

src/language-handlebars/utils.js
```diff
@@ -41,6 +41,7 @@ const voidTags = new Set(htmlVoidElements);
 function isVoid(node) {
   return (
     voidTags.has(node.tag) ||
```
fisker, line 43: Should we remove the html void tag check? 

`` should always self closed.  If anyone wrote `</img>`, we can just ignore.

fisker:
I'm going to merge this, and address https://github.com/prettier/prettier/pull/13691#discussion_r1001731467 in another PR.

PR #13693, opened 2022-10-20
Handlebars: Add tests for `{{! prettier-ignore}}`

dcyriller:
## Description

This PR adds test cases for {{! prettier-ignore}} when it is above or before:
- a text element
- a mustache element
- a block element
- a comment element

@fisker it should help with https://github.com/prettier/prettier/pull/13690

## Checklist


fisker:
Thanks for the test. It seems a little tricky to handle all these cases.  Would you like to work on it?
If you do, please target the `next` branch, code changed in utils.

PR #13699, opened 2022-10-21
Remove `since` from options and languages

fisker:
## Description

Since v2.0.0, we removed `version` option from `getSupportInfo` #7620, `since` became useless, let's remove it.

## Checklist

### DIFF
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -59,7 +59,7 @@ function getSupportInfo({
   return { languages, options };
 
   function filterDeprecated(object) {
-    return showDeprecated || !("deprecated" in object) || object.deprecated;
+    return showDeprecated || !object.deprecated;
   }
 
   function mapInternal(object) {
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -35,7 +35,9 @@ function getSupportInfo({
 
       if (Array.isArray(option.default)) {
         // eslint-disable-next-line no-console
-        console.warn("the default value of option should be a primitive value or a function.")
+        console.warn(
+          "the default value of option should be a primitive value or a function."
+        );
         option.default = option.default[0].value;
       } else if (typeof option.default === "function") {
         option.default = option.default();
--- a/src/main/support.js
+++ b/src/main/support.js
@@ -33,6 +33,14 @@ function getSupportInfo({
     .map((option) => {
       option = { ...option };
 
+      if (Array.isArray(option.default)) {
+        // eslint-disable-next-line no-console
+        console.warn("the default value of option should be a primitive value or a function.")
+        option.default = option.default[0].value;
+      } else if (typeof option.default === "function") {
+        option.default = option.default();
+      }
+
       if (Array.isArray(option.choices)) {
         option.choices = option.choices.filter((option) =>
           filterDeprecated(option)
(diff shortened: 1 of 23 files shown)


fisker:
It can be useful, but if we do https://github.com/prettier/prettier/issues/9879#issuecomment-895183665, we won't need that. Even if we want run multiple version in on page, we can still hardcode there.

thorn0:
I forgot that every version can tell us itself what it supports. So let's proceed.

fisker:
Added old values to playground, so old link won't be broken https://github.com/prettier/prettier/pull/13699/commits/1ba50c73b5465b4fb3d52c67c9ba244e0c49a518

src/main/support.js
```diff
@@ -16,49 +12,30 @@ import currentVersion from "./version.evaluate.cjs";
...
-      if (Array.isArray(option.default)) {
-        option.default =
-          option.default.length === 1
-            ? option.default[0].value
-            : option.default
-                .filter(filterSince)
-                .sort((info1, info2) =>
-                  semverCompare(info2.since, info1.since)
-                )[0].value;
-      }
```
fisker, line 57: Maybe I should keep this array thing, otherwise it will be a breaking change, plugin may use array as default value.

fisker: Deprecated array, introduced function for non-primitive values, it will be safer (to avoid mutate).

thorn0: Shouldn't these things be JSON-serializable?

fisker:
> Also this logic was used to hide unreleased things on the playground.

I know that, but why? Only PR's playground will show unreleased, it should be shown, let people with it.

fisker:
And we already use `showUnreleased: true,`
https://github.com/prettier/prettier/blob/14c58aafd899f9b0b6ab7f7b9713b8f9eed36c09/website/static/worker.js#L82

src/main/support.js
fisker: Make sense, so I'll just put `[{value: []}]` back.

fisker: Wait, we get the support info from `getSupportInfo` function, so it will return `[]`, then it's serializable.

fisker: The option self is not serializable, there are functions for `exception`.

thorn0:
> And we already use `showUnreleased: true,`
> 
> https://github.com/prettier/prettier/blob/14c58aafd899f9b0b6ab7f7b9713b8f9eed36c09/website/static/worker.js#L82

Doesn't this PR remove support for `showUnreleased`?

fisker:
`showUnreleased` checks `since`, there is no `since` to check now.

fisker:
I'm we already show unreleased before this PR, so 

> Also this logic was used to hide unreleased things on the playground.

We are not using.

fisker:
Don't overthink this, nobody is using it, even I thought `prettier-vscode` may use it to support different version of prettier, but they are not.

src/main/support.js
fisker: Let's add a new `defaultValue` field?

thorn0: How will it be different from `default`?

fisker: When they are array, `defaultValue` use the value directly, `default` use `default[0].value`, so we don't need ask plugin to update.

fisker: So I can use `defaultValue: []` instead of `default: [{value: []}]`.

fisker: Forget it, switched back to  `default: [{value: []}]`

thorn0 (approved):

sosukesuzuki (approved):
Good catch

PR #13703, opened 2022-10-22
Tweak gap regex for Flow

thorn0:
## Description

fix for https://github.com/prettier/prettier/pull/13687#issuecomment-1287052686

## Checklist


tests/format/flow/comments/babel-only/type_annotations-3.js
```diff
@@ -1,9 +0,0 @@
-import type { OneType } from './oneFile.js'
-/*::
-import type { HelloType } from './otherFile.js'
-*/
-
-type PropTypes = {|
-  TestComponent: React.AbstractComponent<any>,
-  hello: HelloType,
-|};
```
thorn0, line 9: I forgot to move this test in #13687

sosukesuzuki (approved):

src/language-js/comments/printer-methods.js
```diff
@@ -82,6 +82,14 @@ function willPrintOwnComments({ node, parent }) {
   );
 }
 
+function getGapRegex(enclosingNode, { parser }) {
+  if (parser === "flow" || parser === "babel-flow") {
+    // Example: (a /* b */ /* : c */)
+    //                gap ^^^^
+    return /^[\s(]*(\/\*(\s*::)?[\s(]*)?$/;
+  }
+}
```
fisker, line 91: I remember there is one regexp for babel-ts? Where is it...

fisker: https://github.com/prettier/prettier/commit/c1937cb1b8717d8fe2247de99e7031bd7fd7bab3#diff-112b7d78917580aa0cc6cc3ac0e641c6a9541c0ed4de574f05dba2ae51480ccaR914

fisker: Ah, removed in #10457, fogot.

fisker: Will `/^(\s|\(|\/\*\s*::)*$/` work? This is more readable to me.

fisker: Can we use `isGap`? So we can use a function, the regexp is hard to read when it's long.

```js
function isGap(text) {
	text = text.replace(/[\s\(]/g, '')
	return !text || text === '/*::'
}
```

fisker (approved):

PR #13733, opened 2022-10-26
Update babel to v7.20.0

fisker:
## Description

Babel going to make some change to the location info https://github.com/babel/babel/pull/15032. I downloaded the artifact to see what will break.

## Checklist


fisker:
Note to myself, `babel-ts` range maybe incorrect.

src/language-js/loc.js
```diff
@@ -4,20 +4,16 @@ import isNonEmptyArray from "../utils/is-non-empty-array.js";
...
-  if (!ignoreDecorators) {
-    const decorators =
-      (node.declaration && node.declaration.decorators) || node.decorators;
-
-    if (isNonEmptyArray(decorators)) {
-      return locStart(decorators[0]);
-    }
+  const decorators = node.declaration?.decorators ?? node.decorators;
+  if (isNonEmptyArray(decorators)) {
+    return Math.min(locStart(decorators[0]), start);
```
fisker, line 13: It seems the old one is not right, if decorators after `export`, it will still use range of `decorators[0]`

fisker:
This PR only fixes `hasDecoratorsBeforeExport` check, other changes like

- Parse import reflection
- Add support for expr satisfies Type expressions

Will be done in separate prs.

https://github.com/babel/babel/releases/tag/v7.20.0

sosukesuzuki (approved):

src/language-js/loc.js
nicolo-ribaudo: In which case is `locStart(decorators[0])` less then `start`?

fisker: > In which case is locStart(decorators[0]) less then start?

Other parsers, `flow`/`typescript` I think they followed old behavior of babel.

PR #13746, opened 2022-10-28
Refactor options

fisker:
## Description

[TBD]

## Checklist


thorn0:
All the options are gone on the playground.

> I guess the initial idea is to only pass those options when doing print

Probably not. Maybe the idea was simply to categorize them.

> it should be defined in parser/printer, not plugins, because plugins can have multiple printers

How would configuration files and CLI work in that case? Would users need to specify which parser/printer each option is for?

> think it's possible to define a null plugin, only contains {singleQuote: {default: true}} will make us use singleQuote by default

Does that really need fixing? The biggest problem with the current state of things I can see is that multiple plugins may define options with the same name but different set of values. It's a typical problem with any global name space, but sometimes avoiding such problems creates quite a lot of complexity (e.g., modules in JS). What I'm saying is that sometimes documenting a known issue and, for example, a recommendation to use prefixes for plugin options is a good enough solution.

PR #13788, opened 2022-11-03
Enable `unicorn/prefer-string-replace-all` rule

fisker:
## Description

~I need polyfill this.~ done

## Checklist


sosukesuzuki (approved):
LGTM, however this transform layer is a bit complicated and I would like to remove it soon when Node.js 14 goes EoL...(Of course, I'm happy to use the new ES features with this transform.)

PR #13819, opened 2022-11-08
Support TypeScript 5

fisker:
## Description

https://github.com/microsoft/TypeScript/pull/51387

## Checklist


jakebailey:
Not sure if you saw it in my PR, but https://github.com/jakebailey/prettier-style-ts-tree-shaking is my PoC for getting a similar result for 5.0.

fisker:
I did. But I'm not going to use rollup.

jakebailey:
Oh, I see, you're using esbuild's comments like `// src/compiler/utilities.ts`. That's a really clever idea.

fisker:
Let's remove this https://github.com/prettier/prettier/blob/3fc6298c0ea633cc596b234b49869afa86689ba5/src/language-js/print/class.js#L307

fisker:
Changlog missing.

tests/format/typescript/decorator-auto-accessors/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,65 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`decorator-auto-accessor-like-property-name.ts format 1`] = `
+====================================options=====================================
+parsers: ["typescript"]
+printWidth: 80
```
fisker, line 6: Those tests need run in --no-semi mode, let's move to a separate dir and test.

fisker:
https://github.com/nodejs/node/issues/47096

PR #13828, opened 2022-11-12
Add clarification notes about EditorConfig

SiarheiBobryk:
```
## Description

Although I'm not a Prettier expert, IMHO the section about the EditorConfig integration is not super clear, to be honest. So I'd add some more words about the relationships between Prettier and EditorConfig.

## Checklist

<!-- Please ensure you’ve done all of these things (if applicable). -->
<!-- You can replace the `[ ]` with `[x]` to mark each task as done. -->

<!-- Please DO NOT remove the playground link -->

```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -176,7 +176,11 @@ If you’d like a JSON schema to validate your configuration, one is available h
 
 ## EditorConfig
 
-If `options.editorconfig` is `true` and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc.
+The `prettier` CLI respects `.editorconfig` by default, but you can opt out with `--no-editorconfig`.
+
+However, the API _doesn't_ respect `.editorconfig` by default, in order to avoid potential editor integration issues. To opt in, add `editorconfig: true` to the [`prettier.resolveConfig`](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options) options.
+
+> Note: Rules that are set in `.editorconfig` can be overridden by their corresponding rules from `.prettierrc.js`, therefore rules from `.editorconfig` can be used as a fallback for files that Prettier doesn't support. See the languages list supported by Prettier [here](https://prettier.io/docs/en/index.html)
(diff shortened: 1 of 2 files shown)


SiarheiBobryk:
Can anyone please review it?

fisker (approved):
LGTM! Thanks.

docs/configuration.md
```diff
@@ -176,7 +176,11 @@ If you’d like a JSON schema to validate your configuration, one is available h
 
 ## EditorConfig
 
-If `options.editorconfig` is `true` and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc.
+The `prettier` CLI respects `.editorconfig` by default, but you can opt out with `--no-editorconfig`.
+
+However, the API _doesn't_ respect `.editorconfig` by default, in order to avoid potential editor integration issues. To opt in, add `editorconfig: true` to the [`prettier.resolveConfig`](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options) options.
+
+> Note: Rules that are set in `.editorconfig` can be overridden by their corresponding rules from `.prettierrc.js`, therefore rules from `.editorconfig` can be used as a fallback for files that Prettier doesn't support. See the languages list supported by Prettier [here](https://prettier.io/docs/en/index.html)
```
thorn0, line 183: I don't understand this part: "rules from `.editorconfig` can be used as a fallback for files that Prettier doesn't support". What does this mean?

SiarheiBobryk: Let's say there is a set of rules in `.prettier.js` and an environment that doesn't support prettier (the not widely used programming language that is not in the [Prettier support list](https://prettier.io/docs/en/index.html) for instance). 

However, we still have `.editorconfig`, which has better support for [most IDEs](https://editorconfig.org/).

For example, if, for the reasons above, the `printWidth` Prettier rule cannot be used, we still have its EditorConfig's alternative `max_line_length` that will be used as a fallback value.

Does it answer your question, @thorn0?

thorn0: I see what you mean now, but this isn't clear from the proposed paragraph. A better wording is needed.

Issue #8763, opened 2020-07-15
Offline support for Playground


umeshgmrl:
Prettier playground should be an installable PWA so that it can be used offline

https://prettier.io/playground/

boreme12:
I would love you give this a go. Can I be assigned this?

boreme12:
@fisker For the PWA it is required to have an icon of atleast 512x512px. Do you know where I could look to find this?

fisker:
Thank you.

Icon  https://github.com/prettier/prettier-logo

boreme12:
@fisker I have got the core functionality working for the PWA.

How would you like to make it installable? Do we want a banner which pops up to prompt the user?

I have also noticed that the existing service worker you have implimented is currently depecriated (sw-toolbox) in favor of Workbox. Would you like me to tackle this as well? 
[https://github.com/GoogleChromeLabs/sw-toolbox/](https://github.com/GoogleChromeLabs/sw-toolbox/)

fisker:
I think a <kbd>install</kbd> button is good.

fisker:
> I have also noticed that the existing service worker you have implimented is currently depecriated (sw-toolbox) in favor of Workbox. Would you like me to tackle this as well?

Sure.

umeshgmrl:
any update?

umeshgmrl:
Sorry I've mistakenly closed this issue,

BTW any update?

boreme12:
Hi, sorry I have been struggling with this. I have managed to make it all work offline but I am struggling to convert it all over to sw-toolbox. I can create a PR and see what you guys think about my current implementation?

PR #13851, opened 2022-11-16
Add CodeQL Analysis workflow

fisker:
## Checklist

### DIFF
--- a/website/pages/playground/index.html
+++ b/website/pages/playground/index.html
@@ -28,39 +28,146 @@
 
     <link
       rel="stylesheet"
-      crossorigin
-      href="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.65.6/codemirror.min.css"
+      href="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.65.9/codemirror.min.css"
+      integrity="sha512-uf06llspW44/LZpHzHT6qBOIVODjWtv4MxCricRxkzvopAlSWnTf6hpZTFxuuZcuNE9CBQhqE0Seu1CoRk84nQ=="
+      crossorigin="anonymous"
+      referrerpolicy="no-referrer"
     />
-
     <link
       rel="stylesheet"
-      crossorigin
-      href="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.65.6/addon/fold/foldgutter.min.css"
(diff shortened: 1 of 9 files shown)


website/static/service-worker.js
```diff
@@ -15,33 +17,33 @@ toolbox.precache([
   "lib/parsers-location.js",
   ...plugins,
   "playground.js",
-  "https://unpkg.com/sw-toolbox@3.6.0/sw-toolbox.js",
+  "https://cdnjs.cloudflare.com/ajax/libs/sw-toolbox/3.6.1/sw-toolbox.js",
```
kachkaev, line 20: You probably know that but still worth flagging that `sw-toolbox` is deprecated:
https://github.com/GoogleChromeLabs/sw-toolbox

Also 3.6.1 does not exist on NPM (https://www.npmjs.com/package/sw-toolbox/v/3.6.1), only on GitHub (https://github.com/GoogleChromeLabs/sw-toolbox/releases/tag/v3.6.1). Not sure where CloudFlare pulls the artifact from, but it seem to be doing that correctly 🤔

website/pages/playground/index.html
```diff
@@ -28,45 +28,154 @@
...
+      referrerpolicy="no-referrer"
+    ></script>
     <link rel="stylesheet" href="/separate-css/playground.css" />
 
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <script
       async
       src="https://www.googletagmanager.com/gtag/js?id=UA-111350464-1"
+      crossorigin="anonymous"
+      referrerpolicy="no-referrer"
```
kachkaev, line 178: [gtag.js docs](https://developers.google.com/tag-platform/gtagjs/install) don’t have these extra attributes. They [might break tracking](https://support.google.com/tagmanager/thread/98360689/script-src-for-gtag-and-crossorigin-anonymous-does-not-load-the-js-file-from-cdn?hl=en), but I’m not 100% sure.

fisker: I don't think it's true. But let's trust google "Don't be evil".

website/static/service-worker.js
fisker: Downgraded to 3.6.0

fisker: https://github.com/prettier/prettier/issues/8763#issuecomment-663385515

Issue #13858, opened 2022-11-18
Bug: Separate comments are put onto one line for type "Number | void"


jameshfisher:
**Prettier 2.7.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VACAQgVQOJIZwBucATgJ4wAWAllAOYYNwwDOGb1EpAJhtBgQBUOBgA29OABoM9MOICuvek0gBbdQnYAdWBQAOYgGIQIWAIZkMAXgx6MGAD4YAcovUAjchnQYatBy8EGAe2hw0YgBmtGRsMAIGMLTQDs4YxBC0-H4BQSFhsBHUYmxwkFD8EEkpUGkA3Hp6fgDqEGQA1hwWHHAAHkZg8Lx6MIYmZpbWdmku8WSqvph5GMGhWkX+JRgxcQnVyalQji6Z2UtbgasFG+xbpeXQVTVHjo11UK3tXRg9RAPlYajcYYUzmKy2ezHdLzRa5Oj5dbhe47WLxRKHD4nDBQDzeazwq5rQp3SKcR6VDG1BogaQgA61NjIUBWMgQADuAAUrAgmSgLOJ2RYKEy6Z4yBYwB1WABlCxaAAyUmQUQFZTFEqlsoMktUyBgZEUMhAcC8cF4vHNCosjEUFhYpjI6gsMGSjGQIAsiiEtJA1Bg6nELTo8DYOrAcBlvNoyWIMYoHrAbFFIHoZTIME5EoYzpVauNACs2H0ZapxHAAIqKCDwPPidUgHVxcgesZGNhgBZJX0GBawFrZGjIAAcAAY6b2IGUWhKDB7e3B06RfQBHavwLPVPmetgAWlE5vNvrIcDXsTgWftuaQqvrxrK6lo+sN97LlfXcDrDZgFk8A94Q5IAATHSBoWLQkiMAAwhAmgWB6i4AKy+ooZQACq-nyt4NsQRoAJKVNoMqdrQSQAIKVDKYzll+cAAL50UAA)

```sh
--parser typescript
```

**Input:**

```tsx
// BUG: everything gets shoved onto one line, including comments
type FooBar = 
  | Number // this documents the first option
  | void // this documents the second option
  ;

// Works as expected
type FooBar = 
  | string // this documents the first option
  | void // this documents the second option
  ;

// Works as expected
type FooBar = 
  | string // this documents the first option
[...]

PR #13860, opened 2022-11-19
Fix union type should be printed in the multi-line variant when there are comments

PerfectPan:
## Description

Fix the problem that type should not be considered as hug type when there are comments (#13858)

## Checklist

### DIFF
--- a/changelog_unreleased/typescript/13860.md
+++ b/changelog_unreleased/typescript/13860.md
@@ -3,18 +3,15 @@
 <!-- prettier-ignore -->
 ```tsx
 // Input
-// BUG: everything gets shoved onto one line, including comments
 type FooBar = 
   | Number // this documents the first option
   | void // this documents the second option
   ;
 
 // Prettier stable
-// BUG: everything gets shoved onto one line, including comments
 type FooBar = Number | void; // this documents the first option // this documents the second option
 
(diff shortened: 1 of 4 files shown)


PerfectPan:
I wonder if I want to contribute to the project like bugfix, it is better to base on next branch or main branch?

changelog_unreleased/typescript/13860.md
```diff
@@ -0,0 +1,18 @@
+#### Fix union type should be printed in the multi-line variant when there are comments (#13860 by @PerfectPan)
+
+<!-- prettier-ignore -->
+```tsx
+// Input
+type FooBar = 
+  | Number // this documents the first option
+  | void // this documents the second option
+  ;
```
fisker, line 9: Can you add this case to test?

PerfectPan: OK，I will do it when I get off work today

fisker: Done. [`dcc74d9` (#13860)](https://github.com/prettier/prettier/pull/13860/commits/dcc74d9459cb80e4a016374918f0daed398dba31)

fisker (approved):

thorn0 (approved):

PR #13882, opened 2022-11-22
Simplify doc print

fisker:
## Checklist


fisker:
@thorn0 What do you think this idea using a string instead of `out` array? 

It should makes much easier to trim indentation https://github.com/prettier/prettier/issues/14152#issuecomment-1377689665.

Didn't mark it ready because I'm not sure about the performance impact yet.

thorn0:
If I had to guess, I'd say the array should be faster, but who knows. It'd be best to fix the benchmark script so that it uses a separate copy of the input file for each tested commit to avoid caching bias (https://github.com/prettier/prettier/pull/13224#issuecomment-1228161394).

PR #13894, opened 2022-11-25
Skip measure width of last string in `printDocToString`

fisker:
## Description

Plugins like `@prettier/plugin-pug` prints a big string instead of doc commands, it cost time when measuring width.
Ideally it should only measure the last line, or even better should measure from `out` when use the `pos`, we should improve it in future.

## Checklist


sosukesuzuki (commented):
Sorry, I don't know the impact of this change. I will read the code properly.

sosukesuzuki (approved):

PR #13896, opened 2022-11-26
Changed recommended way of creating a prettier config file

danielturus:
## Description

This PR changes the recommended way to create a Prettier config file. 
We should use the power of autocomplete in current IDEs and this is made fairly easy in JavaScript.
Other tools / frameworks use this as well, one example would be [TailwindCSS](https://tailwindcss.com/docs/installation)

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -20,13 +20,15 @@ yarn add --dev --exact prettier
 
 <!--END_DOCUSAURUS_CODE_TABS-->
 
-Then, create an empty config file to let editors and other tools know you are using Prettier:
+Then, create an empty `prettier.config.js` file to let editors and other tools know you are using Prettier and add the content below:
+```js
+/** @type {import('prettier').Config} */
+module.exports = {
 
-<!-- Note: `echo "{}" > .prettierrc.json` would result in `"{}"<SPACE>` on Windows. The below version works in cmd.exe, bash, zsh, fish. -->
-
-```bash
-echo {}> .prettierrc.json
+}
 ```
+This way, the editor will autocomplete the options you have available according to your installed version of Prettier.
+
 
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -20,13 +20,15 @@ yarn add --dev --exact prettier
...
+/** @type {import('prettier').Config} */
+module.exports = {
 
-<!-- Note: `echo "{}" > .prettierrc.json` would result in `"{}"<SPACE>` on Windows. The below version works in cmd.exe, bash, zsh, fish. -->
-
-```bash
-echo {}> .prettierrc.json
+}
 ```
+This way, the editor will autocomplete the options you have available according to your installed version of Prettier.
```
fisker, line 30: Shouldn't editor also do autocomplete to json file too? https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/prettierrc.json

kachkaev: JSON files don’t allow `require("some-prettier-plugin")` and not using `require` does not work in Yarn PnP. On the other hand `/** @type {import('prettier').Config} */`, only works if [`@types/prettier`](https://www.npmjs.com/package/@types/prettier) is installed.

danielturus: > On the other hand `/** @type {import('prettier').Config} */`, only works if [`@types/prettier`](https://www.npmjs.com/package/@types/prettier) is installed.

Great observation, maybe this should be added in the docs also?

fisker: image
image

karlhorky: This only works with JSON files. With `.prettierrc.js`, `prettier.config.cjs` etc this does not show errors:

Screenshot 2022-12-08 at 13 23 48

Opened a followup PR for this:

PR #13916, opened 2022-11-29
Improve `cursorNode` find logic

fisker:
## Description

Before this change, it use `findNodeAtOffset` it checkes `canAttachComment` and `isSourceElement`, cursorNode should not care.
After this change, the `cursorNode` can be the most inside node, will make the "cursorNodeText" smaller and easier to locate new cursor.

## Checklist

### DIFF
--- a/src/language-css/loc.js
+++ b/src/language-css/loc.js
@@ -232,11 +232,11 @@ function replaceQuotesInInlineComments(text) {
 }
 
 function locStart(node) {
-  return node.source.startOffset;
+  return node.source?.startOffset;
 }
 
 function locEnd(node) {
-  return node.source.endOffset;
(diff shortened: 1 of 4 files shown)


sosukesuzuki (commented):
Is this a just refactoring? If no, please add tests( change logs would not be necessary )

fisker:
> Is this a just refactoring?

No, it makes the cursorNode smaller. 

> If no, please add tests

It can't be tested, even if we always set cursorNode to the AST root, we'll still print the same result. Only difference is this line will be slower https://github.com/prettier/prettier/blob/985b895c47d1b9da25cde0c0b56376606b9c1559/src/main/core.js#L121.

PR #13959, opened 2022-12-08
Add type checking to Prettier JS config example

karlhorky:
Hi, long time no see Prettier team! As always, thanks for your continued effort here 🙌

## Description

The config example using JS does not include types, nor an example of where to get those types (see also https://github.com/prettier/prettier/pull/13896#discussion_r1032830756)

This unfortunately does not work with `module.exports = <object>`, because of https://github.com/microsoft/TypeScript/issues/51819:

Screenshot 2022-12-08 at 13 18 36

So using the common pattern of setting a `config` constant to enable this checking:

Screenshot 2022-12-08 at 13 18 54

## Checklist

cc @kachkaev @fisker

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -33,13 +33,17 @@ JSON:
 JS:
 
 ```js
-// prettier.config.js or .prettierrc.js
-module.exports = {
+// prettier.config.js or .prettierrc.js - install @types/prettier if you want type checking
+
+/** @type {import('prettier').Options} */
+const config = {
   trailingComma: "es5",
   tabWidth: 4,
   semi: false,
   singleQuote: true,
 };
+
+module.exports = config;
 ```
 
 YAML:


kachkaev:
I use typings like this for ESLint and they work without `const config`:

```js
/** @type {import("eslint").Linter.Config} */
module.exports = {
  /* ... */
}
```

I’m not sure if this change is useful for Prettier though. Unlike ESLint configs, Prettier configs usually very small. IDE autocompletion may be enough for them and it’s not worth bothering with installing `@types/prettier`. I have Prettier typings in a couple of projects but only when I need to format something inside my JS code. Not sure we prompt people to install `@types/prettier` as this complicates onboarding but has limited practical gains.

karlhorky:
> IDE autocompletion may be enough for them

autocomplete doesn't work for JS files, as I mentioned here: https://github.com/prettier/prettier/pull/13896#discussion_r1043297304

karlhorky:
> I’m not sure if this change is useful for Prettier though

What about an alternative - instead of editing the only JS example, show a separate section at the bottom for users who want to use types?

fisker:
If you have just one place/project to config, I suggest config while looking at the docs, I don't think it's hard to find.

If you have multiple projects to config, I suggest publish a shared config to npm, so you do't need do the samething in all these projects.
Of course you can install @types/prettier when making your shared config. You can also reexport @types/prettier in this package, so you don't need install extract package if you want make small adjustments in specific project.

karlhorky:
> Of course you can install `@types/prettier` when making your shared config

Right, however this and the JSDoc `import()` syntax / export path are not something that is obvious to everyone, and they are different for every developer tool that has configuration:

- some have built-in types
- some have types under the DefinitelyTyped package (as Prettier does)
- some have types under a completely different package (check `remark` config - comes from `unified` package)
- some do not call the type export `.Config`, as you would imagine (see `eslint`)

It's impossible to know these details about Prettier up front.

So what I am proposing is that in the installation guide for this developer tool show these details. Does not need to be the first JS example, but it should be there somewhere.

karlhorky:
@fisker @kachkaev can I ask for a re-review of the points above which haven't had a response yet?

Summarizing:

1. It can be a separate example, further down in the documentation - does not need to change the only JS example
2. It is not easy to know what export to import from `@types/prettier`, even if you're making a shared config
3. IDE autocompletion does not work in JS files

karlhorky:
I've created a new PR for this approach of including the new config in a separate block: https://github.com/prettier/prettier/pull/14063

PR #14004, opened 2022-12-16
Add `ancestors` getter and `hasAncestor()` to `AstPath`

fisker:
## Description

~It seems accidentally fixed issues.~ Mistake

## Checklist


src/common/ast-path.js
```diff
@@ -208,30 +212,46 @@ class AstPath {
...
+    }
 
-      name = this.stack[stackPointer--];
-      node = this.stack[stackPointer--];
+    return false;
+  }
+
+  /**
+   * @internal Unstable API. Don't use in plugins for now.
+   */
```
thorn0, line 246: It's #private, so this comment isn't needed

PR #14034, opened 2022-12-18
Do not change case of property name if inside a variable declaration in LESS

mvorisek:
## Description

Simple example:
```
@var: {
  camelCase: 0;
};

div {
  margin: @var[camelCase];
};
```

when `preserveCase` is changed to `preservecase`, LESS compilation fails.

This PR fixes such problem and handles also variable declaration with nested ruleset correctly.

## Checklist

### DIFF
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -155,7 +155,12 @@ function genericPrint(path, options, print) {
 
       return [
         node.raws.before.replace(/[\s;]/g, ""),
-        insideICSSRuleNode(path) ? node.prop : maybeToLowerCase(node.prop),
+        (parentNode.type === "css-atrule" &&
+          typeof parentNode.raws.params === "string" &&
+          parentNode.raws.params.startsWith(":")) ||
+        insideICSSRuleNode(path)
+          ? node.prop
+          : maybeToLowerCase(node.prop),
         trimmedBetween.startsWith("//") ? " " : "",
         trimmedBetween,
         node.extend ? "" : " ",
--- a/src/language-css/printer-postcss.js
+++ b/src/language-css/printer-postcss.js
@@ -155,9 +155,8 @@ function genericPrint(path, options, print) {
 
       return [
         node.raws.before.replace(/[\s;]/g, ""),
-        (parentNode.type === "css-atrule" &&
-          typeof parentNode.raws.params === "string" &&
-          parentNode.raws.params.startsWith(":")) ||
+        // Less variable
+        (parentNode.type === "css-atrule" && parentNode.variable) ||
         insideICSSRuleNode(path)
           ? node.prop
           : maybeToLowerCase(node.prop),
--- a/tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
@@ -416,7 +416,7 @@ printWidth: 80
 };
 
 @var: /* comment */ {
-  preservecase: 5;
+  preserveCase: 5;
 };
 
 @var: /* comment */ {
@@ -424,7 +424,7 @@ printWidth: 80
 };
 
 @var: {
-  preservecase: 5;
+  preserveCase: 5;
 };
 
 @var // comment
--- a/tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
@@ -380,8 +380,12 @@ printWidth: 80
 };
 
 @not-var {
-  notVar: @var[@notVarNested][notVar];
-  notVar2: @var[notVar];
+  shouldNotChangeCase: @var[@notVarNested][notVar];
+  shouldNotChangeCase: @var[    @notVarNested][notVar];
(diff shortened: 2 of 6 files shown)


fisker (dismissed):

fisker:
@mvorisek The tests are failing.

mvorisek:
> @mvorisek The tests are failing.

@fisker any idea how to fix it and why does it fail on Node 18 only?

fisker:
`Use  `FULL_TEST` to debug https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md#deeper-testing

mvorisek:
(my/added) test fixed, thanks for `FULL_TEST` hint

fisker:
What cases the AST changes? I can't see what you changed.

mvorisek:
Test to compile the LESS code from the PR description with `camelCase` property name changed to `camelcase`. LESS compilation will fail, case in a declaration must be preserved as it is case sensitive when referenced.

fisker:
I how what are we fixing, I mean how did you fix the FULL_TEST, your code changes the AST at first, that's why the test failing. I want know how did you fix that.

mvorisek:
sure, removed this not needed testcase I initially proposed:

image

fisker:
I'm not familiar with LESS. Is it valid LESS code? Should we fix it instead of removing it.

mvorisek:
here is minimalistic test case:

```
@varFollowedBySpace : {
  camelCase: 0;
};

div {
  margin: @varFollowedBySpace[camelCase];
};
```

in https://lesscss.org/less-preview/ I can compile it /w and /wo space to the same result

in https://prettier.io/playground/ when space is removed, the AST is different

not sure if really a bug in Prettier, maybe a bug in LESS parser? It is not related with this PR nor my needs, so feel free to pursue it, but I have no capacity to do so.

fisker:
Can you add this test back, I can look into it later.

fisker:
The difference is `.value` of `css-atrule`, right?

fisker:
Thanks.

mvorisek:
> The difference is `.value` of `css-atrule`, right?

I have added two failing test cases:

first is definitely in Prettier and should be easy to fix

2nd is the problem we discussed here, ie. different AST when at rule has a whitespace between colon:

```diff
-@var2 : {
+@var2: {
   notVar: 5;
 };
```

fisker:
I'll take a look later, hopefully tomorrow.

tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -321,3 +321,131 @@ TABLE {
...
+: {
+  shouldnotchangecase: 5;
+}
+
+@not-var {
+  shouldnotchangecase: @var[ @notVarNested][notVar];
+  shouldnotchangecase: @var[ @notVarNested][notVar];
+  shouldnotchangecase: @var[ @notVarNested ][notVar];
+  shouldnotchangecase: @var[ @notVarNested][ notVar];
+  shouldnotchangecase: @var[ @notVarNested][notVar ];
```
fisker, line 446: We are fine with these spaces, right?

src/language-css/printer-postcss.js
```diff
@@ -155,7 +155,12 @@ function genericPrint(path, options, print) {
 
       return [
         node.raws.before.replace(/[\s;]/g, ""),
-        insideICSSRuleNode(path) ? node.prop : maybeToLowerCase(node.prop),
+        (parentNode.type === "css-atrule" &&
+          typeof parentNode.raws.params === "string" &&
+          parentNode.raws.params.startsWith(":")) ||
```
fisker, line 160: Should we use `parentNode.variable` instead?

fisker: Ah, fixed cases with comments  https://github.com/prettier/prettier/pull/14034/commits/fd414fbd8dd79bc809f465c19374aec170e2b1c0

fisker (approved):

sosukesuzuki (approved):

sosukesuzuki:
@mvorisek Are you okay with changes by @\fisker ? If you don't have concerns we can merge this PR.

src/language-css/printer-postcss.js
mvorisek: `parentNode.variable` seems fine/better

tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
mvorisek: I would expect no trailing space, especially when space is not present in the original. Can it be fixed easily?

mvorisek:
https://github.com/prettier/prettier/pull/14034#discussion_r1059312395 would be nice to be fixed as well, otherwise I have no objections.

tests/format/less/case/__snapshots__/jsfmt.spec.js.snap
fisker: I didn't notice the space is added.

fisker:
https://github.com/prettier/prettier/pull/14034#discussion_r1059315108

Let's merge this first, we can try to fix it in other PR.

PR #14062, opened 2022-12-25
bash changed to shell to enable copy functionality

mishrasur7:
shell command will enable the copy functionality and copies commands into clipboard.

## Checklist

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -8,13 +8,13 @@ First, install Prettier locally:
 <!--DOCUSAURUS_CODE_TABS-->
 <!--npm-->
 
-```bash
+```shell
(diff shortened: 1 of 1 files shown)


fisker:
Can't see the difference https://deploy-preview-14062--prettier.netlify.app/docs/en/install.html

mishrasur7:
Alright, I now see it does not make difference either its bash or shell in https://deploy-preview-14062--prettier.netlify.app/docs/en/install.html. But we can see the copy button in prettier/docs/install.md file while opening through GitHub. Hmm how can we make this to work? I will do some work if I get some time.

PR #14063, opened 2022-12-25
Add types to config example

karlhorky:
Hi there :) As always, thanks for your continued effort on Prettier 🙌

## Description

The config example using JS does not include types, nor an example of where to get those types (see also https://github.com/prettier/prettier/pull/13896#discussion_r1032830756)

This unfortunately does not work with `module.exports = <object>`, because of https://github.com/microsoft/TypeScript/issues/51819:

Screenshot 2022-12-08 at 13 18 36

So using the common pattern of setting a `config` constant to enable this checking:

Screenshot 2022-12-08 at 13 18 54

As noted in https://github.com/prettier/prettier/pull/13959, it may be better to not change the original JS example.

However, there should be some example using the types via JSDoc, because:

1. [It is not easy to know what export to import from `@types/prettier`, even if you're making a shared config](https://github.com/prettier/prettier/pull/13959#issuecomment-1342968837)
2. IDE autocompletion does not work in JS files

## Checklist


docs/configuration.md
```diff
@@ -42,6 +42,22 @@ module.exports = {
 };
 ```
 
+JS with type checking (eg. `checkJs: true` in `tsconfig.json`):
```
fisker, line 45: Maybe we can use this  `DOCUSAURUS_CODE_TABS` comment https://github.com/prettier/prettier/blob/948b3af232d9be1cf973a143382cad3ebe33725a/docs/install.md?plain=1#LL8C21-L8C21

fisker:
Can you copy changes to this file  `/website/versioned_docs/version-stable/configuration.md`? So we can see it on website.

karlhorky:
Thanks for the tip @fisker, here's the first version ([deployed docs](https://deploy-preview-14063--prettier.netlify.app/docs/en/configuration.html)):

Screenshot 2022-12-27 at 17 15 28

Screenshot 2022-12-27 at 17 15 20

karlhorky:
Seems like it's removing my extra newlines in my code block for whatever reason, maybe there's a way to add extra lines?

docs/configuration.md
```diff
@@ -42,6 +45,25 @@ module.exports = {
 };
 ```
 
+<!--JSDoc types-->
+
+```js
+// prettier.config.js or .prettierrc.js
+// Enable JS type checking (eg. `checkJs: true` in `tsconfig.json`) and install @types/prettier
+
+/** @type {import('prettier').Options} */
```
fisker, line 54:
```suggestion
/** @type {import("prettier").Options} */
```

docs/configuration.md
```diff
@@ -42,6 +45,25 @@ module.exports = {
 };
 ```
 
+<!--JSDoc types-->
+
+```js
+// prettier.config.js or .prettierrc.js
+// Enable JS type checking (eg. `checkJs: true` in `tsconfig.json`) and install @types/prettier
```
fisker, line 52: This part should be improved.

fisker:
@karlhorky We have builtin types now, so `@type/prettier` is not required, would you mind update this? The tabs can be kept, but add both example for CommonJS and ESM format of config.

karlhorky:
@fisker sure, updated to the following (moved out of a comment so that we can link to the [Install docs page](https://prettier.io/docs/en/install.html))

```
Enable JS type checking (eg. `checkJs: true` in `tsconfig.json`) and [install Prettier](install.md)
```

karlhorky:
oh is the `export default <obj>` actually correctly type-checked?

with `module.exports = <obj>` it is not:

- https://github.com/microsoft/TypeScript/issues/47107
- https://github.com/microsoft/TypeScript/issues/51819

karlhorky:
Added a TypeScript Playground above showing what I mean with the object literals not being type checked.

I also usually avoid `export default { ... }` for stylistic reasons ([`eslint-plugin-import/no-anonymous-default-export`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-anonymous-default-export.md))

fisker:
> Added a TypeScript Playground above showing what I mean with the object literals not being type checked.

I was just testing tabs.


After that, I did a test,

image

Does this mean we can use `module.exports = {};` ?

karlhorky:
I don't think that it should be used. In my experience:

- hovering / autocompletion works
- type-checking on incorrect properties breaks
- type-checking on incorrect values breaks

I don't think these breakages are worth it for avoiding the extra `config` variable

fisker:
The reason I try to make it work, because not everyone wants to do type check, for them, I hope they can just copy can delete the first line. But if it doesn't I can accept use a variable.

fisker:
https://deploy-preview-14063--prettier.netlify.app/docs/en/configuration.html#basic-configuration

Good?

fisker (approved):

karlhorky:
@fisker ah sorry, missed the notification about your message above. Yeah, looks good!

One nit: I would put a space before the opening parentheses - but just a super small format thing.

PR #13930, opened 2022-12-01
[handlebars] allow custom "else if"-like with block params

jamescdavis:
## Description

This is a follow-up to #13507 to allow block params in custom "else if"-like blocks.

```hbs
{{! Input }}
{{#when isAtWork as |work|}}
  Ship that
  {{work}}!
{{else when isReading as |book|}}
  You can finish
  {{book}}
  eventually...
{{else}}
  Go to bed!
{{/when}}

{{! Prettier stable }}
{{#when isAtWork as |work|}}
  Ship that
  {{work}}!
{{else when isReading}}
  You can finish
  {{book}}
  eventually...
{{else}}
  Go to bed!
{{/when}}

{{! Prettier main }}
{{#when isAtWork as |work|}}
  Ship that
  {{work}}!
{{else when isReading as |book|}}
  You can finish
  {{book}}
  eventually...
{{else}}
  Go to bed!
{{/when}}
```

## Checklist

- [x] I’ve added tests to confirm my change works.
- [ ] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~~
- [x] (If the change is user-facing) I’ve added my changes to `changelog_unreleased/*/XXXX.md` file following `changelog_unreleased/TEMPLATE.md`.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

```
--parser glimmer
```

CC: @dcyriller

PR #14067, opened 2022-12-27
[handlebars] group params in opening block statements

jamescdavis:
## Description

This is a follow-up to #13930 to establish wrapping consistency between opening block statements and else blocks by grouping params in opening blocks. This causes params to break to a new line together and not be split across lines unless the length of params exceeds the print width. This also updates the else block wrapping to behave exactly the same as opening blocks (after #13930 it attempts to keep params on same line as the "else block" keywords, even when block params cause wrapping, see `else block` in the `Prettier main` example below).
```
--print-width 80
```

```hbs
{{! Input }}
{{#block param param param param param param param param param param as |blockParam|}}
  Hello
{{else block param param param param param param param param param param as |blockParam|}}
  There
{{/block}}

{{! Prettier main (includes unreleased #13930) }}
{{#block
  param
  param
  param
  param
  param
  param
  param
  param
  param
  param
  as |blockParam|
}}
  Hello
{{else block param param param param param param param param param param
  as |blockParam|
}}
  There
{{/block}}

{{! This PR }}
{{#block
  param param param param param param param param param param
  as |blockParam|
}}
  Hello
{{else block
  param param param param param param param param param param
  as |blockParam|
}}
  There
{{/block}}
```

## Checklist

- [ ] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory)~~.

```
--parser glimmer
```


src/language-handlebars/printer-glimmer.js
```diff
@@ -533,21 +533,24 @@ function printOpenBlock(path, print) {
...
-    attributes.push(line, block);
+    blockParams = [line, printBlockParams(node.program)];
   }
 
   return group([
     openingMustache,
-    indent(attributes),
+    indent(blockKeyword),
+    indent(params),
+    indent(blockParams),
```
fisker, line 553: We can use something like

```js
join(line, [
	keywordDoc,
	paramsDoc,
	blockParamsDoc
])
```

fisker:
@jamescdavis I don't use handlebars, but is this a bug?

**Prettier 2.8.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuExgGIBGAbCYDWABAIYC+pAOlIYQIxVpzYDOchAlgGaGblU0AmBsCas+1QgGZhAehx585EABoQEAA4x20ZslDEATgYgB3AAqGEulMWwniAT12rMB4gTgwAysQC2cABl2KDhkTltWFzcPb3V3YIBzZBgDAFc4VThfTDgAE1y8gOIoBNTiBLgAMQgDX2IYLRLkEGJUmAgVEAALGF9sAHUu9nhmOLA4Lyth9gA3YYdmsGZnEGDWAxgzNwS6sIiMkAArZgAPL0TsOABFVIh4PZYDuIN15oTsdl9-A071A2CYP12LkYF1kAAOAAMqj+EFY-Tc6mafzg6xmoVUAEdbvAthprC1mABaEJ5PKdAxwbHsSlbcq7JDhR6qVi+djJNIHZgXOAAQQa-0wbTgZjgBiCIQekRA3JKlxud1CjP2qhgxEwQJBYKQAlVbnYHxKAGEIF9iM1UQBWTqpVgAFXV1iZ0pm6QAklACrAvGB-ppeZ6vDAHJcpXByEA)

```sh
--parser glimmer
```

**Input:**

```hbs
{{#block a}}
  1
{{else if b}}
  2
{{else}}
  3
{{/block}}
```

**Output:**

```hbs
{{#block a}}
  1
{{else}}{{#if b}}
    2
  {{else}}
    3
  {{/if}}{{/block}}
```

I didn't mean `else if` is splited, I mean should `{{/block}}` align with the opening?

fisker (approved):

fisker:
@jamescdavis I made some changes, please take a look.

jamescdavis:
@fisker, your changes look good. Thanks for simplifying the logic!

jamescdavis:
As far as the closing block alignment, it's due to the `--html-whitespace-sensitivity` setting in combination with prettier rewriting the `else if`.

When

```hbs
{{else if b}}
```

is rewritten to

```hbs
{{else}}{{#if b}}
```

and

```hbs
{{/block}}
```

is rewritten to

```hbs
  {{/if}}{{/block}}
```

and `--html-whitespace-sensitivity` is `css` or `strict`, it does not add whitespace between the statements.

See this [playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuExgGIBGAbCYDWABAIYC+pAOlIYQIxVpzYDOchAlgGaGblU0AmBsCas+1QgGZhAehx585EABoQEAA4x20ZslDEATgYgB3AAqGEulMWwniAT12rMB4gTgwAysQC2cABl2KDhkTltWFzcPb3V3YIBzZBgDAFc4VThfTDgAE1y8gOIoBNTiBLgAMQgDX2IYLRLkEGJUmAgVEAALGF9sAHUu9nhmOLA4Lyth9gA3YYdm9gSoGtDVYNYDGDM3BLqwiIyQACtmAA8vROw4AEVUiHgDliO4g03mhOx2X38DTvUDMEYP12LkYF1kAAOAAMqgBEFY-Tc6maALgmxmaxAAEd7vAdhprC1mABaEJ5PKdAxwXHsak7cr7JDhZ6qVi+djJNJHZhXOAAQQagMwbTgZjgBiCISekRAvJK1zuD1CzMOqhgxEwILBEKQAnVbnYXxKAGEID9iM10QBWTqpVgAFU11hZspm6QAklACrAvGBAZp+d6vDAHNcZXByEA) with `--html-whitespace-sensitivity ignore`.

You also see that explicitly writing this with newlines works fine, even with `--html-whitespace-sensitivity strict`: [playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuExgGIBGAbCYDWABAIYC+pAOlIYQIxVpzYDOc5VNa6AlgGaGZ21GoQBMHQoxZtKwmgGYJaAPR8hKnHnzkQAGhAQADjG7RmyUMQBOViAHcACtYTmUxbHeIBPc-sxXiAjgYAGViAFs4ABluKDhkXndWPwCg0MNA2IBzZBgrAFc4fThwzDgAE3KKqOIoLPziLLgAMQgrcOIYEzrkEGJ8mAg9EAALGHDsAHUR7nhmDLA4EJdZ7gA3Wa9e5jzuMBhh2NYrGAcArI6EpKKQACtmAA8Q7Ow4AEV8iHgr6X0Mq2OvSy2G44UiVmGhissRgk245RgI2QAA4AAx-WysSYBQy9KFwY5reL6ACOn3gZyMrj6zAAtHEKhVhlY4GTuCyzo1LkhEr8QKxwtxcgUbswXnAAIJdaGYAZwBxwKwxOI-ZL88UfL7xHnXfQwYiYOEIpFIUR6gLcEF1ADCEDBxF6BIArMN8qwACoG1y8tVrQoASSgVVgITA0OMEqDIRgXleqpkQA)

In any case, this would only be an issue when the `else if` is rewritten, which would only happen when the keyword that follows `else` doesn't match the opening block keyword. And TBH I'm not even sure that construct makes sense.

jamescdavis:
@fisker I did notice one issue after your changes: when the `if` in `{{else if` goes beyond the print width causing the `if` to wrap to the next line, the `if` is not indented properly: [playground link](https://deploy-preview-14067--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuExgGICWAzABAQwF9CAdKXXARjLTgBsBnOXHXAI2LIoCYbh6mncrgDMfAPQ5iIADQgIABxiZoDZKHwAnTRADuABS0I1KfHV34AnmrltN+MAGs4MAMr4AtnAAymKHGRsMyZbeycXVwUHPwBzZBhNAFc4OTgPNjgAE0ys73woGMT8GLgAMQhND3wYZQLkEHxEmAhZEAALGA86AHU2zHgGKLA4V2N+zAA3fst6sAYbED8mTRh9exiqwOCUkAArBgAPV1i6OABFRIh4LcYdqM1l+pi6TA8vTVaFTT8YbsxMmBtZAADjkXwgTG69gU9S+cGWEwCcgAjpd4GtFCYGgwALT+LJZVqaOCozDEtbFTZIIK3ORMDyYeJJHYME5wACCNW+bCacH0cE0vn8NxCIFZBVOFyuAWp2zkMHwbD+AKBSG48vsmBeBQAwhA3vh6vCAKytRJMAAqipMNNFE2SAEkoDlYK4wN8lOzna4YJZTiK4MQgA)

I've added a failing test and a fix for this.

But also, maybe it _shouldn't_ wrap here? Maybe `else if` should always stay on the same line, even when it exceeds the print width? Maybe I was overzealous with `group`? 😆

jamescdavis:
I actually decided to make the change to not group `else if` so that it never wraps the `if` (or other block keyword) to a new line. This feels better and is actually what [happens now](https://prettier.io/playground/#N4Igxg9gdgLgprEAuExgGICWAzABABwEMAnQgWwJPMtIqNoF8GAdKXXARlbTgBsBnOLhw1q9MVTJNW7AEzdgAehxMQAGhAR8MTNH7JQJYhADuABRIJ9KQrxOEAnvo0AjUmADWcGAGVycABlMKDhkbFtBV3cvXyIwYIBzZBhiAFc4DTgyFzgAE1y8gMIoBNTCBLgAMQhiMkIYHRLkEEJUmAh1EAALGDJeAHUuzHh+OLgfK2HMADdhh2bMBKga0I1gwWIYM1IEurCIjJAAK34ADx9E3jgARVSIeH2BQ-oN5oTeTDIyOGJO-GJgjB+phcjAusgABwaf4QQT9Uj4Zr-OAbaarEAARzu8G2WmsLX4AFoQnk8p1iHAsZgKdtynskOEnhpBGRMMk0od+Jc4ABBBoAlxtOBmH5BEKPSIgLklK63e6hBkHDQwQguYGg8FIWTK0iYD4lADCEC+hGaKIArJ1UoIACqq6yMyXTdIASSgBVgPjAAO0PPdPhgDiuErgTCAA) (param wrapping is mess on stable, but should be in good shape after this PR). LMK, if you don't agree.

tests/format/handlebars/block-statement/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -570,6 +624,86 @@ printWidth: 80
   Another thing
 {{~/if~}}
 
+<div>
```
fisker, line 627: You can put this test in a separate dir with small printWidth, instead of deep.nest.

fisker:
> Prettier probably shouldn't change how the code will be interpreted by the compiler and so probably should just leave

Agree, but this problem can be addressed in another PR.

fisker:
>  I did notice one issue after your changes: when the if (or other block keyword) in {{else if goes beyond the print width causing the if to wrap to the next line, the if is not indented properly.

I removed `indent()` on it since `indent` should alway start with `line`, if you want `indent(["else", line, "something else"])`, you should write `["else", indent([line, "something else"])]`, but I can be wrong, to be honest, I don't really understand how `indent` works.

fisker:
> Which do you think is better?

I nerver use handlebars, I can't tell.

@dcyriller Do you have time to review? We plan to make a release, and I was hoping this to be included.

dcyriller (approved):
Hey - I have been afk for a few weeks.

jamescdavis:
@fisker I moved the test as you [suggested](https://github.com/prettier/prettier/pull/14067#pullrequestreview-1233202678). With the approval from @dcyriller, this should be GTG?

fisker (approved):

fisker:
@jamescdavis Thanks for your work. 
@dcyriller Thanks for your time too.

PR #13952, opened 2022-12-05
Add test cases for questioning ternaries

rattrayalex:
## Description

Extracts test cases that I added for https://github.com/prettier/prettier/pull/13183 to make it easier to visualize the changes that PR makes, as @vjeux [requested](https://github.com/prettier/prettier/pull/13183#discussion_r990820889).

## Checklist

- [x] I’ve added tests to confirm my change works.
- [x] I’ve read the [contributing guidelines](https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md).

**✨[Try the playground for this PR](https://prettier.io/playground-redirect)✨**

PR #14143, opened 2023-01-09
Remove and ignore `.node-version` file

fisker:
## Description

It's added in #13952, I don't think we want add this to the repo. @rattrayalex Should I add it to `.gitignore`?

## Checklist


rattrayalex (approved):
Ah shoot, so sorry if I added that on accident! Thank you for fixing!

Yes, I would suggest `.gitignore` instead since some people may need to add `.node-version` to contribute and it's easy to forget to remove it.

But, it's not a big deal, and just deleting it is fine too :) so approving.

PR #14145, opened 2023-01-09
Handlebars:  Preserve triple brackets in attribute modifier

### LINKED ISSUE #1405: define a single rule for operators line break
Hi everyone, how are you?

Playing with prettier (v1.2.2) I noticed that the formatter uses two different rules for the operators' line break and I think it's better if we keep only a single rule for it.

example src file
```javascript
const test1 = typeof URL === "function" && typeof history === "object" && typeof history.replaceState === "function"
  
const test2 = typeof URL === "function" ? "Lorem Ipsum is simply dummy text" : "printing and typesetting industry.";
```

current formatter
```javascript
const test1 =
  typeof URL === "function" &&
  typeof history === "object" &&
  typeof history.replaceState === "function";

const test2 = typeof URL === "function"
  ? "Lorem Ipsum is simply dummy text"
  : "printing and typesetting industry.";
```

using a single style (in my example: `after` style)
```javascript
const test1 =
  typeof URL === "function" &&
  typeof history === "object" &&
  typeof history.replaceState === "function";

const test2 = typeof URL === "function" ?
  "Lorem Ipsum is simply dummy text" :
  "printing and typesetting industry.";
```

I take the idea of this rule from: http://eslint.org/docs/rules/operator-linebreak

What do you think?

fisker:
## Description

Fixes #1405

## Checklist

### DIFF
--- a/src/language-handlebars/printer-glimmer.js
+++ b/src/language-handlebars/printer-glimmer.js
@@ -122,7 +122,24 @@ function print(path, options, print) {
     }
 
     case "ElementModifierStatement": {
-      return group(["{{", printPathAndParams(path, print), "}}"]);
+      const start = locStart(node);
+      const end = locEnd(node);
+      let openingMustache = "{{";
+      let closingMustache = "}}";
+      // https://github.com/prettier/prettier/issues/14090
+      if (
+        options.originalText.slice(start, start + 3) === "{{{" &&
+        options.originalText.slice(end - 3, end) === "}}}"
+      ) {
+        openingMustache = "{{{";
+        closingMustache = "}}}";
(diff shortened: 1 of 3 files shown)


sosukesuzuki (commented):
Prettier Handlebars support is called "Ember/Handlebars". So , I don't think we needs to support syntax that isn't supported by Ember/Glimmer.

Please see 2.3.0 blog https://prettier.io/blog/2021/05/09/2.3.0.html#ember--handlebars

reteps:
Hi there @sosukesuzuki @fisker -- How best to replicate this with a prettier plugin?

Issue #14006, opened 2022-12-17
Copy button for commands on click


mishrasur7:
Hey, is it possible to implement a copy button to copy prettier commands? it provides good user experience. 

**Environments:**

- Prettier Version: 2.8.1
- Running Prettier via: 
- Runtime: 
- Operating System: 
- Prettier plugins (if any): 

**Steps to reproduce:**

**Expected behavior:**

**Actual behavior:**

kachkaev:
Hi @mishrasur7! Could you please elaborate? Not sure I understand what you mean.

mishrasur7:
Hi @kachkaev! 
I mean to say, can we implement a functionality to copy commands into clipboard so its easy for users to just click and it gets copied to clipboard and can be pasted in command line?

mishrasur7:
For example; we can copy this command to install redux toolkit easily from their docs. 

<img width="829" alt="Screenshot 2022-12-17 at 19 33 17" src="https://user-images.githubusercontent.com/75450225/208254204-8c44d3d4-ed89-4294-8600-9c7550268bd4.png">

kachkaev:
Do you mean Prettier docs, e.g. https://prettier.io/docs/en/install.html? I.e. copy buttons here:

<img width="1128" alt="Screenshot 2022-12-17 at 19 28 28" src="https://user-images.githubusercontent.com/608862/208263172-c07347f1-a7aa-44d3-9e62-cee338a13343.png">

mishrasur7:
Exactly!

kachkaev:
Got you, thanks! Feel free to create a PR if you want!

thorn0:
Fixed by #14153

PR #14153, opened 2023-01-10
feat: add copy to clipboard in code blocks

tarunrajput:
## Description

This PR adds a copy to clipboard button to code blocks as part of #14006 

## Checklist

image

### DIFF
--- a/website/static/js/code-block-buttons.js
+++ b/website/static/js/code-block-buttons.js
@@ -1,47 +1,55 @@
+/* global ClipboardJS */
+
 "use strict";
 
-window.addEventListener("load", () => {
-  function button(label, ariaLabel, icon, className) {
-    const btn = document.createElement("button");
-    btn.classList.add("btnIcon", className);
-    btn.setAttribute("type", "button");
-    btn.setAttribute("aria-label", ariaLabel);
-    btn.innerHTML =
-      '<div class="btnIcon__body">' +
-      icon +
-      '<strong class="btnIcon__label">' +
-      label +
-      "</strong>" +
-      "</div>";
-    return btn;
+(function () {
+  const CONTAINER_CLASS_NAME = "code-block-with-actions";
+  const ACTIONS_CONATINER_CLASS_NAME = "code-block-actions";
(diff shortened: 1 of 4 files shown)


website/static/js/code-block-buttons.js
```diff
@@ -0,0 +1,47 @@
+// Turn off ESLint for this file because it's sent down to users as-is.
+/* eslint-disable */
```
fisker, line 2: Please try to fix the problems instead of disable here.

tarunrajput: Sure, I've updated the PR.

thorn0:
@tarunrajput Please make the button's color lighter than the main text.

website/static/overrides.css
```diff
@@ -523,7 +523,7 @@ a {
 }
 
 .hljs {
-  padding: 16px;
```
thorn0, line 526: Let's not change these styles. Instead please move the button up (`top: -5px; right: 16px` so that it doesn't overlap) and show it only when the code block is hovered.

fisker:
docusaurus@2 seems support this feature out of box.

tarunrajput:
> docusaurus@2 seems support this feature out of box.

Yes, docusaurus v2 supports this along with code wrap and many other features. Let me know if you have plans to migrate. thanks.

website/static/js/code-block-buttons.js
```diff
@@ -0,0 +1,47 @@
...
+    btn.setAttribute("aria-label", ariaLabel);
+    btn.innerHTML =
+      '<div class="btnIcon__body">' +
+      icon +
+      '<strong class="btnIcon__label">' +
+      label +
+      "</strong>" +
+      "</div>";
+    return btn;
+  }
```
fisker, line 17: I think the `<div>` is useless here. We can put svg into css (use `::before` or `background-image`), and use `<button>Copy</button>` directly here.

fisker:
Updated to use the same style as docusaurus v2

image
image
image

fisker (approved):

PR #14171, opened 2023-01-13
Rewrite `TypeAnnotation` and `TSTypeAnnotation` print

fisker:
## Checklist


fisker:
This is harder than I thought.

fisker:
@Gregoor I've rewrite the `TSTypeAnnotation` print, should fix #12491 now (tests not added yet).

changelog_unreleased/typescript/14171.md
```diff
@@ -0,0 +1,32 @@
...
+<!-- prettier-ignore -->
+```tsx
+// Input
+let foo/* comment */ : number;
+
+// Prettier stable
+let foo: /* comment */ number;
+
+// Prettier main
+<Same as input>
```
thorn0, line 12: I'm probably missing some context, but why keep it same as input? Looks strange. Prettier usually prints a space in front of a comment and doesn't print a space in front of colons.

thorn0: It makes sense to print the comment in front of `:` as it was there input, but the spacing should be like this:
```ts
let foo /* comment */: number;
```

fisker: Got it.

fisker: @thorn0 the space after comment is printed by [`printLeadingComment`](https://github.com/prettier/prettier/blob/bf406f772802d702d38bda6a961db0a93529c95b/src/main/comments.js#L456) how would you suggest to remove it? Adjust comment attach? Use `willPrintOwnComments`?

thorn0: Yeah, a tricky problem. It should be printed like a trailing comment, but there is no appropriate node to attach it to as trailing. Need to think about this.

fisker: Let's keep it and format as 

`let foo /* comment */ : number;`

for now? This is a edge case after all.

fisker: About the leading space, I also found it's a little bit tricky, the space print inside parent or child also effect the cursor tracking, see [`bb19763`](https://github.com/prettier/prettier/pull/14171/commits/bb197637018eff0e571c81a511e6818e55d15c1c)

I haven't look into it yet, but I think that is a problem should be solved in core.

thorn0: ok

src/language-js/print/type-annotation.js
```diff
@@ -321,6 +296,84 @@ function printJSDocType(path, print, token) {
...
+  if (!typeAnnotation) {
+    return "";
+  }
+
+  let shouldPrintLeadingSpace = false;
+
+  if (
+    (typeAnnotation.type === "TSTypeAnnotation" ||
+      typeAnnotation.type === "TypeAnnotation") &&
+    hasComment(typeAnnotation, CommentCheckFlags.leading) &&
```
thorn0, line 312:
```suggestion
    hasComment(typeAnnotation, CommentCheckFlags.Leading) &&
```

thorn0: Add `/** @enum {number} */` in front of `const CommentCheckFlags = {` for TS to catch this.

fisker: Thanks, but I think I need find a better way to do this, it can exist in other properties, had to cover all cases, maybe use `willPrintOwnComment`.

PR #14191, opened 2023-01-16
Extract `printClassBody`

fisker:
## Description

The `ClassBody` is different than normal statement sequence,

It can't have `directives` and `EmptyStatement` , it have different logic in no-semi mode.

Let's split it.

## Checklist

### DIFF
--- a/src/language-js/print/class.js
+++ b/src/language-js/print/class.js
@@ -9,7 +9,12 @@ import {
   indent,
   ifBreak,
 } from "../../document/builders.js";
-import { hasComment, CommentCheckFlags } from "../utils/index.js";
+import {
+  hasComment,
+  CommentCheckFlags,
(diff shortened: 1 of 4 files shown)


thorn0 (approved):

src/language-js/print/class.js
```diff
@@ -250,9 +265,129 @@ function printClassProperty(path, options, print) {
...
+    return true;
+  }
+
+  switch (nextNode.type) {
+    case "ClassProperty":
+    case "PropertyDefinition":
+    case "TSAbstractPropertyDefinition":
+      return nextNode.computed;
+    case "MethodDefinition": // Flow
+    case "TSAbstractMethodDefinition": // TypeScript
```
thorn0, line 360: Let's remove these comments. `MethodDefinition` is used for TS too.

PR #14199, opened 2023-01-18
Simplify angular expression parser

fisker:
## Checklist

### DIFF
--- a/src/language-js/parse/angular.js
+++ b/src/language-js/parse/angular.js
@@ -1,26 +1,26 @@
-import { locStart, locEnd } from "../loc.js";
+import createParser from "./utils/create-parser.js";
 
-function createParser(_parse) {
-  const parse = async (text, options) => {
-    const ngEstreeParser = await import("angular-estree-parser");
-    const node = _parse(text, ngEstreeParser);
-    return {
-      type: "NGRoot",
-      node:
-        options.parser === "__ng_action" && node.type !== "NGChainedExpression"
-          ? { ...node, type: "NGChainedExpression", expressions: [node] }
-          : node,
-    };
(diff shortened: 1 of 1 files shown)


src/language-js/parse/angular.js
```diff
@@ -1,26 +1,30 @@
...
+/**
+ * @param {"parseAction" | "parseBinding" | "parseInterpolation" | "parseTemplateBindings"} parseMethod
+ */
+function createParse(parseMethod) {
+  return async (text) => {
+    const { [parseMethod]: parse } = await import("angular-estree-parser");
+    let node = parse(text);
+
+    if (parseMethod === "parseAction" && node.type !== "NGChainedExpression") {
+      // @ts-expect-error
```
kachkaev, line 14: As a general rule, it is helpful to clarify the reason for this directive, e.g.

```suggestion
      // @ts-expect-error -- ... is expected to be ... but ... because ...
```

Strictly speaking, there is no need to add ` -- ` but doing so is required in ESLint directives (e.g. `eslint-disable-next-line rule-name -- comment`) It’s a arguably a good thing to add `--` in all directives for ‘symmetry‘.

It is possible to configure [@typescript-eslint/ban-ts-comment](https://typescript-eslint.io/rules/ban-ts-comment/) to so that a comment is required. E.g.:

```js
    "@typescript-eslint/ban-ts-comment": [
      "error",
      {
        "ts-expect-error": "allow-with-description",
        minimumDescriptionLength: 10,
      },
    ],
```

kachkaev (approved):
The diff makes sense, CI is passing, so approving based on that. It might be useful to wait for another reviewer though just in case. My other comment is optional ↓

thorn0 (approved):

Issue #8462, opened 2020-05-31
Proposal: Pretty format multiple css values in row align.


twoco:
**Prettier 2.0.5**

Format multiple values of a css attribute in clean aligned rows.

**Input:**

```css
.some-class {
  box-shadow: 
    0 0 10px 0 black, 
    0 0 0 1px white,
    0 5px 20px -5px rgba(0, 0, 0, 0.3);
}
```

(Also expected.)

**Output:**

```css
.some-class {
  box-shadow: 0 0 10px 0 black, 0 0 0 1px white,
    0 5px 20px -5px rgba(0, 0, 0, 0.3);
}
```

**Expected behavior:**

```css
.some-class {
  box-shadow: 
    0 0 10px 0 black, 
    0 0 0 1px white,
    0 5px 20px -5px rgba(0, 0, 0, 0.3);
}
```

(scss)

thorn0:
Duplicate of #6024

thorn0:
Related: https://github.com/prettier/prettier/issues/7039

Issue #7039, opened 2019-11-23
CSS Multiline declaration should not be on the same line as the property


GrimLink:
**Prettier 1.19.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDABMAOlLMBGEAHgLQDOAFqgCYQDuS6ADM6wIwAOR6ATgOb5UACgDMTADTo2AFglSAnHKYA6ABwBKSSxYAmLlP39BouTNOKta9QG4cAXxDiQEDjACW0MslCoePegAKvgheKKgAbhBu1I4g+DyoYADWcDAAyhyJblB8yDA8AK5wThQwALYANgDqFG7wZJlgcGkhdW7hdQCeyOBkXk7ZZHA8MAEJfGWoyABmqBVDTgBWZEQAQgnJqWmoZXAAMtlwM3MLIMtEadl8FXAAigUQ8MfzxSCZPEM8PWB9sRw82RgVWiMAoyFUEje-iGVQSHB6-zgn3CRycAEcHvAxi5QmgyCQoHA4NRibEeHAMW5yWNUBMpkhZi8nEMym48oVXmQrjd7o8jgyTq8YKh8MDqKDkDonPlUG4KlcAMIQMqTHpQaCokAFIYAFRFoUZp3CRQAklASbA0mAAa4AILmtIwTo3Z5DOx2IA)
```sh
--parser css
```

**Input:**
```scss
a {
  box-shadow: 0 0 0 1px rgba(30, 140, 190, 0.8), 0 0 2px 1px rgba(30, 140, 190, 0.8);
}
```

**Output:**
```css
a {
  box-shadow: 0 0 0 1px rgba(30, 140, 190, 0.8),
    0 0 2px 1px rgba(30, 140, 190, 0.8);
}
```

**Expected behavior:**
```css
a {
  box-shadow:
    0 0 0 1px rgba(30, 140, 190, 0.8),
    0 0 2px 1px rgba(30, 140, 190, 0.8);
}
```

The first value should not be on the same line as the property when wrapped to multiline  declaration.

_Old previous issue #6986_

whatisaphone:
Search keywords: newline comma multiple split lines

I often notice this with `transition` too.

Current formatting:

```css
.box {
  transition: color 0.15s ease-in-out, background-color 0.15s ease-in-out,
    border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
}
```

Ideal formatting:

```css
.box {
[...]

PR #14208, opened 2023-01-20
Fix line break in CSS declaration with comma

### LINKED ISSUE #6024: When box-shadow has multiple values (separated by comma) it'd be prettier to put each value on a separate line

mvorisek:
## Description

Comma in CSS/LESS declaration/value is used to separate group of multiple values of the same meaning.

Currently line break is enforced in selectors like:
```
h1,
h2 {
    ...
}
```

but not for CSS values like:
```
box-shadow: -1px -1px 2px rgba(0, 0, 0, 0.3), 1px -1px 2px rgba(0, 0, 0, 0.3), -1px 2px 2px rgba(0, 0, 0, 0.3);
```

this PR fixes/unifies the behaviour by enforcing line break after comma in CSS declaration/value.

fix #6024, #7039, #8462
close #13260

## Checklist


src/language-css/printer-postcss.js
```diff
@@ -924,12 +924,18 @@ function genericPrint(path, options, print) {
...
+            parentParentParentNode.variable);
+
         const printed = path.map(print, "groups");
-        const res = [];
+        const res = forceHardLine ? [hardline] : [];
 
         for (let i = 0; i < printed.length; i++) {
           if (i !== 0) {
-            res.push([",", line]);
+            res.push([",", forceHardLine ? hardline : line]);
```
mvorisek, line 938: how to fix this lint warning?

mvorisek: @fisker can you please advise how to fix the lint error?

fisker: This may work https://github.com/prettier/prettier/blob/310a5b1ba51ad6a3115670058d53a7ecb1dcd25c/src/language-js/printer-estree.js#L193

mvorisek: Thank you 👍

fisker:
@mvorisek Please check discussion in #6024 and https://github.com/prettier/prettier/issues/6024#issuecomment-1216191051

mvorisek:
I support your https://github.com/prettier/prettier/issues/6024#issuecomment-1209617651 as the wrapping rule should be easy to understand and not dependent on the content IMHO

fisker:
But I agree line breaks in case like [this](https://github.com/prettier/prettier/pull/14208/files#diff-34c9c87dc127dbc18ddefebdc2f51062bb3cace4a73fe78ade1af6ac383c99d0L23) seems unnecessary.

mvorisek:
https://github.com/prettier/prettier/issues/6024#issuecomment-1216191051 proposed to wrap if 2 or more spaces are present in any comma list item.

Where does the 2 spaces come from? What about wrapping when there is at least 1 space?

My reason for 1 space is it will solve the main problem and with zero spaces, any list is well readable on one line/unchanged.

mvorisek:
implemented in 2nd/3rd commit, CSS declarations like `font-family: Arial, sans-serif;` or `$margin: 0, 2em, 0, 1.5em;` are kept on a single line

thank you @fisker for guiding me into the right direction

fisker (approved):
LGTM

src/language-css/printer-postcss.js
```diff
@@ -924,12 +926,27 @@ function genericPrint(path, options, print) {
...
+          (parentParentParentNode.type === "css-atrule" &&
+            parentParentParentNode.variable);
+        if (forceHardLine) {
+          forceHardLine = false;
+          for (let i = 0; i < node.groups.length; i++) {
+            if (node.groups[i].type === "value-comma_group") {
+              forceHardLine = true;
+            }
+          }
+        }
```
fisker, line 941:
```suggestion
    const forceHardLine =
      parentParentParentNode.type === "css-decl" ||
      (parentParentParentNode.type === "css-atrule" &&
        parentParentParentNode.variable &&
        node.groups.some(
          (node) => node.groups[i].type === "value-comma_group"
        ));
```

Same?

mvorisek:
Hi @fisker, can this PR be merged?

fisker:
Need more reviews.

fisker:
@prettier/core Need review.

j-f1 (approved):
Nice improvement!

mvorisek:
@fisker there are more review now, would it be possible to merge this PR so it can be part of the next release?

tests/format/scss/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -224,16 +224,23 @@ selector {
 // #8052
 $font-family-rich:
   // custom
-  "Noto Sans TC", "Noto Sans SC", "Noto Sans JP",
+
```
jerone, line 227: Empty line intentional?

mvorisek: Perfect, thank you ❤️

fisker:
Thank you for your hard work!

fisker:
Excluded custom props for now, but it should allow. Not sure how to fix, 

```
:root {
  --box-shadow:
    inset 0 0 10px #555,
    0 0 20px black;
  box-shadow: var(--box-shadow);
}
```

mvorisek:
> Excluded custom props for now, but it's wrong. Not sure how to fix,

Can you link the fix PR here?

I am not entirely sure what the expected result should be for JSON, I would probably expect the same JSON format as in standard JSON/JS.

fisker:
Commit: https://github.com/prettier/prettier/pull/9583/commits/8ddaa4c53645298a327176a479ac497f8ce5e3b7

> what the expected result should be for JSON

Should stay as-it-is.

alex-kinokon:
This seems a bit too aggressive, see #15070.

viveleroi:
We're using the postcss-modules plugin, which apparently doesn't support multiple lines for `composes`. When prettier rewrites this into multi-line it causes postcss to throw an error.

`composes: body-copy, text-ellipsis from global;`

Error: `referenced class name "body-copy" in composes not found`

But with the inline format above, it works fine. Maybe this is something that the modules plugin needs to support? For now I just use `/* prettier-ignore */` above the line

PR #14210, opened 2023-01-20
Support for CSS language comments; less strict HTML language comments

nsaunders:
## Description

The main objective of this pull request is to add support for CSS language comments for embedded language formatting, e.g.

```javascript
const styleSheet = /* css */`.foo { }`;
```

This builds upon existing support for HTML language comments, e.g.

```javascript
const tmpl = /* HTML */`<button>submit</button>`;
```

> **Note**
> The existing implementation for HTML and GraphQL is case-sensitive and requires leading and trailing spaces. For better compatibility with other tools, CSS language comments are case-insensitive and allow leading/trailing whitespace. I followed suit for HTML; but I left the strict requirements in place for GraphQL because I don't fully understand why they were implemented that way in the first place.

## Motivation/Justification

While the language-comment pattern is undoubtedly less common than using a tag function, it is still a pattern people want to use. Here are a few bits of evidence:
* VS Code extensions
  * [Inline HTML](https://marketplace.visualstudio.com/items?itemName=pushqrdx.inline-html) (~69k downloads)
  * [es6-string-css](https://marketplace.visualstudio.com/items?itemName=bashmish.es6-string-css) (~20k downloads)
* @jaydenseric's [comment](https://github.com/prettier/prettier/issues/4360#issuecomment-383826369) on a Prettier issue
* No-op tagging libraries
  * [jaydenseric/fake-tag](https://github.com/jaydenseric/fake-tag)
  * [lleaff/tagged-template-noop](https://github.com/lleaff/tagged-template-noop)
  * [fabiospampinato/noop-tag](https://github.com/fabiospampinato/noop-tag)
  * [WebReflection/plain-tag](https://github.com/WebReflection/plain-tag)

## Checklist

- [x] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~~ N/A

### DIFF
--- a/src/language-js/clean.js
+++ b/src/language-js/clean.js
@@ -180,9 +180,7 @@ function clean(ast, newObj, parent) {
     const hasLanguageComment = ast.leadingComments?.some(
       (comment) =>
         isBlockComment(comment) &&
-        ["GraphQL", "HTML"].some(
-          (languageName) => comment.value === ` ${languageName} `
-        )
+        comment.value === " GraphQL " || ~["html", "css"].indexOf(comment.value.toLowerCase().trim())
     );
     if (
       hasLanguageComment ||
--- a/src/language-js/embed.js
+++ b/src/language-js/embed.js
@@ -15,7 +15,8 @@ function getLanguage(path) {
     isStyledJsx(path) ||
     isStyledComponents(path) ||
     isCssProp(path) ||
-    isAngularComponentStyles(path)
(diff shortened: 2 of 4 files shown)


nsaunders:
For what it's worth, I just released [babel-plugin-template-css-minifier](https://github.com/nsaunders/babel-plugin-template-css-minifier) which is another tool that facilitates this pattern.

PR #14223, opened 2023-01-23
Support --ignore-pattern as a command line option

Standard8:
## Description

This adds `--ignore-pattern` as a command line option, allowing multiple patterns for ignoring files to be specified on the command line. This echos the [ESLint command line option](https://eslint.org/docs/latest/use/command-line-interface#--ignore-pattern)

This is part of the work for #3460. Related work and discussion is in #12672, however that PR appears to have stalled out.

I'm planning to also add an `ignores` to the configuration, but I'm still figuring out how the structure fits together, as this appears to the first time that option of this type would be added.

I'm happy to do the configuration addition in either this PR or a follow-up, whatever makes most sense - but I thought I'd post this as this part works fine on its own.

## Checklist

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -114,6 +114,14 @@ If you don’t have a configuration file, or want to ignore it if it does exist,
 
 Path to a file containing patterns that describe files to ignore. By default, Prettier looks for `./.prettierignore`.
 
+## `--ignore-pattern`
+
+This allows you to specify patterns of files to ignore (in addition to those in .prettierignore).
+
+```bash
+prettier --ignore-pattern "./ignore**.js" --check .
+```
+
 ## `--list-different`
 
 Another useful flag is `--list-different` (or `-l`) which prints the filenames of files that are different from Prettier formatting. If there are differences the script errors out, which is useful in a CI scenario.
--- a/src/cli/format.js
+++ b/src/cli/format.js
@@ -233,7 +233,8 @@ async function createIgnorerFromContextOrDie(context) {
   try {
     return await createIgnorer(
       context.argv.ignorePath,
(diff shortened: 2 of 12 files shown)


Standard8:
> Are `ignorePatterns` in ESLint compatible with Flat Config? It’s not very clear [from the docs](https://eslint.org/docs/latest/use/command-line-interface#--ignore-pattern). If that’s not the case, I wonder if we should step back and focus on specing a new config file instead, similar to Flat Config in ESLint. Looking at #7073 may be a good start, but this predates Flat Config, so might need further thinking.

For the command line option, I can't see any signs of it being removed.

In the configuration itself, they are moving from `ignorePatterns` in the current to config, to individual `ignores` sections in the Flat Config. It looks like the format of those are the same.

If we're concerned about the ignore pattern future, we could skip adding the command line options and only add an `ignores` option into the existing configuration. I did the command line option first as that happened to be the simpler case.

I think adding an `ignores` option to the existing configuration format wouldn't cause any significant migration issues if Prettier migrated later.

Standard8:
Abandoning, since this doesn't seem to be wanted and #14332 landed.

Standard8:
> Are [`ignorePatterns`](https://eslint.org/docs/latest/use/configure/ignore#ignorepatterns-in-config-files) / [`--ignore-pattern`](https://eslint.org/docs/latest/use/command-line-interface#--ignore-pattern) in ESLint compatible with Flat Config? It’s not very clear from the docs, at least after a quick scan. If that’s not the case, I wonder if we should step back and focus on specing a new config file instead, similar to Flat Config in ESLint. Looking at #7073 may be a good start, but this predates Flat Config, so might need further thinking.

PR #14280, opened 2023-01-31
Fix removal of parentheses for nested typescript unions

### LINKED ISSUE #14271: [Typescript]: Prettier removes the necessary parentheses in union and causes a syntax error
**Prettier 2.8.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAKtgXmwB0ptsAfbAClPPKuDvpYDcBDAGwFc4lso3ALYAjOACdmLAL5TK2JmRb0OPPtgDOMcQEsoAcznlZSgJSkQAGhAQMMHdA3JQ7ceIgB3AAquETlFwe7GhO1iLi7GAA1nAwAMrsQnAAMnpwyABmXBpwYRHRsXEYkXr6yNq81nCicAAmtXXJ7Abc7PpwAGIQ4kLsMPYGyCDs3DAQViAAFjBCnADqkzrwGsVgcHF+SzqsS2hDYBqhIHo54jBeEfq9mdm5IABWGgAecaWccACK3BDwN5w51mK4lOQ3QWA0YF0dgmGF0sDmOlqMEmyAAHAAGQHuHJzCIYIawuCnVjpawAR2+8Autn8ww0AFooHA6nUJuI4BSdOyLm1rkgsv87jkhDpyuJKiANG84ABBfq6ESjOBeCSpJl-AGS6VfH7pfm3awwdgiBFIlFIABMhoiOk4pQAwhAhHyQESAKwTbg5XDG-wCzWsXgASSgDVgcUhOjsMtDcXQ7w1cGk0iAA)

```sh
--parser typescript
```

**Input:**

```tsx
type T = 
  | (
    | {
        value: number
      }
    | {
        value: string
      }
  )

```

**Output:**

```tsx
type T =
  |
      | {
          value: number;
        }
      | {
          value: string;
        };

```

**Expected behavior:**
Expecting the parentheses to stay on their initial place or to remove them along with top-level union to get something like this:
```tsx
type T =
  | {
      value: number;
    }
  | {
      value: string;
    };

```

----
Btw, if union contains only one object the prettier is working as intended
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAKtgXmwB0ptsAfbAClPPKuDvpYDcBDAGwFc4lso3ALYAjOACdmLAL7MAlKRAAaEBAwwAltADOyUO3HiIAdwAKBhLpRdj7NLpUjx7MAGs4MAMrshcADIaUHDIAGZc2nCOzm4enhgugQDmyDDivCpwonAAJtk5fuxQidzsiXAAYhDiQuwwmkXIIOzcMBDKIAAWMEKcAOodGvDa8WBwnpaDGqyDaI1g2g4ggRHiMKbOiTWh4ZEgAFbaAB6eSZxwAIrcEPDbnBEq8eIrjehY2mDiGurtGJ+wvRpsjAOsgABwABgeRgivWcGEavzgK1YwRUAEcrvB1morE1tABaII5HLtcRwDEaMnrUpbJBhO67CJCDQpNKM05wACCdU+IhacFMEgCQVu9xA2g5l2uwTpOxUMHYIgBQJBSAATPLnBpOEkAMIQIS0kBIgCs7W4EVwiqs9LFrF4AEkoHlYJ4Pl8YJznZ50GdRXBpNIgA)

auvred:
## Description

Fixes #14271

Prettier incorrectly prints nested typescript unions. Here're all edge cases that I've found ([link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAdWBPADnABAOQEZcBedXXAH1wApyKrb6HHhmWGBrODJXAZxgAnAJZQA5gG52LAL7sAlPQXSo6GNjz4ATKXrU6UFtTZGOFbrwHCxUmbnlnUME-Ys8+UAK4BbAEZwQqrmjhQq6OqaBADMemYGrPTOlnyCohLBDrjhapg4BAAscRQJpbimFMke1ul2ZrLZ2aqR+fgArMWMhsZMZj3AuCk1tg72JoPVaSOhHONDUxKjfQxzkzaLM2Hsq1YL4kv9E7vr+zM5LVoAbKRd9AD0AFSED3e3y67L7se17GfoSlAQAAaEAQLAwETQfjIUAAQyEQggAHcAArwhDQlCwgA2SNhGGhIL8QlhYG4MAAyrCfHAADJiODIABmOP4cCJJLJcEpWFJtmQwi87JAcH8cAAJuKJbTYRIvLDxHAAGIQIQ+WEwCESZAgWFeGAQYEgAAWMB82IA6saRPB+LywHAKRibSIAG42jA6sD8QkgMRsoQwFEk8Tq5ms4UAK34AA8KbZsXAAIpeCDwcPYtkg3lCAM6jQ4fhgUTgo1YdIwC0icUwY3IAAcAAZs4i2RaSVgdeW4AHXYyQQBHVPwYNgzG6-gAWigcAlEqNQjgQ5Ei+DCrDSBZmeFbJ8IgFQiFIP4CbgAEFNaI-Pq4CjAvSZxmsyATxJEym04zNxGQTBYX4qxrOskG0X8SREbFbAAYQgHwNxFfg2iNLw2QAFX-TEt2fV0hQASSgKVYApYsRHBM8CIpDREyfOBZFkIA)). So I added all these cases to the tests

## Checklist


src/language-js/print/type-annotation.js
```diff
@@ -42,6 +42,7 @@ function shouldHugType(node) {
       (node) =>
         node.type === "ObjectTypeAnnotation" ||
         node.type === "TSTypeLiteral" ||
+        node.type === "TSUnionType" ||
```
fisker, line 45: Should we skip printing `|` here, 

https://github.com/prettier/prettier/blob/37fb53acf33a20379a93bfecea8242c03ce0a4fc/src/language-js/print/type-annotation.js#L123

when union type only have one `types`, instead of changing this?

fisker: As I tested, if I add

```js
  if (node.types.length === 1) {
    return print(["types", 0]);
  }
```

to begin of this function, they all format correctly(though there is a comment print issue to fix).

auvred: Yes, your way of `node.types` checking looks much better. And I'm reached this issue with comments too, it occurs after formatting already formatted code, right?

fisker: I'm not sure what happened, should be something wrong in `willPrintOwnComments`.

If we choose go this approach, the `parent` check part need a big update too. I suggest you test both `typescript` and `flow` parser, they should behave the same. They are not now, because the AST is different.

fisker: Since there is no single type wrapper in flow, we also need update `canAttachComment`, otherwise the comments will attch to different nodes.

fisker: Maybe strip these nodes with single type in `src/language-js/parse/postprocess/index.js` is a simplier way. 

Otherwise this seems be hard to fix, we also need update all places checking `TSUnionType`...

auvred: Hi @fisker! Sorry for mention, but I'm really stuck on this issue. As I can see the stable version also has some troubles with comments inconsistency after re-formatting (like [this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VACAZgSwE4DOM2EeAtgIYww5QDmAOrAJ4AOcGFGAvExhgDIMACj4YGMVACoMARgxTMYicDH9+AazjMkGAG4UANgFc4ajAF8xASiZN0GAnEhQAJiXJUa9JjDYcubgxpOQVUJlUoTW1dAxM4AG4mCwSQABoQCFYaaAJkUAo8PAgAdwAFQoQ8lCMSimY8jIAjPAowLRgAZQoyOAAZWjhkLCMnZtb2uC7WNto6ZBg8Uwy4Mia4V1cNvop6Ywo6OAAxUkpqOeQQCmMYCHSQAAsYMkMAdQeceAIZsDhOqs+OD0n2YlzABEaIFoTjwMDKrTolGGozgGQAVgQAB6dOaGOAARWMEHgyMMYxAM0IcDwlz87AIYDwOGy91YTNgrxwrhgD2QAA4AAwZNkQJyvVqsS5suAwvRDDIARyJ8HhWWqVwIAFooHANht7ng4Er8HB4QckUgRmTUSAnGQcAsljaCLi4ABBahMpo3U3UgY60nkl30PGE4lDS0ojIwChNTnc3lIABM0daOEMcwAwhAyBaQDKAKz3YxOAAqseqVvJelMAEk3AguozmTA3W5On48YG4BYLEA) and [this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VACAZgSwE4DOM2EeAtgIYww5QDmAOrAJ4AOcGFGAvBkxhlQAqPlAEBGfhiGYAPhmABrOMyQYAbhQA2AVzgBfDPKUq1m3QaMLlqjdr2HjNs-ctPTdi-qZN0GAnCQUAAmJORUNPRMMGwcXLzCohJSMlLGGM6eehjeYlbAGR7m2bkC6ZnFHKX5hbaVOWkKtS4WOQDcIAA0IBCsNNAEyKAUeHgQAO4ACiMIgyja4xTMg90ARngUYMowAMoUZHAAMrRwyFjaAWsbW3C7rJu0dMgweHrdcGSrcMHB34cU9B0FDocAAYqRKNRHsgQBQdDAIF0QAALGBkLQAdWROHgBHuYDgO1mOJw6hxzBhYAIKxAtACeBgkw2dEoZwucG6ACsCAAPHaPLRwACKOgg8DZWkuIHuhDgeBhMXYBDAeBwfSRrFVsAxOGCMGRyAAHAAGbqaiABDEbVgwzVwenqU7dACOovgTN6c1hBAAtFA4N9vki8HBXfg4EzgaykOdJRyQAEyDhnq94wQBXAAILUVWreERuXHf0SqXp+iCkVi04x9ndGAUVY6vUGpAAJjrGxwWkeAGEIGRoyB7QBWJE6AIAFQbc1jUvUegAkiEELsVWqYJmQjsYoKSwZ9EA)). Could you please advice me what to do to solve this problem?

fisker: Okay, let's pick the easy path first,

add

```js
      case "TSUnionType":
        if (node.types.length === 1) {
          return node.types[0]
        }
        break;
```

to

https://github.com/prettier/prettier/blob/f602be5a478a5bcdca870b31936dc873ecae1120/src/language-js/parse/postprocess/index.js#L63

the problem should be solved.

(I'm on cellphone, haven't test.)

src/language-js/parse/postprocess/index.js
```diff
@@ -173,6 +173,13 @@ function postprocess(ast, options) {
           });
         }
         break;
+      case "TSUnionType":
+        if (node.types.length === 1 && (node.types[0].type === "TSUnionType" || node.types[0].type === "TSIntersectionType")) {
+          node.types[0].loc = node.loc
+          node.types[0].range = node.range
+          return node.types[0]
+        }
```
fisker, line 181: This is not correct. But before you fix it. Please move tests to a new dir in current one, and add `flow` parser to `jsfmt.spec.js`, and add some comments in code, you'll see what I mean.

auvred: I've added `loc` and `range` expanding to prevent comments detaching

```tsx
// Input
type A6  = /*1*/ | C;
type D1 = /*1*/ | a & b;

// Output without loc and range inheritting
type A6 /*1*/ = C;
type D1 /*1*/ = a & b;

// Output with inheritting (as expected)
type A6 = /*1*/ C;
type D1 = /*1*/ a & b
```

How can I avoid detaching comments without inheritting `loc` and `range` from parent type?

fisker: I'm aware of that, but that's how `flow` works, [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAggNm2wF5sB6AKgEYKzsAfbAYQG4AdWTHAESpPOq0G2AIbYAZNgBGLEABoQEDDACW0AM7JQIgE46IAdwAKuhJpQiANgZFpNCqTpFgA1nBgBlEQFs4AGRUoOGQAMyt1OAcnV3cPDGdAgHNkGB0AV0iQOG8pOAATPPy-EShEtJFEuAAxCB1vERhVUuQQETSYCHkQAAsYb0sAdW6VeHV4sDgPMxGVADcRtBawdXsQQIidGCMnRPrQ8MyAK3UADw8kyzgARTSIeH3LCIV4nQ2WkMtDLowdQJgBlR5GDdZAADgADM99BEBk4MC0fnANrNggoAI63eDbJTmVrqAC0QXy+S6OjgGJUZO2FT2SDCj0yEW8KhS6UZFzguEavyk7TgRjgOgCQQeTxA6g5NzuwTpBwUMBEUgBQJBSAATPKnCpLEkmBBvLSsuoAKxdNIRAAqivM9LFswyAEkoIVYB4wL9lLhnR50JdRXAAL4BoA),

We'll fix that later. First make it print consistently with flow parser.

auvred: If I add 
```tsx
case "TSUnionType":
  if (node.types.length === 1) {
    return node.types[0];
  }
  break;
```
to the `src/language-js/parse/postprocess/index.js`, then `typescript` and `babel-ts` parsers incorrectly attach comments. They attach comments to the `TSTypeAliasDeclaration` in those cases:
https://github.com/prettier/prettier/blob/f602be5a478a5bcdca870b31936dc873ecae1120/tests/format/typescript/union/union-parens.ts#L91
https://github.com/prettier/prettier/blob/f602be5a478a5bcdca870b31936dc873ecae1120/tests/format/typescript/intersection/intersection-parens.ts#L38
So comments are placed before the `=` sign in type declarations

----

But if I'll bring union check back to the 
https://github.com/prettier/prettier/blob/f602be5a478a5bcdca870b31936dc873ecae1120/src/language-js/print/type-annotation.js#L94
all is fine, all tests pass. And it doesn't affect `flow` at all. (Btw `flow` prints all edge cases [correctly](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAggRmwF4AdKbbAH2wAoyKLq7yHHth7WuBrONJbAGcYAJwCWUAOYBuTl2wBfOdgCUnFbNiYcuAEzFOTZdQ4t5vfkNESZyikrNV2dhhYFQArgFsARnBGa8g4UGmToWHgAzAYsTE6mrG5W4lKBiqqa4ToALDFsNE4FJspJwim2LAqqGWHaeACseYXGtC4m2KXWUopt7B18AmU2PY5swP2WQ93BXO2d5SOsao5zA8nDM2MTg12SizVaEbgAbMTNLAD0AFT4VxfnrMWj2+upyjPLIZq1EQBChKRYq1HAAyYHyBLybAANwAhgAbDxwHblNKsTaqdSZOq-fSA-LKMHMeRGZ4UIkuViQqEMOGI5GvCo0+yUzHPT4MDkaEAAGhAEAwMDE0EEyFAsJEIggAHcAAoShCilAI6WwtCivk+ESwsC8GAAZVhXjgABkJHBkAAzBGCOCa7W6uAGjA6mzIURIvlwXxwAAmvr9JthUg8sMkcAAYhARF5YTAhVJkCBYR4YBBeSAABYwLzwgDqmbE8EELrAcH1iqLYmhRbQSbAgg1IAktpEMFl2sksatNrtIAAVoIAB76mzwuAARQ8EHgPfhtr5LpEraTlvhMozGBSMDzYl9MEzyAAHAAGRdS2157UYJNbuCt6EWvkAR2n8A7AqVycEAFooHA-T9DMRDgV8xBAjsw27JBrXnPtbS8MR3RET0QEEMc4FweNxB8VM4FlfwzX-OcFzQjCpxnC0YN7PkYFhHxd33Q8kF0WjtTEeEbAAYQgLxoJAe96gzDxbQAFXopVYNI6EkQASSgANYH1MBxEFXAFP1dBxxIuAFAUIA)) 

----

Maybe I need to perform a check in function `printIntersectionType`?
Because as I understand it if I'll strip redundant unions in `postprocess` without `loc` and `range` inheritance it'll require to change the processing of comments to handle this particular case

fisker: Since in flow parser we can't know where the "redundant" `|`  is, and it's not breaking syntax, only print comments incorrectly.
I think the goal in this PR should be make the output same as flow parser, even the comments print in unexpected position.
After we get it done, we 'll deal with the comment issue together (both `typescript` and `flow` parser). Can be in a follow-up PR.

auvred: Yea, sure, sounds great! I will commit fixes soon

auvred:
I've removed `loc` and `range` expanding, but I left `(node.types.length === 1 && (node.types[0].type === "TSUnionType" || node.types[0].type === "TSIntersectionType"))` unchanged to prevent this case when comment attaches to the nested object:
```tsx
// Input
type A6 = | (
  /*1*/ | (
    | {
      key: string;
    }
  )
  );

// Output
type A6 = {
  /*1*/ key: string,
};
```
Should I remove this condition too?

---

Also, as I mentioned, these test cases now work incorrectly due to comments being detached
https://github.com/prettier/prettier/blob/f602be5a478a5bcdca870b31936dc873ecae1120/tests/format/typescript/intersection/intersection-parens.ts#L38-L43
So I would appreciate some help with how best to fix this problem

src/language-js/parse/postprocess/index.js
```diff
@@ -81,17 +81,9 @@ function postprocess(ast, options) {
...
-      case "TSParenthesizedType": {
-        if (
-          !(
-            isTsKeywordType(node.typeAnnotation) ||
-            node.typeAnnotation.type === "TSThisType"
-          )
-        ) {
-          node.typeAnnotation.range = [locStart(node), locEnd(node)];
-        }
+      case "TSParenthesizedType":
```
fisker, line 84: I'm going to fix this issue first.

fisker:
Blocked by https://github.com/prettier/prettier/issues/14457

fisker:
@auvred Thanks for working on this, after changes in other places, and a few changes in flow-parser, we are able to print union/intersection type consistently between Flow and TS.

I've opened #14654 instead.

PR #14288, opened 2023-02-03
Refactor `printDanglingComments`

fisker:
## Description

- Combine `sameIndent` and `filter` into object
- Change `sameIndent` default to true
- Support `marker`

## Checklist

### DIFF
--- a/src/language-js/print/array.js
+++ b/src/language-js/print/array.js
@@ -33,7 +33,7 @@ function printEmptyArray(path, options, openBracket, closeBracket) {
   }
   return group([
     openBracket,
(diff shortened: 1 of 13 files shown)


sosukesuzuki (commented):
Just in case, `printDanglingComment` isn't exported as a public API right? (I concerned that this change will be a breaking change)

fisker:
> Just in case, `printDanglingComment` isn't exported as a public API right? (I concerned that this change will be a breaking change)

It's not public API.

sosukesuzuki (approved):
> It's not public API.

:+1:

PR #14332, opened 2023-02-10
Accept multiple `--ignore-path`s

### LINKED ISSUE #8048: Multiple files for --ignore-path
I use the [--ignore-path](https://prettier.io/docs/en/cli.html#--ignore-path) argument and point it to an exhaustive `.gitignore` file. It's great that I don't have to duplicate the list of ignored files.

However, there are a couple of files I'd like to exclude from prettifying but still want to commit. Is there any way to specify an additonal `.prettierignore` file so that I can add files without copying the entire `.gitignore`?

fisker:
## Description

Fixes #8048

## Checklist


src/utils/ignore.js
```diff
@@ -44,8 +44,31 @@ async function createIsIgnoredFunction(ignoreFilePath, withNodeModules) {
...
- * @param {{ignorePath?: string, withNodeModules?: boolean}} options
+ * @param {string[]} ignoreFilePaths
+ * @param {boolean?} withNodeModules
+ * @returns {Promise<(string) => boolean>}
+ */
+async function createIsIgnoredFunction(ignoreFilePaths, withNodeModules) {
+  // If `ignoreFilePaths` is empty, we still want `withNodeModules` to work
+  if (ignoreFilePaths.length === 0 && !withNodeModules) {
+    ignoreFilePaths = [undefined];
+  }
```
kachkaev, line 55: What if we first read all files in a loop, combine them and then call `createIgnore({ allowRelativePaths: true }).add(combinedContent);`? That's how I was imagining multiple ignore paths to work from the user perspective. This can simplify how `withNodeModules` works: it will be just a new line appended to `combinedContent`.

Example setup where a combined approach kinda matters: [🐸 njt / .prettierignore](https://github.com/kachkaev/njt/blob/main/.prettierignore)

I can imagine removing seconds named `Same as in .gitignore` and `Shared between linters` and add this to `.prettierrc.cjs`: `ignorePath: [".prettierignore", ".gitignore", ".sharedlintignore"]`

fisker: 1, they can be in different directories, they can't be simply concated. We need to know what the patterns related to.

```
#.ignore1
foo

#dir/.ignore2
bar
```

2. In this case

```
#.ignore1
foo/a.js

#.ignore2
!foo/*
foo/b.js
```

should `foo/a.js` be ignored? I think user may expected it be ignored, because it's ignored `.ignore1` file. But if we treat this case like

```
#.ignore
foo/a.js
!foo/*
foo/b.js
```

I'm not sure what's expected.

kachkaev: Oh I see... I forgot about ignore files sitting in different directories 🤔 You are right then, we need multiple calls to `createIgnore`

kachkaev (commented):
I guess we need to tweak type definitions too:

https://github.com/prettier/prettier/blob/e4349cae71ed5450825ed1121df04166c536ee9d/src/index.d.ts#L788
↓
```
ignorePath?: string | string[] | undefined;
```

Apart from that, LGTM! The change looks backwards-compatible, so it’d be nice to see it in v3.0 or 3.1!

If anyone else has time to take a look at the diff, please do 🙏

kachkaev (approved):

lostinpatterns:
@fisker any chance of cutting a new release now that this is merged?

kachkaev:
@lostinpatterns this change will be included into [v3.0](https://github.com/prettier/prettier/milestone/14)

erikpukinskis:
@lostinpatterns FYI this is in `prettier@next`, it's just not tagged on Github!

src/utils/ignore.js
y-nk: @fisker sorry to ask _so late_ ; it is not told in the docs if the multiple ignore paths would respect their own root folder. for example, how would the final ignore list look like for `prettier --ignore-path .ignore1,subfolder/.ignore2 --check subfolder/`? (usecase is: running prettier at root level of a monorepo)

Issue #13276, opened 2022-08-12
[Bug] File extension patterns not detected when plugin is required explicitly using require() (due to PnP)


Waghabond:
## Steps to Reproduce:
Steps provided in repo readme: [Prettier Plugin Explicit Require Bug demo repo](https://github.com/Waghabond/prettier-plugin-explicit-require-bug)

**Environments:**

- Prettier Version: 2.7.1
- Running Prettier via: CLI (Yarn PnP)
- Runtime: Node.js v16.13.2
- Operating System: WSL2 - Ubuntu 20.04
- Prettier plugins (if any): @prettier/plugin-php, @shufo/prettier-plugin-blade, prettier-plugin-sh (and probably all other plugins)

## Description:

When using PnP plugins are not being loaded as yet (see issue #8474 for that discussion) . A work around for this is using require() to add them explicitly in a .prettierrc.js (this mostly works well). 
However, plugins that are required this way dont seem to seem to have their file extension glob patterns detected by prettier, they're loaded and work correctly if the correct files are passed in eg. with a command like `yarn prettier -w *.php` .
Prettier won't automatically detect the files when it recursively traverses a directory eg. with a command like `yarn prettier -w .`  (.php files wont be detected with this).

  ## Why this problem arises and Solution:
After some digging, I found the reason why i'm facing sunch a weird issue: 
> Doing it this way, the glob pattern resolving stuff will not detect file extensions declared by the plugins. This is because the context object (which resolves the glob into filepaths) is loaded before the .prettierrc file (see #13276). As a result the plugins get loaded due to the configuration in the prettierrc but the files they're supposed to format won't get detected. 
>
> This makes it so that there is no option but to pass the plugins using the `--plugin` cli argument. 
> I think the only sensible way to handle the issue is what has been previously sugested in this thread: 
> - in the `cli.js` file we should introspect the `package.json` file the `.prettierrc` file for plugin names 
> - Then use `require()` to load them.  
>
> It's important that the plugin loading step is done in `cli.js` before the context is loaded because otherwise glob resolution doesn't work correctly
>
> _Originally posted by @Waghabond in https://github.com/prettier/prettier/issues/8474#issuecomment-1213386876_

A solution proposed in https://github.com/prettier/prettier/issues/8474 would fix this problem.

Prettier's `cli.js` should have an aditional step wherein it reads the `package.json` and `.prettierrc` to find any installed prettier modules and then it should try and require them using `require()`
[...]

PR #14348, opened 2023-02-12
feat: support file extentions of the plugins in the configuration file

zouguangxian:
## Checklist


zouguangxian:
This commit can fix issues:  #13276 and #8474.

fisker:
Thanks for working on this.

This problem is a known limitation https://github.com/prettier/prettier/pull/7660#discussion_r383781236 .

Your solution doesn't fix the case with this config

```json
{
	"override": [
		{
	      "files": "*.foo",
	      "options": {
	        "plugins": ["prettier-plugin-foo"]
	      }
		}
   ]
}
```

Before we find a good solution, I suggest you can use `prettier "**/*" --ignore-unknown`.

PR #14354, opened 2023-02-13
Bundle printers in graphql plugin

fisker:
## Description

Attampt to solve https://github.com/prettier/prettier/issues/13256#issuecomment-1208298030

## Checklist


sosukesuzuki (commented):
Overall looks good to me. One question, what about JavaScript parser and printer?

fisker:
Quote from https://github.com/prettier/prettier/issues/13256#issuecomment-1427560139

> For the js printers, we need add a new plugins/estree.mjs plugin, or leave it in core for now.

Let's merge this and decide what to do later.

sosukesuzuki (approved):

PR #14365, opened 2023-02-15
Add a section to explain where the parsers is

fisker:
## Checklist


sosukesuzuki (commented):
For JavaScript parsers, it would be good to have a description of what features each parser has. So that users know which parser they should use.

PR #14408, opened 2023-02-23
Include prettier-plugin-sql-cst to community plugins list

nene:
## Description

This PR adds link to [prettier-plugin-sql-cst](https://github.com/nene/prettier-plugin-sql-cst) to the list of community plugins.

Admittedly, this is an attempt to promote an SQL-formatting plugin that I have written. There already exists an SQL-plugin in this list, so why do I suggest adding another one?

The existing SQL plugin simply wraps an existing [sql-formatter](https://github.com/sql-formatter-org/sql-formatter) library, which doesn't really format the code in the way Prettier itself does. I know its limitations quite well, as I'm actually the author of that library.

The new plugin uses the actual Prettier formatting algorithm to lay out the code and is also much more in line with the philosophy of Prettier, by providing almost no configuration options and having a very opinionated approach to SQL code style.

## Checklist

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -72,6 +72,7 @@ To turn off plugin autoloading, use `--no-plugin-search` when using Prettier CLI
 - [`prettier-plugin-properties`](https://github.com/eemeli/prettier-plugin-properties) by [**@eemeli**](https://github.com/eemeli)
 - [`prettier-plugin-sh`](https://github.com/un-ts/prettier/tree/master/packages/sh) by [**@JounQin**](https://github.com/JounQin)
 - [`prettier-plugin-sql`](https://github.com/un-ts/prettier/tree/master/packages/sql) by [**@JounQin**](https://github.com/JounQin)
+- [`prettier-plugin-sql-cst`](https://github.com/nene/prettier-plugin-sql-cst) by [**@nene**](https://github.com/nene)
 - [`prettier-plugin-solidity`](https://github.com/prettier-solidity/prettier-plugin-solidity) by [**@mattiaerre**](https://github.com/mattiaerre)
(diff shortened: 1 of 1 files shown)


fisker (approved):
Looks great! Can you make it work on Prettier 3? We are close to release it.

[How to migrate my plugin to support Prettier v3?](https://github.com/prettier/prettier/wiki/How-to-migrate-my-plugin-to-support-Prettier-v3%3F)

fisker:
You may want copy change to `/website/versioned_docs/version-stable/plugins.md` as well, so the website will undate, otherwise, it will be updated until next release.

nene:
Thanks. I copied by change to `/website/versioned_docs/version-stable/plugins.md`

Thanks also for pointers about Prettier 3.0. I'll see if I can make it work.

nene:
Just a quick question about 3.0...

Would you expect plugin developers to support both 2.x and 3.x in their plugins, or would you expect the plugins to completely ditch 2.x support and only support 3.x?

fisker:
It's possible to make it work for v2 and v3. Depends on you want your user to use v2 or not.

nene:
Thanks. In my case I will probably just ditch 2.x support as I don't really have any users yet.

nene:
Looked into upgrading to 3.x, but unfortunately the [@types/prettier](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/prettier) doesn't yet support 3.x. As my plugin relies heavily on TypeScript, I'll have to either wait for the new type definitions or contribute my own to DefinitelyTyped before I can start adopting 3.x.

fisker:
Released [3.0.0-alpha.5](https://github.com/prettier/prettier/releases/tag/3.0.0-alpha.5) with type definitions.

PR #14435, opened 2023-03-02
Use `export default` in `.d.ts` files

fisker:
## Description

TS5 seems not allow use `export =`, [CI log](https://github.com/prettier/prettier/actions/runs/4311945857/jobs/7521900909#step:6:5)

I've tested locally, it works.

Also, these modules are prettier plugins, not parsers, so renamed `parser` -> `plugin`.

cc @sosukesuzuki 

## Checklist


sosukesuzuki (approved):
Thank you!

PR #14468, opened 2023-03-07
Perf: Skip `printComments` if printer don't have `printComment` function

fisker:
## Checklist


src/main/comments/print.js
```diff
@@ -203,7 +203,7 @@ function printComments(path, doc, options) {
   return label(
     // Propagate object labels so that the printing logic for ancestor nodes
     // could easily check them.
-    typeof doc.label === "object" && { commented: true, ...doc.label },
+    doc.label,
```
fisker, line 206: @thorn0 `commented` seems not used. Not sure why added in #13532. I think it's safe to use directly.

PR #14475, opened 2023-03-09
feat: add config option to place operators at the beginning of line

magdalipka:
## Description

Added config option to place operators at the beginning of line #3806 .

Defaut option is at the end of line for backward compatibility. User can place `operatorPosition: "beginning"` to place operators at the beginning of line.

## Checklist

### DIFF
--- a/src/language-js/print/binaryish.js
+++ b/src/language-js/print/binaryish.js
@@ -222,10 +222,11 @@ function printBinaryishExpressions(
 
   const shouldInline = shouldInlineLogicalExpression(node);
   const lineBeforeOperator =
-    (node.operator === "|>" ||
+    options.operatorPosition === "beginning" ||
+    ((node.operator === "|>" ||
       node.type === "NGPipeExpression" ||
       (node.operator === "|" && options.parser === "__vue_expression")) &&
-    !hasLeadingOwnLineComment(options.originalText, node.right);
+      !hasLeadingOwnLineComment(options.originalText, node.right));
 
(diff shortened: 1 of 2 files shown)


17LingShan:
hey, I am new here. I just use `prettier@3.0.1` which is the latest version on npm in 7/8/2023. but my code can't be format when I write the .prettierrc.js like this:
```js
module.exports = {
  operatorPosition: 'beginning',
  tabWidth: 2, 
  printWidth: 80, 
  useTabs: false, 
  semi: false, 
  arrowParens: 'avoid',
  singleQuote: true, 
  bracketSpacing: true,
  bracketSameLine: true,
  trailingComma: 'none', 
  operatorPosition: 'beginning',
  overrides: [
    {
      files: '*.json',
      options: {
        printWidth: 140
      }
    }
  ]
}

```

and my eslint has already set the content:
```js
module.exports = {
...
  rules: {
...
    'operator-linebreak': [2, 'before']
...
  },
...
}
```

my code still like this:
```js
     let a =
        12312323 +
        324213324 +
        12324324324 +
        312312312111111111111 +
        11111111111111111111 +
        11111111111111111111111111111 +
        1
```

`operatorPosition` is not work, is my mistake or a bug?

magdalipka:
@12Ling

> hey, I am new here. I just use `prettier@3.0.1` which is the latest version on npm in 7/8/2023.

This option has not been merged, you would need to use https://www.npmjs.com/package/@btmills/prettier.

PR #14506, opened 2023-03-14
Format TypeScript expressions in Vue template attributes

seiyab:
## Description
Fix https://github.com/prettier/prettier/issues/14432

## Checklist


tests/format/vue/ts-expression/filter.vue
```diff
@@ -0,0 +1,8 @@
+<script setup lang="ts"></script>
+
+<!-- vue filters are only allowed in v-bind and interpolation -->
+<template>
+  <div class="allowed">{{value | thisIsARealSuperLongFilterPipe("arg1", arg2 as unknown) | anotherPipeLongJustForFun | pipeTheThird}}</div>
+  <div class="allowed" v-bind:something='value | thisIsARealSuperLongFilterPipe("arg1", arg2 as unknown) | anotherPipeLongJustForFun | pipeTheThird'></div>
+  <div class="allowed" :class='value | thisIsARealSuperLongFilterPipe("arg1", arg2 as unknown) | anotherPipeLongJustForFun | pipeTheThird'></div>
+</template>
```
seiyab, line 8: https://github.com/prettier/prettier/issues/14432 still exists for directives like `v-if`.
But I cannot find a proper parser.
Currently, `__js_expression` is used for Vue directives.
For attribute bindings, I changed parser from `__vue_expression` to `__vue_ts_expression`.
Using `__vue_ts_expression` for directive causes false-positive for Vue filter.

As far as I know, possible approaches are following.
- Approach A: create a new parser `__ts_expression`
- Approach B: leave it until we drop support for Vue filter that is [removed from Vue v3.0](https://v3-migration.vuejs.org/breaking-changes/filters.html)
- Approach C: use `__vue_ts_expression` and somehow detect that `|` is filter or bitwise-or. (I don't know whether it's feasible or not.)

fisker: Approach A: create a new parser `__ts_expression`

Take this path.

fisker: Wait, we already have `__vue_ts_expression` it should work.

seiyab: Thanks, I tried to employ `__vue_ts_expression`. https://github.com/prettier/prettier/pull/14506/commits/e057f090e411e79a9cb5c63c586cc92758876205
It almost works correctly.
The problem I mind is that it considers `|` as [Vue filter](https://v2.vuejs.org/v2/guide/filters.html?redirect=true) rather than bitwise-or operator.
It places operator `|` start of lines rather than end of lines.
I would try to create a `__ts_expression` parser if you mind this too.

But I found that `__vue_ts_expression` is better enough than before.

fisker: We can add node.type check to distinguish, 

```diff
node.operator === "|" &&
+ node.type === "BinaryExpression" && 
        (options.parser === "__vue_expression" ||
          options.parser === "__vue_ts_expression")
```

fisker: If it's not enough. We can also add parent node check, since vue filter can only use on expression root.

seiyab: Thank you for your replies.
I'm sorry but I don't understand you :dizzy_face:

> We can add node.type check to distinguish

I think it's indistinguishable because the code we pass to parser can be completely identical between vue filter and bitwise-or.
I think this is the reason why current implementation uses `__js_expression` for vue-directive and uses `__vue_expression` for vue-bind.

> If it's not enough. We can also add parent node check, since vue filter can only use on expression root.

I didn't know it. I'll add distinction if you want. But it will be complex a bit because chained filter seems to be a nested binary expression.
However, the original problem of this thread https://github.com/prettier/prettier/pull/14506#discussion_r1136369217 won't be solved.

Feel free to correct me.

fisker: I think you are right, let's add `__ts_expression` then.

seiyab: Thanks so much, I missed TS union type 👍

seiyab:
it's ready for review 🚀

tests/format/vue/ts-expression/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -8,20 +43,129 @@ printWidth: 80
...
+    "
+  ></div>
+  <div
+    class="not-allowed"
+    v-if="
+      value |
+        thisIsARealSuperLongFilterPipe('arg1', arg2 as unknown) |
+        anotherPipeLongJustForFun |
+        pipeTheThird
+    "
```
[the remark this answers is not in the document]
seiyab: That's right. I'll change.
Maybe non-ts test also have same confusion.

fisker:
It seems we still missing some places, 

https://github.com/seiyab/prettier/blob/5870d8dd7fa159c175b805fff422f7d68cfd57ed/src/language-html/embed.js#L130-L134

This place seems using wrong parser? 

https://github.com/seiyab/prettier/blob/5870d8dd7fa159c175b805fff422f7d68cfd57ed/src/language-html/syntax-vue.js#L81

Maybe we need `babel-ts`?

fisker:
https://github.com/seiyab/prettier/blob/5870d8dd7fa159c175b805fff422f7d68cfd57ed/src/language-html/embed.js#L56

seiyab:
For the former in https://github.com/prettier/prettier/pull/14506#issuecomment-1477292445, I agree with you. I'll fix it.
https://sfc.vuejs.org/#eNpFkN1OwzAMhV/Fys02iTVD3JWugveIhLrijo7GiRK3XJS8O04zIFe2c87nn1W9el8tM6paNbEPo2eIyLOHqaPr2SiORrWGRutdYFgh4AAJhuAs7MS2M2RoQobezcQY4JwV+9Ph2VCjC1DskjBaP3WMkgE0l5nZEbz009h/ShuA/VsNhAuGA5xbWP95jyf4hqcEsA0C8op54+gSF+j7uNwVTfQdtesvJSWZJVeKp+ga/TeSelBlwaPtfHWLjuQcaxab+4dcoYatkmuyeM6N+mD2sdY6Dn0+4i1WLly1RFWQxqPFCqM9XoL7ihgEbFRGJENJpR+bfH+m

About the latter in https://github.com/prettier/prettier/pull/14506#issuecomment-1477292445 , I need more investigation.

I'll see https://github.com/prettier/prettier/pull/14506#issuecomment-1477293809 too.

seiyab:
I'm still investigating https://github.com/seiyab/prettier/blob/5870d8dd7fa159c175b805fff422f7d68cfd57ed/src/language-html/syntax-vue.js#L81 .

tests/format/vue/ts-expression/hug-string-literal.vue
```diff
@@ -0,0 +1,21 @@
...
+<template>
+<MyComponent :src="'loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooog literal string value'"/>
+</template>
+<template>
+<MyComponent :src="'first loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooog' + 'second loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooog'"/>
+</template>
+<template>
+<MyComponent :src="100000000000000000000000000000000000000000000000000000000000000000000000000"/>
+</template>
+<script lang="ts"></script>
```
fisker, line 21: you can put all attributes in one component, and please add final_new_line.

fisker: Ah, I see it's copied from tests/format/vue/vue/expression-binding.vue, can you put them together instead of move it to another dir. maybe `tests/format/vue/vue/expression-binding-ts.vue`?

fisker:
**Prettier pr-14506**
[Playground link](https://deploy-preview-14506--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzmATgSwA4wAEANgIZQDmAvADogzp0B8qA9JrgUzVD6vAFs8ZeNyiFCqAK7ExEiamI5CANwC0AMwhZaIUkihSBAIzhZCPeVevyIGwgA9CpdIUMmzzYMFIBfX2xKcpKsMmJsgsKkoiAANCAQBDjQ6MigpFhYEADuAAoZCKkopMTZpACeqfHGWKRgANZwMADKpAJwADI4UHDIGiXocNW1DU3NeHXdFMgwWFJDIHAeACbLcMsd5BRSpBRwAGLaAtEwU8h6UjAQcSAAFjACxADqtzjw6BNgcM2Fbzgqb3K5zA6CqIG6gywMFytQoxz6AwWACt0A5mlNiHAAIpSCDwBHEQbxCZYSHnFTzG54XCwJ44ZYwW7IAAcAAZiVlBk9anhztS4JCVL14gBHXHwGGJIp6dBqHrrdY3LBwMU4ZUw3bwpD9QkLQYCHAzOZ6jFwACCMFmOGMlzguTMXR6BKJIHQppxeN62sR8RgpGMdIZTKQACZfbUcEpKABhCACLV6YjEG5SQYAFX9RR1LopcAAklA1rBmth8DAzYXmjBypjnXB-EA)

```sh
--parser vue
```

**Input:**

```vue
<script lang="ts"></script>

<template>
  <ul>
    <li v-for="a:number 
               of x as number">{{a}}</li>
  </ul>
</template>
```

**Output:**

```vue
<script lang="ts"></script>

<template>
  <ul>
    <li
      v-for="a:number 
               of x as number"
    >
      {{ a }}
    </li>
  </ul>
</template>

```

seiyab:
Thanks, I'll also fix v-for.
v-for seems here
https://github.com/seiyab/prettier/blob/6959f30a81d277c7bc6c45b68991dac01dc14c65/src/language-html/syntax-vue.js#L14

I think I need to fix following as you said before.
https://github.com/seiyab/prettier/blob/5870d8dd7fa159c175b805fff422f7d68cfd57ed/src/language-html/syntax-vue.js#L81

However, v-slot with TS seems to work mysteriously. Maybe I'm misunderstanding... 🤔 
https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAzmATgSwA4wAE6cMArnoQDYCGUA5gLwA6IM6rAfM1DgLZ4IWIsEJY4AM0IBfQhKwQ+hAOQA3MnGU9+g4YUgC5CpcoB0AegDCivKfWaePSFHRE+6eoUZjJACmUAEnBUVBCEAOpCVAAmAITKAJQ8qOaYuATcUMnwArTwmYSEqAAWAIyc6KEwKWUFRdE4qnWoBnh1hag5eHlwhKoAtJUQMCwgAB5IhYVkUADWUBAA7lBTANoAulw8U1PAomOrpesy0ttTKV09zeat1w1NyeaXNPkgADQgEAQ40OjIoDQsApFgAFQEIP4oGhURY0ACefw+ACMsDQwLNSABlGh8OAAGRwUDgyAk0JIyNR6KxeDRhPoyBgWA0HzgfCRcGi0Q5eLo9DINHocAAYkI+C8YHTkCAaGQYBB3iBijA+FRwsUcPB0DSwHBMRCNY0NXCpWB0IiQISSMIQaj6GKSWS4B8AFboMaYulUOAARTIw2JSFJVHJIBpWCtUvsCrwuFg4Rw0RgxWQAA4AAwfGMQEjhVF4KUxuBW1TEj4ARz98BtX0h0vQ-SJHI5CvEFZw4htAvtgcdHxIfBwDKZTpA6E9cAAgjBGTgkbK4CC4FgCUSHcGR2OGF7ff61yGYDQkfHE8mkAAmD6Mmg4Kh06x8bsgIsAVgVZBIABVD5CgyH7ABJKAuVgTFsHwGAJyAzEYDhL09zgaRpCAA

seiyab:
Oops v-slot doesn't work on this PR [Playground](https://deploy-preview-14506--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzmATgSwA4wAE6cMArnoQDYCGUA5gLwA6IM6rAfM1DgLZ4IWIsEJY4AM0IBfQhKwQ+hAOQA3MnGU9+g4YUgC5CpcoB0AegDCivKfWaePSFHRE+6eoUZjJACmUAEnBUVBCEAOpCVAAmAITKAJQ8qOaYuATcUMnwArTwmYSEqAAWAIyc6KEwKWUFRdE4qnWoBnh1hag5eHlwhKoAtJUQMCwgAB5IhYVkUADWUBAA7lBTANoAulw8U1PAomOrpesy0ttTKV09zeat1w1NyeaXNPkgADQgEAQ40OjIoDQsApFgAFQEIP4oGhURY0ACefw+ACMsDQwLNSABlGh8OAAGRwUDgyAk0JIyNR6KxeDRhPoyBgWA0HzgfCRcGi0Q5eLo9DINHocAAYkI+C8YHTkCAaGQYBB3iBijA+FRwsUcPB0DSwHBMRCNY0NXCpWB0IiQISSMIQaj6GKSWS4B8AFboMaYulUOAARTIw2JSFJVHJIBpWCtUvsCrwuFg4Rw0RgxWQAA4AAwfGMQEjhVF4KUxuBW1TEj4ARz98BtX0h0vQ-SJHI5CvEFZw4htAvtgcdHxIfBwDKZTpA6E9cAAgjBGTgkbK4CC4FgCUSHcGR2OGF7ff61yGYDQkfHE8mkAAmD6Mmg4Kh06x8bvSkIKsgkAAqh8hQZD9gAklAuVgTFsHwGAJwAzEYDhL09zgaRpCAA).
I'll fix it.

seiyab:
v-slot with type doesn't work on next branch.
[Playground (next)](https://deploy-preview-9583--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAeAzmATgSwA4wAE6cMArnoQDYCGUA5gLwA6IM6rAfM1DgLZ4IWIsEJY4AM0IBfQhKwQ+hAOQA3MnGU9+g4YUgC5CpcoB0AegDCivKfWaePSFHRE+6eoUZjJACmUAEnBUVBCEAOpCVAAmAITKAJQ8qOaYuATcUMnwArTwmYSEqAAWAIyc6KEwKWUFRdE4qnWoBnh1hag5eHlwhKoAtJUQMCwgAB5IhYVkUADWUBAA7lBTANoAulw8U1PAomOrpesy0ttTKV09zeat1w1NyeaXNPkgADQgEAQ40OjIoDQsApFgAFQEIP4oGhURY0ACefw+ACMsDQwLNSABlGh8OAAGRwUDgyAk0JIyNR6KxeDRhPoyBgWA0HzgfCRcGi0Q5eLo9DINHocAAYkI+C8YHTkCAaGQYBB3iBijA+FRwsUcPB0DSwHBMRCNY0NXCpWB0IiQISSMIQaj6GKSWS4B8AFboMaYulUOAARTIw2JSFJVHJIBpWCtUvsCrwuFg4Rw0RgxWQAA4AAwfGMQEjhVF4KUxuBW1TEj4ARz98BtX0h0vQ-SJHI5CvEFZw4htAvtgcdHxIfBwDKZTpA6E9cAAgjBGTgkbK4CC4FgCUSHcGR2OGF7ff61yGYDQkfHE8mkAAmD6Mmg4Kh06x8bvSkIKsgkAAqh8hQZD9gAklAuVgTFsHwGAJwAzEYDhL09zgaRpCAA)

Stable is still mysterious.

src/language-html/embed.js
```diff
@@ -106,11 +106,15 @@ async function printEmbeddedAttributeValue(node, htmlTextToDoc, options) {
 
   if (options.parser === "vue") {
     if (node.fullName === "v-for") {
-      return printVueFor(getValue(), attributeTextToDoc);
+      return printVueFor(getValue(), attributeTextToDoc, {
+        ts: options.__should_parse_vue_template_with_ts,
+      });
```
fisker, line 111: Let's pass `options`, we may need access other props in future.

tests/format/vue/v-for/ts.vue
```diff
@@ -0,0 +1,10 @@
+<script lang="ts"></script>
+<template>
+<div
+v-for="a:number       of x as number[]"
+v-for="[ a  , b]   : [   string,string ]     of x as Array<     [string, string]>"
+v-for="  a  of list.map( (x:any): unknown     => x.foo.bar)"
+v-for="([longLongProp, longLongProp, [longLongProp, longLongProp='Hello, Prettier!', [longLongProp, longLongProp, anotherLongLongProp=[longLongProp, longLongProp, anotherLongLongProp, yetAnotherLongLongProp], yetAnotherLongLongProp], yetAnotherLongLongProp], yetAnotherLongLongProp], index) of longLongLongLongLongLongLongLongList"
+>
+</div>
+</template>
```
fisker, line 10: nit: add a new line at last.

fisker:
Looks great! Thank you for your hard work. I'll take another look tomorrow.

tests/format/vue/v-for/ts.vue
seiyab: Thanks, I found that I had disabled editorconfig plugin mistakenly 😅

fisker (approved):

fisker:
Nice work! Thanks.

You may interest #14575

seiyab:
Thank so much for your friendly supports.
I'll take a look after https://github.com/prettier/prettier/pull/14476 .

Issue #14514, opened 2023-03-15
Support `max_line_length=off` when parsing `.editorconfig`?


josephfrazier:
**TL;DR: Should `max_line_length=off` in `.editorconfig` get interpreted as `printWidth: Number.MAX_SAFE_INTEGER`, or should we continue ignoring it, or something else?**

Hey everyone, I'm the maintainer of [editorconfig-to-prettier](https://www.npmjs.com/package/editorconfig-to-prettier), which [Prettier uses](https://github.com/prettier/prettier/blob/5446a8f32a2717762c4d7a1bd5fe2615d76ebec7/src/config/resolve-config-editorconfig.js#L6) to convert [EditorConfig](https://editorconfig.org/) configuration to Prettier configuration.

Over 5 years ago, bug https://github.com/prettier/prettier/issues/3410 was reported regarding the handling of `max_line_length=off`, and resulted in @duailibe [fixing `editorconfig-to-prettier` to ignore `max_line_length=off` in `.editorconfig` files](https://github.com/josephfrazier/editorconfig-to-prettier/pull/3) (the corresponding dependency version bump can be found [here](https://github.com/prettier/prettier/pull/3412)).

However, there's been some [renewed interest](https://github.com/josephfrazier/editorconfig-to-prettier/issues/7#issuecomment-1468787319) from @celluj34 in having Prettier interpret `max_line_length=off` to mean e.g. `printWidth: A_REALLY_BIG_NUMBER` like [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER).

My question is, would this cause unintended consequences? For example, would Prettier attempt to generate very long lines for files where `max_line_length=off`? I see that https://prettier.io/docs/en/options.html#print-width says:

> In other words, don’t try to use printWidth as if it was ESLint’s [max-len](https://eslint.org/docs/rules/max-len) – they’re not the same. max-len just says what the maximum allowed line length is, but not what the generally preferred length is – which is what printWidth specifies.

[...]

PR #14516, opened 2023-03-16
Support `max_line_length=off` when parsing `.editorconfig`

josephfrazier:
## Description

This fixes https://github.com/prettier/prettier/issues/14514 by upgrading `editorconfig-to-prettier` with the fix from https://github.com/josephfrazier/editorconfig-to-prettier/pull/10

## Checklist

### DIFF
--- a/changelog_unreleased/misc/14516.md
+++ b/changelog_unreleased/misc/14516.md
@@ -1,0 +1,21 @@
+#### Support `max_line_length=off` when parsing `.editorconfig` (#14516 by @josephfrazier)
+
+If an .editorconfig file is in your project and it sets `max_line_length=off` for the file you're formatting,
+it will be interpreted as a `printWidth` of `Infinity` rather than being ignored
+(which previously resulted in the default `printWidth` of 80 being applied, if not overridden by Prettier-specific configuration).
+
+<!-- prettier-ignore -->
+```html
+<!-- Input -->
+<div className='HelloWorld' title={`You are visitor number ${ num }`} onMouseOver={onMouseOver}/>
+
+<!-- Prettier stable -->
+<div
+  className="HelloWorld"
+  title={`You are visitor number ${num}`}
+  onMouseOver={onMouseOver}
+/>;
+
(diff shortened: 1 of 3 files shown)


josephfrazier:
Thanks for the approval, @fisker! Just to confirm, you're ok with this being merged despite not having tests, documentation, or changelog items? I can try to add those if needed, wanted to make sure.

kachkaev:
It’d be nice to have a quick changelog item I guess. Do we want to merge the PR into `main` and then port it into `next` or is it better to backport the change?

josephfrazier:
> It’d be nice to have a quick changelog item

Sure, I added one in https://github.com/prettier/prettier/pull/14516/files?short_path=785b7f6#diff-785b7f6797d53004d9169af9a6c1c8aefa95a2512665c7032c8bc51bd5c46c3f, let me know what you think.

> because we have a mini new feature here (or even a breaking change?). Do we want to merge the PR into `main` and then port it into `next` or is it better to backport the change?

Thanks! Yeah... I was wondering if it should be considered breaking. It does require that the user specify `max_line_length=off` in their `.editorconfig`, which to me would imply that they _desire_ the change this PR makes, so maybe it isn't breaking? I'll defer to y'all, of course, as I'm out of practice with contributing here.

josephfrazier:
> I don't think it will break anything since api support `printWith: Infinity` too. But more tests always good, you can try add some.

Ah, I see the tests with `printWidth: Infinity` now, and think there's no need to add more.

> You already added changelog, which is good. `max_line_length=off` is rarely used, documentation update seems unnecessary to me, but I'm fine if you want mention it.

Similarly, t[he documentation already mentions the general support for `max_line_length`](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options), I don't see a need to include the `off` case there.
 
> I trust your judgement, we may not even release v2 anymore.

Oh, are there no plans to continue further v2 releases? I don't think this should be considered a breaking change, so if it could get released in `2.8.5` or `v2.9.0`, that would be nice.

PR #14573, opened 2023-03-23
Support conditional type and infer type in Flow

SamChou19815:
## Description

This PR adds support for printing conditional types and the companion infer type using `flow-parser`. It uses the existing printTernary helper function that's already implemented for TS. Since Flow's parser produces AST with the same name, it automatically works. For infer type, we also reuse the existing printTypeParameter function.

For testing, I copy the conditional type test suite from TS, and deleted some syntax Flow doesn't support yet (type assertion), and run both Flow and TS prettier on it and ensure they are consistent. I moved the comment one into another suite, because the results are different.

## Checklist

- [x] (N/A) (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).


tests/format/flow/conditional-types/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,455 @@
...
+// @flow
+
+// #13275
+type Foo<T> = T extends ((...a: any[]) => infer R extends string) ? R : never;
+
+=====================================output=====================================
+// @flow
+
+// #13275
+type Foo<T> = T extends (...a: any[]) => infer R: string ? R : never;
```
SamChou19815, line 452: TODO: I need to fix this issue of reprinting the bound syntax. Flow recently added support for the TS syntax of extends, so I need to extend the support to prettier as well.

src/language-js/print/flow.js
```diff
@@ -98,6 +99,10 @@ function printFlow(path, options, print) {
       return printIntersectionType(path, options, print);
     case "UnionTypeAnnotation":
       return printUnionType(path, options, print);
+    case "ConditionalType":
+      return printTernary(path, options, print);
+    case "InferType":
+      return ["infer", " ", print("typeParameter")];
```
SamChou19815, line 105: The naming inconsistency with other annotations isn't ideal. Flow will soon release a patch version that fixes this issue, and I will update the PR after that.

src/language-js/print/flow.js
```diff
@@ -98,6 +99,10 @@ function printFlow(path, options, print) {
       return printIntersectionType(path, options, print);
     case "UnionTypeAnnotation":
       return printUnionType(path, options, print);
+    case "ConditionalType":
+      return printTernary(path, options, print);
+    case "InferType":
```
fisker, line 104: Though it's one line, but please share logic with `TSInferType` extract `printInferType`.

src/language-js/print/flow.js
```diff
@@ -98,6 +99,10 @@ function printFlow(path, options, print) {
       return printIntersectionType(path, options, print);
     case "UnionTypeAnnotation":
       return printUnionType(path, options, print);
+    case "ConditionalType":
+      return printTernary(path, options, print);
```
fisker, line 103: Update comment in printTernary

src/language-js/print/type-parameters.js
```diff
@@ -152,7 +152,11 @@ function printTypeParameter(path, options, print) {
   parts.push(name);
 
   if (node.bound) {
-    parts.push(printTypeAnnotationProperty(path, print, "bound"));
+    if (node.usesExtendsBound) {
+      parts.push(" extends ", print(["bound", "typeAnnotation"]));
```
fisker, line 156: `node.bound` is skipped, better not to do that, there could be comments, cursors.

This won't work?

```js
parts.push(" extends ", print("bound"));
```

fisker: Ah, I understand the problem, we should still use `printTypeAnnotationProperty`. I'll fix it and add comments to explain why it's needed and how it works on monday.

sosukesuzuki (commented):
LGTM except where Fisker reviewed

src/language-js/print/type-parameters.js
fisker: Found a case

**Prettier pr-14573**
[Playground link](https://deploy-preview-14573--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAgtgXmwAoBKQgPmwEsoAzOAJ2wCVs4APeKAEwGcA9ACpskALZiEMbEIHY+MRrQDmAHSggANCAgYY1aH2SgAho0YQA7gAUzCIyhMAbSybRHtAI0YmwAazgYAGUTSQAZWjhkOmc+OC8ff0CgjF8VZEUAV3iQODFPOB4eQrCTKGVMk2U4ADEIRjETGH1y5BATTJgILRAACxgxJwB1Xup4PlSwOCD7MeoANzG0NrA+DxBaOMYYax9lRujYnIArPg4glSc4AEVMiHhDpzjtVMYttronKx6MJVgh6g8GC9ZAADgADC8LHEhj4MG1fnAtvMotoAI53eC7XQOdp8AC0UDghUKPUYcAx1HJuyqByQMSeOTiYmoGUY2W0fEucFwzSUnk6cGsTAiRMezxAXPKV1u9yi9KO2hgJk8AKBIKQACYlT5qE4VABhCASExtZxOHqZOIAFRVDgZEvm2QAkrwpEEwEo9LheEF0FdxXAAL5BoA)

```sh
--parser flow
```

**Input:**

```jsx
type A = () => infer R extends/* comment */ string

```

**Output:**

```jsx
Error: Comment "comment" was not printed. Please report this error!
    at dn (https://deploy-preview-14573--prettier.netlify.app/lib/standalone.js:29:5817)
    at Ze (https://deploy-preview-14573--prettier.netlify.app/lib/standalone.js:29:7308)
    at async kn (https://deploy-preview-14573--prettier.netlify.app/lib/standalone.js:29:11045)
    at async ju (https://deploy-preview-14573--prettier.netlify.app/lib/standalone.js:34:581)
    at async formatCode (https://deploy-preview-14573--prettier.netlify.app/worker.js:159:12)
    at async handleFormatMessage (https://deploy-preview-14573--prettier.netlify.app/worker.js:83:24)
    at async self.onmessage (https://deploy-preview-14573--prettier.netlify.app/worker.js:38:14)
```

src/language-js/needs-parens.js
```diff
@@ -604,6 +607,20 @@ function needsParens(path, options) {
...
+        key === "extendsType" &&
+        parent.type === "ConditionalTypeAnnotation" &&
+        node.type === "ConditionalTypeAnnotation"
+      ) {
+        return true;
+      }
+
+      if (key === "checkType" && parent.type === "ConditionalTypeAnnotation") {
+        return true;
+      }
```
fisker, line 621: Loos like we are missing cases, it's bad we are not sharing logic between TS and Flow.
Maybe better extract some code first to reuse?

SamChou19815: I fixed the logic and pass the test. I'm not sure how to better extract the code, since there are very little code sharing currently in the file.

My personal opinion is that the existing fallthrough logic for TS nodes already made it harder to understand, and I actually like the more explicit logic among those Flow nodes that don't use fallthrough...

fisker (approved):

PR #14619, opened 2023-03-28
Support Flow Mapped Types

jbrown215:
## Description

This PR adds support for Flow's Mapped Types. Flow's Mapped Types are extremely similar to TS's, though we do not support `as` syntax renaming nor read-only-ness removal (`-readonly`). Flow also allows `keyof` to be used in mapped types, thought it does not support keyof as a standalone type yet. Notably, Flow models a Mapped Type as an object property and we may choose to allow mapped types to support additional properties alongside them in the future while TS models a Mapped Type as its own type node.

Due to the significant modeling differences and other more trivial differences in our AST representation, the opportunity for code re-use here seemed small to me. I opted to not try to merge our different approached into one.

I added several tests and included relevant TS tests with some syntax translated. Please let me know if any other cases need to be exercised.

## Checklist


fisker:
Haven't look at the AST, just want point out the AST shape going to be changed in TS. Hope we have similar structure. https://github.com/typescript-eslint/typescript-eslint/issues/6433

jbrown215:
Note also that we probably will not converge with TS on the Plus/MinusToken stuff. It may make sense for the optional field, but Flow has a writeonly concept in addition to readonly, so Plus/MinusToken isn't sufficient to model the variance for mapped types in Flow. We don't support that syntax yet, though.

src/language-js/print/mapped-type.js
```diff
@@ -0,0 +1,103 @@
...
+    case "MinusOptional":
+      return "-?";
+    case "Optional":
+      return "?";
+  }
+}
+
+function printFlowMappedTypeProperty(path, options, print) {
+  const { node } = path;
+  const shouldBreak = hasNewlineInRange(
```
fisker, line 31: Looking at the AST, we use `ObjectTypeAnnotation` and this is a property of it, so the new line check should already handled in `printObject.`

The TS version doing this check because it includes `{}`.

fisker: So we are formatting the following differently.

```jsx
type Mapped = {[key 
               in keyof O]:number};
```

fisker:
Does Flow Mapped type allow multiple properties? 

Like this

```js
type Mapped = {
  [key in keyof O]: number
  another: another
};
```

We already format `ObjectType` differently between Flow and TS. Flow use `,` to separate, TS uses `;`. If Mapped type is a special case, maybe we should use `;` to align with TS.

jbrown215:
> Does Flow Mapped type allow multiple properties?

Yes, though the type checking support for additional properties is blocked on other intermediate steps right now. We do plan to support additional properties in the near future.

Will address feedback shortly. Thanks for the quick review!

jbrown215:
TS-compatibility tests checks that the syntax supported by both parsers formats the same on a single line. I didn't include multi-line tests because of the line break difference you mentioned, where TS uses `;` and flow uses `,`

jbrown215:
@fisker: Heads up that Flow and TS decide to add line breaks differently for objects. The example you provided adds a line break in TS but not in Flow:

```
type Mapped = {[key 
               in keyof O]:number};
```

fisker:
> @fisker: Heads up that Flow and TS decide to add line breaks differently for objects. The example you provided adds a line break in TS but not in Flow:
> 
> ```
> type Mapped = {[key 
>                in keyof O]:number};
> ```

Should be bug. It checks the whole expression, but should mean to check new lines after `{`, like `printObject`.

```js
  const shouldBreak = hasNewlineInRange(
    options.originalText,
    locStart(node),
    locEnd(node)
  );
```

fisker (approved):

PR #14627, opened 2023-03-29
Don’t expand globs via symbolic links

### LINKED ISSUE #12080: endless recursion when dir has symlink which points to its parent dir
**Environments:**

- Prettier Version: 2.5.1
- Running Prettier via: CLI
- Runtime: Node.js v16.13.0
- Operating System: Linux
- Prettier plugins (if any): none

**Steps to reproduce:**

First create the test dirs.
```bash
mkdir -p ./x1 ./y1
ln -sf .. ./x1/x2
ln -sf .. ./y1/y2
```

It should look like:

```bash
$ tree
.
├── x1
│   └── x2 -> ..
└── y1
    └── y2 -> ..

4 directories, 0 files

```

Then run prettier cli
```
prettier -c .
```

Try adding these dirs to `.prettierignore` and run again:
```
$ cat .prettierignore 
x1
y1

```

**Expected behavior:**

No error.

Or these problematic dirs can be excluded by `.prettierignore` file.

**Actual behavior:**

The command will run for a long time, and then produce an error.

```
[error] Unable to expand directory: .
[error] ELOOP: too many symbolic links encountered, scandir '/tmp/t/prettier-bug/x1/x2/x1/x2/x1/x2/x1/x2/x1/x2/y1/y2/y1/y2/x1/x2/y1/y2/x1/x2/y1/y2/y1/y2/x1/x2/x1/x2/y1/y2/y1/y2/x1/x2/y1/y2/x1/x2/x1/x2/x1/x2'
```

andersk:
## Description

Users don’t expect Prettier to follow symlinks:

- If a source tree contains a symlink an external location outside the source tree, Prettier should stay within the tree it was asked to format, and shouldn’t check or modify those external files.

- If a source tree contains a symlink to an internal location within itself, Prettier will already find that location via its normal path; symlinks should not cause Prettier to format those files multiple times.

- If a source tree contains a symlink to a location that’s ignored via `.prettierignore`, Prettier should ignore that location without the user needing to separately list the symlinks as ignored too.

Following symlinks without special care can lead to problems:

- Cycles of symlinks would cause Prettier to recurse infinitely or for a very long time, depending on the filesystem.

- A symlink to (say) your home directory could cause Prettier to rewrite every file in your home directory.

Fixes #12080.

## Checklist


tests/integration/__tests__/patterns-dirs.js
```diff
@@ -151,27 +151,47 @@ if (path.sep === "/") {
...
-    afterAll(() => {
-      fs.unlinkSync(path.resolve(base, "test-a", "one.js"));
-      fs.rmdirSync(path.resolve(base, "test-a"));
-      fs.unlinkSync(path.resolve(base, "test-b", "two.js"));
-      fs.unlinkSync(path.resolve(base, "test-b", "symlink"));
-      fs.rmdirSync(path.resolve(base, "test-b"));
-    });
-
-    testPatterns("", ["test-b"], { stdout: "test-b/two.js\n" });
+describe("Ignore symlinks", () => {
```
andersk, line 156: @fisker We need to skip this test on Windows since Windows doesn’t support symlinks by default, right? That’s why I had put this in the `if (path.sep === "/")` block (but we could also check `process.platform !== "win32"` or something).

fisker:
@andersk Thanks for working on this, we have more than glob patterns, so please fix the new tests I added.

fisker:
Maybe just skip if `stat.isSymbolicLink()` in this `if`

https://github.com/prettier/prettier/blob/053bc69c764d785fbdbe60497b9591f294394ecd/src/cli/expand-patterns.js#L69

tests/integration/__tests__/patterns-dirs.js
andersk: It seems that Windows users can create symlinks [when Developer Mode is enabled](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/).

Maybe we can safely assume that anyone running the Prettier test suite has that enabled?

andersk:
If Prettier throws an error when the user provides an argument that matches only non-supported files, it seems consistent that Prettier should also throw an error when the user provides an argument that matches only symlinks. I don’t think it should silently skip them as a special case.

tests/integration/__tests__/patterns-dirs.js
fisker: We can do a feature test then. Like

```js
function isSymbolicLinkSupported() {
  if (path.sep !== "/") {
    return true;
  }

  try {
    fs.symlinkSync(file);
    fs.statSync(file);
    return fs.isSymbolicLink();
  } catch {}

  return false;
}
```

fisker:
> it seems consistent that Prettier should also throw an error when the user provides an argument that matches only symlinks

We already did that? https://github.com/prettier/prettier/blob/053bc69c764d785fbdbe60497b9591f294394ecd/src/cli/expand-patterns.js#L36

---

Maybe we should discuss what's expected first.

1. Patterns only

```
prettier a/* b/*
```

Skip symlink, if non-match, throws

2. argument is symlink

```
prettier symlink/to/file
```

```
prettier symlink/to/directory
```

throws since no "file" matched.

3. pattern + symlink

```
prettier a/* symlink/to/file symlink/to/directory
```

Skip symlink, if non-match, throws


Sounds good? So, if `stat.isSymbolicLink() === true`, we just skip it should already work as expected.

andersk:
Okay, updated with an specific error for that case (`Explicitly specified symbolic link was ignored` rather than the generic `No matching files`).

tests/integration/__tests__/patterns-dirs.js
```diff
@@ -224,11 +224,61 @@ describe("Ignore symlinks", () => {
...
+    "",
+    ["test-a/symlink-*"],
+    {
+      status: 2,
+      stdout: "",
+      stderr:
+        '[error] No files matching the pattern were found: "test-a/symlink-*".\n',
+    },
+    base
+  );
```
fisker, line 281: What's the result of `prettier test-a/symlink-to-directory-b/*`, let's test it too. (Not really care about the result)

andersk: Added. That actually succeeds, which is perhaps surprising, but `fast-glob` [documents](https://www.npmjs.com/package/fast-glob#followsymboliclinks) that the `followSymbolicLinks` option does not affect the base directory of the pattern.

tests/integration/__tests__/patterns-dirs.js
andersk: Added a feature test.

tests/integration/__tests__/patterns-dirs.js
```diff
@@ -153,7 +153,24 @@ if (path.sep === "/") {
...
 }
 
-describe("Ignore symlinks", () => {
+function isSymlinkSupported() {
+  if (process.platform !== "win32") {
+    return true;
+  }
+
+  const target = path.join(__dirname, "../cli/patterns-symlinks");
+  const symlink = path.join(__dirname, "../cli/patterns-symlinks/symlink");
```
fisker, line 162: Let's link to a file inside to test, prevent accidentally removes the parent dir.

andersk: `unlinkSync` can never delete a directory, but sure, changed.

src/cli/expand-patterns.js
```diff
@@ -65,9 +66,13 @@ async function* expandPatternsInternal(context) {
       continue;
     }
 
-    const stat = await statSafe(absolutePath);
+    const stat = await lstatSafe(absolutePath);
     if (stat) {
-      if (stat.isFile()) {
+      if (stat.isSymbolicLink()) {
+        yield {
+          error: `Explicitly specified symbolic link was ignored: ${pattern}`,
```
fisker, line 73: "was ignored" seems not right, since we exited. 


`Explicitly specified ${pattern} is a symbolic link.` ?

andersk: Fixed.

src/cli/expand-patterns.js
```diff
@@ -65,9 +66,13 @@ async function* expandPatternsInternal(context) {
...
     }
 
-    const stat = await statSafe(absolutePath);
+    const stat = await lstatSafe(absolutePath);
     if (stat) {
-      if (stat.isFile()) {
+      if (stat.isSymbolicLink()) {
+        yield {
+          error: `Explicitly specified ${pattern} is a symbolic link.`,
+        };
```
fisker, line 74: It seems we will still expand other patterns, should we just throw?

tests/integration/__tests__/patterns-dirs.js
```diff
@@ -153,6 +153,160 @@ if (path.sep === "/") {
...
+    fs.symlinkSync(target, symlink);
+  } catch {
+    return false;
+  } finally {
+    fs.unlinkSync(target);
+  }
+  const stat = fs.lstatSync(symlink);
+  fs.unlinkSync(symlink);
+  return stat.isSymbolicLink();
+}
```
fisker, line 174: Let's use a subdir to test. 
Since `test-*` already in `.gitignore`, we don't need clean. And if something goes wrong, them won't get committed either,

```suggestion
function isSymlinkSupported() {
  if (process.platform !== "win32") {
    return true;
  }

  const target = path.join(__dirname, "../cli/patterns-symlinks/test-symlink-feature-detect");
  fs.rmSync(target, { force: true, recursive: true });
  fs.mkdirSync(target);
  const symlink = path.join(target, "symlink");
  try {
    fs.symlinkSync(target, symlink);
  } catch {
    return false;
  }
  return fs.lstatSync(symlink).isSymbolicLink();
}
```

src/cli/expand-patterns.js
andersk: Dunno. We still expand other patterns in similar cases already:

```console
$ echo x > real.js
$ prettier -l nonexistent.js 'r*.js'
[error] No files matching the pattern were found: "nonexistent.js".
real.js
```

tests/integration/__tests__/patterns-dirs.js
andersk: Changed.

src/cli/expand-patterns.js
fisker: That's because we don't exit if file not exits.  and we don't know if we can glob any file.

fisker: Sorry, you are right..

fisker (approved):
Good job.

fisker:
@sosukesuzuki @kachkaev Can you help review?

sosukesuzuki (approved):
LGTM!

sanmai-NL:
@andersk You open with a fine problem statement, but ultimately the solution design is very crude: report an error and exit with a fault exit status whenever a symbolic link happens to be specified on the command line. In CI, it's not always to possible to filter the list of files beforehand based on some special characteristics. Instead, Prettier should handle the scenarios you describe itself.

sanmai-NL:
> If Prettier throws an error when the user provides an argument that matches only non-supported files, it seems consistent that Prettier should also throw an error when the user provides an argument that matches only symlinks. I don’t think it should silently skip them as a special case.

It does not follow that Prettier should skip uncontrollably. Right now, Prettier cannot be controlled in the way it handles symlinks. It now simply fails and fails CI pipelines.

andersk:
@sanmai-NL Neither `prettier .` nor `prettier "*.js"` will error in the presence of skipped symlinks (unless there are *only* symlinks). Is it possible you accidentally wrote `prettier *.js`, where the unquoted wildcard is expanded by the shell before Prettier even sees it?

sanmai-NL:
I didn't specify globs (or rather our tooling), but literal paths.

I note that Prettier does have CLI flags to ignore unmatched paths, etc., but nothing to ignore symlinks.

Whenever something changes to a matched explicit path that is a symlink (e.g., file type changed to symlink), some CI tooling will specify it to Prettier and the fault will trigger.

sanmai-NL:
@andersk What can you do to fix this situation? We have pipelines failing on this with no recourse ...

Otherwise, I'll have to submit PR that restores the original behavior.

andersk:
@sanmai-NL I can’t do anything—I’m not a Prettier maintainer.

But if I understand the linked issue correctly, it looks like your use case will be addressed by evilmartians/lefthook#538, so you might try helping there.

If you don’t want to wait for that, you can use the lefthook [`exclude` option](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#exclude) to avoid passing unsupported files.

sanmai-NL:
@andersk patching other tools to work-around a breakage you introduced (unwittingly) is at best a stop-gap measure. You could've taken on to fix your work in a new PR. At least one compromise solution has been proposed in this thread.

I'll create a PR to fix your PR. Or perhaps revert it, if that takes undue time.

andersk:
@sanmai-NL The change was not made unwittingly. It is consistent with the way Prettier handles other unsupported files, and was discussed with the maintainer [ahove](https://github.com/prettier/prettier/pull/14627#issuecomment-1489597558).

I’m sorry that your CI had been unintentionally relying on the previous behavior, and I understand you’re unhappy that it fails after the Prettier 3 upgrade, but such things happen—major releases sometimes make breaking changes, and the purpose of CI is to fail when things break. Although there is always some value in avoiding breaking changes, here I still believe there was more value gained by eliminating unexpected behavior around symlinks, and throwing an error is better than either silently processing unexpected files or silently skipping them.

fisker:
@sanmai-NL I'm sorry that this change breaks your tool, but please open an new issue to continue discussion. This is an merged PR.

Issue #14290, opened 2023-02-03
Brackets around ternary as twice-curried arrow function body are removed


Josh-Cena:
**Prettier 2.8.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBNAFOglAAgLwB8OAhjgPw4BGOSOYWA3ADpQbb5HuEnlU12MWIADQgIABxgBLaAGdkoYgCclEAO4AFZQnkpiAGzXEAnvNGUlxMAGs4MAMrEAtnAAyUqHGQAzA7LjmljZ29uJWHgDmyDBKAK4BIHBOlHAAJqlprsRQEbHEEXAAYhBKTsQw0jnIIMSxMBAiIAAWME76AOpNUvCyYWBw9jrdUgBu3cbVYLJmIB7+SjAalhFlPn4JAFayAB72kfpwAIqxEPBr+v6iYUrz1ZTEKfqN4koeMO1SqTBNyAAcAAxXVT+dqWcTVF5weYjLyiACOJ3gSwkuhqsgAtJ40mlGko4AipHilvlVkhfBcEv4nFJonFKfs4ABBCqvSh1OAaOBKdyec6XECyBnHU5eMnrUQwB4fL4-JAAJgllik+kiAGEIE5SYlZABWRqxfwAFQeunJ-JG8QAklAMrB7GBXpJGTb7DBjAc+XAAL5eoA)

```sh
--parser babel
```

**Input:**

```jsx
Y(() => a ? b : c);
Y(() => () => a ? b : c);
```

**Output:**

```jsx
Y(() => (a ? b : c));
Y(() => () => a ? b : c);

```

**Expected behavior:**

For consistency:

```jsx
Y(() => (a ? b : c));
Y(() => () => (a ? b : c));
```

fisker:
Same problem as #11593, they use different logic.

PR #14633, opened 2023-03-30
Make curried arrow function and normal arrow function consistent

seiyab:
## Description
Resolves https://github.com/prettier/prettier/issues/14290 .

What I did are...
- unify `printArrowChain()` into `printArrowFunction()` to fix the issue and prevent other existing / future inconsistencies.
- extract some codes from `printArrowFunction()` (and `printArrowChain()`) into new `mayBreakAfterShortPrefix()`, `joinArrowFunctionSignatures()` and `printArrowBody()`.

I'm worrying that it might get somewhat harder to read and change.
But I don't have better ideas.

<details><summary>Old contents. Just for see history</summary>

Still in progress.
I think I need to
- check similar bugs around here
- check consistency
- refactor it to prevent similar bugs
- add more tests

Edit: I'm busier than before. Go ahead if someone can go faster, and feel free to refer to or cherry pick my commits. I myself am still working on it.
Edit 2: I'm still working on it. I'm checking consistency more.

</details>

## Checklist

- [ ] ~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~


src/language-js/print/function.js
```diff
@@ -249,160 +248,243 @@ function printArrowFunctionSignature(path, options, print, args) {
...
+ * @param {*} parent
+ * @param {*} node
+ * @param {Object} flags
+ * @param {boolean} flags.shouldAddParensIfNotBreak
+ * @param {boolean} flags.shouldAlwaysAddParens
+ * @param {boolean} flags.shouldPutBodyOnSameLine
+ * @param {*} args
+ * @param {*} options
+ */
+function printArrowBody(
```
fisker, line 323: printArrowFunctionBody

src/language-js/print/function.js
```diff
@@ -249,160 +248,243 @@ function printArrowFunctionSignature(path, options, print, args) {
...
-      signatures,
-      chainShouldBreak,
+  const { body } = tailNode;
+
+  // In order to avoid confusion between
+  // a => a ? a : a
+  // a <= a ? a : a
+  const shouldAddParensIfNotBreak =
+    body.type === "ConditionalExpression" &&
+    !startsWithNoLookaheadToken(
```
fisker, line 418: This is an expensive call, can we avoid it if possible?

src/language-js/print/function.js
```diff
@@ -249,160 +248,243 @@ function printArrowFunctionSignature(path, options, print, args) {
...
+function printArrowFunctionBody(
+  bodyDoc,
+  bodyComments,
+  parent,
+  node,
+  { shouldAddParensIfNotBreak, shouldAlwaysAddParens, shouldPutBodyOnSameLine },
+  args,
+  options
+) {
+  const printTrailingComma =
```
fisker, line 332: Let's rename it to `shouldPrintTrailingComma`, or just

```js
const trailingComma = 
args?.expandLastArg && shouldPrintComma(options, "all")
? ifBreak(",")
: "";
```

src/language-js/print/function.js
```diff
@@ -249,160 +248,243 @@ function printArrowFunctionSignature(path, options, print, args) {
...
+ * @param {boolean} flags.shouldPutBodyOnSameLine
+ * @param {*} args
+ * @param {*} options
+ */
+function printArrowFunctionBody(
+  bodyDoc,
+  bodyComments,
+  parent,
+  node,
+  { shouldAddParensIfNotBreak, shouldAlwaysAddParens, shouldPutBodyOnSameLine },
```
fisker, line 328: Can we pass `path` here, instead of `node`, `parent`. Maybe we can also move those `should*` check into this function?

src/language-js/print/function.js
seiyab: I'm not sure if I understand your comment correctly.
Anyway, just eliminating this call looks reasonable.
It looks to improve not only performance but simplicity of code and readability of result.
```js
// Input
() => ({} ? b : c);
a => ({}::b()``[''].c++ && 0 ? 0 : 0);

// Stable
() => ({} ? b : c);
(a) => ({}::b()``[""].c++ && 0 ? 0 : 0);

// Next (difference against stable probably comes from #14599)
() => ({}) ? b : c;
(a) => ({})::b()``[""].c++ && 0 ? 0 : 0;

// Before eliminating it in this branch
() => ({}) ? b : c;
(a) => ({})::b()``[""].c++ && 0 ? 0 : 0;

// If we eliminate the call
() => (({}) ? 1 : 2);
(a) => (({})::b()\`\`[""].c++ && 0 ? 0 : 0);
```

seiyab: Or, I wonder if we can add parens for conditional expression even if it breaks like sequence expression.
Then, the logic will get more simpler and we might reduce chance of call of `startsWithNoLookaheadToken()` adding `arrowFunctionBody.type !== "ConditionalExpression"`.

src/language-js/print/function.js
seiyab: Passing `path` instead of `node` and `parent` looks feasible.
I'll change it later.

Moving `should*` into `printArrofFunctionBody()` is not straight forward because it also affect format of signatures.
- `shouldAddParensIfBreak` affects format of signatures.
    - `shouldAddParensIfBreak` → `shouldAddParens` → `shouldPutBodyOnSameLine` → `shouldBreakBeforeChain` → format of signatures
- `shouldAlwaysAddParens` affects format of signatures.
    - `shuoldAlwaysAddParnens` → `shouldAddParens` → `shouldPutBodyOnSameLine` → `shouldBreakBeforeChain` → format of signatures
- `shuoldPutBodyOnSameLine` → `shouldBreakBeforeChain` → format of signatures

Here, I denoted A → B when A affects B, and A → B → C when A affects B and B affects C.

We have some options. I'll write them.

src/language-js/print/function.js
seiyab: Looks good. I'll apply it.

src/language-js/print/function.js
seiyab: Though it might be better, I also think that it looks out of scope to add parens for conditional expression even if it breaks.
It will change output for so many practical codes.

fisker: Sorry for didn't express good. I didn't mean to change logic, but don't all it when we don't need it. I'll take a look.

src/language-js/print/function.js
seiyab: Option E.  Both  `printArrowFunction()` and `printArrowFunctionBody()` calculates `should*` independently even if they should always same.

src/language-js/print/function.js
seiyab: I'm sorry but I don't understand it yet... 🙇 
At the previous revision, `startsWithNoLookaheadToken()` weren't called if `body.type === "ConditionalExpression"` is falsy thanks to short-circuit evaluation, and, I think, avoiding unnecessary call was already achieved.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation

fisker: Now if `body.type === 'SequenceExpression'` or `shouldPutBodyOnSameLine`, we don't need call it [`8ef4b8d` (#14633)](https://github.com/prettier/prettier/pull/14633/commits/8ef4b8d49aedd938a54587a291dcdfe638987e5e).

Make sense?

src/language-js/print/function.js
fisker: I see they both need those checks now.

src/language-js/print/function.js
fisker: You might miss read my comment, by comment is about "SequenceExpression", not "ConditionalExpression" 😄

seiyab: Thank you so much for friendly description.
I missed `shouldPutBodyOnSameLine`.
I meant, if `body.type === 'SequenceExpression'`, `body.type === "ConditionalExpression"` should be false and the function isn't called.
I'm still on the way to understanding it, but it looks to make sense.

fisker:
@seiyab I've made changes I want. Please take a look, and feel free to make changes if you want.

fisker:
`arrow-function.js` even longer than `function.js` :smile:

fisker:
I didn't know curried-arrow-functions not calling `print` before, so the cursor tracking and range format support should be very bad. But it's out scope of this PR.

fisker:
I didn't know curried-arrow-functions not calling `print` before, so the cursor tracking and range format support should be very bad. But it's out scope of this PR.

src/language-js/print/arrow-function.js
```diff
@@ -0,0 +1,335 @@
...
+  }
+  return shouldAddParensIfNotBreakCache.get(node);
+}
+
+// We handle sequence expressions as the body of arrows specially,
+// so that the required parentheses end up on their own lines.
+const shouldAlwaysAddParens = (node) => node.type === "SequenceExpression";
+
+const isArrowFunctionChain = (node) =>
+  node.body.type === "ArrowFunctionExpression";
```
seiyab, line 62: [Question]
As of main, it looked to `currentNode.body.type !== "ArrowFunctionExpression" || args.expandLastArg` reading just `rec()` in `printArrowFunction()`.
Do the difference cause a problem?

Though it doesn't cause any problem as far as I tried, I'm not sure about it because I'm not familiar with `expandLastArg`.
Example of input I tried:
```js
fooooooooooooooooooooooo(action => a => b => 
  (dispatch123456789098765432123451234567789(action), 1, 2)
);
```

seiyab (commented):
Looks good, thanks!

src/language-js/print/arrow-function.js
fisker: > Do the difference cause a problem?

I also worried about it, and it may cause problem, but I forgot about it...

Related logic here  https://github.com/prettier/prettier/blob/7f292bd8f3cda8eade91012b50ab481b32e3d2e8/src/language-js/print/call-arguments.js#L221

seiyab: As my first impression, removing `|| args.expandLastArg` int `rec()` looked to make sense.
But it will change some results, and previous format looks slightly better.
I'm still investigating.
```diff
- request.get(
-   "https://preview-9992--prettier.netlify.app",
-   (head) =>
-     (body) =>
-     (modyLoremIpsumDolorAbstractProviderFactoryServiceModule) => {
-       console.log(head, body);
-     },
- );
+ request.get("https://preview-9992--prettier.netlify.app", (head) =>
+   (body) =>
+   (modyLoremIpsumDolorAbstractProviderFactoryServiceModule) => {
+     console.log(head, body);
+   });
```

seiyab: This yields same results regardless of `|| args.expandLastArg`. So `expandLastArg` looks to work.
```js
request.get("https://preview-9992--prettier.netlify.app", (head) => (body) => {
  console.log(head, body);
})
```

fisker: I read the code again, when `expandLastArg` is true. `functionBody` can be an `ArrowFunctionExpression`, and `functionBody` decides how the body should be printed. Can't find a case to test. [`98c36c2` (#14633)](https://github.com/prettier/prettier/pull/14633/commits/98c36c2e881ffb328dcdec194f35a0f5e2af33b0) should restore the old behavior, I think.

fisker: The following code can make `functionBody` be an `ArrowFunctionExpression` and effect `shouldBreakChain`

```
a("", "", () => ({}) => {
  a;
});
```

- last argument is `ArrrowFunctionExpression` satisfies `couldExpandArg`
- first two arguments satisfies `args.length !== 2` in `shouldExpandLastArg`
- first two arguments is very simple so they will pass `headArgs.some(willBreak)` in `printCallArguments`

fisker: Finally find a case formats differently....

```js
a("", "", ({}) => () => (a ? b : c));
```

fisker:
```console
# this one
>node bin/prettier test.js
a(
  "",
  "",
  ({}) =>
    () => (a ? b : c),
);

# stable
>node node_modules/prettier/bin-prettier.js test.js
a(
  "",
  "",
  ({}) =>
    () =>
      a ? b : c
);
```

fisker:
@seiyab Can you make the following code work as v2?

```js
a(
  "",
  "",
  ({}) =>
    () =>
    () =>
    () =>
    () =>
    () =>
    () => (a ? b : c),
);
```

seiyab:
I'll take a look. Thanks a lot for your detailed investigation.

seiyab:
I don't still understand it yet, but perhaps, we can say, current revision's behavior shows more consistent result with non-curried arrow function?
```js
a(
  "",
  "",
  ({}) =>
    () =>
    () =>
    () =>
    () =>
    () =>
    () => (a ? b : c),
);

a(
  "",
  "",
  (
    {},
    argument,
    argument,
    argument,
    argument,
    argument,
    argument,
    argument,
    argument,
  ) => (a ? b : c),
);
```

fisker:
But we put body on next line [if it's an identifier](https://deploy-preview-14633--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEBDAFAHSgAh5kAgGmzwONJ3WAF8BKHAXgD5K90GW2rPXc8eTPgMFd+7Xtw5Cc8AM4wSUOgG5s2DJXIglZQjsrV6QqZPGjhA6WJHXLEk+buy4CnAH4cARhxIcAJiVVHRAIAAcYAEtoOWRQVAAnBIgAdwAFRIRYlFQAGxTUAE9YohAAIwTUMABrOBgAZVQAWzgAGUioOGQAMzy5OFKKqtqGsKqOgHNkGASAVwGQfqbI6bmFuUncuABFWYh4Hr6FgCs5AA96zZ29g6Re3P7SgEcbuDTksOy0OQBaTrgACaAkIzVCRXKTADCECaTVQyDQuVyIQ2UAmWwAgjAZpEyrN4Gk4Al2p1Dg8FgALGBNXIAdQpkXkYzAcHqWUZkQAbozCgiwHISiBOfMAJJQIGwepgBKRCIY8X1GCFLZkx4gMLJfq0yphBEa1xEzldUodfoJGDvVATOGqhZjBJmhFlVBlODI0oajowWmRAEwCnIAAcAAZSgk4C9IuHLdb4XcjqUYC6fX6A0hAiBZv0ACou7L3NVwJqugFAgGtVBo2ZWuAAMQgCTh2MmCNQ+IgIBoNCAA)

fisker (approved):
Good job.

seiyab:
Thank you for your review.
To be honest, I'm worrying that [0eb4cc7](https://github.com/prettier/prettier/pull/14633/commits/0eb4cc7a241f688cfe092139069b4cfe555edd5b) might not meet your request.
Anyway, I investigated the input and make ternary be printed same as identifier (and many other "default" ones).

(I don't mean to stop merging as long as you accept it. This comment is for information only.)

fisker:
We can always fix bugs when they are found.

Issue #14624, opened 2023-03-29
Markdown nested code blocks delimited by tildes error


KieranHunt:
**Prettier 2.8.7**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEADdBbAhgJwNYAmEA7lADpQBUlAKgBYCWAzgATMvb5GnUUB+AgFZNoFMiBiNW7YaJD8BFdKhAAaEBAAOMBtCbJQuHCQAKuBPpRYANsSwBPfeoBGOLGDxwYAZSwY4ADIMUHDIAGY2THAubh5e3pruwQDmyDA4AK7RIHAYznAEBAUBWFDJGVjJcABiEDjYMDplyCBYGTAQaiB0MBjWAOqM8EyJYHDeFgw6AG5T9i1gTE4gwVE4MCZuydjhkdnCAB7eKdZwAIoZEPC71lHqiThrLZyEJFBdmjjBMP0MBJLIAAcAAZ7sYov03JoWp84GtpqF1ABHS7wTZaSytJgAWhCBQKXRwcBRDCJm0qOyQEVu2SiGAYaUytJOcAAgo0vs52nATHAcEEQjc7iAmCyLldQlS9uoYFhnL9-nRkAAmGVuBjWFIAYQgGEpOSYAFYuhkojQ5ZZqcLplkAJJQIqwbxgL7aVkO7wweynIVwAC+fqAA)

```sh
--parser markdown
```

**Input:**

````markdown
```markdown
**This is markdown**
~~~json
"this is json"
~~~
```
````

**Output:**

````markdown
```markdown
**This is markdown**

```json
"this is json"
```
```

````

**Expected behavior:**

````markdown
```markdown
**This is markdown**

~~~json
"this is json"
~~~
```
````

or

`````markdown
````markdown
**This is markdown**

```json
"this is json"
```
````
`````

Hey Prettier crew. I was [writing a blog post](https://kieran.casa/markdown-in-markdown/) about how to write Markdown code blocks inside Markdown code blocks. There are multiple ways of doing this, the easiest being to use different numbers of backticks (\`) for each nested code block. Another way of doing it is to use, for example, triple backticks for the outer code block and then triple tildes for the inner one. That's what I used in the example above. 

The issue, though, is that Prettier is converting the triple tildes to triple backticks. With that change, my Markdown parser (GFM through Kramdown on Jekyll) is no longer able to delimit between the outer and inner code blocks. 

Suggestions:
[...]

PR #14634, opened 2023-03-30
Add support for nested Markdown code blocks with backticks and tildes (#14624)

KieranHunt:
## Description

With this change the following code will now format correctly:

````markdown
```markdown
This is markdown

~~~json
["json", "in", "markdown", "in", "markdown"]
~~~
```
````

## Checklist

- [ ] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~~

## Open questions:

- RE "If the change is user-facing". Does this warrant an update to the CHANGELOG? I'm happy to write one if it does.
- Would you like any more tests cases from me?

### DIFF
--- a/src/language-markdown/embed.js
+++ b/src/language-markdown/embed.js
@@ -14,7 +14,7 @@ function embed(path, options) {
       return async (textToDoc) => {
         const styleUnit = options.__inJsTemplate ? "~" : "`";
         const style = styleUnit.repeat(
-          Math.max(3, getMaxContinuousCount(node.value, styleUnit) + 1)
+          Math.max(3, getMaxContinuousCount(node.value, "`") + 1, getMaxContinuousCount(node.value, "~") + 1)
         );
         const newOptions = { parser };
         if (node.lang === "tsx") {
--- a/tests/format/markdown/code/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/markdown/code/__snapshots__/jsfmt.spec.js.snap
@@ -272,6 +272,32 @@ proseWrap: "always"
 ================================================================================
 `;
 
+exports[`nested-code.md - {"proseWrap":"always"} format 1`] = `
+====================================options=====================================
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/language-markdown/embed.js [hunk 1]
```diff
@@ -14,7 +14,7 @@ function embed(path, options) {
       return async (textToDoc) => {
         const styleUnit = options.__inJsTemplate ? "~" : "`";
         const style = styleUnit.repeat(
-          Math.max(3, getMaxContinuousCount(node.value, styleUnit) + 1)
+          Math.max(3, getMaxContinuousCount(node.value, "`") + 1, getMaxContinuousCount(node.value, "~") + 1)
         );
```
fisker, line 18:
```suggestion
        const style = styleUnit.repeat(
          Math.max(2, getMaxContinuousCount(node.value, "`"), getMaxContinuousCount(node.value, "~")) + 1
        );
```

fisker (changes requested):
Can you add a changelog entry?

KieranHunt:
> Can you add a changelog entry?

Yep will do later today.

fisker:
**Input:**

`````markdown
````markdown
```json
{}
```
~~~~~~~~~~json
{}
~~~~~~~~~~
````
`````

**Before:**

````````````markdown
```````````markdown
```json
{}
```

```json
{}
```
```````````

````````````


**After: **

`````markdown
````markdown
```json
{}
```

```json
{}
```
````

`````

fisker:
Unfortunately... it's a known issue, the `printDocToString` mutate docs...

KieranHunt:
Sorry that it's a bit late, but I've now added a CHANGELOG.md entry.

fisker:
https://github.com/prettier/prettier/blob/54add8cfe830af260f1f0186135ff965fe250afc/changelog_unreleased/TEMPLATE.md?plain=1#L3-L21

PR #14638, opened 2023-03-31
Improve doc printer when `printWidth: Infinity`

fisker:
## Description

Ref: https://github.com/prettier/prettier/issues/14514#issuecomment-1471000708

In #14634, maybe we should also print the doc with `printWidth: Infinity` and check the printed string instead?

## Checklist

### DIFF
--- a/src/document/printer.js
+++ b/src/document/printer.js
@@ -201,6 +201,10 @@ function fits(
   groupModeMap,
   mustBeFlat
 ) {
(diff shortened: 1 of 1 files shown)


sosukesuzuki (commented):
Add tests?

fisker:
The test didn't fail means it works. It's a pref improvement, can't test. We already have `printWidth: Infinity` in code and tests.

sosukesuzuki (approved):

PR #14639, opened 2023-04-01
Fix incorrect formatting of member chains where last call has a single template literal argument

### LINKED ISSUE #13533: Incorrect formatting of member chain if last call has template literal as single argument
**Prettier 2.7.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEACdG4A8AOcwwAUhAlKgLwB8qUEATgLYCGANgJYBecA8jjG9ADOhYKlZJUMOgFc4qAL4kSAOgBmECKWUAjJhy266hVCQA6UDJdTKYEACoALOhADuAUTrO6AWSYwwDmxQAOYAklDsUHAAylBMOIIOEEQABuZWGKYgAAK62nAsAPQ4dHCCcDAAtKosrhIA5Kz1qGyCNMliqABurGwAJpIQOJUscF0FqEP80MrpGaiVqAAi-agAnhDSqAxwTBaNLCz1APxZc5YpZlAgADQgUwJQgsigTJ6uAApvCM8orC5MNbPO7aOhMMAAawq0SYOwAMkE4MhVKxyiCwZDoThwUFgsgpLI7nAGPk+n04H04XtgtImME4AAxejMGD8ELIEBMaS2W4gBwwBgsADqgXggmxYBiPzY-C6MrWHLAgmBICC5ToMA+YOCzGRqLgdwAVoIsNFcaMAIrSZJIpAolhokDYujqjl5Aq8kpBGBC-owBzIAAcAAY7iUIOUhWCcBySmU4HRxryAI7W+Baoa-TmCSpRCkU3mlVNsUpaum6u36u7lBhsfEyA0gQTmuBWm16h2NmBMbS+vr+5AAJjuUiYbEiwQAwhAGBWQGUAKy86TlOw9372x1dWThcmwaJgOhsPgAQSgfWiMDWow75Xk8iAA)

```sh
--parser babel
```

**Input:**

```jsx
      expect(() => normalizeOptions({ al: true })).foo().baz().bar( )
        .toThrowErrorMatchingInlineSnapshot(`
        "@babel/preset-flow: 'al' is not a valid top-level option.
        - Did you mean 'all'?"
      `)

```

**Output:**

```jsx
expect(() => normalizeOptions({ al: true }))
  .foo()
  .baz()
  .bar().toThrowErrorMatchingInlineSnapshot(`
        "@babel/preset-flow: 'al' is not a valid top-level option.
        - Did you mean 'all'?"
      `);

```

**Expected behavior:**

```jsx
expect(() => normalizeOptions({ al: true }))
  .foo()
  .baz()
  .bar()
  .toThrowErrorMatchingInlineSnapshot(`
        "@babel/preset-flow: 'al' is not a valid top-level option.
        - Did you mean 'all'?"
      `);

```

solarized-fox:
## Description

Fixes #13533

## Checklist

### DIFF
--- a/changelog_unreleased/javascript/14639.md
+++ b/changelog_unreleased/javascript/14639.md
@@ -1,0 +1,27 @@
+#### Fix incorrect formatting of member chains where last call has a single template literal arguement (#14639 by @solarized-fox)
+
+<!-- Optional description if it makes sense. -->
+
+<!-- prettier-ignore -->
+```jsx
+// Input
+foo("long" + "set").bar("of" + "arguments").baz("will" + "break")
+.qux(`
+    template literal
+`)
+
+// Prettier stable
+foo("long" + "set")
+  .bar("of" + "arguments")
+  .baz("will" + "break").qux(`
+    template literal
+`);
+
+// Prettier main
+foo("long" + "set")
+  .bar("of" + "arguments")
+  .baz("will" + "break")
+  .qux(`
+    template literal
+`);
+```
--- a/changelog_unreleased/javascript/14639.md
+++ b/changelog_unreleased/javascript/14639.md
@@ -1,7 +1,5 @@
 #### Fix incorrect formatting of member chains where last call has a single template literal argument (#14639 by @solarized-fox)
 
-<!-- Optional description if it makes sense. -->
-
 <!-- prettier-ignore -->
 ```jsx
 // Input
@@ -10,12 +8,20 @@ foo("long" + "set").bar("of" + "arguments").baz("will" + "break")
     template literal
 `)
 
+insertRule(`*, *:before, *:after {
+  box-sizing: inherit;
+}`);
+
 // Prettier stable
 foo("long" + "set")
   .bar("of" + "arguments")
   .baz("will" + "break").qux(`
     template literal
 `);
+
+insertRule(`*, *:before, *:after {
+   box-sizing: inherit;
+}`);
 
 // Prettier main
 foo("long" + "set")
@@ -24,4 +30,9 @@ foo("long" + "set")
   .qux(`
     template literal
 `);
+
+insertRule(
+   `*, *:before, *:after {
+   box-sizing: inherit;
+ }`);
 ```
--- a/changelog_unreleased/javascript/14639.md
+++ b/changelog_unreleased/javascript/14639.md
@@ -1,4 +1,4 @@
-#### Fix incorrect formatting of member chains where last call has a single template literal arguement (#14639 by @solarized-fox)
+#### Fix incorrect formatting of member chains where last call has a single template literal argument (#14639 by @solarized-fox)
 
 <!-- Optional description if it makes sense. -->
 
--- a/src/language-js/print/call-arguments.js
+++ b/src/language-js/print/call-arguments.js
@@ -22,7 +22,6 @@ import {
   isTSTypeExpression,
   isArrayOrTupleExpression,
   isObjectOrRecordExpression,
-  isTemplateOnItsOwnLine,
 } from "../utils/index.js";
 
 import {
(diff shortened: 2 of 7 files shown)


fisker:
**Prettier pr-14639**
[Playground link](https://deploy-preview-14639--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAzCEAUAdEAbaAc1wAIBqE3AZzhlwEoA6AIwEMAnHECVUi3DoQCuAWwQwqDFqwBeXAO4BLPHj6UQzdnFYBrBtiiMAjkIAemAAYGSNkvBEAHPK3gk8i+O1aqoFpjAgAZRh2RShCTHoQABpuBxhFaCpkUA52CHkABQ4EZJRveVYAT2TYzVYwHVpA1jEAGTC4ZFRvGjKvSuqHCrDCZBChOFiaEUV+9kHh3rw4AEUhCHhm1qGQACsqU0DpuYWlpBa8NpATRbhM9Ic8kFYqAFooODgAExeYkBDWZV6AYQgRESsZA3FTvKg7ACCMBCimYQngmTg7Aaj2WR1WAAsYCI8AB1DEeOBUbpgOCBXIeRQANw8RWBYCopRAVMGAEkoK9YIEwKF4hCOcEijM0ccHOkaLivA5gWKiUiqU1YmEaOwYBdWIRASLVt12CrgWxmHA8O8xWEYLjFM8YBjkAAOAAMsS0JkUWnVmqBBxWsRgrGYluttqQACZYkIaAAVf15Q7HOAiI3PV7POqscJCDVwABiEHYgOhvWBrHhEBAAF9y0A)

```sh
--parser babel
```

**Input:**

```jsx
foo("long" + "set").bar("of" + "arguments").baz("will" + "break")
.qux(`
    template literal
`).toString()
```

**Output:**

```jsx
foo("long" + "set")
  .bar("of" + "arguments")
  .baz("will" + "break")
  .qux(
    `
    template literal
`,
  )
  .toString();

```

Maybe 

```js
foo("long" + "set")
  .bar("of" + "arguments")
  .baz("will" + "break")
  .qux(`
    template literal
`)
  .toString();
```

Should be preferred.

solarized-fox:
ah, nice catch!! it would seem that that turns it back into an internally-recognized member chain element… if this is the case the only sensible way to deal with this might be by removing the `isTemplateLiteralSingleArg` special case in call-expression.js. if anybody knows precisely why that special case exists, that would speed up my investigation

solarized-fox:
alright, I implemented the change suggested by @fisker. I decided to move the `isTemplateLiteralSingleArg` check into `call-expression.js` and was able to move my logic there as well. effectively I chose a different point in the code to implement the same changes, so I force pushed a new version of the branch for simplicity's sake. the relevant tests have been updated as well, everything should be good to go

fisker:
These two should format the same

**Prettier pr-14639**
[Playground link](https://deploy-preview-14639--prettier.netlify.app/playground/#N4Igxg9gdgLgprEAuEAzCEAUAdEAbaAc1wAIBqE3AZzhlwEoA6AIwEMAnHECVUi3DoQCuAWwQwqDFqwBeXAO4BLPHj6UQzdnFYBrBtiiMAjkIAeOKCRIADA1avwRABzyt4JPIvjtWqqNfoDA3QsXAIoYhBydRo6ECY2TlweNQF2YTFYSXjpOVwlFVSNLV19QxNzW0t7Rxc3OA8vOB8-AJAAGm4nGEVoKmRQDnYIeQAFDgR+lF95VgBPfs7NVjAdWgBlVjEAGUUoOGRUXxoln1WNpxW9wmQYdiE4TpoRRVv7x5Aqa7w4AEUhCDwQ7HD4AKyopnW3z+AKBSCOeBOIBMgLgo2GTimIFYVAAtPs4AATIkdEB3VjKa4AYQgIhErGQ2JUpK+ER+AEEYHdFMwhPBRs1dvtgYiPgALGAiPAAdTFTSolzAcHWky8igAbl45oywFRFiB1Q8AJJQYmwdZgdiKbrs03rGBzH4ipFOYY0aU+JyM11wGjsdUHTp7P0wdGsQj050fS7sP2MtjMOB4UmuvYwaWKQkwMXIAAcAAZOloTIotGGIwz4SDOjBWMwM1mc0gAEydIQ0AAqdamCKRcBEicJxMJ21YESE4bgADEIOx6VzrozWHyICAAL5roA)

```sh
--parser babel
```

**Input:**

```jsx
foo("long" + "set").bar("of" + "arguments").baz("will" + "break")
.qux(
  `
    template literal
`)

foo("long" + "set").bar("of" + "arguments").baz("will" + "break")
.qux(`
    template literal
`)
```

**Output:**

```jsx
foo("long" + "set")
  .bar("of" + "arguments")
  .baz("will" + "break")
  .qux(
    `
    template literal
`,
  );

foo("long" + "set")
  .bar("of" + "arguments")
  .baz("will" + "break")
  .qux(`
    template literal
`);

```

solarized-fox:
> These two should format the same

hmm, I see. I tried naively implementing this but it causes tests to fail in e.g. tests/format/typescript/end-of-line/multiline.ts where

```
console.log(
  `Multiline \n string\
   Multiline string\
   Multiline string`
);
 ```
 
 becomes the incorrect:
 
 ```
 console.log(`Multiline \n string\
   Multiline string\
   Multiline string`);
```

it seems to me that the rule should be that the delimiters "stick" to the parens if there isn't a newline between them in the original source OR there is but the first line is whitespace AND the string ends in a newline. I will try and come up with a workable solution based on this rule.

solarized-fox:
alright, the way I implemented this is basically that a template string like `({[\n bar \n]})` will when appearing as the lone argument to a function call will be formatted as:
```
foo(`({[
 bar 
]})`)
```

as a consequence of this, a couple existing snapshots needed to be updated, for example:
```
-insertRule(\`*, *:before, *:after {
+insertRule(
+  \`*, *:before, *:after {
   box-sizing: inherit;
 }\`);
```
without picking an arbitrary number and measuring the length of the first line of a template, it doesn't seem possible to determine the difference between the above and cases where the template string should start on its own line, so the change appears necessary to me.

the tests have also been updated with the the examples from @fisker's last comment, and as requested, they now format the same.

I added these as additional commits for ease of viewing but if need be I can squash this branch back down to fewer commits before they are merged

solarized-fox:
hmm, I have fixed the listing error, but yarn test shows total success, even when using node 18. I wonder if this is a ubuntu specific test failure? that'd be strange, though

PR #14661, opened 2023-04-03
Forbid decorators on object property

fisker:
## Checklist


CvX:
@fisker could you provide a bit of context for this PR? 😃  It's a breaking change but it wasn't highlighted in the release notes of 3.0.0.

Decorators on properties of object literals are non-standard but are a part of earlier decorators proposals and some code still relies on them. For example Ember.js apps that still contain parts of code from before tooling supported native js classes:

```js
EmberObject.extend({
  @computed("name")
  get hello() {
    return `hi, ${this.test}!`;
  }
}
```

PR #14664, opened 2023-04-04
Don't apply plugin default options from unused plugins

fisker:
## Checklist

### DIFF
--- a/src/main/normalize-format-options.js
+++ b/src/main/normalize-format-options.js
@@ -71,17 +71,13 @@ async function normalizeFormatOptions(options, opts = {}) {
 
   rawOptions.printer = printer;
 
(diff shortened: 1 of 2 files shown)


src/main/normalize-format-options.js
```diff
@@ -71,17 +71,13 @@ async function normalizeFormatOptions(options, opts = {}) {
...
+    ? Object.fromEntries(
+        Object.entries(printerPlugin.defaultOptions).filter(
+          ([, value]) => value !== undefined
+        )
       )
-      .map((optionInfo) => [
-        optionInfo.name,
-        optionInfo.pluginDefaults[printerPlugin.name],
-      ])
-  );
```
fisker, line 84: This problem exits because plugins don't have `name`.

PR #14691, opened 2023-04-11
Update install.md

kumarasinghe:
## Description

Removed depreciations from installation instructions.
Stylelint 15 has depreciated rules that could conflict with Prettier. This config will no longer be needed. [annoucement](https://stylelint.io/migration-guide/to-15/#deprecated-stylistic-rules)

## Checklist

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -99,7 +99,7 @@ See [Editor Integration](editors.md) for how to set up your editor. If your edit
 
 ## ESLint (and other linters)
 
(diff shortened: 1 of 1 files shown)


fisker (approved):

kachkaev (commented):
We should probably update this page too:
https://prettier.io/docs/en/integrating-with-linters.html

https://github.com/prettier/prettier/blob/fb948bb15f31ade3b35e66b720f8d44260de45f8/docs/integrating-with-linters.md?plain=1#L8-L13

It might be also worth searching for more occurrences of `stylelint-config-prettier` in the codebase, just in case.

PR #14731, opened 2023-04-20
Ignore `.gitignore`d files

### LINKED ISSUE #14115: Ignore files that are not tracked by git (by default)
## The current behavior

Prettier is often used in a git repository, and we often have a `.gitignore` or `.prettierignore` file.
As a result, we often end up with a setup like that (in `package.json`):

```json
{
  "scripts": {
    "lint:prettier": "prettier \".\" --check --ignore-path \".gitignore\""
  }
}
```

However, only files/globs specified in the root `.gitignore` file are ignored. If we have `.gitignore` files in subdirectories they will not be taken into account.

In a monolith project or in a project with only one `package.json` it should be uncommon to use `.gitgnore` in subdirectories (not really needed in most Node.js/JavaScript applications).
However, in monorepo and with projects with not only languages supported by Prettier (JavaScript, HTML, CSS, etc.), it doesn't make sense to put everything in the one single `.gitignore`.
By using Prettier, we don't want to lose the convenience we have with `git`. Also even if [ignore npm package](https://www.npmjs.com/package/ignore) can be 100% spec compliant, it will never mirror with perfection what git really ignores or not.

Also, there are already enough dotfiles in our project, if we can avoid them when we can, it's better to do so, hence it's often better to specify  `--ignore-path \".gitignore\"` instead to rely on the `.prettierignore` file.

*Note:* This issue could be applied to many linters in the ecosystem, and by opening this issue on Prettier, I hope that others tools will follow (e.g: ESLint).

## The expected behavior

We want to use Prettier like this (in `package.json`):
```json
{
  "scripts": {
    "lint:prettier": "prettier \".\" --check"
  }
}
```
**By default** Prettier, will ignore all the files ignored by git (if the code is in a git repository of course) without any configuration and with exactly the same behavior as git.
One could argue that it could be a **BREAKING CHANGE** as we're changing *default-behavior* but, it should very uncommon to willing to lint ignored files by git with Prettier.

## Potential solution

Instead to rely on the `ignore` npm package, and doing manually the same algorithm as git to take into account `.gitignore` in subdirectories, we want to use git CLI.
Doing the exact same algorithm as git manually with the `ignore` npm package can get quite complicated to reimplement from scratch.

Similarly to how it used for https://github.com/editorconfig-checker/editorconfig-checker.

Related links:
- https://github.com/editorconfig-checker/editorconfig-checker/issues/99
- https://github.com/editorconfig-checker/editorconfig-checker/pull/135

fisker:
## Description

Fixes #14115

## Checklist


src/cli/cli-options.evaluate.js
```diff
@@ -195,7 +195,7 @@ const options = {
   ignorePath: {
     array: true,
     category: optionCategories.CATEGORY_CONFIG,
-    default: [{ value: [".prettierignore"] }],
+    default: [{ value: [".prettierignore", ".gitignore"] }],
```
kachkaev, line 198: How about ordering these file names alphabetically?

sosukesuzuki (approved):

fisker:
I'm going to wait for #14627, since we'll need update once it land, and I don't want ask the author to make the change.

rotu:
I don't know that this can be said to fix #14115. It only happens to work if:
* you have a single `.gitignore`
* you happen to run prettier from the same directory containing this file
* your `.gitignore` doesn't contain special syntax (e.g. I think character ranges `[a-z]` are supported in gitignore but not globby)
* You don't use `$GIT_DIR/info/exclude` nor `core.excludesFile` https://git-scm.com/docs/gitignore

Perhaps it would be better to filter files with [`git check-ignore`](https://git-scm.com/docs/git-check-ignore), which will reflect git's *actual* behavior.

rotu:
I think this should be reworked to not use .gitignore as a second .eslintignore: https://github.com/prettier/prettier/issues/14115#issuecomment-1529316066

fisker:
> I think this should be reworked to not use .gitignore as a second .eslintignore: [#14115 (comment)](https://github.com/prettier/prettier/issues/14115#issuecomment-1529316066)

I think we'll go with this approach for now.  See https://github.com/prettier/prettier/issues/14115#issuecomment-1517641375

Issue #14207, opened 2023-01-20
Handlebars / Glimmer: Square brackets removed from array index - causes SyntaxError


davidtaylorhq:
**Prettier 2.8.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuExgAc4XQGzgOgG0BGAXXygEMBbOAXzoB0o1Ns8iyGQAaEbGAEtoAZ2ShKAJ0kQA7gAUpCMSko5ZlAJ5i+AI0mUwAazgwAyjTgAZQVDjIAZmpFw9B46bPpDtgObIYSQBXVxA4al04ABMo6KtKKF8gyl84ADEISWpKGCFE5BBKIJgIXhAACxhqHAB1csF4EW8wODNlBsEANwbNArARHRBbF0kYeQNfbMdnUIArEQAPMz88AEUgiHhpnBc+b0kRgt8cQWpaSTL0SVsYGsEomHLkAA4ABj2ZFxqDdAKruBGnXsfAAjht4ONsCpCiIALR2aLRMqSOBgwQo8YpKZIJw7UIuaiCALBfErOAAQVy110xTg8jgkhsdm2uxAIjJ6029hxMz4MEoujuDyeSAATHyDIITokAMIQM6UAoAgCsZSCLgAKgKVLjWZ0QgBJKCxWBmMDXdAwcnGswwTR4Fn0OhAA)

```sh
--parser glimmer
```

**Input:**

```hbs
{{people.[1].name}}
{{people.[1]}}
```

**Output:**

```hbs
{{people.1.name}}
{{people.1}}
```

When parsed, this output produces a syntax error on the second line

```
SyntaxError: Parse error on line 2:
...e.1.name}}{{people.1}}
----------------------^
Expecting 'ID', got 'NUMBER' (2:10)
  1 | {{people.1.name}}
> 2 | {{people.1}}
```

**Expected behavior:**

Prettier should only remove square brackets when the array index is followed by another property.

The correct output for the given input would be:

```hbs
{{people.1.name}}
{{people.[1]}}
```

shogohida:
Hi guys! Can I work on this?

I assume we need to change this file (Not sure which function yet)
https://github.com/prettier/prettier/blob/main/src/language-handlebars/printer-glimmer.js

fisker:
[...]

PR #14737, opened 2023-04-22
Fix square brackets removed from end of array

kennyhngo:
## Description

Fixes issue #14207 and handles literal segments.

## Checklist


Ekinsxd:
LGTM!

src/language-handlebars/printer-glimmer.js
```diff
@@ -365,8 +365,66 @@ function print(path, options, print) {
...
+        "|",
+        "}",
+        "~",
+      ];
+
+      if (node.this) {
+        parts.push("this");
+      }
+
+      for (const [idx, part] of node.parts.entries()) {
```
fisker, line 411: Let's use `node.parts.map(...).join(".")`

PR #14746, opened 2023-04-23
Bundle actual cjs entry

fisker:
## Description

VSCode extention can't use cjs proxy to `import()` esm version, try to bundle CommonJS version of index.cjs

## Checklist


kachkaev:
Bundling CJS in parallel to ESM can help us with https://github.com/prettier/prettier-vscode/pull/2947, but will significantly increase package download size for everyone. A CJS bundle will probably be incompatible with Prettier plugins using ESM. I feel like we need to seek for a solution on the VS Code side, probably as suggested by @sosukesuzuki in https://github.com/prettier/prettier-vscode/pull/2947#issuecomment-1519065479.

fisker:
Closing since it didn't work as expected.

PR #14766, opened 2023-04-27
"browser" field in exports["."] for standalone entrypoint (fixes #14765)

### LINKED ISSUE #14765: Standalone version isn't loaded with ESM and Webpack
**Environments:**

- Prettier Version: 3.0 alpha
- Prettier plugins (if any): 

**Steps to reproduce:**

- import `prettier` in ESM-only React code (`"type": "module"`, and Typescript in my case, but I'm not sure that's relevant)
- try to bundle it with webpack 5

**Expected behavior:**

Bundle is built.

**Actual behavior:**

Multiple errors of this type:

```
ERROR in ../../node_modules/prettier/index.mjs 1:0-66
Module not found: Error: Can't resolve 'module' in '/Users/berekuk/coding/quri/squiggle/node_modules/prettier'
 @ ./src/components/CodeEditor.tsx 11:0-37 41:50-75
 @ ./src/components/SquiggleEditor.tsx 5:0-45 20:32-42
 @ ./src/index.ts 4:0-64 4:0-64
```

This could be solved with `prettier: "prettier/standalone"` in webpack's `resolve.alias` field, but that's a solution only for the current package; I have multiple apps that use the same components package, and I don't want to impose this requirement on all consumers of my package.

Importing `prettier/standalone` directly in the React code doesn't help either because I depend on the custom prettier plugin that imports `prettier` instead of `prettier/standalone`.

I think that adding `"browser": "./standalone.mjs"` to `exports["."]` field in `package.json` will fix this. I'm less sure if it would cause any problems for CJS builds.

berekuk:
## Description

Attempt to fix #14765. Seems to help, but I'm not 100% sure if it's a safe change for everyone (should it be `standalone.js` and not `standalone.mjs`?).

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     ".": {
       "types": "./src/index.d.ts",
       "require": "./src/index.cjs",
+      "browser": "./src/standalone.mjs",
       "default": "./src/index.js"
     },
     "./*": "./*"


fisker:
Not sure we want this change yet. Anyway change should make in this https://github.com/prettier/prettier/blob/1107ae6bd099f063d8edfb20a3905a22af01f0e1/scripts/build/build-package-json.js#L41-L45

package.json
```diff
@@ -16,6 +16,7 @@
     ".": {
       "types": "./src/index.d.ts",
       "require": "./src/index.cjs",
+      "browser": "./src/standalone.mjs",
```
fisker, line 19: Can you try if this work?

```json
{
	"exports": {
		".": {
			"browser": {
			  "module": "./standalone.mjs",
			  "default": "./standalone.js"
			}
		}
	}
}
```

PR #14768, opened 2023-04-27
Fix long comments causing unexpected newlines in concise arrays

### LINKED ISSUE #14160: Number array and Long comment
**Prettier 2.8.2**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAdKB6DACANtAc20gFsSEZ1IoBnGbKAVxICM4AnG7AXmwG102bFmIQyFQdgAMADylSANNLmLl8hZJGlysGHv0HDR4ydNnz+ybPVrV1xZpzaJUALoBudCAUgIABxgAS2gaZFAAQ3Z2CAB3AAVIhFCUcNwY8IBPUJ8WdnCwAGs4GABlcPIAGUCoOGQAM1SaOBy8wuKSv3zqgmQYdkZmkDhWOAATUbGK8KgCRnCCOAAxCHYScL1u5BBwxhgIbxAACxgSXAB1Q8D4Gk6wOBKkq8CANyuMrbAabJBqpvYYOJ5AhreqNQYAKxoMhK3VwcAAiowIPBQbgmj5OpwOFsWOE2LgDn52NUYGdAqMYIdkAAORQgIkQJpnPJ+LZEuB-Z61HwARyR8EB-mS2xoAFoamMxgd2HA+YEZYD5iCkA00YMmiRAr1+urYXAAIJ6YksXZwOIcKo1VHokA0PWI5G1FVgnwwPFkilUpAAJldeUCuG6AGExMqhjQAKwHRhNAAqeOSqptzwGAEkoBNYCUwMSAvqMyUYBk4da4ABfMtAA)

```sh
--parser babel
```

**Input:**

```jsx

// long comment
const numbers = [
  // comment
  0x00, 0x00, 0x00,
  // commentttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt
  0x00, 0x00, 0x00,
  // comment
];

```

**Output:**

```jsx
// long comment
const numbers = [
  // comment
  0x00, 0x00, 0x00,
  // commentttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt
  0x00,
  0x00, 0x00,
  // comment
];

```

**Expected behavior:**

solarized-fox:
## Description

fixes #14160

as detailed in the issue, long comments would cause the array elements they are attached to to break unexpectedly. this appears to be due to how src/document/printer.js handles breaking, i.e. if the comment is in the same array `Doc` as the element, it counts toward line width, apparently. for leading line comments on concise arrays this produces the odd newline outlined in #14160

the only existing snapshots that have been updated are some array literals that can fit more on a given line because their comments aren't being incorrectly counted toward line width

## Checklist


fisker:
This is not the correct fix. There should be something wrong in the doc printer.

fisker:
Bug should be here https://github.com/prettier/prettier/blob/7f64e920de81f97f942e12682061309f37fb8337/src/document/printer.js#L515


<details>
<summary>test script</summary>

```js
import { fill, hardline, line } from "./src/document/builders.js";
import { printDocToString } from "./src/document/printer.js";

const doc = fill([["/".repeat(81), hardline, "1"], line, "2"]);

console.log(printDocToString(doc, { printWidth: 80 }));
````

</details>

**Comment by solarized-fox** [comment]
it looks like the problem is that in order to correctly determine if the next doc will fit, you need to look backwards and see if the text between the start of the second and end of the first contents fit on the same line. I added an extra step to the process which should resolve this

**Comment by solarized-fox** [comment]
wait, after sleeping on it I realized I could just do this. ignore the previous version

**Comment by solarized-fox** [comment]
wait, I was half-right before. it appears that checking `secondContent` in a separate pass is necessary, for cases such as
```
const numbers2 = [
 	/////////////////////////////////////////////////////////////////////////////////
 	1,
 	/*/////////////////////////////////////////////////////////////////////////////////*/
 	2,
 ]
 ```

seiyab:
IMO, stable `printDocToString()` works fine for #14160 in a sense.  I think a comment for `fill()` in `printDocToString()` and documentation for `fill` is slightly inconsistent and confusing so it might need discussion.
https://github.com/prettier/prettier/blob/1ea2fd059a2ef8762f062e3540b9d48596c43b58/src/document/printer.js#L450-L458
https://github.com/prettier/prettier/blob/56f5f4c736f35506e8a256131df2cb50d87391ab/commands.md#fill
> This is an alternative type of [group](https://github.com/prettier/prettier/blob/56f5f4c736f35506e8a256131df2cb50d87391ab/commands.md#group) which behaves like text layout: it's going to add a break whenever the next element doesn't fit in the line anymore. The difference with group is that it's not going to break all the separators, just the ones that are at the end of lines.

Let's consider #14160 and now we are printing `fill([["// long comment", hardline, "0x00", ","], line, "0x01", line, ...])` (slightly modified for explanation).
- According to the comment:
    -  the first element `["// long comment", hardline, "0x00", ","]` doesn't fit on the line without breaking. Following `line` should be broken. Stable Prettier is right.
- According to the documentation:
    - The documentation doesn't mention for `line` after breaking element.  And following `"0x01"` can be fit in the line without breaking so `line` between `"0x00"` and `"0x01"` shouldn't be broken. Stable Prettier has a bug.

IMO, I believe the comment shows more preferable rule.
Consider following case: [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4D4A6UABEQO4QBO6qAzgA4CGUu5VZl67brnPqA9PSYsOhAVjyEQAGhAQ6MTNBrJQDChQikACuoQqUDADakGATxWyARhQZgA1nBgBlBgFs4AGUxQ4yAGbGNHDWtg5OzoxgPgDmyDAUAK4hIMFumPFJKXAAHnRwFJgesMYAKgVQ6phwBoFGwbI0sUZwAIqJEPABQSkAVjQ5zs1tHV1IdQ0gAI6jcNqadAYgDDQAtL5w6JsyIAkMmEaxAMIQbm4MyCA1AKw7TVAxLQCCMAmYVonw2gXevt31KQAFjA3EYAOqAzDwIRgODOfRQnBQsyXMA0SwgbDJACSUC2sGcYEKCieeOcMDMLX+kzommCYNsdEutJqBWwflkPmCFBg8wYMXO1JSjAo3MuVgYVjgRh2tJ8MDBmHQMEByAAzAAGWQUOAzTA6vkCi7jHqyGCSxXK1VIABMskSwVKktqpqubil6C26E8TBiiX5cAAYpRzq9YpcGJ8ICAAL4xoA)
Doc of children of `<div>` is approximately following:
```js
fill([
  "word",
  softline,
  group(/* doc for <span>word word word word word</span> */)
  softline,
  "word"
])
```
Stable Prettier prints pretty output:
```jsx
<div>
  word
  <span>
    word word word word word
  </span>
  word
</div>;
```
Considering only next element of line-like as documentation suggests, it will go:
```jsx
<div>
  word
  <span>
    word word word word word
  </span>word
</div>;
```
Isn't it unnatural?

Of course, I also agree  #14160 shows ugly format. For #14160, can't we yield doc like following instead of changing `printDocToString()`?
```js
fill([
  "//comment",
  hardline,
  "0x00",
  line,
  "0x00",
  line,
  "0x00",
  hardline,
  "// commentttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt",
  hardline,
  "0x00",
  line,
  "0x00",
  line,
  "0x00",
])
```

PR #14770, opened 2023-04-27
Update configuration.md (Add spaces for readability)

DerTimonius:
## Description

This will add just two spaces in the `configuration.md` file to improve formatting a bit, see [this PR comment](https://github.com/prettier/prettier/pull/14063#issuecomment-1525046241)
e.g. `JS(ESM) -> JS (ESM)`

## Checklist


docs/configuration.md
```diff
@@ -32,7 +32,7 @@ JSON:
 }
 ```
 
-JS(ESM):
+JS (ESM):
```
fisker, line 35:
```suggestion
JS (ES Module)
```

fisker (approved):

docs/configuration.md
DerTimonius: Good idea, added this real quick!

karlhorky: @fisker The long form of `ESM` is more commonly `ES Modules` (with `s`) - what do you think about that instead?

fisker: I don't think so, the config is a single module, why `s`? I'm a native English speaker, I'll let other maintainer decide.

karlhorky: This is what I was thinking was your reasoning for it, and technically,  it kind of makes sense

However, in the context of JS, I would argue it's more common for people to look for the project module format (not a description of a single file) - JS module formats are often described as "CommonJS" vs "ES Modules"

Also, people may search for the more common search term "ES Modules" and not be able to find this part of the page

DerTimonius:
@fisker Is there anything left I can do to get this merged?

PR #14775, opened 2023-04-29
Add note about pre-commit mirror versions to `docs/precommit.md`

kaechele:
## Description

Add notes on how to use prettier with the pre-commit mirror for when certain stable version tags are unavailable.

## Checklist


louwers:
pre-commit no longer supports prettier, so this section can be removed entirely 

https://github.com/prettier/prettier/pull/16229

Issue #4708, opened 2018-06-17
config option to ignore paths


tvald:
It's desirable to place rules for ignoring/excluding paths in `.prettierrc` rather than in `.prettierignore`, so that configuration for Prettier is self-contained within a single file.

(Of secondary concern, the `prettier-vscode` extension should soon support [configuring the location](https://github.com/prettier/prettier-vscode/pull/433) of `.prettierrc`, but not `.prettierignore`. In general, more configuration files are just harder to deal with.)

For prior art, consider TypeScript's "exclude" [configuration option](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html):
```js
// tsconfig.json
{
  "exclude": [
    "node_modules",
    "**/*.spec.ts"
  ]
}
```

ikatyang:
Related: #3460, #2691, and #3764.

saadq:
May be worth mentioning that one of the benefits for allowing this is to improve the usage of shareable Prettier configs. My team has a shareable Prettier config that we're using across our packages, but we still need to create separate `.prettierignore` files that list out all our globs in each one. Allowing this to be store inside the shareable config would be great.

charles-allen:
When running Prettier in a directory, I believe it will look for `.prettierrc` in ancestor directories, but it will not do the same for `.prettierignore`. As a result, I'm currently duplicating ignores (to support executing prettier module-by-module).

I believe, in addition to merging rc & ignore, **this issue would also allow me to eliminate duplicate ignores** by centralizing ignores under one `.prettierrc` (I could delete all the files with comments; `.eslintrc` shown for comparison).
```yaml
.eslintrc.js
.prettierrc.js
.prettierignore           # separate file
admin/.prettierignore     # duplicate
common/.prettierignore    # duplicate
docs/.prettierignore      # duplicate
functions/prettierignore  # duplicate
hosting/.prettierignore   # duplicate
```

ssbarnea:
At this moment, for practical reasons, `.prettierignore` is a symlink to `.gitignore` but that prevents us from adding other exclude for files that are tracked in git but for which we do not want to use prettier.

The current options make it very hard to use prettier or to adopt prettier on bigger codebases as reformatting cannot happen over the night, it can easily take many months, especially if you happen to have at least one developer in the team that dislikes formatters. They will do their best to fight every related change.

eddiemonge:
Is this a feature someone can work on or does the team not want this option?

TomSssM:
[...]

Issue #12923, opened 2022-05-24
.prettierignore is ignored when inside of a subdirectory


Zamiell:
tfw your ignore file is ignored.

Consider the following sequence:

```sh
# Set up a new JavaScript project
mkdir monorepo
cd monorepo
npm init --yes
npm install prettier --save

# Set up the first package in the monorepo, "foo"
mkdir -p ./packages/foo

# Add a second "package.json" file that is not formatted correctly (since it has a leading newline)
echo "" > ./packages/foo/package.json
cat package.json >> ./packages/foo/package.json

# Add a ".prettierignore" file
echo "package.json" > .prettierignore

# Check every file using prettier
npx prettier --check . # Passes!

# Check every file using prettier again, but this time in the foo directory
cd ./packages/foo
npx prettier --check . # Fails!
```

Why the doesn't Prettier use the `.prettierignore` file here?

- When I type `--show-config`, Prettier seems to be able to correctly work backwards to find the `.prettierrc.js` file at the root of the monorepo.
- However, it does not seem to be able to work backwards to get the `.prettierignore` file.

Thus, I strongly suspect that this behavior is a bug in the program.

Zamiell:
Related: Having #12922 would make diagnosing this kind of thing much less painful.

Zamiell:
> Facing this issue as well.

@ScarletMcLearn Then use the thumbs up feature to avoid pinging everyone on the thread. Replying like this with no additional content is considered extremely rude.

PR #14785, opened 2023-05-01
Add parser "ignore" as alternative to .prettierignore

rotu:
## Description

Add a new parser "ignore". This is similar to `.prettierignore`, but:
1. can be included directly in `.prettierrc`, removing a file (#4708)
2. does not require prettier to to be run from the project root to be respected (#4081, #12923)

## Checklist


rotu:
I'm not totally happy with having made two not-so-obvious alternate ways to ignore files. I think this makes config easier to understand, but I'd like to eventually see `.prettierignore` deprecated in favor of this plus vcs integration.

fisker:
Thanks for your contribution. I don't have a good solution for #4708, but I don't think we can accept this solution either.

rotu:
@fisker Why don't you think this is acceptable? It doesn't aim to fix #4708 (which should be somewhat more automatic). It *does* aim to be be a better option over `.prettierignore`, which has a few footguns (especially e.g. in a monorepo)

WillsterJohnson:
With all due respect, it's been 5 years since #4708 was requested.

This PR is the only one which has attempted to fulfill even a portion of #4708's request.

If this isn't acceptable, then a top-level option named `ignore` or `exclude` of type `string[]` must be.
If that still isn't, what is?

If there hasn't been any solution for five years, I think this PR qualifies as "good enough". The pain point isn't "we have one additional file in our project", the pain point is "using prettier in a monorepo feels like more hassle than manually formatting files"

Below are the current options for sharing prettier ignores across projects. All of them are worse than what this PR offers.

**Multiple `.prettierignore` files, one per workspace project**

This does not require justification as to why it's awful.

**Use 3rd party tooling to handle prettier ignores**

Declaring this as a recommended method is just declaring prettier an incomplete tool.
Prettier is `prettier`, not `eslint-plugin-opinions`, it was built to stand on it's own, and it ought to continue to do so.

**Hack other config options**

This is objectively worse than what this PR enables. Compare;
```jsonc
{
  "overrides": [
    {
      "files": "dist/**/*",
      "options": { "requirePragma": true },
    },
  ],
  // vs
  "overrides": [
   {
     "files": "dist/**/*",
     "options": { "parser": "ignore" }
    }
  ]
}
```
This PR makes the config significantly easier to understand, and as an unnecessary bonus it's a few characters shorter.

**`../../`**

This is what I'm trying out currently in my monorepo. It's truely awful, and because under `packages/` I have some directories for grouping related packages, I constantly end up with an invalid path.

```json
{
  "scripts": {
    "lint": "prettier --ignore-path ../../.prettierignore --check ."
  }
}
```

---

For monorepos, prettier is more trouble than it's worth.

After this PR, prettier will be the obviously easiest option. 

This PR ought to have been merged in 2018.

31i0t:
Also, run into this issue when working within a monorepo 🙋🏽‍♂️. Was really surprised Prettier didn't have this option yet.

With [ESLint](https://eslint.org/docs/latest/use/configure/ignore#ignorepatterns-in-config-files), it's as simple as this:
```json
```

alexander-akait:
The ideal solution is implement `ignorePatterns`, but I don't see any problems with `"parser": "ignore"`

rotu:
> The ideal solution is implement `ignorePatterns`, but I don't see any problems with `"parser": "ignore"`

`ignorePatterns` is necessary for eslint because you can have multiple rulesets and need a way to turn them all off. But a parser precludes other parsers, so you don't *need* it in the same way.

I considered using `"ignore": true` instead of `"parser": "ignore"` but this seems less intrusive and we can always sugar it up later.

WillsterJohnson:
What are the issues blocking this from being accepted?

rotu:
@fisker could you please take a look at this and let me know either (1) what I need to to do in order to move this PR forward or (2) that this PR is unwelcome and should be closed? Thank you.

fisker:
Personally, I really don't like the idea, I'd rather implement `ignorePatterns` mentioned in https://github.com/prettier/prettier/pull/14785#issuecomment-1563454701, but if other maintainers think it's okay to add this special parser, I'll be fine.

One important thing worth to mention, current implementation not really "ignore" the file, it will replace EOL(maybe more than this, not sure). And it will read/write files.

WillsterJohnson:
I agree @fisker, `ignorePatterns` is the ideal solution. But after 5 years, we don't have it.

> I considered using `"ignore": true` instead of `"parser": "ignore"` but this seems less intrusive and we can always sugar it up later.

\- *Rotu, https://github.com/prettier/prettier/pull/14785#issuecomment-1563518522*

Let's merge this PR so we can have an implementation for this feature request, and work on making it perfect later. It's taken over 5 years (#4708 - June 17 2018) to get to this stage, who knows how long it will be before `ignorePatterns` is implemented?

From what you've said, the only blocker is;
- [ ] `ignore` parser should fully ignore files (currently may or may not do other things - needs further investigation)

@rotu given that you implemented and tested this PR I think it's best to come to you for this - what are your thoughts/findings on the above inquiry?

alexander-akait:
> One important thing worth to mention, current implementation not really "ignore" the file, it will replace EOL(maybe more than this, not sure). And it will read/write files.

We need to fix this before the merge, ignore should ignores all

rotu:
The decision to *not* do `ignorePatterns` was on purpose, as it's confusing. If I specify both `ignorePatterns` and a matching pattern with options for some files, it's not clear which should win out (or how multiple matching `prettierrc` files should interact). Additionally, this makes it equally easy to ignore certain files or ignore all *but* certain files.

I changed the approach and it now leaves whitespace alone. BUT I think the test fixture is mangling the data before it gets fed into the test case (i.e. the `input` part of the `jsfmt.spec.js.snap` file is not what I expect). So I may yet need to adjust the tests.

WillsterJohnson:
> The decision to not do `ignorePatterns` was on purpose, as it's confusing [...]

This was something I had considered, but I didn't know how to articulate that concern. I suppose it might work similar to the priority `.prettierignore` has, but providing it as a config option means there may be multiple instances in different contexts, making it difficult to clearly see what is and is not ignored.

Having an ignorant parser is much clearer, and while I stand by agreeing with Fisker that `ignorePatterns` is the ideal solution, figuring out how exactly that ought to work will take time, and deciding to wait for that could mean we wait a total of a decade for prettier to be viable in monorepos.

`ignorePatterns` would be great; we seriously cannot afford to wait until it's well defined to have this feature, let's take this PR (assuming the issue found last week is resolved - which it appears to be) and we can break it in a few years time once `ignorePatterns` has a sensible and clear definition and behavior.

WillsterJohnson:
Any decision on this? Is it going to be merged or are we going to continue with poor monorepo support?

WillsterJohnson:
@fisker since you last commented the concern you rose has been addressed by a new commit. Assuming that commit correctly addresses that concern (which it appears to do), can we have this merged and released in the next minor bump?

fisker:
What do you think we add support for `ignore: boolean` to `overrides`? #15322

fredericrous:
Following use case for ignorePatterns seem to be missing in this discussion:

We have a global prettier config for the whole company. the same way we do for eslint.

our prettier config looks like this:

```js
module.exports = require('@teckro/prettier-config/prettier.config');
```

Today, I cannot have a global config to ignore build/ , public/ and coverage/ folders.
I am required to create a .prettierignore file per repo. If I want to do a global change to the list of ignored folders, I need to edit each .prettierignore file

WillsterJohnson:
An issue with this was raised in #15322 which I actually agree with (https://github.com/prettier/prettier/pull/15322#issuecomment-1720468761)

To be clear;

### I still very much think `ignore: true` is an anti-pattern we should avoid

However I can't ignore that `"parser": "ignore"` could actually be a name conflict, or at the very least be misleading to a degree. I think `"parser": false` resolves this quite well, and would make this PR an absolute no-brainer to merge if it somehow wash't already the better of the two.

rotu:
I’m rather fond of `parser: ignore` because 1) it’s more discoverable/searchable 2) it’s less magical (even if we were to admit a `false` literal in the config file, I’d probably want to desugar it internally since each parser has a unique string identifier.

I say ship it. I just care that this functionality exists, not how you spell the config.

rotu:
@fisker What needs to happen to get this merged in?

rotu:
Merged in master, freshened up the description.

@fisker, @sosukesuzuki Could we please get this merged in?

PR #14787, opened 2023-05-02
fix incorrect formatting of attribute selectors with url protocols

romainmenke:
## Description

Attribute selectors with url protocols are incorrectly formatted :

```css
a[href^='http://example.com'] {
}

/* becomes */

a[href^='http://example.com']
{
}
```

## Checklist

- [x] ~~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~~

### DIFF
--- a/tests/format/css/attribute/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/css/attribute/__snapshots__/jsfmt.spec.js.snap
@@ -280,6 +280,10 @@ href
 *=
 'example'
 ] {}
+a[ href ^= 'http://example.com' ] {}
+a[ href $= 'https://example.com' ] {}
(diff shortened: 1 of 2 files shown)


src/language-css/parse/parse-selector.js
```diff
@@ -6,7 +6,7 @@ function parseSelector(selector) {
   // the content of the comment as selectors which turns it into complete
   // garbage. Better to print the whole selector as-is and not try to parse
   // and reformat it.
-  if (/\/\/|\/\*/.test(selector)) {
+  if (/^\/\/|[^:]\/\/|\/\*/.test(selector)) {
```
romainmenke, line 9: `//` at the start or `//` when the it doesn't follow `:`

PR #14812, opened 2023-05-10
Improve cursor tracking

### LINKED ISSUE #14811: Wrong cursorOffset with multi-line comment
**Environments:**

- Prettier Version: 2.8.8
- Running Prettier via: Node.js API
- Runtime: Node.js 18.16.0
- Operating System: macOS
- Prettier plugins (if any): 

**Steps to reproduce:**

Save this script and run it:

```js
import prettier from "prettier";

const code = `
// All messages are represented in JSON.
// So, the prettierd.py controls a subprocess which spawns "node {this_file}".
import {  } from "fs"
`; //   ^ cursor here

console.log(
  prettier.formatWithCursor(code, {
    cursorOffset: 129,
    parser: "babel",
  })
); // => { cursorOffset: 65 }
```

**Expected behavior:**

Expect the returned `cursorOffset` to be around 129, right in the `{}`.

**Actual behavior:**

It returns 65, which is in the comments.

Frankly, I don't even find a minimal case to reproduce this issue (the code above is all I could make, changing any line of the comment may make the result correct).

fisker:
## Description

~Attempt to fix #14811, didn't run test locally, not sure about the result.~

Fixes #14811

## Checklist


fisker:
Though there are unsolved questions left in #14811, this can still merge since leading comments doesn't belong to node.

PR #14829, opened 2023-05-14
Boolean checks should not be inverted

cesarkohl:
Hey there, I've found this small issue.

By the way, this is a great project! I've been using it since its beginning :)

## Description

It is needlessly complex to invert the result of a boolean comparison. The opposite comparison should be made instead.

In other words, instead of `!(a === b)` use `a !== b`.

## Checklist


fisker:
Thanks!

PR #14836, opened 2023-05-19
Consider Flow type cast expressions when expanding call arguments

pieterv:
## Description

Comparing formatting of prettier v2 vs prettier v3 on the Meta codebases i noticed argument expansion no longer works when the second arg is a Flow type cast. My guess is it was related to the changes in #13341 which explicitly handles TS `TSAsExpression` nodes. This change adds similar support for Flow `TypeCastExpression` nodes.

```js
// Input
const bar1 = [1,2,3].reduce((carry, value) => {
  return [...carry, value];
}, ([]: Array<string>));
// Prettier stable
const bar1 = [1,2,3].reduce(
  (carry, value) => {
    return [...carry, value];
  },
  ([]: Array<string>)
);
// Prettier main
const bar1 = [1,2,3].reduce((carry, value) => {
  return [...carry, value];
}, ([]: Array<string>));
```

This change doesn't bring full parity to the current v2.8 logic but it should cover most of the cases and at least match TS behaviour. 

## Checklist

### DIFF
--- a/src/language-js/print/call-arguments.js
+++ b/src/language-js/print/call-arguments.js
@@ -281,8 +281,12 @@ function isHopefullyShortCallArgument(node) {
     return isHopefullyShortCallArgument(node.expression);
   }
 
-  if (isTSTypeExpression(node)) {
+  if (isTSTypeExpression(node) || node.type === "TypeCastExpression") {
     let { typeAnnotation } = node;
+    if (typeAnnotation.type === "TypeAnnotation") {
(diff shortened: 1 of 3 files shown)


pieterv:
Do you want a changelog entry for this change? Its not really an added feature over the last release, just improving the #13341 change.

fisker (approved):

fisker:
@pieterv Tests are failing

thorn0 (approved):

PR #14982, opened 2023-06-25
add git-diff option

amidkoruzam:
## Description

Add --git-diff option to show git diff. Can only be used with --check.
Nothing fancy but I think thats what people at [#6885](https://github.com/prettier/prettier/issues/6885) have been requesting?

I can add tests/docs, is it an alright solution to the problem?

Screenshot 2023-06-25 at 22 44 12

## Checklist

### DIFF
--- a/src/cli/cli-options.evaluate.js
+++ b/src/cli/cli-options.evaluate.js
@@ -157,6 +157,12 @@ const options = {
     default: 0,
     type: "int",
   },
+  gitDiff: {
+    default: false,
+    description:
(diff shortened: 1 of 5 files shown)


sosukesuzuki (commented):
Sorry. I don't think this feature is necessary, please put it in the PR description.

amidkoruzam:
Hey, thanks for the reply. 
I have edited the description with 'feature is not necessary' if I got you right.

muescha:
If git is not installed or if an error occurs, I recommend adding a simple JavaScript diff for this situation.

Issue #14261, opened 2023-01-29
Handlebars / Glimmer: reformatted whitespace inside `<pre>` tags


CvX:
**Prettier 2.8.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeADgJzgPgDpQAEhYAJoQH4HEA2AztYXGABYSF4gtwCenBqAPRZcIADQgI6GAEtodZKACGmTBADuABRUIFKJTXVKeCiQCNMSsAGs4MAMpKAtnAAyMqHGQAzA3TjmljZ29uhWHgDmyDCYAK4BIHBOZnCkpKmuSlARsUoRcABiEJhOSjCy2cggSrEwEOJcME40AOosMvB0YWBw9rodMgBuHTxVYHSmIB7+mDCalhGlPn4JAFZ0AB72kTRwAIqxEPDL9AlhmDNVETQyTi6YDVgeMC0ypDAsyAAcAAwSWBB-C1LOgqiIZoMvBIAI6HeDzKR6ap0AC0nlSqQa2FhMmw8zySyQvlOEn8Thk0TiCToOzgAEFypgZGZanBNHBMO5PCd-KTaQcjl4iSsJDAlGZXu9PkgAEyiywyG7ZADCEDuSiqcDoAFYGrF-AAVcV6Ym8kCDeIASSg6Vg9jATOkdJt9hgPF2PLgAF8vUA)

```sh
--parser glimmer
```

**Input:**

```hbs
<pre>
  cd ~
  ls
  echo "hey"
</pre>
```

**Output:**

```hbs
<pre>
  cd ~ ls echo "hey"
</pre>
```

**Expected behavior:**
It should not change whitespace inside `<pre>` tags

```hbs
<pre>
  cd ~
  ls
  echo "hey"
</pre>
```

(this issue does not affect the html parser)

ef4:
The same goes for `<style>`, which should either be left alone or (ideally) delegate to the CSS rules for prettier.

jurgenwerk:
It looks like the glimmer printer has a flaw where it doesn't check whether the node is whitespace sensitive, or indentation sensitive. In case of html parser, it checks for these things before trying to apply formatting. I'm checking to see if I can submit a fix for that.

jurgenwerk:
Fixed by https://github.com/prettier/prettier/pull/15087/files

jurgenwerk:
[...]

PR #15068, opened 2023-07-10
Glimmer: don't format <pre> and <style> tags

jurgenwerk:
## Description

This PR fixes the bug described in issue https://github.com/prettier/prettier/issues/14261

As mentioned in the issue, In addition to `<pre>`, `<style>` is also causing problems - the printer is collapsing CSS code into one line.

Ideally, we should also properly format the `<style>` content but I suppose that would have to be implemented - I was looking at how html printer is formatting that and I don't see an easy way of reusing that here. 

In hopes of finding support for formatting CSS in the glimmer printer was also checking the printer from the glimmer source code that this printer was based on (https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/syntax/lib/generation/printer.ts) but it looks like it was rewritten entirely (in an incompatible manner). 

So for now I think ignoring the `<style>` content is still an improvement over the current behaviour. 

## Checklist

### DIFF
--- a/changelog_unreleased/handlebars/15068.md
+++ b/changelog_unreleased/handlebars/15068.md
@@ -1,0 +1,43 @@
+#### Glimmer: don't format <pre> and <style> tags (#15068 by @jurgenwerk)
+
+<!-- prettier-ignore -->
+```handlebars
+// Input
+<pre>
+  cd ~
+  ls
+  echo "hey"
+</pre>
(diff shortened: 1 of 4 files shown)


jurgenwerk:
Closing this PR in favor of a better one that is coming, with proper CSS formatting support.

Issue #12683, opened 2022-04-22
Missing `preprocess` from `Printer` types


siefkenj:
The documentation states that in a `Printer` plugin, a `preprocess` attribute is allowed. However, the 2.6.0 types from `@types/prettier` do not list a `preprocess` type for the `Printer` type.

Harry-Hopkinson:
Where in the documents is this stated?

Harry-Hopkinson:
@siefkenj ah sorry, I can work on this.

marcgibbons:
@siefkenj Opened a PR here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65635 to add the `preprocess` method to the `Printer` type.

PR #15123, opened 2023-07-17
fix: add missing `printer.preprocess` definition

so1ve:
## Description

> Also, the preprocess method of a printer can return a promise now.

In the blog, preprocess in allowed to return a promise now. However the types are wrong.

I guess the blog is wrong too because preprocess belongs to parser?

Maybe address #12683

(I'm unable to provide a changelog now but soon)

## Checklist

### DIFF
--- a/changelog_unreleased/api/15123.md
+++ b/changelog_unreleased/api/15123.md
@@ -1,0 +1,3 @@
+#### fix: add missing `printer.preprocess` definition (#15123 by @so1ve)
+
+Adds missing types for `printer.preprocess`.
--- a/changelog_unreleased/api/15123.md
+++ b/changelog_unreleased/api/15123.md
@@ -5,7 +5,7 @@ Adds missing types for `printer.preprocess`.
 ```diff
 export interface Printer<T = any> {
 +  preprocess?:
-    | ((ast: T, options: ParserOptions<T>) => T | Promise<T>)
-    | undefined;
++    | ((ast: T, options: ParserOptions<T>) => T | Promise<T>)
++    | undefined;
 }
 ```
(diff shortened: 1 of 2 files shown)


so1ve:
It doesn't look right because the document says printer has a `preprocess` method but the type doesn't include it.

fisker:
The blog post is correct. We only support `printer.preprocss` to be async. https://github.com/prettier/prettier/blob/24dfe5afe768b627f09484426b26fca764251225/src/main/ast-to-doc.js#L141

so1ve:
@fisker Yeah I agree that the blog post is correct, however the types are wrong. (This method is not in the type!)

And I think async preprocess should be supported in parser too. Just add an `await` keyword.

fisker:
So, there are two separate things to do.

1. Fix types for `printer.preprocss`. (You can do it in this PR)
2. Request `parser.preprocss` to be async.  (Please raise an issue to discuss first)

src/index.d.ts
```diff
@@ -485,6 +485,9 @@ export interface Printer<T = any> {
         | Doc
         | null)
     | undefined;
+  preprocess?:
+    | ((ast: T, options: ParserOptions<T>) => T | Promise<T>)
```
fisker, line 489: The second parameter should be `Options` lile other methods.

fisker: Oh, sorry, print has same options. should be correct.

so1ve:
It's ready for review now!

changelog_unreleased/api/15123.md
```diff
@@ -0,0 +1,3 @@
+#### fix: add missing `printer.preprocess` definition (#15123 by @so1ve)
```
fisker, line 1:
```suggestion
#### Add missing type definition for `printer.preprocess` (#15123 by @so1ve)
```

changelog_unreleased/api/15123.md
```diff
@@ -0,0 +1,3 @@
...
+
+Adds missing types for `printer.preprocess`.
```
fisker, line 3: We can just put type defination here. Something like:

````text
```diff
export interface Printer<T = any> {
+  preprocss
}
```
````

fisker (approved):

sosukesuzuki (approved):

PR #15130, opened 2023-07-18
Support `as` and `satisfies` expressions for Flow

gkz:
## Description

Adds support for  printing `as` and `satisfies` expressions in Flow.
```
const x = y as T;
```

This is accomplished first by factoring out TypeScript's logic and using/extending that.

Tests are mostly copied from TS's tests. I did a `diff` of Flow's and TS's snapshot file to confirm that that differences were expected (from input change).

## Checklist

### DIFF
--- a/src/language-js/print/flow.js
+++ b/src/language-js/print/flow.js
@@ -10,6 +10,7 @@ import {
   rawText,
 } from "../utils/index.js";
 import isFlowKeywordType from "../utils/is-flow-keyword-type.js";
+import { printBinaryCastExpression } from "./cast-expression.js";
 import { printClass } from "./class.js";
 import {
   printOpaqueType,
@@ -276,6 +277,11 @@ function printFlow(path, options, print) {
           ? ["(", print("value"), ")"]
           : []),
       ];
+
+    case "AsExpression":
+    case "AsConstExpression":
+    case "SatisfiesExpression":
+      return printBinaryCastExpression(path, options, print);
   }
 }
 
--- a/src/language-js/utils/index.js
+++ b/src/language-js/utils/index.js
@@ -55,7 +55,7 @@ function hasNakedLeftSide(node) {
     node.type === "TaggedTemplateExpression" ||
     node.type === "BindExpression" ||
     (node.type === "UpdateExpression" && !node.prefix) ||
-    isTSTypeExpression(node) ||
+    isBinaryCastExpression(node) ||
     node.type === "TSNonNullExpression" ||
     node.type === "ChainExpression"
   );
@@ -795,6 +795,9 @@ function startsWithNoLookaheadToken(node, predicate) {
     case "TSSatisfiesExpression":
     case "TSAsExpression":
     case "TSNonNullExpression":
+    case "AsExpression":
+    case "AsConstExpression":
+    case "SatisfiesExpression":
       return startsWithNoLookaheadToken(node.expression, predicate);
     default:
       return predicate(node);
@@ -1101,9 +1104,14 @@ const markerForIfWithoutBlockAndSameLineComment = Symbol(
   "ifWithoutBlockAndSameLineComment",
 );
 
-const isTSTypeExpression = createTypeCheckFunction([
+const isBinaryCastExpression = createTypeCheckFunction([
+  // TS
   "TSAsExpression",
   "TSSatisfiesExpression",
+  // Flow
+  "AsExpression",
+  "AsConstExpression",
+  "SatisfiesExpression",
 ]);
 
 export {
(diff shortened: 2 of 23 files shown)


gkz:
Does anyone have an idea why "Prod / Node.js 18 on ubuntu-latest (Full Test) (pull_request)" (and only that) is failing? 
Running the tests locally passes

thorn0:
The names `isTypeExpression` and `printCastExpression` are inconsistent and also both aren't really good names. How about `isBinaryCastExpression`/`printBinaryCastExpression`?

gkz:
> The names `isTypeExpression` and `printCastExpression` are inconsistent and also both aren't really good names. How about `isBinaryCastExpression`/`printBinaryCastExpression`?

All casts are binary, but you're right it should be `isCastExpression` & `printCastExpression` - will update

fisker:
But there is already a node type called `TypeCastExpression`. https://github.com/prettier/prettier/blob/24dfe5afe768b627f09484426b26fca764251225/src/language-js/print/flow.js#L248

gkz:
> But there is already a node type called `TypeCastExpression`.
> 
> https://github.com/prettier/prettier/blob/24dfe5afe768b627f09484426b26fca764251225/src/language-js/print/flow.js#L248

You're right, I can call it `isBinaryCastExpression `. The other one requires parens so isn't like a binary expression.

src/language-js/print/flow.js
```diff
@@ -276,6 +277,18 @@ function printFlow(path, options, print) {
...
+
+    case "AsExpression":
+    case "AsConstExpression":
+    case "SatisfiesExpression": {
+      const operator =
+        node.type === "AsExpression"
+          ? "as"
+          : node.type === "AsConstExpression"
+          ? "as-const"
+          : "satisfies";
```
fisker, line 289: Let's get `operator` in `printBinaryCastExpression`.

fisker: So we can use `node.type === "AsExpression" || node.type === "TSAsExpression"`.

src/language-js/utils/index.js
```diff
@@ -1101,9 +1104,14 @@ const markerForIfWithoutBlockAndSameLineComment = Symbol(
...
 );
 
-const isTSTypeExpression = createTypeCheckFunction([
+const isBinaryCastExpression = createTypeCheckFunction([
+  // TS
   "TSAsExpression",
   "TSSatisfiesExpression",
+  // Flow
+  "AsExpression",
+  "AsConstExpression",
```
fisker, line 1113: I'm not familiar with flow AST, any reason why we are not parsing it as `AsExpression` like TS? 
There should be edge cases format differently between Flow and TS.

gkz: `x as const` is really just a one argument (`x`) operation. TS does something weird where they parse it as a binary `as` expression where the right hand side is a type identifier `const`, but then they match on it and ignore it?

gkz: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

fisker: Make sense.

fisker: `let x = "hello" as /* comment */ const;` format different. 😄 

Allow comment in "operator" also wired to me...

tests/format/flow/as-satisfies-expression/jsfmt.spec.js
```diff
@@ -0,0 +1 @@
+run_spec(import.meta, ["flow"], { errors: { "babel-flow": true } });
```
fisker, line 1: If tests are both valid in TS and Flow  we can  add `typescript` parser here, so we can make them print consistently if possible.  If code not valid in TS, please ignore.

src/language-js/utils/index.js
gkz: Yeah I think it's weird that TS supports a comment in between `as` and `const` - it's just a side effect of how they decided to parse this

tests/format/flow/as-satisfies-expression/jsfmt.spec.js
gkz: The tests in general are not valid for both TS and Flow. We still have a variety of syntactic differences (though over time, it's getting a bit closer as you see).

gkz:
Sorry I went on vacation and when I got back started working on other stuff.
I've updated the PR now as per your latest comment.

src/language-js/print/flow.js
fisker: I see thumbs up here, but code didn't change.

fisker:
https://github.com/prettier/prettier/pull/15130#discussion_r1270982236

src/language-js/print/flow.js
gkz: Changes should be amended and force pushed now

gkz:
I'm used to using Mercurial every day so forgot to add the files before amending!

fisker (approved):

fisker:
Do we need a release after merge, or we can wait?

gkz:
> Do we need a release after merge, or we can wait?

I'm not in a rush with this

sosukesuzuki (approved):

PR #15134, opened 2023-07-20
Update docs to mention gitignore changes

moiri-gamboni:
## Description

Updates docs to mention that prettier follows .gitignore in the CWD.

## Checklist


awsr:
Any chance of getting this merged? I spent hours trying to figure out why Prettier wasn't working only to finally stumble upon the answer on my own because searching the documentation for ignoring code showed _nothing_ about it being used by default. It only gets mentioned elsewhere under the description for the `--ignore-path` path option (which I wasn't using, so why would I look there?), and even _that_ **doesn't show up in the results** when searching for "gitignore" in the documentation.

sosukesuzuki (approved):

docs/ignore.md
```diff
@@ -24,7 +24,7 @@ coverage
 
 It’s recommended to have a `.prettierignore` in your project! This way you can run `prettier --write .` to make sure that everything is formatted (without mangling files you don’t want, or choking on generated files). And – your editor will know which files _not_ to format!
 
-By default prettier ignores files in version control systems directories (".git", ".svn" and ".hg") and `node_modules` (if [`--with-node-modules` CLI option](cli.md#--with-node-modules) not specified)
+By default prettier ignores files in version control systems directories (".git", ".svn" and ".hg") and `node_modules` (if [`--with-node-modules` CLI option](cli.md#--with-node-modules) not specified). Prettier will also follow rules specified in .gitignore if it exists in the same directory from which it is run.
```
fisker, line 27: We also ignores `.sl`  https://github.com/prettier/prettier/blob/56945ecf12a5b2e3b70e13bbcec956ccaf7bd460/src/cli/expand-patterns.js#L47

fisker: If we are going to mention `.gitignore` here, maybe we should put `.prettierignore` here too.

fisker, line 27:
```suggestion
By default prettier ignores files in version control systems directories (".git", ".sl", ".svn", and ".hg") and `node_modules` (unless [`--with-node-modules` CLI option](cli.md#--with-node-modules) is specified). Prettier will also follow rules specified in ".gitignore" file if it exists in the same directory from which it is run.
```

awsr, line 27: Slightly better wording would probably be:

```suggestion
By default prettier ignores files in version control systems directories (".git", ".sl", ".svn" and ".hg") and `node_modules` (unless the [`--with-node-modules` CLI option](cli.md#--with-node-modules) is specified). Prettier will also follow rules specified in the ".gitignore" file if it exists in the same directory from which it is run.
```

fisker (approved):

fisker:
Hello @moiri-gamboni Can you copy change to files in `website/versioned_docs/version-stable/`? So the website can be updated when we merge.

PR #15157, opened 2023-07-24
Update Babel to v8

nicolo-ribaudo:
## Description

Let's see if there are any big problems :)

I'm also working on re-enabling prettier tests in Babel's CI for Babel 8, which we had to disable when moving to native ESM. This is _veary early_, so at least until when we them working do not even think about merging this PR 😛 

These are the breaking changes in `@babel/parser`:
  - Dropped support for the `moduleAttributes` plugin (prettier does not use it)
  - AST nodes do not have a `__clone` method anymore (prettier does not use it)
  - When the `estree` plugin is enabled, we now generate ESTree-compliant AST for class fields (**prettier already supports proper estree**)
  - In AST nodes `TSCallSignatureDeclaration`, `TSConstructSignatureDeclaration`, `TSFunctionType`, `TSConstructorType` and `TSMethodSignature` we renamed `parameters` to `params` and `typeAnnotation` to `returnType` **(prettier already supports the new names)**
  - Tokens changes: **(this might affect prettier?)**
  	- `#foo` is a single `tt.privateName` token instead of `#` + identifier
  	- `` `start `` opening template literals is now a single token instead of two
  - New recoverable errors: **(this might affect prettier?)**
  	- Disallow `>`/`}` in JSX text, which is invalid according to the JSX spec (`<div>></div>`)
  	- Disallow unparenthesized SequenceExpression in JSX (`<div>{a,b}</div>`)

Breaking changes in the other packages do not affect prettier, other than:
- we ship Babel as native ESM (but Prettier bundles Babel, so it doesn't matter)
- we dropped support for Node <16 **(prettier compiles babel in its released bundle so this is not a problem for the published version. However, prettier uses babel to run some tests in Node.js 14 and thus we need to compile babel on-the-fly)**

Fill changelog: https://next.babeljs.io/docs/v8-migration-api

## Checklist


nicolo-ribaudo:
It looks like CI is failing on chalk, it might be fixed by https://github.com/babel/babel/pull/15814.

nicolo-ribaudo:
Yep, the problem comes from `@babel/code-frame`.

However, will have to update `parser`/`types`/`generator` at the same time, because of the AST changes.

fisker:
The dev version passed, means the build script need update, 

> TypeError: Cannot read properties of undefined (reading 'platform')

Should be `process.platform` , caused by https://github.com/prettier/prettier/blob/6ed37ad85731e4e62d81481d40dfd98c6a675a10/scripts/build/build-javascript-module.js#L116

fisker:
This package size increased https://github.com/prettier/prettier/actions/runs/5648163263/job/15299782481?pr=15157#step:4:289

This replacement may not working.

https://github.com/prettier/prettier/blob/6ed37ad85731e4e62d81481d40dfd98c6a675a10/scripts/build/config.js#L397-L400

scripts/build/config.js
```diff
@@ -426,11 +426,21 @@ const nonPluginUniversalFiles = [
...
+        }),
         path: path.join(dirname, "./shims/babel-highlight.js"),
       },
       {
-        module: createRequire(require.resolve("vnopts")).resolve("chalk"),
+        module: require.resolve("chalk", {
+          paths: [require.resolve("@babel/code-frame")]
+        }),
+        path: path.join(dirname, "./shims/chalk.js"),
+      },
```
nicolo-ribaudo, line 439: This change (and the `getChalk` removal in `shims/babel-highlight.js`) are because of https://github.com/babel/babel/pull/15812.

You will need to do the same changes when we release the next Babel 7 version, as `@babel/code-frame` now has a `chalk` dependency.

nicolo-ribaudo:
Would it be possible for Prettier to compile `@babel/parser` down to Node.js 14 when bundling it? We were planning to only support Node.js 16+, but this would be a problem given that prettier supports Node.js 14+.

fisker:
> Would it be possible for Prettier to compile @babel/parser down to Node.js 14 when bundling it?

Yes, we can do when bundling.

> The alternative could be that we publish @babel/parser compiled down to Node.js 14 just for prettier

Not nessary, we can figure how to do it ourself.

fisker:
esbuild should be able to transform `??=`. Maybe our target is wrong? https://github.com/prettier/prettier/blob/825425c57f00268abaf3964b441a7649c8004ba0/scripts/build/build-javascript-module.js#L227


We can also use `supported` option https://esbuild.github.io/api/#supported

nicolo-ribaudo:
Debugging the error, it looks like the problem is that `@babel/code-frame` and `@babel/runtime` are being loaded at runtime by `tests/config/format-test-setup.js`, and they use syntax not supported by Node.js 14 (`??=`).

fisker:
We used bundle modules into `vendors/` and load from there. https://github.com/prettier/prettier/pull/12217

scripts/build/config.js
fisker: [`de94bcc` (#15236)](https://github.com/prettier/prettier/pull/15236/commits/de94bccca752f3e2db07ecd6c470735f9dd8590a)

fisker:
@nicolo-ribaudo 

I think everything is working as expected now. 
Since we are not going to release new version in near future, I think we can merge.  

Questions:

1. Will we release v8 for every change in babel repo?
2. Should we do something for the e2e test in babel repo?

PR #15171, opened 2023-07-29
2548 long long member and method chain

### LINKED ISSUE #2548: Long chaining property?
**Version**

1.5.3

**Code**

([Playground](https://prettier.io/playground/#%7B%22content%22%3A%22this.is.a.long.long.long.long.long.long.long.long.long.long.long.long.long.long.long.long.method()%3B%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D))

```js
this.is.a.long.long.long.long.long.long.long.long.long.long.long.long.long.long.long.long.method();
```

**Expected**

Split it somewhere.

**Actual**

No changes

seiyab:
## Description
Fix #2548.

## Checklist

### DIFF
--- a/src/language-js/print/member-chain.js
+++ b/src/language-js/print/member-chain.js
@@ -24,11 +24,10 @@ import {
   conditionalGroup,
   breakParent,
   label,
-  softline,
 } from "../../document/builders.js";
 import { willBreak } from "../../document/utils.js";
 import printCallArguments from "./call-arguments.js";
-import { printMemberLookup, shouldInlineMember } from "./member.js";
+import { printMemberLookup } from "./member.js";
 import {
   printOptionalToken,
   printFunctionTypeParameters,
@@ -106,19 +105,21 @@ function printMemberChain(path, options, print) {
       });
     } else if (isMemberish(node)) {
       path.call((object) => rec(object), "object");
-      let doc;
-      if (isMemberExpression(node)) {
-        const lookupDoc = printMemberLookup(path, options, print);
-        doc = shouldInlineMember(path, printedNodes.at(-1).printed)
-          ? lookupDoc
-          : group(indent([softline, lookupDoc]));
-      } else {
-        doc = printBindExpressionCallee(path, options, print);
-      }
       printedNodes.push({
         node,
         needsParens: pathNeedsParens(path, options),
-        printed: printComments(path, doc, options),
+        printed: printComments(
+          path,
+          isMemberExpression(node)
+            ? printMemberLookup(
+                path,
+                options,
+                print,
+                printedNodes.at(-1)?.printed,
+              )
+            : printBindExpressionCallee(path, options, print),
+          options,
+        ),
       });
     } else if (node.type === "TSNonNullExpression") {
       path.call((expression) => rec(expression), "expression");
--- a/src/language-js/print/member-chain.js
+++ b/src/language-js/print/member-chain.js
@@ -24,10 +24,11 @@ import {
   conditionalGroup,
   breakParent,
   label,
+  softline,
 } from "../../document/builders.js";
 import { willBreak } from "../../document/utils.js";
 import printCallArguments from "./call-arguments.js";
-import { printMemberLookup } from "./member.js";
+import { printMemberLookup, shouldInlineMember } from "./member.js";
 import {
   printOptionalToken,
   printFunctionTypeParameters,
@@ -87,7 +88,8 @@ function printMemberChain(path, options, print) {
       isCallExpression(node) &&
       (isMemberish(node.callee) || isCallExpression(node.callee))
     ) {
-      printedNodes.unshift({
+      path.call((callee) => rec(callee), "callee");
+      printedNodes.push({
         node,
         printed: [
           printComments(
@@ -102,28 +104,30 @@ function printMemberChain(path, options, print) {
           shouldInsertEmptyLineAfter(node) ? hardline : "",
         ],
       });
-      path.call((callee) => rec(callee), "callee");
     } else if (isMemberish(node)) {
-      printedNodes.unshift({
+      path.call((object) => rec(object), "object");
+      let doc;
+      if (isMemberExpression(node)) {
+        const lookupDoc = printMemberLookup(path, options, print);
+        doc = shouldInlineMember(path, printedNodes.at(-1).printed)
+          ? lookupDoc
+          : group(indent([softline, lookupDoc]));
+      } else {
+        doc = printBindExpressionCallee(path, options, print);
+      }
+      printedNodes.push({
         node,
         needsParens: pathNeedsParens(path, options),
-        printed: printComments(
-          path,
-          isMemberExpression(node)
-            ? printMemberLookup(path, options, print)
-            : printBindExpressionCallee(path, options, print),
-          options,
-        ),
+        printed: printComments(path, doc, options),
       });
-      path.call((object) => rec(object), "object");
     } else if (node.type === "TSNonNullExpression") {
-      printedNodes.unshift({
+      path.call((expression) => rec(expression), "expression");
+      printedNodes.push({
         node,
         printed: printComments(path, "!", options),
       });
-      path.call((expression) => rec(expression), "expression");
     } else {
-      printedNodes.unshift({
+      printedNodes.push({
         node,
         printed: print(),
       });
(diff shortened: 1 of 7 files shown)


src/language-js/print/member-chain.js
```diff
@@ -102,28 +103,32 @@ function printMemberChain(path, options, print) {
           shouldInsertEmptyLineAfter(node) ? hardline : "",
         ],
       });
-      path.call((callee) => rec(callee), "callee");
     } else if (isMemberish(node)) {
-      printedNodes.unshift({
+      path.call((object) => rec(object), "object");
+      printedNodes.push({
```
seiyab, line 108: I reversed the order:
- before: `unshift()` current property -> recursive call for object
- after: recursive call for object -> `push()` current property

This is because I need the doc for the object to print member lookup.
I think the performance and readability might be worried about. My thoughts are following, though I haven't benchmarked.
- Performance
  - Tail call optimization doesn't apply for both before and after.
  - `push()` should be faster than `unshift()`
- Readability
  - Former one goes right to left and latter one goes left to right. I feel it get more readable.

src/language-js/print/member-chain.js
```diff
@@ -102,28 +103,32 @@ function printMemberChain(path, options, print) {
...
         needsParens: pathNeedsParens(path, options),
         printed: printComments(
           path,
           isMemberExpression(node)
-            ? printMemberLookup(path, options, print)
+            ? printMemberLookup(
+                path,
+                options,
+                print,
+                printedNodes.at(-1)?.printed,
```
seiyab, line 118: I found I'm misunderstanding something here, though it might works fine.

tests/format/js/arrow-call/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -255,26 +249,23 @@ const testResults = results.testResults.map((testResult) =>
...
 
 expect(() =>
   asyncRequest({ url: "/test-endpoint-but-with-a-suuuuuuuuper-long-url" })
 ).toThrowError(/Required parameter/);
 
-expect(() =>
-  asyncRequest({ type: "foo", url: "/test-endpoint" })
-).not.toThrowError();
+expect(() => asyncRequest({ type: "foo", url: "/test-endpoint" })).not
+  .toThrowError();
```
seiyab, line 268: It looks to get worse to read.
In my opinion, this is an unfortunate edge case.
We need better heuristics, if this can't be accepted.

src/language-js/print/member.js
```diff
@@ -6,16 +6,65 @@ import {
...
+      return [optional, "[", property, "]"];
+    }
+
+    return group([optional, "[", indent([softline, property]), softline, "]"]);
+  }
+}
+
+/**
+ * @param {AstPath} path
+ * @param {*} objectDoc
```
seiyab, line 57: I want to put "Doc" type here but it requires too redundant check like:
```diff
- objectDoc.label?.memberChain
+ typeof objectDoc === "object" && "label" in objectDoc &&  typeof objectDoc.label === "object" &&  "memberChain" in objectDoc.label && objectDoc.label?.memberChain
```
Putting `@property {*} [label]` for `Doc` (it should be right), it get a bit better:
```diff
- objectDoc.label?.memberChain
+ typeof objectDoc === "object" && "label" in objectDoc &&  objectDoc.label?.memberChain
```
But I think this might still make developers to avoid typing.
Then, defining `@typedef {Partial<DocObject> & (string | DocObject | DocArray)} LooseDoc`, or just define `Doc` as it make just `objectDoc.label?.memberChain` works.
I'd like to hear member's opinion

src/language-js/print/member-chain.js
seiyab: I reverted this change because objectDoc doesn't look necessary here.

src/language-js/print/member-chain.js
seiyab: Outdated. Probably resolved.

seiyab:
Maybe I accidentally hit another independent issue about idempotency...
```diff
-   .Even// comment<LF>
+   .Even // comment<LF>
```

src/language-js/print/member.js
seiyab: I mean that I will open a discussion or an issue if it is worth to discuss, or open a PR if it looks better than the current one.

tests/format/js/arrow-call/__snapshots__/jsfmt.spec.js.snap
fisker: We suppose to use installed v3.0.1 not the dev version to format, something is wrong.

seiyab:
I consider it ready for review.
(Just for a reminder. I'm patiently waiting for review.)

fisker:
Sorry for the long delay, I'll take a look ASAP.

tests/format/js/method-chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,87 @@
...
+So.why.choose.the.Prettier./* comment */ style.guide /* comment */.over.any
+  .other.random.style
+  .guide?.()
+  .Because.Prettier.is./* comment */ the.only./* comment */ style.guide.that.is
+    .fully.automatic()
+  // commment
+  .Even // comment
+  .if.Prettier.does.not.format.all.code["100%"].the.way["you’d"].like["it’s"]
+    .worth.the.sacrifice.given.the.unique.benefits.of.Prettier["don’t"].you
+    .think?.();
```
fisker, line 80: Can you find some real world case? These tests doesn't make sense to me at all. Nobody write code like this.

tests/format/js/method-chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,87 @@
...
+    "CSS, Less and SCSS",
+    "HTML",
+    "Ember/Handlebars",
+    "JSON",
+    "GraphQL",
+    "Markdown, including GFM and MDX",
+    "YAML",
+  )
+  .it.removes.all.original.styling.and.ensures.that.all.outputted.code.conforms
+    .to.a.consistent.style(see, this, blogpost);
```
fisker, line 60: This indention seems unnessary.

tests/format/js/method-chain/__snapshots__/jsfmt.spec.js.snap
seiyab: I see. I'll consider a more practical test case. This is just a test case that is so long chain and not likely to have a copyright / license issue.

tests/format/js/method-chain/__snapshots__/jsfmt.spec.js.snap
seiyab: I'll fix it.

seiyab:
Finally it looks to "work".
But I'm not sure whether the introduced complexity is reasonable to fix the bug.
I intend to look for simpler way. This might be more costly than leaving the bug existing.

seiyab:
I give up to make it easier to read.
I consider it's ready for review.

src/language-js/print/member-chain.js
```diff
@@ -106,7 +107,9 @@ function printMemberChain(path, options, print) {
     } else if (isMemberish(node)) {
       printedNodes.unshift({
         node,
+        path,
```
fisker, line 110: `ASTPath` is a state machine, it make no sense to add to the array.

seiyab: Thank you, you're right.
I added it once and it didn't work. Then added `shouldInline` instead and forgot to remove it.

fisker:
I was looking at prs send by you, and found this, sorry that we lost track.
The new output doesn't look so good. But I'm not sure what we can do. We'll proberly just ignore such edge cases.

seiyab:
I agree that this change is not worth applying enough. Probably I had encountered a practical problematic case to be improved as I worked on it. But I completely forget it. Closing this PR 👍

PR #15204, opened 2023-08-03
Break after equal on assignment if rhs is await/yield expression

### LINKED ISSUE #15208: line breaks after assignment instead of between parentheses / around parameters
### Discussed in https://github.com/prettier/prettier/discussions/15196

<div type='discussions-op-text'>

<sup>Originally posted by **macgyver** August  1, 2023</sup>
Just wondering if anybody can explain the formatting philosophy of converting a line like 
```js
  const { section, rubric, authors, tags } = await utils.upsertCommonData(mainData);
```

into this:
```js
  const { section, rubric, authors, tags } = await utils.upsertCommonData(
    mainData
  );
```

vs. this:
```js
  const { section, rubric, authors, tags } =
    await utils.upsertCommonData(mainData);
```

or even
```js
  const { section, rubric, authors, tags }
    = await utils.upsertCommonData(mainData);
```
(though I recognize that [putting operators at the start of a line](https://github.com/prettier/prettier/issues/3806) is controversial)

it seems like keeping the code on two lines vs. three is more parsimonious in terms of vertical space, and groups the concepts better by dividing the code neatly into variable names, and assigned values. Having  `mainData` dangling by itself at the end there just doesn't read that well (to me at least).

What do other folks think? Is there a bigger concept at play here that prevents this from being the behavior? Thanks for your time and insights.  

My guess is that Prettier is trying to produce the longest line it can without exceeding the maximum line length, but in cases like this my feeling is that balancing the length of the lines it produces or trying to reduce the total number of lines produced would be a _prettier_ result.</div>

seiyab:
## Description
Works on https://github.com/prettier/prettier/discussions/15196.

Feel free to stop this if it needs more discussions.

Fixes  #15208

## Checklist


tests/format/js/assignment/await-discussion-15196.js
```diff
@@ -0,0 +1,9 @@
+async function f() {
```
seiyab, line 1: This is because Flow yields error on top-level await.

src/language-js/print/assignment.js
```diff
@@ -205,7 +205,7 @@ function shouldBreakAfterOperator(path, options, print, hasShortKey) {
   let node = rightNode;
   const propertiesForPath = [];
   for (;;) {
-    if (node.type === "UnaryExpression") {
+    if (node.type === "UnaryExpression" || node.type === "AwaitExpression") {
```
fisker, line 208: If we are doing this, I think we should treat `yield` the sameway, though it's not very common to be on RHS.

seiyab: Ah, I agree with you,thank you.
Maybe `typeof` too？

tests/format/js/assignment/discussion-15196.js
```diff
@@ -0,0 +1,27 @@
...
+  const loooooooooooooooooooooooooong1 = yield looooooooooooooong.looooooooooooooong.loooooong;
+  const loooooooooooooooooooooooooong2 = yield looooooooooooooong.looooooooooooooong.loooooong();
+  const loooooooooooooooooooooooooong3 = yield looooooooooooooooooooooooooooooooooooooooooooog();
+  const loooooooooooooooooooooooooong4 = !(yield looooooooooooooong.looooooooooooooong.loooooong);
+  const loooooooooooooooooooooooooong5 = void !!(yield looooooooooooooong.looooooooooooooong.loooooong);
+
+  const longlonglonglonglonglonglong1 = yield qwertyuiop(asdfghjkl, zxcvbnm, qwertyuiop, asdfghjkl);
+  const longlonglonglonglonglonglong2 = yield { qwertyuiop: 1234567890, asdfghjkl: 1234567890, zxcvbnm: 123456789 };
+
+  const x = yield;
```
fisker, line 26: one test for `yield* ...` please.

fisker (approved):
Looks good to me.

tests/format/js/assignment/discussion-15196.js
seiyab: I have missed it, thank you for feedback.

sosukesuzuki (approved):

aarongoldenthal:
While this change is definitely the better formatting option, changing this types of formatting is a breaking change.

k4rli:
Definitely a breaking change since it can affect much of codebase. Is there a way to use the old way?

seiyab:
There isn't way to avoid this on 3.0.2.
Prettier have usually been changed format as a patch release.
<details><summary>Examples</summary>

**Add parentheses for TypeofTypeAnnotation to improve readability**
```ts
// Input
type A = (typeof node.children)[];

// Prettier 2.8.4
type A = typeof node.children[];

// Prettier 2.8.5
type A = (typeof node.children)[];
```

**Break the LHS of type alias that has complex type parameters**
```ts
// Input
type FieldLayoutWith<
  T extends string,
  S extends unknown = { width: string }
> = {
  type: T;
  code: string;
  size: S;
};

// Prettier 2.3.0
type FieldLayoutWith<T extends string, S extends unknown = { width: string }> =
  {
    type: T;
    code: string;
    size: S;
  };

// Prettier 2.3.1
type FieldLayoutWith<
  T extends string,
  S extends unknown = { width: string }
> = {
  type: T;
  code: string;
  size: S;
};
```

</details>

Personally, I automatically format only changed files to avoid changing huge amount of files by updating Prettier. https://stackoverflow.com/q/47610213/13195622
I hope it helps.

PR #15290, opened 2023-08-24
refactor: use `esbuild-plugins-node-modules-polyfill`

MichaelDeBoey:
This PR replaces `@esbuild-plugins/node-modules-polyfill` with the up-to-date & maintained `esbuild-plugins-node-modules-polyfill`

The `esbuild-plugins` repo itself points towards using `esbuild-plugin-polyfill-node` instead
https://github.com/remorses/esbuild-plugins/blob/373b44902ad3e669f7359c857de09a930ce1ce90/README.md?plain=1#L15-L16

After doing this in the Remix repo (see https://github.com/remix-run/remix/pull/5274), we got quite some new bugs, so we chose to go for @imranbarbhuiya's  `esbuild-plugins-node-modules-polyfill` instead (see https://github.com/remix-run/remix/pull/6562), which is an up-to-date and well maintained alternative

Added benefit is that we won't get the following deprecation warnings when installing `@esbuild-plugins/node-modules-polyfill`:

```sh
npm WARN deprecated rollup-plugin-inject@3.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-inject.
npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
```


fisker:
This plugin looks good, however if I remember correctly we don't really use it to replace nodejs objects now. The plugin is kept just for safe. What we actually need is "a plugin find out what is used" 😄

MichaelDeBoey:
@fisker I don't think it hurts to replace the out-of-date and deprecated package with an up-to-date and well maintained one (for now) though 🤷‍♂️

fisker:
Agree, I'll take a better look later.

PR #15330, opened 2023-08-29
Do not remove cache file

fisker:
## Description

I don't understand, why we remove cache, and only removes when `cacheLocation` not provide.
What it `cacheLocation` happened to be the default location?

## Checklist


sosukesuzuki (commented):
I forgot why I decide to do this.... I think there is a some reason... Please wait.

PR #15357, opened 2023-09-05
Missing attributes in ParserOptions type

Janther:
## Description

found these missing attributes in the ParserOptions type while working on the typescript migration in our plugin.

I based this from `src/main/normalize-format-options.js`

## Checklist

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -441,6 +441,8 @@ export interface ParserOptions<T = any> extends RequiredOptions {
   locStart: (node: T) => number;
   locEnd: (node: T) => number;
   originalText: string;
+  astFormat: string;
+  printer: Printer<T>;
 }
 
 export interface Plugin<T = any> {


src/index.d.ts
```diff
@@ -441,6 +441,8 @@ export interface ParserOptions<T = any> extends RequiredOptions {
   locStart: (node: T) => number;
   locEnd: (node: T) => number;
   originalText: string;
+  astFormat: string;
+  printer: Printer<T>;
```
fisker, line 445: I don't think they should be used by anyone. 
Can you share how they used in plugin?

Janther: We use it to get the [comment printer function](https://github.com/prettier-solidity/prettier-plugin-solidity/blob/924efb60a17076d568e87cc78ff04b8680b1ebed/src/common/printer-helpers.js#L20) but I guess we can just import that directly.
Not really needed now that I look at it.
Up to you if you wanna merge or not.

fisker: Suggest import it directly.

fisker:
These two properties should not be used, since they were added by accident, we may remove them in future.

PR #15367, opened 2023-09-07
Fix empty Prettier config file create command

### LINKED ISSUE #11751: [docs] Add note to installation instructions about PowerShell defaulting to UTF-16
The [installation docs](https://prettier.io/docs/en/install.html) make the following suggestion for creating an empty prettier config file:

```sh
echo {}> .prettierrc.json
```

This is a useful suggestion in most cases, but breaks for people using Windows and PowerShell, which ["uses the Unicode UTF-16LE encoding by default"](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_character_encoding?view=powershell-7.1#character-encoding-in-windows-powershell).

I gave students a lab this week, and asked them to set up prettier. A number of them struggled with errors that they couldn't make sense of--all of them were encoding related, and all of the people affected were on Windows and PowerShell.

It looks like this has been [discussed in the past](https://github.com/prettier/prettier/issues/8815#issuecomment-671222957), but the suggestions then seem to be "convert the file to UTF8."

I wonder if it would be better to suggest another method for doing this (there's nothing sacred about using the `echo` style above) and/or add a note about PowerShell and the need for UTF8 in JSON.

### LINKED ISSUE #13926: Prettier does not support UTF-16LE config files, which is the default in powershell
As I reported in #8815, prettier doesn't support UTF-16LE config files. The documentation still gives instructions that will result in a UTF-16LE file and give users very confusing instructions. I ran into this again, 2 years later while setting up a new project. Powershell is a very common terminal for Windows developers to work in and I'm sure this will result in confusion by others as well.
**Environments:**

- Prettier Version: 2.8.0
- Running Prettier via: `npx prettier --write .`
- Runtime: Powershell inside of vscode
- Operating System: Windows 10
- Prettier plugins (if any): N/A

**Steps to reproduce:**

Follow the installation instructions in a powershell prompt:
- `npm install --save-dev --save-exact prettier`
- `echo {}> .prettierrc.json`  <--- the problematic command
- `npx prettier --write .`

**Expected behavior:**
Prettier recognizes an empty json config file
**Actual behavior:**
You get the hideous error message:
```
[error] Invalid configuration file `.prettierrc.json`: JSON Error in <file path>\.prettierrc.json: 
[error]
[error] > 1 | ��
[error]     | ^
[error]   2 |
[error]   3 |
[error]
[error] Unexpected token "�" (0xFFFD) in JSON at position 0 while parsing "��\r\u0000\n\u0000"
[error]
[error] > 1 | ��
[error]     | ^
[error]   2 |
[error]   3 |
[error]
```

Saving the file with UTF-8 encoding produces the expected result of running with an empty config file.

fisker:
## Description

Tested on `cmd`, `powershell`, work as expected.

```console
node --eval "fs.writeFileSync('.prettierrc','{}\n')" && node -p "JSON.stringify(fs.readFileSync('.prettierrc','utf8'))"
# "{}\n"
```

Fixes #13926
Fixes #11751
Closes #11692

Note to myself: copy to stable version after we decide what to use.

## Checklist

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -29,7 +29,7 @@ pnpm add --save-dev --save-exact prettier
 Then, create an empty config file to let editors and other tools know you are using Prettier:
 
 ```bash
-node -e "fs.writeFileSync('.prettierrc.json', '{}\n')"
+node -e "fs.writeFileSync('.prettierrc.json','{}\n')"
 ```
 
 Next, create a [.prettierignore](ignore.md) file to let the Prettier CLI and editors know which files to _not_ format. Here’s an example:
--- a/docs/install.md
+++ b/docs/install.md
@@ -31,7 +31,7 @@ Then, create an empty config file to let editors and other tools know you are us
 <!--
 Note:
 - `echo "{}" > .prettierrc` would result in `"{}"<SPACE>` on Windows.
-- `echo {}> .prettierrc` would result the file in UTF-16LE encoding.
+- `echo {}> .prettierrc` would result the file in UTF-16LE encoding in PowerShell.
 The below version works in cmd.exe, bash, zsh, fish, powershell.exe.
 -->
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -32,7 +32,7 @@ Then, create an empty config file to let editors and other tools know you are us
 Note:
 - `echo "{}" > .prettierrc` would result in `"{}"<SPACE>` on Windows.
 - `echo {}> .prettierrc` would result the file in UTF-16LE encoding in PowerShell.
-The below version works in cmd.exe, bash, zsh, fish, powershell.exe.
+The below version works in cmd.exe, bash, zsh, fish, PowerShell.
 -->
 
 ```bash
--- a/docs/install.md
+++ b/docs/install.md
@@ -36,7 +36,7 @@ The below version works in cmd.exe, bash, zsh, fish, powershell.exe.
 -->
 
 ```bash
-node -e "fs.writeFileSync('.prettierrc.json','{}\n')"
+node --eval "fs.writeFileSync('.prettierrc','{}\n')"
 ```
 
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -28,10 +28,8 @@ pnpm add --save-dev --save-exact prettier
 
 Then, create an empty config file to let editors and other tools know you are using Prettier:
 
-<!-- Note: `echo "{}" > .prettierrc.json` would result in `"{}"<SPACE>` on Windows. The below version works in cmd.exe, bash, zsh, fish. -->
```
lydell, line 31: Might be useful to keep some kind of comment around in case someone opens a PR to “simplify” this to using `echo` again.

lydell (approved):

docs/install.md
fisker: https://github.com/prettier/prettier/pull/15367/commits/51effd971dea63958eb4c2a062bebcbf1e246548

docs/install.md
```diff
@@ -28,10 +28,15 @@ pnpm add --save-dev --save-exact prettier
...
+<!--
+Note:
+- `echo "{}" > .prettierrc.json` would result in `"{}"<SPACE>` on Windows.
+- `echo {}> .prettierrc.json` would result `.prettierrc.json` file in UTF-16LE encoding.
+The below version works in cmd.exe, bash, zsh, fish, powershell.exe.
+-->
 
 ```bash
-echo {}> .prettierrc.json
+node -e "fs.writeFileSync('.prettierrc.json','{}\n')"
```
fisker, line 39:
```suggestion
node --eval "fs.writeFileSync('.prettierrc.json','{}\n')"
```

?

fisker: Maybe suggest `.prettierrc` filename?

lydell (approved):

sosukesuzuki (approved):

Issue #15369, opened 2023-09-07
Nested SCSS function calls with leading dash are interpreted as unary minus


threema-danilo:
**Prettier 3.0.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAZgVymGBLaAAgFoATCTAIwBs4AKAEgDcBDazOASkOAB0pChVACc4MTMIHM2HQgCpCAJgDc-AL79+AOgDOcWrgjCe-QemgxiOvAC84SEuSq06ZCjXoBGOAFtOnVSg1EAAaEAgAB3xoHWRQFmFhCAB3AAUEhFiUNmSWAE9YsMphFjAAazEAZRYfOAAZPCg4ZHQ2PSKS8qqI0saAc2QYYQ4wvR88QeG4Uf7aAEVMCHgWtumQACsdAA9K2bgFpeakVup2kABHRfhUpIiskBYdYia4UlfQkCGWPGp+gGEID4fCxkA9qNQPtYoH1aABBGBDPCUTDXODCBpNFanNYACxgPmoAHUcXh4DoemA4JVMqS8ExSXlQWAdIUQEwOABJKBvWCVMDCPBRWHcyowPK0LFnCJJPSEkoRUHSuB6YRMZphRoqmA3Fh9YGStY9YQq0E6Zms6WNGCEvCkGA45AADgADGFRJc8KIdXqQcdVmEYCxKDa7Q6kIowpg9AAVINZE5nXyUV5vUh1FjQzC6uAAMSMwIR-VBLBREBAajUQA)

```sh
--parser scss
```

**Input:**

```scss
@function -double($value) {
  @return $value * 2;
}

.selector {
  font-size: -double(-double(1em));
}
```

**Output:**

```scss
@function -double($value) {
  @return $value * 2;
}

.selector {
  font-size: -double(- double(1em));
}

```

**Expected behavior:**

Dashes are valid parts of SCSS function names. You can test this by pasting the following snippet at https://www.sassmeister.com/:

```scss
@function -double($value) {
  @return $value * 2;
}

.selector {
  font-size: -double(-double(1em));
}
```

However, if a function call is nested, the inner leading dash is interpreted as unary minus, and a space is inserted. This breaks the SCSS function call.

threema-danilo:
[...]

PR #15370, opened 2023-09-08
Do not split call of scss function with leading dash

auvred:
## Description

Fixes https://github.com/prettier/prettier/issues/15369

## Checklist

### DIFF
--- a/src/language-css/print/comma-separated-value-group.js
+++ b/src/language-css/print/comma-separated-value-group.js
@@ -270,7 +270,7 @@ function printCommaSeparatedValueGroup(path, options, print) {
       isMathOperator &&
       iNode.value === "-" &&
       iNextNode.type === "value-func" &&
-      locEnd(iNode) !== locStart(iNextNode)
+      (options.parser !== "scss" || locEnd(iNode) !== locStart(iNextNode))
     ) {
       parts.push(" ");
       continue;
--- a/src/language-css/print/comma-separated-value-group.js
+++ b/src/language-css/print/comma-separated-value-group.js
@@ -269,7 +269,8 @@ function printCommaSeparatedValueGroup(path, options, print) {
       options.parser === "scss" &&
       isMathOperator &&
       iNode.value === "-" &&
-      iNextNode.type === "value-func"
+      iNextNode.type === "value-func" &&
+      locEnd(iNode) !== locStart(iNextNode)
     ) {
       parts.push(" ");
       continue;
--- a/src/language-css/print/comma-separated-value-group.js
+++ b/src/language-css/print/comma-separated-value-group.js
@@ -270,7 +270,7 @@ function printCommaSeparatedValueGroup(path, options, print) {
       isMathOperator &&
       iNode.value === "-" &&
       iNextNode.type === "value-func" &&
-      (options.parser !== "scss" || locEnd(iNode) !== locStart(iNextNode))
+      locEnd(iNode) !== locStart(iNextNode)
     ) {
       parts.push(" ");
       continue;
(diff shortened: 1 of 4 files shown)


src/language-css/print/comma-separated-value-group.js
```diff
@@ -269,7 +269,8 @@ function printCommaSeparatedValueGroup(path, options, print) {
       options.parser === "scss" &&
       isMathOperator &&
       iNode.value === "-" &&
-      iNextNode.type === "value-func"
+      iNextNode.type === "value-func" &&
+      locEnd(iNode) !== locStart(iNextNode)
```
fisker, line 273: Is this `-foo()` syntax only supported by SCSS? I don't think LESS support it. So, we should check `options.parser` here.

auvred: Thanks, updated it

fisker:
@auvred I guess there is no way to tell

```css
width: -double(1em);

// Mean
width: -(double(1em));

// Or
width: (-double)(1em);
```

So I guess we can use your logic to check if there was space there.
Maybe we can remove that part added in #15129?

fisker:
Sorry for not noticing this PR and #15129 are changing the same part. We are good..

src/language-css/print/comma-separated-value-group.js
```diff
@@ -269,7 +269,8 @@ function printCommaSeparatedValueGroup(path, options, print) {
       options.parser === "scss" &&
       isMathOperator &&
       iNode.value === "-" &&
-      iNextNode.type === "value-func"
+      iNextNode.type === "value-func" &&
+      (options.parser !== "scss" || locEnd(iNode) !== locStart(iNextNode))
```
fisker, line 273: Didn't notice the parser already checked above.

auvred: Ohh.. I missed that too

tests/format/scss/scss/scss.scss
```diff
@@ -1090,6 +1090,16 @@ div {
   margin: - pow(2, 2) * 100px;
 }
 
+div {
+  margin: -double(-double(1em));
```
fisker, line 1094:
```suggestion
  margin: -double(-double(1em));
  margin: -double(- double(1em));
  margin: - double(-double(1em));
```

auvred: https://github.com/prettier/prettier/pull/15370/commits/fed17b6174a4efd3e4554a2811880fda38147527

fisker (approved):
LGTM, thank you!

tests/format/scss/scss/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4835,6 +4869,16 @@ div {
...
 }
 
+div {
+  margin: -double(-double(1em));
+  margin: -double(- double(1em));
+  margin: - double(-double(1em));
+}
+
+@function asdf() {
+  @return (width: -double(1em) ;);
```
fisker, line 4879: This looks like a bug to me too.

tests/format/scss/scss/scss.scss
```diff
@@ -1090,6 +1090,18 @@ div {
...
+  margin: -double(- double(1em));
+  margin: - double(-double(1em));
+}
+
+@function asdf() {
+  @return (
+    width: -double(1em);
+  )
+}
```
fisker, line 1103:
```suggestion
@function asdf() {
  @return width: -double(1em);
}
```

auvred: Changed it https://github.com/prettier/prettier/pull/15370/commits/3f9921e94db056236ff5a32c583ce37023ae698d

fisker (approved):

sosukesuzuki (approved):

Issue #16, opened 2017-01-10
Add Vim Support


fingermark:
Hope this is the appropriate place for tracking this.

For anyone interested in adding this, please see these potential references:

https://github.com/Chiel92/vim-autoformat/blob/master/plugin/autoformat.vim

https://github.com/facebook/reason/blob/0883fc938802a028c4e0c07ad8da67faf16625e7/editorSupport/VimReason/plugin/reason.vim

jlongster:
Thanks! Yep here is good for now. I bet somebody will do vim soon. If not I'll do it when I get time.

bspaulding:
Heyo, I use vim and would love to try to tackle this. I'm a bit light on vim script though, does this feel like a good or bad first project? heh.

colbydehart:
[Neoformat](https://github.com/sbdchd/neoformat) has added this functionality today if anyone is wanting to implement prettier in another formatting plugin and needs inspiration/reference.

vramana:
There is note for vim usage in Readme now. So, I am closing this issue.

rauchg:
@vramana you might want to re-open this in view of: https://github.com/jlongster/prettier/issues/743 specially https://github.com/jlongster/prettier/issues/743#issuecomment-280958650

vjeux:
This has been addressed. https://github.com/prettier/prettier#vim

PR #15394, opened 2023-09-14
Move editorconfig-to-prettier package into repo

josephfrazier:
I'd like to vendor/upstream this code because it's been the topic of a handful of issues in its repo over the years, including https://github.com/josephfrazier/editorconfig-to-prettier/issues/9, https://github.com/josephfrazier/editorconfig-to-prettier/issues/21, and most recently https://github.com/josephfrazier/editorconfig-to-prettier/issues/24, and I feel that having it in-repo would be more appropriate (not to mention [the issues in _this_ repo that pertain to editorconfig parsing](https://github.com/search?q=repo%3Aprettier%2Fprettier+editorconfig++&type=issues&state=open)). I've elaborated on my thoughts in https://github.com/josephfrazier/editorconfig-to-prettier/issues/24#issuecomment-1718405985, which I'll quote below:

> hey @Drakmyth, thanks for continuing this conversation! I appreciate all the links you've provided that indicate existing intention and implementations of parsing non-standard/custom properties. It seems like this question might not be as simple as I thought it was previously.
> 
> @fisker, thanks for chiming in as well!
> 
> As you (@fisker) mentioned in [#16 (comment)](https://github.com/josephfrazier/editorconfig-to-prettier/pull/16#issuecomment-1486130040), this package was originally intended for use by Prettier, and while there are some other dependents (according to [github](https://github.com/josephfrazier/editorconfig-to-prettier/network/dependents?dependent_type=PACKAGE) and [npm](https://www.npmjs.com/package/editorconfig-to-prettier?activeTab=dependents)), Prettier is still the vast majority of the usage this package sees.
> 
> Given that, I think it might make sense for this package's code/functionality to move into the Prettier repository, so that any issues relating to it can be discussed where it's used. Personally, I don't use `.editorconfig` with Prettier, so I don't feel especially well-positioned to make decisions on how things should work, especially when [the output of this package is subsequently modified by Prettier](https://github.com/prettier/prettier/blob/7c512651f8814924a042034ff821190c68ef409d/src/config/resolve-editorconfig.js#L13-L16) (I think you had also [mentioned](https://github.com/prettier/prettier/issues/14514#issuecomment-1471056464) this previously, @fisker), indicating that its behavior is already divergent from the intended use case.
> 
> What do y'all think of that proposal? If we have https://github.com/prettier/prettier/ vendor this repository's code, then I can archive this repo and direct people to discuss functionality in the Prettier issues.

This also implements @fisker's suggested change to ES module syntax from https://github.com/josephfrazier/editorconfig-to-prettier/pull/16/, as it's required for lint/test etc to pass.

Once this is merged, I'll likely archive the https://github.com/josephfrazier/editorconfig-to-prettier/ repo after adding a note directing any visitors here instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "diff": "5.1.0",
     "eastasianwidth": "0.2.0",
     "editorconfig": "0.15.3",
-    "editorconfig-to-prettier": "1.0.0",
     "emoji-regex": "10.2.1",
     "escape-string-regexp": "5.0.0",
     "espree": "9.6.1",
--- a/src/config/editorconfig-to-prettier.js
+++ b/src/config/editorconfig-to-prettier.js
@@ -1,0 +1,75 @@
+module.exports = editorConfigToPrettier;
+
+function removeUnset(editorConfig) {
+  const result = {};
+  const keys = Object.keys(editorConfig);
+  for (let i = 0; i < keys.length; i++) {
+    const key = keys[i];
+    if (editorConfig[key] === "unset") {
+      continue;
+    }
+    result[key] = editorConfig[key];
+  }
+  return result;
+}
+
+function editorConfigToPrettier(editorConfig) {
+  if (!editorConfig) {
+    return null;
+  }
+
+  editorConfig = removeUnset(editorConfig);
+
+  if (Object.keys(editorConfig).length === 0) {
+    return null;
+  }
+
+  const result = {};
+
+  if (editorConfig.indent_style) {
+    result.useTabs = editorConfig.indent_style === "tab";
+  }
+
+  if (editorConfig.indent_size === "tab") {
+    result.useTabs = true;
+  }
+
+  if (result.useTabs && editorConfig.tab_width) {
+    result.tabWidth = editorConfig.tab_width;
+  } else if (
+    editorConfig.indent_style === "space" &&
+    editorConfig.indent_size &&
+    editorConfig.indent_size !== "tab"
+  ) {
+    result.tabWidth = editorConfig.indent_size;
+  } else if (editorConfig.tab_width !== undefined) {
+    result.tabWidth = editorConfig.tab_width;
+  }
+
+  if (editorConfig.max_line_length) {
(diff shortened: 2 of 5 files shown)


fisker:
[tests/unit/](https://github.com/prettier/prettier/tree/main/tests/unit) should be a good place for the tests.

josephfrazier:
Thanks @fisker, I was just about to ask about that. The tests currently use `assert` rather than Jest, will that be an issue, or should they "just work"?

fisker (approved):
Thank you for your contribution to Prettier for years!

josephfrazier:
Thanks @fisker, I'm glad to be a part of Prettier! Should I wait for any other reviews (I know you mentioned @sosukesuzuki in https://github.com/josephfrazier/editorconfig-to-prettier/issues/24#issuecomment-1718608567), or are you comfortable with this being merged as-is?

fisker:
I'm fine to merge, the only thing I can image that @sosukesuzuki may concern is that we plan to change the repo to a monorepo https://github.com/prettier/prettier/issues/13912, so he may prefer keep it in a sepeate repo temporally. Let's wait for him for a while.

josephfrazier:
That works for me, but I see that there hasn't been any comments on https://github.com/prettier/prettier/issues/13912 in almost 10 months now. Are you sure we should wait on that before adding a couple hundred lines of code that (I believe) will be needed by Prettier regardless of which parsers are used?

fisker:
I'm not saying we should wait for #13912, we can also transfer https://github.com/josephfrazier/editorconfig-to-prettier/ to prettier org, he may prefer that.

josephfrazier:
Oh, I see, thanks

sosukesuzuki (approved):
Thank you for waiting for me.

Yes, I'm thinking about make Prettier to monorepo, but I don't have time for that yet. And, the code added by this change is sufficiently small and modular.

LGTM.

josephfrazier:
Thanks @sosukesuzuki , merging

Issue #15435, opened 2023-09-20
Unstable formatting for chained calls


so1ve:
**Prettier 3.0.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAxCEB0BDAFASgB0pjMAjA4kAGhAgAcYBLaAZ2VGwCcuIB3AArcE7FNgA2fbAE92tMl2xgA1nBgBlbAFs4AGSZQ4yAGYTWceYpVr19JQYDmyGFwCuFkOa1Nnbj60dxOABFVwh4EzMPACtWAA91QJCwiKRTcXNaAEcUuAFeelEQbFYAWkM4ABMqmhAXbCZxRwBhCC0tbGRi8XFagKgHIIBBGBcmMld4ATgufUNIjI8ACxgtcQB1JaZ4VjswOHURbaYAN23pLrBWORAT9wBJKGrYdTAuJkYhp-UYaSCFzIgei8czrRT0LrAuDmLgnIy0AwwmD5bAODoAjx2Lgwrq-ejQt4fGC1YEGGDrJiVGBLZAADgADLQuHAckxmSi0Z00lFaDBsGQKVSaUgAEy0VzmAAq-NE6UBcC0ZCq1UqumwA1cqLg6C4HVGji62EmEBAAF9TUA)

```sh
--parser typescript
```

**Input:**

```tsx
Foo.a()

.b()

```

**Output:**

```tsx
Foo.a()
.b();

```

**Second Output:**

```tsx
Foo.a().b();

```

**Expected behavior:**

```tsx
Foo.a().b();

```

so1ve:
I just saw #15437, maybe we should make the output the same as input?

Issue #2803, opened 2017-09-13
Idempotency


frontsideair:
**Prettier Version:** 1.6.1 / master

**Code**

([playground1](https://prettier.io/playground/#%7B%22content%22%3A%22const%20testEpic%20%3D%20action%24%20%3D%3E%20action%24.ofType(test.getType())%5Cn%20%20.switchMap((%7B%20payload%20%7D)%20%3D%3E%20(%5Cn%20%20%20%20handleResponse(%5Cn%20%20%20%20%20%20ajaxPost(%5Cn%20%20%20%20%20%20%20%20'%2Ftest'%2C%5Cn%20%20%20%20%20%20%20%20%7B%20a%3A%20'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'%20%7D%5Cn%20%20%20%20%20%20)%2C%5Cn%20%20%20%20%20%20'Tested%20successfully!'%5Cn%20%20%20%20)%5Cn%20%20))%5Cn%22%2C%22options%22%3A%7B%22printWidth%22%3A80%2C%22tabWidth%22%3A2%2C%22singleQuote%22%3Afalse%2C%22trailingComma%22%3A%22none%22%2C%22bracketSpacing%22%3Atrue%2C%22jsxBracketSameLine%22%3Afalse%2C%22parser%22%3A%22babylon%22%2C%22semi%22%3Atrue%2C%22useTabs%22%3Afalse%2C%22doc%22%3Afalse%2C%22ast%22%3Afalse%7D%7D))
[...]

PR #15522, opened 2023-10-19
Stabilize member chain with empty line

seiyab:
## Description
Fix https://github.com/prettier/prettier/issues/15435

Fortunately, this doesn't conflict with https://github.com/prettier/prettier/pull/15171.
I actually implemented this on the top of https://github.com/prettier/prettier/pull/15171 ([99d389d](https://github.com/prettier/prettier/pull/15171/commits/99d389dc0137fa3e31d82be7cecc9c6a470c076b)) and rebased it into master without seeing any problems.

## Checklist


seiyab:
I might add test cases a bit.
Other meaningful test cases might not be found since it looks to reproduce under limited situation, though.

src/language-js/print/member-chain.js
```diff
@@ -331,7 +340,11 @@ function printMemberChain(path, options, print) {
 
   // If we only have a single `.`, we shouldn't do anything fancy and just
   // render everything concatenated together.
-  if (groups.length <= cutoff && !nodeHasComment) {
+  if (
+    groups.length <= cutoff &&
+    !nodeHasComment &&
+    !groups.some((g) => g.at(-1).hasTrailingEmptyLine)
```
seiyab, line 346: I'm not sure whether it should be something like `group.some(willBreak)` as https://github.com/prettier/prettier/blob/51e48e5584f055221d0d7c7cf1da04102a32df29/src/language-js/print/member-chain.js#L398.
Probably it will affect code like following:
```jsx
this.bar(a).baz(
  <ul>
    <li>a</li>
  </ul>,
);
```

seiyab:
I found following inconsistency around here and this PR doesn't resolve it. Anyway, this PR resolves idempotency issue https://github.com/prettier/prettier/issues/15435. I don't matter closing this PR if member consider it needs discussion.

**Prettier 3.0.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEU4HcAEBlOMAUAlADpSaankB0AhgCZ34CsJZmtD+xIAzhALZxM8AB4xuhANylSaLLgKFyldvUYsZbDo259BwuGInSoIADQgIABxgBLaD2SgaAJxcR0ABVcJHKGgA26DQAno4WAEYuNGAA1njYNIIAMrZoyABmgTxwkdFxCVYxaQDmyDAuAK65vHD8tuVVNTylAXAAipUQ8JnZNQBWPCLYrR1dPUhZATkWAI7jcJ7uVn4gNDwAtGhwdDvmIBU0tgGlAMIC-DTIawEB+y1QJW0AgjAVthGV8J5wLqnpkz6FgAFjB+AEAOrA2zwHhFMBwXBQFp2ABuMJC1zAPHCIFR1QAklBdrBsGAXLYbM9idgYCE2r1pjUrO4chDolZriy4DkXKi4Ps0ryYEsaCVLoyZiAii5edc6VYeeTKTB9iy0jAIbY6DBgcgABwABgsLjg81sptF4qugKZFhgNAiWp1eqQACYLJUcgAVR1+KZSuoRHa7OjJGiPSpiuAAMQgLkub1K1xoXwgIAAvhmgA)

```sh
--parser typescript
```

**Input:**

```tsx
new Set()
  
  .add(5)
  .add("some text");

new Set()  
  .add(5)

  .add("some text");

```

**Output:**

```tsx
new Set().add(5).add("some text");

new Set()
  .add(5)

  .add("some text");

```

src/language-js/print/member-chain.js
fisker: Should we name the element as `printedNodes`?

seiyab: Ah, I was wrong. Trying to add test for it, I found it isn't affected by `groups.some((g) => g.at(-1).hasTrailingEmptyLine)` vs `willBreak(oneLine)` (`group.some(willBreak)` that I suggested  isn't correct).
Changing `groups.some((g) => g.at(-1).hasTrailingEmptyLine)` into `willBreak(oneLine)`, some existing test are affected.

<details><summary>Example</summary>

**stable and this PR**
```js
function f() {
  return this._getWorker(workerOptions)({
    filePath,
    hasteImplModulePath: this._options.hasteImplModulePath,
  }).then((metadata) => {
    // \`1\` for truthy values instead of \`true\` to save cache space.
    fileMetadata[H.VISITED] = 1;
    const metadataId = metadata.id;
    const metadataModule = metadata.module;
    if (metadataId && metadataModule) {
      fileMetadata[H.ID] = metadataId;
      setModule(metadataId, metadataModule);
    }
    fileMetadata[H.DEPENDENCIES] = metadata.dependencies || [];
  });
}
```

**willBreak(oneLine)**
```js
function f() {
  return this._getWorker(workerOptions)({
    filePath,
    hasteImplModulePath: this._options.hasteImplModulePath,
  })
    .then((metadata) => {
      // \`1\` for truthy values instead of \`true\` to save cache space.
      fileMetadata[H.VISITED] = 1;
      const metadataId = metadata.id;
      const metadataModule = metadata.module;
      if (metadataId && metadataModule) {
        fileMetadata[H.ID] = metadataId;
        setModule(metadataId, metadataModule);
      }
      fileMetadata[H.DEPENDENCIES] = metadata.dependencies || [];
    });
}
```

</details>


I'd like to hear your opinion about
- whether we should find and add test case specifically focusing on it.
    - Just for information: [Zod](https://github.com/colinhacks/zod#strings) looks good library to search for practical method chains.
- which is better `groups.some((g) => g.at(-1).hasTrailingEmptyLine)` vs `willBreak(oneLine)`
    - IMO, `groups.some((g) => g.at(-1).hasTrailingEmptyLine)` looks slightly better because:
        - it looks slightly more readable to me for existing test cases
        - it doesn't change formatting which doesn't relate to https://github.com/prettier/prettier/issues/15435

For naming, another variable is already named as `printedNodes`. I'm sorry but probably my mistake (`group.some(willBreak)`, which is not correct) confused you.

seiyab: @fisker Could you respond this, please?

fisker: Sorry for been busy, will take a look.

seiyab: I just worried you might have missed notification. There's no hurry. Thank you.

fisker (approved):
LGTM from the result.

thorn0:
These changes don't look good https://github.com/prettier/prettier-regression-testing/issues/342#issuecomment-1839765105

seiyab:
@thorn0 
It's because this branch wasn't up-to-date. The difference comes from https://github.com/prettier/prettier/pull/15547.
I added test and merge latest main.

src/language-js/print/member-chain.js
thorn0: It's best to avoid adding more uses of `willBreak`. Eventually, the existing use should be somehow removed from the member chain printing code too because it causes #2803.

thorn0 (approved):

fisker:
Maybe the empty line only makes sense when the chain is long enough.

PR #15531, opened 2023-10-21
Allow setting arbitrary Prettier options via .editorconfig

benblank:
### Description

This is a potential implementation of the feature request in #15530.

When reading key-value pairs from an EditorConfig file, any key starting with `prettier/` is translated into an equivalently-named Prettier option, e.g. `prettier/prose_wrap = always` becomes `{ proseWrap: "always" }`. Values which appear to represent booleans (i.e. `true` and `false`) or numbers (e.g. `7` or `-.5`[^1]) are parsed as such. All other values remain strings.

[^1]: While Prettier itself doesn't use floating-point or negative numbers, they are supported here for the benefit of Prettier plugins which may need them.

Prefixed keys are not restricted to those which correspond to Prettier options, so that Prettier plugins can be configured via this mechanism as well. It also creates the potential for the same option to be set via both a common key (e.g. `max_line_length`) and a prefixed key (e.g. `prettier/print_width`), but the documentation warns against doing so. In the case of a conflict, the prefixed key takes precedence.

The slash character was chosen to separate the `prettier` "namespace" from the setting name because, while there doesn't seem to be any real consensus among EditorConfig plugins on what character to use, I felt the slash was at least nicely visually distinct and that using a character other than an underscore reduces the likelihood of accidentally conflicting with other EditorConfig plugins' custom keys or future versions of the EditorConfig spec. I'm very open to selecting a different character (or entire prefix) if another would be preferable.

### Checklist

### DIFF
--- a/changelog_unreleased/misc/15531.md
+++ b/changelog_unreleased/misc/15531.md
@@ -1,0 +1,3 @@
+#### Support all Prettier options in EditorConfig files (#15531 by @benblank)
+
+Any Prettier option can be set via `.editorconfig` by converting its name to snake case and prefixing it with `prettier/`. For example, the option `--trailing-comma none` / `{ trailingComma: "none" }` can be translated to `prettier/trailing_comma = none`.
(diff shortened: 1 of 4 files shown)


fisker:
Close since #15530 has been rejected.

PR #15548, opened 2023-10-27
Upgrade to yarn 4

fisker:
## Checklist


fisker:
yarn dropped support for Node.js < 18, but we  can use [`YARN_IGNORE_NODE`](https://github.com/prettier/prettier/actions/runs/6667404209/job/18120833777?pr=15548#step:3:19) not sure if I should continue..

fisker:
Turns out it can't work on v14. We'll have to close.

PR #15592, opened 2023-11-03
Update package.json to add "browser" conditional export

### LINKED ISSUE #15591: Missing "browser" conditional export in "exports"
Importing from `"prettier"` fails when modern bundlers build for browser and use the `"exports"` pattern, because there's no `"browser"` condition - so it resolved to `"default"` instead.

Reproducible with esbuild version 0.19.5 and prettier version 3.0.3:

Run: `echo 'import "prettier";' | npx esbuild --bundle --platform=browser`

Output:
```
✘ [ERROR] Could not resolve "module"

    node_modules/prettier/index.mjs:1:57:
      1 │ import { createRequire as __prettierCreateRequire } from "module";
        ╵                                                          ~~~~

  The package "module" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "--platform=node" to do that, which will remove this error.

✘ [ERROR] Could not resolve "os"

    node_modules/prettier/index.mjs:13725:15:
      13725 │ import os from "os";
            ╵                ~~~~

  The package "os" wasn't found on the file system but is built into node. Are you trying to bundle
  for node? You can use "--platform=node" to do that, which will remove this error.

...
```

victoragnez:
## Description

Update package.json to add "browser" conditional export when using "exports" pattern

Fixes #15591

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     ".": {
       "types": "./src/index.d.ts",
       "require": "./src/index.cjs",
+      "browser": "./standalone.js",
       "default": "./src/index.js"
     },
     "./*": "./*"


fisker:
https://github.com/prettier/prettier/pull/14766/files#r1178532190

PR #15594, opened 2023-11-06
Add blog post for experimental ternaries

rattrayalex:
## Description

Supersedes https://github.com/prettier/prettier/pull/15502 from @sosukesuzuki who had pulled my post out from https://github.com/prettier/prettier/pull/13183 and made some helpful edits (thank you!)

I am keen for feedback on this post!

A few things I'm wondering:

1. Should we do more to explain why formatting ternaries is hard? I wanted/tried to but couldn't think of reasonably concise examples.
2. Should we give visual examples of flat and indented nested ternaries?
3. Should we do more to define our terms here? nested, indented, chained, curious, case… it can all get washed together
4. Is it clear what the new thing is and how to read it?
5. Is there a ToC automatically added? If not can we add it?
5. Should we try to explain why it took us so damn long? the real reasons are basically:
  a. I/we were afraid of community blowback from anything other than indented ternaries, but also felt that indented ternaries weren't good enough to ship
  b. I'd proposed at one point that if we didn't align on a new style soon, we just ship the revert to indented, but we as a team didn't reach consensus on that, I think
  c. We had a hard time reaching consensus as a team in general
  d. I have a very busy job (founder/ceo of https://stainlessapi.com), many of us also have very busy jobs
  e. … but none of that feels very satisfying to share in a blog post :smile: perhaps we could at least express apologies/regret that it took so long? (I certainly feel that way).

BTW if preferred we could change the target of this PR to sosuke's branch, merge it, and and work on it from there.


sosukesuzuki (approved):
Overall, LGTM! Let's wait for @fisker and @vjeux's reviews.
Also I will merge this PR when 3.1 is released, so please wait and don't merge.

Here are the answers to your concerns:

> Should we do more to explain why formatting ternaries is hard? I wanted/tried to but couldn't think of reasonably concise examples.

Okay, I think it would be nice to have that, but I understand that it's difficult.

> Should we give visual examples of flat and indented nested ternaries?

Yes, it would be nice to have that!

> Should we do more to define our terms here? nested, indented, chained, curious, case… it can all get washed together

I personally think this is unnecessary.

> Is it clear what the new thing is and how to read it?

Yes:+1:

> Is there a ToC automatically added? If not can we add it?

Yes, ToC is automatically generated.

> Should we try to explain why it took us so damn long? the real reasons are basically:

I think this doesn't need to be explained.

rattrayalex:
Okay, I'm just going to merge this so I don't block anything, but I will try to follow up and add examples of flat/nested ternaries too.

rattrayalex:
Reverted in https://github.com/prettier/prettier/pull/15602 so it can be timed with the release, will open a new version of the PR soon.

PR #15596, opened 2023-11-06
Add "browser" field in exports["."]

### LINKED ISSUE #14765: Standalone version isn't loaded with ESM and Webpack

fisker:
## Description

Closes #15592
Closes #14766
Fixes #14765

## Checklist

### DIFF
--- a/scripts/build/build-package-json.js
+++ b/scripts/build/build-package-json.js
@@ -42,6 +42,10 @@ async function buildPackageJson({ file, files }) {
       ".": {
         types: "./index.d.ts",
         require: "./index.cjs",
+        browser: {
+          import: "./standalone.mjs",
+          default: "./standalone.js",
+        },
         default: "./index.mjs",
       },
       "./*": "./*",


sosukesuzuki (approved):
I didn't know about `browser` in conditional exports..

fisker:
Local tested, it works as expected. 

I was going to add test for it, but it's hard..

PR #15649, opened 2023-11-13
Improve `getVisitorKeys` runtime performance

pmdartus:
## Description

After profiling Prettier against a large repo like [mui/material-ui](https://github.com/mui/material-ui), the `getVisitorKeys` function is flagged as the function with the highest self-time. 

Screenshot 2023-11-13 at 06 31 52

After drilling into the file, it becomes clear that most of the time is spent checking `process.env.NODE_ENV`. This PR improves `getVisitorKeys` runtime performance by swapping the two conditions in the if branch. The `undefined`  check is cheaper than the `process.env.NODE_ENV` property lookup and the string comparison. 

Rerunning Prettier against the material-ui repo shows a ~1% performance improvement.

**Before:**

Screenshot 2023-11-13 at 06 32 46

**After:**

Screenshot 2023-11-13 at 06 36 33

## Checklist

### DIFF
--- a/src/utils/create-get-visitor-keys.js
+++ b/src/utils/create-get-visitor-keys.js
@@ -7,7 +7,7 @@ function createGetVisitorKeys(visitorKeys, typeProperty = "type") {
     const type = node[typeProperty];
 
     /* c8 ignore next 5 */
-    if (process.env.NODE_ENV !== "production" && type === undefined) {
+    if (type === undefined && process.env.NODE_ENV !== "production") {
       throw new Error(
         `Can't get node type, you must pass the wrong typeProperty '${typeProperty}'`,
       );


src/utils/create-get-visitor-keys.js
```diff
@@ -7,7 +7,7 @@ function createGetVisitorKeys(visitorKeys, typeProperty = "type") {
     const type = node[typeProperty];
 
     /* c8 ignore next 5 */
-    if (process.env.NODE_ENV !== "production" && type === undefined) {
+    if (type === undefined && process.env.NODE_ENV !== "production") {
```
fisker, line 10: It's intentionally write like this, since `process.env.NODE_ENV !== "production"` will be `false` after we bundled.

fisker: And this branch will be removed in published version.

pmdartus: Oh, I see. I should have spent more time looking into the build process. 😉
Thanks for the prompt review.

PR #15675, opened 2023-11-18
【WIP】 Do not look for config file outside project

### LINKED ISSUE #8303: Don't look for a configuration file outside the project directory
When looking for the `.editorconfig` file, Prettier doesn't climb all the way up the file system. It finds the '[project (repository) root](https://www.npmjs.com/package/find-project-root)' and stops there. It would make sense to adjust the resolution of configuration files the same way.

* Prettier should be configured on the project-level, its configuration should be part of the project.
* It's encouraged to use Prettier with default settings, which means without a configuration file. It also means that if there is a configuration file in some parent directory, it can unexpectedly affect formatting.
* It's inconsistent that `.editorconfig` and `.prettierrc` are resolved differently.

But this would be a breaking change.

Thoughts?

fisker:
## Description

Fixes #8303

## Checklist

### DIFF
--- a/src/config/get-prettier-config-explorer.js
+++ b/src/config/get-prettier-config-explorer.js
@@ -105,8 +105,13 @@ async function transform(result) {
 /**
  * @return {ReturnType<import("lilconfig").lilconfig>}
  */
-function getExplorer() {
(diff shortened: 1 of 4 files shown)


fisker:
Not sure if we want consider this as breaking change.

PR #15683, opened 2023-11-19
Use `.prettierrc` consistently in docs

KristjanESPERANTO:
It is the appropriate file extension. In addition, the extension is already used in line 170. So this creates consistency.


fisker:
I'd like to use the non-extension version. You can update L170 to use `.prettierrc`.

PR #15709, opened 2023-11-23
Reduce how much text needs to be diffed when computing cursor positions

ExplodingCabbage:
## Description

This implements roughly the tactic I proposed at https://github.com/prettier/prettier/issues/4801#issuecomment-1816869699.

It's a draft for now. I have a bunch of outstanding concerns I want to dig into before it's ready for review (in addition to the stuff in the PR template checklist), many of which probably require tweaks to the logic here:

- [ ] I've changed the signatures of a bunch of functions. Are any of them part of the package's API, meant to be used either by end users of Prettier or authors of third-party plugins? I don't have a clear idea yet of what's an internal implementation detail vs part of the interface for plugin authors vs part of the interface for end users.
  - [ ] In particular: I removed `opts.cursorNode`, but perhaps that's something that third party plugins might use or modify, in which case I've broken them?
- [ ] I seem to have broken a couple of the tests that combine cursor positioning with formatting only a target range. I am a bit confused about why those tests exist because the Pretter CLI `--help` text specifically says that those two features can't be used together. (I also have no idea why my changes broke that scenario in particular.) I guess I should investigate...
- [ ] I've added some logic to `printAstToDoc` to insert the `cursor` symbol at beginning or end of the doc in the scenario where the cursor is located before the first AST node. But what happens when there are embedded docs, or when `printAstToDoc` gets called from the multiparser (are these the same scenario?)? Since I didn't consider these scenarios at all when writing the code, I presume the behaviour of my code in them is currently broken.
- [ ] Have I introduced a possible scenario where only one cursor placeholder gets printed in `printDocToString` (e.g. because the other is part of a group that doesn't get printed at all for some reason)? If this happens it will cause `printDocToString` to throw, so I need to either ensure it's impossible or modify `printDocToString` to handle it when it happens.
- [ ] What happens when the `nodeBeforeCursor` ends up _after_ `nodeAfterCursor` in the formatted doc, e.g. because some sort of sorting plugin for Prettier was used like https://github.com/trivago/prettier-plugin-sort-imports? This obviously breaks the intended logic for computing the cursor position, and right now I don't detect it or recover from it.
- [ ] Wherever it makes sense, I should add tests exploring the precise doubts I list above
  
## Checklist

### DIFF
--- a/src/document/printer.js
+++ b/src/document/printer.js
@@ -648,8 +648,8 @@ function printDocToString(doc, options) {
 
     return {
       formatted: beforeCursor + aroundCursor + afterCursor,
-      cursorNodeStart: beforeCursor.length,
-      cursorNodeText: aroundCursor,
+      cursorRegionStart: beforeCursor.length,
+      cursorRegionText: aroundCursor,
     };
   }
 
--- a/src/utils/ast-utils.js
+++ b/src/utils/ast-utils.js
@@ -47,6 +47,17 @@ function* getDescendants(node, options) {
 
 /**
  * @param {Node} node
+ * @param {{getVisitorKeys: GetVisitorKeys}}
+ */
+function isLeaf(node, options) {
+  for (const _ of getChildren(node, options)) {
+    return false;
+  }
+  return true;
+}
+
+/**
+ * @param {Node} node
  * @param {{getVisitorKeys: GetVisitorKeys, predicate: Predicate}} options
  */
 function hasDescendant(node, { getVisitorKeys, predicate }) {
@@ -59,4 +70,4 @@ function hasDescendant(node, { getVisitorKeys, predicate }) {
   return false;
 }
 
-export { hasDescendant, getDescendants, getChildren };
+export { hasDescendant, getDescendants, getChildren, isLeaf };
--- a/src/document/public.d.ts
+++ b/src/document/public.d.ts
@@ -192,8 +192,8 @@ export namespace printer {
     options: Options,
   ): {
     formatted: string;
-    cursorNodeStart?: number | undefined;
-    cursorNodeText?: string | undefined;
+    cursorRegionStart?: number | undefined;
+    cursorRegionText?: string | undefined;
   };
   interface Options {
     /**
--- a/src/language-js/print/jsx.js
+++ b/src/language-js/print/jsx.js
@@ -244,14 +244,28 @@ function printJsxElementInternal(path, options, print) {
     : group(multilineChildren, { shouldBreak: true });
 
   /*
-  `printJsxChildren` won't call `print` on `JSXText`
-  When the cursorNode is inside `cursor` won't get print.
+  `printJsxChildren` won't call `print` on `JSXText`, so when the cursorNode,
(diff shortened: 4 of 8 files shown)


commands.md
```diff
@@ -305,13 +305,14 @@ Examples:
 - `hardlineWithoutBreakParent` is used for printing tables in Prettier's Markdown printer. With `proseWrap` set to `never`, the columns are aligned only if none of the rows exceeds `printWidth`.
 - `literallineWithoutBreakParent` is used in the [Ruby plugin](https://github.com/prettier/plugin-ruby) for [printing heredoc syntax](https://github.com/prettier/plugin-ruby/blob/b6e7bd6bc3f70de8f146aa58ad0c8310518bf467/src/ruby/nodes/heredocs.js).
 
-### `cursor`
+### `cursorLeft`, `cursorRight`
```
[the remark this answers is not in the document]
ExplodingCabbage: Sure. Done.

fisker:
> What happens when the nodeBeforeCursor ends up after nodeAfterCursor in the formatted doc

Is this the reason why you add `cursorStart`/`cursorEnd`?

ExplodingCabbage:
Ooh, sorry, missed the comment above somehow.

Yeah, I think the new `cursorStart` & `cursorEnd` *only* need to exist to handle the scenario where a third-party plugin reorders AST nodes. In every other situation, it'd be fine if we just had `cursor`.

fisker:
I'm not sure it worth to make the change (`cursor` -> `cursor{Start,End}`).

a) plugins maybe already using `cursor`
b) builtin plugin doesn't reorder nodes, and we don't encourage plugin to do that either

ExplodingCabbage:
Fair enough; I've removed the new `cursorStart` and `cursorEnd` in favour of just using `cursor`.

src/utils/ast-utils.js
```diff
@@ -45,6 +45,17 @@ function* getDescendants(node, options) {
...
 
+/**
+ * @param {Node} node
+ * @param {{getVisitorKeys: GetVisitorKeys}} options
+ */
+function isLeaf(node, options) {
+  for (const _ of getChildren(node, options)) {
+    return false;
+  }
+  return true;
```
fisker, line 56:
```suggestion
  return getChildren(node, options).next().done;
```

src/document/printer.js
```diff
@@ -648,8 +680,8 @@ function printDocToString(doc, options) {
 
     return {
       formatted: beforeCursor + aroundCursor + afterCursor,
-      cursorNodeStart: beforeCursor.length,
-      cursorNodeText: aroundCursor,
+      cursorRegionStart: beforeCursor.length,
+      cursorRegionText: aroundCursor,
```
fisker, line 684: `printDocToString` is a public API, let's not change the signature.

ExplodingCabbage: Trouble is, the region we return here is no longer guaranteed to correspond to a single node. I guess I can leave the identifiers the same and add comments noting that they're misnomers if we want to preserve the signature, though.

ExplodingCabbage: (Done)

fisker (approved):
Good job

PR #1637, opened 2017-05-20
Add `cursorOffset` option for cursor translation

josephfrazier:
This addresses https://github.com/prettier/prettier/issues/93 by
adding a new option, `cursorOffset`, that tells prettier to determine
the location of the cursor after the code has been formatted. This is
accessible through the API via a new function, `formatWithCursor`, which
returns a `{formatted: string, cursorOffset: ?number}`.

Here's a usage example:

```js
require('prettier').formatWithCursor(' 1', {cursorOffset: 1})
// -> { formatted: '1;\n', cursorOffset: 0 }
```

This implementation works roughly as described in https://github.com/prettier/prettier/issues/93#issuecomment-286011921.
It doesn't yet have tests, but I wanted to go ahead and get some
feedback on the API to see if this seems like the right way to go.

Thoughts?

PR #15750, opened 2023-12-04
Remove claim, untrue since over 5 years ago, that cursorOffset is incompatible with rangeStart/rangeEnd

ExplodingCabbage:
## Description

In May 2017, @josephfrazier [added the `cursorOffset` option](https://github.com/prettier/prettier/pull/1637). This initial implementation of `cursorOffset` was incompatible with `rangeStart`/`rangeEnd`, and so Joseph added lines to the docs about `cursorOffset` and `rangeStart`/`rangeEnd` saying:

> This option ... cannot be used with `rangeStart` and `rangeEnd`.

and

> This option cannot be used with `cursorOffset`.

This was all true and reasonable at the time, but the next year, in April 2018, the actual incompatibility was fixed by @ds300 in https://github.com/prettier/prettier/pull/4397. But I guess that neither he nor anyone who reviewed his PR noticed the *documentation* about the incompatibility still exists, because it wasn't removed at the time and still exists to this day, even though David's fix made it obsolete.

I suggest we finally remove those false statements, 5 years later. This PR does that. :)

## Checklist

- [ ] I’ve added tests to confirm my change works. - *No reason to, though I've updated existing snapshots as needed.*
- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory). - *N/A*


thorn0:
Thanks! Could you also change the types (`src/index.d.ts`) accordingly?

fisker (approved):

ExplodingCabbage:
> Could you also change the types (`src/index.d.ts`) accordingly?

Ah, interesting - I hadn't looked at the types and hadn't noticed that `CursorOptions` used properties of type `never` to forbid `rangeStart` or `rangeEnd` from being set (or at least to purport to - clearly it still actually worked!). I've [removed those properties](https://github.com/prettier/prettier/pull/15750/commits/0de9fc565088ae166567f79387dfc40e10b8871f).

Hmm... but wait, there's nothing special any more about `cursorOffset` to justify there being a dedicated `CursorOptions` type at all, is there? So it probably makes sense to eliminate the type entirely, in favour of simply adding a `cursorOffset` property to `RequiredOptions` akin to all the other properties on there. So I've gone ahead and [done that too](https://github.com/prettier/prettier/pull/15750/commits/43c69445e92e319ab4614e83e534de743e569455).

(One conceivable objection I see to this is that it might break compatibility with packages that are currently importing `CursorOptions`. I defer to you guys as Prettier's maintainers on whether it is likely that such packages exist and, if so, whether breaking compatibility with them is acceptable if so; I have no idea on either point.)

Tangential aside: what is the sense in which `RequiredOptions` are "required"? That naming is puzzling to me and I hesitated a bit about adding `cursorOffset` to the "required" options because passing a `cursorOffset` when calling Prettier is of course entirely optional before realising that of course the same is true of literally every option and so "required" must mean something else. It might be worth chucking a one- or two-sentence doc comment above the type explaining its meaning!

fisker:
Sometimes, `cursorOffset` is useless, see  https://github.com/prettier/prettier/pull/12803

ExplodingCabbage:
> In `format()` and `check()`, `cursorOffset` is useless, see https://github.com/prettier/prettier/pull/12803

... and therefore you'd like to keep a separate `CursorOptions` type, simply to communicate via the type system that some functions that take options respect `cursorOffset` and others don't? Okay, I'll revert https://github.com/prettier/prettier/pull/15750/commits/43c69445e92e319ab4614e83e534de743e569455 then.

(If I'm not inferring your wishes correctly, please go ahead and just edit the PR as you see fit - I've got it set to allow edits by maintainers!)

thorn0:
I can't see the type changes in the resulting diff

thorn0:
@ExplodingCabbage Oops, somehow overlooked it
@fisker What's that CI error? Something known?

fisker:
Type test is failing, don't know why. https://github.com/prettier/prettier/tree/main/tests/dts/unit

thorn0:
Figured it out https://github.com/prettier/prettier/pull/15766
@ExplodingCabbage please remove `@ts-expect-error` in `tests/dts/unit/cases/standalone.ts`

thorn0 (approved):

josephfrazier:
Thanks for catching and fixing this, @ExplodingCabbage!

PR #15757, opened 2023-12-05
Keep empty import attributes and assertions

fisker:
## Checklist


Conaclos:
Is there any reason to keep import assertions/attributes empty?

kachkaev:
> Is there any reason to keep empty import assertions/attributes?

I believe that removing `assert {}` is [out of Prettier’s scope](https://prettier.io/docs/en/rationale#what-prettier-is-_not_-concerned-about). This can be done by a linter (e.g. `eslint --fix`).

PR #15770, opened 2023-12-07
Reenable parsers.ts test case (since the OOM issue that was the reason for disabling it no longer exists)

ExplodingCabbage:
## Description

I noticed this disabled test case. If I checkout [6d21de1](https://github.com/prettier/prettier/commit/6d21de11fd532cc8c4ed42f957d24c45dddff3ec), I can reproduce the OOM issue alluded to in the comment, but not on main; it seems to me, therefore, that the case should be reenabled.

That still turned out to require a bit of thought since:
* the test was legitimately failing. `Parser.parse` was typed as _requiring_ the `options` parameter, but the implementations and the disabled test case both disagreed with that and treated the `options` parameter as optional. I've fixed this by modifying the type declaration of `parse`.
* <s>the test relies on importing generated types from `dist/`, and therefore requires that the types have been built before you run `yarn test` - something which isn't true of any other tests. I feared this would lead to new contributors being confused or learning to distrust local test failures if they run `yarn test` locally without having run `yarn build` and see a failure. I therefore think it makes sense to gate this test case behind the `FULL_TEST` environment variable and update the "Deeper testing" instructions in CONTRIBUTING.md to include running `yarn build` before `FULL_TEST=1 jest`.</s> - never mind; this was the wrong way to solve the reliance on generated types and has been undone. See review discussion with fisker below.

I don't have much deep insight into these tests; I just noticed some tech debt and randomly decided to clean it up. Feel free to make edits to this PR; I've got it set to allow edits by maintainers.

## Checklist


.github/workflows/dev-test.yml
```diff
@@ -60,6 +60,9 @@ jobs:
       - name: Install Dependencies
         run: yarn install --immutable
 
+      - name: Build Package
+        run: yarn build --clean
```
fisker, line 64: If the tests need build, it should be in prod test.

fisker: https://github.com/prettier/prettier/blob/5f7aedc1cf0a1b3b4ec2c5a5ca0c09d1e0d00660/jest.config.js#L34

fisker (changes requested):
Need update

.github/workflows/dev-test.yml
ExplodingCabbage: Aha! So _that's_ why there was some commented out stuff with an `isProduction` variable in the test before! I had not made the connection that "production" tests were ones where `yarn build` had been run.

Change made; should be ready for re-review.

tests/dts/unit/cases/plugins.ts
```diff
@@ -16,7 +16,7 @@ const plugin: prettier.Plugin<PluginAST> = {
   ],
   parsers: {
     lines: {
-      parse(text, options) {
+      parse(text, options = {}) {
```
fisker, line 19: Why?

src/language-js/parse/typescript.js
```diff
@@ -58,7 +58,7 @@ function getParseOptionsCombinations(text, options) {
   ];
 }
 
-function parse(text, options) {
+function parse(text, options = {}) {
```
fisker, line 61: Changes in this file seems unnecessary, let's not do unrelated changes.

tests/dts/unit/cases/plugins.ts
ExplodingCabbage: If the second parameter doesn't have a default argument, TypeScript thinks it's mandatory, and thus thinks that `parse` here doesn't provide the first overload in the definition of `Parser` - namely:

```
parse(text: string): T | Promise<T>;
```

You can see this locally by deleting the ` = {}` that I add here and running:

```
yarn build && NODE_ENV=production yarn test dts
```

- it'll fail.

docs/plugins.md
```diff
@@ -118,6 +118,7 @@ export const parsers = {
 The signature of the `parse` function is:
 
 ```ts
+function parse(text: string): Promise<AST> | AST;
```
fisker, line 121: I think this signals what the arguments this function will receive, I don't think we should add this, many plugins define the following functions/methods with less parameters, we didn't mark they are optional. I'm not familiar with  TS, teach me if there is a good explanation.

tests/dts/unit/cases/plugins.ts
fisker: options alway provide when calling inside prettier. we allow it omitted only because we crushed jest once. not really meaningful.

src/language-js/parse/typescript.js
ExplodingCabbage: They're not *strictly* necessary but only because the typechecker isn't actually checking that `parse` here conforms to the signature given in `Parser.parse` (I guess because `createParser` has no type annotations and so isn't being checked?). If this function *were* getting typechecked, it would fail without the change I make here, just like https://github.com/prettier/prettier/pull/15770#discussion_r1424531564 does. I think it's therefore better to make the change.

tests/dts/unit/cases/plugins.ts
ExplodingCabbage: If you want to enforce that `options` is mandatory, we can do that. I figured that it shouldn't be mandatory because the test that I'm restoring here was specifically testing that we could call `parse(text)` without a second argument, and so I assumed that was something important to support and that the types were wrong. If actually it's the _test_ that's wrong, I can leave the types alone and change the test instead to pass an `options` argument.

docs/plugins.md
ExplodingCabbage: So, basically here I'm saying that there are TWO valid ways to call `parse`:

1. You can call it with a single argument, `parse(text)`, OR
2. You can call it with two arguments, `parse(text, options)`

Prior to this PR, the codebase can't make up its mind about whether the first option is supposed to be allowed or not. The [old declaration of the `Parser` interface](https://github.com/prettier/prettier/pull/15770/files#diff-619cc2239a7eb4ef05041767e6ab0e4fb069e98ee9d2c83c2d8e1865c475e141L450) said that the `options` argument was mandatory and that you cannot just call `parse(text)`. But *all* of our actual implementations of `parse` supported calling `parse(text)` one way or another (either they checked for options being null and handled that case, or they had a default argument of `{}`, both of which effectively mean you can call `parse(text)` without providing a second argument), and the unit test that I restore in this PR specifically tested that it's possible to call `parse(text)` without a second argument.

Clearly this inconsistency should be resolved one way or another. *Either* the previous types were wrong and needed changing to say that it's okay to call `parse(text)` (that's what I've assumed in this PR), *or* the test where we were calling `parse(text)` is wrong and should be calling `parse(text, options)` instead.

tests/dts/unit/cases/plugins.ts
fisker: I don't really know TS well, but if it's for bultin parsers, then yes, we should force user provide options, since we change parsers  anytime we want, user call it manually is not encouraged, they are on their own risk, though we allow omit the `options` but we don't want document/add type for this.

ExplodingCabbage: 👍 - I'll rework this PR tomorrow morning to modify the test instead of the types.

fisker: In short, built-in plugins are considered private, the parse function should only be called by prettier, and we always provide `options`.

docs/plugins.md
fisker: `parse(test)` was added by me, but I forgot where it is...

fisker: Found, https://github.com/prettier/prettier/issues/11888, we don't really support `parse(text)`.

PR #15789, opened 2023-12-13
Re-enable tests/dts/unit/cases/parsers.ts - take 2

ExplodingCabbage:
## Description

This is a second try at https://github.com/prettier/prettier/pull/15770. As noted in that PR, the OOM issue that was the reason for disabling this test no longer exists, but the test fails because the types forbid calling `parse` without an `options` argument and the test case tries to do exactly that. To resolve that this time around, I've (mostly) modified the test case instead of the types.

(I still needed to mess with the types a bit to be able to call `parse` with a realistic options object, because the objects that get passed when I format something with `bin/prettier.js` weren't quite implementing the `RequiredOptions` interface.)

## Checklist

- [ ] (If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).  - N/A


tests/dts/unit/cases/parsers.ts
```diff
@@ -14,26 +14,74 @@ import * as prettierPluginGlimmer from "../../../../dist/plugins/glimmer.js";
...
+  locStart: (node) => 0,
+  printer: {
+    print: (
+      path: prettier.AstPath,
+      options: prettier.ParserOptions,
+      print: (path: prettier.AstPath) => prettier.Doc,
+      args?: unknown,
+    ) => [],
+  },
+  originalText: "bla bla bla",
```
fisker, line 62: This looks crazy to me... I think type of `ParserOptions` is not correct. Not really matters, like I said, we should consider it's private.

fisker (approved):

tests/dts/unit/cases/parsers.ts
ExplodingCabbage: FWIW I constructed this object basically by logging a real example of what was getting passed to parseText in core.js (from running `bin/prettier.js` to format something with some log lines added to core.js) and then replacing the functions with stubs. So, unless I screwed up, this is a realistic example of what actually gets passed - not something crazy!

sosukesuzuki (approved):
Thank you

src/index.d.ts
```diff
@@ -430,6 +424,22 @@ export interface RequiredOptions extends doc.printer.Options {
...
+  /**
+   * Put the `>` of a multi-line JSX element at the end of the last line instead of being alone on the next line.
+   * @default false
+   * @deprecated use bracketSameLine instead
+   */
+  jsxBracketSameLine?: boolean;
+  /**
+   * Arbitrary additional values on an options object are always allowed.
+   */
+  [_: string]: unknown;
```
Sec-ant, line 442: @ExplodingCabbage I'm a plugin developer and I use `keyof Options` to get the known keys in Prettier Options and this line of code prevents me from doing that again. I think this is a little too loose. Would you consider revert this change?

JounQin: Waht means it's actually a BREAKING CHANGE for other plugin authors or users.

ExplodingCabbage: Ah, man! What a mess!

> I think this is a little too loose

I mean, given the places we currently annotate arguments as having type `ParserOptions`, the old way was *definitely* too tight; in practice arguments typed as `ParserOptions` *always* included keys the ones explicitly listed here (see discussion at https://github.com/prettier/prettier/pull/15789/files#r1425688411). I also suspect that, *taking as given the places Prettier currently uses these types*, they really do need to be this loose (and that merely listing more keys like the `printer` key wouldn't cut it), since I figure third-party plugins may define their own options whose keys Prettier itself can't know and thus can't possibly be listed here. So IMO the change I made here is a fix and the type was incorrect before. However...

> I use `keyof Options` to get the known keys

... this *does* seem like it might be a legit use-case that I've broken. It might make sense to split this into two types - one that just has the known keys, and one that allows arbitrary additional keys. (I'd be curious to know what you need to use `keyof Options` for, though?)

A quick tangent (that will ultimately be relevant): I want to comment that I kinda hate *lots* of things about the options objects & types as they exist right now:
* the confusing type names (e.g. `RequiredOptions` - "required" in what sense?)
* the confusing & arbitrary type hierarchy (`ParserOptions` extends `RequiredOptions` which extends `doc.Printer.Options`, so all printing options are also mandatory parsing options but not vice versa - why?!)
* the fact that at runtime a load of extra stuff like a `printer` or `cursorNode` gets tacked onto the `options` object, and the existence of these extra keys isn't reflected in any of the types but *is* relied on by the underlying JS code. This is more fundamental than just being an issue with the TypeScript types; it's an issue with the underlying JS code, and I find it horrible for multiple reasons:
  * some of these things, like `cursorNode`, seem to me like they are really part of the internal state of the formatting algorithm; it's unintuitive to call them "options"
  * others are just methods of the parser being used that are getting shoved onto the options object as a way of passing them around (e.g. the `locStart` and `locEnd` functions); it's *again* unintuitive to me to call these "options".
  * the gradual building up of options as Prettier runs makes it hard to know what keys are present on an "options" object when looking at a bit of the code in isolation. You need to understand the order in which things get added to options and where in the sequence the code you're looking at runs.
  * as a consequence of the points above, it's *very* hard to figure out how to manually build a valid object to pass directly to some function that takes an options argument

Note that the point above about stuff getting tacked onto options objects is why I had to liberalize the `RequiredOptions` type in this PR to even be able to pass a realistic options object to a parser in my test cas without causing a type error, and thus is part of the chain of causation that leads to the backwards compatibility breakage.

I was trying to be surgical with my changes in this PR here to avoid any breaking change (obviously I still failed!) but because of the points above I found touching this code at all pretty frustrating. If backwards compatibility weren't a concern, I'd want to refactor the innards a bunch, ripping the parser methods and state variables off the "options" objects and passing them around some other way. Since backwards compatibility _is_ a concern, I am not really sure what to do. But in my opinion the handling of "options" right now in the Prettier code is a mess, this PR only scratches the surface of it, and I don't really see a way to make it clean without significant breaking changes.

Whether to make those changes or not, and whether to revert this PR or not, is a choice I leave to the maintainers; I've offered my two cents above but it's not my call to make!

Sec-ant: > I figure third-party plugins may define their own options whose keys Prettier itself can't know and thus can't possibly be listed here. So IMO the change I made here is a fix and the type was incorrect before.

> It might make sense to split this into two types - one that just has the known keys, and one that allows arbitrary additional keys.

I think the typical TypeScript way is to use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)?

> I'd be curious to know what you need to use `keyof Options` for, though?

If you're interested, the place where I use it is [here](https://github.com/Sec-ant/prettier-plugin-embed/blob/3ffbe984fc6dec6194503f5e4f3ab6b2f7f93cb7/src/types.ts#L57). The `keyof` is indirectly called by [`Omit`](https://github.com/microsoft/TypeScript/blob/6add96fd18d126d1505de5d96f0df6a86e5f2c21/src/lib/es5.d.ts#L1605). I'm providing an `overrides`-like option for my plugin for users to override some of the options, and there're ones that cannot be overridden because of the way Prettier applies them. So I created this type to omit those options wanting to give users a hint of which options are supported and which are not.

> I want to comment that I kinda hate lots of things about the options objects & types as they exist right now.

Believe me, I get it XD. I've also used lots of patches in my codebase to fight with the types in Prettier.

> Whether to make those changes or not, and whether to revert this PR or not, is a choice I leave to the maintainers; I've offered my two cents above but it's not my call to make!

Sounds fair. The good news for me is that I can always strip all the index signatures first and then use `keyof` to get the literal keys, with the help of [a utility type](https://github.com/sindresorhus/type-fest/blob/main/source/omit-index-signature.d.ts).

And thanks for your two cents. Have a good day!

JounQin: I also think `Declaration Merging` is the right way to do this, a reality example here: https://github.com/eslint-types/define-config-plugin-types/blob/2c2ae3be59b31928e078de1199606e18d6acc957/types/import/index.d.ts#L3-L8

JounQin: The current `[_: string]: unknown;` will make typos unable to be checked.

PR #15806, opened 2023-12-15
Fix inconsistencies for ChainElements

### LINKED ISSUE #15785: Different output for optional call/member expression using `typescript` parser vs `babel` and `babel-ts`
**Prettier 3.1.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VACAwgQQMoCiGAjEgDpToYByEATgLYCGANhgEZwAWTAbgJb0ANBhhc4GAAb8AzgQCOAV1aSMYVm3Z04TANYyAdBQBmiqGBiCoGGRAZwAYmYtXqTewAoKGG3bgAZaABzACFtPX4oIIAFJjp3OHg6N3shbwwmKAgxODpAqNj4+ySUuDSoAEoMYHTtGEU6a1kFZRYPJgNmAAcPDwBtGBEAfQBdKoBeAD5RAH4DfgATCpF2TqYe-sGMUYnpmDnFioqAbgoAXwoKKgB1cWsmMDA4GRlRCCkmVUyFqXZVWQwEC6lmgrBEMAAnl1nmA6PxgRgFhBnlAAOQwDjhXRXTA5NQaFaKDHsJicNjfDikuAsAC0MFeSKMUFM5hB1ls9icrNcCS81l89nyoSxkRicQSJQS5R8mWy4jywUKEtypXKVRq-LqDSaciUrHac26vQGwzGGCms3mSxWhvWxq2O3NewOSxO50ulEwuEIGAATORPTR6MxNNw+II6ODxFJjPwWElVOoWGxFDJnhkMEE6BBFNCfiwmBCcxiPF0ia8cjiMnRswB3DBwAAeXW0LysGC64uKuVe0FE0ZkCQwLEiEiYrzIIFj8dyk4qTMgUBkGJyADVWIoJOMMAB5dgAKzgFgMCBgcOeHg5AWCe8PFlK8+nST5Pn6I+XooAkgszRar-5ZBgAAVCAsDsTttBAgAJXI4HmcwWEUBYL3fSwom-ZYKDdKAq2ufgxAyawgTZVgMHsBhODoBtm1bGQrBWKlaXpDBdDgOAugrLgAXHDMsxzPMqwLIsiUJFcoRhOEETMPjc1efDAVgd4mAwKA4HrEdVIXaBl37OB10QrddwPI8YBPWBzxkS8-CFW8TIfAwn1yF8MDfQCvx-XYBQCQCQLAhgILgaDYIOBCkJQtz0OtLDThwwNvSIABmAMqFoRhSM4HgBGEXTRDgfyC3gaiW2eOi+y0HR9AyFdo1jOgdMiJIuggAqrCZFgICCIJcgMdqgmcyQcGHYJcvyphCuXOEogwWt8K4DNFwWfCrFYJAMAAEmAMAeCgVSWCtM4RApHJ2T8Mj6AkRd4FgfsxrUaA0M3V4zEsNgNoAKTwHdqAMCbRX4YwIQ8dr1DZCozkkaKPRuWbCMBYFlrYcjKKK2j6NEcSZFheEMSRFF0UxCqctq5cqwa3ImpavsKUiZcdB+cq9E4iQ0wWtqOq6ugeo6-rBvaqb4FG8az1FaaYeUhaltBFhVo2rbMl2l0DsIn5jq8s7tDu2BTxujFLsiR6MGeuN1uAD6vp+4Won+wHgbGqwwYhypjhAIQQGIqwZGQUA4jrQoEE9lBWFrQtPddrQHlYmA8ASADVOQYxWDTMP4jASO8E7MBRWQM9N1dtMGH4bO6FzkAm2hOF7FgVggNyKA4n4Z548TsoQDoqIWDgABFRRsjgJuWCTkB9xkRs8FFDvu97-vB6UXvomzDjkBAccaVUuBkIWF2QDPJg41FPzmCXjQt7boIO5wGBhfYIk4GiXJY77pAE4HluuBgBgWFufDngzuA8H9paAhIRLzAC8LevBNyfigMhWAeAsbAhwNAvAkIO7Txbi2CAaZrjxC6EvYqaY6C8D7q7GmuQYDzyYEEQ+T9m6uwggQpeJIyRb20Eofg2gKFUKYGg12MBSR4QWGIZAvpXapkCqSAOz9B55U4AsDe-hMhBGUF1BwwYxpoSCEfIkEAt4tgagIoRSBiC+gAAxnDOEAA)

```sh
--parser babel
--print-width 120
```

**Input:**

```jsx
// CASE 1:
// Normal behavior, the `isEqual` call breaks.
function someFunctionName(
  someLongBreakingParameterName,
  anotherLongParameterName,
) {
  return isEqual(a.map(([t, _]) => t?.id), b.map(([t, _]) => t?.id));
}

// When access to `a` and `b` is optional, typescript doesn't break
// the call, but babel and babel-ts do.
function someFunctionName(
  someLongBreakingParameterName,
  anotherLongParameterName,
) {
  return isEqual(a?.map(([t, _]) => t?.id), b?.map(([t, _]) => t?.id));
}

// CASE 2:
// Normal behavior, the `filter` call uses a grouped layout (puts the
// arrow expression parameters on the same line as "filter").
const theValue = Object.entries(someLongObjectName).filter(
  ([listingId]) => someListToCompareToHere.includes(listingId),
);

// With an optional member expression, babel-ts keeps this as a grouped
// layout, but typescript ungroups it onto a new line.
const theValue = Object.entries(someLongObjectName).filter(
  ([listingId]) => someListToCompareToHere?.includes(listingId),
);

// CASE 3:
// Normal behavior, the template expression breaks at the first interpolation.
logger.log(
  `A long template string with a conditional: ${channel.id}, and then some more content that continues until ${JSON.stringify(location)}`
);

// With an optional member expression, typescript doesn't break the first
// interpolation and instead breaks the second.
logger.log(
  `A long template string with a conditional: ${channel?.id}, and then some more content that continues until ${JSON.stringify(location)}`
);
```

**Output:**

```jsx
// CASE 1:
// Normal behavior, the `isEqual` call breaks.
function someFunctionName(someLongBreakingParameterName, anotherLongParameterName) {
  return isEqual(
    a.map(([t, _]) => t?.id),
    b.map(([t, _]) => t?.id),
  );
}

// When access to `a` and `b` is optional, typescript doesn't break
// the call, but babel and babel-ts do.
function someFunctionName(someLongBreakingParameterName, anotherLongParameterName) {
  return isEqual(
    a?.map(([t, _]) => t?.id),
    b?.map(([t, _]) => t?.id),
  );
}

// CASE 2:
// Normal behavior, the `filter` call uses a grouped layout (puts the
// arrow expression parameters on the same line as "filter").
const theValue = Object.entries(someLongObjectName).filter(([listingId]) =>
  someListToCompareToHere.includes(listingId),
);

// With an optional member expression, babel-ts keeps this as a grouped
// layout, but typescript ungroups it onto a new line.
const theValue = Object.entries(someLongObjectName).filter(([listingId]) =>
  someListToCompareToHere?.includes(listingId),
);

// CASE 3:
// Normal behavior, the template expression breaks at the first interpolation.
logger.log(
  `A long template string with a conditional: ${
    channel.id
  }, and then some more content that continues until ${JSON.stringify(location)}`,
);

// With an optional member expression, typescript doesn't break the first
// interpolation and instead breaks the second.
logger.log(
  `A long template string with a conditional: ${
    channel?.id
  }, and then some more content that continues until ${JSON.stringify(location)}`,
);

```

**Expected behavior:**

The output from each of these 3 cases should match regardless of which parser is used (`babel`, `babel-ts`, or `typescript`).

The consistent difference here is whether the expression is `optional` or not. Inspecting the parsed ASTs, Babel (and thus babel-ts) parses these as distinct node types (`OptionalCallExpression` and `OptionalMemberExpression`), but the TypeScript parser just uses the plain `CallExpression` and `MemberExpression` nodes with an extra `optional` property on them.

From https://github.com/prettier/prettier/issues/10244, it seems like maybe in the past (or in a different situation), these were parsed as `ChainExpression`, which then got transformed by Prettier into the appropriate optional expression nodes, but that is no longer the case, so the transform doesn't pick them up and convert as needed.

fisker:
## Description

Fixes #15785

## Checklist


fisker:
Changes in https://github.com/prettier/prettier-regression-testing/issues/342#issuecomment-1857537533 are expected, since babel/babel-ts print them this way.

sosukesuzuki (approved):

PR #15814, opened 2023-12-16
Add `isUnionType` and `isIntersectionType` utilities

fisker:
## Checklist

### DIFF
--- a/src/language-js/print/assignment.js
+++ b/src/language-js/print/assignment.js
@@ -19,6 +19,8 @@ import {
   isLoneShortArgument,
   isObjectProperty,
   createTypeCheckFunction,
+  isUnionType,
+  isIntersectionType,
 } from "../utils/index.js";
 import { shouldInlineLogicalExpression } from "./binaryish.js";
 import { printCallExpression } from "./call-expression.js";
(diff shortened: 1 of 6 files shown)


src/language-js/print/assignment.js
```diff
@@ -19,6 +19,8 @@ import {
   isLoneShortArgument,
   isObjectProperty,
   createTypeCheckFunction,
+  isUnionType,
+  isIntersectionType,
```
kachkaev, line 23: WDYT of adding https://github.com/lydell/eslint-plugin-simple-import-sort to the codebase? This will automatically make import lists alphabetically ordered. The diffs will be more predictable and specific import names will be easier to find. These lists are currently ordered based on the introduction time, which makes sense historically, but not so much when you look at the code snapshot.

kachkaev (approved):

src/language-js/print/assignment.js
fisker: Good idea.

kachkaev: Do you want to do it? I can craft a quick PR if you’re busy with other stuff.

fisker: It will be great, if you can work on it.

kachkaev: → https://github.com/prettier/prettier/pull/15817

PR #15885, opened 2024-01-06
Add VSCode settings

fisker:
## Checklist


.vscode/settings.json
liuxingbaoyu, line 1: Some people may have custom settings, perhaps `settings.example.json`?

fisker: Good point, I'm going to rename it as `settings.example.json`, and add a `prepare` script to copy to `settings.json` if it's not already exist.

liuxingbaoyu: Good idea!

package.json
```diff
@@ -184,7 +184,8 @@
     "build": "node ./scripts/build/build.js",
     "build:website": "node ./scripts/build-website.js",
     "gen:changelog": "node ./scripts/generate-changelog.js",
-    "debug": "node bin/prettier.js --ignore-path=.prettierignore"
+    "debug": "node bin/prettier.js --ignore-path=.prettierignore",
+    "postinstall": "node -e \"try{fs.existsSync('.vscode/settings.json')||fs.writeFileSync('.vscode/settings.json',fs.readFileSync('.vscode/settings.example.json'))}catch(_){}\""
```
fisker, line 188: Turns out we can't use an extra script file to copy the file, because this will run when `yarn install prettier@prettier/prettier`

liuxingbaoyu: Is there any reason we don't use `prepare`?

fisker: yarn>=2 don't run `prepare`.

liuxingbaoyu: I am surprised, which means that the `Prepare` in the Babel warehouse did not work as expected.
But when we use Yarn 4, maybe we can use it to complete this.
https://yarnpkg.com/configuration/yarnrc#enableConstraintschecks

fisker: I don't understand how `enableConstraintschecks` can do this. But yarn plugin should be able to do it.

liuxingbaoyu: When `EnableConstraintschecks` is true, Yarn runs every time` yarn.config.cjs`.
https://github.com/babel/babel/blob/main/yarn.config.cjs

fisker: I'm going to use yarn plugin instead.

PR #15896, opened 2024-01-08
Fix css embed in JSX when using (only) a spread attribute

eelco:
## Description

Prettier *crashes* on this JSX/TSX code:
```
<style {...spread}>{`.{}`}</style>
```

This is fixed by improving the detection mechanism used in `isStyledJsx`.

## Checklist


src/language-js/embed/css.js
```diff
@@ -75,7 +75,10 @@ function isStyledJsx({ node, parent, grandparent }) {
       grandparent.type === "JSXElement" &&
       grandparent.openingElement.name.name === "style" &&
       grandparent.openingElement.attributes.some(
-        (attribute) => attribute.name.name === "jsx",
+        (attribute) =>
+          (attribute.type === "JSXAttribute" &&
+            attribute.name.name === "jsx") ||
+          attribute.type === "JSXSpreadAttribute",
```
fisker, line 81: I don't think we should add this `JSXSpreadAttribute` here.

fisker, line 81:
```suggestion
      (attribute) =>
        attribute.type === "JSXAttribute" && attribute.name.name === "jsx",
````

eelco: Why not? Without it the CSS within the template string will not format in the example given in the description and the test.

eelco: That said, I’m not entirely sure why the `isStyledJsx` function looks at the attributes of the `style` element. Perhaps that whole check can be removed?

fisker: `jsx` is required in [`styled-jsx`](https://github.com/vercel/styled-jsx). We don't want to break something we don't know.

eelco: Updated the code and the snapshot.

fisker (approved):
Thank you!

sosukesuzuki (approved):

Issue #15877, opened 2024-01-05
Doc of paragraph in markdown violates a rule of `fill()`


seiyab:
**Prettier 3.1.1**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDABB9AdK6BGB6OeY6ZJuxIANCBAA4wCW0AzsqKgE7cQDuABR4IOKVABt+qAJ4c6+bqjABrODADKqALZwAMsyhxkAM0ls4CpavUaGywwHNkMbgFdLIC9uYv3nuAAPBjhuZl1YSQAVUKgeZjgxMwkLOjYnCTgARTcIeFNzTwArNkCNDOzc-KRk1JAARyq4QT4GMTQ2AFojOAATPtoQV1RmCScAYQhtbVRkNAkJQfSoR0yAQRhXZnw3eEFQgyMClM8ACxhtCQB1U+Z4NnswOA1RO+YANzuZObA2eRB3h4AJJQfqwDRgMJMNagjQwGSZY51Bh8CxXJQMOaSaRyQaGCzcGAtVCOGZIzz2bgEuYzbgqXoCKCDFGGGBXZi9GCnZAADgADHRuHBGswhcTSbMaoU6DBUPh2ZzuUgACx0NwWKJypLSkBwbT4Pr9Xp6VArNwkuAAMQg3BmmycWN2EBAAF8XUA)

```sh
--parser markdown
--tab-width 4
--prose-wrap always
```

**Input:**

```markdown
a a  
 b b  
 c c  

  
```

**Doc:**

```markdown
[
  fill([
    "a",
    line,
    "a  ",
    markAsRoot(literalline),
    line,
    "b",
    line,
    "b  ",
    markAsRoot(literalline),
    line,
    "c",
    line,
    "c",
  ]),
  hardline,
]

```

**Expected behavior:**
The result doc should preserve the following invaliant:
https://github.com/prettier/prettier/blob/bab892242a1f9d8fcae50514b9304bf03f2e25ab/src/document/printer.js#L442-L443

I haven't found user-facing problem yet. I found it investigating https://github.com/prettier/prettier/discussions/15836#discussioncomment-7917975.

seiyab:
Oh it can be reproduced without `--prose-wrap: always`. I changed the issue title.

**Prettier 3.1.1**
[...]

PR #15909, opened 2024-01-11
delete `normalizeDoc()`

seiyab:
## Description
Resolves https://github.com/prettier/prettier/issues/15877 and https://github.com/prettier/prettier/issues/9503.
`flattenFill()` is the key.

## Checklist

- [ ] ~(If changing the API or CLI) I’ve documented the changes I’ve made (in the `docs/` directory).~


tests/format/mdx/mdx/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -416,7 +416,8 @@ A {/* JS-style comment */} comment.
 </Hello>
 <Hello>
   test <World /> test
-</Hello>123
+</Hello>
+123
```
seiyab, line 420: I consider this get correct.

src/document/utils/fill-builder.js
```diff
@@ -0,0 +1,30 @@
...
+      parts.push(doc);
+      return self;
+    },
+    build() {
+      flush();
+      return fill(parts);
+    },
+  };
+  return self;
+}
```
seiyab, line 28: **Possible alternative API**
```js
const result = buildFill(($) => {
  $.append("a");
  $.append("b");
  $.appendLine(line);
  $.append("c");
})
```
instead of
```js
const builder = fillBuilder();
builder.append("a");
builder.append("b");
builder.appendLine(line);
builder.append("c");
const result = builder.build();
```

src/document/utils/fill-builder.js
```diff
@@ -0,0 +1,30 @@
+import { fill } from "../builders.js";
+
+function fillBuilder() {
```
seiyab, line 3: It might be better to encourage using `fillBuilder()` instead of `fill()` for whole the repository since this should be safer.
If you agree, let's discuss on another discussion or issue.

src/language-markdown/printer-markdown.js
```diff
@@ -686,90 +683,6 @@ function printIgnoreComment(node) {
   }
 }
 
-/** @return {false | 'next' | 'start' | 'end'} */
-function isPrettierIgnore(node) {
```
seiyab, line 690: some functions went to util.

src/language-markdown/leading-hardlines.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {*} options
+ * @returns {Doc | null}
+ */
+function leadingHardlines(path, options) {
+  if (!shouldPrePrintHardline(path)) {
+    return null;
+  }
+  let length = 1;
+  if (shouldPrePrintTripleHardline(path)) {
+    length = 3;
```
seiyab, line 29: I can't find input that cover this path. I left it anyway.
https://app.codecov.io/gh/prettier/prettier/blob/main/src%2Flanguage-markdown%2Fprinter-markdown.js#L663

src/document/utils/flatten-fill.js
```diff
@@ -0,0 +1,49 @@
...
+        break;
+      default:
+        builder.append(doc);
+        break;
+    }
+  }
+
+  drain(doc);
+
+  return builder.build();
```
fisker, line 46: Will something like this work?

```js
  const parts = []
  
  // Check last part / parts.index is ODD-EVEN
  switch (getDocType(doc)) {
      case DOC_TYPE_FILL: {
        // concat or turn the last element to array and push the first
        // push the rest
        break;
      }
      default:
        builder.append(doc);
        break;
    }

return fill(parts)
```

src/language-markdown/printer-markdown.js
```diff
@@ -842,6 +754,20 @@ function printFootnoteReference(node) {
...
+ * @param {AstPath} path
+ * @param {*} options
+ * @param {*} print
+ * @returns {Doc}
+ */
+function printParagraph(path, options, print) {
+  const parts = [];
+  path.each(() => {
+    parts.push(print());
+  }, "children");
```
fisker, line 767:
```suggestion
  const parts = path.map(print, "children")
```

Maybe we should add an ESLint rule for this.

seiyab: Oh I wonder why I avoided `path.map()` though I love declarative style. I have thought I saw a comment like "map is deprecated" but it is just a phantom... Thank you to point it out.

src/document/utils/flatten-fill.js
seiyab: @fisker 
I'm sorry but I don't understand what does it mean / aim at.
Do you mean
- it can be faster?
- it can be more readable?
- prefer cleaner doc? (e.g. prefer `fill("a", line, "b")` to `fill(["a"], line, ["b"])`)
- recursion should be avoided?
- fillBuilder looks over-engineering?

I'm not sure whether I understand the pseudo code or not but possibly following case can be a counter-example.
input: `["a", fill(["b", line, "c", line]), "b"]`
expected output: `fill(join(line, [["a", "b"], ["c"], ["b"]]))`

fisker: It's simply, so we don't need `fillBuilder`. Can you find other places that we want a "flat" fill?

src/language-markdown/print-paragraph.js
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * @param {Doc[]} docArray
+   */
+  (function rec(docArray) {
+    for (const doc of docArray) {
+      switch (getDocType(doc)) {
+        case DOC_TYPE_FILL: {
+          const [head, ...rest] = getDocParts(doc);
+          parts.at(-1).push(head);
+          parts.push(...rest.map((doc) => [doc]));
```
seiyab, line 38: `.map((doc) => [doc]))` is just for type safety.

src/document/utils/flatten-fill.js
seiyab: @fisker 
Does [866b257](https://github.com/prettier/prettier/pull/15909/commits/866b257e4719b3ab1c5f27f7c78b731a3ca62f49) meet your request?

fisker:
@seiyab Does https://github.com/prettier/prettier/pull/15909/commits/be4b1c6d68b4b209d33a0cf893ba1b744d2708ed make sense to you ?

src/language-markdown/print-paragraph.js
```diff
@@ -0,0 +1,54 @@
...
+      const docType = getDocType(doc);
+      if (docType === DOC_TYPE_ARRAY) {
+        rec(doc);
+        continue;
+      }
+
+      let head = doc;
+      let rest = [];
+      if (docType === DOC_TYPE_FILL) {
+        [head, ...rest] = doc.parts;
```
fisker, line 43: I'm not sure about this part, do we need go deeper? `head` and docs in `rest` can also be `DOC_TYPE_ARRAY` or `DOC_TYPE_FILL`.

fisker: I'm going to run some test first, https://github.com/prettier/prettier-regression-testing/issues/435#issuecomment-1908654047

fisker:
Wow, we are breaking (or maybe improving) a LOT... https://github.com/prettier/prettier-regression-testing/issues/435#issuecomment-1908695983 Good opportunity to add more tests.

seiyab:
Maybe this PR makes embedded code and normal code consistent. I'll check them.

<details><summary>Memo about regression test</summary>

### content/files/en-us/learn/css/building_blocks/backgrounds_and_borders/index.md
✅ It get consistent
```diff
diff --git ORI/content/files/en-us/learn/css/building_blocks/backgrounds_and_borders/index.md ALT/content/files/en-us/learn/css/building_blocks/backgrounds_and_borders/index.md
index f9114358..dd09ac79 100644
--- ORI/content/files/en-us/learn/css/building_blocks/backgrounds_and_borders/index.md
+++ ALT/content/files/en-us/learn/css/building_blocks/backgrounds_and_borders/index.md
@@ -46,7 +46,8 @@ The CSS {{cssxref("background")}} property is a shorthand for a number of backgr
         105deg,
         rgb(255 255 255 / 20%) 39%,
         rgb(51 56 57 / 100%) 96%
-      ) center center / 400px 200px no-repeat,
+      )
+      center center / 400px 200px no-repeat,
     url(big-star.png) center no-repeat,
     rebeccapurple;
 }
```
[Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA6ARhAHgAmAHShx3QEMwBrAcwCcIBXKAEyUOOIBsBLKOUmgLS1STLghgAKNuxkBGAAwBWJnCoAaaTOI0q6CQCZFinIeOmcAehPyApAEocAZgCcNjUS3bdExbJyKANn8AdkscBVsHZwCbTXYHMHE4GhxE2GSwgBZ5eQAHXH0c-JwoCAEaOFy+GHcZehoOCXQuKgEAZxh+VFyoKgSklNLyyura9gr0ODAwUlz63I44AG5CAF8QNRAIXJguaDbkUH46AHcABX4EA5RSDhPSAE8DzfQacgo4GABlUgBbOAAMjw4MgAGa3NpwF5vSifL65cg8KjIGA0ehQkCQ35cFFojFwLBVGhcf6wW4AFWSUH4Ymu4I4kM2bSRiwAivQIPAwRCMQArNpYL4suDszkgpD0xkgACOHPgZzouWuIFIbQEvDgKiYGxAqNIXG4vQAwhBfr9SMgVRwODrmb1FgBBGCorjoejy5JA3jchkYgAWMF+HAA6n6uPA2gjEl8ruGuAA3cMPS1gNrPEDx9EASWY4i+YGJOwdzC+MAeix9UtydEhwbeuUt1bgkJo8ZBmx4LZgCtIVHNlYxCJoLZTaZ11Z4MGDXCYMD9yAAHPJNhVZVwKj2+xaJTzNp10NPZ-OkPpNvRIeTSOg6buQHBfpMmFqAaRevRe3AAGIQGjm51Iy1SHdCAQFWVYgA)

</details>

seiyab:
https://github.com/prettier/prettier/commit/be4b1c6d68b4b209d33a0cf893ba1b744d2708ed can some sense to me but it looks technically incorrect (but it can be safe for `printParagraph()`). Technically saying, removing `if (rest.length % 2 === 1) {{ parts.push([]); }` can cause bug. I'll fix the problem and type error.

src/language-markdown/print-paragraph.js
seiyab: I'm not very sure it. But probably `fill()` isn't nested in paragraph. If we want to go deeper, we should take care that we should go deeper only even (odd for `doc.parts`) index element.

seiyab:
Feel free to revert [e85797f](https://github.com/prettier/prettier/pull/15909/commits/e85797f53a2dc10b078510bbdb444534e7ab95a4) if you regard type safety as important.

fisker:
> Feel free to revert [e85797f](https://github.com/prettier/prettier/pull/15909/commits/e85797f53a2dc10b078510bbdb444534e7ab95a4) if you regard type safety as important.

`getDocParts` should not be used anymore. It's a legacy function should be removed.

fisker:
If you worry about edge cases,

1. odd length should be thrown
2. zero-length should be ignored, otherwise `undefined` will be pushed to doc, and it make sense to have zero-length fill, eg `fill(join(line, path.map(print, "children"))`

seiyab:
@fisker 
I've wrote tests https://github.com/prettier/prettier/pull/15985.

seiyab:
I see. I'll remove the edge case consideration. I have considered fill can have even length because printer cares even length:
https://github.com/prettier/prettier/blob/1ea2fd059a2ef8762f062e3540b9d48596c43b58/src/document/printer.js#L491-L498
(I consider "fill should alway have even length" is typo of "rest should alway have even length" or "fill should alway have odd length")

fisker:
The `printParagraph` part looks good and fixed lots of bugs, do you think we can extract this part to seperate PR?

seiyab:
@fisker 
Do you mean reverting some refactoring that doesn't affect behavior? Or reverting other bug fixes such as `printSentence()` so that only fix code block bug leaving wrong wrapping in sentences? For former separation, I think I can.
For latter one, I think it can cause a new minor bug. Changes in `printSentence()` and `shouldRemainTheSameContent` block are also parts of related bug fix (making `fill()` valid and harmonize with `flattenFill()` in `printParagraph()`). `nomalizeDoc()` have yielded invalid `fill()` in codeblock and paragraph. Reverting `printSentence()`, codeblock stays valid and some paragraph can get valid but some paragraph can be new invalid form. So if we separate PR for other changes but `printParagraph()`, I believe it’s better to release them in a same version.

seiyab:
@fisker 
I found two smaller step without refactoring. I pushed the steps into https://github.com/seiyab/prettier/commits/rearrange-history-of-15909/.
Should we start from a PR at the point of https://github.com/seiyab/prettier/commit/064f77df8ca0dc9e3a086169af7716971da6ff35 as the first step?

fisker:
> Should we start from a PR at the point of https://github.com/seiyab/prettier/commit/064f77df8ca0dc9e3a086169af7716971da6ff35 as the first step?

Yes, please

seiyab:
I opened https://github.com/prettier/prettier/pull/15993 🚀

seiyab:
Splitted into following PRs. Close this.
https://github.com/prettier/prettier/pull/15993
https://github.com/prettier/prettier/pull/16158

